PR #38399, opened 2025-03-03
feat: use Lefthook for pre-commit checks

bsmth:
### Description

This PR proposes that we use https://lefthook.dev/ instead of husky and lint-staged.

### Motivation

The motivation is mostly for performance reasons. Making changes to more than 50 files can be slow from an authoring perspective. The pre-commit tasks that we run (linting / formatting) can be done in parallel, and are faster with lefthook, based on testing.

## Documentation

https://lefthook.dev/configuration/index.html

## Troubleshooting

If there are problems removing husky, try running:

```bash
rm -rf ./node_modules
git config --unset core.hooksPath
yarn
```

PR #38718, opened 2025-03-19
feat: cache Prettier writes

bsmth:
### Description

Change occurrences of `prettier -w` to `prettier --write --cache`

### Motivation

Better performance between formatting runs. CI will probably not benefit much from this, but authors may.

See:

- [x] https://github.com/mdn/content/pull/38399


Josh-Cena (approved):
@mdn-bot is an interesting party to request reviews for. Let's try this anyway.

PR #38722, opened 2025-03-19
Add Compression Dictionary Transport docs

tunetheweb:
### Description

[Shared Compression Dictionary](https://datatracker.ietf.org/doc/draft-ietf-httpbis-compression-dictionary/) is a new standard to allow much smaller resource sizes by reusing either previous resources as dictionaries, or having a separate disctionary.

It is available using Brotli or the newer ZStd compression standards and has been [available in Chrome since 130](https://developer.chrome.com/release-notes/130#compression_dictionary_transport_with_shared_brotli_and_shared_zstandard)

This PR adds the relevant documentation.

### Motivation

This feature is poorly documented at the moment outside of specs and GitHub repos.

Chrome wrote and old blog post on it, but the syntax changed a bit when it was launched:
https://developer.chrome.com/blog/shared-dictionary-compression
Rather then redo that, we'd rather document on MDN and point people from that post to here.

### Additional details

https://chromestatus.com/feature/5124977788977152

BCD updates in https://github.com/mdn/browser-compat-data/pull/26257


wbamberg:
Thanks for contributing these great-looking docs! Would you consider making https://pr38722.review.mdn.allizom.net/en-US/docs/Glossary/Compression_dictionary_transport a guide page under https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides, instead of a glossary page? I think it would be a better fit there.

The sweet spot for glossary pages is things that are:
- not specific to one tech area (i.e. JS, CSS, HTML, HTTP, ...)
- a term that needs defining rather than a whole chunk of tech that needs walking through
- something we can cover in a fairly short page

So the use case for readers is really when they are reading something else and thinking "what does this term mean" and we can explain it relatively quickly and they can get back to what they were reading before.

This looks much more like a usage guide: how to use a set of web platform features to accomplish a thing.

This is a drive-by comment since the PR is already assigned to Hamish and Estelle, but since I have stuck my oar in I am happy to review the whole thing if they would like, or not, if they would prefer that :).

tunetheweb:
Done! I was wondering where exactly to put all this guidance, since I agree it was stretching the Glossary beyond it's intent, but also doesn't have a single API to put this. Guides is a perfect place for it.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+Compression Dictionary Transport allows the use of specifying a resource which can be used as a dictionary for future requests. This can either be a specific dictionary file, or an existing resource (for example, using `app.v1.js` as a dictionary when downloading `app.v2.js`). This can result in a huge savings in subsequent file download sizes and hence time. In the `app.vX.js` example, basically only the delta would need to be download and the common bytes could be referenced from the original `app.v1.js` file that is already downloaded.
+
+## Existing resource as a dictionary
+
+To use a resource as a dictionary the resource will include the {{HTTPHeader("Use-As-Dictionary")}} HTTP Header:
```
pmeenan, line 17:
```suggestion
To use a resource as a dictionary the resource will include the {{HTTPHeader("Use-As-Dictionary")}} HTTP response Header:
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+```http
+Use-As-Dictionary: match="/js/app.*.js"
+```
+
+When a resource is requested (for example, `app.v2.js`) that matches this header, the request will include a SHA-356 hash of the available disctionary in the {{HTTPHeader("Available-Dictionary")}} HTTP Header, along with a `dcb` or `dcz` {{HTTPHeader("Accept-Encoding")}} HTTP Header (for delta compression brotli or zstd as appropriate):
```
pmeenan, line 23:
```suggestion
When a subsequent resource is requested (for example, `app.v2.js`) that matches the specified pattern, the request will include a SHA-256 hash of the available disctionary in the {{HTTPHeader("Available-Dictionary")}} HTTP Header, along with `dcb` and/or `dcz` values in the {{HTTPHeader("Accept-Encoding")}} HTTP Header (for delta compression brotli or zstd as appropriate):
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+```http
+Accept-Encoding: gzip, br, zstd, dcb, dcz
+Available-Dictionary: :pZGm1Av0IEBKARczz7exkNYsZb8LzaMrV7J32a2fFG4=:
+```
+
+The server can then respond {{HTTPHeader("Content-Encoding")}} HTTP Header:
```
pmeenan, line 30:
```suggestion
The server can then respond with an appropriately-encoded response and content encoding in the {{HTTPHeader("Content-Encoding")}} HTTP Header:
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,83 @@
...
+Vary: accept-encoding, available-dictionary
+```
+
+An optional `id` can also be provided in the {{HTTPHeader("Use-As-Dictionary")}} header:
+
+```http
+Use-As-Dictionary: match="/js/app.*.js", id="dictionary-12345"
+```
+
+In which case it will be used in the future requests along with the SHA-256 hash of the dictionary enclosed in colons:
```
pmeenan, line 48:
```suggestion
In which case the value will be sent in future requests in the {{HTTPHeader("Dictionary-ID")}} header:
```

files/en-us/web/http/reference/headers/accept-encoding/index.md
```diff
@@ -46,6 +46,8 @@ Accept-Encoding: compress
 Accept-Encoding: deflate
 Accept-Encoding: br
 Accept-Encoding: zstd
+Accept-Encoding: cdb
```
pmeenan, line 49:
```suggestion
Accept-Encoding: dcb
```

files/en-us/web/http/reference/headers/accept-encoding/index.md
```diff
@@ -46,6 +46,8 @@ Accept-Encoding: compress
...
+Accept-Encoding: cdz
```
pmeenan, line 50: dcz

files/en-us/web/http/reference/headers/accept-encoding/index.md
```diff
@@ -65,6 +67,10 @@ Accept-Encoding: deflate, gzip;q=1.0, *;q=0.5
   - : A compression format that uses the [Brotli](https://en.wikipedia.org/wiki/Brotli) algorithm.
 - `zstd`
   - : A compression format that uses the [Zstandard](https://en.wikipedia.org/wiki/Zstd) algorithm.
+- `dcb`
+  - : A {{glossary("Compression Dictionary Transport")}} format that uses the [Brotli](https://en.wikipedia.org/wiki/Brotli) algorithm with a custom dictionary.
+- `dcz`
+  - : A {{glossary("Compression Dictionary Transport")}} format that uses the [Zstandard](https://en.wikipedia.org/wiki/Zstd) algorithm with a custom dictionary.
```
pmeenan, line 73: These should probably link to the compression dictionary spec because the `dcb` and `dcz` stream formats include a header that is defined in the spec (and not part of the brotli or ZStandard formats).

files/en-us/web/http/reference/headers/available-dictionary/index.md
```diff
@@ -0,0 +1,36 @@
...
+{{HTTPSidebar}}
+
+The HTTP **`Available-Dictionary`** header allows the browser to specify the best matching dictionary it has to allow the server to use {{glossary("Compression Dictionary Transport")}} for a resource request.
+
+## Example
+
+> [!NOTE]
+> See the [Compression Dictionary Transport guide](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) for more information.
+
+Clients can send an `Available-Dictionary` HTTP Header when they support `dcb` or `dcz` encodings. The header is a colon-surrounded SHA-256 {{glossary("Hash")}} of the dictionary contents:
```
pmeenan, line 17:
```suggestion
Clients can send an `Available-Dictionary` HTTP Header when they support `dcb` or `dcz` encodings. The header is a colon-surrounded base-64 encoded SHA-256 {{glossary("Hash")}} of the dictionary contents:
```

files/en-us/web/http/reference/headers/available-dictionary/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+```http
+Accept-Encoding: gzip, br, zstd, dcb, dcz
+Available-Dictionary: :pZGm1Av0IEBKARczz7exkNYsZb8LzaMrV7J32a2fFG4=:
+```
+
+## Specifications
+
+{{Specifications}}
```
pmeenan, line 26: Does this need to link to the IETF spec?

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -0,0 +1,43 @@
...
+> [!NOTE]
+> See the [Compression Dictionary Transport guide](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) for more information.
+
+An optional `id` can also be provided in the {{HTTPHeader("Use-As-Dictionary")}} header:
+
+```http
+Use-As-Dictionary: match="/js/app.*.js", id="dictionary-12345"
+```
+
+In which case the `id` will be used in the {{HTTPHeader("Dictionary-ID")}} HTTP Header of future requests, along with the SHA-256 hash of the dictionary enclosed in colons in the {{HTTPHeader("Available-Dictionary")}} HTTP Header:
```
pmeenan, line 23:
```suggestion
In which case the `id` will be used in the {{HTTPHeader("Dictionary-ID")}} HTTP Header of future requests:
```

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Use-As-Dictionary`** header lists the matching criteria that the {{glossary("Compression Dictionary Transport")}} dictionary can be used for, for future requests.
+
+## Directives
+
+- `match`
+  - : A string value that provides the {{domxref("URL Pattern API", "URL Pattern")}} to use for future request matching.
```
pmeenan, line 15:
```suggestion
  - : A string value that provides the {{domxref("URL Pattern API", "URL Pattern")}} to use for future request matching (regex patterns are not supported).
```

pmeenan (commented):
LGTM, just a few tweaks.

files/en-us/web/http/reference/headers/accept-encoding/index.md
tunetheweb:
```suggestion
Accept-Encoding: dcz
```

files/en-us/web/http/reference/headers/available-dictionary/index.md
tunetheweb: This will be picked up from the BCD change in https://github.com/mdn/browser-compat-data/pull/26257 once that's merged.

files/en-us/web/http/reference/headers/accept-encoding/index.md
tunetheweb: Updated in https://github.com/mdn/content/pull/38722/commits/865f0bd9ca87eb6e5eb78ca3b23c051e8ec5e527

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+Modern compression formats such as {{glossary("Brotli compression")}} and {{glossary("Zstandard compression")}} allow the use of dictionaries of frequently used data to further increase compression over just referencing those from within the file being compressed. Typically, for HTTP requests, this uses the pre-defined static dictionary included in that format (for example [the Brotli static dictionary is available in the source code](https://github.com/google/brotli/blob/master/csharp/org/brotli/dec/Dictionary.cs)).
+
+Compression Dictionary Transport allows the use of specifying a resource which can be used as a dictionary for future requests. This can either be a specific dictionary file, or an existing resource (for example, using `app.v1.js` as a dictionary when downloading `app.v2.js`). This can result in a huge savings in subsequent file download sizes and hence time. In the `app.vX.js` example, basically only the delta would need to be download and the common bytes could be referenced from the original `app.v1.js` file that is already downloaded.
```
wbamberg, line 13: It's possible that I'm just an idiot, but I think this guide would benefit from stepping back a bit and explaining more about compression dictionaries, shared compression dictionaries, and how and why a developer might want to use them. This guide is quite bare-bones: "these are the controls you need to twiddle in order to use this feature", assuming that someone already understands what the feature is and how at a use-case level we might want to use it.

I mean naively, someone might ask something like "if I want to provide a compression dictionary, what format does it need to be in?".

As I understand it, the story is something like this:

-------------

Compression algorithms are used in HTTP to reduce the size of resources downloaded over the network, reducing bandwidth cost and the time taken to load pages. HTTP compression algorithms work by finding redundancy in the source: for example, places where text like `function` is repeated. They then include just one copy of the redundant string, and replace occurrences of it in the resource with identifiers for the string. Since the identifiers are shorter than the string, the compressed version is shorter. Algorithms like Brotli achieve even greater efficiency by shipping with a dictionary containing commonly encountered strings, so you don't need any copies of them in the compressed resource (aside, I thought this: https://www.debugbear.com/blog/shared-compression-dictionaries#text-compression-with-and-without-a-shared-compression-dictionary-whats-the-difference was a very helpful diagram here).

Compression dictionary transport builds on this by enabling you to provide your own dictionary which is especially applicable to a particular set of resources, which the compression algorithm can then use as a source of strings when compressing and decompressing the resource.

For example, suppose your website uses a JavaScript library. You would typically load a specific version of the library, and might include the version name in the name of the library, like `<script src="my-library.v1.js">`. When the browser loads your page, it will fetch a copy of the library as a subresource.

If you then update to `v2` of the library, most of the library's code will probably have stayed the same. So you can greatly reduce the size of the download for `my-library.v2.js` by telling the browser to use `my-library.v1.js` as a compression dictionary for `my-library.v2.js`. Then all strings that are common between `v1` and `v2` don't need to be included in the download for `v2`, because the browser already has them. Most of the download size of `my-library.v2.js` is then just the delta between the two versions.

-------------

Is that more or less correct? If so, I think we ought to say something like that here.

tunetheweb: Yeah that's more or less correct. Will expand to basically include that with a few tweaks.

And those diagrams are from a Googler's presentations (link in that post) so we can include them as they are not owned by DebugBear.

tunetheweb: OK I've added that (including the diagram but as an inline code sample, so it's more easily editted in the future). Also added a few more details on formats, examples on how to create the resources, and some restrictions. PTAL.

tunetheweb:
BTW what does this mean?

image

Flaws (1)
Note! 8 documents with no flaws that don't need to be listed. 🎉

wbamberg:
> BTW what does this mean?
> image
> 
> Flaws (1) Note! 8 documents with no flaws that don't need to be listed. 🎉

In general, flaws are a specific set of errors that the build system checks for, but in this case the message is the rather opaque "Unknown: No generic content config found" which I'm going to assume we can ignore.

files/en-us/glossary/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Compression Dictionary Transport
+slug: Glossary/Compression_dictionary_transport
+page-type: glossary-definition
+---
+
+{{GlossarySidebar}}
+
+**Compression Dictionary Transport** is a way of using a shared compression dictionary, rather than the standard static dictionary in {{glossary("Brotli compression")}} or {{glossary("Zstandard compression")}}, to dramatically reduce the transport size of HTTP requests.
```
pmeenan, line 9:
```suggestion
**Compression Dictionary Transport** is a way of using a shared compression dictionary, rather than the standard static dictionary in {{glossary("Brotli compression")}} or {{glossary("Zstandard compression")}}, to dramatically reduce the transport size of HTTP responses.
```

files/en-us/web/http/guides/compression/index.md
```diff
@@ -76,6 +76,12 @@ As compression brings significant performance improvements, it is recommended to
 
 Apache supports compression and uses [mod_deflate](https://httpd.apache.org/docs/current/mod/mod_deflate.html); for Nginx there is [ngx_http_gzip_module](https://nginx.org/en/docs/http/ngx_http_gzip_module.html); for IIS, the [`<httpCompression>`](https://learn.microsoft.com/en-us/iis/configuration/system.webServer/httpCompression/) element.
 
+## Compression dictionary transport
+
+Modern compression formats such as {{glossary("Brotli compression")}} and {{glossary("Zstandard compression")}} allow the use of dictionaries of frequently used data to further increase compression over just referencing those from within the file being compressed. Typically, for HTTP requests, this uses the pre-defined static dictionary included in that format (for example [the Brotli static dictionary is available in the source code](https://github.com/google/brotli/blob/master/csharp/org/brotli/dec/Dictionary.cs)).
```
pmeenan, line 81:
```suggestion
Modern compression formats such as {{glossary("Brotli compression")}} and {{glossary("Zstandard compression")}} allow the use of dictionaries of frequently used data to further increase compression over just referencing those from within the file being compressed. Typically, for HTTP responses, this uses the pre-defined static dictionary included in that format (for example [the Brotli static dictionary is available in the source code](https://github.com/google/brotli/blob/master/csharp/org/brotli/dec/Dictionary.cs)).
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,169 @@
...
+[0:9]b[10:20]I am here[42:]
+```
+
+Where `[0:9]` refers to copying the 9 characters starting at character 0. Note this is a simple example to demonstrate the concept and the actual algorithms are more complex than this.
+
+Clients can then easily reverse the compression after download to recreate the original, uncompressed resource.
+
+Algorithms like {{glossary("Brotli compression")}} and {{glossary("Zstandard compression")}} achieve even greater efficiency by allowing the use of dictionaries of commonly encountered strings, so you don't need any copies of them in the compressed resource. These algorithms ship with a pre-defined, default dictionary that is used when compressing HTTP responses.
+
+Compression Dictionary Transport builds on this by enabling you to provide your own dictionary which is especially applicable to a particular set of resources. The compression algorithm can then use as a source of strings when compressing and decompressing the resource.
```
pmeenan, line 41:
```suggestion
Compression Dictionary Transport builds on this by enabling you to provide your own dictionary which is especially applicable to a particular set of resources. The compression algorithm can then reference it as a source of bytes when compressing and decompressing the resource.
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,169 @@
...
+
+Assuming the references from the previous example are included in that common dictionary, this could be further reduced to this:
+
+```Compressed
+[d1]a[d2]Hello World![d3]
```
pmeenan, line 46:
```suggestion
[d0:9]a[d10:20]Hello World![d42:46]
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+This could be compressed by replacing repeated strings with references to a previous location and number of characters, like in this example:
+
+```Compressed
+function a() {
+  console.log("Hello World!");
+}
+
+[0:9]b[10:20]I am here[42:]
```
pmeenan, line 32:
```suggestion
[0:9]b[10:20]I am here[42:46]
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,169 @@
...
+
+Algorithms like {{glossary("Brotli compression")}} and {{glossary("Zstandard compression")}} achieve even greater efficiency by allowing the use of dictionaries of commonly encountered strings, so you don't need any copies of them in the compressed resource. These algorithms ship with a pre-defined, default dictionary that is used when compressing HTTP responses.
+
+Compression Dictionary Transport builds on this by enabling you to provide your own dictionary which is especially applicable to a particular set of resources. The compression algorithm can then use as a source of strings when compressing and decompressing the resource.
+
+Assuming the references from the previous example are included in that common dictionary, this could be further reduced to this:
+
+```Compressed
+[d1]a[d2]Hello World![d3]
+[d1]b[d2]I am here[d3]
```
pmeenan, line 47:
```suggestion
[d0:9]b[d10:20]I am here[d42:46]
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
pmeenan: Just trying to make it clear that the dictionary is just a reference byte stream where `d1` might imply specific tokenization.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,169 @@
...
+
+For example, suppose your website uses a JavaScript library. You would typically load a specific version of the library, and might include the version name in the name of the library, like `<script src="my-library.v1.js">`. When the browser loads your page, it will fetch a copy of the library as a subresource.
+
+If you then update to v2 of the library, most of the library's code will probably have stayed the same. So sites can greatly reduce the size of the download for `my-library.v2.js` by telling the browser to use `my-library.v1.js` as a compression dictionary for `my-library.v2.js`. Then all strings that are common between v1 and v2 don't need to be included in the download for v2, because the browser already has them. Most of the download size of `my-library.v2.js` is then just the delta between the two versions.
+
+While text compression is incredibly effiction, creating files that are are often 25% the size of uncompressed files, delta compression allows for an order of magnitude more compression with [examples files of only 2% site](https://github.com/WICG/compression-dictionary-transport/blob/main/examples.md).
```
pmeenan, line 56:
```suggestion
While text compression is incredibly effiction, creating files that are are often 25% the size of uncompressed files, delta compression allows for an order of magnitude more compression with [examples files of only 2% size](https://github.com/WICG/compression-dictionary-transport/blob/main/examples.md).
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,169 @@
...
+<link rel="compression-dictionary" href="/dictionary.dat" />
+```
+
+This dictionary is then downloaded by the browser during idle time, and that response must include the {{HTTPHeader("Use-As-Dictionary")}} HTTP Header:
+
+```http
+Use-As-Dictionary: match="/js/app.*.js"
+```
+
+From here the process is similar to the previous example when a matching resources is requested.
```
pmeenan, line 117: Worth adding a Link HTTP header example?

```
Link: </dictionary.dat>; rel="compression-dictionary"
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Creating dictionary compressed responses
+
+Delta-compressed dictioresourcesnaries use the Brotli or ZStandard algothms, a custom dictionary, but also include a magic header and embedded dictionary hash.
```
pmeenan:
```suggestion
Delta-compressed dictionaries use the Brotli or ZStandard algothms with a custom dictionary, but also include a magic header and embedded dictionary hash.
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,169 @@
...
+
+These can be created dynamically, but it is often better to create these in advance at build time. This will require deciding how many delta-compressed versions to create — for the last version only, or for the last X versions.
```
pmeenan, line 123:
```suggestion
These can be created dynamically, but for static resources it can be better to create these in advance at build time. This will require deciding how many delta-compressed versions to create — for the last version only, or for the last X versions.
```

pmeenan (commented):
LGTM, mostly nits. Probably worth adding the `compression-dictionary` link type to the [Link HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Link) page too.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
pmeenan: Yeah, but it wasn't actually until the end of the file since you're still in the same file ;-)

tunetheweb: 🤦‍♂️

files/en-us/web/http/guides/compression_dictionary_transport/index.md
tunetheweb: Had spotted this one already myself!

files/en-us/web/http/guides/compression_dictionary_transport/index.md
tunetheweb: Done!

tunetheweb:
> LGTM, mostly nits. Probably worth adding the compression-dictionary link type to the [Link HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Link) page too.

There aren't any other examples on that page.

If you mean in the Browser Compatibility section at the bottom, that's down in BCD so will add it to https://github.com/mdn/browser-compat-data/pull/26257

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+```http
+Use-As-Dictionary: match="/js/app.*.js"
+```
+
+From here the process is similar to the previous example when a matching resources is requested.
+
+## Creating dictionary-compressed responses
+
+Dictionary-compressed (also known as delta-compressed) responses use the Brotli or ZStandard algothms, a custom dictionary, but also include a magic header and embedded dictionary hash.
```
wbamberg, line 127: > Dictionary-compressed (also known as delta-compressed)

I think it's always best to be explicit and consistent about definitions. This page sometimes calls them dictionary-compressed and sometimes delta-compressed. It would be better to choose just one. If we want to introduce the term "delta-compressed" as well, I think we should do so up where we describe dictionary-based compression, and we could say something like "Responses that are compressed using a dictionary are called _dictionary-compressed responses_ or sometimes _delta-compressed responses_.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+These can be created dynamically, but for static resources it can be better to create these in advance at build time. This will require deciding how many delta-compressed versions to create — for the last version only, or for the last X versions.
```
wbamberg, line 129: That aside I'm not super-sure what the first paragraph here is saying. We already know that they use Brotli or ZStandard, and we already know that they use a dictionary. Perhaps we could simplify?

```suggestion
Dictionary-compressed responses can be created dynamically, but for static resources it can be better to create these in advance at build time. This will require deciding how many delta-compressed versions to create — for the last version only, or for the last X versions.
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,176 @@
...
+### Delta-compressed Brotli (`dcz`) a resource
+
+Assuming the dictionary file is `dictionary.txt` and the file to be compressed is `data.text`, this will create a `dcb`-compressed file `data.txt.dcb` that includes the magic header and embedded dictionary hash:
+
+```bash
+echo -en '\x5e\x2a\x4d\x18\x20\x00\x00\x00' > data.txt.dcz && \
+openssl dgst -sha256 -binary dictionary.txt >> data.txt.dcz && \
+zstd -D dictionary.txt -f -o tmp.zstd data.txt && \
+cat tmp.zstd >> data.txt.dcz
+```
```
wbamberg, line 152: There's a lot of repetition here which makes it hard to read. Maybe:

````suggestion
Given a dictionary file named `dictionary.text` and a file to compress named `data.text`, the following Bash command will compress the file using Brotli, producing a compressed file named `data.txt.dcb`:

```bash
echo -en '\xffDCB' > data.txt.dcb && \
openssl dgst -sha256 -binary dictionary.txt >> data.txt.dcb && \
brotli --stdout -D dictionary.txt data.txt >> data.txt.dcb
```

Given the same input files, the following Bash command will compress the file using ZStandard, producing a compressed file named `data.txt.dcz`:

```bash
echo -en '\x5e\x2a\x4d\x18\x20\x00\x00\x00' > data.txt.dcz && \
openssl dgst -sha256 -binary dictionary.txt >> data.txt.dcz && \
zstd -D dictionary.txt -f -o tmp.zstd data.txt && \
cat tmp.zstd >> data.txt.dcz
```

Note that you will need {{glossary("OpenSSL")}} installed locally as well as Brotli or ZStandard.
````

File: files/en-us/web/http/guides/compression_dictionary_transport/index.md [hunk 26]
```diff
@@ -0,0 +1,176 @@
...
+
+## Restrictions
+
+Compression algorithms are at risk of security attacks, so there are a number of restrictions for Compression Dictionary Transport, including:
+
+- Dictionaries must be served on the same-origin as the resource using the dictionary.
+- Delta-compressed resources must be same-origin to the document origin, or follow the [CORS](/docs/Web/HTTP/Guides/CORS) rules, and so be requested with the [`crossorigin`](/docs/Web/HTML/Attributes/crossorigin) attribute and served with an appropriate {{HTTPHeader("Access-Control-Allow-Origin")}} header.
+- Dictionaries are bound by the usual HTTP Cache partitioning and so cannot be shared between origins even if they download the same resources. The dictionary will need to be downloaded again for each origin.
+
+Additionally, Compression Dictionary Transport could themselves becoming tracking vectors so browsers may restrict this feature when cookies are disabled or when other extra privacy protections are enabled.
```
wbamberg, line 162:
```suggestion
Additionally, dictionaries could themselves becoming tracking vectors so browsers may restrict this feature when cookies are disabled or when other extra privacy protections are enabled.
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,176 @@
...
+zstd -D dictionary.txt -f -o tmp.zstd data.txt && \
+cat tmp.zstd >> data.txt.dcz
+```
+
+## Restrictions
+
+Compression algorithms are at risk of security attacks, so there are a number of restrictions for Compression Dictionary Transport, including:
+
+- Dictionaries must be served on the same-origin as the resource using the dictionary.
+- Delta-compressed resources must be same-origin to the document origin, or follow the [CORS](/docs/Web/HTTP/Guides/CORS) rules, and so be requested with the [`crossorigin`](/docs/Web/HTML/Attributes/crossorigin) attribute and served with an appropriate {{HTTPHeader("Access-Control-Allow-Origin")}} header.
```
wbamberg, line 159:
```suggestion
- Dictionaries must same-origin with the resource using the dictionary.
- Delta-compressed resources must be same-origin with the document origin, or follow the [CORS](/docs/Web/HTTP/Guides/CORS) rules, and so be requested with the [`crossorigin`](/docs/Web/HTML/Attributes/crossorigin) attribute and served with an appropriate {{HTTPHeader("Access-Control-Allow-Origin")}} header.
```

I think we usually express this as "same-origin with"? But open to better alternatives...

wbamberg (commented):
OK thanks for your updates! This is a detailed review of just the guide page.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+For example, suppose your website uses a JavaScript library. You would typically load a specific version of the library, and might include the version name in the name of the library, like `<script src="my-library.v1.js">`. When the browser loads your page, it will fetch a copy of the library as a subresource.
+
+If you then update to v2 of the library, most of the library's code will probably have stayed the same. So sites can greatly reduce the size of the download for `my-library.v2.js` by telling the browser to use `my-library.v1.js` as a compression dictionary for `my-library.v2.js`. Then all strings that are common between v1 and v2 don't need to be included in the download for v2, because the browser already has them. Most of the download size of `my-library.v2.js` is then just the delta between the two versions.
+
+While text compression is incredibly effiction, creating files that are are often 25% the size of uncompressed files, delta compression allows for an order of magnitude more compression with [examples files of only 2% size](https://github.com/WICG/compression-dictionary-transport/blob/main/examples.md).
```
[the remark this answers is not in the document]
tunetheweb: I understand the concern, but think with the removal of the specific 2% example (which was a bit cherry picked!) it's less of a concern. Being an "order of magnitude" better is an important part of this technology and the fact it uses standards like Brotli and Zstd underneath (but with a better dictionary!) means it should be the "better" than those. If it was a closed-source, non-standard, or a completely new algorithm I'd more understand your concern.

So will go with your suggestion.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
tunetheweb: There's a few intents behind the first paragraph:

1) I don't think I have explicitly mentioned that it needs Brotli or ZStd compression, and it explains why there's two following sections under this section. In the Overview I've said that "Algorithms _like_ [Brotli compression](https://pr38722.review.mdn.allizom.net/en-US/docs/Glossary/Brotli_compression) and [Zstandard compression](https://pr38722.review.mdn.allizom.net/en-US/docs/Glossary/Zstandard_compression) achieve even greater efficiency by allowing the use of dictionaries" (added emphasis on _like_), and also briefly mentioned the Accept/Content-Encoding values for these, but until now I've not explicitly stated that only those two can be used for this. I want to be clear that things like `gzip` are not supported (since they don't allow for a separate dictionary) and that if a new compression standard comes out that also supports dictionaries (or there is one out there already!) then it's not automatically supported, since it's currently limited to just those two. At the very least we'd need to register new Accept/Content-Encoding values to add them to the standard.

2) It's Brotli and Zstd with a couple of twists: you need a magic header and the hash to be embedded in the file.

The middle point (you need a dictionary) is probably more obvious and was more used for the way I write the sentence ("You need A, B, and C"). Happy to drop that one.

So I'd like to keep the first paragraph or some variant of it to get those points across. I'm open a better way of wording this to make those intents clearer.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
tunetheweb: Looks good, but spotted another grammar issue:

```suggestion
Additionally, dictionaries could themselves become tracking vectors so browsers may restrict this feature when cookies are disabled or when other extra privacy protections are enabled.
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
tunetheweb: Checking out [the spec](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-compression-dictionary) both are used (with dictionary-compressed being more prevalent at 12 examples compared to 4 for delta-compressed).

I think dictionary-compressed is technically more accurate but delta-compressed is more often used in conversation.

@pmeenan has there been any discussion during standardisation on these two terms?

tunetheweb:
Thanks for the thorough review on the guidde @wbamberg ! Two outstanding comments: one for you and one for @pmeenan .

pmeenan:
> > LGTM, mostly nits. Probably worth adding the compression-dictionary link type to the [Link HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Link) page too.
> 
> There aren't any other examples on that page.
> 
> If you mean in the Browser Compatibility section at the bottom, that's done in BCD so will add it to [mdn/browser-compat-data#26257](https://github.com/mdn/browser-compat-data/pull/26257)

Sorry, I didn't look closely enough - ignore me.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
pmeenan: Delta-compressed is more applicable for the case where a static resource is used as a dictionary for a newer version of itself. Dictionary-compressed is the more generic case that covers both that and a stand-alone bag-of-bytes dictionary.

tunetheweb: OK I've cleaned up most instances of "delta-compressed" in https://github.com/mdn/content/pull/38722/commits/67d3a3d2ec4c33966064e8504e33f6cdf7be1032 leaving just one. Resolving.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
tunetheweb: Changed the first paragraph to:

> Dictionary-compressed responses can use either the Brotli or ZStandard algothms, with two extra requirements: they must also include a magic header and embedded dictionary hash.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,181 @@
...
+{{HTTPSidebar}}
+
+**Compression Dictionary Transport** is a way of using a shared compression dictionary to dramatically reduce the transport size of HTTP responses.
+
+## Overview
+
+Compression algorithms are used in HTTP to reduce the size of resources downloaded over the network, reducing bandwidth cost and the time taken to load pages. Lossless HTTP compression algorithms work by finding redundancy in the source: for example, places where text like the string `"function"` is repeated. They then include just one copy of the redundant string, and replace occurrences of it in the resource with references to that copy. Since the references are shorter than the string, the compressed version is shorter.
+
+> [!NOTE]
+> A previous attempt at this technology was called SDCH (Shared Dictionary Compression for HTTP) but was never widely supported and was removed in 2017. Compression Dictionary Transport is a better specified and more robust implementation with broader industry consensus.
```
tunetheweb, line 16: @wbamberg , since you've already reviewed the guide, not that I made this addition. There are still a few references to SDCH around the web (e.g. https://caniuse.com/sdch) which would be confused with this so I thought it best to explain the relationship.

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Use-As-Dictionary`** header lists the matching criteria that the {{glossary("Compression Dictionary Transport")}} dictionary can be used for, for future requests.
+
+## Directives
+
+- `match`
+  - : A string value that provides the {{domxref("URL Pattern API", "URL Pattern")}} to use for future request matching (regex patterns are not supported).
```
wbamberg, line 15:
```suggestion
  - : A string value containing a [URL Pattern](/en-US/docs/Web/API/URL_Pattern_API): only resources whose URLs match this pattern may use this resource as a dictionary.
```

("URL Pattern" is not an identifier so it should not get code formatting).

I find the URL Pattern docs really hard to understand tbh, but that's outside the scope of this PR.

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,95 @@
...
+- `match-dest`
+  - : An [Inner List of String values](https://www.rfc-editor.org/rfc/rfc8941#name-inner-lists) that provides a list of [Fetch request destinations](/docs/Web/API/Request/destination) for the dictionary to match.
```
wbamberg, line 17: I don't understand "Inner List of String values" and don't think we can give people a link that will make them understand ABNF. Is this actually effectively just a space-separated list of strings, with each string in quotres and the whole value enclosed in parentheses?

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,95 @@
...
+- `id`
+  - : A string value that specifies a server identifier for the dictionary. This will then be added as a {{HTTPHeader("Dictionary-ID")}} to resource requests where that dictionary can be used for.
```
wbamberg, line 19:
```suggestion
  - : A string value that specifies a server identifier for the dictionary. This ID value will then be added in the  {{HTTPHeader("Dictionary-ID")}} request header when the browser requests a resource which can use this dictionary.
```

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,95 @@
...
+- `type`
+  - : A token value that describes the file format of the supplied dictionary. Currently only `raw` is supported (which is the default) so this is more for future compatibility.
```
wbamberg, line 21:
```suggestion
  - : A string value that describes the file format of the supplied dictionary. Currently only `raw` is supported (which is the default) so this is more for future compatibility.
```

?

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Examples
+
+> [!NOTE]
+> See the [Compression Dictionary Transport guide](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) for more information.
```
wbamberg, line 26: Again put this in the general description, not with examples.

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+This uses a wildcard to match multiple versions of a file.
+
+### Destinations
+
+```http
+Use-As-Dictionary: match="/product/*", match-dest=("document")
+```
+
+This uses`match-dest` to ensure the dictionary is inly used for `document` requests so `/product/js/app.js` would not match and hence would not use this dictionary.
```
wbamberg, line 50:
```suggestion
This uses`match-dest` to ensure the dictionary is only used for `document` requests, so `/product/js/app.js` would not match and hence would not use this dictionary.
```

But also, isn't it more accurate to say "resource requests originating from a `<script>` tag"  - i.e. it's the way the resource is requests, not its extension, that determines `destination`?

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+```http
+Use-As-Dictionary: match="/product/*", match-dest=("document" "frame")
+```
+
+This would allow the dictionary to match both top-level documents and iframes.
+
+> [!NOTE]
+> As a [structured field](https://www.rfc-editor.org/rfc/rfc8941), the parameters should be space-separated, quoted strings — as shown above — and not comma-separated, which developers may be more used to.
```
wbamberg, line 59: This should be in Directives, not here. Also I would just describe the syntax directly: "The value of `match-dest` should be a space-separated list of quoted strings, enclosed in parentheses".

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Id
+
+```http
+Use-As-Dictionary: match="/product/*", id="dictionary-12345"
+```
+
+In which case the `id` will be used in the future requests in the {{HTTPHeader("Dictionary-ID")}} HTTP header along with the SHA-256 hash of the dictionary surrounded by colons in the {{HTTPHeader("Available-Dictionary")}} HTTP header:
```
wbamberg, line 67:
```suggestion
When `Use-As-Dictionary` includes an `id` directive, as in this example, the `id` value will be included in the {{HTTPHeader("Dictionary-ID")}} request header for resources that can use this dictionary. The resource request will also include the SHA-256 hash of the dictionary surrounded by colons in the {{HTTPHeader("Available-Dictionary")}} header:
```

wbamberg (commented):
Comments for the other pages.

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
tunetheweb: Yup! Changed it to:

> This uses`match-dest` to ensure the dictionary is only used for `document` requests so `<script src="/product/js/app.js">` resource requests for example would not match.

files/en-us/web/html/element/link/index.md
```diff
@@ -184,6 +184,10 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Global_attrib
   - : This attribute explicitly indicates that certain operations should be blocked on the fetching of an external resource. It must only be used when the `rel` attribute contains `expect` or `stylesheet` keywords. The operations that are to be blocked must be a space-separated list of blocking tokens listed below.
     - `render`: The rendering of content on the screen is blocked.
 
+- `compression-dictionary`
+
+  - : This attribute is used for {{glossary("Compression dictionary transport")}} to specify a dictionary that can be used for future requests.
+
```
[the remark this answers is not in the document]
tunetheweb: Good spot! Fixed.

files/en-us/web/http/reference/headers/accept-encoding/index.md
```diff
@@ -103,6 +109,7 @@ Accept-Encoding: br;q=1.0, gzip;q=0.8, *;q=0.1
 
 ## See also
 
+- [Compression Dictionary Transport guide](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport)
```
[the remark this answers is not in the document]
tunetheweb: Moved

files/en-us/web/http/reference/headers/available-dictionary/index.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Available-Dictionary
```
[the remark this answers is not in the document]
tunetheweb: Done.

files/en-us/web/http/reference/headers/content-encoding/index.md
```diff
@@ -60,6 +62,10 @@ Content-Encoding: deflate, gzip
   - : A format using the {{glossary("Brotli compression","Brotli")}} algorithm structure (defined in {{rfc(7932)}}).
 - `zstd`
   - : A format using the {{glossary("Zstandard compression","Zstandard")}} algorithm structure (defined in {{rfc(8878)}}).
+- `dcb`
+  - : A {{glossary("Compression Dictionary Transport")}} format that uses the [Dictionary-Compressed Brotli format](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-compression-dictionary#name-dictionary-compressed-brotl)
+- `dcz`
+  - : A {{glossary("Compression Dictionary Transport")}} format that uses the [Dictionary-Compressed Zstandard format](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-compression-dictionary#name-dictionary-compressed-zstan)
```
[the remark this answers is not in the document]
tunetheweb: Done. But changed the second "format" to algorithm.

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -0,0 +1,43 @@
...
+Use-As-Dictionary: match="/js/app.*.js", id="dictionary-12345"
+```
+
+In which case the `id` will be used in the {{HTTPHeader("Dictionary-ID")}} HTTP Header of future requests:
+
+```http
+Accept-Encoding: gzip, br, zstd, dcb, dcz
+Available-Dictionary: :pZGm1Av0IEBKARczz7exkNYsZb8LzaMrV7J32a2fFG4=:
+Dictionary-ID: "dictionary-12345"
+```
```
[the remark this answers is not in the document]
tunetheweb: SGTM! Done.

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Dictionary-ID
+slug: Web/HTTP/Reference/Headers/Dictionary-id
```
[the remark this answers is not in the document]
tunetheweb: Yes. Fixed.

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -0,0 +1,43 @@
...
+> [!NOTE]
+> See the [Compression Dictionary Transport guide](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) for more information.
+
+An optional `id` can also be provided in the {{HTTPHeader("Use-As-Dictionary")}} header:
+
+```http
+Use-As-Dictionary: match="/js/app.*.js", id="dictionary-12345"
+```
+
+In which case the `id` will be used in the {{HTTPHeader("Dictionary-ID")}} HTTP Header of future requests:
```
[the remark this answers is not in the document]
tunetheweb: Done.

files/en-us/web/http/reference/headers/available-dictionary/index.md
```diff
@@ -0,0 +1,36 @@
...
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Available-Dictionary`** header allows the browser to specify the best matching dictionary it has to allow the server to use {{glossary("Compression Dictionary Transport")}} for a resource request.
+
+## Example
+
+> [!NOTE]
+> See the [Compression Dictionary Transport guide](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) for more information.
```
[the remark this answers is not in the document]
tunetheweb: Moved.

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -0,0 +1,43 @@
...
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Dictionary-ID`** header used in {{glossary("Compression Dictionary Transport")}} to reference a dictionary in later requests.
+
+## Examples
+
+> [!NOTE]
+> See the [Compression Dictionary Transport guide](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) for more information.
```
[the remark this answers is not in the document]
tunetheweb: Moved.

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Dictionary-ID
+slug: Web/HTTP/Reference/Headers/Dictionary-id
+page-type: http-header
+browser-compat: http.headers.Dictionary-id
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Dictionary-ID`** header used in {{glossary("Compression Dictionary Transport")}} to reference a dictionary in later requests.
```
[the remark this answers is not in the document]
tunetheweb: Done.

files/en-us/web/http/reference/headers/available-dictionary/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+Clients can send an `Available-Dictionary` HTTP Header when they support `dcb` or `dcz` encodings. The header is a colon-surrounded base-64 encoded SHA-256 {{glossary("Hash")}} of the dictionary contents:
```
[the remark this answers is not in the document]
tunetheweb: Done.

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
tunetheweb: Yup. Changed it to that.

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
tunetheweb: Done.

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
tunetheweb: Removed.

files/en-us/web/http/reference/headers/accept-encoding/index.md
```diff
@@ -65,7 +67,10 @@ Accept-Encoding: deflate, gzip;q=1.0, *;q=0.5
   - : A compression format that uses the [Brotli](https://en.wikipedia.org/wiki/Brotli) algorithm.
 - `zstd`
   - : A compression format that uses the [Zstandard](https://en.wikipedia.org/wiki/Zstd) algorithm.
-- `identity`
```
wbamberg, line 68: This seems to be an error? Like the new items got spliced between the dt and the dd for `identity`?

files/en-us/web/http/reference/headers/available-dictionary/index.md
```diff
@@ -0,0 +1,46 @@
...
+
+{{HTTPSidebar}}
+
+The HTTP **`Available-Dictionary`** request header allows the browser to specify the best matching dictionary it has to allow the server to use {{glossary("Compression Dictionary Transport")}} for a resource request.
+
+See the [Compression Dictionary Transport guide](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) for more information.
+
+## Syntax
+
+Clients can send an `Available-Dictionary` Header when they support `dcb` or `dcz` encodings. The header is a colon-surrounded base-64 encoded SHA-256 {{glossary("Hash")}} of the dictionary contents.
```
wbamberg, line 16:
```suggestion
Clients can send an `Available-Dictionary` header when they support `dcb` or `dcz` encodings. The header is a colon-surrounded base-64 encoded SHA-256 {{glossary("hash")}} of the dictionary contents.
```

The platform will give you a flaw for lower-casing "hash" here, but let's ignore it.

files/en-us/web/http/reference/headers/available-dictionary/index.md
```diff
@@ -0,0 +1,46 @@
...
+
+```http
+Available-Dictionary: :<base64-hash>:
+```
+
+## Directives
+
+- `<base64-hash>`
+  - : A base-64 encoded SHA-256 {{glossary("Hash")}} of the dictionary contents.
```
wbamberg, line 25:
```suggestion
  - : A base-64 encoded SHA-256 {{glossary("hash")}} of the dictionary contents.
```

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -0,0 +1,55 @@
...
+## Syntax
+
+```http
+Dictionary-ID: "<string-identifier>"
+```
+
+## Directives
+
+- `<string-identifier>`
+  - : A string identifier
```
wbamberg, line 25:
```suggestion
  - : A string representing the dictionary's server-assigned ID.
```

?

files/en-us/web/http/reference/headers/use-as-dictionary/index.md
```diff
@@ -0,0 +1,103 @@
...
+// Multiple, in any order
+Content-Encoding: match="<urlpattern>", match-dest=("<destination1>")
+```
+
+## Directives
+
+- `match`
+  - : A string value containing a [URL Pattern](/en-US/docs/Web/API/URL_Pattern_API): only resources whose URLs match this pattern may use this resource as a dictionary.
+- `match-dest`
+  - : A space-separated list of strings, with each string in quotres and the whole value enclosed in parentheses, that provides a list of [Fetch request destinations](/docs/Web/API/Request/destination) for the dictionary to match.
```
wbamberg, line 31:
```suggestion
  - : A space-separated list of strings, with each string in quotes and the whole value enclosed in parentheses, that provides a list of [Fetch request destinations](/docs/Web/API/Request/destination) that requests must match if they are to use this dictionary.
```

(I think?) The dictionary doesn't match, the request does.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,181 @@
...
+
+```html
+<link rel="compression-dictionary" href="/dictionary.dat" />
+```
+
+Alternatively a dictionary can be included using the {{HTTPHeader("Link")}} header:
+
+```http
+Link: </dictionary.dat>; rel="compression-dictionary"
+```
```
wbamberg, line 124:
````suggestion
Sometimes an HTML document will want to provide a compression dictionary to the browser which isn't a resource that the browser is downloading anyway via an element such as a {{htmlelement("script")} tag. In this situation you have two options:

- Include a {{HTMLElement("link")}} element whose [`rel`](/en-US/docs/Web/HTML/Attributes/rel) attribute is set to `compression-dictionary`:

   ```html
   <link rel="compression-dictionary" href="/dictionary.dat" />
   ```

- Reference the dictionary using the {{HTTPHeader("Link")}} header:

    ```http
    Link: </dictionary.dat>; rel="compression-dictionary"
    ```
````

Something like this anyway. I think it's good to introduce this use case first, and the "Alternatively/alternatively" duplication was a bit confusing (the first is an alternative to the preceding section, but the second is an alternative to the first method).

**Review by wbamberg — COMMENTED** [review]
I still had a few more comments, sorry! But this is really close.

File: files/en-us/web/http/guides/compression_dictionary_transport/index.md (hunk 54 above)
Comment (tunetheweb) [reply files/en-us/web/http/guides/compression_dictionary_transport/index.md:124]: Sounds good, but gonna tweak the intro text a bit:

```suggestion
An HTML document can also provide a compression dictionary to the browser which isn't a resource that the browser is downloading anyway via an element such as a {{htmlelement("script")} tag. There are two methods to do this:

- Include a {{HTMLElement("link")}} element whose [`rel`](/en-US/docs/Web/HTML/Attributes/rel) attribute is set to `compression-dictionary`:

    ```html
    <link rel="compression-dictionary" href="/dictionary.dat" />
    ```

- Reference the dictionary using the {{HTTPHeader("Link")}} header:

    ```http
    Link: </dictionary.dat>; rel="compression-dictionary"
    ```
```

files/en-us/web/http/reference/headers/accept-encoding/index.md
tunetheweb: Oops. Added back in. Good spot.

tunetheweb:
Thanks again for all your efforts on this @wbamberg ! Let me know if you spot anything else.

files/en-us/web/http/reference/headers/content-encoding/index.md
wbamberg:
```suggestion
- `dcb`
  - : A format that uses the [Dictionary-Compressed Brotli algorithm](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-compression-dictionary#name-dictionary-compressed-brotl). See [Compression Dictionary Transport](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport).

- `dcz`
  - : A format that uses the [Dictionary-Compressed Zstandard algorithm](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-compression-dictionary#name-dictionary-compressed-zstan). See [Compression Dictionary Transport](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport).
```

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -0,0 +1,183 @@
...
+
+```http
+Accept-Encoding: gzip, br, zstd, dcb, dcz
+Available-Dictionary: :pZGm1Av0IEBKARczz7exkNYsZb8LzaMrV7J32a2fFG4=:
+Dictionary-ID: "dictionary-12345"
+```
+
+## Separate dictionary
+
+An HTML document can also provide a compression dictionary to the browser which isn't a resource that the browser is downloading anyway via an element such as a {{htmlelement("script")} tag. There are two methods to do this:
```
wbamberg, line 114:
```suggestion
An HTML document can also provide a compression dictionary to the browser which isn't a resource that the browser is downloading anyway via an element such as a {{htmlelement("script")}} tag. There are two methods to do this:
```

wbamberg (approved):
OK this looks great. Thank you for adding these docs, and for your patience with me!

PR #38723, opened 2025-03-20
Improve SVG entry in Image types

abacabadabacaba:
### Description

Improve the entry for SVG in https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Image_types.

### Motivation

The change improves the description of how SVG images can be used. It also adds a link to SVG documentation, helping with navigation.

### Additional details

None

None

### DIFF
--- a/files/en-us/web/media/guides/formats/image_types/index.md
+++ b/files/en-us/web/media/guides/formats/image_types/index.md
@@ -953,7 +953,7 @@ SVG can be used in web content in three ways:
 
 1. An {{SVGElement("svg")}} element can appear directly within the HTML. It can contain [SVG elements](/en-US/docs/Web/SVG/Reference/Element) to draw the image.
 2. An SVG image can be embedded in HTML using elements such as {{HTMLElement("iframe")}}, {{HTMLElement("object")}} and {{HTMLElement("embed")}}.
-3. It is possible to use SVG images anywhere where other image types can be used, including with the {{HTMLElement("img")}} and {{HTMLElement("picture")}} elements, the {{cssxref("background-image")}} CSS property, and so forth. However, there are [additional restrictions](/en-US/docs/Web/SVG/Guides/SVG_as_an_image) when SVG is used in this way.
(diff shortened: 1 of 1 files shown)


files/en-us/web/media/guides/formats/image_types/index.md
```diff
@@ -949,10 +949,11 @@ For instance, this example defines an drawing area with initial size 100 by 100
...
 ```
 
-SVG can be used in web content in two ways:
+SVG can be used in web content in three ways:
 
-1. You can directly write the {{SVGElement("svg")}} element within the HTML, containing [SVG elements](/en-US/docs/Web/SVG/Reference/Element) to draw the image.
-2. You can display an SVG image anywhere you can use any of the other image types, including with the {{HTMLElement("img")}} and {{HTMLElement("picture")}} elements, the {{cssxref("background-image")}} CSS property, and so forth.
+1. An {{SVGElement("svg")}} element can appear directly within the HTML. It can contain [SVG elements](/en-US/docs/Web/SVG/Reference/Element) to draw the image.
+2. An SVG image can be embedded in HTML using elements such as {{HTMLElement("iframe")}}, {{HTMLElement("object")}} and {{HTMLElement("embed")}}.
+3. It is possible to use SVG images anywhere where other image types can be used, including with the {{HTMLElement("img")}} and {{HTMLElement("picture")}} elements, the {{cssxref("background-image")}} CSS property, and so forth. However, there are [additional restrictions](/en-US/docs/Web/SVG/Guides/SVG_as_an_image) when SVG is used in this way.
```
pepelsbey, line 956:
```suggestion
3. It is possible to use SVG images anywhere where other image types can be used, including with the {{HTMLElement("img")}} element, the {{cssxref("background-image")}} CSS property, and so forth. However, there are [additional restrictions](/en-US/docs/Web/SVG/Guides/SVG_as_an_image) when SVG is used in this way.
```

I’d suggest keeping only `` element in this list, since `<picture>` is just a wrapper that supplies it with additional features and it’s `` that’s rendered on the page.

Looks great, otherwise!

pepelsbey (approved):
Looks good! Thank you :)

PR #38730, opened 2025-03-20
Add getPathData and setPathData methods

pepelsbey:
### Description

Documents new `getPathData` and `setPathData` methods shipped behind a flag in Firefox 136, enabled by default in Firefox 137.

### Additional details

Already in BCD and in all relevant release notes.

- https://bugzilla.mozilla.org/show_bug.cgi?id=1934525
- https://bugzilla.mozilla.org/show_bug.cgi?id=1945312

### Related issues and pull requests

- https://github.com/mdn/content/issues/37940

PR #38731, opened 2025-03-20
Update SVG path data methods in Firefox release notes

pepelsbey:
### Description

Only `getPathData` and `setPathData` methods were implemented for the `SVGPathElement` interface.

Goes along with the new docs: https://github.com/mdn/content/pull/38730

### Additional details

- https://bugzilla.mozilla.org/show_bug.cgi?id=1934525
- https://bugzilla.mozilla.org/show_bug.cgi?id=1945312

- https://github.com/mdn/content/issues/37940

### DIFF
--- a/files/en-us/mozilla/firefox/releases/136/index.md
+++ b/files/en-us/mozilla/firefox/releases/136/index.md
@@ -98,7 +98,7 @@ These features are newly shipped in Firefox 136 but are disabled by default. To 
   The {{svgelement("discard")}} SVG element allows developers to specify a trigger, such as the elapsed time since the SVG was loaded into DOM or the end of a particular animation, at which a specified element and its children should be removed from the DOM. This allows an SVG viewer to conserve memory by discarding animated elements that no longer needed.
   ([Firefox bug 1069931](https://bugzil.la/1069931)).
 - **SVG path API methods**: `dom.svg.pathSegment.enabled`.
-  The `getPathData()` and `setPathData()` methods of the {{domxref("SVGPathElement")}} interface are now supported. These methods provide a convenient way to work with SVG path data instead of parsing raw string data. ([Firefox bug 1934525](https://bugzil.la/1934525)).
+  The {{domxref("SVGPathElement/getPathData", "getPathData()")}}, {{domxref("SVGPathElement/setPathData", "setPathData()")}}, and {{domxref("SVGPathElement/getPathSegmentAtLength", "getPathSegmentAtLength()")}} methods of the {{domxref("SVGPathElement")}} interface are now supported. These methods provide a convenient way to work with SVG path data instead of parsing raw string data. ([Firefox bug 1934525](https://bugzil.la/1934525)).
 
 ## Older versions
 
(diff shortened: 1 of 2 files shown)


files/en-us/mozilla/firefox/releases/136/index.md
```diff
@@ -98,7 +98,7 @@ These features are newly shipped in Firefox 136 but are disabled by default. To
   The {{svgelement("discard")}} SVG element allows developers to specify a trigger, such as the elapsed time since the SVG was loaded into DOM or the end of a particular animation, at which a specified element and its children should be removed from the DOM. This allows an SVG viewer to conserve memory by discarding animated elements that no longer needed.
   ([Firefox bug 1069931](https://bugzil.la/1069931)).
 - **SVG path API methods**: `dom.svg.pathSegment.enabled`.
-  The `SVGPathSegment` interface now supports the `getPathData()`, `setPathData()`, and `getPathSegmentAtLength()` methods. These methods provide a convenient way to work with SVG path data instead of parsing raw string data. ([Firefox bug 1934525](https://bugzil.la/1934525)).
+  The `getPathData()` and `setPathData()` methods of the {{domxref("SVGPathElement")}} interface are now supported. These methods provide a convenient way to work with SVG path data instead of parsing raw string data. ([Firefox bug 1934525](https://bugzil.la/1934525)).
```
hamishwillee, line 101: What makes you think `SVGPathElement.getPathSegmentAtLength()` isn't implemented? 

I can see it in searchfox here https://searchfox.org/mozilla-central/source/dom/webidl/SVGPathElement.webidl#33 (the blames says added in [1934525](https://bugzilla.mozilla.org/show_bug.cgi?id=1934525)) - also the BCD collector detected it in 137 https://github.com/mdn/browser-compat-data/pull/26214/files#diff-6be09bc1af7d4b61152a3db70c81254f874fbd7eb7d012eab6db53a5e8599aceR1124

hamishwillee: @pepelsbey ^^^

pepelsbey:
@hamishwillee the `getPathSegmentAtLength` is back :)

hamishwillee (commented):
Thanks @pepelsbey. There is a comment from Estelle in your corresponding docs PR that we need to document the interface returned by `getPathSegmentAtLength` - it is an interface in the IDL so we probably should (even though it looks like it is really a dictionary).

My assumption is that you will want to link that from here - otherwise I would approve this.

pepelsbey:
> it is an interface in the IDL so we probably should (even though it looks like it is really a dictionary)

I’m not convinced it’s worth documenting, [see my comment in the PR](https://github.com/mdn/content/pull/38796/files#r2018845619). I don’t think we should document purely theoretical structures from the spec that browsers don’t really support/expose.

hamishwillee:
> > it is an interface in the IDL so we probably should (even though it looks like it is really a dictionary)
> 
> I’m not convinced it’s worth documenting, [see my comment in the PR](https://github.com/mdn/content/pull/38796/files#r2018845619). I don’t think we should document purely theoretical structures from the spec that browsers don’t really support/expose.

@pepelsbey As per my response, you are right. If this was actually returned as object of the indicated type in any browser we would have to document it as such and probably here explicitly note the omission here in the relnote. But that is not the case - it is actually implemented everywhere as an object - a defacto standard.

If you want to avoid these questions you might note this kind of thing in the discussion. Apologies if you did and I just missed it.

hamishwillee (approved):

PR #38636, opened 2025-03-14
move layout mode to glossary

estelle:
this is more a definition than a guide, so moved it to glossary

[#218](https://github.com/openwebdocs/project/issues/218)

PR #218, opened 2020-12-15
Normalize to "Specifications" (plural) heading

sideshowbarker:
This change replaces all instances of the heading text "Specification" (singular) with "Specifications" (plural).

The vast majority of our existing articles consistently have "Specifications" (plural), so otherwise, without this change, we have some articles which inconsistently contain "Specification" (singular).

That editorial inconsistency is suboptimal for end users, and even more suboptimal for any downstream tool developers who might want to programatically consume MDN content to extract specification data.

PR #38737, opened 2025-03-20
move layout mode to glossary

estelle:
replaces https://github.com/mdn/content/pull/38636

this is more a definition than a guide, so moved it to glossary

[#218](https://github.com/openwebdocs/project/issues/218)


chrisdavidmills (approved):
LGTM; let's do this.

PR #38740, opened 2025-03-20
Fix code block language from C++ to C in markdown

low-perry:
### Description
Changed the syntax highlighting in a Markdown code block from C++ to C for more accurate language representation

In three separate occasions the code snippets are in C but the block code says it's C++. I just corrected that.

### Motivation
This change ensures the code block is correctly highlighted according to its programming language, improving readability and accuracy for readers.

### DIFF
--- a/files/en-us/webassembly/guides/c_to_wasm/index.md
+++ b/files/en-us/webassembly/guides/c_to_wasm/index.md
@@ -30,7 +30,7 @@ This is the simplest case we'll look at, whereby you get emscripten to generate 
 
 1. First we need an example to compile. Take a copy of the following simple C example, and save it in a file called `hello.c` in a new directory on your local drive:
 
-   ```cpp
+   ```c
    #include <stdio.h>
 
    int main() {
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Hi, C++ is a syntax superset of C, so all valid C code can be marked as C++. We always use `cpp` as the language so people can easily write C++ syntax.

low-perry:
Hi @Josh-Cena , thank you for taking the time to respond to my pull request. I appreciate your explanation regarding the use of `cpp` as the language for markdown. I understand that C++ is a syntax superset of C, but the code in this instance was written in C. I'm curious—wouldn't it be clearer and more consistent to use C++ syntax directly when marking something as `cpp` ? Just wondering your thoughts on that.

Josh-Cena:
We have zero code blocks marked as `c` though. Rather than introducing another point of inconsistency, it seems easier to enforce using `cpp` everywhere, since it does not result in poorer rendering.

However, if you are interested, I'd definitely be open to reviewing a PR auditing our usage of `cpp` everywhere! For example, I see code blocks that are in C, C#, and GLSL all marked up as `cpp`. I think it would be great to at least get the latter two cases fixed.

low-perry:
Thanks again for clarifying.  I might look into submitting a PR to address cases like C# and GLSL being marked as `cpp`—I agree that fixing those is more important than the C case.

PR #38747, opened 2025-03-21
plural consistency

derekdilger:
goes with the above two 'references' words

### Description

plural consistency
### Motivation
consistency


dipikabh (approved):
Thanks!

PR #38753, opened 2025-03-21
Link from `CSSStyleSheet()` constructor to `Document.adoptedStyleSheets`

caugner:
### Description

Link from `CSSStyleSheet()` constructor to `Document.adoptedStyleSheets`

### Motivation

These are commonly used together.


hamishwillee (approved):
Yes, this makes sense. Thanks. 

FWIW we really should have a section on this stuff in our docs like the "Constructable Stylesheets" web.dev article.

PR #38761, opened 2025-03-22
`@keyframes` not `@keyframe`

Josh-Cena:
Little typo


hamishwillee (approved):
Thanks @Josh-Cena - good catch - that would be an easy typo to make.

PR #38762, opened 2025-03-22
Merge Browser compatibility for manifest.json page

Josh-Cena:
This page actually is blank now because it does not contain an H2 header which seems required in the new infra. I don't know how long it has been broken for, but instead of fixing it, I'm just removing this page because it has exactly the same information as the actual `manifest.json` page.


rebloor (approved):
@Josh-Cena nice spot, thanks

PR #38766, opened 2025-03-22
Fix set property invalid description of priority parameter

OlaviSau:
### Description
This change corrects the parameter description for CSSStyleDeclaration setProperty method.

### Motivation
These changes reflect the full list of accepted values and more accurately describe what happens when an invalid value is passed.

### Additional details
-
- I am not sure if there are related issues.

### DIFF
--- a/files/en-us/web/api/cssstyledeclaration/setproperty/index.md
+++ b/files/en-us/web/api/cssstyledeclaration/setproperty/index.md
@@ -29,8 +29,10 @@ setProperty(propertyName, value, priority)
     > **Note:** `value` must not contain `"!important"`, that should be set using the `priority` parameter.
 - `priority` {{optional_inline}}
   - : A string allowing the CSS priority to be set to important. Only the values listed below are accepted:
+
     - `"important"` (case-insensitive) for setting the property as `!important`;
     - `""`, `undefined`, or `null` for removing the `!important` flag if present.
+
     Anything else causes the method to return early and no change to happen (unless `value` is empty, in which case the property is removed regardless of the `priority` value). `false`, for example, is not a valid priority value.
 
 ### Return value
--- a/files/en-us/web/api/cssstyledeclaration/setproperty/index.md
+++ b/files/en-us/web/api/cssstyledeclaration/setproperty/index.md
@@ -29,12 +29,20 @@ setProperty(propertyName, value, priority)
     > **Note:** `value` must not contain `"!important"`, that should be set using the `priority` parameter.
 - `priority` {{optional_inline}}
 
-  - : A string allowing the "important" CSS priority to be set. If not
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
Sure, thanks

OlaviSau:
@Josh-Cena If the value is empty and the priority is invalid, no change will be applied even if the value is empty, so the changed description is invalid. Verified on chrome / firefox.

Josh-Cena:
@olavisau It works on Firefox. To try this, open the live example on this page. Change the stylesheet line to `const stylesheet = document.styleSheets[2];` (an annoying bug: https://github.com/mdn/yari/issues/12809) and change the `setRandomBorder` function to:


```js
function setRandomBorder() {
  boxParaRule.style.setProperty("border", "", "foo");
}
```

And press the "border" button. You will see that the border disappears.

The algorithm in the spec: https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setproperty also says if value is empty then priority is not validated at all.

That it does not work in Chrome or Safari may be a bug to be reported to these browsers, but Firefox is the only compliant browser for now.

OlaviSau:
> @OlaviSau It works on Firefox. To try this, open the live example on this page. Change the stylesheet line to `const stylesheet = document.styleSheets[2];` (an annoying bug: [mdn/yari#12809](https://github.com/mdn/yari/issues/12809)) and change the `setRandomBorder` function to:
> 
> ```js
> function setRandomBorder() {
>   boxParaRule.style.setProperty("border", "", "foo");
> }
> ```
> 
> And press the "border" button. You will see that the border disappears.
> 
> The algorithm in the spec: https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setproperty also says if value is empty then priority is not validated at all.
> 
> That it does not work in Chrome or Safari may be a bug to be reported to these browsers, but Firefox is the only compliant browser for now.

It seems it's actually a bit more complicated - the value is not validated if it is a string 😄 Try - boxParaRule.style.setProperty("border", "", false); instead of using "foo".

Josh-Cena:
I don't see a difference between `boxParaRule.style.setProperty("border", "", false)` and `boxParaRule.style.setProperty("border", "", "false")`. Both do nothing on Chrome and remove the border on Firefox. This is expected: any non-string parameter will get coerced to string per IDL semantics.

OlaviSau:
> I don't see a difference between `boxParaRule.style.setProperty("border", "", false)` and `boxParaRule.style.setProperty("border", "", "false")`. Both do nothing on Chrome and remove the border on Firefox. This is expected: any non-string parameter will get coerced to string per IDL semantics.

You're right, I did the test, but must have missed something as boxParaRule.style.setProperty("border", "", false) also removed the property on a new test that I did.

PR #38767, opened 2025-03-22
Replace link with Glossary Macro in web/privacy/guides/state_partitioning

cheeslide:
### Description

Replace link with Glossary Macro in web/privacy/guides/state_partitioning

### Motivation

replacing with macro help the maintainer understand and edit.

### DIFF
--- a/files/en-us/web/privacy/guides/state_partitioning/index.md
+++ b/files/en-us/web/privacy/guides/state_partitioning/index.md
@@ -91,7 +91,7 @@ As such, the following network APIs and caches are **permanently** partitioned b
 - Favicon Cache
 - Connection Pooling
 - Stylesheet Cache
-- [DNS](/en-US/docs/Glossary/DNS)
+- {{Glossary("DNS")}}
 - HTTP Authentication
 - [Alt-Svc](/en-US/docs/Web/HTTP/Reference/Headers/Alt-Svc)
 - Speculative Connections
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Hi, there is zero preference for one or the other. As a baseline we don't accept PRs that merely switch from one to the other, to avoid churn.

cheeslide:
> Hi, there is zero preference for one or the other. As a baseline we don't accept PRs that merely switch from one to the other, to avoid churn.

Thanks for your reply.

Glossary Macro will automatically get the right language for the link, and it is shorter.

Change the language code in link is the simple job that can totally be done by the platform, rather than search and replace when translating.

If they are exactly the same, why there is a macro for it?

Josh-Cena:
> If they are exactly the same, why there is a macro for it?

Because when MDN was a wiki and was authored with HTML, `<a>` links were a PITA to write. Now that we have Markdown links, both are essentially the same.

I don't understand the point about localization. You either don't have the localized version of the page at all, or you have an outdated version, or you copy over the latest English version and you translate it. On no occasion would you end up with an en-US link in your localized page. Sure it means you have to change one more thing in your link, but you have to translate everything else anyway, I fail to see how that is a significant undertaking.

We have people who really dislike macros (because it's not standard Markdown and not checkable by tooling), and people who really like macros (for example myself). To make both sides happy, we are not accepting PRs that change one to the other.

PR #38769, opened 2025-03-22
Replace http with https

lfrancke:
### Description

Replace two http links with https versions.

### Motivation

The image won't load otherwise


hamishwillee (approved):
Thanks very much @lfrancke - and welcome to MDN.

FWIW this is due to changes in [mixed content handling over time](https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content).

PR #37740, opened 2025-01-21
Add finalResponseHeadersStart docs

tunetheweb:
### Description

Add `finalResponseHeadersStart` docs coming to Chrome in Chrome 133.

This also updates the outdated TTFB docs.

### Motivation

New header starting to be made available in browsers to clarify the meaning of TTFB.

### Additional details

https://chromestatus.com/feature/5158830722514944

Also various discussions in W3C WebPerf WG:
- https://docs.google.com/document/d/1wL8Re4atVHVpfKTxhFfh8TNkkK4VCBbVJ-U70q5PWjo/edit?tab=t.0
- https://github.com/w3c/resource-timing/issues/345#issuecomment-2450475244

### Related issues and pull requests

See also:
- https://github.com/mdn/shared-assets/pull/39
- https://github.com/mdn/browser-compat-data/pull/25733

PR #38775, opened 2025-03-23
Fix TTFB definition

tunetheweb:
### Description

I made a mistake in #37740 and stated TTFB was `requestStart` instead of `responseStart`. Note the code snippet, just below this, correctly uses `responseStart`.

### Motivation

It's wrong

Error made in  #37740


Josh-Cena (approved):
Sure, thanks

PR #38778, opened 2025-03-23
Fixed typo and added await

erosman:
### Description
Fixed typo missing `s` in `getContext` in Examples
Added `await` as it returns a Promise


rebloor (approved):
Thanks @erosman

PR #38784, opened 2025-03-24
fix(accessibility docs): Remove "Best practices" section for roles that should not be used

dipikabh:
### Description

This PR removes the "Best practices" section from pages where there the advice is to not use the role, which is already conveyed via a note at the top of the page.

As a drive-by, also fixed the following problems:
- Incorrect placement of a section inside "Examples" (alertdialog_role) => Moved it into "Description" as an h3
- Inconsistent heading level for the section "Associated WAI-ARIA roles, states, and properties" => changed it to h3 where it was h2 (menuitem_role)
- Title casing for section titles => changed to sentence casing
- Typos with missing period (composite_role), stray word (roletype_role), and empty section (term_role)

### Motivation

Keep content relevant, non-confusing, and consistent


bsmth (approved):
All changes look great.

>advice is to not use the role, which is already conveyed via a note at the top of the page.

Maybe this could be a warning? Something to consider as a follow-up?

dipikabh:
> All changes look great.
> 
> > advice is to not use the role, which is already conveyed via a note at the top of the page.
> 
> Maybe this could be a warning? Something to consider as a follow-up?

Excellent suggestion, @bsmth. I've changed those notes to warnings in https://github.com/mdn/content/pull/38784/commits/55649e21bbf13d19618d474753bc2d9975bc979f.
Thanks for the review!

PR #38788, opened 2025-03-25
Clarify Object.assign() behavior with primitive targets

### LINKED ISSUE #659: Clarify primitive `target` behavior in `Object.assign()` documentation
### Proposal

The current documentation for `Object.assign()` doesn't explicitly state what happens when a primitive value is used as the `target` parameter. While the documentation mentions that primitives are wrapped to objects when used as source parameters, it doesn't clearly address the `target` parameter behavior.
When a primitive value is used as the `target` in `Object.assign()`, it gets automatically boxed into an object using the appropriate wrapper constructor (e.g., `Number`, `String`). Properties are then assigned to this wrapper object, and the wrapper object is returned.

For example:

```js
const number = Object.assign(3, { a: 1 });
console.log(number); // Number {3, a: 1}
console.log(typeof number); // object
console.log(number.a); // 1
```

I propose updating the documentation to:

- Clearly state in the "Parameters" section that primitives as targets are converted to objects
- Add an example showing this behavior in the "Primitives will be wrapped to objects" section

### Browser support

_No response_

### Tasks

- [ ]

### Dependencies

- [ ]

### Additional information

See: https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-object.assign

_No response_

codersjj:
### Description

- Clearly state in the "Parameters" section that primitives as targets are converted to objects
- Add an example showing this behavior in the "Primitives will be wrapped to objects" section

### Additional details

See: https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-object.assign

Fixes https://github.com/mdn/mdn/issues/659


Josh-Cena (approved):
Nice. Thank you!

PR #38791, opened 2025-03-25
SVGTransform switched to newer structure

yashrajbharti:
Restructured [SVGTransform Web API](https://developer.mozilla.org/en-US/docs/Web/API/SVGTransform) following the example [SVGLength Web API](https://developer.mozilla.org/en-US/docs/Web/API/SVGLength).


Josh-Cena (approved):
Nice, thanks

PR #38803, opened 2025-03-25
Remove ordinal number example from sup element documentation

dschuessler:
### Description

The [HTML5 specification of the `sup` element](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-sub-and-sup-elements) states that "In general, authors should use these elements only if the absence of those elements would change the meaning of the content."

The "Ordinal numbers" example in MDN stipulates that the `sup` element is a proper choice to mark up ordinal numbers in English like so: `4<sup>th</sup>`.

However, the absence of the elements would not change the meaning of the content. (`4<sup>th</sup>` has the same meaning as `4th`).

Hence, the example is not compliant with the spec (or at least its compliance is highly questionable). This MR removes the example.

### Motivation

The motivation is to keep the examples consistent with the specification.

### Additional details

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-sub-and-sup-elements


Josh-Cena:
The condition you quoted is used to clarify the first sentence:

> These elements must be used only to mark up typographical conventions with specific meanings,

The example given is "M<sup>lle</sup>", but in casual text it is also commonly written as "Mlle". Similarly, ordinal numbers themselves are a form of superior lettering, such as French "XXI<sup>e</sup> siècle" or English "21<sup>st</sup> century". The English version without superscript is also considered casual and perhaps inappropriate for professional typesettings, so in a way removing the superscript makes the typesetting incorrect.

This spec paragraph clearly allows any typographic convention to use `<sup>`. What it's meant to reject is "typographical presentation for presentation's sake", such as stylized icons/logos like $\LaTeX$. Ordinal numbers are not stylized text; they are proper content.

PR #38804, opened 2025-03-25
Make explanation of the cite element more spec compliant

dschuessler:
### Description

The [HTML5 specification of the cite element](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-sub-and-sup-elements) states that "The [cite](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-cite-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) the title of a work [...]. This can be a work that is being quoted or [referenced](https://html.spec.whatwg.org/multipage/dom.html#referenced) in detail (i.e., a citation), or it can just be a work that is mentioned in passing."

The MDN page in contrast defines the `cite` element more strictly as "The `<cite>` [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element is used to mark up the title of a cited creative work."

Since the work does not necessarily have to be cited as per spec, this MR changes the definition of the `cite` element.

### Motivation

The motivation is to keep the documentation consistent with the specification.


estelle (approved):

PR #38805, opened 2025-03-25
Change the second argument deleteCount of Array.prototype.toSpliced to skipCount.

yossydev:
### Description

I changed all instances of deleteCount to skipCount in the documentation for toSpliced.

### Motivation

In the first place, the ECMA-262 specification for toSpliced actually uses skipCount, not deleteCount.
The word “delete” implies mutation, which is why it was likely avoided. (https://github.com/tc39/ecma262/pull/2997#discussion_r1123772935)

To align with the ECMA spec, and because “skip” is semantically more accurate than “delete” in this context, I believe it would be better to update it.

### Additional details

Also, it seems that the term deleteCount was used during the initial proposal stage of this method.
(https://github.com/tc39/proposal-change-array-by-copy?tab=readme-ov-file#overview)

- ecma262: https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.tospliced
- mdn: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSpliced
- toSpliced add ecma262 pr: https://github.com/tc39/ecma262/pull/2997#discussion_r1123772935

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/array/tospliced/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/array/tospliced/index.md
@@ -30,7 +30,7 @@ toSpliced(start, skipCount, item1, item2, /* …, */ itemN)
 
 - `skipCount` {{optional_inline}}
 
-  - : An integer indicating the number of elements in the array to remove from `start`.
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
Nice. Thank you

PR #38816, opened 2025-03-26
`~` (bitwise NOT): Clarify 32-bit signed integer overflow and comparison with `Math.trunc`

lionel-rowe:
### Description

* Explain why `Math.trunc` is better for general-purpose integer truncation
* Tidy up wording around 32-bit signed integer overflow
* Move examples into side-by-side comparison table
* Split Number and BigInt details into own subsections

### Motivation

Current version just says "Do not use `~~x` to truncate numbers to integers; use `Math.trunc()` instead" without explaining why or what the tradeoffs might be.

### Additional details

Added wording "typically has near-identical performance in modern JavaScript engines" tested in Chrome, FF, Deno, Node, Bun:

```js
const randInts = crypto.getRandomValues(new Int32Array(1e4))
const randFloats = [...randInts].map((n) => n * Math.random() / 0xfff)
const all = [
    ...randInts,
    ...randFloats,
].sort(() => Math.random() - 0.5)

const fns = {
    ['~~'](n) {
        return ~~n
    },
    ['| 0'](n) {
        return n | 0
    },
    ['Math.trunc'](n) {
        return Math.trunc(n)
    },
    ['Math.floor'](n) {
        return Math.floor(n)
    }
}

for (const [name, fn] of Object.entries(fns)) {
    const start = performance.now()
    for (const n of all) fn(n)
    console.info(`${name}: ${performance.now() - start}`)
}
```

N/A


Josh-Cena:
I can't immediately see what you added because there's a lot of shuffling around, which I don't particularly like; this is the same format as used by https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR. We start by saying that number bitwise NOT discards the higher bits, then that bigint NOT does not discard, then gives a good mental model for what it does arithmetically (which is not critical for understanding its behavior and therefore placed further down), then gives a warning for a potential thing that people do (which is even less useful for the majority of users and therefore placed last). This order is very carefully thought to maximize utility for most readers wanting to systematically understand its behavior.

We are very careful not to give any performance hints, unless a feature is specifically designed for performance. Engines can optimize anything at any time, so these are incidental implementation reality, not something set in stone. If there are two solutions, we always prefer the one that is 100% correct, not the one that is 80% correct and fast, because our readers are not always going to consider if that "80% correct" is acceptable for their use case and may introduce subtle bugs.

lionel-rowe:
@Josh-Cena The motivation for the edit was that the advice about `Math.trunc` seemed very opinionated without justification. It seems odd to tell readers to "do x not y" without any explanation of what that decision means either way.

The only reason I mentioned performance is that a common justification for using bit manipulation tricks is performance, but it doesn't seem to have any measurable benefit here at all. Given that, I think there's very little reason to ever use `~~` outside of code golf or obfuscation (but it's not uncommon in the wild, so worth mentioning as a thing readers might encounter and want to know what it means).

The reason I shuffled the content around was because it seemed confusing to talk about numbers, then bigints, then 32-bit ints (which really refers to any number representable as such, plus all bigints), then numbers again. And I moved the truncation examples into a table for clarity, because the two examples are taking about the same phenomenon but were presented in a way that didn't make that obvious.

I'm not married to everything in the edit, but I definitely think a justification should be given for "use `Math.trunc` instead" (maybe without mentioning performance, as that's not central to the justification). As for the rest, I can propagate those edits to the other relevant articles if there are any you agree on, or otherwise revert.

Josh-Cena:
> It seems odd to tell readers to "do x not y" without any explanation of what that decision means either way.

The paragraph says this:

> Bitwise NOTing any number `x` twice returns `x` converted to a 32-bit integer. Do not use `~~x` to truncate numbers to integers; use [`Math.trunc()`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#using_bitwise_no-ops_to_truncate_numbers) instead. Due to using 32-bit representation for numbers, both `~-1` and `~4294967295` (2<sup>32</sup> - 1) result in `0`.

The first sentence tells you that `~~x` is a `toInt32` operation. Second sentence tells you it's not a truncation device, and if you want to do truncation, use `Math.trunc` instead. Third sentence gives you a concrete case where using it as a truncation device does not do what you want.

We are not really talking about numbers then bigints then numbers. We are first introducing its behavior, then giving a mental model of the behavior, then providing guidelines on its usage.

Josh-Cena:
If you think going from "it does `toInt32`" to "it doesn't do truncation" sounds like a non-sequitur, perhaps we can reorder this paragraph only?

> Bitwise NOTing any number `x` twice returns `x` converted to a 32-bit integer. It does _not_ always perform integer truncation. Due to using 32-bit representation for numbers, both `~-1` and `~4294967295` (2<sup>32</sup> - 1) result in `0`. Use [`Math.trunc()`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#using_bitwise_no-ops_to_truncate_numbers) instead for truncation.

lionel-rowe:
Ah I see the logic of the original now, but I still think the wording is a little unclear (doesn't state the logical connection).

> Bitwise NOTing any number x twice returns x converted to a 32-bit integer. It does not always perform integer truncation. Due to using 32-bit representation for numbers, both ~-1 and ~4294967295 (232 - 1) result in 0. Use [Math.trunc()](https://github.com/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#using_bitwise_no-ops_to_truncate_numbers) instead for truncation.

This is worse IMO — "It does not always perform integer truncation" is wrong — it _does_ always perform integer truncation, it's just that for numbers outside of `[-2147483648, 2147483647]` it _additionally_ truncates their most significant digits.

TBH I don't think `~~x` is even the best way to convert a number to a 32-bit integer — `| 0` seems both more explicit and potentially more performant as it's only one operation (not that the perf gain would matter in 99.9% of scenarios). So I think the explanation for `~~` should be more along the lines of "this is a thing you'll sometimes see in other people's code".

Josh-Cena:
> it _does_ always perform integer truncation,

"Integer truncation" means one very specific thing: removing the fractional part of the number. Any change to the integer part is not integer truncation.

Josh-Cena:
I swapped the paragraphs a bit because I can see how it does not progress most naturally. How does this look?

lionel-rowe:
> I swapped the paragraphs a bit because I can see how it does not progress most naturally. How does this look?

I still don't love the opinionated "use `Math.trunc`" (given that `| 0` et al aren't deprecated and are valid enough solutions in some cases), but I guess that's just my stylistic preference. And in general, [positive > negative wording](https://www.cypressmedia.net/articles/article/26/six_principles_of_technical_writing#positive), so "does not perform simple integer truncation" -> "additionally removes leading bits" (I just pushed this change). Looks good otherwise 👍

lionel-rowe:
1 last change — I think both range bounds should be specified. I've spelled the range bounds as plain decimal numbers, but spelling as powers of 2 would be fine too, up to you if you want to revert that.

Josh-Cena (approved):
Nice, thank you!

We have to recommend something that works 100% of the time, not something that seems to work 95% of the time but breaks on some valid input. We can't rely on an average reader's discretion to know that they won't run into that 5%. If you do have enough discretion you are welcome to ignore warnings.

PR #222, opened 2020-12-15
Fix typo: "effectst"

talsafran:
MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/EffectTiming/endDelay

#### What information was incorrect, unhelpful, or incomplete?

I found a lil typo!

```diff
- note, however, that many of the sequence effectst that will benefit most...
+ note, however, that many of the sequence effects that will benefit most...
```

#### Specific section or headline?

Second paragraph!

#### What did you expect to see?

The word "effects" 💥 

#### Did you test this? If so, how?

I will confess that I did not 😊

Issue #36508, opened 2024-10-26
All `ElementInternals` docs for `ARIAMixin` properties are incorrect


knowler:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals

### What specific section or headline is this issue about?

Instance properties included from ARIA

### What information was incorrect, unhelpful, or incomplete?

The issue I am reporting isn’t just for the page I’ve reported the issue for, but for all of the linked pages in the noted section as well.

`ARIAMixin` properties are implemented differently for `ElementInternals` than they are for `Element`, however, the docs describe them how they work for `Element` (i.e. as _reflecting_ content attributes). For `ElementInternals` the properties set default semantics or “implicit ARIA” for the element. There are a number of notes that correctly point that out, but the main content remains incorrect.

### What did you expect to see?

The docs should correctly describe how `ARIAMixin` properties for `ElementInternals` work: that they don’t reflect any attributes, but rather represent the default semantics of the custom element.

### Do you have any supporting links, references, or citations?

Relevant spec work (that does not describe how `ARIAMixin` should work for `ElementInternals`):

- https://html.spec.whatwg.org/multipage/custom-elements.html#elementinternals
- https://w3c.github.io/aria/#ARIAMixin

[The AOM explainer describes how `ElementInternals` works](https://wicg.github.io/aom/explainer.html#default-semantics-for-custom-elements-via-the-elementinternals-object) and notes that they do not reflect.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/api/elementinternals`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/elementinternals/index.md
* Last commit: https://github.com/mdn/content/commit/8b920a5e7567dcc9d642dfbd704b0ddbe2005d30
* Document last modified: 2024-07-27T01:21:26.000Z

</details>

PR #38819, opened 2025-03-26
New pages: ariaInvalid properties

estelle:
added these two features now supported in all browsers.

[#222](https://github.com/openwebdocs/project/issues/222)


files/en-us/web/api/elementinternals/ariainvalid/index.md
```diff
@@ -0,0 +1,49 @@
...
+  - : The element is in an invalid state because grammatical error was detected.
+- `"spelling"`
+  - : The element is in an invalid state because spelling error was detected.
+
+## Examples
+
+In this example the value of `ariaInvalid` is set to "true".
+
+```js
+this.internals_.ariaInvalid = "true";
```
Josh-Cena, line 34: While correct as per https://github.com/mdn/content/issues/36508, I think this is a poor example when taken out of context. At least we should show the full class declaration and where `internals_` is defined.

estelle: All the `elementInternals.ariaX` props have bad examples, .... or did. Hopefully this one is a bit better.
updating all those is a project for a different day

Josh-Cena (approved):
I think this is good enough to be merged. But the `ElementInternals` docs, as mentioned, is very incorrect. We would need to spend more time later to fix everything.

PR #38826, opened 2025-03-26
Remove links to images

Josh-Cena:
The images are already visible inline; there's no point to open them on their own page. Plus the SVG links don't work anyway.


bsmth (approved):
LGTM, thanks 👍🏻

PR #38831, opened 2025-03-27
fixed minor code bug at index.md

Marvellous890:
### Description
Direct copy any paste won't run due to canvas not defined. I fixed this by breaking the one-liner into two separate lines.

### Motivation
I am making these changes to inprove the quality of code snippets. I also don't want readers to waste time debugging why the code won't run because of a simple error.

### DIFF
--- a/files/en-us/web/api/canvas_api/tutorial/applying_styles_and_colors/index.md
+++ b/files/en-us/web/api/canvas_api/tutorial/applying_styles_and_colors/index.md
@@ -451,7 +451,7 @@ In this example we are creating a marching ants effect. It is an animation techn
 ```
 
 ```js
-const canvas = document.getElementById("canvas")
+const canvas = document.getElementById("canvas");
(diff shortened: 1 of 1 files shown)


wbamberg:
I'm not against splitting the code into two lines, but it's the same, isn't it? I'm not sure how this fixes anything. If there isn't an element with the ID `canvas`, then the variable is `null` and `getContext()` will be an error.

Josh-Cena:
@wbamberg later in this file it accesses `canvas.width`. However, this will happen to work in browsers because element IDs are added as global variables. Not sure we should encourage that though.

wbamberg:
> @wbamberg later in this file it accesses `canvas.width`. However, this will happen to work in browsers because element IDs are added as global variables. Not sure we should encourage that though.

Oh, I didn't see the reference to `canvas.width`. Like I say, I'm happy with the change, I just didn't understand, from the description, what it was fixing.

PR #38832, opened 2025-03-27
Rationalize glossary entries for hash functions

wbamberg:
We have three different glossary pages:

- https://developer.mozilla.org/en-US/docs/Glossary/Hash
- https://developer.mozilla.org/en-US/docs/Glossary/Cryptographic_hash_function
- https://developer.mozilla.org/en-US/docs/Glossary/Digest

In particular, the one on "Hash" gives bad advice for anyone doing crypto:

> It is commonly in the form of a 128-bit "fingerprint" or "message digest". Hashes are very useful for [cryptography](https://developer.mozilla.org/en-US/docs/Glossary/Cryptography)

There are AFAIK no algorithms commonly used for crypto that generate 128-bit hashes, and the ones that do, like MD5, are considered insecure. But many of our crypto pages link to https://developer.mozilla.org/en-US/docs/Glossary/Hash because, well, it's a natural thing to do.

So I've expanded the hash page to talk about crypto, and generally updated it, and deleted the specific "hashing for crypto" page. I also merged the "digest" page, which is a bit more tenuous I suppose but the page already duplicated much of the hash/crypto hash pages, and really if you want to know what a digest is you're going to have to know what a hash function is.


files/en-us/glossary/hash/index.md
```diff
@@ -6,8 +6,31 @@ page-type: glossary-definition
 
 {{GlossarySidebar}}
 
-The hash function takes a variable-length message input and produces a fixed-length hash output. It is commonly in the form of a 128-bit "fingerprint" or "message digest". Hashes are very useful for {{glossary("cryptography")}} — they ensure the integrity of transmitted data, and provide the basis for {{glossary("HMAC")}}, which enables message authentication.
+A hash function is a function that takes a variable-length input and produces a fixed-length output, also called a _digest_. Hash functions should be quick to compute, and different inputs should as far as possible produce different outputs (this is called _collision-resistance_).
```
hamishwillee, line 9: So the digest is the output right?

```suggestion
A hash function is a function that takes a variable-length input and produces a fixed-length output called a _digest_. Hash functions should be quick to compute, and different inputs should as far as possible produce different outputs (this is called _collision-resistance_).
```

wbamberg: That's all fair enough, yes. I agree that perhaps "Hash function" is a better name for the entry than "Hash" and I will make that change in this PR. But why capitalize "Hash" in your suggestion?

hamishwillee: No reason/mistake. 

Otherwise I am happy with the changes in this PR.

wbamberg:
I just realized we also have a page on [Signature/Security](https://developer.mozilla.org/en-US/docs/Glossary/Signature/Security) and one on [Digital Signature](https://developer.mozilla.org/en-US/docs/Glossary/Digital_signature), and I think this is my fault.

wbamberg:
> Error: Invalid 'from' URL for redirect: From-URL '/en-us/docs/glossary/hash' resolves to an existing folder at 'en-us/glossary/hash/index.md' for locale 'en-US'.

I don't know what this error means, as usual.

files/en-us/web/api/subtlecrypto/digest/index.md
```diff
@@ -9,7 +9,7 @@ browser-compat: api.SubtleCrypto.digest
 {{APIRef("Web Crypto API")}}{{SecureContext_header}}{{AvailableInWorkers}}
 
 The **`digest()`** method of the {{domxref("SubtleCrypto")}}
-interface generates a {{Glossary("digest")}} of the given data. A digest is a short
+interface generates a _digest_ of the given data, using the specified {{glossary("hash function")}}. A digest is a short
```
hamishwillee, line 12:
```suggestion
interface generates a _digest_ ("hash") of the given data, using the specified {{glossary("hash function")}}. A digest is a short
```

hamishwillee: Actually ignore. This wouldn't hurt, but nor is it needed. And if we were to do it here, we'd reasonably do it in a lot of the other cases too.

hamishwillee (approved):
@wbamberg Look good to me thanks.

Note that I'm not sure what of the MDN policy on synonyms. I mean you could reasonably have Digest, digest function, etc as entries with very thin links that point to the hash functions topic. 

I personally think your approach here is good, because we're not trying to be exhaustive in our glossary - there is no intent that people will be reading the glossary to get a good grip on common terms. So as the linked topic covers the synonyms (which this does), we're gold.

hamishwillee:
Yes, it is a bit of a can of worms. I think you'd probably have some standard header for "this is a synonym - see X for detail".

But I quite like our current "policy" - let's leave it to the internet dictionaries and wikipedia to cover the world, and we restrict ourself as far as possible to the subset of topics we link to.

PR #38833, opened 2025-03-27
pushState/replaceState called too frequently throws SecurityError

zcorpan:
See https://github.com/whatwg/html/pull/11169

### Motivation

The spec change no longer allows "ignore the call" behavior.


wbamberg (approved):
👍 thank you!

PR #38838, opened 2025-03-28
chore(uri): Get rid of 'reference' page type in URI section

bsmth:
### Description

__removals:__
* `reference` page type in fm config for URIs

__Changes:__
* `reference` to `uri-scheme` and `uri-component`

### Motivation

`reference` is not a good page type


Josh-Cena (approved):

bsmth:
Thanks!

PR #38845, opened 2025-03-28
improve description

theking2:
### Description
The long sentences and convolute writing makes the effect difficult to understand. The relation between inline-size and the more classic width or height is improved.

### Motivation
Many readers will have met `width` and `height` and are quite intimate with these. The suggested change makes the relation between those and `inline-size easier to understand.

### DIFF
--- a/files/en-us/web/css/inline-size/index.md
+++ b/files/en-us/web/css/inline-size/index.md
@@ -7,9 +7,7 @@ browser-compat: css.properties.inline-size
 
 {{CSSRef}}
 
-The **`inline-size`** [CSS](/en-US/docs/Web/CSS) property defines the horizontal or vertical size of an element's block, depending on its writing mode. It corresponds to either the {{cssxref("width")}} or the {{cssxref("height")}} property, depending on the value of {{cssxref("writing-mode")}}.
-
-If the writing mode is vertically oriented, the value of `inline-size` relates to the height of the element; otherwise, it relates to the width of the element. A related property is {{cssxref("block-size")}}, which defines the other dimension of the element.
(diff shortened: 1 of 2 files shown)


files/en-us/web/css/inline-size/index.md
```diff
@@ -7,9 +7,7 @@ browser-compat: css.properties.inline-size
 
 {{CSSRef}}
 
-The **`inline-size`** [CSS](/en-US/docs/Web/CSS) property defines the horizontal or vertical size of an element's block, depending on its writing mode. It corresponds to either the {{cssxref("width")}} or the {{cssxref("height")}} property, depending on the value of {{cssxref("writing-mode")}}.
-
-If the writing mode is vertically oriented, the value of `inline-size` relates to the height of the element; otherwise, it relates to the width of the element. A related property is {{cssxref("block-size")}}, which defines the other dimension of the element.
+The **`inline-size`** [CSS](/en-US/docs/Web/CSS) property defines the size of an element's block, depending on its writing mode. If the {{cssxref("writing-mode")}} is horizonal-tb or horizonal-bt it corresponds to the {{cssxref("width")}} if the writing mode vertical-lr or vertical-rl it corresponds to the {{cssxref("height")}} property.
```
Josh-Cena, line 10:
```suggestion
The **`inline-size`** [CSS](/en-US/docs/Web/CSS) property defines the size of an element's block, and whether it's the horizontal or vertical size depending on its writing mode. If the {{cssxref("writing-mode")}} is horizontal, it corresponds to the {{cssxref("width")}}; if the writing mode is vertical, it corresponds to the {{cssxref("height")}}. A related property is {{cssxref("block-size")}}, which defines the other dimension of the element.
```

Josh-Cena:
Could you propagate the same change to the `block-size` page as well?

Josh-Cena (approved):

PR #38848, opened 2025-03-28
Fix closing tag

mbergou:
Text was enclosed in `<strong>...<strong>` instead of `<strong>...</strong>`.

### Description

Very minor fix to example html

### Motivation

Copy/paste of the example html lead to error about invalid closing tag


Josh-Cena (approved):
Great catch. Thank you

PR #38853, opened 2025-03-29
Use `Object.hasOwn` on undefined and null property values

DanKaplanSES:
### Description

Use `Object.hasOwn` on undefined and null property values

### Motivation

- In my experience, the output of `Object.hasOwn` on `undefined`/`null` property values can be unexpected, so I wanted to document it in the example.


Josh-Cena:
Thanks. This is an interactive example, whose goal is to offer a quick taste of the API, not every aspect of its functionality. We demonstrate three things here: an existing own property, an existing property on the prototype, and a nonexistent property, which is more than enough to cover the whole algorithm anyway. I agree we should have examples demonstrating nullish properties, but they should go into the actual example section.

Josh-Cena:
Actually we already have examples for these below. Thanks for the PR!

PR #38854, opened 2025-03-29
Fix broken/redirected links

Josh-Cena:
After this round, nearly all inaccessible links on MDN should be fixed.


CONTRIBUTING.md
```diff
@@ -216,9 +216,9 @@ This will make it easier to submit a pull request for your changes.
...
 
-If you have a [local checkout](#forking-and-cloning-the-repository) of the repository and have [installed the dependencies](#preparing-the-project), or you are using [github.dev](https://github.dev), a pre-commit hook will be installed which automatically runs while making a commit. To save some headache and improve your work flow while authoring, you may wish to [configure your editor to automatically run Prettier](https://prettier.io/docs/editors.html). Alternatively, you may run `yarn fix:md` in the command line to manually format all Markdown files.
+If you have a [local checkout](#forking-and-cloning-the-repository) of the repository and have [installed the dependencies](#preparing-the-project), or you are using [github.dev](https://github.dev/), a pre-commit hook will be installed which automatically runs while making a commit. To save some headache and improve your work flow while authoring, you may wish to [configure your editor to automatically run Prettier](https://prettier.io/docs/editors.html). Alternatively, you may run `yarn fix:md` in the command line to manually format all Markdown files.
```
hamishwillee, line 221: This redirects me to https://github.dev/github/dev

CONTRIBUTING.md
```diff
@@ -391,7 +391,7 @@ Here are some things to keep in mind at this point:
 
 - Your pull request must be **reviewed and approved** before it's merged into the `main` branch.
 - **You do not need to request a review**; one or more reviewers will be selected for you automatically.
-- **It can be up to 48 hours** for merged pull requests to have their changes published to [MDN Web Docs](https://developer.mozilla.org).
+- **It can be up to 48 hours** for merged pull requests to have their changes published to [MDN Web Docs](https://developer.mozilla.org/).
```
hamishwillee, line 394: Does it matter? Either redirect me to https://developer.mozilla.org/en-US/

hamishwillee: Note, I don't disagree, just wondering whether the logic is just to add trailing backslash more or less everywhere there is no path

hamishwillee (approved):
This looks good IMO. Approving with an FMI?
@chrisdavidmills Note the scrimba link fixes.

CONTRIBUTING.md
Josh-Cena: > whether the logic is just to add trailing backslash more or less everywhere there is no path

Yeah that's normal browser behavior. `new URL("https://developer.mozilla.org").href` automatically adds a trailing slash. All these paths without a trailing slash, I surmise, come from authors manually writing the URL as opposed to copying it from the address bar.

> Either redirect me to https://developer.mozilla.org/en-US/

This would be different for people with different `Accept-Language` preferences.

CONTRIBUTING.md
Josh-Cena: Yeah, my tooling allows redirecting from `[url]` to `[url]/some/subpath`, in case they decide to change which route their root domain is an alias of.

hamishwillee:
Thanks @Josh-Cena !

PR #38855, opened 2025-03-29
Improve typesetting

Josh-Cena:
Put quotes inside code


files/en-us/mozilla/add-ons/webextensions/api/extension/index.md
```diff
@@ -21,7 +21,7 @@ Utilities related to your extension. Get URLs to resources packages with your ex
 - {{WebExtAPIRef("extension.lastError")}} {{deprecated_inline}}
   - : Set for the lifetime of a callback if an asynchronous extension API has resulted in an error. If no error has occurred, `lastError` will be {{jsxref("undefined")}}.
 - {{WebExtAPIRef("extension.inIncognitoContext")}}
-  - : `True` for content scripts running inside incognito tabs, and for extension pages running inside an incognito process. (The latter only applies to extensions with '`split`' `incognito_behavior`.)
+  - : `True` for content scripts running inside incognito tabs, and for extension pages running inside an incognito process. (The latter only applies to extensions with `incognito_behavior: "split"`.)
```
rebloor, line 24: I assume this is referring to the manifest key, for reference, see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/incognito
```suggestion
  - : `True` for content scripts running inside incognito tabs, and for extension pages running inside an incognito process. (The latter only applies to extensions with `"incognito": "split"` set in their manifest.json file.)
```

files/en-us/mozilla/add-ons/webextensions/api/i18n/index.md
```diff
@@ -14,7 +14,7 @@ See the [Internationalization](/en-US/docs/Mozilla/Add-ons/WebExtensions/Interna
 ## Types
 
 - {{WebExtAPIRef("i18n.LanguageCode")}}
-  - : A [language tag](https://www.rfc-editor.org/rfc/rfc9110.html#name-language-tags) such as `"en-US"` or "`fr`".
+  - : A [language tag](https://www.rfc-editor.org/rfc/rfc9110.html#name-language-tags) such as `"en-US"` or `"fr"`.
```
rebloor, line 17: @dotproto can you confirm that these are without quotes? (The Chrome documentation suggests they aren't. https://developer.chrome.com/docs/extensions/reference/api/i18n#type-LanguageCode.)
```suggestion
  - : A [language tag](https://www.rfc-editor.org/rfc/rfc9110.html#name-language-tags) such as `en-US` or `fr`.
```

files/en-us/mozilla/add-ons/webextensions/api/i18n/languagecode/index.md
```diff
@@ -7,7 +7,7 @@ browser-compat: webextensions.api.i18n.LanguageCode
 
 {{AddonSidebar}}
 
-A [language tag](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.10) such as `"en-US"` or "`fr`".
+A [language tag](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.10) such as `"en-US"` or `"fr"`.
```
rebloor, line 10: As per previous comment
```suggestion
A [language tag](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.10) such as `en-US` or `fr`.
```

rebloor (dismissed):

files/en-us/mozilla/add-ons/webextensions/api/i18n/index.md
Josh-Cena: Isn't it a string? In general string values are denoted with quotes

rebloor: It's because we specifically name these as "language tabs" these tabs or codes don't include the quotes, e.g. the language code is fr not "fr". I'm happy to defer to your judgment, however, and have removed my request for changes.

files/en-us/web/http/reference/headers/prefer/index.md
```diff
@@ -40,7 +40,7 @@ Prefer: <preference>
 
 - `respond-async`
   - : The client prefers asynchronous processing.
-    For example, the server might respond with {{httpstatus("202", "202 (Accepted)")}} response indicated that the request has been accepted, along with the {{httpheader("Location")}} header that has a URL that the client can use to monitor the state of the processing.
+    For example, the server might respond with {{httpstatus("202", "202 Accepted")}} response indicated that the request has been accepted, along with the {{httpheader("Location")}} header that has a URL that the client can use to monitor the state of the processing.
```
bsmth, line 43: These are okay seeing as the "reason phrase" part is recommended: https://www.rfc-editor.org/rfc/rfc9110.html#name-overview-of-status-codes

But it would be nice to have this uniform or even documented in writing guidelines

Josh-Cena: Yeah, currently my checker flags spaces in code unless it's in the form `\d+ [\w '-]+`, so parentheses are currently not allowed. I can make it require the parentheses if you want, but I might even want to simply remove the code formatting from the macro.

bsmth: > simply remove the code formatting from the macro

Right, that's a good suggestion, also. I wouldn't be opposed to it

bsmth:
> Anything else pending review?

Not from my side 👍

bsmth (approved):

PR #38862, opened 2025-03-30
Change example to clearly demonstrate tabular-nums

jruderman:
### Description

Change numbers in the example to clearly demonstrate `tabular-nums`.

### Motivation

With this change, readers can see that digits are constant-width only when using `tabular-nums`.


Josh-Cena (approved):
Sure, thanks

PR #38865, opened 2025-03-30
Remove unnecessary 'stream: true' argument from TextEncoder.encode call

### LINKED ISSUE #38824: WritableStreamAPI example typo
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_writable_streams#writing

Writing example

TextEncoder::encode method does not have parameters and returns Uint8Array

```javascript
const encoder = new TextEncoder();
const encoded = encoder.encode(message, { stream: true }); // does not have stream parameter
```

```javascript
sendMessages(["Hello, world."], writableStream);

async function sendMessage(messages, writableStream) {
  // defaultWriter is of type WritableStreamDefaultWriter
  const defaultWriter = writableStream.getWriter();
  const encoder = new TextEncoder();
  // TODO: close stream in case of encoder error
  const chunks = messages.map((it) => encoder.encode(it));
  chunks.forEach((chunk) => {       
       try {
           await defaultWriter.write(chunk);
           console.log("Chunk written to sink.");
       } catch(error) {
           console.error("Chunk error:", error);
       }
  });
  // Call ready again to ensure that all chunks are written
  //   before closing the writer.
  try {
     await defaultWriter.ready;
     console.log("All chunks written.");
  } catch(error) {
      console.error("Stream error:", error);
   }
}
```

https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encode

_No response_

low-perry:
### Description
Removes unnecessary argument from the TextEncoder.encode call. The encode method accepts only an [optional string argument](https://encoding.spec.whatwg.org/#interface-textencoder) .

### Motivation
correct technical issues on TextEncoder.encode method call

### Additional details
The example on [dom-examples](https://github.com/mdn/dom-examples/blob/main/streams/simple-writer/index.html) doesn't use that argument.
The [Encoding Api Spec](https://encoding.spec.whatwg.org/#interface-textencoder) says that the method only accepts an optional string arguement.

Fixes #38824


Josh-Cena (approved):
Nice, thank you

PR #38871, opened 2025-03-31
Express: Glitch hosting moved links to help centre

hamishwillee:
Glitch hosting used in the Express tutorial moved some of the existing helpdesk links behind a login. The same content is available at different (public) links if you go through the search function.

Thanks for the heads up @Josh-Cena


Josh-Cena (approved):
Great, thank you!

PR #38896, opened 2025-03-31
Update index.md

Stxxal:
Stacking context with a small 3D view demonstration

### Description
This update enhances the stacking context tutorial with a small 3D view demonstration by the readers' developer tools to help users visualize how elements are layered.

### Motivation
Provides readers a way to see the layers.


chrisdavidmills (commented):
@Stxxal Hi there!

I don't think this page addition is going to work. You are not demonstrating anything; you tell the reader to go and open a tool, but provide no guidance on how to use it.

The main issue, however, is that this tool doesn't exist across browser developer tools in general. Firefox used to have a 3D DOM viewer, but got rid of it years ago. From a Google search, I see that Edge has such a tool, but no other browsers do. So, the instruction is not going to work for the majority of users.

Stxxal:
Yes you are right! Thanks for the clarification. I will proceed with closing the pull request.

PR #38898, opened 2025-03-31
New page: HTMLElement.role property

estelle:
added this feature now supported in all browsers.

[#222](https://github.com/openwebdocs/project/issues/222)

removed the FF3.5 note as it is old and should be on the instance page not the index page.


files/en-us/web/api/element/role/index.md
```diff
@@ -0,0 +1,45 @@
...
+slug: Web/API/Element/role
+page-type: web-api-instance-property
+browser-compat: api.Element.role
+---
+
+{{ ApiRef("DOM") }}
+
+The **`role`** property of the {{domxref("Element")}} interface returns the explicily set [WAI-ARIA role](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles) for the element.
+
+All HTML elements have an implicit ARIA role, even if that role is [`generic`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/generic_role). This semantic association allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type. The `role` attribute is used to explicitly set the element's ARIA role, overriding the implicit role. For example, a {{htmlelement("ul")}}, which has an implicit [`list`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/list_role) role, might have [`role="treegrid"`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/treegrid_role) explicitly set. The `role` property reflects the explicitly set value of the `role` attribute &mdash; in this case `treegrid`; it does not return the element's implicit `list` role unless explicitly set.
```
Josh-Cena, line 13:
```suggestion
All HTML elements have an implicit ARIA role, even if that role is [`generic`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/generic_role). This semantic association allows tools to present and support interaction with the object in a manner that is consistent with user expectations about other objects of that type. The `role` attribute is used to explicitly set the element's ARIA role, overriding the implicit role. For example, a {{htmlelement("ul")}}, which has an implicit [`list`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/list_role) role, might have [`role="treegrid"`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/treegrid_role) explicitly set. The `role` property reflects the explicitly set value of the `role` attribute—in this case `treegrid`; it does not return the element's implicit `list` role unless explicitly set.
```

files/en-us/web/api/element/role/index.md
```diff
@@ -0,0 +1,45 @@
...
+title: "Element: role property"
+short-title: role
+slug: Web/API/Element/role
+page-type: web-api-instance-property
+browser-compat: api.Element.role
+---
+
+{{ ApiRef("DOM") }}
+
+The **`role`** property of the {{domxref("Element")}} interface returns the explicily set [WAI-ARIA role](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles) for the element.
```
wbamberg, line 11:
```suggestion
The **`role`** property of the {{domxref("Element")}} interface returns the explicitly set [WAI-ARIA role](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles) for the element.
```

wbamberg (approved):
👍 thanks Estelle!

PR #38900, opened 2025-03-31
New pages: HTMLFormElement.rel and .relList

estelle:
Created two new docs
added themn to the index page

part of [#222](https://github.com/openwebdocs/project/issues/222): add missing newly available baseline pages


Josh-Cena (approved):
This looks great, thanks

PR #38903, opened 2025-04-01
feat(writing style guide): Add guideline for using accessible language

dipikabh:
### Description

**Context**: https://github.com/mdn/content/pull/38846#discussion_r2021887803 highlighted the importance of avoiding spatial and directional language to keep our content accessible.

This PR adds a new section to define the guideline for using accessible language on MDN:

- Adds a new section titled "Use accessible language" within "General writing guidelines".
- Updates examples to use descriptive phrases instead of directional language.
- Refactors existing instances of "above", "below", "right", "left", and "here" on the page to align with the new guidance.
- Moves the guidance about "here" from [Cross-references (linking)](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide#cross-references_linking) section to the new section.

### Motivation

- To bring accessibility best practices to MDN, as also promoted in other major style guides (eg, [Microsoft](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/b/below), [Google](https://developers.google.com/style/word-list#below))
- To improve the usability of MDN documentation for all readers


files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
```diff
@@ -125,6 +125,31 @@ This last example of dealing with the problem is arguably better.
 Not only is it grammatically more correct, but removes some of the complexity associated with dealing with genders across different languages that may have wildly different gender rules.
 This solution can make translation easier for both readers and translators.
 
+### Use accessible language
+
+Avoid using spatial and directional words, such as "above", "below", "left", "right", or "here". These terms assume a specific visual layout, which may not apply to all users. They can also be unclear or misleading–especially for users relying on screen readers or those reading translated content, where directional language can be ambiguous or difficult to translate accurately. In responsive layouts, where the position of content can change depending on screen size, such directional references may become inaccurate. This kind of language can hinder accessibility and make it harder for all users to navigate or understand content.
```
hamishwillee, line 130: These arguments are IMO false logic. They are true for HTML authors, but very unlikely to be true for any content generated from markdown.

- ~99%~ 99.99% of users will have a layout where "above" and "below" will work.
- "Here" is also a great shorthand for "this section". 
- Right/left are problematic in HTML land, but when was the last time anyone used them in markdown?

hamishwillee: Don't get me wrong, I prefer "the following", but I think we're making a lot of potential work for relatively little improvement, if any.

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
```diff
@@ -125,6 +125,31 @@ This last example of dealing with the problem is arguably better.
...
+
+Instead, use descriptive phrases that clearly identify the section, concept, or element being referenced. Refer to sections by their titles or headings, and refer to examples or code snippets by what they demonstrate or contain.
+
+For example:
+
+- **Correct**: "Refer to the [Accessibility](/en-US/docs/Web/CSS/gradient/repeating-conic-gradient#accessibility) section later on this page."
+- **Incorrect**: "Refer to the Accessibility section below."
+
+- **Correct**: "In the following code example, we animate a circle using CSS transitions."
+- **Incorrect**: "In the code example below, we animate a circle."
```
hamishwillee, line 140: The actual improvement here is "using CSS transitions.". The "below" bit is irrelevant. Find me one case where it won't be below. I dare you :-).

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
```diff
@@ -125,6 +125,31 @@ This last example of dealing with the problem is arguably better.
...
+
+- **Correct**: "This concept is explained in the earlier section titled Creating a media query."
+- **Incorrect**: "This concept is explained in the section above."
```
hamishwillee, line 143: The "correct" case would suck if the concept is actually immediately above, while the incorrect case would make heaps of sense.

If it is a long way above, then yes, you might do this, but why not link then?

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
```diff
@@ -125,6 +125,31 @@ This last example of dealing with the problem is arguably better.
...
+
+Additionally, avoid using vague link text like "Click here" or "Read this article". Descriptive link text provides better context for all readers and improves the experience for users of assistive technologies.
```
hamishwillee, line 145: This point is useful.

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
```diff
@@ -136,7 +161,7 @@ The following checklist is good to keep in mind while writing and reviewing cont
 - **Ensure that pages aren't too similar**: If the content on different pages is similar textually, search engines will assume that the pages are about the same thing even if they aren't.
   For example, if an interface has the properties `width` and `height`, it's easy for the text to be surprisingly similar on the two pages documenting these two properties, with just a few words swapped out and using the same example. This makes it hard for search engines to know which is which, and they wind up sharing page rank, resulting in both being harder to find than they ought to be.
 
-  It's important, then, to ensure that every page has its own content. Here are some suggestions to help you accomplish that:
+  It's important, then, to ensure that every page has its own content. The following suggestions can help you accomplish that:
```
hamishwillee, line 164: I see this and think - "no better".

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
```diff
@@ -300,16 +325,16 @@ Do not use "curly" quotes and quotation marks. On MDN Web Docs, we only use stra
...
 
-The list below describes some of the common situations where we need to be aware of the comma usage rules:
+The following list describes some of the common situations where we need to be aware of the comma usage rules:
 
 - **After introductory clauses**: An introductory clause is a dependent clause, usually found at the beginning of a sentence. Use a comma after an introductory clause to separate it from the following independent clause.
 
   - Example 1:
-    - **Correct**: "In this example, you will see how to use a comma."
-    - **Incorrect**: "In this example you will see how to use a comma."
+    - **Correct**: "In this example, you learn how to use a comma."
```
hamishwillee, line 333: Why? Is this grammar correct?

hamishwillee (commented):
@dipikabh I've added some comments. The update is well done, I just would find it irritating if someone "fixed" my text with these changes.

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
hamishwillee:
```suggestion
    - **Correct**: "In this example, you will learn how to use a comma."
```

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
dipikabh: I'm not sure I follow why we're looking at HTML and markdown authors.

What I am focusing on is the readers of MDN and depending on the devices they use to access our website, "above" and "below" can be vague or unhelpful, like for people using screen readers.

I realize we're adding an overhead for ourselves, but I think it would be for a good cause. We recommend using the Microsoft style guide, and by making this change, we'd making our content more consistent with that style guide (and also more accessible).

https://developers.google.com/style/accessibility#document-rendering:

> "Don't use directional language to orient the reader, such as above, below, or right-hand side. This type of language doesn't work well for accessibility or for localization reasons. For example, what's on the right side for left-to-right languages appears on the left side for right-to-left languages.
Don't use directional language to refer to a position in a document. For example, the text isn't below if it's being read by a screen reader. Instead, use earlier, preceding, or following."

We can add more examples of words that can be used as alternatives so it won't feel restrictive, just a switch in our writing habits.

Let me know if I'm missing something or have missed your point.

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
```diff
@@ -125,6 +125,31 @@ This last example of dealing with the problem is arguably better.
...
+### Use accessible language
+
+Avoid using spatial and directional words, such as "above", "below", "left", "right", or "here". These terms assume a specific visual layout, which may not apply to all users. They can also be unclear or misleading–especially for users relying on screen readers or those reading translated content, where directional language can be ambiguous or difficult to translate accurately. In responsive layouts, where the position of content can change depending on screen size, such directional references may become inaccurate. This kind of language can hinder accessibility and make it harder for all users to navigate or understand content.
+
+Instead, use descriptive phrases that clearly identify the section, concept, or element being referenced. Refer to sections by their titles or headings, and refer to examples or code snippets by what they demonstrate or contain.
+
+For example:
+
+- **Correct**: "Refer to the [Accessibility](/en-US/docs/Web/CSS/gradient/repeating-conic-gradient#accessibility) section later on this page."
+- **Incorrect**: "Refer to the Accessibility section below."
```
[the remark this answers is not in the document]
dipikabh: Yeah, your second version works perfectly as well. The example is trying to illustrate in case someone really wants to emphasize the location of the section on the page.

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
dipikabh: I agree, we should keep that part the same
```suggestion
- **Incorrect**: "In the code example below, we animate a circle using CSS transitions."
```
I know what you mean – I don't think "below" is wrong, but "following" is slightly more precise and it avoids accessibility issues. To me:
"below" = visual positioning
"following" = part of the flow

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
dipikabh: Good point. What do you think about including both those examples like so:

- **Correct**: "This concept is explained in a previous section called [Creating a media query](url)." (for a section further up the page)
- **Correct**: "This concept is explained in the previous section about creating media queries." (for a section that immediately precedes this one)

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
dipikabh: Thanks for the correction

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
dipikabh, line 143:
```suggestion
- **Correct**: "This concept is explained in a previous section called [Creating a media query]()." (for a section further up the page)
- **Correct**: "This concept is explained in the previous section about creating media queries." (for a section that immediately precedes this one)
- **Incorrect**: "This concept is explained in the section above."
```

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
dipikabh: I don't have a strong argument against "here" except that avoiding it can help us be more direct and clear. Plus I imagine the result will be slightly easier to localize.

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
```diff
@@ -146,7 +171,7 @@ The following checklist is good to keep in mind while writing and reviewing cont
 
 - **Ensure that pages aren't too short**: If the content on a page is too little (called "thin pages" in SEO parlance), search engines will not catalog such pages accurately (or at all). Overly-short content pages are hard to find. As a guiding principle, ensure that pages on MDN Web Docs are not shorter than around 300 words or so. Don't artificially inflate a page, but treat this guideline as a minimum target length when possible.
 
-  Here are some basic guidelines to help you create pages that have enough content to be properly searchable without resorting to cluttering them up with unnecessary text:
+  These basic guidelines can help you create pages that have enough content to be properly searchable without resorting to cluttering them up with unnecessary text:
```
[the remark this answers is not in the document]
dipikabh: Yeah, precisely, so maybe this doesn't fall so much in the accessibility net but is just making the text more clear and direct

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
hamishwillee: Yes, you are missing my point!

The advice _only_ makes sense to HTML authors. In HTML or word or whatever it is common to arrange your content in floating boxes, divs, whatever. Since the arrangement of these depends on your screen size you shouldn't make assumptions about layout.

But it is wrong markdown authors.
When you're working with markdown (that does not include HTML) the flow is always linear. That is true for screen readers, mobile, or whatever.

Yes, there might be a case or two in the whole of the thousands of pages on MDN where this advice makes some difference, but it is definitely not worth any change to our writing.

dipikabh:
Thanks a lot for the review, @hamishwillee. Sorry it took me a while to get to your feedback.

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
hamishwillee: I like the approach of separating the instruction based on how far away the link is. But 
- as per https://github.com/mdn/content/pull/38903#discussion_r2038627534 it really can only be "above", so I'd say that.
- If you're providing a link you don't need to say above or below.

So

- **Correct**: "This concept is explained in [Creating a media query](url)." (for a section further up the page)
- **Correct**: "This concept is explained in the section above" (for a section that immediately precedes this one)

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
hamishwillee: Yeah, it's a little more readable. Again, I have a preference for it, but it's cruel to enforce the style.

hamishwillee:
Hope my follow on explanation helps. I'm away for two weeks from now, so there won't be an instant response from me either sorry!

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
```diff
@@ -300,16 +325,16 @@ Do not use "curly" quotes and quotation marks. On MDN Web Docs, we only use stra
...
-The list below describes some of the common situations where we need to be aware of the comma usage rules:
+The following list describes some of the common situations where we need to be aware of the comma usage rules:
 
 - **After introductory clauses**: An introductory clause is a dependent clause, usually found at the beginning of a sentence. Use a comma after an introductory clause to separate it from the following independent clause.
 
   - Example 1:
-    - **Correct**: "In this example, you will see how to use a comma."
-    - **Incorrect**: "In this example you will see how to use a comma."
+    - **Correct**: "In this example, you will learn how to use a comma."
+    - **Incorrect**: "In this example you learn how to use a comma."
```
dipikabh, line 334:
```suggestion
    - **Incorrect**: "In this example you will learn how to use a comma."
```

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
pepelsbey: I think we would benefit from such guidelines since MDN content is not just linear prose but also demos/examples where more diverse directionality is more common. It's not a linting rule that will scream at you; it's a guide to help you start thinking more inclusively.

pepelsbey (approved):
I like it! Thank you 👏

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
dipikabh: Thanks a lot Vadim for sharing your perspective! Yes precisely, "inclusivity" (not a word it seems) is what we want to target with the language we use on MDN.

Hamish: We discussed this in the content team and the general consensus is to follow the guideline, be mindful going forward of the language we use in prose.
Since our output is ultimately rendered as HTML–regardless of whether we write in HTML or Markdown–that's the context we're considering when using this guidance.
Happy to discuss more when you're back.

dipikabh:
Thanks for reviewing @pepelsbey.
 @hamishwillee - thanks for explaining more. Please see my response in https://github.com/mdn/content/pull/38903#discussion_r2042235195. I'm open to revisiting the topic when you're back.
Merging this for now, and we can revise if needed.

files/en-us/mdn/writing_guidelines/writing_style_guide/index.md
dipikabh: @hamishwillee Sorry, I'm still trying to understand your perspective.
Yeah, I agree - prose and code are always together and rendered in a linear order. I'm not asking for us to change the flow of our docs in any way.

My concern is with using words like "below" and "above". These make sense if you're looking at the page, but for people using screen readers or other assistive tech, these words don't help if those folks are navigating the content by headings, landmarks, or code blocks. And even when prose and code are directly next to each other, "below" assumes the reader is looking at the page and can visually scan down to spot the relevant example.

While trying to argue both for and against the usage of these words, I realized that for a screen reader user, "above" is even trickier. If someone lands in the middle of a page, "see the code above" could refer to any number of things, and there's no quick way for a non-visual user to know what it refers to except by moving through each item one at a time. Sighted readers can quickly glance up the page and see what "above" refers to. It would be more helpful to refer to the code block by what it does.

I think using "following" is a step up from "below" since it tells the reader that the example/code block comes next in the reading order and isn't tied to the visual layout. That to me makes "following" a better word choice for both accessibility and general clarity; it aligns better with the linear experience. Using "following" is still not perfect, but I guess it's less ambiguous in the context of logical reading order. Similarly, "preceding" can be used instead of "above". We can use these alternate words or be more descriptive about our references.

This is one of those small tweaks that can make the docs a bit more accessible and easier to follow for everyone. Like with our other inclusive language choices, this is just about being a bit more thoughtful about the words we use so our docs work well for as many people as possible, no matter how they're reading.
Even if we haven't had complaints, it’s worth being proactive and thoughtful about our language.

hamishwillee: I guess this feels far more like virtue signaling that providing any real value - mostly because I can't imagine anyone using a screen reader that doesn't know that below means forward and above means preceding. The words might be technically more accurate, but they flow poorly.

PR #38911, opened 2025-04-01
Remove closing slashes from <wbr> examples

adamchainz:
### Description

Remove the closing space and forward slash from the `<wbr>` examples.

### Motivation

They don’t do anything per https://jakearchibald.com/2023/against-self-closing-tags-in-html/ .

### Additional details

n/a

n/a


adamchainz:
Oh okay this is linted for, never mind then.

PR #38920, opened 2025-04-01
Fix hash glossary links

tunetheweb:
### Description

Fix linting error from #38722.

@wbamberg sorry I spotted this but was just to slow to raise it before you merged! :-(

### Motivation

Prevent linting errors for anyone editing this doc in the future.

#38722.

### DIFF
--- a/files/en-us/web/http/reference/headers/available-dictionary/index.md
+++ b/files/en-us/web/http/reference/headers/available-dictionary/index.md
@@ -13,7 +13,7 @@ See the [Compression Dictionary Transport guide](/en-US/docs/Web/HTTP/Guides/Com
 
 ## Syntax
 
-Clients can send an `Available-Dictionary` header when they support `dcb` or `dcz` encodings. The header is a colon-surrounded base-64 encoded SHA-256 {{glossary("hash")}} of the dictionary contents.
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for the PR, but I don't think we should make this change: I think the linter is wrong here. Giving us a casing flaw for glossary text is not helpful, because it means 9/10 we will need to provide 2 macro arguments instead of one, for no real benefit. There is a related issue here which I will try to find :).

tunetheweb:
Actually is the issue using `"Hash"` instead of `"Hash_function"`? In which case there may be more of an argument to use two arguments.

But if you're happy to ignore then suits me!

wbamberg:
> Actually is the issue using `"Hash"` instead of `"Hash_function"`? In which case there may be more of an argument to use two arguments.
> 

Oh no you are right. I renamed that page myself [just the other day](https://github.com/mdn/content/pull/38832#discussion_r2017770444), while your PR was still in flight!

wbamberg (approved):
👍 thank you!

tunetheweb:
And my OCD is sated :-)

Still wonder what that "No generic content config found" is all about!

PR #38923, opened 2025-04-01
Reorganize graphics software list

kaipepler:
Added distinction to recommended graphics packages.

### Description

Separated Photoshop from Sketch as industry standards in their respective area, and expanded newcomers section.

### Motivation

Sketch isn't really a newcomer, as it has been around for 15 years and has historically been the standard choice for UI and icon work, where photoshop is much better for raster editing. I wanted to make this distinction to allow users to pick the best graphics package for their goals.


hamishwillee (approved):
Thank you. Seems reasonable. MDN doesn't really like having lists like this (at all) because they are hard to maintain and change over time. But all these appear to have stability and a track record (I questioned The Affinity Suite but basic checks show it has been around for a while).

kaipepler:
That's good to know, thank you for the information!

PR #38925, opened 2025-04-02
Fix invalid links

Josh-Cena:
Was about to add to https://github.com/mdn/content/pull/38924 but it was merged too quickly


wbamberg (approved):
👍 oops

PR #38926, opened 2025-04-02
Replace DL anchors with real target

Josh-Cena:
In general, real links are easier to check via all sort of automation. They also provide more substantive content for the reader and do not require one extra click if they want to read more. Furthermore, in-page anchor links are bad reader experience because they have to scroll back, whereas a page navigation can be quickly reverted to the same position on the original page.


bsmth (approved):
These are good changes. I presume a lot of this happens because the links are created to (same) page sections which are later extracted out into separate documents.

Thanks 👍🏻

PR #38876, opened 2025-03-31
feat(meta): Remove usage of EmbedInteractiveExample in meta docs

bsmth:
### Description

This PR removes references to `EmbedInteractiveExample` and replaces instances with `InteractiveExample`.

__Changes:__
* Replace `EmbedInteractiveExample` instances with `InteractiveExample`
* Writing markdown on MDN has small incidental prose tidy
* Kitchensink uses `InteractiveExample` instead of `EmbedInteractiveExample`.

### Motivation

We're no longer using the `EmbedInteractiveExample` macro in content.

### Related issues and pull requests

- [x] https://github.com/mdn/content/pull/38552
- [x] https://github.com/mdn/content/pull/38356
- [x] https://github.com/mdn/content/pull/38355
- [x] https://github.com/mdn/content/pull/38257
- [x] https://github.com/mdn/content/pull/38023

PR #38927, opened 2025-04-02
Use uniform code block language

Josh-Cena:
I'm adding checks for code blocks. This fixes some abnormal code block language markup.


files/en-us/learn_web_development/extensions/server-side/django/models/index.md
```diff
@@ -134,7 +134,7 @@ The following common arguments can be used when declaring many/most of the diffe
   > Apps created using **manage.py** set the type of the primary key to a [BigAutoField](https://docs.djangoproject.com/en/5.0/ref/models/fields/#bigautofield).
   > You can see this in the local library **catalog/apps.py** file:
   >
-  > ```py
+  > ```python
```
bsmth, line 137: These mistakes are too easy to make as authors, especially when some are `python` v `py` and others are `md` instead of `markdown`. I guess we should have this in checks (mdlint / precommit / whatever).

bsmth (dismissed):
LGTM, thanks 👍🏻

Josh-Cena:
Added a lint config

Josh-Cena:
@bsmth Please check my last few commits. I've updated the writing guide slightly, and I've additionally normalized `wasm` to `wat` because `wasm` is not a textual format

files/en-us/mdn/writing_guidelines/howto/markdown_in_mdn/index.md
bsmth, line 1: I think we have a race here: https://github.com/mdn/content/pull/38876

Nothing too complex, though, so no preference which lands first

files/en-us/mozilla/add-ons/webextensions/manifest.json/chrome_settings_overrides/index.md
```diff
@@ -24,7 +24,7 @@ browser-compat: webextensions.manifest.chrome_settings_overrides
 
   <tr>
       <th scope="row">Example</th>
-      <td><pre class="brush:json">
+      <td><pre class="brush: json">
```
caugner, line 27: Is this auto-fixed by markdownlint, or did you manually fix these?

files/en-us/mdn/writing_guidelines/howto/markdown_in_mdn/index.md
```diff
@@ -68,7 +56,7 @@ On MDN, writers will use code fences for example code blocks. They must specify
     - `rust` - Rust
     - `glsl` - GLSL (OpenGL Shaders)
     - `sql` - SeQueL commands
-    - `wasm` - WebAssembly
+    - `wat` - WebAssembly
```
bsmth, line 59: For Prism, the keyword is `wasm` (which you can alias), but `wat` doesn't match a language, currently.

files/en-us/mozilla/add-ons/webextensions/manifest.json/chrome_settings_overrides/index.md
Josh-Cena: Manually fixed these. There aren't many cases. This PR isn't generated by markdownlint; I added the lint rule after the fact by Brian's suggestion.

.markdownlint.jsonc
```diff
@@ -104,7 +104,58 @@
   "no-bare-urls": false,
   // Produces too many false positives
   "no-space-in-emphasis": false,
-  "fenced-code-language": true,
+  "fenced-code-language": {
+    // Your language is not in this list? Please check https://github.com/orgs/mdn/discussions/170#discussioncomment-3404366
```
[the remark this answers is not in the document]
Josh-Cena: @caugner This decision is now solidified by the writing guide: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#example_code_blocks, so it is pretty much timeless.

> nd that yari afaik no longer uses an allowlist, can you clarify why you're doing this change now

Currently we use a blocklist for certain identifiers, especially those for plaintext. As you can see in the diff, this approach is not scalable because people keep inventing new ways to say plaintext. There's also no consistency: identifiers like `javascript` or `markdown` keep coming back. So it's better to just settle on an allowlist, which is the same as the list given in the writing guide.

> (I'm specifically surprised about changing py to python.)

I'm indifferent either way. I only chose `python` because it's much more commonly used than `py`. I'm very happy to do otherwise if others agree, but I don't want to start a longer bikeshedding about which is better; I only want consistency.

files/en-us/mdn/writing_guidelines/howto/markdown_in_mdn/index.md
Josh-Cena: @bsmth The interactive examples already use `wat` as the language, so this is not an entirely new things. Thanks for pointing about not mapping to a Prism language. I'll revert it for now.

rebloor (approved):
Looks good for the web extension related changes

.markdownlint.jsonc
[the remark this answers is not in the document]
bsmth: >I don't want to start a longer bikeshedding about which is better; I only want consistency

I also agree on this approach for these kinds of changes.

>yari afaik no longer uses an allowlist

This is less about what's technically possible, but about code style / conventions in the repo, so these changes are great, IMO.

bsmth (approved):
Thanks a lot!

files/en-us/mdn/writing_guidelines/howto/markdown_in_mdn/index.md
bsmth: > The interactive examples already use wat as the language

Good point, I'll bring it up with eng

Josh-Cena: For the record, I prefer `wat` to `wasm` because `wasm` is a binary format, not a text format, but I don't have strong opinions.

PR #38928, opened 2025-04-02
Add live sample embeds for SVG path examples

### LINKED ISSUE #38495: Incorrect column title in arc flags illustration: "Large Arc Sweep Flag"
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Paths

Illustration of large arc and sweep flag combinations (svgarcs_flags.png)

The column label reads "Large Arc Sweep Flag".

https://github.com/mdn/content/blob/5f50e55f9c64273f69bd01ed16698fe32f1c20c2/files/en-us/web/svg/tutorial/paths/svgarcs_flags.png

The column label should read "Large Arc Flag".

The SVG 2 Paths spec: https://svgwg.org/svg2-draft/paths.html#PathElement

The error can be found on a small number of other (non-MDN) sites and is likely a result of accidentally merging `large-arc sweep-flag` into ``large-arc-sweep-flag``.

### MDN metadata

<details>

* Folder: `en-us/web/svg/tutorial/paths`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Paths
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/svg/tutorial/paths/index.md
* Last commit: https://github.com/mdn/content/commit/01b8471b84e1d157cbddbb3ffaf560a86b082070
* Document last modified: 2024-11-14T23:45:45.000Z

</details>

AshmitGupta:
### Description

- Fixed the label in `svgarcs_flags.png` to reflect "Large Arc Flag"
- Updated the page to include `{{EmbedLiveSample}}` for better user experience

### Motivation

- The previous image used an incorrect label which could confuse readers about the `large-arc-flag` and `sweep-flag` parameters. 
- Also, the addition of live sample embeds maies it easier for users to experiment directly with the examples.

Fixes #38495


AshmitGupta:
I think the original intention behind this issue may have been to replace the PNG diagrams with live SVG samples. I've gone ahead and added `{{EmbedLiveSample}}` blocks for interactivity wherever possible.

That said, some of the PNG illustrations - especially those with blue and red lines—convey additional contextual information (like inferred control points or ellipse boundaries) that’s not as easily represented through live code alone. So instead of removing those, I’ve kept them alongside the live samples to preserve that extra clarity.

Happy to hear your thoughts on this approach!

chrisdavidmills (approved):
Hi @AshmitGupta, and thank you so much for converting all of these demos into live samples. I really think it improves the experience on the page. 

The only thing I wondered is whether it would be good to add a paragraph before each live sample embed saying something like "The above code is rendered as follows:", so it is really obvious what the connection is in each case. What do you think?

files/en-us/web/svg/tutorials/svg_from_scratch/paths/svgarcs_flags.png
Josh-Cena, line 1: The idea is that we should just demonstrate the SVG code that generated it, with a live sample, so we don't need this file anymore.

AshmitGupta:
> Hi @AshmitGupta, and thank you so much for converting all of these demos into live samples. I really think it improves the experience on the page.
> 
> The only thing I wondered is whether it would be good to add a paragraph before each live sample embed saying something like "The above code is rendered as follows:", so it is really obvious what the connection is in each case. What do you think?

Thanks @chrisdavidmills! That’s a great suggestion.

Just wanted to note that pages like https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform-origin and https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-width also use similar live SVG embeds, and they don’t include that extra explanatory line either.

I think that extra sentence might have made more sense if we were keeping the PNG images. But since, as @Josh-Cena pointed out, we’re replacing them with live SVG samples and possibly incorporating the necessary markup directly into them, it feels consistent to follow the simpler pattern used in those other pages. What do you think?

files/en-us/web/svg/tutorials/svg_from_scratch/paths/svgarcs_flags.png
AshmitGupta: Thanks for the clarification!

Just a quick thought - adding visual aids like inferred control points and guide lines directly into the SVG sample (like we did for the Bézier curves) significantly increases the amount of code needed to reproduce what was previously a simple image. For instance, the SVG for the Bézier diagram grew quite a bit after including the red lines and control points for clarity (screenshot attached).

Is this added complexity okay from a documentation perspective? I’m happy to proceed this way for other images too if that's the preferred direction, but wanted to double-check before replacing similar PNGs with code-heavy alternatives.

Let me know what you think!

**Original Code**
Original Bezier Curve SVG Code


**Code with lines**
Bezier Curve with lines (Part 1)
Bezier Curve with lines (Part 2)

chrisdavidmills:
> Thanks @chrisdavidmills! That’s a great suggestion.
> 
> Just wanted to note that pages like https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform-origin and https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-width also use similar live SVG embeds, and they don’t include that extra explanatory line either.
> 
> I think that extra sentence might have made more sense if we were keeping the PNG images. But since, as @Josh-Cena pointed out, we’re replacing them with live SVG samples and possibly incorporating the necessary markup directly into them, it feels consistent to follow the simpler pattern used in those other pages. What do you think?

@AshmitGupta OK, that is totally fair. I'm not that familiar with the SVG docs; in the API and CSS docs, we usually include a bit more structure and explanation about the code listing and live example, but I think you are right — it makes sense to be consistent.

I'm satisfied with this, but I'll not merge yet in case @Josh-Cena has any more thoughts, or you want to make any more changes.

files/en-us/web/svg/tutorials/svg_from_scratch/paths/svgarcs_flags.png
Josh-Cena: @AshmitGupta You don't need to change the other examples; just that particular example that produces the PNG diagram you are replacing. I think extra markup is fine.

Josh-Cena (approved):
I'm having a lot of fun :) I now migrated all PNGs to SVGs, and made the reference markup hidden from the page. Thank you!

PR #38930, opened 2025-04-02
Added link to Web Speech API article written by the Pipe team

remusnegrota:
### Description
A link to a comprehensive article which covers all the aspects of  the Web Speech API together with fully working code examples for both Speech Recognition and Speech Synthesis

### Motivation
Added a link to an article written by us (addpipe.com)

### Additional details
The article provides additional context on how to implement the Web Speech API and what kind of use cases this API could enable.
This is part of our publishing effort for the article that we've written.


wbamberg:
Thanks for your PR!

We're very conservative about adding external links to MDN, as our experience is that they are often not actively maintained. We do have an external link policy at https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide#external_links. I'm not sure it meets the criteria for "good external links" there:

> You should prefer adding links to external content that is:
> 
> - Unique or indispensable (e.g., an IETF RFC)
> - Necessary for attribution, citation, or acknowledgement (e.g., as part of a Creative Commons attribution)
> - More likely to be maintained for the topic than incorporating such content on MDN Web Docs itself (e.g., a vendor's release notes)
> - Open source or community-driven, like MDN Web Docs itself

That is, while your post is a good article about the API, I'm not sure what it adds that isn't on MDN already. (FWIW I feel the same way about the other external links in this page, and would actively remove at least the HTML5Rocks one).

PR #38931, opened 2025-04-02
add l10n for new sources footer

fiji-flo:
### Description

This adds translations for the css syntax macro.

See: https://github.com/mdn/yari/pull/12861


bsmth (approved):
Thanks!

PR #38934, opened 2025-04-02
Fix: change `iSheetIndex` to the correct `i`

liuxuelian:
### Description

Change `iSheetIndex` to the correct `i`


wbamberg (approved):
👍 thank you!

PR #38936, opened 2025-04-02
Add a json validation tool which scores higher for a11y and is wcag compliant.

focal-local-bg:
The current link for JSON validation should be updated. https://json-validate.com offers a more modern and inclusive solution. Our tool scores significantly higher on A11y Lighthouse reports, ensuring better accessibility and WCAG compliance.

In addition to the point above, it offers features like tree mode editing, YAML conversion, formatting, and a random JSON generator. The alternatives are slower, have a less accessible interface, require more steps for validation, and appears to be no longer actively maintained. I suggest adding the tool https://json-validate.com to the list of suggested tools.

- Adding reference to a a11y friendly json validation tool.


chrisdavidmills (approved):
@focal-local-bg LGTM; I'm happy to add another tool choice here. Looks pretty good!

focal-local-bg:
Thanks for the feedback and the merge @chrisdavidmills ! 🙌

PR #38937, opened 2025-04-02
Slightly rewrite Constructor glossary

Josh-Cena:
I noticed a lot of technical inaccuracies about this page, so this PR rewrites it.

### DIFF
--- a/files/en-us/glossary/constructor/index.md
+++ b/files/en-us/glossary/constructor/index.md
@@ -8,7 +8,7 @@ page-type: glossary-definition
 
 A **constructor** is a specialized {{glossary("function")}} that generates {{glossary("object", "objects")}} with the same shape and behavior. The constructor initializes this object with some data specific to the object. The concept of a constructor can be applied to most {{glossary("OOP","object-oriented programming")}} languages.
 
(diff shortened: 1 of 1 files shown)


files/en-us/glossary/constructor/index.md
```diff
@@ -6,28 +6,9 @@ page-type: glossary-definition
...
-To call the constructor of the class in JavaScript, use a `new` operator to assign a new {{glossary("object reference")}} to a {{glossary("variable")}}.
-
-```js
-function Default() {}
-
-// A new reference of a Default object assigned to a
-// local variable defaultReference
-const defaultReference = new Default();
-```
+In {{glossary("JavaScript")}}, a constructor is usually declared within a [class](/en-US/docs/Web/JavaScript/Reference/Classes), but [functions](/en-US/docs/Web/JavaScript/Reference/Functions) can be constructors as well. In fact, any function that can be called with the [`new`](/en-US/docs/Web/JavaScript/Reference/Operators/new) operator is a constructor.
```
chrisdavidmills, line 11: This paragraph sounds a bit confused to me. Isn't `Class` just syntactic sugar that creates a constructor function behind the scenes, which can then be called with `new` to create an object instance? You are making it sound like they are two different things.

chrisdavidmills (changes requested):
@Josh-Cena looking mostly good; just one query for you to look at.

files/en-us/glossary/constructor/index.md
chrisdavidmills: I think we are on the same page; it is the "but [functions](/en-US/docs/Web/JavaScript/Reference/Functions) can be constructors as well" language that is confusing me. Maybe update it to "but a constructor can also be declared with a [function](/en-US/docs/Web/JavaScript/Reference/Functions)." ?

chrisdavidmills (approved):
Looking good; thanks @Josh-Cena!

PR #38940, opened 2025-04-02
Normalize event syntax, add -nolint

Josh-Cena:
The events reference is the only part that doesn't add `-nolint` to the example code, despite the code being pseudocode. This PR reverts the syntax back to what it was before Prettier was installed and added `-nolint`.


wbamberg (approved):
👍 sorry I put off reviewing this incredibly boring PR for so long.

PR #38941, opened 2025-04-02
Eliminate many syntax errors

Josh-Cena:
In general, code examples with valid syntax is more easily integrated with all sorts of analysis. They are also more user-friendly for readers because it sets up all the necessary context instead of taking an excerpt that's not valid by itself.

The PR may seem bigger than it is. A lot of the changes are just caused by the examples now being properly formattable.


files/en-us/web/api/attribution_reporting_api/generating_reports/index.md
```diff
@@ -124,13 +124,16 @@ The properties are as follows:
...
 
-        ```js
+        ```json
         {
-          "operation": "histogram",  // Allows for the service to support other operations in the future
-          "data": [{
-            "bucket": <bucket, encoded as a 16-byte (i.e. 128-bit) big-endian bytestring>,
-            "value": <value, encoded as a 4-byte (i.e. 32-bit) big-endian bytestring>
-          }, ...]
+          "operation": "histogram", // Allows for the service to support other operations in the future
```
bsmth, line 129: Any nicer way to move the comments into prose? Not blocking

bsmth (approved):
All looks good. One non-blocking comment to check, but happy to merge regardless.

PR #38942, opened 2025-04-02
docs(base64) clarify how base64 digit is 6 bits

hchiam:
Just adding a quick clarification that helped me understand an intermediate/implicit step in the explanation.

https://developer.mozilla.org/en-US/docs/Glossary/Base64#encoded_size_increase

### DIFF
--- a/files/en-us/glossary/base64/index.md
+++ b/files/en-us/glossary/base64/index.md
@@ -24,7 +24,7 @@ Base64 encoding schemes are commonly used to encode binary data for storage or t
 
 ## Encoded size increase
 
-Each Base64 digit represents 6 bits of data (64 = 2 to the power of 6). So, three 8-bit bytes of the input string/binary file (3×8 bits = 24 bits) can be represented by four 6-bit Base64 digits (4×6 = 24 bits).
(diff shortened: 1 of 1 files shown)


files/en-us/glossary/base64/index.md
```diff
@@ -24,7 +24,7 @@ Base64 encoding schemes are commonly used to encode binary data for storage or t
 
 ## Encoded size increase
 
-Each Base64 digit represents 6 bits of data. So, three 8-bit bytes of the input string/binary file (3×8 bits = 24 bits) can be represented by four 6-bit Base64 digits (4×6 = 24 bits).
+Each Base64 digit represents 6 bits of data (64 = 2 to the power of 6). So, three 8-bit bytes of the input string/binary file (3×8 bits = 24 bits) can be represented by four 6-bit Base64 digits (4×6 = 24 bits).
```
Josh-Cena, line 27:
```suggestion
Each Base64 digit represents 6 bits of data (64 = 2<sup>6</sup>). So, three 8-bit bytes of the input string/binary file (3×8 bits = 24 bits) can be represented by four 6-bit Base64 digits (4×6 = 24 bits).
```

Josh-Cena (approved):
Sure, thank you

hchiam (commented):
LGTM! :+1:

PR #38943, opened 2025-04-03
Eliminate many syntax errors (2)

Josh-Cena:
Like https://github.com/mdn/content/pull/38941, but with half of the work split out for ease of review

### DIFF
--- a/files/en-us/web/api/webgl_api/tutorial/creating_3d_objects_using_webgl/index.md
+++ b/files/en-us/web/api/webgl_api/tutorial/creating_3d_objects_using_webgl/index.md
@@ -67,9 +67,9 @@ const faceColors = [
 
 // Convert the array of colors into a table for all the vertices.
 
-var colors = [];
+const colors = [];
 
-for (var j = 0; j < faceColors.length; ++j) {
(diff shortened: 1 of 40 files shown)


files/en-us/web/api/webgl_api/tutorial/creating_3d_objects_using_webgl/index.md
```diff
@@ -92,43 +92,14 @@ function initIndexBuffer(gl) {
   // indices into the vertex array to specify each triangle's
   // position.
 
+  // prettier-ignore
```
bsmth, line 95: This is good, I think we have many cases of matrices that were mangled during initial formatting runs, but not all have the comment lines like this to help bring back the illustrative spacing.

bsmth (approved):
Great, thanks 👍🏻

files/en-us/web/api/webgl_api/tutorial/creating_3d_objects_using_webgl/index.md
Josh-Cena: Previously we were very anti-disable comments so these cases were either overlooked or entirely nolinted. Now I think disabling one line is a better idea.

PR #38944, opened 2025-04-03
Update redirected links

Josh-Cena:
Google has moved their privacy docs, yet again.


hamishwillee (approved):
Thanks very much @Josh-Cena .

PR #38948, opened 2025-04-03
chore(http): Add Compression Dictionary Transport headers

bsmth:
### Description

Adds Compression Dictionary Transport to headers page

Follow-up from #38722


files/en-us/web/http/reference/headers/index.md
```diff
@@ -429,6 +429,18 @@ Network client hints allow a server to choose what information is sent based on
 - {{HTTPHeader("Save-Data")}} {{experimental_inline}}
   - : A string `on` that indicates the user agent's preference for reduced data usage.
 
+### Compression Dictionary Transport
+
+Compression Dictionary Transport is a way of using a shared compression dictionary to reduce the transport size of HTTP responses rather than using the standard static dictionary in {{glossary("Brotli compression")}} or {{glossary("Zstandard compression")}}.
```
tunetheweb, line 434: Worth linking to this guide that I added for more info?:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport

bsmth: Good call, it's probably best to link the first occurrence:

```suggestion
[Compression Dictionary Transport](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) is a way of using a shared compression dictionary to reduce the transport size of HTTP responses rather than using the standard static dictionary in {{glossary("Brotli compression")}} or {{glossary("Zstandard compression")}}.
```

files/en-us/web/http/reference/headers/index.md
```diff
@@ -31,7 +31,7 @@ Headers can also be grouped according to how {{Glossary("Proxy_server", "proxies
 - End-to-end headers
   - : These headers _must_ be transmitted to the final recipient of the message: the server for a request, or the client for a response. Intermediate proxies must retransmit these headers unmodified and caches must store them.
 - Hop-by-hop headers
-  - : These headers are meaningful only for a single transport-level connection, and _must not_ be retransmitted by proxies or cached. Note that only hop-by-hop headers may be set using the {{httpheader("Connection")}} header.
+  - : These headers are meaningful only for a single transport-level connection, and _must not_ be retransmitted by proxies or cached. Note that only hop-by-hop headers may be set using the {{HTTPHeader("Connection")}} header.
```
hamishwillee, line 34: Strongly disagree with all these changes to the macro casing. The system doesn't and shouldn't care about case for macros, and neither should we.

files/en-us/web/http/reference/headers/index.md
```diff
@@ -429,6 +429,18 @@ Network client hints allow a server to choose what information is sent based on
 - {{HTTPHeader("Save-Data")}} {{experimental_inline}}
   - : A string `on` that indicates the user agent's preference for reduced data usage.
 
+### Compression Dictionary Transport
+
+[Compression Dictionary Transport](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) is a way of using a shared compression dictionary to reduce the transport size of HTTP responses rather than using the standard static dictionary in {{glossary("Brotli compression")}} or {{glossary("Zstandard compression")}}.
+
+- {{HTTPHeader("Available-Dictionary")}} {{experimental_inline}}
+  - : Lets the browser indicate that it already has a dictionary available that the server can use for compression.
```
hamishwillee, line 437: IMO useful to indicate the kind of header here, there and everywhere :-)

```suggestion
  - : Request header indicating that the browser has a dictionary available that the server can use for compression.
```

hamishwillee:
```suggestion
  - : A browser can use this request header to indicate the best dictionary it has available for the server to use for compression.
```

files/en-us/web/http/reference/headers/index.md
```diff
@@ -500,9 +512,7 @@ See the [Topics API](/en-US/docs/Web/API/Topics_API) documentation for more info
 
 ## Contributing
 
-You can help by [writing new entries](/en-US/docs/MDN/Writing_guidelines/Howto/Document_an_HTTP_header) or improving the existing ones.
-
-<!-- Check https://github.com/mdn/content/issues/1458 for known missing pages -->
+You can help by [writing new entries](/en-US/docs/MDN/Writing_guidelines/Howto/Document_an_HTTP_header) or improving existing ones.
```
hamishwillee, line 515: I'm wondering if we should have this section at all. I mean we don't generally sprinkle calls-to-action.

files/en-us/web/http/reference/headers/index.md
```diff
@@ -429,6 +429,18 @@ Network client hints allow a server to choose what information is sent based on
...
 
+### Compression Dictionary Transport
+
+[Compression Dictionary Transport](/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport) is a way of using a shared compression dictionary to reduce the transport size of HTTP responses rather than using the standard static dictionary in {{glossary("Brotli compression")}} or {{glossary("Zstandard compression")}}.
+
+- {{HTTPHeader("Available-Dictionary")}} {{experimental_inline}}
+  - : Lets the browser indicate that it already has a dictionary available that the server can use for compression.
+- {{HTTPHeader("Dictionary-ID")}} {{experimental_inline}}
+  - : Used when a browser already has a dictionary available for a resource and the server provided an `id` for the dictionary in the `Use-As-Dictionary` header.
+    Requests for resources that can use the dictionary have an `Available-Dictionary` header and the server-provided dictionary `id` in the `Dictionary-ID` header.
```
hamishwillee, line 440: FMI what is the point of `Available-Dictionary`, and does it always require `Dictionary-ID` too?

I ask because you get and return the server id. Why do you need a hash of the dictionary too when you already have an identity for it?

hamishwillee (approved):
@bsmth I've added a suggestion and a question that is mostly FMI https://github.com/mdn/content/pull/38948/files#r2027917100

But this is probably all fine, so approving in case you think these are not useful.

files/en-us/web/http/reference/headers/index.md
tunetheweb: I can take this one. @pmeenan correct me if I'm wrong.

The point of `Available-Dictionary` is for the client to tell the server which Dictionary it has that can be used for compression. It is a crypto-graphic hash to ensure the correct dictionary is actually used.

No it does not require the optional `Dictionary-ID` too. This is an optional, extra field that, from my understanding, servers may use if they don't want to store dictionaries by the hash value but by some other identifier. In this case calculating the hash of all available dictionaries to find the match to the `Available-Dictionary` would be expensive. And storing a key-value store of these might be large. Once the dictionary is found, the hash must still be calculated and checked against the `Available-Dictionary` to ensure a match.

The identifier is provided in the `Use-As-Dictionary` header (when the dictionary is originally sent to the client) in the `id` field and and must be reflected in the `Dictionary-ID` header for any subsequent request wanting to use that dictionary. 

See: https://github.com/httpwg/http-extensions/pull/2687 (and the issue it closes).

pmeenan: Correct. The `Dictionary-ID` was an added helper to make it easier for origins and CDN's to find/fetch the dictionary using whatever key they want (for example, the URL of the original resource). It saves them from having to hash every response that includes a `Use-As-Dictionary` response header just in case it might eventually be used as a dictionary (and to avoid double-storing).

The client doesn't do anything with it except echo whatever value was used back. The hard-verification (and failure in case of a mismatch) is on the dictionary hash.

They still verify the hash against Available-Dictionary before using it but that's when they're already initializing the dictionary itself anyway.

tunetheweb: OK updated the main docs in this PR with some information on this: https://github.com/mdn/content/pull/38974

hamishwillee: Thank you both. I will look at that.

files/en-us/web/http/reference/headers/index.md
bsmth: >The system doesn't and shouldn't care about case for macros, and neither should we.

Yeah, I have no strong opinions on what convention we settle for this syntax as long as there's only one way to write it, so "all lower, no spaces" is fine with me. This is what we're currently looking at:

* 1x `{{HttpHeader(…)}}`
* 3x `{{ HTTPHeader(…) }}`
* 3x `{{ httpheader(…) }}`
* 3x `{{HTTPheader(…)}}`
* 422x `{{httpheader(…)}}`
* 1814x `{{HTTPHeader(…)}}`

files/en-us/web/http/reference/headers/index.md
bsmth: Good, I like the second one

PR #38957, opened 2025-04-03
Used proper minus sign to match + width in example

woodyrew:
### Description

The standard - is more narrow than the +. When switching between open and closed states we get a layout shift or an unexpected look. − is the proper counter part to the +.

### Motivation

The layout shifts when interacting with the example distracting from the example.


Josh-Cena (approved):
Looks great, thank you :)

PR #38434, opened 2025-03-04
FF137 Relnote: SVG <discard> element

hamishwillee:
FF127 ships `<discard>` in https://bugzilla.mozilla.org/show_bug.cgi?id=1945330. This adds a release note and removes from experimental features.

Related docs work can be tracked in #38409

PR #38961, opened 2025-04-04
FF137 relnote/expr feat - remove SVG discard

hamishwillee:
FF137, FF138 has disabled the [`<discard>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/discard) element behind a pref in https://bugzilla.mozilla.org/show_bug.cgi?id=1954608
This was supposed to be added in 137 nightly and release in 138 but is still in discussion specification wise. 

The change reverts #38434, removing discard from the release note and adding back to the experimental features. 

Corresponding BCD work done in https://github.com/mdn/browser-compat-data/pull/26398

### DIFF
--- a/files/en-us/mozilla/firefox/experimental_features/index.md
+++ b/files/en-us/mozilla/firefox/experimental_features/index.md
@@ -786,7 +786,48 @@ The parts that have been implemented include:
 
 ## SVG
 
-None.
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #38963, opened 2025-04-04
FF138 Relnote: JS Error.captureStackTrace() and .isError()

hamishwillee:
FF138 supports `Error.captureStackTrace()` and `Error.isError()` in https://bugzilla.mozilla.org/show_bug.cgi?id=1952249 and
https://bugzilla.mozilla.org/show_bug.cgi?id=1950508

This adds a release note and removes from Experimental features.

Related docs work can be tracked in #38879

### DIFF
--- a/files/en-us/mozilla/firefox/experimental_features/index.md
+++ b/files/en-us/mozilla/firefox/experimental_features/index.md
@@ -844,48 +844,6 @@ This includes:
   </tbody>
 </table>
 
-### Error.captureStackTrace
-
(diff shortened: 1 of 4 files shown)


Josh-Cena:
Btw, if you are still interested, https://github.com/tc39/proposal-error-capturestacktrace is a real proposal now, so you can link to it

hamishwillee:
> Btw, if you are still interested, https://github.com/tc39/proposal-error-capturestacktrace is a real proposal now, so you can link to it

Thanks @Josh-Cena - I saw that. I've got a BCD update with spec url to that. If/when it goes through I'll see if the doc also needs a little tidy.

bsmth (approved):
Nice, thanks 👍🏻

PR #38971, opened 2025-04-04
Update code example related to above text

vladsadretdinov:
### Description

Update code example related to above text

### Motivation

It's confusing


Josh-Cena (approved):
Good catch, thanks

PR #38972, opened 2025-04-04
Update async_function_star_ to have a working example

jaenster:
### Description

The current example puts in the console ` [object Promise]`

### Motivation

The example now does what it says it does


Josh-Cena (approved):
Great catch. I've been meaning to fix this for a while. Thank you!

jaenster:
> Great catch. I've been meaning to fix this for a while. Thank you!

Oh the then example is way better, good change

PR #38973, opened 2025-04-04
Add a json patch generator tool

cyrilbois:
I suggest adding the tool https://jsoning.com/jsonpatch/ (my website) to the list of resources.
It allows to generate a JSON Patch from two JSON files. I suppose it could be useful for people consulting the PATCH page.


hamishwillee (approved):
Thanks @cyrilbois 

This looks useful to me - thank you, and welcome to MDN.

FYI, normally we're pretty conservative about adding external links to MDN, as our experience is that they are often not actively maintained. We do have an [external link policy](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide#external_links) which I think this meets (or at least isn't actively non cmopliant):

> You should prefer adding links to external content that is:
> 
> - Unique or indispensable (e.g., an IETF RFC)
> - Necessary for attribution, citation, or acknowledgement (e.g., as part of a Creative Commons attribution)
> - More likely to be maintained for the topic than incorporating such content on MDN Web Docs itself (e.g., a vendor's release notes)
> - Open source or community-driven, like MDN Web Docs itself

PR #38974, opened 2025-04-04
Compression Dictionary Transport updates

tunetheweb:
### Description

Follow on from #38722 (reviewed by @wbamberg ) I noticed the new guide wasn't showing in the sidebar.

I've also added more front matter to flag this as experimental and also add Spec and Compatibility table.

Finally, I've also added a bit more explanation of `id`/`Dictionary-ID` as raised at https://github.com/mdn/content/pull/38948#discussion_r2027917100 by @hamishwillee

### Motivation

More helpful to devs when viewing the guide.

An update to #38722 
Also discussed in #38948


Josh-Cena:
This is a minor point. https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel does not link to https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/compression-dictionary. Could you link to it somehow, such as in the table (following `dns-prefetch`), or in the DL below?

tunetheweb:
> This is a minor point. https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel does not link to https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/compression-dictionary. Could you link to it somehow, such as in the table (following `dns-prefetch`), or in the DL below?

Done!

files/en-us/web/html/attributes/rel/index.md
```diff
@@ -18,38 +18,38 @@ The type of relationships is given by the value of the `rel` attribute, which, i
...
+| [`canonical`](#canonical)                                                    | Preferred URL for the current document.                                                                                                                                                                                                                                                 | Link                    | Not allowed                                      | Not allowed             |
+| [`compression-dictionary`](/en-US/docs/Web/HTML/Attributes/rel/dns-prefetch) | Link to a {{glossary("Compression dictionary transport", "compression dictionary")}} that can be used to compress future downloads for resources on this site.                                                                                                                          | Link                    | Not allowed                                      | Not allowed             |
```
Josh-Cena, line 27:
```suggestion
| [`compression-dictionary`](/en-US/docs/Web/HTML/Attributes/rel/compression-dictionary) | Link to a {{glossary("Compression dictionary transport", "compression dictionary")}} that can be used to compress future downloads for resources on this site.                                                                                                                          | Link                    | Not allowed                                      | Not allowed             |
```

Unfortunately you would need to reformat it, perhaps better to not apply this suggestion but directly commit a well formatted one

tunetheweb: Oops. Done.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -2,9 +2,21 @@
...
+browser-compat:
+  - html.elements.link.rel.compression-dictionary
+  - http.headers.Accept-Encoding.dcb
+  - http.headers.Accept-Encoding.dcz
+  - http.headers.Available-Dictionary
+  - http.headers.Content-Encoding.dcb
+  - http.headers.Content-Encoding.dcz
+  - http.headers.Dictionary-ID
+  - http.headers.Use-As-Dictionary
+spec-urls: https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-compression-dictionary
```
wbamberg, line 16: We don't generally include spec tables or BCD in guide pages, only in reference pages.

wbamberg (commented):
I just had one comment on the spec/BCD additions - we don't generally include this in guide pages.

I'll leave the comments on DictionaryId to @hamishwillee because it was his comment :).

files/en-us/web/http/guides/compression_dictionary_transport/index.md
tunetheweb: That's what I thought too — hence why I didn't include it originally.

However, I noticed for [Permissions Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Permissions_Policy) we do include these.

I think that while a technology is "experimental" (like Permissions Policy and Compression Dictionary Transport are) that's a pretty good idea to mark it as such and include the full things.

The rest of the HTTP guides really are for core concepts that are supported by all browsers, so including BCD is not really necessary — though specs still seem like good ideas to me in some case. The only other similar case (that probably should be marked as experimental and treated similarly) is [HTTP Client Hints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Client_hints), but will leave that for now and stick to Compression Dictionary Transport for this PR.

WDYT?

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -13,6 +13,8 @@ The HTTP **`Dictionary-ID`** request header references a dictionary that can be
 
 In Compression Dictionary Transport, a server can indicate that a resource can be used as a dictionary by sending the {{httpheader("Use-As-Dictionary")}} header with the response. The server may include an `id` directive in the `Use-As-Dictionary` header, thus assigning an ID value to the dictionary. If the server does this, then when the browser requests a resource that can be compressed using the dictionary, then the resource request must include the `Dictionary-ID` header, and its value must be the ID that was given in `Use-As-Dictionary`.
```
hamishwillee, line 14:
```suggestion
A server can indicate that a resource can be used as a dictionary by sending the {{httpheader("Use-As-Dictionary")}} header with the response. The server may include an `id` directive in the `Use-As-Dictionary` header, thus assigning an ID value to the dictionary. If the server does this, then when the browser requests a resource that can be compressed using the dictionary, the resource request must include the `Dictionary-ID` header, and its value must match the ID that was given in `Use-As-Dictionary`.
```

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -13,6 +13,8 @@ The HTTP **`Dictionary-ID`** request header references a dictionary that can be
...
 
+The server must still check the hash from the `Available-Dictionary` header — the `Dictionary-ID` is additional information for the server to identify the dictionary but does not replace the need for the `Available-Dictionary` header.
```
hamishwillee, line 16: @tunetheweb Awesome. _Suggestion_ that we capture just a little more detail as provided by you in https://github.com/mdn/content/pull/38948#discussion_r2028188961 

```suggestion
This allows the server to identify and find a dictionary that is referenced by some arbitrary key, rather than having to use the  {{glossary("hash function","dictionary hash")}} as a key (if that approach is used, the server will have to hash every response that includes the `Use-As-Dictionary` header just in case the resource might eventually be used as a dictionary).

Note that while the server can identify and locate the dictionary from its `Dictionary-ID`, it must still check the hash from the `Available-Dictionary` header to confirm that it is a correct match!
```

tunetheweb: LGTM but don't think we need that final exclamation mark.

files/en-us/web/http/reference/headers/dictionary-id/index.md
```diff
@@ -11,7 +11,11 @@ browser-compat: http.headers.Dictionary-ID
...
+A server can indicate that a resource can be used as a dictionary by sending the {{httpheader("Use-As-Dictionary")}} header with the response. The server may include an `id` directive in the `Use-As-Dictionary` header, thus assigning an ID value to the dictionary. If the server does this, then when the browser requests a resource that can be compressed using the dictionary, the resource request must include the `Dictionary-ID` header, and its value must match the ID that was given in `Use-As-Dictionary`.
+
+This allows the server to identify and find a dictionary that is referenced by some arbitrary key, rather than having to use the  {{glossary("hash function","dictionary hash")}} as a key (if that approach is used, the server will have to hash every response that includes the `Use-As-Dictionary` header just in case the resource might eventually be used as a dictionary).
+
+Note that while the server can identify and locate the dictionary from its `Dictionary-ID`, it must still check the hash from the `Available-Dictionary` header to confirm that it is a correct match!
```
tunetheweb, line 18:
```suggestion
Note that while the server can identify and locate the dictionary from its `Dictionary-ID`, it must still check the hash from the `Available-Dictionary` header to confirm that it is a correct match.
```

bsmth:
Thank you! There's enough eyes on this so I'm taking myself off review 🙏🏻

hamishwillee (approved):

files/en-us/web/http/guides/compression_dictionary_transport/index.md
hamishwillee: @wbamberg I looked at the bits that were associated with other work I am doing, but not the rest since you reviewed that.

This bit is the only outstanding discussion and is waiting on you.

FWIW We are not consistent on this. I think the rule is understood, but isn't followed because it isn't that helpful to readers.

wbamberg: I've tried to ask for more feedback from other maintainers on this but not got much in the way of definitive responses.

But I'm tired of arguing about consistency and structure, and I don't think there's anything to say that hasn't already been said, so I'm not going to block the PR on this.

hamishwillee: Fair enough. I tend to accept things if I think they are good or not actively harmful in cases like this. Not worth banging your head against the wall for.

PR #38975, opened 2025-04-04
Add missing {{Specifications}} macros

Josh-Cena:
My tooling automatically detects pages that have associated spec URLs (via `spec-urls` or `browser-compat`) but don't have the `{{Specifications}}` macro. These pages have browser compat, and therefore should have an appropriate link to specs.


wbamberg (approved):
👍

PR #38977, opened 2025-04-04
Add a guide on xs-leaks

### LINKED ISSUE #3526: Suggestion: New article on cross-site leaks (xsleaks)
https://xsleaks.dev/ has lots of content, a bit of a bummer that MDN has nothing. It's interesting in the context of CORB, COEP, COOP and Sec-Fetch metadata.

wbamberg:
Here's a PR to add a guide to cross-site leaks. I feel like this might need some further work but it should be good enough for a review.

To understand the rationale for some of the choices I've made it would be worth looking at https://github.com/mdn/content/issues/3526#issuecomment-2715961146 and the response.

XS-leaks are a bit different to the other attacks because they are very diverse. So I've chosen three quite different attacks, to try to give an idea of their diversity, and tried to explain how the attacks have similarities and how various defenses can help with them. I'm reasonably happy with the first two attacks - error events and frame counting - but wasn't sure which to pick for a third, and not sure the CSP one is a good choice.

I haven't talked about CORP since AFAICT CORP is mostly presented in connection with Spectre/Meltdown, which I haven't discussed here (again see https://github.com/mdn/content/issues/3526#issuecomment-2715961146 for why). But perhaps I should after all introduce it, maybe as a defense against the error event attack?

I'm conscious as I write these guides to different that we're getting a lot of duplication from one guide to another in the "Defenses" section, outlining for example how to do framing protection, or set SameSite. I think we might want to factor these out into separate guides at some point.

I mentioned at the start that "side channels" can be a vector for xs-leaks, and think we should have a glossary page for that, but didn't write it yet.

Fixes https://github.com/mdn/content/issues/3526.


files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+slug: Web/Security/Attacks/XS-leaks
+page-type: guide
+---
+
+Cross-site leaks (also called XS-leaks) are a class of attack which exploit the ways in which a target site can leak information to an attacker's site, either through web platform APIs or through {{glossary("side channels")}} such as timing attacks. The information leaked could include, for example:
+
+- Whether the user has visited the target site.
+- Whether the user is logged into the target site.
+- What the user's ID on the site is.
+- What the user has recently searched for on the site.
```
hamishwillee, line 12: It is worth a following paragraph as to why a user or developer should care if any of this leaks, and perhaps some specific dangers from this knowledge [I had the same problem with the discussion of timing attacks on the wiki page - to motivate change people have to understand not just what the risk is, but the impact of the risk]. 

My first reaction was that I don't see the risk. On further thought, I guess if I know that the user has a login with bank say, that makes them a better target for trying some of the other attacks, or even an old-style scam where I send a letter - but in this case I know that they will be in that bank, and I might know their id, which adds to the trustworthyness.

Further, if I know their id, perhaps that allows a dark web search for passwords etc. Dunno, but worth making it real world.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
+---
+title: Cross-site leaks (XS-leaks)
+slug: Web/Security/Attacks/XS-leaks
+page-type: guide
+---
+
+Cross-site leaks (also called XS-leaks) are a class of attack which exploit the ways in which a target site can leak information to an attacker's site, either through web platform APIs or through {{glossary("side channels")}} such as timing attacks. The information leaked could include, for example:
```
hamishwillee, line 7: I think a web platform API is a side channel in these cases (i.e. is it worth differentiating)?

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+
+- Whether the user has visited the target site.
+- Whether the user is logged into the target site.
+- What the user's ID on the site is.
+- What the user has recently searched for on the site.
+
+Unlike other attacks such as [XSS](/en-US/docs/Web/Security/Attacks/XSS) or [Clickjacking](/en-US/docs/Web/Security/Attacks/Clickjacking), cross-site leaks are not a single technique. Instead, they are a term for a whole class of attack which exploit weaknesses in the ways that browsers isolate websites from each other.
```
hamishwillee, line 14: Do we need to say something about how these are typically deployed? I mean how do you get someone to visit evil.com - does it matter? [what I'm thinking here is that for some of the other attacks we said things like "send target an email that does X, Y]]

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Sample cross-site leaks
+
+In this section we'll describe three different cross-site leaks, to give an idea of how they work.
+
+- [Leaking page existence using error events](#leaking_page_existence_using_error_events): in this attack, an attacker can determine whether particular pages in the target site exist, by attempting to load them as resources and listening for the {{domxref("HTMLElement/error_event", "error")}} and {{domxref("HTMLElement/load_event", "load")}} events. If certain pages are only available to logged-in users, the attacker can determine whether the user is signed into the target site.
+- [Frame counting using window references](frame_counting_using_window_references): in this attack, the attacker gets a reference to a {{domxref("Window", "window")}} object hosting a page in the target site, for example as the return value of a call to {{domxref("Window.open()", "window.open()")}}. The attacker can then determine the number of {{htmlelement("iframe")}} elements in the target page, which again might reveal whether the user is signed into the target.
```
hamishwillee, line 23: Just FMI - how can the number of iframes in a page reveal whether the user is logged in? I guess if the page displays different content for logged in users.

It might be worth saying something along the lines at the top level that the attacker does need to have some knowledge of the target site or you wouldn't know this many frames means logged in, or what page to test for your load event. 

Note, I'm not sure, this is one of those "what went though my head as I read this reviews". What went through my head here is that this was a very specific search about a user interaction with a particular site, and therefore the risk (and hence protection effort) kind of depends on what your site is hiding.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+- [Leaking redirects with a CSP](leaking_redirects_with_a_csp): in this attack, the attacker creates a [Content Security Policy](/en-US/docs/Web/HTTP/Guides/CSP) which effectively blocks redirects, then tries to load a page in the target site. If the page is blocked, the attacker knows that the target redirected the request, and again, this may reveal that the user is signed in.
```
hamishwillee, line 24:
```suggestion
- [Leaking redirects with a CSP](#leaking_redirects_with_a_csp): in this attack, the attacker creates a [Content Security Policy](/en-US/docs/Web/HTTP/Guides/CSP) which effectively blocks redirects, then tries to load a page in the target site. If the page is blocked, the attacker knows that the target redirected the request, and again, this may reveal that the user is signed in.
```

hamishwillee, line 24: My understanding is that if a page tries to load a resource that is served with a CSP policy that the calling page violates, the browser simply doesn't load it, and fires a `securityviolation` event. What is this redirect we're talking about?

hamishwillee, line 24: >  If the page is blocked, the attacker knows that the target redirected the request, and again, this may reveal that the user is signed in.

I don't understand how this inference can be drawn. Might be obvious in the section. Is it that some pages are only served with CSP for logged in users or something similar - so you won't get the violation report for a non-logged in user?

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+
+script.src = url;
+document.head.appendChild(script);
+```
+
+An attacker may even be able to discover a user's ID, by iteratively trying to load pages to see if pages like `https://example.org/users/my_username` exist.
+
+### Frame counting using window references
+
+In a frame counting attack, the attacker finds out the number of frames currently loaded in the target page. In turn, that leaks information about the state of the target site, which may enable to attacker to learn, for example, whether the user is currently logged into the site.
```
hamishwillee, line 67: OK, so yes the attacker can draw inferences about a site when they know characteristics of the side that reflect certain states, such as logged in. IN this case the characteristic is the number of loaded frames.

files/en-us/web/security/attacks/xs-leaks/index.md
hamishwillee: Actually, maybe you're saying that the logged in user will be taken to the trusted page, and IF this page is served with CSP there will be a violation. If they are not logged in the user will be taken to another page - a redirect has occurred. So if that redirect page is not protected by some kind of CSP, then you will be seeing different behaviour, which tells you something about state.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+```js
+const target = document.querySelector("iframe").contentWindow;
+const frames = target.length;
+```
+
+### Leaking redirects with a CSP
+
+In this attack, the attacker creates a page with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that will generate a {{domxref("Element.securitypolicyviolation_event", "securitypolicyviolation")}} event if the page tries to embed a document that is not from the target URL.
+
+The attacker's page then creates an {{htmlelement("iframe")}} that embeds the target URL, and listens for the violation event. If the violation event fires, the attacker knows that a redirect took place, and this in turn may reveal something about the user's relationship with the site (for example, that they are signed in, or that they have some other status on the site).
```
hamishwillee, line 93: "attacker knows that a redirect took place" - see comment above on redirect https://github.com/mdn/content/pull/38977/files#r2030329489 - I'm not sure what is happening here and why is crystal clear - at least it isn't for me.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+</html>
+```
+
+Note that this attack works even if the target site disallows embedding using a mechanism such as [`frame-ancestors`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors).
+
+## Defenses against cross-site leaks
+
+Cross-site leaks exploit mechanisms in the web platform which enables websites to interact with each other. Correspondingly, the defenses against cross-site leaks all involve _isolating_ the target website from potential attackers, by disabling or controlling these cross-site interactions.
+
+Since cross-site leaks can work in many different ways, there isn't a single defense that will work against all of them. However, there are several practices that will work against many of them, and we will summarise them here.
```
hamishwillee, line 122: Don't we use US spellings?
```suggestion
Since cross-site leaks can work in many different ways, there isn't a single defense that will work against all of them. However, there are several practices that will work against many of them, and we will summarize them here.
```

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+app.get("/admin", (req, res) => {
+  if (isAllowed(req)) {
+    // Respond with the admin page if the user is admin
+    getAdminPage(req, res);
+  } else {
+    res.status(404).send("Not found.");
+  }
+});
+```
+
```
hamishwillee, line 177: So what specifically happens here that is different

```suggestion
With the above server code the page is never loaded: the problematic code will always get an error whether the user is logged in or not.
```

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+This means it is a good practice to prevent a site from being embeddable unless you need to allow embedding, and if you do need to allow embedding, restrict it as far as you can.
+
+There are two relevant tools here:
+
+- The [`frame-ancestors` directive](/en-US/docs/Web/HTTP/CSP#clickjacking_protection) in a [content security policy](/en-US/docs/Web/HTTP/CSP).
+- The {{httpheader("X-Frame-Options")}} response header.
+
+The `frame-ancestors` directive is a replacement for `X-Frame-Options`. By setting `X-Frame-Options` as well as `frame-ancestors`, you can prevent embedding in browsers that don't support `frame-ancestors`. As [browser support for `frame-ancestors` is very good](/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors#browser_compatibility), this is not a major concern.
+
+If `frame-ancestors` and `X-Frame-Options` are both set, then browsers that support `frame-ancestors` will ignore `X-Frame-Options`.
```
hamishwillee, line 203:
```suggestion
The `frame-ancestors` directive is a replacement for `X-Frame-Options`. By setting `X-Frame-Options` as well as `frame-ancestors`, you can prevent embedding in browsers that don't support `frame-ancestors` (if both are both set, then browsers that support `frame-ancestors` will ignore `X-Frame-Options`). As [browser support for `frame-ancestors` is very good](/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors#browser_compatibility), this is not a major concern.
```

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+const target = window.open("https://example.com");
+```
+
+The {{httpheader("Cross-Origin-Opener-Policy")}} response header determines whether a document will be opened in the same {{glossary("Browsing context", "browsing context group")}} as the document that opened it.
+
+If your server sends this header and sets it to any value except the default of `"unsafe-none"`, then if a document from a different origin tries to open your page using `window.open()`, your page will be loaded into a different browsing context group. Among other things, this means that the opener will not get a reference to the `window` object for your page, and will therefore not be able to use it in a frame counting attack.
+
+### Defense summary checklist
+
+As we've seen, cross-site leaks include a range of attacks targeting different parts of the web platform: a single defense doesn't work against any of them, and some leaks don't have any defenses yet.
```
hamishwillee, line 219: > and some leaks don't have any defenses yet.

Pedantic, but we have not seen this in this doc.

hamishwillee:
Looks good. The choices are reasonable and I think this should give a good flavour of the problem and the kinds of solutions. In particular when you expand a _little_ on how this information leaking might actually be a real problem. 

I don't fully understand spectre or any of the timing attacks. That makes it hard to understand whether or not it should be here. But my leaning is that if you try write a document and it looks a lot like this one, you should be able to slot it in later without much difficulty.

> which to pick for a third, and not sure the CSP one is a good choice.

I don't "know" either. I kind of like it because it's a side channel in a security mechanism.

As per comments, I'm still getting clear on exactly how this works. Either way though, what is the defense?

> I haven't talked about CORP since AFAICT CORP is mostly presented in connection with Spectre/Meltdown, which I haven't discussed here (again see https://github.com/mdn/content/issues/3526#issuecomment-2715961146 for why). But perhaps I should after all introduce it, maybe as a defense against the error event attack?

I would mention it as a single line - it is part of a defense in depth to restrict any cross origin loading to the minimum required so that the resource can still be used as intended.

> I'm conscious as I write these guides to different that we're getting a lot of duplication from one guide to another in the "Defenses" section, outlining for example how to do framing protection, or set SameSite. I think we might want to factor these out into separate guides at some point.

I do think it might be nice to have a separate guide "defense it depth handbook" that covers the common things. But you'll still have to link it from each of these sections, and I really think as long as we keep them constrained like this, the duplication does not hurt.

files/en-us/web/security/attacks/xs-leaks/index.md
torgo: I think the wording is a little unclear: "a class of attack which exploit the ways in which a target site can leak information to an attacker's site". Maybe start with a typical user story, an example of what information is (presumably unintentionally) leaking and then how an attacker exploits that leak?

torgo:
One general comment that comes to mind: there is a lot of discussion of iframes but no context of how these are used that might make it easier to understand the risks. E.g. iframes are used in payment flows and advertising, but there is no mention of either in the doc...?

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: > It is worth a following paragraph as to why a user or developer should care if any of this leaks, and perhaps some specific dangers from this knowledge [I had the same problem with the discussion of timing attacks on the wiki page - to motivate change people have to understand not just what the risk is, but the impact of the risk].
> 
> My first reaction was that I don't see the risk. On further thought, I guess if I know that the user has a login with bank say, that makes them a better target for trying some of the other attacks, or even an old-style scam where I send a letter - but in this case I know that they will be in that bank, and I might know their id, which adds to the trustworthyness.
> 
> Further, if I know their id, perhaps that allows a dark web search for passwords etc. Dunno, but worth making it real world.

It's a good point. AIUI these attacks are usually less severe than something like an XSS or even a CSRF attack. There are a couple of things worth mentioning:
- the mere fact that someone knows you have an active account somewhere could put you in peril, and certainly might be something you don't want people to know about you. It could open you up to retaliation from an oppressive government (for example, if you are seeking an abortion) or open you up to extortion.
- as you've noted, it would make a phishing attack (for example) much more convincing if someone knows your ID

Worth mentioning these I agree.

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: This is all true, and yes, AIUI the idea is that the _if_ target include a a different number of iframes for logged-in users (which is very plausible), then the number of iframes gives that away, and so yes, this is a bespoke attack for a particular target (but many of these attacks are like that).

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: So yes, the idea is that the attack leaks whether there was a redirect, and that lets the attacker know whether the user is signed in.

For instance:

- if I am signed into https://frontendmasters.com/ and I visit https://frontendmasters.com/dashboard/, there's no redirect - it just shows my dashboard.
- if I am not signed in, then it redirects me to https://frontendmasters.com/login/

But it's also not just signed in/not signed in of course. There might be some special status someone has on a site that this might leak.

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: Actually the CSP one doesn't have any defenses AFAIK. Should probably say that.

wbamberg:
> Looks good. The choices are reasonable and I think this should give a good flavour of the problem and the kinds of solutions. In particular when you expand a _little_ on how this information leaking might actually be a real problem.
> 
> I don't fully understand spectre or any of the timing attacks. That makes it hard to understand whether or not it should be here. But my leaning is that if you try write a document and it looks a lot like this one, you should be able to slot it in later without much difficulty.
> 
> > which to pick for a third, and not sure the CSP one is a good choice.
> 
> I don't "know" either. I kind of like it because it's a side channel in a security mechanism.
> 

Yeah I like that aspect of it as well. One of the things that make xs-leaks hard to defend against is that things that should protect you can end up hurting you.

> As per comments, I'm still getting clear on exactly how this works. Either way though, what is the defense?

As I say, AFAIK there isn't one.

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: -> https://github.com/mdn/content/pull/38977/commits/575018586eb90ef291a59889b6b3cbc46c0156da

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: -> https://github.com/mdn/content/pull/38977/commits/67f993d20f8ca01aaf6eb43f1f0019942b752a3e

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: The target doesn't have the CSP, the attacker does. The attacker creates a CSP - for its page - which only allows the target page to be loaded, then tries to load the target page.

wbamberg: The CSP is controlling the attacker page, not the target page.

See the example: https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/XS-leaks#leaking_redirects_with_a_csp. 

This is the attacker page, and it contains a CSP in the `<meta>` tag. The CSP says that this page (i.e. the attacker page) can only load the target page to be loaded into an `<iframe>`. The attacker page then tries to load the target page into an `<iframe>`.

- If this does not generate a `securitypolicyviolation` event, then the target did not redirect.
- If it does generate a `securitypolicyviolation` event, then the target did redirect.

The doc does say this:

> In this attack, the attacker creates a page with a [Content Security Policy (CSP)](https://pr38977.review.mdn.allizom.net/en-US/docs/Web/HTTP/Guides/CSP) that will generate a [securitypolicyviolation](https://pr38977.review.mdn.allizom.net/en-US/docs/Web/API/Element/securitypolicyviolation_event) event if the page tries to embed a document that is not from the target URL.

hamishwillee: Yes, I see it now, but I needed this explanation to understand the code. I think sometimes I'm a bit dim, but I can't be along. Maybe we can improve it. https://github.com/mdn/content/pull/38977/files#r2032522590

wbamberg: I'm happy to make the change you suggested, but I don't expect people to be able to understand the attack from the little intro paragraph at https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/XS-leaks#sample_cross-site_leaks - if I did, I wouldn't need the full description and code sample further down.

This `<ul>` is really just giving a preview of what we will talk about in the section. It might even be better to remove the explanation and just make it a list of fragment links:

> - [Leaking page existence using error events](https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/XS-leaks#leaking_page_existence_using_error_events)
> - [Frame counting using window references](https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/frame_counting_using_window_references)
> - [Leaking redirects with a CSP](https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/XS-leaks#leaking_redirects_with_a_csp)

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: Well for example, someone might have special status that's more fine-grained than just "logged in", like platinum status on a banking site (whatever that means).

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,242 @@
...
+- [Frame counting using window references](frame_counting_using_window_references): in this attack, the attacker gets a reference to a {{domxref("Window", "window")}} object hosting a page in the target site, for example as the return value of a call to {{domxref("Window.open()", "window.open()")}}. The attacker can then determine the number of {{htmlelement("iframe")}} elements in the target page, which again might reveal whether the user is signed into the target.
+- [Leaking redirects with a CSP](#leaking_redirects_with_a_csp): in this attack, the attacker's page has a [Content Security Policy](/en-US/docs/Web/HTTP/Guides/CSP) which effectively blocks redirects, then tries to load a page in the target site. If the page is blocked, the attacker knows that the target redirected the request, and again, this may reveal that the user is signed in.
```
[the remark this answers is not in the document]
wbamberg: I'm happy to make the change you suggested, but I don't expect people to be able to understand the attack from the little intro paragraph at https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/XS-leaks#sample_cross-site_leaks - if I did, I wouldn't need the full description and code sample further down.

This `<ul>` is really just giving a preview of what we will talk about in the section. It might even be better to remove the explanation and just make it a list of fragment links:

> - [Leaking page existence using error events](https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/XS-leaks#leaking_page_existence_using_error_events)
> - [Frame counting using window references](https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/frame_counting_using_window_references)
> - [Leaking redirects with a CSP](https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/XS-leaks#leaking_redirects_with_a_csp)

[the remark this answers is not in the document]
wbamberg: Hm, I just noticed that your suggestion added a new paragraph, I don't really like this here, I think it slows things down to go into too much detail about the attacks at this point. Maybe a bare list as suggested above is better.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,235 @@
...
+
+### SameSite cookies
+
+The [`SameSite`](/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value) cookie attribute determines whether or not the cookie will be sent in requests originating from a different site.
+
+The `Lax` value of `SameSite` means that cross-site requests will only include the cookie if the request is a top-level navigation (meaning, essentially, that the value in the browser's address bar changes to the target site) and uses a {{glossary("Safe/HTTP", "safe")}} method (most notably, this excludes {{httpmethod("POST")}} requests).
+
+This can protect against some cross-site leaks. For example, the [Leaking page existence using error events](#leaking_page_existence_using_error_events) attack depends on the attacker making cross-site resource requests that include the user's cookies. Setting `SameSite` to `Lax` on the user's session cookie would prevent this attack, because the cookie would not be included in the attacker's request, so the admin page would never be returned.
```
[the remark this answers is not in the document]
wbamberg: I think 'even" should be "ever" here, because the point is that the page is not returned whether or not the user is logged in.

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: I've tried to break this down some more in https://github.com/mdn/content/pull/38977/commits/e04b3dedd67c478fb423169c545163789c52ad02.

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: -> https://github.com/mdn/content/pull/38977/commits/c854152cd32e9498d05a4db5dcd65887644dad70

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -95,17 +95,25 @@ const frames = target.length;
 
 ### Leaking redirects with a CSP
 
-In this attack, the attacker creates a page with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that will generate a {{domxref("Element.securitypolicyviolation_event", "securitypolicyviolation")}} event if the page tries to embed a document that is not from the target URL.
+In some websites, the server will redirect a request, or not, based on whether the user is signed in (or has some special status on the site). For example, imagine a site which shows administrators a page at `https://example.org/admin`. If the user is not signed in, and requests this page, then the server might redirect them to `https://example.org/login`.
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
```

files/en-us/web/security/attacks/xs-leaks/index.md
hamishwillee: Yes, this is really good. I like the use of the admin here and explanation - previously admin login was used in an example and it looked pointless. Now it it is clear that this tells you not just that they are logged in but that they are an admin - compromising this login is more worthwhile than others.

files/en-us/web/security/attacks/xs-leaks/index.md
hamishwillee: Or admin, as you showed in the other example.

files/en-us/web/security/attacks/xs-leaks/index.md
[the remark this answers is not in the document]
hamishwillee: FWIW I like the way it is now. The previous explanation confused me, and this does not.

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: -> https://github.com/mdn/content/pull/38977/commits/beae2f35d07741d4946a9c1c380ff4d2f551a4f8

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: > I think a web platform API is a side channel in these cases (i.e. is it worth differentiating)?

yes, tbh I started being uncertain about what exactly counts as a side channel, so I reworked the definition a bit in https://github.com/mdn/content/pull/38977/commits/36d919efdac4826b7fbc47630e26471a912e894b .

> Maybe start with a typical user story, an example of what information is (presumably unintentionally) leaking and then how an attacker exploits that leak?

I'm happy to do more work on the definition if it is unclear but I really do want to start with a concise definition rather than a story.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,254 @@
...
+const target = window.open("https://example.com");
+```
+
+The {{httpheader("Cross-Origin-Opener-Policy")}} response header determines whether a document will be opened in the same {{glossary("Browsing context", "browsing context group")}} as the document that opened it.
+
+If your server sends this header and sets it to any value except the default of `"unsafe-none"`, then if a document from a different origin tries to open your page using `window.open()`, your page will be loaded into a different browsing context group. Among other things, this means that the opener will not get a reference to the `window` object for your page, and will therefore not be able to use it in a frame counting attack.
+
+### Defense summary checklist
+
+As we've seen, cross-site leaks include a range of attacks targeting different parts of the web platform: a single defense doesn't work against any of them. Indeed, some leaks, such as the one that exploits CSP to leak redirects, don't have any defenses yet.
```
hamishwillee, line 238: I created a comment against a commit, but it doesn't seem to be visible here. Might be irrelevant

But while there is no way to stop CSP from revealing the redirect, you aren't forced to have redirects. You might, for example, use the same page with frames and different content for a logged in users.

hamishwillee (approved):
IMO this is a useful entry point that positions the problem appropriate, doesn't swamp the reader, and doesn't duplicate other resource. I feel I have a good idea of what kinds of information might leak, what an attacker might do with that information, broad steps I might take to reduce the risk, and I can work out where to go for more information.

Approving, so you can merge when you think inflow of useful review has ceased.

wbamberg:
@mozfreddyb , would love you to take a look when you get the chance.

hamishwillee:
It would be great to get this in. @mozfreddyb Gentle ping for review as per ^^^

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,254 @@
...
+const url = "https://example.org/admin";
+const script = document.createElement("script");
+
+script.addEventListener("load", (e) => {
+  console.log(`${url} exists`);
+});
+
+script.addEventListener("error", (e) => {
+  console.log(`${url} does not exist`);
+});
```
mozfreddyb, line 64: I think you need to explain why script? Can you give an example of a response the web page may give that is an error/non-error depending on user state?

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,254 @@
...
+
+script.src = url;
+document.head.appendChild(script);
+```
+
+An attacker may even be able to discover a user's ID, by iteratively trying to load pages to see if pages like `https://example.org/users/my_username` exist.
+
```
mozfreddyb, line 71: I think it's important to note here (or above) that this is supposed to be a user-specific page, not "my_username" as a thing that is to be replaced. Maybe something like "example.org/personal-profile-settings" would be less nuanced?

mozfreddyb (commented):
This looks great (sorry about the long wait)

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: > I think you need to explain why script? 

Isn't this explained at line 42:

> The attack relies on the ability of a website to load a resource from another site, for example by setting the `src` attribute of a {{htmlelement("script")}} element to the URL of the resource:

...or am I misunderstanding you? 

>  Can you give an example of a response the web page may give that is an error/non-error depending on user state?

I think I might need you to explain this a bit more for me. AIUI, in this scenario:
- if the user is logged in, then the page is returned, and `load`is fired
- if the user isn't logged in, then the server returns an error (e.g. 404) and `error` is fired

I sort of say that in 52, but is there something else/something more I need to say?

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: To check my understanding here: in this case we have a website that includes the the username in the URL for a page, so if the username is `fishfingers` and the user is logged in, then the page `https://example.org/users/fishfingers` exists. So an attacker can repeatedly try to load pages with test IDs, and if a page exits, then they know not only that the current user is logged in, but what their username is.

So (if this is right) maybe it's worth explaining it a bit more, with some version of that paragraph? Does that address your comment?

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,254 @@
...
+The attacker can get a `Window` object by calling {{domxref("Window.open()", "window.open()")}}:
+
+```js
+const target = window.open("https://example.org");
+const frames = target.length;
+```
+
+Alternatively, the attacker can embed the target site in an {{htmlelement("iframe")}}, and retrieve the frame's {{domxref("HTMLIFrameElement.contentWindow", "contentWindow")}} property:
+
```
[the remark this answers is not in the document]
wbamberg: I'd prefer to keep defenses and attacks separate. I mean otherwise, why mention XFO here and not COOP? I do mention frame protection as a defense against this attack, in https://pr38977.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/XS-leaks#framing_protection.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,254 @@
...
+- Send the `Cross-Origin-Opener-Policy` response header to prevent other sites from accessing your `window` global object.
+
+## See also
+
+- [XS-Leaks Wiki](https://xsleaks.dev/) (xsleaks.dev)
+- [Cross-site leaks Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/XS_Leaks_Cheat_Sheet.html) (OWASP)
+
+<section id="Quick_links">
+{{ListSubpages("/en-US/docs/Web/Security", "1", "0", "1")}}
+</section>
```
bsmth, line 254: We can use the `security` sidebar now:

```suggestion
```

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,254 @@
+---
+title: Cross-site leaks (XS-leaks)
+slug: Web/Security/Attacks/XS-leaks
+page-type: guide
+---
```
bsmth, line 5:
```suggestion
sidebar: security
---
```

martinakraus:
> This results in an HTTP request to the https://example.org/ website. If the request includes cookies that the site uses to identify users, and the page requested is only available to logged-in users, then the success or failure of the request reveals whether or not the user is logged in.

Is that really the only possible conclusion? I'm just wondering—if the SameSite attribute is set to Strict, the cookie wouldn't be sent either, right?
Or does it mean that the URL https://example.org/admin is already considered same-origin in this context?

Overall, the introduction to these types of attacks is very well written. The section on defense mechanisms in particular provides a clear and comprehensible overview of possible mitigations.

wbamberg:
> > This results in an HTTP request to the https://example.org/ website. If the request includes cookies that the site uses to identify users, and the page requested is only available to logged-in users, then the success or failure of the request reveals whether or not the user is logged in.
> 
> Is that really the only possible conclusion? I'm just wondering—if the SameSite attribute is set to Strict, the cookie wouldn't be sent either, right? Or does it mean that the URL https://example.org/admin is already considered same-origin in this context?

Right, AFAIK SameSite is a defense here, which I do note at line 206: 

> This can protect against some cross-site leaks. For example, the [Leaking page existence using error events](#leaking_page_existence_using_error_events) attack depends on the attacker making cross-site resource requests that include the user's session cookies. Setting `SameSite` to `Lax` on the user's session cookie would prevent this attack, because the cookie would not be included in the attacker's request, and no pages that require a login would ever be returned.

So in the first half of the doc, that describes the attacks, there's an implicit "if you don't implement any defenses". I think it's hard to avoid from this, given the overall doc structure of presenting first attacks, then defenses.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+- If the user is logged in as an admin, then the `<iframe>` will load, and the browser will not fire `securitypolicyviolation`.
+- If the user is not logged in as admin, the server redirects to `https://example.org/login`. Because this URL is not allowed by the attacker's CSP, the browser will block the `<iframe>` and fire the `securitypolicyviolation` event, and the attacker's event handler will run.
+
+Note that this attack works even if the target site disallows embedding using a mechanism such as [`frame-ancestors`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors).
+
+## Defenses against cross-site leaks
+
+Cross-site leaks exploit mechanisms in the web platform which enables websites to interact with each other. Correspondingly, the defenses against cross-site leaks all involve _isolating_ the target website from potential attackers, by disabling or controlling these cross-site interactions.
```
terjanq, line 138: > the defenses against cross-site leaks all involve

Not all protections involve isolating the target. For example, adding XSRF tokens to subresources do not isolate the target but prevent an attacker from probing these resources.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,251 @@
...
+const script = document.createElement("script");
+script.src = "https://example.org/admin";
+document.head.appendChild(script);
+```
+
+A server can use Fetch metadata to deny these requests, as in the following [Express](/en-US/docs/Learn_web_development/Extensions/Server-side/Express_Nodejs) code:
+
+```js
+function isAllowed(req) {
+  // Allow same-origin, same-site, and user-initiated requests
```
terjanq, line 166: "none" are either user initiated or browser initiated, _I think_.

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,251 @@
...
+There are two relevant tools here:
+
+- The [`frame-ancestors` directive](/en-US/docs/Web/HTTP/CSP#clickjacking_protection) in a [content security policy](/en-US/docs/Web/HTTP/CSP).
+- The {{httpheader("X-Frame-Options")}} response header.
+
+The `frame-ancestors` directive is a replacement for `X-Frame-Options`. Although [browser support for `frame-ancestors` is very good](/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors#browser_compatibility), some very old browsers, notably Internet Explorer, don't support `frame-ancestors`.
+
+If `frame-ancestors` and `X-Frame-Options` are both set, then browsers that support `frame-ancestors` will ignore `X-Frame-Options`. This means that there's no reason not to set `X-Frame-Options` as well as `frame-ancestors`, and thus prevent embedding even in browsers that don't support `frame-ancestors`.
+
+### Cross-Origin Opener Policy (COOP)
```
terjanq, line 225: I would maybe suggest mentioning that COOP in many cases limits the attacks to 1 bit of information per user activation because the reference is "burned" after COOP headers are encountered. Example 1 bit attacks that are possible with COOP are timing based attacks - you can measure the time it takes to lose reference to the window.

terjanq (commented):
Hey! I left some comments, but overall looks really good! :)

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,254 @@
...
+      document.addEventListener("securitypolicyviolation", () => {
+        console.log("Page was redirected");
+      });
+      const frame = document.createElement("iframe");
+      document.body.appendChild(frame);
+      frame.src = "https://example.org/admin";
+    </script>
+  </body>
+</html>
+```
```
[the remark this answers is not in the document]
wbamberg: Ah, could be. I don't think I tested this with a same-origin redirect, I tested with GitLab, which redirects gitlab.com to about.gitlab.com for not-logged-in visitors. So I suppose I could change this to admin.example.org?

Still I am surprised it wouldn't work same-origin. Doesn't `frame-src https://example.org/admin` forbid `https://example.org/login`, per https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#host-source ?

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: Could you elaborate what the difference is? The spec only talks about user-initiated (https://w3c.github.io/webappsec-fetch-metadata/#sec-fetch-site-header)  and links to come more detail at https://w3c.github.io/webappsec-fetch-metadata/#directly-user-initiated.

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: I wonder if this is a bit too advanced/nuanced for this guide.

files/en-us/web/security/attacks/index.md
```diff
@@ -12,6 +12,8 @@ This page links to pages explaining how some common attacks work, and how they c
   - : In a clickjacking attack, an attacker creates a decoy site which embeds the target site inside an {{htmlelement("iframe")}} element. It hides the `<iframe>`, and overlays some decoy elements on top of it. When the user interacts with these decoy elements, they are inadvertently interacting with the target site, and may be tricked into performing actions with the target site which they did not intend.
 - [Cross-site request forgery (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF)
   - : In a cross-site request forgery (CSRF) attack, an attacker tricks the user or the browser into making an HTTP request to the target site from a malicious site. The request includes the user's credentials and causes the server to carry out some harmful action, thinking that the user intended it.
+- [Cross-site leaks (XS-leaks)](/en-US/docs/Web/Security/Attacks/XS-leaks)
+  - : Cross-site leaks (XS-leaks) are a class of attack which exploit the ways in which a target site can leak information to an attacker's site, either through web platform APIs or through {{glossary("side channels")}} such as timing attacks.
```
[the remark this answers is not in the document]
wbamberg: -> https://github.com/mdn/content/pull/38977/commits/5ce8cb68c283343ad99cac82d33759ecf4341488

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,254 @@
...
+
+In this guide we will not attempt to describe every cross-site leak attack and defense. Instead, we'll start by describing a few example attacks, then outline the common underlying weaknesses that enable them, then describe some general defenses that can work against many known attacks.
+
+## Sample cross-site leaks
+
+In this section we'll describe three different cross-site leaks, to give an idea of how they work.
+
+- [Leaking page existence using error events](#leaking_page_existence_using_error_events): in this attack, an attacker can determine whether particular pages in the target site exist, by attempting to load them as resources and listening for the {{domxref("HTMLElement/error_event", "error")}} and {{domxref("HTMLElement/load_event", "load")}} events. If certain pages are only available to logged-in users, the attacker can determine whether the user is signed into the target site.
```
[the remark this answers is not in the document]
wbamberg: -> https://github.com/mdn/content/pull/38977/commits/3566fa161c65e7ef7e107a794a9ddf03ed05ed17

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,254 @@
...
+
+All three attacks are deployed in the same way: the attacker crafts a page that implements the attack, then persuades the user to visit the page, for example by sending them an email or sharing a post containing the link. When the user visits the page the attack executes automatically.
+
+In the rest of this section we'll describe these three attacks in a little more detail, to give you a concrete sense of how they work. Although the three attacks target quite different parts of the web platform, they have a common underlying cause: the extent to which the browser enables websites to connect to and interact with each other by mechanisms such as framing, loading subresources, or opening new windows.
+
+> [!NOTE]
+> For a more complete catalog of cross-site leaks, see the [XS-Leaks Wiki](https://xsleaks.dev/) and the [OWASP Cross-site Leaks Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/XS_Leaks_Cheat_Sheet.html).
+
+### Leaking page existence using error events
```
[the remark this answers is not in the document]
wbamberg: -> https://github.com/mdn/content/pull/38977/commits/3566fa161c65e7ef7e107a794a9ddf03ed05ed17

files/en-us/web/security/attacks/xs-leaks/index.md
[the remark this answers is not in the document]
wbamberg: -> https://github.com/mdn/content/pull/38977/commits/944a50597215f4a80737aa2e47abd5e474310835

files/en-us/web/security/attacks/xs-leaks/index.md
wbamberg: -> https://github.com/mdn/content/pull/38977/commits/ccf9aa32ecf5abddbf518ec386369d40d34fe221

wbamberg:
Thank you for the review @terjanq ! I made most of the changes and commented on a couple. Please let me know if you're happy with this.

files/en-us/web/security/attacks/xs-leaks/index.md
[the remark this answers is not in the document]
terjanq: > Still I am surprised it wouldn't work same-origin. Doesn't frame-src https://example.org/admin forbid https://example.org/login

It does, but it onlly applies to the original URL not the one after redirects. This is to prevent information leaks if I remember correctly.

files/en-us/web/security/attacks/xs-leaks/index.md
terjanq: Didn't know that the spec tries to be that explicit about "none" and I'm surprised there is no mention about browser initiated requests but maybe I'm hallucitating. I thought that some of the background requests the browser sends come with `Sec-Fetch-Site: none` but maybe the headers are just omitted in that scenario. Let's leave it how it is for now :)

files/en-us/web/security/attacks/xs-leaks/index.md
terjanq: From the experience COOP cause a little confusion because it's presented as a complete mitigation against navigation based xsleaks but it's a great to mitigate the risks rather than fully mitigate them. It is a little bit advanced/nuanced so it's your call in the end whether it's worth mentioning the nuance :)

terjanq:
> Thank you for the review @terjanq ! I made most of the changes and commented on a couple. Please let me know if you're happy with this.

Looks great, thanks!

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -0,0 +1,251 @@
...
+Cross-site leaks (also called XS-Leaks) are a class of attack in which an attacker's site can derive information about the target site, or about the user's relationship with the target site, by using web platform APIs that enable sites to interact with one another. The information leaked could include, for example:
+
+- Whether the user has visited the target site.
+- Whether the user is logged into the target site.
+- What the user's ID on the site is.
+- What the user has recently searched for on the site.
+
+This might seem to be a much less damaging problem than, for example, a [cross-site scripting](/en-US/docs/Web/Security/Attacks/XSS) attack, but it can still have serious consequences for users. For example:
+
+- A user might have accounts on websites that they don't want to make public. Leaking this information to an attacker could expose them to extortion or retaliation from an oppressive government (for example, against a user seeking information about abortions).
```
bsmth, line 17: Could we rephrase this?

```suggestion
- A user might have accounts on websites that they don't want to make public. Leaking this information to an attacker could expose them to extortion or retaliation from an oppressive government (for example, against a user seeking information about specific medical procedures).
```

wbamberg:
I'm merging this: it's has a tick from Hamish and an implicit tick from one expert reviewer. There are a few unresolved comments but I'd like to take these as follow-ups if needed - in general people seem pretty happy with it.

mozfreddyb:
Sorry I did not find the time to go through the article more thoroughly before it was merged. This looks great!
I really appreciate you jumping in @terjanq 😍

PR #38979, opened 2025-04-04
Fix <hue-interpolation-method> BCD query

Josh-Cena:
The BCD is broken, chiefly because `css.types.image.gradient` is now `css.types.gradient`; still I think it's not too useful to have so many queries, so I'm reducing the list to a single item.


bsmth (approved):
Looks good to me, tnx.

CCing @yarusome who worked on these previously

PR #38980, opened 2025-04-04
Remove external Web Speech API articles

wbamberg:
Based on a conversation about the external links in this page, and related to https://github.com/mdn/content/pull/38930.

Why? It doesn't add anything that's not in the page, all the code is 2014-style and has not been updated.

I would also be happy to remove the SitePoint article if people want me to do that.


Josh-Cena:
> I would also be happy to remove the SitePoint article if people want me to do that.

I won't object to that. We already have good coverage of this area with some great examples, so I don't think external articles on the same topic can be much greater.

Josh-Cena (approved):
:+1:

PR #38981, opened 2025-04-04
Fix grammar

toocomputer:
### Description

This is a simple grammatical fix.

### Motivation

Enforces correct grammar, which leads to better readability and clarity.

N/A


Josh-Cena (approved):
Good catch, thanks

PR #38986, opened 2025-04-05
Remove uncovered topic from the introduction

youssef-el-atmani:
The introduction claims that the article will cover the `description list` element, but the article actually don't.


Josh-Cena (approved):
Good catch, thanks

PR #38029, opened 2025-02-07
Be a lot more precise when talking about URL component getter/setter

Josh-Cena:
- Fix https://github.com/mdn/content/issues/37834
- Fix https://github.com/mdn/content/issues/37165
- Fix https://github.com/mdn/content/issues/37003
- Fix https://github.com/mdn/content/issues/8895

PR #39000, opened 2025-04-06
Remove `port` mention for URL `protocol` documentation

Marcono1234:
### Description

Remove the mention of the `port` property from the URL `protocol` documentation

### Motivation

Current documentation for `protocol` was confusing because it said "**this property** contains an empty string", but actually referred to the `port` property?

Possibly a copy & paste error?

Was introduced by #38029
@Josh-Cena, was the original change intentional, and am I maybe misunderstanding something here?


Josh-Cena (approved):
Thanks, great catch! I believe this was a copy paste mistake

PR #39001, opened 2025-04-06
Update syntax info of `view-timeline`

skyclouds2001:
### Description

`view-timeline` can also accept `view-timeline-inset` as a longhand property

### Additional details

https://drafts.csswg.org/scroll-animations/#view-timelines-named


estelle (approved):
Thanks! 🎉

PR #39002, opened 2025-04-06
Editorial review: Add moveBefore() docs

chrisdavidmills:
### Description

Chrome 133 supports the `moveBefore()` method; see https://chromestatus.com/feature/5135990159835136.

This PR documents the new method on `Document`, `DocumentFragment`, and `Element`.


noamr:
Thanks! A few technical comments:

- What is preserved for `<dialog>` is its modal state. See https://github.com/web-platform-tests/wpt/blob/e3e6eee31d98d5b5dea1a3159febe139280c10ad/dom/nodes/moveBefore/modal-dialog.html
- Selection is a bit of an odd one; It only preserves selection in certain situations - namely when the anchor node of the selection is the target of the `moveBefore`.

- This is missing a section about custom elements. By default, a custom elements that get moved using `moveBefore`, receive their `disconnectedCallback` and `connectedCallback` invoked. To have custom support for state-preserving move, an author can add a `connectedMoveCallback` that would be called instead of connected/disconnected. It can be empty to avoid the disconnection/connection side-effects, or contain logic that responds to a move.

- We should say something about the constraints: `moveBefore` only works within the same document, and it cannot work as a way to connect or disconnect an element. Moving a disconnected element to a connected parent or vice versa, would result in an exception. Users of this API that want to be resilient to this need to know in advance whether they're acting across disconnected/connected state, across documents, and use `insertBefore` instead in these cases (or try/catch the `moveBefore`).

chrisdavidmills:
> Thanks! A few technical comments:
> 
>     * What is preserved for `<dialog>` is its modal state. See https://github.com/web-platform-tests/wpt/blob/e3e6eee31d98d5b5dea1a3159febe139280c10ad/dom/nodes/moveBefore/modal-dialog.html

Cool, thanks for the tip; I've updated the list of preserved state things to account for this.

Note that, in light of your comments, I've moved this content to a new "Description" section on the page and added the other bits you wanted me to include there. 

> 
>     * Selection is a bit of an odd one; It only preserves selection in certain situations - namely when the anchor node of the selection is the target of the `moveBefore`.

I did a bit of testing on this but couldn't get the selection to preserve. I've just removed the bullet about preserving selection, as this doesn't seem to me to be a case that will come up often or behavior that can be particularly relied upon. What do you think?

> 
>     * This is missing a section about custom elements. By default, a custom elements that get moved using `moveBefore`, receive their `disconnectedCallback` and `connectedCallback` invoked. To have custom support for state-preserving move, an author can add a `connectedMoveCallback` that would be called instead of connected/disconnected. It can be empty to avoid the disconnection/connection side-effects, or contain logic that responds to a move.

I've added a section to cover this inside the "Description" section, and added some info about `connectedMoveCallback()` to the custom elements page. I didn't add it to the `Document.moveBefore()` page as it didn't seem so relevant there, just `DocumentFragment...` and `Element...`
 
>     * We should say something about the constraints: `moveBefore` only works within the same document, and it cannot work as a way to connect or disconnect an element. Moving a disconnected element to a connected parent or vice versa, would result in an exception. Users of this API that want to be resilient to this need to know in advance whether they're acting across disconnected/connected state, across documents, and use `insertBefore` instead in these cases (or try/catch the `moveBefore`).

I've added a section about constraints to the "Description" section, and also tried to move sure the relevant exceptions are represented in the "Exceptions" section.

> 
>     * As far as mutation observers go, there is no particular change, but it might be worthwhile to note that the moved element would appear as removed/re-inserted in mutation observers.

Good call; I've added a note about this.

chrisdavidmills:
@noamr ^

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,84 @@
...
+- `TypeError` {{jsxref("TypeError")}}
+  - : The second argument was not supplied.
+
+## Description
+
+The `moveBefore()` method moves a given node to a new place in the DOM. It provides similar functionality to the {{domxref("Node.insertBefore()")}} method, except that it doesn't remove and then reinsert the node. This means that the state of the node (which would be reset if moving it with `insertBefore()` and similar mechanisms) is preserved after the move. This includes:
+
+- [Animation](/en-US/docs/Web/CSS/CSS_animations) and [transition](/en-US/docs/Web/CSS/CSS_transitions) state.
+- {{htmlelement("iframe")}} loading state.
+- Interactive node {{cssxref(":focus")}} and {{cssxref(":active")}} states.
```
noamr, line 49: Perhaps reword to "interactivity"

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,217 @@
...
+- [Fullscreen](/en-US/docs/Web/API/Fullscreen_API) element state.
+- Open/close state of [popovers](/en-US/docs/Web/API/Popover_API).
+- Modal state of {{htmlelement("dialog")}} elements (modal dialog will not be closed).
+
+The play state of {{htmlelement("video")}} and {{htmlelement("audio")}} elements is not included in the above list, as these elements retain their state when removed and reinserted, regardless of the mechanism used.
+
+> [!NOTE]
+> When observing changes to the DOM using a {{domxref("MutationObserver")}}, nodes moved with `moveBefore()` will be recorded with a [removed node](/en-US/docs/Web/API/MutationRecord/removedNodes) and an [added node](/en-US/docs/Web/API/MutationRecord/addedNodes).
+
+### `moveBefore()` constaints
```
noamr, line 59: constraints

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,217 @@
...
+There are some constraints to be aware of when using `moveBefore()`:
+
+- It can only work when moving a node within the same document.
+- It won't work if you try to move a node that is not connected to the DOM to an already connected parent, or vice versa.
+
+In such cases, `moveBefore()` will fail with a `HierarchyRequestError` exception. If the above constraints are requirements for your particular use case, you should use {{domxref("Node.insertBefore()")}} instead, or use [`try...catch`](/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) to handle the errors that arise from such cases.
+
+### Moving custom elements
+
+When moving a [custom element](/en-US/docs/Web/API/Web_components/Using_custom_elements), care needs to be taken to handle the `connectedCallback()` and `disconnectedCallback()` [lifecycle callbacks](/en-US/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks) appropriately, if any are defined in the custom element's constructor.
```
noamr, line 70: "care needs to be taken" is not the right framing here.
By default, custom elements don't support state-preserving move, and their `disconnected` and `connected` callbacks are fired.

Authors of web components can *opt in* to state-preserving move behavior, by implementing the `connectedMoveCallback`, which would be called instead of the disconnected+connected default.

Note: this default behavior is introduced in order to maintain backwards compatibility with existing custom elements who might rely on being disconnected and reconnected when moved around the DOM.

files/en-us/web/api/web_components/using_custom_elements/index.md
```diff
@@ -57,6 +57,7 @@ Custom element lifecycle callbacks include:
 
 - `connectedCallback()`: called each time the element is added to the document. The specification recommends that, as far as possible, developers should implement custom element setup in this callback rather than the constructor.
 - `disconnectedCallback()`: called each time the element is removed from the document.
+- `connectedMoveCallback()`: called _instead_ of `connectedCallback()` and `disconnectedCallback()` each time the element is moved to a different place in the DOM via {{domxref("Element.moveBefore()")}}. This can be used to overcome functionality problems encountered as a result of initialization/cleanup code being run multiple times when the element is not actually being added to or removed from the DOM. See [Moving custom elements](/en-US/docs/Web/API/Element/moveBefore#moving_custom_elements) for more details.
```
noamr, line 60: * When defined, called _instead etc..

files/en-us/web/api/document/movebefore/index.md
chrisdavidmills: Updated to

> - Interactivity states (for example, {{cssxref(":focus")}} and {{cssxref(":active")}}).

files/en-us/web/api/documentfragment/movebefore/index.md
chrisdavidmills: Oops! Fixed.

files/en-us/web/api/documentfragment/movebefore/index.md
chrisdavidmills: OK, I get this now; thanks for the explanation. I've reworded the section as follows:

---

### Moving custom elements

Each time a [custom element](/en-US/docs/Web/API/Web_components/Using_custom_elements) is moved via `moveBefore()` (or similar methods such as {{domxref("Node.insertBefore()")}}), the `disconnectedCallback()` and `connectedCallback()` [lifecycle callbacks](/en-US/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks) are fired.

Consider this minimal example:

```js
class MyCustomElement extends HTMLElement {
  constructor() {
    super();
  }

  connectedCallback() {
    console.log("Custom element added to page.");
  }

  disconnectedCallback() {
    console.log("Custom element removed from page.");
  }
}

customElements.define("my-custom-element", MyCustomElement);
```

In this case, "Custom element removed from page." and "Custom element added to page." are logged to the console with each move. This might be your intended behavior. However, if you use the callbacks to run initialization and cleanup code, it might cause problems with the state of the moved element.

Custom elements can be opted in to state-preserving moves via the `connectedMoveCallback()` lifecycle callback. If defined in the constructor, this will run instead of `connectedCallback()` and `disconnectedCallback()` when an element instance is moved via `moveBefore()`. You could add an empty `connectedMoveCallback()` to stop the other two callbacks running, or include some custom logic to handle the move:

```js
connectedMoveCallback() {
  console.log("Custom element moved with moveBefore()");
}
```

noamr: Perfect thanks!

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,215 @@
...
+
+In this case, "Custom element removed from page." and "Custom element added to page." are logged to the console with each move. This might be your intended behavior. However, if you use the callbacks to run initialization and cleanup code, it might cause problems with the state of the moved element.
+
+Custom elements can be opted in to state-preserving moves via the `connectedMoveCallback()` lifecycle callback. If defined in the constructor, this will run instead of `connectedCallback()` and `disconnectedCallback()` when an element instance is moved via `moveBefore()`. You could add an empty `connectedMoveCallback()` to stop the other two callbacks running, or include some custom logic to handle the move:
+
+```js
+connectedMoveCallback() {
+  console.log("Custom element moved with moveBefore()");
+}
+```
```
wbamberg, line 100: I think this could be more concise and sort of buries the lede, especially with:

> This might be your intended behavior. However, if you use the callbacks to run initialization and cleanup code, it might cause problems with the state of the moved element. 

...when AIUI these callbacks are primarily intended for setup and teardown code. From the guide:

> The specification recommends that, as far as possible, developers should implement custom element setup in this callback

So I think it's worth considering sharpening the language here, something like:

```suggestion
Each time a [custom element](/en-US/docs/Web/API/Web_components/Using_custom_elements) is moved via methods such as {{domxref("Node.insertBefore()")}}), the `disconnectedCallback()` and `connectedCallback()` [lifecycle callbacks](/en-US/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks) are fired, because the element is being removed from and then added back to the DOM.

These callbacks are typically intended to allow you to implement any initialization or cleanup code.

The same callbacks are, by default, fired when the element is moved using `moveBefore()`. However, since the reason to use `moveBefore()` is that the element's state is preserved, it may not be appropriate to run initialization and setup code. To handle this situation, you can define the `connectedMoveCallback()` for the element: if you do, then neither `disconnectedCallback()` or `connectedCallback()` are called when the element is moved using `moveBefore()`.
```

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Examples
+
+### Basic `moveBefore()` usage
+
+In this demo we illustrate basic usage of `moveBefore()`.
+
+#### HTML
+
+The HTML features three {{htmlelement("button")}} elements, and an {{htmlelement("article")}} element. We will use the buttons to control inserting `DocumentFragments` into the `<article>`, and emptying it.
```
wbamberg, line 110:
```suggestion
The HTML features three {{htmlelement("button")}} elements, and an {{htmlelement("article")}} element. We will use the buttons to control inserting `DocumentFragments` into the `<article>` and emptying it.
```

Also though, I find it grating to mix identifiers into prose like "`DocumentFragment`s", rather than something like "`DocumentFragment` instances". This is a very pedantic comment but I have checked and at least Hamish feels the same way.

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### JavaScript
+
+In our script, we define a function, `createFragment()`, which creates a `DocumentFragment` containing a {{htmlelement("div")}} element and two {{htmlelement("section")}} elements as immediate children.
+
+We then attach a click event listener to each `<button>` via {{domxref("EventTarget.addEventListener", "addEventListener()")}}:
+
+- The first button appends the `DocumentFragment` to the `#wrapper` `<article>` element, unmodified.
+- The second button appends the `DocumentFragment` to the `#wrapper` `<article>` element, but first uses `moveBefore()` to move the `<div>` to be the second child of the `DocumentFragment` rather than the first.
+- The third button empties the `#wrapper` `<article>` element using {{domxref("Element.innerHTML", "innerHTML()")}}.
```
wbamberg, line 152:
```suggestion
- The third button empties the `#wrapper` `<article>` element using {{domxref("Element.innerHTML", "innerHTML")}}.
```

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,215 @@
...
+
+```js live-sample___movebefore-basic
+const wrapper = document.getElementById("wrapper");
+const insertBtn1 = document.getElementById("insert1");
+const insertBtn2 = document.getElementById("insert2");
+const clearBtn = document.getElementById("clear");
+
+function createFragment() {
+  let fragment = new DocumentFragment();
```
wbamberg, line 161:
```suggestion
  const fragment = new DocumentFragment();
```

wbamberg (commented):
Thanks Chris! I think some of my comments are going to apply in a few places.

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,215 @@
...
+### `moveBefore()` constraints
+
+There are some constraints to be aware of when using `moveBefore()`:
+
+- It can only work when moving a node within the same document.
+- It won't work if you try to move a node that is not connected to the DOM to an already connected parent, or vice versa.
+
+In such cases, `moveBefore()` will fail with a `HierarchyRequestError` exception. If the above constraints are requirements for your particular use case, you should use {{domxref("Node.insertBefore()")}} instead, or use [`try...catch`](/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) to handle the errors that arise from such cases.
+
+### Moving custom elements
```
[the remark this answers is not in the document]
wbamberg: FWIW I don't really understand why `connectedCallback()` and `disconnectedCallback()` are fired when using `moveBefore()`.

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,215 @@
...
+- `movedNode`
+  - : A {{domxref("Node")}} representing the node to be moved.
+- `referenceNode`
+  - : A {{domxref("Node")}} that `movedNode` will be moved before, or `null`. If the value is `null`, `movedNode` is inserted at the end of the invoking `DocumentFragment`'s child nodes.
+    > [!NOTE]
+    > The `referenceNode` parameter is mandatory. You must explicitly pass a {{domxref("Node")}} or `null`.
+
+### Return value
+
+None ({{jsxref("Undefined")}}).
```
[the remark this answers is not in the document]
chrisdavidmills: Updated on all three pages.

files/en-us/web/api/web_components/using_custom_elements/index.md
```diff
@@ -55,10 +55,11 @@ Once your custom element is registered, the browser will call certain methods of
...
-- `adoptedCallback()`: called each time the element is moved to a new document.
-- `attributeChangedCallback()`: called when attributes are changed, added, removed, or replaced. See [Responding to attribute changes](#responding_to_attribute_changes) for more details about this callback.
+- `connectedCallback()`: Called each time the element is added to the document. The specification recommends that, as far as possible, developers should implement custom element setup in this callback rather than the constructor.
+- `disconnectedCallback()`: Called each time the element is removed from the document.
+- `connectedMoveCallback()`: When defined, called _instead_ of `connectedCallback()` and `disconnectedCallback()` each time the element is moved to a different place in the DOM via {{domxref("Element.moveBefore()")}}. This can be used to overcome functionality problems encountered as a result of initialization/cleanup code being run in the `connectedCallback()` and `disconnectedCallback()` callbacks when the element is not actually being added to or removed from the DOM. See [Moving custom elements](/en-US/docs/Web/API/Element/moveBefore#moving_custom_elements) for more details.
```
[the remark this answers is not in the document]
chrisdavidmills: Like it; updated.

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+### Parameters
+
+- `movedNode`
+  - : A {{domxref("Node")}} representing the node to be moved.
+- `referenceNode`
+  - : A {{domxref("Node")}} that `movedNode` will be moved before, or `null`. If the value is `null`, `movedNode` is inserted at the end of the `Document`'s child nodes.
+    > [!NOTE]
+    > The `referenceNode` parameter is mandatory. You must explicitly pass a {{domxref("Node")}} or `null`.
```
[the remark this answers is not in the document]
chrisdavidmills: This is a good point; I wouldn't normally add this. I only did it because `insertBefore()` has a similar note on it.

I have removed all three instances from this PR (but not the `insertBefore()` one, as that has additional bug information that I'm not sure about removing.)

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Return value
+
+None ({{jsxref("Undefined")}}).
```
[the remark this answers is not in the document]
chrisdavidmills: done

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Exceptions
+
+- `HierarchyRequestError` {{jsxref("TypeError")}}
+  - : Thrown when:
```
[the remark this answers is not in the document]
chrisdavidmills: updated, on all three pages

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,84 @@
...
+## Description
+
+The `moveBefore()` method moves a given node to a new place in the DOM. It provides similar functionality to the {{domxref("Node.insertBefore()")}} method, except that it doesn't remove and then reinsert the node. This means that the state of the node (which would be reset if moving it with `insertBefore()` and similar mechanisms) is preserved after the move. This includes:
+
+- [Animation](/en-US/docs/Web/CSS/CSS_animations) and [transition](/en-US/docs/Web/CSS/CSS_transitions) state.
+- {{htmlelement("iframe")}} loading state.
+- Interactivity states (for example, {{cssxref(":focus")}} and {{cssxref(":active")}}).
+- [Fullscreen](/en-US/docs/Web/API/Fullscreen_API) element state.
+- Open/close state of [popovers](/en-US/docs/Web/API/Popover_API).
+- Modal state of {{htmlelement("dialog")}} elements (modal dialog will not be closed).
```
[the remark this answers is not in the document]
chrisdavidmills: Good spot, thanks! I've fixed it in all three places.

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+The play state of {{htmlelement("video")}} and {{htmlelement("audio")}} elements is not included in the above list, as these elements retain their state when removed and reinserted, regardless of the mechanism used.
+
+> [!NOTE]
+> When observing changes to the DOM using a {{domxref("MutationObserver")}}, nodes moved with `moveBefore()` will be recorded with a [removed node](/en-US/docs/Web/API/MutationRecord/removedNodes) and an [added node](/en-US/docs/Web/API/MutationRecord/addedNodes).
```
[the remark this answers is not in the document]
chrisdavidmills: OK, WFM. I've moved it out of a note, in all three pages.

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,84 @@
...
+There are some constraints to be aware of when using `moveBefore()`:
+
+- It can only work when moving a node within the same document.
+- It won't work if you try to move a node that is not connected to the DOM to an already connected parent, or vice versa.
+
+In such cases, `moveBefore()` will fail with a `HierarchyRequestError` exception. If the above constraints are requirements for your particular use case, you should use {{domxref("Node.insertBefore()")}} instead, or use [`try...catch`](/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) to handle the errors that arise from such cases.
+
+## Examples
+
+The `moveBefore()` method is not particularly useful when invoked on the `Document` node. There are some non-element uses for it, for example you could use `moveBefore()` to move comment nodes around the root of the `Document`. However, you are much more likely to find a use for invoking it on an individual {{domxref("Element")}} — see {{DOMxRef("Element.moveBefore()")}} for some examples.
```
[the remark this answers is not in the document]
chrisdavidmills: I've moved the paragraph to the description section.

And, I've added a demo that moves a comment node to the end of the document. It's pretty pointless, and not something you'd ever really do, but it's about the best I can come up with (I previously spent hours trying to even get `document.moveBefore()` to run, but now I have ... hurrah!)

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Exceptions
+
+- `HierarchyRequestError` {{jsxref("TypeError")}}
+  - : Thrown when:
+    - The specified `referenceNode` is not yet added to the DOM, and you are trying to move it inside a node that is part of the DOM, or vice versa.
+    - An attempt is made to move `movedNode` between two different documents.
```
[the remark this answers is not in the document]
chrisdavidmills: Ah, good point. These need updating. I did some testing, and edited them to:

- The specified `referenceNode` is already added to the DOM, and you are trying to move it inside a `DocumentFragment`.
- You are trying to move `movedNode` between two different document fragments.

I found that both of these cases produced the exception in question.

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Return value
+
+None ({{jsxref("Undefined")}}).
+
+### Exceptions
+
+- `HierarchyRequestError` {{jsxref("TypeError")}}
+  - : Thrown when:
+    - The specified `referenceNode` is not yet added to the DOM, and you are trying to move it inside a node that is part of the DOM, or vice versa.
```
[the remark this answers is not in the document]
chrisdavidmills: I think the confusion here was partly around `referenceDOM`, which should have been `movedNode`, I think. I've updated the sentence in both places it appears, to:

> The specified `movedNode` is not part of the DOM, and you are trying to move it inside a node that is part of the DOM, or vice versa.

Does this clear up the confusion, or no? In which case, please state your case again ;-)

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,215 @@
...
+- `NotFoundError` {{jsxref("TypeError")}}
+  - : The specified `referenceNode` is not a child of the `DocumentFragment` you are calling `moveBefore()` on, that is, the fragment you are trying to move `movedNode` inside.
+- `TypeError` {{jsxref("TypeError")}}
+  - : The second argument was not supplied.
+
+## Description
+
+The `moveBefore()` method moves a given node to a new place in the DOM. It provides similar functionality to the {{domxref("Node.insertBefore()")}} method, except that it doesn't remove and then reinsert the node. This means that the state of the node (which would be reset if moving it with `insertBefore()` and similar mechanisms) is preserved after the move. This includes:
```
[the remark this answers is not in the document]
chrisdavidmills: Oops, updated to `DocumentFragment`

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,215 @@
...
+The play state of {{htmlelement("video")}} and {{htmlelement("audio")}} elements is not included in the above list, as these elements retain their state when removed and reinserted, regardless of the mechanism used.
+
+> [!NOTE]
+> When observing changes to the DOM using a {{domxref("MutationObserver")}}, nodes moved with `moveBefore()` will be recorded with a [removed node](/en-US/docs/Web/API/MutationRecord/removedNodes) and an [added node](/en-US/docs/Web/API/MutationRecord/addedNodes).
+
+### `moveBefore()` constraints
+
+There are some constraints to be aware of when using `moveBefore()`:
+
+- It can only work when moving a node within the same document.
```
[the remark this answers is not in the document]
chrisdavidmills: updated to document fragment

files/en-us/web/api/documentfragment/movebefore/index.md
chrisdavidmills: I've used some of your wording.

files/en-us/web/api/documentfragment/movebefore/index.md
chrisdavidmills: Well, in this case, it introduced an actual mistake. Updated to "...inserting `DocumentFragment` instances...".

And I also removed the comma as per the suggestion.

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,127 @@
...
+  <body>
+    <!-- This comment should be at the end of the document -->
+    <p>Some content</p>
+  </body>
+</html>
+```
+
+#### JavaScript
+
+In our script, we loop through all the {{domxref("Node.childNodes", "childNodes")}} of the `<body>` element. When we find a node with a {{domxref("Node.nodeType", "nodeType")}} value of `8` (which indicates a comment node), we store a reference to it insidea variable called `commentNode`. We then invoke `document.moveBefore()`, specifying that we want to move the comment node, and specifying a second argument of `null` to inserted our comment the end of the `Document`'s child nodes.
```
wbamberg, line 93:
```suggestion
In our script, we loop through all the {{domxref("Node.childNodes", "childNodes")}} of the `<body>` element. When we find a node with a {{domxref("Node.nodeType", "nodeType")}} value of `8` (which indicates a comment node), we store a reference to it inside a variable called `commentNode`. We then invoke `document.moveBefore()`, specifying that we want to move the comment node, and specifying a second argument of `null` to insert our comment the end of the `Document`'s child nodes.
```

files/en-us/web/api/document/movebefore/index.md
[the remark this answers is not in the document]
wbamberg: I no longer understand my own previous comment (!) but:

> The specified movedNode is not part of the DOM, and you are trying to move it inside a node that is part of the DOM, or vice versa.

This reads weirdly for `**Document**.moveBefore()`, because with that API, per the overview description, we are moving "a given [Node](https://pr39002.review.mdn.allizom.net/en-US/docs/Web/API/Node) inside the Document DOM node as a direct child".

Could we replace both bullets with "The specified `movedNode` is not part of this document."

Also there seem to be some other conditions in which this exception gets thrown: https://dom.spec.whatwg.org/#move.

files/en-us/web/api/documentfragment/movebefore/index.md
wbamberg: I still think this https://pr39002.review.mdn.allizom.net/en-US/docs/Web/API/Web_components/Using_custom_elements#moving_custom_elements_lifecycle_side-effects is confusing.

AIUI, this issue, and the solution of implementing `connectedMoveCallback()`, is specific to `moveBefore()`. It's not considered an issue for `insertNode()`, because in that case the element actually is removed and reinserted, and its state is reset. And, in that situation,  implementing `connectedMoveCallback()` does not help you. Is that right?

But this section doesn't say that. It says you can move elements using either API, and that the `disconnectedCallback`/`connectedCallback` will fire in these situations. It then says:

> since these callbacks are typically used to implement any required initialization or cleanup code, moving the element may cause problems with its state.

It doesn't indicate that this is specifically an issue for `moveBefore()`. And "may cause problems with its state" is vague. What we mean AIUI is: if you move an element using `moveBefore()`, its state is preserved, so you don't want to run cleanup/initialization code. This is IMO a much more precise way to express it.

It then says:

> Custom elements can be opted in to state-preserving moves via the [Element.moveBefore()](https://pr39002.review.mdn.allizom.net/en-US/docs/Web/API/Element/moveBefore) method and the connectedMoveCallback() lifecycle callback.

...but it's not at all obvious what "opted in to state-preserving moves" means. For example, calling `moveBefore()` doesn't "opt into" a state preserving move, it performs a state preserving move. It would be much clearer to say something like "In contrast to `insertNode()`, if you move an element using `moveBefore()` then its state is preserved".

FWIW I also don't see the value in the example, which is just an entirely generic presentation of the callbacks, as a cut down copy/paste of the example in https://pr39002.review.mdn.allizom.net/en-US/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks. And I think it's odd in that context to say "This might be your intended behavior." - surely logging `"Custom element added to page."` is almost never going to be the intended behavior.

Anyway, these are the kinds of considerations I had in mind when making my comment. I don't want this to be like "you must make exactly the change I've suggested" - I'm happy to discuss. But I'd like to understand why you're not keen on accepting the changes I've suggested.

files/en-us/web/api/web_components/using_custom_elements/index.md
```diff
@@ -236,6 +241,44 @@ The `:state()` pseudo-class can also be used within the [`:host()`](/en-US/docs/
 
 There are several live examples in {{domxref("CustomStateSet")}} showing how this works.
 
+## Moving custom elements: lifecycle side-effects
+
+The position of a custom element in the DOM can be manipulated just like any regular HTML element, but there are [lifecycle](/en-US/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks) side-effects to consider.
+
+Each time a custom element is moved via methods such as {{domxref("Element.moveBefore()")}} or {{domxref("Node.insertBefore()")}}, the `disconnectedCallback()` and `connectedCallback()` lifecycle callbacks are fired, because the element is being removed from and then added back to the DOM.
```
wbamberg, line 248: This:

> Each time a custom element is moved via methods such as {{domxref("Element.moveBefore()")}} ... the `disconnectedCallback()` and `connectedCallback()` lifecycle callbacks are fired, because the element is being removed from and then added back to the DOM.

seems to contradict [this](https://pr39002.review.mdn.allizom.net/en-US/docs/Web/API/DocumentFragment/moveBefore#description):

> The moveBefore() method moves a given node to a new place in the DocumentFragment. It provides similar functionality to the [Node.insertBefore()](https://pr39002.review.mdn.allizom.net/en-US/docs/Web/API/Node/insertBefore) method, except that it doesn't remove and then reinsert the node.

files/en-us/web/api/web_components/using_custom_elements/index.md
```diff
@@ -236,6 +241,44 @@ The `:state()` pseudo-class can also be used within the [`:host()`](/en-US/docs/
...
+    console.log("Custom element removed from page.");
+  }
+}
+
+customElements.define("my-custom-element", MyCustomElement);
+```
+
+In this case, "Custom element removed from page." and "Custom element added to page." are logged to the console each time an instance of the custom element is moved. This might be your intended behavior. However, since these callbacks are typically used to implement any required initialization or cleanup code, moving the element may cause problems with its state.
+
+Custom elements can be opted in to state-preserving moves via the {{domxref("Element.moveBefore()")}} method and the `connectedMoveCallback()` lifecycle callback. If defined in the element constructor, `connectedMoveCallback()` will run instead of `connectedCallback()` and `disconnectedCallback()` when an element instance is moved via `moveBefore()`.
```
wbamberg, line 272: Callbacks are not defined in the constructor.

chrisdavidmills: Ah, good spot. Updated to "element class"

files/en-us/web/api/web_components/using_custom_elements/index.md
```diff
@@ -236,6 +241,44 @@ The `:state()` pseudo-class can also be used within the [`:host()`](/en-US/docs/
 
 There are several live examples in {{domxref("CustomStateSet")}} showing how this works.
 
+## Moving custom elements: lifecycle side-effects
```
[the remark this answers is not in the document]
chrisdavidmills: OK, sounds good to me. Updated.

files/en-us/web/api/document/movebefore/index.md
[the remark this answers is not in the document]
chrisdavidmills: OK, I've made the suggested change. I've also looked through the spec and updated the situations under which this exception gets thrown.

files/en-us/web/api/documentfragment/movebefore/index.md
chrisdavidmills: I didn't use all of your wording because the last paragraph seemed a bit wordy and bottom heavy.

I've taken all of your follow-up comments here on board, and rewritten the section quite heavily. I do think the state stuff is an issue for `insertBefore()` (and other similar DOM move methods), because they are how people have moved elements for a long time before `moveBefore()` was created.

I want to be careful to frame it as an issue with `moveBefore()` and other similar move methods, which can be fixed using the `connectedMoveCallback()` lifecycle callback, specifically when you use `moveBefore()` method to move elements. 

Rather than just an issue with `moveBefore()`.

> AIUI, this issue, and the solution of implementing connectedMoveCallback(), is specific to moveBefore(). It's not considered an issue for insertNode(), because in that case the element actually is removed and reinserted, and its state is reset. And, with insertNode(), connectedMoveCallback() does not suppress the other callbacks. Is that right?

Your summary is correct, but it _is_ still an issue for people using `insertBefore()` when they are trying to use it to move nodes rather than insert them, which is often the case.

wbamberg: I think this version is great. 👍

files/en-us/web/api/web_components/using_custom_elements/index.md
```diff
@@ -92,6 +97,24 @@ class MyCustomElement extends HTMLElement {
 customElements.define("my-custom-element", MyCustomElement);
 ```
 
+#### Lifecycle callbacks and state-preserving moves
+
+The position of a custom element in the DOM can be manipulated just like any regular HTML element, but there are [lifecycle](/en-US/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks) side-effects to consider.
```
wbamberg, line 102:
```suggestion
The position of a custom element in the DOM can be manipulated just like any regular HTML element, but there are lifecycle side-effects to consider.
```

remove self-link

files/en-us/web/api/documentfragment/movebefore/index.md
```diff
@@ -0,0 +1,179 @@
...
+## Syntax
+
+```js-nolint
+moveBefore(movedNode, referenceNode)
+```
+
+### Parameters
+
+- `movedNode`
+  - : A {{domxref("Node")}} representing the node to be moved.
```
wbamberg, line 22:
```suggestion
  - : A {{domxref("Node")}} representing the node to be moved. Note that this must be an {{domxref("Element")}} or a {{domxref("CharacterData")}} node.
```

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,128 @@
...
+## Syntax
+
+```js-nolint
+moveBefore(movedNode, referenceNode)
+```
+
+### Parameters
+
+- `movedNode`
+  - : A {{domxref("Node")}} representing the node to be moved.
```
wbamberg, line 22:
```suggestion
  - : A {{domxref("Node")}} representing the node to be moved. Note that this must be an {{domxref("Element")}} or a {{domxref("CharacterData")}} node.
```

Because I would be annoyed if I tried to use something else, which was correct according to the parameter description, and it didn't work.

files/en-us/web/api/element/movebefore/index.md
```diff
@@ -0,0 +1,282 @@
...
+## Syntax
+
+```js-nolint
+moveBefore(movedNode, referenceNode)
+```
+
+### Parameters
+
+- `movedNode`
+  - : A {{domxref("Node")}} representing the node to be moved.
```
wbamberg, line 22:
```suggestion
  - : A {{domxref("Node")}} representing the node to be moved. Note that this must be an {{domxref("Element")}} or a {{domxref("CharacterData")}} node.
```

files/en-us/web/api/document/movebefore/index.md
```diff
@@ -0,0 +1,128 @@
...
+
+None ({{jsxref("undefined")}}).
+
+### Exceptions
+
+- `HierarchyRequestError` {{jsxref("TypeError")}}
+  - : Thrown in any of the following situations:
+    - The specified `movedNode` is not part of this document.
+    - The specified `movedNode` is not an {{domxref("Element")}} or {{domxref("CharacterData")}} node.
+    - You are trying to move `movedNode` before the document's [Doctype](/en-US/docs/Glossary/Doctype) (represented by a {{domxref("DocumentType")}} object).
```
wbamberg, line 36:
```suggestion
    - You are trying to move `movedNode` before the document's {{glossary("doctype")}} (represented by a {{domxref("DocumentType")}} object).
```

(note that this might give you a flaw because the case of "doctype" does not match the slug case, but I think we ought to ignore that).

wbamberg (approved):
I had just a few comments, so am approving this so as not to hold you up. I think the only comment that's more or less mandatory is to remove the self-link, although I'd also be very happy for the casing of "doctype" to be fixed.

Thanks a lot for your work on this!

chrisdavidmills:
Yay, thanks for the fab review @wbamberg!

PR #39010, opened 2025-04-07
FF138 Relnote: Certificate returns fingerprints

hamishwillee:
FF138 adds support for the [`RTCCertificate.getFingerprints()`](https://developer.mozilla.org/en-US/docs/Web/API/RTCCertificate/getFingerprints) method in https://bugzilla.mozilla.org/show_bug.cgi?id=1525241

This adds the release not and update to docs to make it clear what this is used for.
I am "fairly sure" this is correct based on discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1525241

Note I found this confusing because there was lots of discussion around certificate/fingerprint exchange and DTLS. The key here is that WebRTC will automatically exchanges certs/fingerprints as part of the offer and DTLS handshake, and an app doesn't have to do anything special for that.

This is useful if you want the app do do something special - such as associated a particular user with a fingerprint. 

Related docs can be tracked in #38882


files/en-us/web/api/rtccertificate/getfingerprints/index.md
```diff
@@ -41,39 +41,49 @@ Each fingerprint is represented by an object with the following properties:
...
 
-First we create a connection and get the fingerprints.
+First we create a connection and get certificates and their fingerprints.
 We also get the fingerprints from the server using "some mechanism".
 
 ```js
-const rtcPeerConnection = new RTCPeerConnection();
 
 // Get the certificate fingerprints from the client.
-const fingerprintsFromClient = rtcPeerConnection.certificate.getFingerprints();
```
hamishwillee, line 51: This was definitely wrong, because there is no `RTCPeerConnection.certificate` property. 

You can either the `RTCCertificate` from the `RTCPeerConnection.getConfiguration()` (which gets your current configuration, as set up in the peer connection constructor) or from a certificate you generate using the static method and pass in.

I think the intent is that the peer connection gets created and creates you some certificates (or you create your own). You then exchange their fingerprints out of band, then start the connection up by creating offer etc. As part of all this DTLS gets setup and the fingerprints sent are used for identity verification.
If that is true, then this is OK - though it could be more informative.

files/en-us/web/api/rtccertificate/getfingerprints/index.md
```diff
@@ -41,22 +41,32 @@ Each fingerprint is represented by an object with the following properties:
...
 
-// Get the certificate fingerprints from the server (pseudo code)
+// Get the certificate fingerprints from the server for particular certificate (pseudo code)
 const fingerprintsFromServer = [
   /* … */
 ];
 ```
 
-There are numerous ways to compare the fingerprint arrays.
+There are numerous ways to compare the fingerprint arrays for a particular certificate
```
hamishwillee, line 69:
```suggestion
There are numerous ways to compare the fingerprint arrays for a particular certificate.
```

pepelsbey (approved):
Looks good! Thank you :)

PR #39011, opened 2025-04-07
FF138 Relnote: RTCSender degradationPreference

hamishwillee:
FF138 adds support for the `degradationPreference` setting on WebRTC senders in https://bugzilla.mozilla.org/show_bug.cgi?id=1329847

This adds a release note.

Related docs work can be tracked in #38880

### DIFF
--- a/files/en-us/mozilla/firefox/releases/138/index.md
+++ b/files/en-us/mozilla/firefox/releases/138/index.md
@@ -42,7 +42,7 @@ Firefox 138 is the current [Beta version of Firefox](https://www.mozilla.org/en-
 
 #### Media, WebRTC, and Web Audio
 
-- WebRTC applications can now set a preference for maintaining framerate or resolution when both the configured frame rate and resolution cannot be maintained.
+- WebRTC applications can now set a preference for prioritizing framerate or resolution when both cannot be maintained at the configured levels due to network degradation.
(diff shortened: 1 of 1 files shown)


files/en-us/mozilla/firefox/releases/138/index.md
```diff
@@ -42,6 +42,11 @@ Firefox 138 is the current [Beta version of Firefox](https://www.mozilla.org/en-
 
 #### Media, WebRTC, and Web Audio
 
+- WebRTC applications can now set a preference for maintaining framerate or resolution when both the configured frame rate and resolution cannot be maintained.
```
bsmth, line 45: Small one for you

```suggestion
- WebRTC applications can now set a preference for maintaining either framerate or resolution when both the configured frame rate and resolution cannot be maintained due to bandwidth constriction.
```

bsmth (approved):
Looking good, small suggestion (based on the bug) if you like it:

files/en-us/mozilla/firefox/releases/138/index.md
hamishwillee: Thanks. It's not actually either, but I don't want to draw that distinction here - there is a balanced option where both fail. 

```suggestion
- WebRTC applications can now set a preference for prioritizing framerate or resolution when network degradation means that they can't both cannot be maintained at the configured levels.
```

bsmth: OK good, thanks for clarifying!

PR #39017, opened 2025-04-07
Fix typo

arlgnv:
### Description

Fixes small typo

### DIFF
--- a/files/en-us/learn_web_development/getting_started/your_first_website/adding_interactivity/index.md
+++ b/files/en-us/learn_web_development/getting_started/your_first_website/adding_interactivity/index.md
@@ -452,7 +452,7 @@ Next, let's change the page heading to a personalized welcome message when the u
    }
    ```
 
-   This first line of this block uses the negation operator (logical NOT, represented by the `!`) to check whether the `name` data item is already stored in localStorage. If not, the `setUserName()` function runs to create it. If it exists (that is, the user set a user name during a previous visit), we retrieve the stored name using `getItem()` and set the `textContent` of the heading to a string, plus the user's name, as we did inside `setUserName()`.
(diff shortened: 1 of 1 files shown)


files/en-us/learn_web_development/getting_started/your_first_website/adding_interactivity/index.md
```diff
@@ -452,7 +452,7 @@ Next, let's change the page heading to a personalized welcome message when the u
    }
    ```
 
-   This first line of this block uses the negation operator (logical NOT, represented by the `!`) to check whether the `name` data item is already stored in in localStorage. If not, the `setUserName()` function runs to create it. If it exists (that is, the user set a user name during a previous visit), we retrieve the stored name using `getItem()` and set the `textContent` of the heading to a string, plus the user's name, as we did inside `setUserName()`.
+   This first line of this block uses the negation operator (logical NOT, represented by the `!`) to check whether the `name` data item is already stored in localStorage. If not, the `setUserName()` function runs to create it. If it exists (that is, the user set a user name during a previous visit), we retrieve the stored name using `getItem()` and set the `textContent` of the heading to a string, plus the user's name, as we did inside `setUserName()`.
```
Josh-Cena, line 455:
```suggestion
   This first line of this block uses the negation operator (logical NOT, represented by the `!`) to check whether the `name` data item is already stored in `localStorage`. If not, the `setUserName()` function runs to create it. If it exists (that is, the user set a user name during a previous visit), we retrieve the stored name using `getItem()` and set the `textContent` of the heading to a string, plus the user's name, as we did inside `setUserName()`.
```

Josh-Cena (approved):
Good catch, thanks

PR #39020, opened 2025-04-07
chore: Remove references to deprecated params in compat and specification macros

bsmth:
## Description

We've only two places we see `{{compat()}}` and `{{specifications()}}`.

Also removing a place where we talk about macro for sidebars seeing as we do this in yaml now.

See also:

- [x] https://github.com/mdn/content/pull/20480
- [x] https://github.com/mdn/yari/pull/6576


pepelsbey (approved):
Looks good! Thank you :)

PR #39022, opened 2025-04-07
Fix CSP upgrade-insecure-requests typo

a-kriya:


wbamberg (approved):
Good catch, thank you!

PR #39026, opened 2025-04-07
Remove unreferenced images

Josh-Cena:
Discovered by my new automation.

To review this, you probably need to search through the entire content to confirm that these files are not referenced anywhere.


Josh-Cena:
I'm not sure about that because my own habit is to remove images whenever they are unused in our content—I'm fairly sure some others do that too. I think it's a bit of a hazzle to have to care about translated content when maintaining the main content, and any out-of-sync should be a bug in itself, even without a broken image, so I'd argue that those pages are already broken as-is. This PR just performs cleanup that ought to have happened with the content change.

bsmth:
@mdn/localization-team-leads FYI there is some cleanup incoming here - could you please take a look at the removals and try to plan for these deletions so you don't have broken images in downstream pages?

bsmth:
> I'm not sure about that because my own habit is to remove images whenever they are unused in our content—I'm fairly sure some others do that too. I think it's a bit of a hassle to have to care about translated content when maintaining the main content, and any out-of-sync should be a bug in itself, even without a broken image, so I'd argue that those pages are already broken as-is. This PR just performs cleanup that ought to have happened with the content change.

I understand, yes. Basically, I'm +1 on the changes, but I'd like to give the translated content teams some time to react before we remove them, how does that sound?

Josh-Cena:
Yeah sure.

bsmth (approved):
Leaving a +1 on this, I will remind translated-content folks.

Josh-Cena:
Going to merge this now

PR #39027, opened 2025-04-08
Remove KeyboardEvent.initKeyEvent page

Josh-Cena:
This API has not been exposed by default for many years now. It has no associated BCD.


bsmth (approved):
Thanks! (This was discussed on Discord already)

PR #39029, opened 2025-04-08
Remove XPathException

Josh-Cena:
This API has never been implemented in any browser. See also: https://github.com/whatwg/dom/issues/67


bsmth (approved):
Great, tnx

PR #39041, opened 2025-04-08
Fix example code for picker-icon and checkmark pseudo-elements

lucasweng:
### Description

This PR updates the example code for [picker-icon](https://developer.mozilla.org/en-US/docs/Web/CSS/::picker-icon#animating_the_picker_icon) and [checkmark](https://developer.mozilla.org/en-US/docs/Web/CSS/::checkmark#customizing_the_checkmark) to match their respective descriptions. The correct selector should be ::picker(select) for the styling to work properly.

> ### Animating the picker icon
>
> To opt-in to customizable select functionality, the `<select>` element and its picker both need to have an [appearance](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance) value of `base-select` set on them:
>
> ```css
> select,
> ::picker-icon(select) {
>  appearance: base-select;
> }
> ```

> ### Customizing the checkmark
>
> To opt-in to customizable select functionality, the `<select>` element and its picker both need to have an [appearance](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance) value of `base-select` set on them:
>
> ```css
> select,
> ::checkmark(select) {
>   appearance: base-select;
> }
> ```

### DIFF
--- a/files/en-us/web/css/_doublecolon_checkmark/index.md
+++ b/files/en-us/web/css/_doublecolon_checkmark/index.md
@@ -99,7 +99,7 @@ To opt-in to customizable select functionality, the `<select>` element and its p
 
 ```css
 select,
-::checkmark(select) {
+::picker(select) {
   appearance: base-select;
 }
 ```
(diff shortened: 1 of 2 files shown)


Josh-Cena (approved):
Nice catch, thanks

PR #39051, opened 2025-04-09
Use WAT as code block language instead of Wasm

Josh-Cena:
Follow up from https://github.com/mdn/yari/pull/12903


files/en-us/webassembly/guides/understanding_the_text_format/index.md
```diff
@@ -26,15 +26,15 @@ represents a tree with the root node "module" and two child nodes, a "memory" no
...
+```wat
 (module)
 ```
 
 This module is totally empty, but is still a valid module.
 
 If we convert our module to binary now (see [Converting WebAssembly text format to Wasm](/en-US/docs/WebAssembly/Guides/Text_format_to_Wasm)), we'll see just the 8 byte module header described in the [binary format](https://webassembly.github.io/spec/core/binary/modules.html#binary-module):
 
-```wasm
+```wat
```
bsmth, line 37: I think this is the only place we show binary format, maybe `plain` is appropriate:

```suggestion
```plain
```

bsmth (approved):
LGTM, thanks!

Issue #4025, opened 2021-04-12
Issue with "getAvailability": needs improvement (currently just contains a dump of the spec for the method)


hamishwillee:
MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/PresentationRequest/getAvailability

#### What information was incorrect, unhelpful, or incomplete?

This just contains a dump of the spec for the method. 

#### What did you expect to see?

MDN style explainer

<details>
<summary>MDN Content page report details</summary>

* Folder: `en-us/web/api/presentationrequest/getavailability`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/PresentationRequest/getAvailability
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/presentationrequest/getavailability/index.html
* Last commit: https://github.com/mdn/content/commit/dc9089417852530f233c345bd6821d6e06f11b1e
* Document last modified: 2020-12-18T10:05:08.000Z

</details>

PR #39053, opened 2025-04-09
Add `navigator.presentation` example

low-perry:
### Description
This pull request adds an example to demonstrate the usage of the `navigator.presentation` property in the documentation. The example checks if the browser supports the Presentation API, retrieves the `navigator.presentation` object, and logs it to the console. If the API is unavailable, an error message is displayed.


Josh-Cena (changes requested):
Thanks but I don't think this is a useful demo. I would prefer an actual usage example that invokes at least one method on the presentation object.

low-perry:
@Josh-Cena I'll try to improve the example.

Josh-Cena:
IMO this is still not a useful example because it doesn't pertain to `navigator.presentation`: you are just using it for feature testing but then going to something else. I understand all of this is very hard because MDN's documentation on Presentation API is extremely bad (https://github.com/mdn/content/issues/4025). Is there a motivation for adding this example? A lot of these global singletons don't have examples anyway because you can just go to the `Presentation` interface page.

low-perry:
Thanks for your feedback! From what I understand, `navigator.presentation` is just a read-only attribute of the `Navigator` interface that serves as a gateway to the Presentation API, with no methods to use on it. That’s why my first example focused on checking if the browser supports the Presentation API. The idea was simply to show `navigator.presentation` in action—demonstrating that it exists and providing a basic example of its usage.

Josh-Cena:
The `Presentation` object seems to contain two properties. I have no idea what they do though.

low-perry:
The `defaultRequest` represents a `PresentationRequest` object that can be used to start a presentation, mainly to establish a connection to an external display. However, it’s not fundamentally different from directly instantiating a `PresentationRequest`, since you still need to create the object before setting it as the `defaultRequest`. This redundancy might explain why the W3C specification doesn't emphasize `defaultRequest` much—it’s simply another way to associate a `PresentationRequest` with `navigator.presentation`.

Josh-Cena (approved):

PR #39059, opened 2025-04-10
HTTP caching: add quotes around ETag

mina86:
The format of the ETag header requires the entity tags to be surrounded in double quotes.  Add them when missing in the example HTTP responses in the document.


bsmth (approved):
Looks good, thanks!

PR #39070, opened 2025-04-10
Remove experimental flag from Reporting API doc

tunetheweb:
### Description

Safari 16.4 now supports this, making two browser engines supporting it and so no longer meeting the definition of Experimental. Firefox also has it behind a flag btw.

This was recently updated in https://github.com/mdn/browser-compat-data/issues/26361 and https://github.com/mdn/browser-compat-data/pull/26380 but it's not linked to BCD here as far as I can tell so doing a manual update in this PR (and also linking it for future reference).

I also removed the reference to the deprecated `Report-To` header since it [is no longer recommended](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Report-To) and not even part of the Spec.

FYI: @yoavweiss 

### Motivation

Correcting data.

### Additional details

https://github.com/WebKit/WebKit/pull/3613

https://github.com/mdn/browser-compat-data/issues/26361
https://github.com/mdn/browser-compat-data/pull/26380

### DIFF
--- a/files/en-us/web/api/reporting_api/index.md
+++ b/files/en-us/web/api/reporting_api/index.md
@@ -10,9 +10,6 @@ spec-urls:
   - https://w3c.github.io/webappsec-csp/#cspviolationreportbody
   - https://wicg.github.io/deprecation-reporting/#deprecationreportbody
   - https://wicg.github.io/intervention-reporting/#intervention-report
-browser-compat:
-  - http.headers.Reporting-Endpoints
-  - http.headers.Report-To
 ---
 
 {{DefaultAPISidebar("Reporting API")}}{{AvailableInWorkers}}
--- a/files/en-us/web/api/reporting_api/index.md
+++ b/files/en-us/web/api/reporting_api/index.md
@@ -2,6 +2,9 @@
 title: Reporting API
 slug: Web/API/Reporting_API
 page-type: web-api-overview
+browser-compat:
+  - http.headers.Reporting-Endpoints
+  - http.headers.Report-To
 spec-urls:
   - https://w3c.github.io/reporting/#intro
   - https://w3c.github.io/webappsec-csp/#cspviolationreportbody
--- a/files/en-us/web/api/reporting_api/index.md
+++ b/files/en-us/web/api/reporting_api/index.md
@@ -2,9 +2,7 @@
 title: Reporting API
 slug: Web/API/Reporting_API
 page-type: web-api-overview
-browser-compat:
-  - http.headers.Reporting-Endpoints
-  - http.headers.Report-To
+browser-compat: http.headers.Reporting-Endpoints
 spec-urls:
   - https://w3c.github.io/reporting/#intro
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/reporting_api/index.md
```diff
@@ -2,16 +2,17 @@
 title: Reporting API
 slug: Web/API/Reporting_API
 page-type: web-api-overview
-status:
-  - experimental
+browser-compat:
+  - http.headers.Reporting-Endpoints
+  - http.headers.Report-To
```
tunetheweb, line 7: Actually `Report-To` is deprecated so let's just have `Reporting-Endpoints`:

```suggestion
browser-compat: http.headers.Reporting-Endpoints
```

files/en-us/web/api/reporting_api/index.md
```diff
@@ -101,8 +100,6 @@ These HTTP response headers define the endpoints where reports are sent.
 - {{HTTPHeader("Reporting-Endpoints")}}
   - : Sets the name and URL of reporting endpoints.
     These endpoints can be used in the `report-to` directive, which may be used with a number of HTTP headers including {{httpheader("Content-Security-Policy")}} and or {{HTTPHeader("Content-Security-Policy-Report-Only")}}.
-- {{HTTPHeader("Report-To")}} {{deprecated_inline}}
-  - : Sets the name and URL of reporting endpoint groups, which may be used with a number of HTTP headers including `Content-Security-Policy`.
```
hamishwillee, line 103:
```suggestion
While your point on removal makes some sense, MDN keeps things around while their are browsers that support them. The trigger for removal would be removal from BCD.

```suggestion
- {{HTTPHeader("Report-To")}} {{deprecated_inline}}
  - : Sets the name and URL of reporting endpoint groups, which may be used with a number of HTTP headers including `Content-Security-Policy`.

```
```

Comment (hamishwillee) [reply files/en-us/web/api/reporting_api/index.md:103]: Otherwise this is great thank you. Note that keeping the compatibility table in the guide is "not to template", as discussed in the other case we were involved in. We would definitely remove when FF supports this.

Comment (tunetheweb) [reply files/en-us/web/api/reporting_api/index.md:103]: I've mixed feelings about this. Technically it's not part of the Reporting API anymore (which this page is about) but agree it's still technically supported.

Let's add it back but with a stronger note:

```suggestion
- {{HTTPHeader("Report-To")}} {{deprecated_inline}}
  - : No longer part of the Reporting API but still supported by some browsers. This sets the name and URL of reporting endpoint groups, which may be used with a number of HTTP headers especially for [Network Error Logging](/en-US/docs/Web/HTTP/Guides/Network_Error_Logging) that has not yet been updated to support `Reporting-Endpoints`.

```

hamishwillee: Reasonable compromise

hamishwillee (approved):

PR #39055, opened 2025-04-10
chore(ia): Reorganize HTML pages

dipikabh:
## Description

As part of Phase 2, reorganizing HTML docs to structure them into reference / guides / how-to / tutorials.

#### Before
```
files/en-us/web/html/
├── attributes
├── comments
├── constraint_validation
├── content_categories
├── cors_enabled_image
├── date_and_time_formats
├── element
├── global_attributes
├── index.md
├── microdata
├── microformats
├── quirks_mode_and_standards_mode
├── reference
├── responsive_images
└── viewport_meta_tag
```

#### After

```
files/en-us/web/html/
├── guides
├── how_to
├── index.md
├── reference
└── tutorials
```

```
files/en-us/web/html/
├── guides
│   ├── comments
│   ├── constraint_validation
│   ├── content_categories
│   ├── date_and_time_formats
│   ├── index.md
│   ├── microdata
│   ├── microformats
│   ├── quirks_mode_and_standards_mode
│   ├── responsive_images
│   └── viewport_meta_element
├── how_to
│   ├── cors_enabled_image
│   └── index.md
├── index.md
├── reference
│   ├── attributes
│   ├── elements
│   ├── global_attributes
│   └── index.md
└── tutorials
    └── index.md
```

### Related issues and pull requests

Discussion: https://github.com/orgs/mdn/discussions/776
Tracker issue: https://github.com/mdn/mdn/issues/662

PR #39072, opened 2025-04-10
fix(htmlsidebar): Make top-level items links

dipikabh:
### Description

This is is follow-up to https://github.com/mdn/content/pull/39055.

- Added the `link` key so that the landing pages for those items can be reached from the sidebar
- `code` for Attributes subpages somehow got missed in the previous iteration

### Motivation

Make content reachable from side-bar

### DIFF
--- a/files/sidebars/htmlsidebar.yaml
+++ b/files/sidebars/htmlsidebar.yaml
@@ -29,6 +29,7 @@ sidebar:
     path: /Web/HTML/Reference/Attributes
     title: Attributes
     details: closed
+    code: true
   - type: listSubPages
     path: /Web/HTML/Reference/Global_attributes
     title: Global_attributes
--- a/files/sidebars/htmlsidebar.yaml
+++ b/files/sidebars/htmlsidebar.yaml
@@ -17,21 +17,25 @@ sidebar:
     title: Reference
   - type: listSubPages
     path: /Web/HTML/Reference/Elements
+    link: /Web/HTML/Reference/Elements
     title: HTML_Elements
     details: closed
     code: true
   - type: listSubPages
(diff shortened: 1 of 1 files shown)


files/sidebars/htmlsidebar.yaml
```diff
@@ -17,20 +17,25 @@ sidebar:
     title: Reference
   - type: listSubPages
     path: /Web/HTML/Reference/Elements
+    link: /Web/HTML/Reference/Elements
     title: HTML_Elements
     details: closed
     code: true
   - type: listSubPages
     path: /Web/HTML/Reference/Elements/input
+    link: /Web/HTML/Reference/Elements/input
```
bsmth, line 26: We can remove this because we already list it above under `/Web/HTML/Reference/Elements`

```suggestion
```

This is different from `script/type` because the directory structure uses a different strategy:

```bash
$ tree -d ./files/en-us/web/html/reference/elements/script
./files/en-us/web/html/reference/elements/script
└── type
    ├── importmap
    └── speculationrules

$ tree -d ./files/en-us/web/html/reference/elements/input/
./files/en-us/web/html/reference/elements/input/
├── button
├── checkbox
├── color
…
└── week
```

bsmth (approved):
One small fix for linking the listsubpages, but otherwise good to go!

files/sidebars/htmlsidebar.yaml
dipikabh: True, it does appear under the `Elements` drop-down, and I thought why not repeat it.
But let's go with this (link removal).
This requirement might go away anyway if we can add the nested drop-down inside `Elements`.

dipikabh:
Thanks for reviewing @bsmth!

PR #39083, opened 2025-04-11
Fix CSS syntax errors

liuxuelian:
### Description

Remove the extra whitespace and replace the semicolon(`;`) with braces(`{}`).


Josh-Cena (approved):
I took the chance to fix some other CSS syntax errors. Thank you!

PR #39085, opened 2025-04-11
ElementInternals.aria* examples

estelle:
The examples in elementInternals for ARIA does not define a variable used. This makes the examples a little bit better. (Not perfect, but hopefully a definite improvement)

addresses some of https://github.com/mdn/content/issues/36508


wbamberg (approved):
👍

PR #39089, opened 2025-04-12
Synchronize with BCD v6.0.5

OnkarRuikar:
[BCD release v6.0.5](https://github.com/mdn/browser-compat-data/releases/tag/v6.0.5)

/cc cc/ @bsmth @wbamberg

### DIFF
--- a/files/en-us/web/api/invoker_commands_api/index.md
+++ b/files/en-us/web/api/invoker_commands_api/index.md
@@ -20,7 +20,7 @@ Historically creating these kinds of controls has required JavaScript event list
 
 ## HTML attributes
 
-- [`commandfor`](/en-US/docs/Web/HTML/Reference/Elements/button#commandfor) {{experimental_inline}}
+- [`commandfor`](/en-US/docs/Web/HTML/Reference/Elements/button#commandfor)
   - : Turns a {{htmlelement("button")}} element into a button, controlling the given interactive element; takes the ID of the element to control as its value.
 - [`command`](/en-US/docs/Web/HTML/Reference/Elements/button#command)
   - : Specifies the action to be performed on an element being controlled by a control `<button>`, specified via the `commandfor` attribute.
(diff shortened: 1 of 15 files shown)


files/en-us/web/api/invoker_commands_api/index.md
```diff
@@ -24,12 +22,12 @@ Historically creating these kinds of controls has required JavaScript event list
 
 - [`commandfor`](/en-US/docs/Web/HTML/Reference/Elements/button#commandfor) {{experimental_inline}}
```
Josh-Cena, line 23: Why is this not removed?

OnkarRuikar: This is a landing page, so the bot used only the first BCD query `api.CommandEvent`. As we have discussed multiple times before, we haven't decided how to handle multiple queries on landing pages.

Josh-Cena: > This is a landing page, so the bot used only the first BCD query `api.CommandEvent`.

I would expect the _page status_ to be derived from the page's BCD query, but the DL items' inline status badges to be derived from the feature it is linking to, no?

Josh-Cena (approved):

files/en-us/web/api/invoker_commands_api/index.md
OnkarRuikar: What would be the best way to reach the `html.elements.button` BCD query from `Web/HTML/Reference/Elements/button#commandfor` slug? Hardcoding?
Also, the `HTML.elements.button` query is not mentioned in the page's front matter. To keep it simple and avoid conflicts, ATM the bot processes only the BCD queries listed in the front matter.

Josh-Cena: > What would be the best way to reach the `html.elements.button` BCD query from `Web/HTML/Reference/Elements/button#commandfor` slug? Hardcoding?

My hunch is (based on my experience of parsing content): you can assume that `Web/HTML/Reference/Elements/(\w+)#(\w+)` is always a reference to an attribute, and convert it to `html.elements.$1.$2`.

PR #39090, opened 2025-04-12
Avoid double DD

Josh-Cena:
Noticed some bad typesetting

### DIFF
--- a/files/en-us/web/html/reference/elements/button/index.md
+++ b/files/en-us/web/html/reference/elements/button/index.md
@@ -63,32 +63,22 @@ This element's attributes include the [global attributes](/en-US/docs/Web/HTML/R
   - : Specifies the action to be performed on an element being controlled by a control `<button>`, specified via the `commandfor` attribute. The possible values are:
 
     - `"show-modal"`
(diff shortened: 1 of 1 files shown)


wbamberg (approved):
👍 I'm surprised that markup worked at all!

Josh-Cena:
Yeah I filed https://github.com/mdn/rari/issues/190

wbamberg:
Hm, it is documented though: it's not allowed. From https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#definition_lists:

> Definition lists written using this syntax must consist of pairs of <dt>/<dd> elements. Using this syntax, it's not possible to write a list with more than one consecutive <dt> element or more than one consecutive <dd> element: the parser will treat this as an error. We expect almost all definition lists on MDN will work with this limitation, and for those that do not, authors can fall back to raw HTML.

wbamberg:
I mean in these cases it's obviously an error because they're not semantically multiple `<dd>` elements: that is, they're not two separate definitions. But either way, it's not permitted markup per our"specification".

PR #39096, opened 2025-04-12
Add more links to the Wasm sidebar

Josh-Cena:
A lot of pages are not reachable from the sidebar, making the Wasm docs extremely hard to navigate. This adds pretty much everything to the sidebar.

### DIFF
--- a/files/sidebars/webassemblysidebar.yaml
+++ b/files/sidebars/webassemblysidebar.yaml
@@ -133,7 +133,7 @@ l10n:
     Compiling_Rust_to_Wasm: Rust zu WebAssembly kompilieren
     JavaScript_API: WebAssembly-JavaScript-API verwenden
     Text_format: WebAssembly-Textformat verstehen
-    Text_format_to_Wasm: Converting text format to binary
+    Text_format_to_Wasm: Converting WebAssembly text format
     Loading_and_running: WebAssembly-Code laden und ausführen
     Exported_functions: Exportierte WebAssembly-Funktionen
     JavaScript_builtins: JavaScript-Standardobjekte
@@ -146,7 +146,7 @@ l10n:
     Compiling_Rust_to_Wasm: Compiling from Rust to WebAssembly
     JavaScript_API: Using the WebAssembly JavaScript API
     Text_format: Understanding WebAssembly text format
-    Text_format_to_Wasm: Converting text format to binary
+    Text_format_to_Wasm: Converting WebAssembly text format
     Loading_and_running: Loading and running WebAssembly code
     Exported_functions: Exported WebAssembly functions
     JavaScript_builtins: JavaScript builtins
@@ -159,7 +159,7 @@ l10n:
     Compiling_Rust_to_Wasm: Compiler du Rust en WebAssembly
     JavaScript_API: Utiliser l'API JavaScript WebAssembly
     Text_format: Comprendre le format texte WebAssembly
-    Text_format_to_Wasm: Converting text format to binary
+    Text_format_to_Wasm: Converting WebAssembly text format
     Loading_and_running: Chargement et exécution de code WebAssembly
     Exported_functions: Fonctions WebAssembly exportées
     JavaScript_interface: JavaScript interface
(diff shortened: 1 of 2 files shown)


files/sidebars/webassemblysidebar.yaml
```diff
@@ -4,58 +4,132 @@ sidebar:
...
+    title: Exported_functions
+  - link: /WebAssembly/Guides/JavaScript_builtins
+    title: JavaScript_builtins
+  - link: /WebAssembly/Guides/Imported_string_constants
+    title: Imported_string_constants
+  - type: section
+    link: /WebAssembly/Reference
+    title: Reference
   - title: JavaScript_interface
     details: open
```
dipikabh, line 36: I think we should have this collapsed in the initial state
```suggestion
    details: closed
```

files/sidebars/webassemblysidebar.yaml
```diff
@@ -68,6 +142,7 @@ l10n:
   en-US:
     WebAssembly_concepts: WebAssembly concepts
     Compiling_to_Wasm: Compiling from C/C++ to WebAssembly
+    Existing_C_to_Wasm: Compiling an existing C module to WebAssembly
     Compiling_Rust_to_Wasm: Compiling from Rust to WebAssembly
     JavaScript_API: Using the WebAssembly JavaScript API
     Text_format: Understanding WebAssembly text format
```
dipikabh, line 148: we need an entry for:
```suggestion
    Text_format: Understanding WebAssembly text format
    Text_format_to_Wasm: Converting text format to webAssembly
```
The page title on https://developer.mozilla.org/en-US/docs/WebAssembly/Guides/Text_format_to_Wasm can also be changed

files/sidebars/webassemblysidebar.yaml
```diff
@@ -4,58 +4,132 @@ sidebar:
...
+        details: closed
         code: true
+  - title: Instructions
+    details: open
+    children:
+      - type: listSubPages
+        path: /WebAssembly/Reference/Control_flow
+        link: /WebAssembly/Reference/Control_flow
+        title: Control flow
+        details: closed
```
dipikabh, line 112: - I think we can keep the initial state of `Instructions` as `closed`
- What do you think about moving all children under `Reference' to directly under Reference instead of nesting under `Instructions`?

dipikabh (changes requested):
Great work, looks much neater and better organized, not to mention that all pages are now discoverable via sidebar. Thank you!

I've added a few comments, lmk what you think.

files/sidebars/webassemblysidebar.yaml
Josh-Cena: Are you suggesting we should change the page title? The current title doesn't look too bad for me.

files/sidebars/webassemblysidebar.yaml
dipikabh: Thanks for clarifying. Keeping the organization this way makes sense.

files/sidebars/webassemblysidebar.yaml
dipikabh: Yes, I was suggesting we update both the page title as well as the sidebar entry - mainly to replace "Wasm" with "WebAssembly".
This is the only sidebar entry that uses "Wasm" instead of "WebAssembly", and it is pulling this directly from the page title "Converting WebAssembly text format to Wasm"

dipikabh (approved):
Approving so you can merge when it's ready. Thanks again!

Josh-Cena:
How does "Converting WebAssembly text format to binary" look?

dipikabh:
> How does "Converting WebAssembly text format to binary" look?

Good for the page title but long for the sidebar. Can we strip "to binary" for the sidebar title?

Josh-Cena:
How about "Converting text format to binary" for the sidebar?

Josh-Cena:
After sitting on it some more, I think I prefer your suggestion. Thanks for the reviews!

bsmth:
Great changes, thanks a lot!

PR #39097, opened 2025-04-12
Make Element and Attributes list pages reachable in MathML sidebar

Josh-Cena:
These two pages are not linked in the sidebar


dipikabh (approved):
Thanks!

PR #39105, opened 2025-04-13
Fix anchors

PassionPenguin:
anchors all point to `#details_2` now point to `#details` instead:

- these anchors are links under `addListener syntax > Parameters > listener > details`;
- they are pointing to the details section below which covers the specific parameters of the `details` object;
- however, the subheading `details` seems to have a higher priority when assigning a heading hashtag, this cause the heading itself to be `#details`, while the definition list are to be `#details_2`;
- this causes the links are pointing to themselves.

evidences are below:

image
image


Josh-Cena (approved):
Nice, thank you

PR #39108, opened 2025-04-13
Remove reference to yeared ES spec

nikoandpiko:
### Description

Updates the current ECMAScript version. However, with its inclusion, this document will need updated every June annually.

### Motivation

Provides accurate information.

### Additional details

https://262.ecma-international.org/15.0/

### DIFF
--- a/files/en-us/glossary/javascript/index.md
+++ b/files/en-us/glossary/javascript/index.md
@@ -16,7 +16,7 @@ JavaScript is primarily used in the browser, enabling developers to manipulate w
 
 Conceived as a server-side language by Brendan Eich (then employed by the Netscape Corporation), JavaScript soon came to Netscape Navigator 2.0 in September 1995. JavaScript enjoyed immediate success and {{glossary("Microsoft Internet Explorer", "Internet Explorer 3.0")}} introduced JavaScript support under the name JScript in August 1996.
 
-In November 1996, Netscape began working with Ecma International to make JavaScript an industry standard. Since then, the standardized JavaScript is called ECMAScript and specified under ECMA-262, whose latest (fourteenth, ES2023) edition is available as of June 2023.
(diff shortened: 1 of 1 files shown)


files/en-us/glossary/javascript/index.md
```diff
@@ -16,7 +16,7 @@ JavaScript is primarily used in the browser, enabling developers to manipulate w
 
 Conceived as a server-side language by Brendan Eich (then employed by the Netscape Corporation), JavaScript soon came to Netscape Navigator 2.0 in September 1995. JavaScript enjoyed immediate success and {{glossary("Microsoft Internet Explorer", "Internet Explorer 3.0")}} introduced JavaScript support under the name JScript in August 1996.
 
-In November 1996, Netscape began working with Ecma International to make JavaScript an industry standard. Since then, the standardized JavaScript is called ECMAScript and specified under ECMA-262, whose latest (fourteenth, ES2023) edition is available as of June 2023.
+In November 1996, Netscape began working with Ecma International to make JavaScript an industry standard. Since then, the standardized JavaScript is called ECMAScript and specified under ECMA-262, whose latest (fifteenth, ES2024) edition is available as of June 2024.
```
Josh-Cena, line 19:
```suggestion
In November 1996, Netscape began working with Ecma International to make JavaScript an industry standard. Since then, the standardized JavaScript is called ECMAScript and specified under ECMA-262. The standard is constantly updated and implemented (a _living standard_).
```

nikoandpiko: Yeah, that would take care of the date issue. This solves the issue of the PR. If some semblance of what used to be written needs to be referenced, an external link could be useful (easiest reading would be https://en.wikipedia.org/wiki/ECMAScript_version_history). Other than that, lgtm.

Josh-Cena: We generally prefer timeless documentation so they don't have to be updated annually like this, so this would be the preferred way.

Josh-Cena (approved):

PR #39109, opened 2025-04-13
Remove SVG font API

Josh-Cena:
These features were removed by https://github.com/mdn/browser-compat-data/pull/26443


dipikabh (approved):
Thanks for propagating these updates in content 🙌. 

The following pages have been removed but don't have any deprecated banner – but they also don't have a BCD section. Can you verify if they should be deleted:
- `overline-position` and `overline-thickness` 
- `strikethrough-position` and `strikethrough-thickness` 
- `underline-position` and `underline-thickness`

This one should probably be deleted as well:
- `requiredFeatures`

Josh-Cena:
> * `overline-position` and `overline-thickness`
> * `strikethrough-position` and `strikethrough-thickness`
> * `underline-position` and `underline-thickness`

These features are only usable with the `<font-face>` element, which is removed, so it appears that they no longer make any sense.

> This one should probably be deleted as well:
> 
> * `requiredFeatures`

This attribute contains so many features other than fonts, and it would take some work to verify if it still exists.

dipikabh:
> > * `overline-position` and `overline-thickness`
> > * `strikethrough-position` and `strikethrough-thickness`
> > * `underline-position` and `underline-thickness`
> 
> These features are only usable with the `<font-face>` element, which is removed, so it appears that they no longer make any sense.

Makes sense 👍 

> > This one should probably be deleted as well:
> > 
> > * `requiredFeatures`
> 
> This attribute contains so many features other than fonts, and it would take some work to verify if it still exists.

Sure! There is no BCD data though. We can leave it for a follow-up.

Merging this now. Thank you!

PR #39112, opened 2025-04-14
fix: add missing part of br instruction

familyboat:
### Description

According to https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Control_flow/br, add missing part of `br` instruction.


dipikabh (approved):
Thank you!

PR #39119, opened 2025-04-14
fix(AuthenticatorAttestationResponse): quote BLE transport constant

aomader:
Fixes probably a small left-over.


wbamberg (approved):
👍 thank you @aomader ! I think some people like to quote in this situation and others don't, but we should definitely be consistent within a single page, even if we can't yet agree on a site-wide rule :).

PR #39128, opened 2025-04-15
Update groupby

pimskie:
Remove unnecessary declared objects

### Description
Removed unnecessary declared objects to simplify the example. 

### Motivation

The objects were unneeded, and only lead to confusion


Josh-Cena (approved):
Nice, thank you

PR #39129, opened 2025-04-15
Update "Django Tutorial Part 2: Creating a skeleton website" page to prevent a bug in "Django Tutorial Part 8: User authentication and permissions" 

LittleBangTheory:
Move the catalog as the first line in the installed app to override default behavior

### Description

I moved `'catalog.apps.CatalogConfig',` as the first line of `INSTALLED_APPS` in the setup of the Django skeleton website tutorial, and made the following change in the description just above and below the code section.

### Motivation

When implementing the part 8 of the tutorial, I ran into an issue : the `logged_out.html` page I set up wasn't overriding the default Django page, which meant that when using the log out form in the tutorial, I was sent to the admin console logout page instead of mine.

I found that it was because the order of applications in the settings mattered, so I needed to move my app the top of my list. I hope this change will avoid others the same headache as me ! 

English is not my first language, so my apologies if there are any mistakes in the way I rewrote the description :) 

### Additional details

I found the solution [here](https://stackoverflow.com/questions/35153108/why-is-logged-out-html-not-overriding-in-django-registration), which is coherent with [Django Documentation](https://docs.djangoproject.com/en/5.2/ref/applications/) : 

> At each stage, Django processes all applications in the order of [INSTALLED_APPS](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-INSTALLED_APPS).

None that I could find.

### DIFF
--- a/files/en-us/learn_web_development/extensions/server-side/django/skeleton_website/index.md
+++ b/files/en-us/learn_web_development/extensions/server-side/django/skeleton_website/index.md
@@ -139,22 +139,24 @@ In addition we now have:
 
 Now that the application has been created, we have to register it with the project so that it will be included when any tools are run (like adding models to the database for example). Applications are registered by adding them to the `INSTALLED_APPS` list in the project settings.
 
-Open the project settings file, **django-locallibrary-tutorial/locallibrary/settings.py**, and find the definition for the `INSTALLED_APPS` list. Then add a new line at the end of the list, as shown below:
+Open the project settings file, **django-locallibrary-tutorial/locallibrary/settings.py**, and find the definition for the `INSTALLED_APPS` list. Then add a new line at the top of the list, as shown below:
 
 ```bash
 INSTALLED_APPS = [
+    # Add our new application
+    'catalog.apps.CatalogConfig', # This object was created for us in /catalog/apps.py
+
+    # Previous applications
     'django.contrib.admin',
     'django.contrib.auth',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
-    # Add our new application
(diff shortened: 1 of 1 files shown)


bsmth:
@hamishwillee - PR here for you to have a look at when you have time 👀

hamishwillee:
@LittleBangTheory Firstly, thank you for posting this issue and for posting a solution as a PR. That makes it much easier to evaluate.

My problem with this is that
1. This tutorial has been around for 6 or 7 years and no one else has reported this problem
2. We have a worked tutorial https://github.com/mdn/django-locallibrary-tutorial/ that works.

So we know that if you follow the instructions properly the code will work.
The bit you are omitting is this one https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Server-side/Django/Authentication#template_directory.

It is true that Django looks at INSTALLED_APPS, but that isn't the full story. Django follows this order:
- DIRS (Project-level templates): It iterates through the directories specified in `TEMPLATES['DIRS']`. If it finds a template with the requested name, it uses that one and stops searching. So in the above link we set this to be our templates, which get found.
- APP_DIRS (App-level templates): If `APP_DIRS` is True and the template wasn't found in DIRS, Django then checks the templates directory within each application listed in INSTALLED_APPS, in the order they appear. Even though the admin files have some of the same names, I believe the apps are expecting the more specific path. 

The admin part of the site will be looking for the more specific path like admin/registration/logged_out and will not be found in the templates you specify.

I don't really want to make this change because while it might be a little "safer", it isn't the right way to manage finding templates. If we did do this we'd have to update the app https://github.com/mdn/django-locallibrary-tutorial and fully test it first. 

Therefore I am going to close this. But will reopen if you strongly disagree.

Thank you again for contributing, and apologies for not getting to this earlier.

PR #39134, opened 2025-04-15
Replace strings with string in webRequest docs

PassionPenguin:
array of strings => array of string


Rob--W (approved):
Thanks for spotting the issue and fixing up the types!

Tip for future pull requests (generally commit messages): Try to make the commit message descriptive without it being too long or too short. This enables people who read the pull requests and/or commit history to guess what a PR/commit is doing.

In this specific case, I would have chosen something like:
- `Drop trailing s from strings in webRequest docs`
- `Replace strings with string in webRequest docs`

I'll change the commit title as part of merging, so you don't have to do anything else right now. Thanks again!

PR #39143, opened 2025-04-16
Make all JS articles reachable via sidebar

Josh-Cena:
Add a few missing articles, also make the categories themselves clickable.

### DIFF
--- a/files/sidebars/jssidebar.yaml
+++ b/files/sidebars/jssidebar.yaml
@@ -15,7 +15,8 @@ sidebar:
         title: Dynamic_scripting_with_JavaScript
       - link: /Learn_web_development/Core/Frameworks_libraries
         title: JavaScript_frameworks_and_libraries
-  - title: Guide
+  - link: /Web/JavaScript/Guide
+    title: Guide
     details: closed
     children:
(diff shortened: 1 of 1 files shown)


files/sidebars/jssidebar.yaml
```diff
@@ -87,37 +88,43 @@ sidebar:
...
     path: /Web/JavaScript/Reference/Global_Objects
-    includeParent: true
+    link: /Web/JavaScript/Reference/Global_Objects
     title: Global_Objects
     details: closed
   - type: listSubPages
     path: /Web/JavaScript/Reference/Operators
-    includeParent: true
+    depth: 2
+    nested: true
```
dipikabh, line 97: Yeah, I notice this but the indentation does not stand out visually. I wish there was a drop-down available for the children.
```
import.meta
   import.meta.resolve()
```

dipikabh (approved):
Nice! LGTM. Thank you

files/sidebars/jssidebar.yaml
Josh-Cena: I do, too :/

PR #39144, opened 2025-04-16
Synchronize with BCD v6.0.6

OnkarRuikar:
[BCD release v6.0.6](https://github.com/mdn/browser-compat-data/releases/tag/v6.0.6)

/cc cc/ @bsmth @wbamberg


files/en-us/web/css/@media/index.md
```diff
@@ -164,17 +164,17 @@ Media feature expressions test for their presence, value, or range of values, an
   - : Combination of brightness, contrast ratio, and color depth that are supported by the video plane of user agent and the output device. Added in Media Queries Level 5.
 - {{cssxref("@media/width", "width")}}
   - : Width of the viewport including width of scrollbar.
-- {{cssxref("@media/-moz-device-pixel-ratio", "-moz-device-pixel-ratio")}} {{deprecated_inline}} {{non-standard_inline}}
+- {{cssxref("@media/-moz-device-pixel-ratio", "-moz-device-pixel-ratio")}}
```
Josh-Cena, line 167: As I said before, I think:

1. "Deprecated" and "non-standard" are relatively stable statuses and writers should be free to add them without worrying about them going stale.
2. I understand it's hard to tell where these macros should be _added_ if needed, so how about this: we divide DLs into three kinds:
   - Always-sync: what you already have. Like "parameters" or "instance properties". Macros are added and removed according to BCD.
   - Respect: these DLs contain links to pages and the pages have `browser-compat`, but they are not in a location you defined as always-sync. In this case: if the DL contains at least one status macros, then we sync as usual, and otherwise we don't add macros.
   - Always-remove: if the DL is not homogeneous (not every item is a link and/or they don't link to feature pages, etc.), then you remove status macros.

OnkarRuikar: > I understand it's hard to tell where these macros should be added if needed,

In this case it was simple: we don't use inline status macros on CSS pages. 

You'll have to convince @wbamberg to reinstate deprecated and non-standard inline macros on CSS pages.

bsmth (approved):
As discussed in the content call and on Discord, this is good to merge for now given we want to be able to automate "all or nothing" in this section as far as I understand. That means a +1 from me on this, and we're going to look at automating the macros like we do in other section again in future.

OnkarRuikar:
Thanks @bsmth !

PR #39162, opened 2025-04-16
chore(htmlsidebar): Add links to nested pages

dipikabh:
### Description

This is a follow-up to the recent HTML docs reorg (https://github.com/mdn/content/pull/39055).

1. This PR adds more pages to the sidebar.

| **Before** | **After** |
| --- | --- |
| <kbd>Sidebar before</kbd> | <kbd>Sidebar after adding more pages</kbd> |

2. I noticed that the `rel` pages were missing quotes around the values, so added those. I've also updated the title for `Alternate stylesheet` page to make it consistent with the others:

```diff
-- title: Alternate stylesheet
++ title: rel="alternate stylesheet"
```

### Motivation

Make pages in nested folders discoverable via the sidebar


files/sidebars/htmlsidebar.yaml
```diff
@@ -38,10 +33,39 @@ sidebar:
...
+        path: /Web/HTML/Reference/Elements/meta/name
+        tags: html-attribute-value
+        title: Meta_names
+        details: closed
+        code: true
+  - title: Attribute_values
+    details: closed
+    children:
+      - type: listSubPages
+        path: /Web/HTML/Reference/Attributes/rel
```
Josh-Cena, line 61: Instead of this, why don't we just increase the depth of listSubPages for attributes? We can do that for elements too.

dipikabh: I've tried that I don't like the rendering. That's what we have in Glossary as well.
I would prefer a drop-down detail-summary kind of functionality. Will discuss with Florian next week, he's away currently.

files/sidebars/htmlsidebar.yaml
```diff
@@ -38,10 +33,39 @@ sidebar:
...
+        tags: html-attribute-value
+        title: Input_types
+        details: closed
+        code: true
+      - type: listSubPages
+        path: /Web/HTML/Reference/Elements/script/type
+        tags: html-attribute-value
+        title: Script_types
+        details: closed
+        code: true
```
dipikabh, line 50: Note: For this and other `listSubPages`, I've deliberately not added the `link` key because the link takes you to another part of the sidebar.

files/sidebars/htmlsidebar.yaml
Josh-Cena: OK; I'm good with this being an interim solution, but I would prefer eventually having these under the elements/attributes lists

dipikabh: Yes, this is a temporary solution hopefully. Same, I want them to list under Elements/Attributes and I would prefer there be logic to list children if there are any nested pages, instead of maintaining this manually.
CSS sidebar uses `type: listSubPagesGrouped` and `tags` but these don't seem to work in the htmlsidebar.

pepelsbey (approved):
Looks good! I think it’s a perfectly viable solution until we have better rendering for subpages. Thank you :)

dipikabh:
Thanks a lot for the review, Vadim

PR #39166, opened 2025-04-16
background-clip: Fix interactive example

### LINKED ISSUE #38596: background-clip example using text-shadow is unnecessary

https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip

text-shadow on an interactive element prevents the "background-clip: text" property from being demonstrated
Screenshot_7

You just need to remove the text-shadow from the text "This is the content of the element."

so that it would be like this:
Screenshot_8

_No response_

_No response_

estelle:
all browsers now support the `text` example, so a a shadow is not longer needed behind the transparent text.
changed the text and shadow color to black and white to improve contrast on the dark image, and removed the no-longer required and breaking shadow when clipping to text.

Fixes https://github.com/mdn/content/issues/38596

### DIFF
--- a/files/en-us/web/css/background-clip/index.md
+++ b/files/en-us/web/css/background-clip/index.md
@@ -26,6 +26,7 @@ background-clip: content-box;
 ```css interactive-example-choice
 background-clip: text;
 color: transparent;
+text-shadow: none;
 ```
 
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
LGTM

PR #39170, opened 2025-04-17
Fix method name typo

wakfi:
`Math.expm1` was typo'd as `Math.exp1m` in one example


Josh-Cena (approved):
Great catch. Thank you!

PR #39173, opened 2025-04-17
deprecated clip property

estelle:
clarified that the rect() value is different from the rect() basic shape.
made the see also more relevant to the property.
removed "new" language.

part of https://github.com/openwebdocs/project/issues/224 and https://github.com/mdn/content/pull/39151


chrisdavidmills (approved):
@estelle LGTM, nice work.

PR #39185, opened 2025-04-18
Change CPP to other languages where appropriate

Josh-Cena:
Most of the code blocks marked as "CPP" are not real C++, but other C-like languages. This PR makes them more technically precise so readers know what kind of file to copy the code into.


Josh-Cena:
Please disregard the markdownlint warnings; they will be resolved once this config change is merged.

sideshowbarker (approved):

PR #39190, opened 2025-04-18
Typo in "visibility" Page

danielrentz:
Typo: "tr" -> "th"

### Motivation

Typo: Table cell elements are td and th (not tr)


Josh-Cena (approved):
Nice catch, thanks

PR #39205, opened 2025-04-20
Remove inherit value from some css properties

skyclouds2001:
### Description

the inherit keyword is a global keyword, so it should not list in the *Value* section


estelle (approved):
Thanks! 🎉

PR #39208, opened 2025-04-20
Clarify that privacy sandbox enrollment is currently a Chrome-only thing

Psychpsyo:
### Description

This makes some of the wording around the Privacy Sandbox Enrollment more specific to Chrome since these features are currently only available as experimental features in Chrome.
Furthermore, Mozilla is opposed to 4/5 of the listed features, while Apple is opposed to just the Topics API.
They both have yet to give a standards position on the remaining features.

### Motivation

I think it is disingenuous to frame Google's additions to their browser as a general concept for "Browsers" (plural) when other implementers seem to overall take issue with the features at hand.
Framing it that way seems to give too much credibility to something that is still up in the air and with a seemingly undecided future.

### Additional details

https://mozilla.github.io/standards-positions/ includes Mozilla's negative standards position for the [Fenced Frame API](https://developer.mozilla.org/en-US/docs/Web/API/Fenced_frame_API), Protected Audience API, [Shared Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Shared_Storage_API) and [Topics API](https://developer.mozilla.org/en-US/docs/Web/API/Topics_API).
https://webkit.org/standards-positions/#position-111 has Apple's opposition to the [Topics API](https://developer.mozilla.org/en-US/docs/Web/API/Topics_API).

Could not find any.

### DIFF
--- a/files/en-us/web/privacy/guides/privacy_sandbox/enrollment/index.md
+++ b/files/en-us/web/privacy/guides/privacy_sandbox/enrollment/index.md
@@ -6,7 +6,8 @@ page-type: guide
 sidebar: privacy
 ---
 
-To access certain privacy sandbox features, browsers require developers to complete an **enrollment** process.
+To access certain privacy sandbox features, Google Chrome requires developers to complete an **enrollment** process.
+These features are currently still experimental and only available in Chrome and Chromium-based browsers and may or may not make it into other browsers.
 
 Enrollment provides a mechanism to verify the entities that call privacy sandbox features, and to gather the developer-specific data needed to properly configure and use them. The enrollment process adds an additional layer of protections on top of the structural restrictions enforced within each feature by adding transparency to who is collecting data, and mitigating attempts to misuse features to gather more data than intended.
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @Psychpsyo . Despite some discussion MDN still does not handle the capture of standards positions particularly well - there is no easy or standard way to indicate that something is not just "not broadly supported", but also will never be broadly supported. It is even worse, because after two years being released in Chrome, the compatibility would change such that the whole thing is no longer experimental - so there isn't even any trigger for readers to question this.

At the top level of Privacy sandbox this is captured as Google Chrome specific, but not at this level.

Upshot, I agree the point, but not sure of the "solution". Let me see if I can get some advice from others on the team.

hamishwillee:
@Psychpsyo Following discussion with reviewers I'm closing this in favour of #39300

The background is that there is a discussion on this in https://github.com/orgs/mdn/discussions/463 which culminated in this proposal for reporting standards positions https://github.com/orgs/mdn/discussions/463#discussioncomment-8283921.
Right now we can make clear statements when there is opposition as per the linked PR.

I've added the same disclaimer up the top of the enrollment doc. This is probably not as clear on the current status quo as the changes you made, but it is inline with MDN policy and should be enough to highlight that there are issues.

Thanks very much for raising this. It's a bit of a can of worms.

hamishwillee:
> > because after two years being released in Chrome, the compatibility would change such that the whole thing is no longer experimental
> 
> Something is experimental as long as one and only one browser implements it, for no matter how long. The two-year rule is only relevant for feature deletion, when the last browser unships it for two years.

No. From https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/index.md#setting-experimental :

> A single-engine feature's experimental status may expire and switch to false when the following conditions are met:
> - The feature has been supported by default and without major changes by some browser for two or more years.

Josh-Cena:
Oh I didn't know that, TIL!

PR #39214, opened 2025-04-21
remove extra page word

familyboat:
### Motivation

To use consistent description for same thing, add page word. See https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/CompileError#see_also

### DIFF
--- a/files/en-us/webassembly/reference/javascript_interface/compile_static/index.md
+++ b/files/en-us/webassembly/reference/javascript_interface/compile_static/index.md
@@ -99,6 +99,6 @@ fetch("log-concat.wasm")
 
 ## See also
 
-- [WebAssembly](/en-US/docs/WebAssembly) overview page
+- [WebAssembly](/en-US/docs/WebAssembly) overview
 - [WebAssembly concepts](/en-US/docs/WebAssembly/Guides/Concepts)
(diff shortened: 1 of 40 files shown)


Josh-Cena:
Our see also guidelines suggest against adding extra words like "page" or "guide". Therefore we should remove "page" where possible

familyboat:
> Our see also guidelines suggest against adding extra words like "page" or "guide". Therefore we should remove "page" where possible

Where is the see also guidelines you mentioned? I just found [this related guideline](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide#see_also_section).
 
I will work on it.

Josh-Cena:
The "spirit" is here:

> - The link text should be the same as the title of the page or the section being linked to. For example, the link text to this [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes) page with the page title "ARIA states and properties" will be:
> - No article ("A", "An", "The") is needed at the beginning of the link list item. No punctuation is required at the end of the list item because it will invariably be a term or a phrase.

We generally discourage all extra words unless for disambiguation

familyboat:
> We generally discourage all extra words unless for disambiguation

Ok. Because English is not my mother language, so I decided to accept your advice.

Josh-Cena (approved):
Nice, thank you!

PR #39215, opened 2025-04-21
Document checkbox value

Zarel:
### Description

I added documentation for the thing where checkbox and radio elements actually have JavaScript value `"on"` rather than `""` if the value attribute is left off their HTML.

### Motivation

This is how browsers actually work, and the current text is incorrect as far as I can tell.

I'm not sure about this wording, but it seems clear to me that this behavior should be mentioned in _some_ way here, because of how confusing it is when not explicitly mentioned. I think you could make an argument that the current text is technically correct (it feels like a huge stretch), but that doesn't change how relevant it is to describe this behavior here.

### Additional details

I mean, here's the spec:

https://html.spec.whatwg.org/multipage/input.html#dom-input-value-default-on

### DIFF
--- a/files/en-us/web/api/htmlinputelement/value/index.md
+++ b/files/en-us/web/api/htmlinputelement/value/index.md
@@ -14,7 +14,7 @@ This property can also be set directly, for example to set a default value based
 
 ## Value
 
-A string containing the value of the {{htmlelement("input")}} element, or the empty string if the input element has no value set, unless the input element has a type of `checkbox` or `radio`, in which case it will be the string `"on"` if it has no value set (although the value can still be manually set to the empty string).
+A string specifying the default value of the {{htmlelement("input")}} element.
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/htmlinputelement/value/index.md
```diff
@@ -14,7 +14,7 @@ This property can also be set directly, for example to set a default value based
 
 ## Value
 
-A string containing the value of the {{htmlelement("input")}} element, or the empty string if the input element has no value set.
+A string containing the value of the {{htmlelement("input")}} element, or the empty string if the input element has no value set, unless the input element has a type of `checkbox` or `radio`, in which case it will be the string `"on"` if it has no value set (although the value can still be manually set to the empty string).
```
sideshowbarker, line 17:
```suggestion
A string specifying the default value of the {{htmlelement("input")}} element.
```

sideshowbarker (approved):

files/en-us/web/api/htmlinputelement/value/index.md
Zarel: I'm a little confused, to be honest? This seems like the opposite of where I was trying to do. And this also seems even less correct to me. This is the entry for the `.value` class property of the JavaScript `HTMLInputElement` DOM interface, not the HTML `value` attribute.

PR #39224, opened 2025-04-21
Add missing SourceBuffer event pages

low-perry:
### Description

This PR addresses the absence of [SourceBuffer](https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer) event documentation. The following changes have been made:
- Added detailed event pages for SourceBuffer
- Included code examples to demonstrate usage.

### Additional details

The event summary in the specification states that the `error` event is fired whenever an error occurs during `SourceBuffer.appendBuffer()`, with `SourceBuffer.updating` changing from `true` to `false`.
 However, as highlighted [here](https://www.w3.org/TR/media-source-2/#sourcebuffer-events:~:text=object%20to%20true.-,remove,-Removes%20media%20for), the `error` event is also fired during the `remove` method. This behaviour is described on the `error` page


Josh-Cena:
I don't have the bandwidth to review this and I didn't really intend to review it. I think the examples are too long and if we want to go with one like this we should have a single one and link to it from every event page, or we could simply remove the examples. I would leave to someone else in @mdn/web-api to review.

low-perry:
@Josh-Cena I will try to come up with a better solution for the examples.

Josh-Cena (approved):
I'm happy with this version. Thank you!

PR #39230, opened 2025-04-22
Add reference to Number()

sastani:
https://dev.to/darkmavis1980/you-should-stop-using-parseint-nbf

### DIFF
--- a/files/en-us/web/javascript/guide/grammar_and_types/index.md
+++ b/files/en-us/web/javascript/guide/grammar_and_types/index.md
@@ -293,24 +293,20 @@ With all other operators, JavaScript does _not_ convert numeric values to string
 
 In the case that a value representing a number is in memory as a string, there are methods for conversion.
 
-- {{jsxref("Number()")}}
 - {{jsxref("parseInt()")}}
 - {{jsxref("parseFloat()")}}
-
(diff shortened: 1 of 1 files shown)


files/en-us/web/javascript/guide/grammar_and_types/index.md
```diff
@@ -293,14 +293,19 @@ With all other operators, JavaScript does _not_ convert numeric values to string
 
 In the case that a value representing a number is in memory as a string, there are methods for conversion.
 
+- {{jsxref("Number()")}}
 - {{jsxref("parseInt()")}}
 - {{jsxref("parseFloat()")}}
 
+  While both `parseInt` and `Number` can be used to convert strings to numbers, `Number` is more efficient.
```
Josh-Cena, line 300: We avoid talking about performance in documentation unless something is designed to be more performant than something else. This is an engine implementation detail.

sastani: @Josh-Cena I think at the very least the article should mention `Number`

Josh-Cena: Yes, so my suggestion is to remove this paragraph and keep the mention

Josh-Cena (approved):
Thanks :)

PR #39232, opened 2025-04-22
Update PWA manifest id example: id and start_url must be the same origin

kynikos:
### Description

I updated the "Updating an existing app" example because the id and start_url values were on different origins, which as I understand it is not supported. 

### Motivation

The id and start_url values were on different origins, which as I understand it is not supported. 

### Additional details

-

-

### DIFF
--- a/files/en-us/web/progressive_web_apps/manifest/reference/id/index.md
+++ b/files/en-us/web/progressive_web_apps/manifest/reference/id/index.md
@@ -84,17 +84,17 @@ Consider a scenario where you deploy a web app with the following manifest:
 {
   "name": "My Weather Application",
   "id": "https://example.com/weatherapp/",
-  "start_url": "https://old-domain.com/app"
+  "start_url": "https://example.com/old-app"
 }
 ```
 
-However, you later decide to move the app to a different domain. You would then update the manifest as follows:
+However, you later decide to move the app to a different path. You would then update the manifest as follows:
 
 ```json
 {
   "name": "My Weather Application",
   "id": "https://example.com/weatherapp/",
-  "start_url": "https://new-domain.com/app"
+  "start_url": "https://example.com/new-app"
 }
 ```
 
--- a/files/en-us/web/progressive_web_apps/manifest/reference/id/index.md
+++ b/files/en-us/web/progressive_web_apps/manifest/reference/id/index.md
@@ -42,13 +42,29 @@ The `id` manifest member serves as a unique identifier for your web app. It allo
 
 The `id` can also be used by services that collect lists of web apps to uniquely identify applications.
 
+### Usage notes
+
 A few key points to remember while using the `id` member:
 
-- Use a leading `/` to specify that the `id` is a root-relative URL path.
+- As a recommended practice, use a leading `/` to specify that the `id` is a root-relative URL path.
 - Since `id` is resolved against `start_url`'s origin, `id` values such as `../foo`, `foo`, `/foo`, and `./foo` all resolve to the same identifier relative to the origin. For example, if `start_url` is `https://example.com/app/`, all these `id` values will resolve to `https://example.com/foo/`.
 - Standard URL encoding and decoding rules apply when resolving the `id` value.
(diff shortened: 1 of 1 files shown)


files/en-us/web/progressive_web_apps/manifest/reference/id/index.md
```diff
@@ -84,17 +84,17 @@ Consider a scenario where you deploy a web app with the following manifest:
...
 
-However, you later decide to move the app to a different domain. You would then update the manifest as follows:
+However, you later decide to move the app to a different path. You would then update the manifest as follows:
 
 ```json
 {
   "name": "My Weather Application",
   "id": "https://example.com/weatherapp/",
-  "start_url": "https://new-domain.com/app"
+  "start_url": "https://example.com/new-app"
```
janbrasna, line 97: > I updated the "Updating an existing app" example because the id and start_url values were on different origins, which as I understand it is not supported.

IIUC `id` doesn't have to actually match `start_url` hostname, it's usually kept intact even when moving across domains — the origin that _must_ match is the location of the manifest file itself related to the `start_url`…

janbrasna: I'm actually trying to infer what the spec says: https://www.w3.org/TR/appmanifest/#id-member 

_"The manifest's id member is a string that represents the identity for the application. The identity takes the form of a URL, which is same origin as the start URL."_

vs.

_"NOTE: The identity is processed like a URL but it doesn't point to a resource that can be navigated to, so it's not required to be within scope."_

(that is defined https://www.w3.org/TR/appmanifest/#dfn-within-scope e.g. _"A URL target is said to be within scope of URL scope if the following algorithm returns true: 1.) If target and scope are not same origin, return false. […]"_ which in this case sounds like it's allowed…)

Is it possible this has somehow changed over time?

kynikos: I'm really not sure about the official interpretation... However I'd like to point out that the "Updating an existing app" [1] section currently seems inconsistent with the "Understanding id resolution" [2] section, since the latter says "Cross-origin URL not allowed, falls back to start_url" in the table's penultimate row: with this patch I was mostly trying to solve that.

[1] https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/id#updating_an_existing_app
[2] https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/id#understanding_id_resolution

hamishwillee:
@dipikabh You worked on this recently right? I've replaced my assignment with yours - if you'd like to push it back you me, then that is OK.

dipikabh:
> You worked on this recently right? I've replaced my assignment with yours

Yes, that's right. I can take a look at these changes

files/en-us/web/progressive_web_apps/manifest/reference/id/index.md
dipikabh: @kynikos Thanks - these changes look good to me! The revised example now correctly keeps `id` and `start_url` on the same origin, which aligns with the spec and what we've also documented in the [`id` description](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/id#id):
> The URL must be same-origin with the [start_url](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/start_url) of your web app.

dipikabh: I took the liberty of pushing a few additional changes to your branch @kynikos, mainly to update some text for clarity and to move the section with the table to the "Description" section. (https://github.com/mdn/content/pull/39232/commits/28756aa8bb303f5335dc67a74777a1708a50da9e)

dipikabh (approved):
Thanks for sending this fix! Great job, especially for your first contribution 🙌 
Welcome to MDN 🎉

kynikos:
Thank you for all your work guys!

PR #39249, opened 2025-04-23
Fix typo

andrewkincaid:
I tested the example code in Google Chrome.  The `groups` actually returns `{ color: "Brown" }` with an uppercase `B` since the `input` is `The Quick Brown Fox Jumps Over The Lazy Dog`.  The docs currently show `brown` with a lowercase `b`.

### Description
Typo in `brown`.  Should be `Brown` given the `input` string.

### Motivation
The example code should be accurate.

### Additional details
None

None


Josh-Cena (approved):
Great catch, thanks

PR #39253, opened 2025-04-23
add relevant external link

estelle:
we don't cover this, or have the skills to do so, so think this external link is really useful.


chrisdavidmills (approved):
LGTM!

PR #39268, opened 2025-04-24
Correct an article overview

youssef-el-atmani:
### Description
I have just copied the correct overview from the  [Advanced styling effects](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Advanced_styling_effects) article.


Josh-Cena (approved):
Great catch, thank you

PR #39269, opened 2025-04-24
Remove polyfill link for :focus-visible

glmvc:
### Description

This pr removes the [link to the polyfill for :focus-visible in the "See also" section](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#see_also).

### Motivation

No need for a polyfill because of widely available (baseline) browser support.

### Additional details

[Browser support details for the :focus-visible CSS pseudo-class on Can I use...](https://caniuse.com/css-focus-visible)

I hope it's fine to make a small pr without an issue :)


estelle (approved):
Thanks! 🎉

PR #39271, opened 2025-04-25
Fix wrong ID

MrBrain295:
### Description
The ID wasn’t consistent with the things that referenced it.

### Motivation
Saw it was wrong.


Josh-Cena:
Hi, the current code is correct. `start` and `stop` are button IDs; `.go` and `.stop` are class names applied to the div via JavaScript.

PR #39272, opened 2025-04-25
Fix content issues

Josh-Cena:
Fix issues including:

- Typography issues
- shortened links
- CSS syntax errors
- HTML syntax errors


files/en-us/web/css/css_values_and_units/value_definition_syntax/index.md
```diff
@@ -112,9 +106,7 @@ But not:
 
 Separating two or more components by a _double bar_, `||`, means that all entities are options: **at least one must be present, and they may appear in any order**. Typically this is used to define the different values of a [shorthand property](/en-US/docs/Web/CSS/CSS_cascade/Shorthand_properties).
 
-```css
-<'border-width'> || <'border-style'> || <'border-color'>
-```
+{{CSSSyntaxRaw(`example = <'border-width'> || <'border-style'> || <'border-color'>`)}}
```
bsmth, line 109: This one expands out the definition for each (existing) property, so I'm not a fan of how it looks. What happens if we use hypothetical constituents, like:

```
{{CSSSyntaxRaw(`example = <'example-width'> || <'example-style'> || <'example-color'>`)}}
```

files/en-us/web/css/reference/index.md
```diff
@@ -12,7 +12,7 @@ Use this **CSS reference** to browse an [alphabetical index](#index) of all of t
 
 ### Style rule syntax
 
-```css
+```plain
```
bsmth, line 15: Outside of the scope of this PR, but noting for future us / discussion: I think we're surfacing a small problem here, is that BNF? Would something like this be better?

````md
```bnf
<style-rule> ::=
    <selectors-list> "{"
      <properties-list>
    "}"
```
````

And if so, I wonder how helpful this is for readers as it stands in the current place, it's one of the first thing you see, and it's maybe a little intimidating or potentially confusing if you're not familiar with the notation.

**Review by bsmth — APPROVED** [review]
All the changes are good, so +1 with comments about:

* linting / search-replace rule for shortlinks
* syntax blocks on the CSS reference page (editorial comment)

**Review by bsmth — CHANGES_REQUESTED** [review]
Sorry, I forgot one blocker for me which is the `||` example:

https://pr39272.review.mdn.allizom.net/en-US/docs/Web/CSS/CSS_Values_and_Units/Value_definition_syntax#double_bar

We should re-think that before merging

File: files/en-us/web/css/reference/index.md (hunk 2 above)
Comment (Josh-Cena) [reply files/en-us/web/css/reference/index.md:15]: I would prefer it to stay as-is because I think the CSS formal syntax doesn't use `"`.

Comment (bsmth) [reply files/en-us/web/css/reference/index.md:15]: OK, let's leave for now

File: .markdownlint.jsonc [hunk 3]
```diff
@@ -290,6 +290,13 @@
         "replace": "e.g., ",
         "searchScope": "text",
       },
+      {
+        "name": "short-link",
+        "message": "Use shortened links for bug links",
+        "searchPattern": "/https:\\/\\/bugzilla\\.mozilla\\.org\\/show_bug\\.cgi\\?id=/g",
```
bsmth, line 296: great, thanks 👍🏻

bsmth (approved):
Looks way better now, thanks 👍🏻

PR #39065, opened 2025-04-10
chore: Use front matter for learn sidebar

bsmth:
This PR moves the learn sidebar into front matter. Also including some hints for authoring.

PR #38269, opened 2025-02-21
Move Privacy guides into subfolder

pepelsbey:
### Moves Privacy guides into the `guides` folder

```diff
- privacy/firefox_tracking_protection/index.md
+ privacy/guides/firefox_tracking_protection/index.md
etc.
```

### Creates a landing pages for guides

```diff
+ privacy/guides/index.md
```

### Creates and uses front-matter sidebar

```diff
+ sidebar: privacy
- {{ListSubpages("/en-US/docs/Web/Privacy", "2", "0", "0")}}
```

### Refactor error pages and its landing

Error summaries are moved to the top:

```diff
+ A request to access cookies or storage was blocked because the browser is blocking all storage access.
```

The landing page now uses a macro to list all errors:

```diff
+ {{SubPagesWithSummaries}}
```

### Motivation

Docs IA & maintenance

PR #39274, opened 2025-04-25
chore: Move mdnsidebar into front matter

bsmth:
### Description

There are leftovers in this section that we should clean up.
Using front matter in all `/mdn` pages instead of macro has the side-effect that the page content should be updated, too.

Updating templates to use front matter for non-parameterized sidebars, this will likely be the default going forward.

### Motivation

Fewer macros embedded in prose and page layout / type defined in front matter.

- [x] https://github.com/mdn/content/pull/39065


files/en-us/mdn/writing_guidelines/page_structures/page_types/http_header_page_template/index.md
```diff
@@ -192,8 +198,9 @@ Developers can set and get HTTP headers using `fetch()` in order to provide appl
 
 ## See also
 
-Include links to reference pages and guides related to the current HTTP header. For more guidelines, see the [See also section](/en-US/docs/MDN/Writing_guidelines/Writing_style_guide#see_also_section) in the _Writing style guide_.
-You can link to relevant response statuses like `\{{HTTPStatus("123", "123 Response Status")}}` and headers like `\{{HTTPHeader("Header-Name")}}`.
+Include links to reference pages and guides related to the current HTTP header.
+For more guidelines, see the [See also section](/en-US/docs/MDN/Writing_guidelines/Writing_style_guide#see_also_section) in the _Writing style guide_.
+You can link to relevant response statuses like `\{{HTTPStatus("123", "123 (Reason)")}}` and headers like `\{{HTTPHeader("Header-Name")}}`.
```
Josh-Cena, line 203: Note, currently my checker only allows `100 Continue` and not `100 (Continue)`. Do you intend to change the convention? (I'm happy with that, if you think that is more appropriate)

Josh-Cena (approved):

files/en-us/mdn/writing_guidelines/page_structures/page_types/html_element_page_template/index.md
```diff
@@ -64,17 +66,14 @@ sidebar: mdnsidebar
 > - `\{{SecureContext_Header}}` — this generates a **Secure context** banner that indicates the technology is only available in a [secure context](/en-US/docs/Web/Security/Secure_Contexts).
```
Josh-Cena, line 66: I think we should remove this (and the one in SVG element). Elements don't make sense to be secure-context-only. And then we can remove the whole "You should update or delete the following macros according to the advice below:" part.

files/en-us/mdn/writing_guidelines/learning_content/index.md
```diff
@@ -3,10 +3,9 @@ title: Learn web development writing guidelines
 short-title: Learning content
 slug: MDN/Writing_guidelines/Learning_content
 page-type: mdn-writing-guide
+sidebar: mdnsidebar
```
hamishwillee, line 6: Just "FWIW", we no longer need to include sidebar in the name if we're always referencing a sidebar. I.e. this would have been better as

```
sidebar: mdn
```

files/en-us/mdn/writing_guidelines/page_structures/page_types/css_function_page_template/index.md
```diff
@@ -73,7 +72,7 @@ sidebar: mdnsidebar
 >
 > _Remember to remove this note block before publishing._
 
-{{CSSRef}}{{SeeCompatTable}}{{Deprecated_Header}}{{Non-standard_Header}}
+{{SeeCompatTable}}{{Deprecated_Header}}{{Non-standard_Header}}
```
hamishwillee, line 75: FWIW I wonder if it is time to also remove these macros and entirely generate the the headers from the status.

files/en-us/mdn/writing_guidelines/page_structures/page_types/css_property_page_template/index.md
```diff
@@ -39,6 +40,9 @@ sidebar: mdnsidebar
 >   - : Flags describing the status of this feature. An array which may contain one or more of the following: `experimental`, `deprecated`, `non-standard`. This key should not be set manually: it is set automatically based on values in the browser compatibility data for the feature. See ["How feature statuses are added or updated"](/en-US/docs/MDN/Writing_guidelines/Page_structures/Feature_status#how_feature_statuses_are_added_or_updated).
 > - **browser-compat**
 >   - : Replace the placeholder value <code>css.properties.NameOfTheProperty</code> with the query string for the property in the [Browser compat data repo](https://github.com/mdn/browser-compat-data/tree/main/css/properties). Check the _Other macros in the page_ section of this note block to see how this key-value is used to generate content for the _Specifications_ and _Browser compatibility_ sections.
+> - **sidebar**
+>   - : This is always `cssref`.
```
hamishwillee, line 44: Perhaps everywhere something like

```suggestion
>   - : This is `cssref` for CSS reference pages of any type.
```

files/en-us/mdn/writing_guidelines/page_structures/page_types/glossary_page_template/index.md
```diff
@@ -19,6 +19,7 @@ sidebar: mdnsidebar
 > title: Term being defined
 > slug: Glossary/Term_being_defined
 > page-type: glossary-definition OR glossary-disambiguation
+> sidebar: glossarysidebar
```
hamishwillee, line 22: Again, would have been better as 

```suggestion
> sidebar: glossary
```

files/en-us/mdn/writing_guidelines/page_structures/page_types/html_element_page_template/index.md
hamishwillee: Yes ^^^

files/en-us/mdn/writing_guidelines/page_structures/page_types/http_header_page_template/index.md
```diff
@@ -108,13 +108,19 @@ Two or three paragraphs in this section is appropriate, and if there are substan
...
 NameOfTheHeader: <directive1>
 NameOfTheHeader: <directive1>, <directive2>, …
 ```
 
+If the header has a lot of available directives, feel free to include multiple syntax boxes, subsections and explanations as appropriate:
+
+```http
+NameOfTheHeader: <directive3>,
+                 <directiveN>
+```
```
hamishwillee, line 122: Should this be multi-line? It feels non-standard to do it like this.

files/en-us/mdn/writing_guidelines/page_structures/page_types/http_header_page_template/index.md
hamishwillee: @bsmth We use `100 Continue` format in a lot of places. I think that does reflect the actual content you get in a response, at least in HTTP 1/1. So while I quite like the bracket format I don't think we should change this.

files/en-us/mdn/writing_guidelines/learning_content/index.md
bsmth: > Just "FWIW", we no longer need to include sidebar in the name if we're always referencing a sidebar. I.e. this would have been better as
> 
> ```
> sidebar: mdn
> ```

Yeah this is really leftovers due to migrating everything from Yari land, naming was kept so existing sidebars worked in the new world. I think we've broken that convention in some cases (like in the Privacy section -> https://github.com/mdn/content/pull/38269) - and totally agree this is much nicer not to have to duplicate 'sidebar' if possible :)

files/en-us/mdn/writing_guidelines/page_structures/page_types/css_function_page_template/index.md
bsmth: This is on Ruth's list of things to derive, yeah (under the "Banner overload" umbrella). We'll be looking at this pattern more seriously later in '25 to find a nicer solution, also how this information is displayed on the page.

files/en-us/mdn/writing_guidelines/page_structures/page_types/glossary_page_template/index.md
bsmth: totally agree

files/en-us/mdn/writing_guidelines/page_structures/page_types/http_header_page_template/index.md
bsmth: I believe we have some examples that do this for readability https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#strict_csp although it breaks syntax highlighting.

I don't mind making it one line only here so we encourage this over multi-line examples

files/en-us/mdn/writing_guidelines/page_structures/page_types/http_header_page_template/index.md
```diff
@@ -108,13 +108,19 @@ Two or three paragraphs in this section is appropriate, and if there are substan
...
 ```http
 NameOfTheHeader: <directive1>
 NameOfTheHeader: <directive1>, <directive2>, …
 ```
 
+If the header has a lot of available directives, feel free to include multiple syntax boxes, subsections and explanations as appropriate:
+
+```http
+NameOfTheHeader: <directive3>,
+                 <directiveN>
```
bsmth, line 121:
```suggestion
NameOfTheHeader: <directive3>, …, <directiveN>
```

files/en-us/mdn/writing_guidelines/page_structures/page_types/http_header_page_template/index.md
bsmth: OK then, let's not change this convention. It's true that readers are more likely to see the reason without parens, so the "optional" aspect is moreso "good to know" that we can describe in more detail on the message anatomy docs instead of every HTTPStatus macro.

bsmth, line 202:
```suggestion
You can link to relevant response statuses like `\{{HTTPStatus("123", "123 Reason")}}` and headers like `\{{HTTPHeader("Header-Name")}}`.
```

files/en-us/mdn/writing_guidelines/page_structures/page_types/css_property_page_template/index.md
bsmth: I think the whole CSS section uses one sidebar, so the second one applies

files/en-us/mdn/writing_guidelines/page_structures/page_types/css_module_landing_page_template/index.md
```diff
@@ -44,13 +45,9 @@ sidebar: mdnsidebar
...
 >
-> ---
->
-> **Top-of-page macros**
->
-> The `\{{CSSRef}}` macro call appears at the top of the content section (immediately after the page front matter).
-> This macro must be present on every CSS module landing page. It generates a suitable CSS sidebar, depending on the tags included on the page.
-> Remove the `\{{MDNSidebar}}` macro when you use this template.
+> - **sidebar**
+>   - : This is always `cssref`.
```
bsmth, line 49:
```suggestion
>   - : This is `cssref` for all CSS guide and reference pages.
```

files/en-us/mdn/writing_guidelines/page_structures/page_types/css_selector_page_template/index.md
```diff
@@ -45,6 +46,10 @@ sidebar: mdnsidebar
 >     Note that you may first need to create/update an entry for the selector and its specification in our <a href="https://github.com/mdn/browser-compat-data">Browser compat data repo</a>.
 >     See our [guide on how to do this](/en-US/docs/MDN/Writing_guidelines/Page_structures/Compatibility_tables).
 >
+> - **sidebar**
+>   - : This is always `cssref`.
```
bsmth, line 50:
```suggestion
>   - : This is `cssref` for all CSS guide and reference pages.
```

files/en-us/mdn/writing_guidelines/page_structures/page_types/css_module_landing_page_template/index.md
bsmth: done in 8c34575469b12152ae817962644d891d1504a010

bsmth:
Thanks, both! I'm going to merge now to get the changes in, but I think we can revisit the templates in general. I don't really like stuffing everything into a note block in the top of the page, and there's a lot of duplication. Compared to https://pr39274.review.mdn.allizom.net/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types/ARIA_Page_Template which is a bit easier to read, that's an improvement, IMO. Or we can look at actual templating that rari can consume / produce.

files/en-us/mdn/writing_guidelines/page_structures/page_types/css_property_page_template/index.md
hamishwillee: Great. In https://github.com/mdn/content/pull/39284#discussion_r2064563456 this was changed for the attribute template to:

> Keep as `htmlsidebar` (all pages under `/web/html/` use this sidebar).

FWIW I am a little ambivalent about what approach is better. What I suggested is more helpful, but this variant is more strictly likely to be true.

PR #39287, opened 2025-04-26
CSS properties: mask-mode and mask-type

estelle:
part of https://github.com/openwebdocs/project/issues/224 and https://github.com/mdn/content/pull/39151


files/en-us/web/css/mask-type/index.md
```diff
@@ -19,149 +17,121 @@ mask-type: luminance;
...
+## Formal syntax
 
-mask {
-  mask-type: alpha;
-}
-```
+{{csssyntax}}
 
-#### Result
+## Examples
```
dipikabh, line 55: - Though optional, would it be helpful for readers if there's a second example using `<div>` elements instead of ``? It could make it easier to focus directly on the masking effect without image details getting in the way.
- Also, how about demo'ing the effect of different `mask-mode` values, especially the effect of the non-default value. (similar to what's in the spec)

files/en-us/web/css/mask-type/index.md
```diff
@@ -19,149 +17,121 @@ mask-type: luminance;
...
-        fill="yellow"
-        fill-opacity="0.7" />
-    </mask>
-  </defs>
-</svg>
-```
+If the `<mask>` is defined as the mask image's source, and the `mask-mode` is set or defaults to `match-source`, the `mask-mode` will resolve to the `<mask>` element's `mask-type` value; `luminance` or `alpha`. If not explicitly set, the value defaults to `luminance`.
 
-#### CSS
+If the mask image source is not an SVG `<mask>`, this property has no effect.
```
dipikabh, line 46: - and then somewhere here, move the `mask-mode` from earlier
- Your earlier sentence, "The property defines the resolved value of the {{cssxref("mask-mode")}} property value if the `mask-mode` property is explicitly set to, or defaults to, `match-source`.", is a bit hard to follow. WDYT about this phrasing.
- Also adding another sentence to contrast between mask-type and mask-mode
```suggestion
### Effect of `mask-mode` on `mask-type`

While the `mask-type` property is set on the SVG `<mask>` element, the {{cssxref("mask-mode")}} property is set on the element being masked (the element you're applying the mask to).

The default value of `mask-mode` is `match-source`, which means the browser uses the `mask-type` value from the `<mask>` element to determine how to interpret it. If `mask-mode` is set to a value other than `match-source`, that value takes precedence and overrides the `mask-type` value of the applied mask.
```

files/en-us/web/css/mask-type/index.md
```diff
@@ -19,149 +17,121 @@ mask-type: luminance;
...
+- `alpha`
+  - : Indicates that the alpha values of the `<mask>` should be used.
+- `luminance`
+  - : Indicates that the luminance values of the `<mask>` should be used.
 
-{{csssyntax}}
+## Description
 
-## Examples
+The `mask-type` property is only relevant when applied to the `<mask>` element. The property defines the resolved value of the {{cssxref("mask-mode")}} property value if the `mask-mode` property is explicitly set to, or defaults to, `match-source`. If `mask-mode` is set to any `<masking-mode>` value other than `match-source`, that value takes precedence, overriding the `mask-type` value of the applied mask.
```
dipikabh, line 37: - I suggest we first explain mask-type, its values before explaining how `mask-mode` interacts with it. So this part about `mask-mode` could be moved later.
```suggestion
The `mask-type` property is only relevant for the SVG `<mask>` element. If you set `mask-type: luminance`, the mask will use how bright each part of the mask is; if you set `mask-type: alpha`, it will use how transparent or opaque each part is.
```

dipikabh (changes requested):
Review for the mask-type page

files/en-us/web/css/mask-mode/index.md
```diff
@@ -28,19 +28,33 @@ mask-mode: revert-layer;
...
+
 - `luminance`
-  - : This keyword indicates that the luminance values of the mask layer image should be used as the mask values.
+
+  - : Indicates that the luminance values of the mask layer image should be used as the mask values.
+
 - `match-source`
+  - : If the {{cssxref("mask-image")}} property values is a reference to an SVG {{svgelement("mask")}}, the `<mask>` element's {{cssxref("mask-type")}} property value is used, or it's {{svgattribute("mask-type")}} attribute, if present. If neither is explicitly set, in the case of the `<mask>` element as the mask source, this value defaults to `luminance`. Otherwise, if the mask image source is an {{cssxref("image")}}, rather than an SVG `<mask>`, the `alpha` values of the mask layer image is used.
+
+### Description
```
dipikabh, line 46: There are several concepts covered in this section. It would be easier for readability (quick scan) and clarity if the prose here is split into logical subsections.

files/en-us/web/css/mask-mode/index.md
```diff
@@ -28,19 +28,33 @@ mask-mode: revert-layer;
...
+A mask transfers its transparency and, depending on the mask type, it's luminescence, to the element it is masking.
+If the mask is of type {{cssxref("&lt;image&gt;")}}, by default the alpha values of the mask layer image is used as the mask values; where the mask is opaque, the pixels of the masked object are opaque. Where the mask is translucent, the element is as well, with those areas of the element being hidden. This is what happens with `<image>` masks, when the `mask-mode` property is set or defaults to `match-source`, and in all cases when the `mask-mode` is explicitly set to `alpha`.
 
-    If it is of type {{cssxref("&lt;image&gt;")}}, the alpha values of the mask layer image should be used as the mask values.
+In the case of `luminance` masks, whether the areas masked by an opaque mask are opaque themselves depends partially on the lightness of the color of the opaque areas. White opaque areas (100% luminance) will be masked and visible, transparent and black areas (0% luminance) will be clipped, and anything in between will be partially masked, reflecting the luminance and alpha-transparency of each color making up the mask.
```
dipikabh, line 51: - How about simplifying the 1st sentence slightly
- Can you check the accuracy of "black areas (0% luminance) transparent (alpha = 0) will be clipped."?
```suggestion
In the case of `luminance` masks, the visibility of the masked element depends on both the opacity of the mask and the lightness of the color of the opaque areas. White (100% luminance) opaque areas (alpha = 1) will be masked and visible, and black areas (0% luminance) transparent (alpha = 0) will be clipped. Areas with colors in between white and black and with partial opacity will be partially masked, reflecting the luminance and alpha transparency of each color making up the mask.
```

files/en-us/web/css/mask-mode/index.md
```diff
@@ -52,24 +66,61 @@ The `mask-mode` property is specified as one or more of the keyword values liste
...
+  mask-mode: match-source;
+}
 ```
 
-{{EmbedLiveSample("mask-mode-example", "", "250px")}}
+#### Results
+
+{{EmbedLiveSample("Using mask-mode", "", "250px")}}
+
+Because the mask source is an `<image>`, and not a `<mask>`, the `match-source` value resolves to `alpha`.
```
dipikabh, line 123:
```suggestion
Because the mask source is an `<image>` and not an SVG `<mask>`, the `match-source` value resolves to `alpha`.
```

files/en-us/web/css/mask-mode/index.md
```diff
@@ -28,19 +28,33 @@ mask-mode: revert-layer;
...
+In the case of `luminance` masks, whether the areas masked by an opaque mask are opaque themselves depends partially on the lightness of the color of the opaque areas. White opaque areas (100% luminance) will be masked and visible, transparent and black areas (0% luminance) will be clipped, and anything in between will be partially masked, reflecting the luminance and alpha-transparency of each color making up the mask.
+
+The opacity of a `luminance` mask is determined by the `R`, `G`, `B`, and `A` values of an `rgb()` color. The equation is `((0.2125 * R) + (0.7154 * G) + (0.0721 * B)) * A`. In a `luminance` mask, the color `green` being `#008000`, or `rgb(0% 50% 0% / 1)`, any area masked by a solid `green` mask will be `35.77%` opaque. If the `mask-mode` for this mask is set to `alpha`, as `green` is a fully opaque color, the masked area will be `100%` opaque.
+
+Each `mask-mode` value in the comma-separated list of values. When multiple values are present, each value will match up to the image in the same position in the {{cssxref("mask-image")}} value. The value defines whether the associated mask image is treated as a `luminance` or `alpha` mask.
```
dipikabh, line 55:
```suggestion
Each `mask-mode` value is a comma-separated list of values. When multiple values are present, each value corresponds to a mask layer in the same position in the {{cssxref("mask-image")}} property. The values define whether the associated mask images are treated as a `luminance` or an `alpha` mask.
```

files/en-us/web/css/mask-mode/index.md
```diff
@@ -28,19 +28,33 @@ mask-mode: revert-layer;
...
+
+In the case of `match-source`, whether `luminance` or `alpha` is used depends on the mask mode of the mask source. If the `mask-image` property values is a reference to an SVG {{svgelement("mask")}} element, the `<mask>` element's {{cssxref("mask-type")}} property value is used. If there is no CSS `mask-type` property set on the `<mask>` element, the value of the `<mask>` element's {{svgattribute("mask-type")}} attribute is used, if present and supported. If neither is explicitly set, this value defaults to `luminance`; but only in the case of the `<mask>` element as the mask source. Otherwise, as noted before, if the mask image source is an {{cssxref("image")}}, rather than an SVG `<mask>`, the `alpha` values of the mask layer image is used.
```
dipikabh, line 57: "{{svgattribute("mask-type")}}" rendering is an issue

files/en-us/web/css/mask-mode/index.md
dipikabh, line 46:
```suggestion
## Description
```

dipikabh (changes requested):
Review for mask-mode

files/en-us/web/css/mask-mode/index.md
estelle, line 57:
```suggestion
In the case of `match-source`, whether `luminance` or `alpha` is used depends on the mask mode of the mask source. If the `mask-image` property values is a reference to an SVG {{svgelement("mask")}} element, the `<mask>` element's {{cssxref("mask-type")}} property value is used. If there is no CSS `mask-type` property set on the `<mask>` element, the value of the `<mask>` element's {{svgattr("mask-type")}} attribute is used, if present and supported. If neither is explicitly set, this value defaults to `luminance`; but only in the case of the `<mask>` element as the mask source. Otherwise, as noted before, if the mask image source is an {{cssxref("image")}}, rather than an SVG `<mask>`, the `alpha` values of the mask layer image is used.
```

files/en-us/web/css/mask-type/index.md
```diff
@@ -19,149 +17,121 @@ mask-type: luminance;
 mask-type: alpha;
 
 /* Global values */
-mask-type: inherit;
```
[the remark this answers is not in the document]
estelle: oops, that was a mistake. Thanks.

files/en-us/web/css/mask-type/index.md
```diff
@@ -173,4 +161,7 @@ mask {
 
 ## See also
 
-- Other mask-related properties: {{cssxref("mask")}}, {{cssxref("mask-mode")}}
+- {{cssxref("mask")}}
+- {{cssxref("mask-mode")}}
+- [CSS masking](/en-US/docs/Web/CSS/CSS_masking) module
+- SVG {{svgattr("mask-type")}} attribute
```
estelle, line 167: not sure why the preview is showing a missing end quote here.

files/en-us/web/css/mask-mode/index.md
estelle: fixed. {{svgattr("mask-type")}}
that page is also now live.

estelle:
Thanks @dipikabh. I think i caught everything and SVG `mask-type` attribute page is now live, so red should be gone.

files/en-us/web/css/mask-mode/index.md
```diff
@@ -28,19 +28,41 @@ mask-mode: revert-layer;
...
+
+  - : Indicates that the alpha (transparency) values of the mask image should be used.
+
 - `luminance`
-  - : This keyword indicates that the luminance values of the mask layer image should be used as the mask values.
+
+  - : Indicates that the [luminance (brightness)](#understanding-luminance) values of the mask image should be used.
+
 - `match-source`
+  - : Indicates that the type of mask is determined by the source. The default property value.
```
dipikabh, line 44:
```suggestion
  - : Indicates that the type of mask is determined by the source. This is the default property value.
```

files/en-us/web/css/mask-mode/index.md
```diff
@@ -28,19 +28,41 @@ mask-mode: revert-layer;
...
+In the case of `luminance` masks, the visibility of the masked element depends on both the opacity of the mask and the lightness of the color of the opaque areas. White (100% luminance) opaque areas (alpha = 1) will be masked and visible, and black areas (0% luminance) transparent (alpha = 0) will be clipped. Areas with colors in between white and black and with partial opacity will be partially masked, reflecting the luminance and alpha transparency of each color making up the mask.
 
-    If it is of type {{cssxref("&lt;image&gt;")}}, the alpha values of the mask layer image should be used as the mask values.
+The opacity of a `luminance` mask is determined by the `R`, `G`, `B`, and `A` values of an `rgb()` color. The equation is `((0.2125 * R) + (0.7154 * G) + (0.0721 * B)) * A`. In a `luminance` mask, the color `green` being `#008000`, or `rgb(0% 50% 0% / 1)`, any area masked by a solid `green` mask will be `35.77%` opaque. If the `mask-mode` for this mask is set to `alpha`, as `green` is a fully opaque color, the masked area will be `100%` opaque.
```
dipikabh, line 57: - making the formula more explicit
- clarifying the example slightly
```suggestion
The opacity of a `luminance` mask is determined by the `R`, `G`, `B`, and `A` values of an `rgb()` color using the formula:

`((0.2125 * R) + (0.7154 * G) + (0.0721 * B)) * A`

For example, the color `green` is `#008000` or `rgb(0% 50% 0% / 1)`. In a `luminance` mask, any area masked by a solid `green` mask will be `35.77%` opaque. If the `mask-mode` for this mask is set to `alpha`, since `green` is a fully opaque color, the masked area will be `100%` opaque.
```

files/en-us/web/css/mask-type/index.md
```diff
@@ -26,142 +24,132 @@ mask-type: revert-layer;
...
-  - : Is a keyword indicating that the associated mask image is an alpha mask, i.e., that its [alpha channel](https://en.wikipedia.org/wiki/Alpha_compositing) values must be used when applying it.
+  - : Indicates that the alpha (transparency) values of the `<mask>` should be used.
+- `luminance`
+  - : Indicates that the [luminance (brightness) values](#understanding-luminance) of the `<mask>` should be used.
 
-## Formal definition
+## Description
 
-{{cssinfo}}
+The `mask-type` property is only relevant for the SVG `<mask>` element. If you set `mask-type: luminance`, the mask will use how bright each part of the mask is; if you set `mask-type: alpha`, it will use how transparent or opaque each part is.
```
dipikabh, line 36:
```suggestion
The `mask-type` property is only relevant for the SVG `<mask>` element. If you set `mask-type: luminance`, the mask will use how bright each part of the mask is; if you set `mask-type: alpha`, it will use how transparent or opaque each part of the mask is.
```

files/en-us/web/css/mask-type/index.md
```diff
@@ -26,142 +24,132 @@ mask-type: revert-layer;
...
 
-## Formal syntax
+### Default behavior
 
-{{csssyntax}}
+By default, the `<mask>` element uses `mask-type: luminance`. This means both the color and the transparency of the mask content affect masking. Whether the mask is opaque partially depends on the lightness of the color of the opaque areas:
```
dipikabh, line 40:
```suggestion
By default, the SVG `<mask>` element uses `mask-type: luminance`. This means both the color and the transparency of the mask content affect masking. Whether the mask is opaque partially depends on the lightness of the color of the opaque areas:
```

files/en-us/web/css/mask-type/index.md
```diff
@@ -26,142 +24,132 @@ mask-type: revert-layer;
...
 
-#### CSS
+For example, the color `green` (`#008000` or `rgb(0% 50% 0% / 1)`) has a luminance value of `35.77%`. Any area masked by a solid `green` luminance mask will be `35.77%` visible. If the `mask-type` is set to `alpha`, the same fully opaque `green` color will make the masked area `100%` visible.
 
-```css live-sample___mask-type-alpha-example
-* {
-  margin: 0px;
-  padding: 0px;
-}
+If the SVG `<mask>` element has `fill="rgb(0 0 0 / 0.5)"`, which is a 50% transparent black, the corresponding shape on the masked element will display at 50% opacity when using `alpha` as the value, as it is a color at 50% opacity, but if the `mask-type` defaults or is explicitly set to `luminance`, as the luminance is `0`, the masked area will be clipped and invisible.
```
dipikabh, line 54: Very long sentence
```suggestion
If the SVG `<mask>` element has `fill="rgb(0 0 0 / 0.5)"`, which is a 50% transparent black, the corresponding shape on the masked element will display at 50% opacity when `mask-type ` is set to `alpha` because the alpha values is `0.5` (50% opacity). But if the `mask-type` defaults to or is set as `luminance` the masked area will be fully clipped and invisible because its luminance is `0`.
```

files/en-us/web/css/mask-type/index.md
```diff
@@ -26,142 +24,132 @@ mask-type: revert-layer;
...
+If the mask image source is not an SVG `<mask>`, this property has no effect.
 
-mask {
-  mask-type: alpha;
-}
-```
+The default value of `mask-mode` is `match-source`, which means the browser uses the `mask-type` value from the `<mask>` element to determine how to interpret it. If `mask-mode` is set to a value other than `match-source`, that value takes precedence and overrides the `mask-type` value of the applied mask.
 
-#### Result
+If the `<mask>` is defined as the mask image's source, and the `mask-mode` is set or defaults to `match-source`, the `mask-mode` will resolve to the `<mask>` element's `mask-type` value; `luminance` or `alpha`. If not explicitly set, the value defaults to `luminance`.
```
dipikabh, line 63:
```suggestion
If the `<mask>` is defined as the mask image's source, and the `mask-mode` is set as or defaults to `match-source`, the `mask-mode` will resolve to the `<mask>` element's `mask-type` value: `luminance` or `alpha`. If not explicitly set, the value defaults to `luminance`.
```

files/en-us/web/css/mask-type/index.md
```diff
@@ -26,142 +24,132 @@ mask-type: revert-layer;
...
-        x=".2"
-        y=".1"
-        width=".4"
-        height=".8"
-        fill="yellow"
-        fill-opacity="0.7" />
-    </mask>
-  </defs>
+We've included two images that will be masked. Other than their class names, the two images are identical.
+We've also included an SVG with two `<mask>` elements. Other than their `id` values, the two masks are also identical; each with a green and white heart shape and a semi-transparent white and black filled circle.
```
dipikabh, line 82:
```suggestion
We've also included an SVG with two `<mask>` elements. Other than their `id` values, the two masks are also identical: each has a green and white heart shape and a semi-transparent white-and-black-filled circle.
```

dipikabh (approved):
Thanks a lot for all the updates, @estelle!
I've suggested a few more improvements. But leaving a +1.

PR #39292, opened 2025-04-27
add a example `children` 

zhangenming:
### Description
These two methods are the most commonly used, so I add this one

and more
Document.getElementsByTagName
Element.getElementsByTagName
Element.getElementsByClass
Element.children


Josh-Cena (commented):
Thanks. When we added this section, I was very deliberate to only add one example for every such interface, because most of them have more than one source, and I want to avoid the question of "why did you show this one and not the other? Why don't you add X as well?"

zhangenming:
image
but this page list two api... I think its more better

>why did you show this one and not the other? Why don't you add X as well?
Because these two are the most commonly used, I think listing them here is the most meaningful

zhangenming:
I now find that a better approach might be to put this more useful information at the top for easier reading

PR #39293, opened 2025-04-27
add more info about NodeList

zhangenming:
https://dom.spec.whatwg.org/#ref-for-dom-parentnode-queryselectorall%E2%91%A0:~:text=The%20querySelectorAll(selectors)%20method%20steps%20are%20to%20return%20the%20static%20result%20of%20running%20scope%2Dmatch%20a%20selectors%20string%20selectors%20against%20this.

I search spec, only this item return static nodelist

### DIFF
--- a/files/en-us/web/api/nodelist/index.md
+++ b/files/en-us/web/api/nodelist/index.md
@@ -7,7 +7,7 @@ browser-compat: api.NodeList
 
 {{APIRef("DOM")}}
 
-**`NodeList`** objects are collections of [nodes](/en-US/docs/Web/API/Node), usually returned by properties such as {{domxref("Node.childNodes")}} and methods such as {{domxref("document.querySelectorAll()")}}.
+**`NodeList`** objects are collections of [nodes](/en-US/docs/Web/API/Node), usually returned by properties such as {{domxref("Node.childNodes")}} and methods such as {{domxref("document.querySelectorAll()")}}. only {{domxref("document.querySelectorAll()")}} returns a static NodeList, others APIs returns a live NodeList.
 
 This interface was an [attempt to create an unmodifiable list](https://stackoverflow.com/questions/74630989/why-use-domstringlist-rather-than-an-array/74641156#74641156) and only continues to be supported to not break code that's already using it. Modern APIs represent list structures using types based on JavaScript [arrays](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array), thus making many array methods available, and at the same time imposing additional semantics on their usage (such as making their items read-only).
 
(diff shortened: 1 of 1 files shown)


zhangenming:
and this my note... 
image

Josh-Cena:
I merged this sentence into where live/static are actually introduced. What do you think?

zhangenming:
I just felt that thoses knowledge was very fragmented. It took me some effort to figure it out and summarize this table. I just hope it will be helpful to others, or The next time I check


I think these two apis HTMLCollection/NodeList
The most crucial information is
1. Which apis are called
2. return what (node or element)
3. live or static 

but i fell those two page did not so good

maybe we can Introduce a new paragraph to summarize the differences between the two apis

thanks

hope MDN better

wbamberg:
I don't think this PR is an improvement. What kind of thing an API returns is a matter for the API that returns it, not for the thing that is returned. So pages that return (e.g.) static `NodeList` instances should document that that is what they return.

Exhaustively listing the APIs that return a type is a trap, because if a new API is created that returns the type, we have no way to know we have to update the list, so the list will become incorrect. I don't think there was anything wrong with the existing page, here.

Josh-Cena:
> Exhaustively listing the APIs that return a type is a trap, because if a new API is created that returns the type, we have no way to know we have to update the list, so the list will become incorrect. I don't think there was anything wrong with the existing page, here.

This interface is essentially deprecated in the spec though, so we won't see more things that return it.

wbamberg:
> > Exhaustively listing the APIs that return a type is a trap, because if a new API is created that returns the type, we have no way to know we have to update the list, so the list will become incorrect. I don't think there was anything wrong with the existing page, here.
> 
> This interface is essentially deprecated in the spec though, so we won't see more things that return it.

Yes, in this particular case it is unlikely (although not impossible, if a new API wants to return the same thing as an existing API for compatibility/congruence with it). But even so, the general principle applies.

zhangenming:
> Exhaustively listing the APIs that return a type is a trap, because if a new API is created that returns the type, we have no way to know we have to update the list, so the list will become incorrect. I don't think there was anything wrong with the existing page, here.

if MDN can find a way to listing all the APIs, and Keep updated at all times, will so good
Of course, if the effect is not usefully, it might be necessary to weigh it up



My strategy is not to add all the apis either
Instead, several of the most commonly used apis were added (childNodes, getElementsByName, querySelectorAll)
This might be the most helpful

Josh-Cena:
@wbamberg IMO this is an improvement, is non-obstrusive, and does not pose maintenance burden. WDYT?

Josh-Cena (approved):

PR #39294, opened 2025-04-27
Fix case of `Document.styleSheets` reference in DOM scripting introduction

noahlemen:
the `styleSheets` attribute on `Document` is not all-lowercase: https://drafts.csswg.org/cssom/#extensions-to-the-document-or-shadow-root-interface


Josh-Cena (approved):
Nice, thank you!

PR #39299, opened 2025-04-28
Fix typo: Add missing quote

dipikabh:
### Description

Adds the missing quote in the `glossary` macro

### Motivation

To correctly render the glossary link


estelle (approved):
Thanks! 🎉

PR #39300, opened 2025-04-28
Privacy sandbox APIs - add more standards positions

hamishwillee:
This attempts to fix the problems highlighted in #39208 with the pattern agreed in https://github.com/orgs/mdn/discussions/463#discussioncomment-8283921

In summary, where there is clear opposition to a spec we add a warning up the top, and in the specifications provide a link to the known standards position.

Replaces #39208


files/en-us/web/privacy/guides/privacy_sandbox/enrollment/index.md
```diff
@@ -6,6 +6,10 @@ page-type: guide
 sidebar: privacy
 ---
 
+> [!WARNING]
```
hamishwillee, line 9: This enrollment process is chrome specific - not in any spec - and the position of browsers on this is unknown. I think it likely they consider it irrelevant at this point since they haven't yet done any implementation on any of this. 

It may be that this becomes some kind of specification relevant thing in future (though I doubt it) but at this point we simply don't know. I have therefore done the same thing as in the top level Privacy sandbox docs and highlighted that this is opposed. They should infer from this that the enrollment process is Chrome specific.

bsmth (approved):
Thanks, Hamish. I discussed out-of-band with the team, and relaying a +1 that this is good to merge 👍🏻

PR #39316, opened 2025-04-28
Fix plaindate.compare return value explanation

codegeordie:
### Description & Motivation

There's a typo/error in the explanation of the return value of Temporal.PlainDate.compare(), as it reads currently it repeats the first set of conditions.


codegeordie:
Actually the error exists in all of the various `.compare()` pages.

Josh-Cena (approved):
Great changes. Thank you!

PR #39318, opened 2025-04-28
New page: SVG mask-type attribute

estelle:
There are a few links to this page, but they were red, and the attribute is not in the SVG spec, but rather in masking. So I asked, and got confirmed:

"SVG presentational attributes have the identical definition to a CSS property of the same name."

https://mastodon.scot/@svgeesus/114405776420182252

So, this attribute does exist and should be documented.


files/en-us/web/svg/reference/element/mask/index.md
```diff
@@ -45,6 +45,8 @@ svg {
 - {{SVGAttr("height")}}
   - : This attribute defines the height of the masking area.
     _Value type_: [**\<length>**](/en-US/docs/Web/SVG/Guides/Content_type#length); _Default value_: `120%`; _Animatable_: **yes**
+- {{SVGAttr("height")}}
```
hamishwillee, line 48:
```suggestion
- {{SVGAttr("mask-type")}}
```

files/en-us/web/svg/reference/element/mask/index.md
```diff
@@ -45,6 +45,8 @@ svg {
...
+- : This attribute defines the mask mode for the contents for the contents of the `<mask>`.
```
hamishwillee, line 49: Can we also have a `_Value type_: ` line as for the other attributes? I presume luminance and alpha

hamishwillee: Note, rendering of the list is broken

files/en-us/web/svg/reference/attribute/mask-type/index.md
```diff
@@ -0,0 +1,98 @@
...
+      <th scope="row">Animatable</th>
+      <td>Discrete</td>
+    </tr>
+  </tbody>
+</table>
+
+- `alpha`
+  - : This value indicates that the {{SVGElement('mask')}} element's alpha values should be used; the pixels of the masked object match the opaqueness of the mask areas, without regard to the luminance of the colors of the mask.
+- `luminance`
+  - : This value indicates that the {{SVGElement('mask')}} element's luminance values should be used; the opaquess of the masked object depend on the opacity and lightness of the mask. The opacity of a `luminance` mask is determined by the `R`, `G`, `B`, and `A` channels of the mask, using the equation `((0.2125 * R) + (0.7154 * G) + (0.0721 * B)) * A`.
```
hamishwillee, line 85:
```suggestion
  - : This value indicates that the {{SVGElement('mask')}} element's luminance values should be used; the opaquess of the masked object depends on the opacity and lightness of the mask. The opacity of a `luminance` mask is determined by the `R`, `G`, `B`, and `A` channels of the mask, using the equation `((0.2125 * R) + (0.7154 * G) + (0.0721 * B)) * A`.
```

files/en-us/web/svg/reference/attribute/mask-type/index.md
```diff
@@ -0,0 +1,98 @@
+---
+title: mask-type
+slug: Web/SVG/Reference/Attribute/mask-type
+page-type: svg-attribute
+browser-compat: svg.elements.mask.mask-type
```
hamishwillee, line 5: Is this BCD entry in the works? In particular, I'm interested in the spec link, because I wasn't aware that this would point at the CSS spec.

files/en-us/web/svg/reference/attribute/mask-type/index.md
```diff
@@ -0,0 +1,98 @@
...
+sidebar: svgref
+---
+
+The **`mask-type`** attribute indicates which mask mode, _alpha_ or _luminance_, to use for the contents of the {{ SVGElement("mask") }} element when masking.
```
hamishwillee, line 9: This is interesting:

> "SVG presentational attributes have the identical definition to a CSS property of the same name."

Do we define what those are anywhere? Would be great to make that statement and then change the entry point to something like this:

```suggestion
The **`mask-type`** [SVG presentational attribute](LINK) indicates which mask mode, _alpha_ or _luminance_, to use for the contents of the {{ SVGElement("mask") }} element when masking.
```

files/en-us/web/svg/reference/attribute/mask-type/index.md
estelle: The property is not in the SVG spec. which is why i confirmed it with speccies

files/en-us/web/svg/reference/attribute/mask-type/index.md
estelle: project for a different day.

files/en-us/web/svg/reference/attribute/mask-type/index.md
hamishwillee: Yes, but that doesn't stop us putting the CSS spec into the SVG property docs, right?

hamishwillee (approved):
Thanks @estelle - we should address https://github.com/mdn/content/pull/39318/files#r2071096784 but that's a separate issue. If you're not planning on doing that let me know and I will.

PR #39319, opened 2025-04-28
update to mask-clip

estelle:
removed prefixed values.
added two values to the example.
added a description.


bsmth (approved):
LGTM thanks, one lint fix by the bot pending, but otherwise good to go!

PR #39326, opened 2025-04-29
Clarify unexpected bahaviour of using layered `@import` and `@layer` statement together

mahmoudmoravej:
### Description

The following 
```cs
@layer A, B, C;

@import url('/blob.css') B;
```

will result in `B, A, C` order **NOT** the defined one. 
The reason behind it is that `@import` runs sooner and its layered defined in layers stack. The the `@layer` runs and it predefine the layers in the stack if they are not already defined. 

While CSS doesn't raise any error or warning, this could end up in unexpected behaviour. 

### Motivation

It took me 2 days to understand the issue!

### DIFF
--- a/files/en-us/web/css/@layer/index.md
+++ b/files/en-us/web/css/@layer/index.md
@@ -121,8 +121,8 @@ Another way to create a cascade layer is by using {{cssxref("@import")}}. In thi
 @import "theme.css" layer(utilities);
 ```
 
-> [!IMPORTANT]
-> Using `@layer` to declare the layers order while using layered `@import`s in a same file could result an unexpected bahaviour. `@import` runs sooner and sets its layer before the `@layer` statement and makes its set order ineffective. As the result, the imported layer has the least order among other defined layers in the `@layer` statement. It is a good practice to keep them separated. 
+> [!NOTE]
+> Using `@layer` to declare the layers order while using layered `@import` rules in the same file could result in unexpected behavior. `@import` runs sooner and sets its layer before the `@layer` statement, making the latter's set order ineffective. As a result, the imported layer has the least order among other defined layers in the `@layer` statement. It is a good practice to keep them separated.
 
 ### Nesting layers
 
--- a/files/en-us/web/css/@layer/index.md
+++ b/files/en-us/web/css/@layer/index.md
@@ -115,7 +115,7 @@ The third way is to create an unnamed layer using a `@layer` block at-rule witho
 
 This creates an _anonymous cascade layer_. This layer functions in the same way as named layers; however, rules cannot be assigned to it later. The order of precedence for anonymous layers is the order in which layers are declared, named or not, and lower than the styles declared outside of a layer.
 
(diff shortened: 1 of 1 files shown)


mahmoudmoravej (commented):
remove space

files/en-us/web/css/@layer/index.md
```diff
@@ -121,6 +121,9 @@ Another way to create a cascade layer is by using {{cssxref("@import")}}. In thi
 @import "theme.css" layer(utilities);
 ```
 
+> [!IMPORTANT]
+> Using `@layer` to declare the layers order while using layered `@import`s in a same file could result an unexpected bahaviour. `@import` runs sooner and sets its layer before the `@layer` statement and makes its set order ineffective. As the result, the imported layer has the least order among other defined layers in the `@layer` statement. It is a good practice to keep them separated.
```
Josh-Cena, line 125:
```suggestion
> [!NOTE]
> Using `@layer` to declare the layers order while using layered `@import` rules in the same file could result in unexpected behavior. `@import` runs sooner and sets its layer before the `@layer` statement, making the latter's set order ineffective. As a result, the imported layer has the least order among other defined layers in the `@layer` statement. It is a good practice to keep them separated.
```

Just drive-by typo fixes. Not constituting a review.

files/en-us/web/css/@layer/index.md
```diff
@@ -121,6 +121,9 @@ Another way to create a cascade layer is by using {{cssxref("@import")}}. In thi
 @import "theme.css" layer(utilities);
 ```
 
+> [!NOTE]
+> Using `@layer` to declare the layers order while using layered `@import` rules in the same file could result in unexpected behavior. `@import` runs sooner and sets its layer before the `@layer` statement, making the latter's set order ineffective. As a result, the imported layer has the least order among other defined layers in the `@layer` statement. It is a good practice to keep them separated.
```
estelle, line 125: This is super useful and important to know. 
 I think it may be a better fit for the `@import` page, which currently reads "An @import rule must be defined at the top of the stylesheet, before any other at-rule (except [@charset](https://developer.mozilla.org/en-US/docs/Web/CSS/@charset) and [@layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer)) and style declarations, or it will be ignored." 

I am not sure if this is a feature or a bug. The spec reads: "The layer is added to the [layer order](https://drafts.csswg.org/css-cascade-5/#layer-ordering) even if the import fails to load the stylesheet, but is subject to any [import conditions](https://drafts.csswg.org/css-cascade-5/#import-conditions) ..." but doesn't say anything about the import layer order.

That said, i think we can make this more concise.
how about 
 ```suggestion
> While `@charset` and `@layer` declarations are the only declarations that can precede an `@import` declaration, layers created via an `@import` declaration will precede any layers declared in a `@layer` declaration."
```
This is more concise, but I am not sure it's not a bug and actually intentional.

mahmoudmoravej: Oh, thank you @estelle. I read the draft and I actually found [the example 31](https://drafts.csswg.org/css-cascade-5/#example-aa4a55a5) clearly says that it is possible. So I assume there should be a bug in the Browser or tooling. So I close this issue. Thanks again!

estelle: Please consider filing a bug report with the browser if the layer order is incorrect per the spec

PR #39331, opened 2025-04-30
Fix GFM alert syntax, add markdownlint check

Josh-Cena:
Without a line break after `[!NOTE]`, it's rendered weirdly without spaces around HTML tags.


bsmth (approved):
LGTM, thanks 👍🏻

PR #38359, opened 2025-02-27
Addressing issue #38325

low-perry:
### Description

Updated the definition of the `returnValue` property, to better align with it's function.
Updated the example of `returnValue` usage for better clarity

### Motivation

I made these changes to help readers better understand the `returnValue` property of the `<dialog>` element, and provided an example that demonstrates in a clearer way how this property is used.

### Related issues and pull requests

"Fixes #38325"

PR #39336, opened 2025-04-30
rm reference to undefined function `openCheck()`

jezmck:
### Description
Remove reference to undefined function `openCheck()`.

### Motivation
Prevent confusion.

### DIFF
--- a/files/en-us/web/api/htmldialogelement/returnvalue/index.md
+++ b/files/en-us/web/api/htmldialogelement/returnvalue/index.md
@@ -66,7 +66,6 @@ Closing the dialog with the <kbd>Esc</kbd> key does not set the `returnValue`.
   }
 
   dialog.addEventListener("close", () => {
-    openCheck(dialog);
     handleUserInput(dialog.returnValue);
   });
 </script>


wbamberg (approved):
👍 thank you! Looks like the function was removed in https://github.com/mdn/content/pull/38359#pullrequestreview-2650419581 but the call wasn't.

wbamberg:
That said I don't really understand this example. Why is it calling `handleUserInput` twice when it is closed, once directly in `closeDialog`, and then again in the `close` event handler?

wbamberg:
> @wbamberg I wrote that function. I agree with you that it is redundant. It can safely be removed from the `closeDialog` function.

Oh, thanks for the response @low-perry ! I also wondered why `handleUserInput()` gets called again when the dialog is opened. The first time the dialog is opened, this is always going to display "There was no return value", because the return value hasn't been set. On any subsequent occasion, it will update the status to the same thing it already was, because we updated status to reflect return value the last time the dialog was closed.

low-perry:
You're right, @wbamberg. When I first tackled this code, I admit I didn't look too closely at the openDialog part that was already there and just kind of took it as given. Appreciate you pointing out how that part behaves as well!
I think I should remove all the redundancies from that example.

wbamberg:
Thanks for the update @low-perry . I'll make a PR :).

PR #39338, opened 2025-04-30
mask-composite property

estelle:


bsmth:
Thanks, Estelle. I'm guessing this is still work in progress? Do you want to flip it to draft PR until it's ready or should I have a look already?

estelle:
> Thanks, Estelle. I'm guessing this is still work in progress? Do you want to flip it to draft PR until it's ready or should I have a look already?

Hi @bsmth It is out of draft

files/en-us/web/css/mask-composite/index.md
```diff
@@ -67,11 +85,211 @@ For the composition the current mask layer is referred to as _source_, while all
...
+</table>
+```
+
+```html
+<img
+  src="https://mdn.github.io/shared-assets/images/examples/progress-pride-flag.jpg"
+  alt="Pride flag" />
+```
+
+And an SVG with 4 masks; an alpha heart and circle and a luminance heart and circle. The heart masks are defined using solid colors. The circle masks are created using semi-transparent white and black {{SVGAttr("stroke")}} and {{SVGAttr("fill")}} colors.
```
estelle, line 183: the alpha / luminance is something they may want to consider, but there are no "rules" or "issues", they just might be like "WTF is this happening?" so want to include an example but there's nothing really to say that isn't said in the description. people don't really realize that luminence exists, and the effect can be profound, but basically it's not an "exception" (see intersect i think...)

files/en-us/web/css/mask-composite/index.md
```diff
@@ -26,20 +26,28 @@ mask-composite: revert-layer;
 mask-composite: unset;
 ```
 
-One or more of the keyword values listed below, separated by commas.
-
 ### Values
 
-For the composition the current mask layer is referred to as _source_, while all layers below it are referred to as _destination_.
+The property accepts a comma-separated list of `<compositing-operator>` keyword value, each representing a Porter-Duff compositing operator which defines the compositing operation used on the current mask layer with the mask layers below it, including:
```
bsmth, line 31:
```suggestion
The property accepts a comma-separated list of `<compositing-operator>` keyword values, each representing a Porter-Duff compositing operator which defines the compositing operation used on the current mask layer with the mask layers below it, including:
```

files/en-us/web/css/mask-composite/index.md
```diff
@@ -67,11 +85,211 @@ For the composition the current mask layer is referred to as _source_, while all
...
+mask.luminance {
+  mask-type: luminance;
+}
+
+mask.alpha {
+  mask-type: alpha;
+}
+```
+
+We then apply the heart and circle masks as the comma-separated {{cssxref("mask-image")}} property values . These are applied to each {{htmlelement("img")}} element, with all the images in a row getting the same masks.
```
bsmth, line 236:
```suggestion
We then apply the heart and circle masks as the comma-separated {{cssxref("mask-image")}} property values. These are applied to each {{htmlelement("img")}} element, with all the images in a row getting the same masks.
```

files/en-us/web/css/mask-composite/index.md
bsmth: Looks interesting, thanks for adding

bsmth (approved):
Thanks, Estelle. A couple of suggestions for you to look at, but +1 for when you've had a look 👀

Issue #38860, opened 2025-03-30
SVG: Restructuring to newer format


yashrajbharti:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/SVGPreserveAspectRatio

### What specific section or headline is this issue about?

Switching to newer content format 

### What information was incorrect, unhelpful, or incomplete?

- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedNumber
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedNumberList
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedPreserveAspectRatio
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedRect
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedTransformList
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGComponentTransferFunctionElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGFEBlendElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGFEColorMatrixElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGFECompositeElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGFEConvolveMatrixElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGFEDisplacementMapElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGFEMorphologyElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGFETurbulenceElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGGradientElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGPreserveAspectRatio
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGTextContentElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGTextPathElement
- [x] https://developer.mozilla.org/en-US/docs/Web/API/SVGTransformList

### What did you expect to see?

Good example to follow:
https://developer.mozilla.org/en-US/docs/Web/API/SVGLength

hamishwillee:
[...]

PR #39345, opened 2025-05-01
Reformat SVGAnimatedRect

yashrajbharti:
Part of #38860

PR #39352, opened 2025-05-01
Reformat SVGTextContentElement

yashrajbharti:
Part of #38860

### DIFF
--- a/files/en-us/web/api/svgtextcontentelement/index.md
+++ b/files/en-us/web/api/svgtextcontentelement/index.md
@@ -20,7 +20,9 @@ _This interface also inherits properties from its parent, {{domxref("SVGGraphics
 - {{domxref("SVGTextContentElement.lengthAdjust")}} {{ReadOnlyInline}}
   - : An {{domxref("SVGAnimatedEnumeration")}} reflecting the {{SVGAttr("lengthAdjust")}} attribute of the given element. The numeric type values represent one of the constant values above.
 
-<table class="standard-table">
+### Constants
+
+<table>
   <tbody>
     <tr>
       <td>Constant</td>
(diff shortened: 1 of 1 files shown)


wbamberg:
With this and the other PRs, I'm not sure what the newer content format referred to is or how or where it was agreed. https://github.com/mdn/content/issues/38860#issuecomment-2764757816 says:

> FWIW All of this should start by updating and agreeing the templates in https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types#page_templates

...has this been done?

Josh-Cena (approved):

Josh-Cena:
@wbamberg: the newer format is the format in the page template. The old format, which you can see has got removed, is the top-level H2 heading called "constants" (which shouldn't exist; it should be called "static properties"). Even worse are pages like https://github.com/mdn/content/pull/39345.

wbamberg:
In https://github.com/mdn/content/issues/38860, there's a reference to https://developer.mozilla.org/en-US/docs/Web/API/SVGLength as a "good example to follow", but I don't think that page is at all a good example to follow (it just dumps the table at the end of "instance properties", with no explanation), and it doesn't match the format in this PR. 

> it should be called "static properties"

But this PR doesn't use "static properties", it has an H3 "constants" under "instance properties".

And as I said above, @hamishwillee asked to update the template in advance of doing this work, but [the template hasn't been updated in this regard](https://github.com/mdn/content/commits/main/files/en-us/mdn/writing_guidelines/page_structures/page_types/api_reference_page_template/index.md). So is Hamish confused about this as well?

Josh-Cena:
> But this PR doesn't use "static properties", it has an H3 "constants" under "instance properties".

Yeah, I updated them together in https://github.com/mdn/content/pull/39349. The current content should reflect the page templates well.

PR #39360, opened 2025-05-01
Update a few TC39 proposal links

Josh-Cena:
Update TC39 links that are either already standard or changed significantly


estelle (approved):
Thanks! 🎉

PR #39368, opened 2025-05-02
Fix typo in "Range: setEnd() method"

gurdiga:
### Description

Fixing what seems like a typo.

### Motivation

It was confusing to me, and is probably confusing to others.


Josh-Cena (approved):
Nice catch, thanks

PR #39374, opened 2025-05-03
Use reduce() method to get the total size of Uint8Arrays

def00111:
### Description

Use reduce() method to get the total size of Uint8Arrays instead of using `for...of` loop.

### Motivation

Replace 4 lines of code with 1 line.


dotproto (approved):

rebloor:
Thanks @def00111

PR #39375, opened 2025-05-03
Update 1 and 4 example from /docs/Mozilla/Add-ons/WebExtensions/API/w…

def00111:
…ebRequest/StreamFilter/ondata

### Motivation

In the first example the "*" is not needed. In the 4 example just store the string directly instead of creating an array and storing the string there.


willdurand:
> In the first example the "*" is not needed. In the 4 example just store the string directly instead of creating an array and storing the string there.

That doesn't look particularly meaningful, or am I missing something?

def00111:
@willdurand Can you look at the patch again? Is this change worth?

bsmth:
Thanks all, I agree that the changes are correct, but so is the original, so it's not worth the churn. Going to close for now. If you think I've made a mistake, feel free to let me know and we can take a look again. Thanks!

PR #39377, opened 2025-05-03
Use the new blob.bytes() method

def00111:
### Motivation

Using the newer blob.bytes() method.


dotproto (commented):
[`TextDecoder.decode()`](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode) works with either an ArrayBuffer or a TypedArray, so it's not clear to me what benefit this change provides. Given the age of this PR, I'm inclined to close it out as is.

PR #39388, opened 2025-05-04
Minor edits to the "Third-party cookies" section

DanKaplanSES:
- I think there was some issues with pluralization of the word "Cookie" at the end of this section, so I attempted to reword it.
- I ended up changing more than I originally intended
- I did not intend to change any of the original meaning. If I did, that was a mistake on my part.

### DIFF
--- a/files/en-us/web/http/guides/cors/index.md
+++ b/files/en-us/web/http/guides/cors/index.md
@@ -360,7 +360,12 @@ Also note that any `Set-Cookie` response header in a response would not set a co
 
 Note that cookies set in CORS responses are subject to normal third-party cookie policies. In the example above, the page is loaded from `foo.example` but the `Set-Cookie` header in the response is sent by `bar.other`, and would thus not be saved if the user's browser is configured to reject all third-party cookies.
 
-Third-party cookie policies (e.g., the [SameSite](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute) may suppress certain cookies from being sent to the server. The enforced cookie policy may therefore nullify the capability described in this chapter, effectively preventing you from making credentialed requests whatsoever.
+Cookies set in CORS requests and responses are subject to normal third-party cookie policies.
+
+Third-party cookie policies may prevent third party cookies being sent in requests, effectively stopping a site from making credentialed requests even if permitted by the third party server (using `Access-Control-Allow-Credentials`).
+The default policy differs between browsers, but may be set using the [SameSite](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute.
+
+Even if credentialed requests are allowed, a browser may be configured to reject all third-party cookies in responses.
 
 ## The HTTP response headers
 
--- a/files/en-us/web/http/guides/cors/index.md
+++ b/files/en-us/web/http/guides/cors/index.md
@@ -360,9 +360,7 @@ Also note that any `Set-Cookie` response header in a response would not set a co
 
 Note that cookies set in CORS responses are subject to normal third-party cookie policies. In the example above, the page is loaded from `foo.example` but the `Set-Cookie` header in the response is sent by `bar.other`, and would thus not be saved if the user's browser is configured to reject all third-party cookies.
 
-Cookie in the request may also be suppressed in normal third-party cookie policies. The enforced cookie policy may therefore nullify the capability described in this chapter, effectively preventing you from making credentialed requests whatsoever.
-
-Cookie policy around the [SameSite](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute would apply.
+Third-party cookie policies (e.g., the [SameSite](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute) may suppress certain cookies from being sent to the server. The enforced cookie policy may therefore nullify the capability described in this chapter, effectively preventing you from making credentialed requests whatsoever.
 
 ## The HTTP response headers
 


files/en-us/web/http/guides/cors/index.md
```diff
@@ -364,9 +364,7 @@ Also note that any `Set-Cookie` response header in a response would not set a co
 
 Note that cookies set in CORS responses are subject to normal third-party cookie policies. In the example above, the page is loaded from `foo.example` but the `Cookie` header in the response is sent by `bar.other`, and would thus not be saved if the user's browser is configured to reject all third-party cookies.
 
-Cookie in the request may also be suppressed in normal third-party cookie policies. The enforced cookie policy may therefore nullify the capability described in this chapter, effectively preventing you from making credentialed requests whatsoever.
-
-Cookie policy around the [SameSite](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute would apply.
+Third-party cookie policies (e.g., the [SameSite](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute) may suppress certain cookies from being sent to the server. The enforced cookie policy may therefore nullify the capability described in this chapter, effectively preventing you from making credentialed requests whatsoever.
```
hamishwillee, line 367: I prefer your version, but it does kind of change the "flow". How about this:

```suggestion
Cookies set in CORS requests and responses are subject to normal third-party cookie policies.

Third-party cookie policies may prevent third party cookies being sent in requests, effectively stopping a site from making credentialed requests even if permitted by the third party server (using `Access-Control-Allow-Credentials`).
The default policy differs between browsers, but may be set using the [SameSite](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute.

Even if credentialed requests are allowed, a browser may be configured to reject reject all third-party cookies in responses.
```

bsmth:
Hi @DanKaplanSES - there's some review comments above for you. Do you want to come back to have a look at this again? Thanks!

bsmth:
@hamishwillee do you feel like applying the suggestions and co-authoring this one to bring it over the line? 🙏🏻

files/en-us/web/http/guides/cors/index.md
DanKaplanSES: Hello @hamishwillee  Sorry it took me so long to reply. 

1. Could you help me better understand the motivation behind your suggestion? Was my original too terse, did it leave out important details (e.g., `Access-Control-Allow-Credentials` and server side cookie policies), etc.? You mentioned "flow," but I am not sure what you meant by that in _this_ context.
1. Also, when you say, "How about this (for the whole section)," do you mean the paragraph before my edit would be removed too? I'm referring to this one:
    > Note that cookies set in CORS responses are subject to normal third-party cookie policies. In the example above, the page is loaded from `foo.example` but the `Cookie` header in the response is sent by `bar.other`, and would thus not be saved if the user's browser is configured to reject all third-party cookies.
    
    If so, I see the improvement of your suggestion, but I don't understand the difference between your first two paragraphs.

---

About the original paragraph, I'm more confused than ever. Is it supposed to be about 3rd party Cookies sent in the preflight request, after preflight request, or both? 

Here's the type of scenario I _think_ it's talking about: The server responds to a preflight request with `Access-Control-Allow-Headers: Cookie`, but the client has a `SameSite=Strict` value for the Cookie the site uses for authentication. If the server is on a different domain, the client won't send the cookie in the real request even though the preflight allows it.

hamishwillee: @DanKaplanSES I'll look at this next week - been OOO for three weeks and am a bit jetlagged. Not forgotten.

DanKaplanSES: NP @hamishwillee . I appreciate it

hamishwillee: CORS makes me jetlagged even when I'm not.

1. I am not sure what I meant by the flaws in your original flow - which indicates I was wrong. 
  At a guess I didn't like the original flow because it repeats "_ cookies set in CORS [requests | responses] are subject to normal third-party cookie policies_" and your fix didn't improve this flow. But who knows - this was months ago :-)
2. > Note that cookies set in CORS responses are subject to normal third-party cookie policies. In the example above, the page is loaded from `foo.example` but the `Set-Cookie` header in the response is sent by `bar.other`, and would thus not be saved if the user's browser is configured to reject all third-party cookies.

   Yes, I am proposing that if my version were chosen then this sentence is not needed.

   >If so, I see the improvement of your suggestion, but I don't understand the difference between your first two paragraphs.
   
   - "Cookies set in CORS requests and responses are subject to normal third-party cookie policies.". This summarizes that third party request and responses are subject the normal policies
   - "Third-party cookie policies may prevent third party cookies being sent in request ..." - this then covers what those policies mean for requests. The next paragraph covers what those policies mean for responses.
   
   Does that show the difference?



--- 

> About the original paragraph, I'm more confused than ever. Is it supposed to be about 3rd party Cookies sent in the preflight request, after preflight request, or both?

Which original para? This one? : "Cookie in the request may also be suppressed in normal third-party cookie policies. The enforced cookie policy may therefore nullify the capability described in this chapter, effectively preventing you from making credentialed requests whatsoever." 

The answer is **after** the preflight request because credentials aren't ever send in the preflight request. But it doesn't matter. The paragraph is just saying that you might indicate that you want to send credentials, for example by using `credentials: "include" }` in your request, but that will be ignored if blocked by the normal third party rules (Same-Site).

> Here's the type of scenario I think it's talking about: The server responds to a preflight request with Access-Control-Allow-Headers: Cookie, but the client has a SameSite=Strict value for the Cookie the site uses for authentication. If the server is on a different domain, the client won't send the cookie in the real request even though the preflight allows it.

Correct.

Again, your version is better than what was there. I just kind of prefer my structure. Might be worth adding a note "e.g. by setting  `credentials: "include" }` " in there too to make it clear how we are expecting credentials to be sent, but I think overkill.

DanKaplanSES: > Which original para? This one?

Yep, that one. 

> Correct.

Okay, I understand the original paragraph now, thanks!

Mentioning [credentials: "include"](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials) is confusing me a bit. Isn't that a Fetch API concept? I may be mistaken, but aren't there ways to send HP requests that don't use the Fetch API?

Regardless, I'd like to give this edit another shot to see what you think. I'll do that in my next reply.

DanKaplanSES: @hamishwillee Actually, I think I'd like to increase the scope of my edit beyond this part of the page. I'm happy with your suggestion, but I might create a PR in the future that changes more, including this paragraph. 

In the meantime, I'll accept your suggestion. Thanks!

hamishwillee: Thanks @DanKaplanSES . I'll merge this and look forward to your next PR :-)

hamishwillee (approved):

PR #39396, opened 2025-05-05
fix: Move lang attributes to bdi elements

### LINKED ISSUE #39252: Lang attribute bad example
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang

Code block of the menu

Reading [this document from W3](https://www.w3.org/International/questions/qa-html-language-declarations) they provide an example on how to handle lang attributes combined with title attributes, especially when their languages are mismatching. 

The idea is that you only add the lang attribute on elements actually in that language. Meanwhile the example provided here (and probably the implementation) has title attributes in english WITHIN a non-english lang tag to essentially tag the bdi element. So, the lang attributes need to be moved onto the bdi tags.

No contradicting information

_No response_

_No response_

enumura1:
### Description
I have corrected the placement of the lang attribute in the language menu example. The lang attribute has been moved from the li element to the bdi element, which actually contains the text in that language.

### Motivation
[Declaring language in HTML](https://www.w3.org/International/questions/qa-html-language-declarations) contains examples of how to handle lang attributes and title attributes when their languages don't match. Following this reference, I updated the code to apply the lang attribute to elements that actually contain text in that language.

### Additional details
[W3C - Declaring language in HTML](https://www.w3.org/International/questions/qa-html-language-declarations#details)

Fixes #39252

### DIFF
--- a/files/en-us/web/html/reference/global_attributes/lang/index.md
+++ b/files/en-us/web/html/reference/global_attributes/lang/index.md
@@ -86,74 +86,74 @@ For example, the language menu on this site (MDN) includes a **`lang`** attribut
     class="dropdown-menu-items right show"
     aria-expanded="true"
     role="menu">
-    <li lang="ca" role="menuitem">
+    <li role="menuitem">
       <a
         href="/ca/docs/Web/HTML/Reference/Global_attributes/lang"
         title="Catalan">
-        <bdi>Català</bdi>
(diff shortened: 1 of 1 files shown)


estelle (approved):
Thank you. Also, congratulations on your first merged MDN content PR. Welcome to the team!

PR #39397, opened 2025-05-05
Unify console output format

douglasimcabral:
### Description

I believe the detail of the semicolon and quotation marks went unnoticed, as it are very subtle, which is why I am submitting this pull request to the "Number.isInteger()" documentation.

### Motivation

Improve documentation standardization.


Josh-Cena (approved):
Sure, thanks

PR #39400, opened 2025-05-05
fix(learn): use `inherit` for skills solution in CSS cascade

arlgnv:
### Description

On 65th line `inherit` keyword is being mentioned, so, I guess, that has to be used here


bsmth (approved):
That's a good fix, thanks. They're both correct in terms of the goal of the challenge, but the prose has an explanation that expects `inherit`, so let's add this.

Well done! 👍🏻

PR #39402, opened 2025-05-05
Update default useragent stylesheet for view-transition-*

skyclouds2001:
### Additional details

https://drafts.csswg.org/css-view-transitions/#ua-styles


bsmth (approved):
LGTM, thanks a lot!

PR #39403, opened 2025-05-05
Add missing text

AdedejiAdelanwa:
### Description

Added  missing text 'an error' under the import explanation

### Motivation

It makes it easier for a beginner reading through the docs not to be confused as to the idea being expressed when they don't have to fill the blanks in their head


bsmth (approved):
Thanks a lot. Just to clarify: it's idiomatic to write "will throw if…" because it's implied that throwing is about exceptions / errors (see language here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#return_value). That means, the original is not incorrect, but the additions may make the document more accessible, so I'm happy to approve.

Thanks for the contribution, well done and welcome aboard!

PR #29746, opened 2023-10-20
Update promise usage example for clarity

collinsnji:
### Description
Some edits to the promises examples to highlight the differences in the bad code example vs the good code example. PR addresses issue: https://github.com/mdn/content/issues/29743. Here's a summary of the changes

* Clarifies the difference between the good promise examples vs the bad example related to return statements in the arrow function
* Minor grammar and punctuation changes.
### Motivation
Looking at the code examples for the promises, it is not exactly clear what makes a good promise call different from a bad promise call when considering the `return` statement. To avoid floating promises and race conditions, promises should always be returned either implicitly or explicitly, and this PR changes highlights that need. Changes will help the reader know that promises can have unexpected behavior if not properly returned.

### Additional details
N/A

### Related issues and pull requests
Fixes: #29743
Fixes #29742 
Fixes https://github.com/mdn/content/issues/28236

PR #39407, opened 2025-05-05
fix: Remove "Initial" log from promise example output

hyungjikim:
### Description

This PR removes "Initial" from the example output in the promise chaining section to better reflect the current code and maintain clarity.

### Motivation

The example output included 'Initial', but the current code no longer logs it.
I considered adding console.log('Initial'), but decided to remove it from the output instead, since:

- The example focuses on chaining after a .catch(), not synchronous logs.
- Adding 'Initial' back could distract from the promise flow being demonstrated.

This change keeps the output aligned with the current code and its intent.

### Additional details

Attaching previous update made in [#29746](https://github.com/mdn/content/pull/29746), where the console.log('Initial') statement was removed from the code example.

### DIFF
--- a/files/en-us/web/javascript/guide/using_promises/index.md
+++ b/files/en-us/web/javascript/guide/using_promises/index.md
@@ -313,7 +313,6 @@ doSomething()
 This will output the following text:
 
 ```plain
-Initial
 Do that
 Do this, no matter what happened before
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
Great catch, thank you!

PR #39410, opened 2025-05-05
Merge Introduction_to_HTML5_Game_Development into Games/Tools

Josh-Cena:
This page is unreachable via any page, unreachable via sidebar, and still has "HTML5" in its title. It's just a list of Web APIs, so we can merge it into the other article that talks about tools for game dev.


chrisdavidmills (approved):
@Josh-Cena makes sense to me, thanks.

PR #39411, opened 2025-05-05
Add content covering the limitations of <selectedcontent>

philipwalton:
### Description

This PR updates the `<selectedcontent>` reference document to:
1. Include more details about the limitations of <selectedcontent> cloning, and
2. Add a note emphasizing that sites built using JavaScript frameworks may prefer not to use `<selectedcontent>` and instead rely on the framework's existing UI synchronization feature.

### Motivation

After seeing [this Angular bug](https://github.com/angular/angular/issues/60636) related to `<selectedcontent>` as well as testing the behavior of `<selectedcontent>` in several other frameworks and discovering issues, I believe it's important to update the MDN docs to clarify both the use cases for `<selectedcontent>` as well as its limitations. This should help ensure developers aren't confused by issues they run into when trying to use this new element.

FYI, I've discussed this recommendation with folks on the Chrome team working on the customizable `<select>` feature, and they're comfortable with this recommendation.

### DIFF
--- a/files/en-us/web/html/reference/elements/selectedcontent/index.md
+++ b/files/en-us/web/html/reference/elements/selectedcontent/index.md
@@ -39,7 +39,7 @@ When creating a [Customizable select element](/en-US/docs/Learn_web_development/
 Any subsequent `<select>` content will be included in the drop-down picker.
 
 Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `<option>`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
-Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
+Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element, and must be manually updated by the developer.
 
 > [!NOTE]
 > The `<selectedcontent>` element's built-in synchronization features that make it ideal for websites built with minimal or no JavaScript. However, its use is not required when building customizable selects. Developers using JavaScript frameworks may prefer to use the UI synchronization features provided by their framework instead of using `<selectedcontent>`.
--- a/files/en-us/web/html/reference/elements/selectedcontent/index.md
+++ b/files/en-us/web/html/reference/elements/selectedcontent/index.md
@@ -38,7 +38,8 @@ When creating a [Customizable select element](/en-US/docs/Learn_web_development/
 
 Any subsequent `<select>` content will be included in the drop-down picker.
 
-Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `<option>`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}. Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
+Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `<option>`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
+Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
 
 > [!NOTE]
 > The `<selectedcontent>` element's built-in synchronization features that make it ideal for websites built with minimal or no JavaScript. However, its use is not required when building customizable selects. Developers using JavaScript frameworks may prefer to use the UI synchronization features provided by their framework instead of using `<selectedcontent>`.
--- a/files/en-us/web/html/reference/elements/selectedcontent/index.md
+++ b/files/en-us/web/html/reference/elements/selectedcontent/index.md
@@ -38,7 +38,10 @@ When creating a [Customizable select element](/en-US/docs/Learn_web_development/
 
 Any subsequent `<select>` content will be included in the drop-down picker.
 
-Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected <code>option</code>, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
+Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected <code>option</code>, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}. Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
+
+> [!NOTE]
+> The `<selectedcontent>` element's built-in synchronization features that make it ideal for websites built with minimal or no JavaScript. However, its use is not required when building customizable selects. Developers using JavaScript frameworks may prefer to use the UI synchronization features provided by their framework instead of using `<selectedcontent>`.
 
 ## Styling with CSS
 
--- a/files/en-us/web/html/reference/elements/selectedcontent/index.md
+++ b/files/en-us/web/html/reference/elements/selectedcontent/index.md
@@ -38,7 +38,7 @@ When creating a [Customizable select element](/en-US/docs/Learn_web_development/
 
 Any subsequent `<select>` content will be included in the drop-down picker.
 
-Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected <code>option</code>, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}. Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
+Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `<option>`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}. Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
 
 > [!NOTE]
 > The `<selectedcontent>` element's built-in synchronization features that make it ideal for websites built with minimal or no JavaScript. However, its use is not required when building customizable selects. Developers using JavaScript frameworks may prefer to use the UI synchronization features provided by their framework instead of using `<selectedcontent>`.
--- a/files/en-us/web/html/reference/elements/selectedcontent/index.md
+++ b/files/en-us/web/html/reference/elements/selectedcontent/index.md
@@ -41,8 +41,8 @@ Any subsequent `<select>` content will be included in the drop-down picker.
 Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `<option>`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
 Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element, and must be manually updated by the developer.
 
-> [!NOTE]
-> The `<selectedcontent>` element's built-in synchronization features that make it ideal for websites built with minimal or no JavaScript. However, its use is not required when building customizable selects. Developers using JavaScript frameworks may prefer to use the UI synchronization features provided by their framework instead of using `<selectedcontent>`.
+> [!WARNING]
+> In particular, this may cause issues with sites that use popular front-end JavaScript frameworks where {{htmlelement("option")}} elements are dynamically updated after creation, as these updates will not be cloned to the `<selectedcontent>` element.
 
 ## Styling with CSS
 


files/en-us/web/html/reference/elements/selectedcontent/index.md
```diff
@@ -38,7 +38,10 @@ When creating a [Customizable select element](/en-US/docs/Learn_web_development/
...
+Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected <code>option</code>, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}. Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
+
+> [!NOTE]
+> The `<selectedcontent>` element's built-in synchronization features that make it ideal for websites built with minimal or no JavaScript. However, its use is not required when building customizable selects. Developers using JavaScript frameworks may prefer to use the UI synchronization features provided by their framework instead of using `<selectedcontent>`.
```
estelle, line 44:
```suggestion
```
Let's remove this.
My thought is, if` <selectedcontent>` is useful and has semantic meaning, it should be used even by frameworks. If it doesn't, it shouldn't be used by anyone. We definitely don't want to start making suggestions that markup produced by frameworks should ever differ.

files/en-us/web/html/reference/elements/selectedcontent/index.md
```diff
@@ -38,7 +38,10 @@ When creating a [Customizable select element](/en-US/docs/Learn_web_development/
 
 Any subsequent `<select>` content will be included in the drop-down picker.
 
-Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected <code>option</code>, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
+Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected <code>option</code>, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}. Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
```
estelle, line 41:
```suggestion
Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `option`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}. Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
```
or `<option>`?

estelle (commented):
I am not even sure <selectedcontent> should be included yet on MDN. This review is based on us keeping it for now

files/en-us/web/html/reference/elements/selectedcontent/index.md
philipwalton: Good catch. This was supposed to be `<option>`. I've pushed an update.

files/en-us/web/html/reference/elements/selectedcontent/index.md
philipwalton: While I understand this objection (I initially felt the same way), I believe this is one of those cases where the element being documented has a very specific use case, and the documentation should be clear about what what use case is, and when it makes sense to use this element versus some other mechanism to achieve the same result.

> We definitely don't want to start making suggestions that markup produced by frameworks should ever differ.

The suggestion here isn't that the markup produced by frameworks should differ, it's that people using a framework likely do not need to use this element (since it doesn't meet their use case).

The goal of this note is to signal to framework _users_ that it's perfectly okay not to use the `<selectedcontent>` element and instead rely on the framework's existing APIs to mirror content in multiple places in the DOM.

> My thought is, if `<selectedcontent>` is useful and has semantic meaning, it should be used even by frameworks.

The `<selectedcontent>` element has no semantic meaning (or ARIA role, or anything like that), and its only use is to enable authors to mirror the content of the selected `<option>` within the `<select>` button—without having to write any JavaScript code to do so.

This is a very useful feature for authors who don't want to write JavaScript. However, due to the element's limitations (that I mention in this PR), it may not work as expected for highly dynamic sites (which we're seeing in the wild based on the bug reports already coming from framework users). I think the documentation should warn author of that possibility.

files/en-us/web/html/reference/elements/selectedcontent/index.md
```diff
@@ -38,7 +38,10 @@ When creating a [Customizable select element](/en-US/docs/Learn_web_development/
 
 Any subsequent `<select>` content will be included in the drop-down picker.
 
-Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected <code>option</code>, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
+Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `<option>`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}. Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
```
hamishwillee, line 41: Just me catering to my own line break tastes.

```suggestion
Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `<option>`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
```

files/en-us/web/html/reference/elements/selectedcontent/index.md
```diff
@@ -38,7 +38,11 @@ When creating a [Customizable select element](/en-US/docs/Learn_web_development/
 
 Any subsequent `<select>` content will be included in the drop-down picker.
 
-Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected <code>option</code>, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
+Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `<option>`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
+Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element and must be manually updated by the developer.
```
hamishwillee, line 42:
```suggestion
Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element, and must be manually updated by the developer.
```

files/en-us/web/html/reference/elements/selectedcontent/index.md
hamishwillee: Thanks @philipwalton 

I tend to think the chrome team need a swat around the ears - they should push this whole thing through a proper specification process and work with angular to clearly address this case. Generally though I agree with @estelle - the tail (framework) should not wag the dog (Angular). Angular have already stated that they will look at fixing this in due course.

Anyway, my take on this is that I very much like the added paragraph, but this note is not acceptable.
It too general and after the issue is addressed it would live here forever with no trigger to remove.

I propose we are much more specific (if you and @estelle are both OK with it):

```suggestion

> [!NOTE]
> The `<selectedcontent>` may not work properly with JavaScript front-end frameworks that dynamically modify options after creation, such as Angular.
> For more information and workarounds see the issue [angular#60636: Angular breaks `<selectedcontent>`](https://github.com/angular/angular/issues/60636#top)
```

philipwalton: > I tend to think the chrome team need a swat around the ears

I would ask that we try to keep this conversation respectful and constructive, in accordance with the [Mozilla Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/
).

> they should ... work with angular to clearly address this case.

I've met with the engineering leads from both Chrome and Angular about this issue, and after lots of discussion, the conclusion we all came to was that developers building highly dynamic sites that may need to programmatically update select `<option>` values after page load, should probably not use the `<selectedcontent>` element, as it's not built for that use case.

> Angular have already stated that they will look at fixing this in due course.

To be clear, this is not a bug in Angular, this is a fundamental limitation of the `<selectedcontent>` element. While Angular may choose to work around the specific problem reported in the linked issue, they will not be able to address the larger issue without deviating from the platform behavior (which I'm not sure is a good idea). I've tested the use of `<selectedcontent>` in React, Vue, Svelte, Solid, and Angular, and all of them do not work as expected if the <option> value is dynamically updated.

I believe developers—particularly developers using frameworks—will get tripped up by this behavior, so it's important that the limitations and quirks of the element are clearly called out in its documentation.

> I propose we are much more specific (if you and @estelle are both OK with it):

I'm happy to consider alternative language; however, I don't think your suggestion addresses the core issue, which is that dynamic modifications to the selected `<option>` value do not get automatically cloned to the `<selectedcontent>` element. Your language also implies that this problem is specific to Angular, which is not true (it applies to any site that dynamically modifies `<option>` values, regardless of whether its using a framework or not).

For framework users it's much simpler and easier to just not use `<selectedcontent>` in the first place, which is why my original note suggested that. Like I said, if there is better language that makes the same point. I'm happy to consider other options.

hamishwillee: @philipwalton Apologies for offending you. I shouldn't have been flippant. A poor reaction to my surprise that the element isn't being modified to handle dynamic updates of the options.

> For framework users it's much simpler and easier to just not use `<selectedcontent>` in the first place.

So how do they fall back if the user disables javascript? 

I'll have a think about this. Perhaps @estelle will come up with some better words.

estelle:
> I don't think your suggestion addresses the core issue, which is that dynamic modifications to the selected <option> value do not get automatically cloned to the `<selectedcontent>` element. 

Is this a specification bug or an implementation bug?

files/en-us/web/html/reference/elements/selectedcontent/index.md
hamishwillee: Thinking about this a bit more, my specific problems with the note are:
- "synchronization features" is browser spec/blog jargon. I find it hard to parse this, and not actually necessary.
- "minimal javascript" - isn't very clear direction. This could be a problem for a website that only updates the options dynamically and does nothing else.
- This isn't in a public spec, so anyone looking at it will wonder at the reasoning and source of the advice.

I'd accept a suggestion that was very clear on the source of the recommendation, and very clear on the case that causes the problem. So maybe something like:

```suggestion

> [!NOTE]
> Chrome recommend that this element not be used with sites that use JavaScript to dynamically update `<options>` after creation, as these updates will not be cloned to the `<selectedcontent>` element.
> This may include sites that use popular front end JavaScript frameworks.
```

If that doesn't work for you, please suggest an alternative that addresses my concerns. Estelle may of course have other comments following your response to her queries.

hamishwillee: > "Developers using JavaScript frameworks may prefer to use the UI synchronization features provided by their framework instead of using <selectedcontent>."

@chrisdavidmills Please see my comment above about what I consider acceptable for this https://github.com/mdn/content/pull/39411#discussion_r2096632661

I wouldn't accept your suggestion for the reasons given:

- "synchronization features" is browser spec/blog jargon. I find it hard to parse this, and not actually necessary.
- This isn't in a public spec, so anyone looking at it will wonder at the reasoning and source of the advice.

chrisdavidmills: I don't necessarily agree that "synchronization features" is difficult jargon — it is pretty obvious to me what it means regarding how frameworks operate. But I agree don't think it is vital to include.

So, how about we delete the note, and then add:

- An extra sentence along the lines of "In particular, this may cause issues with sites that use popular front-end JavaScript frameworks where the included `<options>` are dynamically updated after creation. These updates will not be cloned to the `<selectedcontent>` element. The [specification is being updated](LINK TO SPEC FIX) to fix such issues".
- A note on the BCD for the feature that brief explains the issue and links to the spec fix.

hamishwillee: That works for me :-)

philipwalton: I do worry that without the extra visual call out of a note, this will be missed by developers trying to add this element to their existing framework-based codebase. But I'm less familiar with the norms of MDN, so happy to defer to you all of that one.

> The [specification is being updated](LINK TO SPEC FIX) to fix such issues

I'm not actually sure that it is being updated (cc: @josepharhar), so I'd suggest we remove that last part. My understanding is the behavior originally shipped in Chrome wouldn't have been a problem for frameworks, but then after concerns were raised that the Chrome behavior could result in lots of unnecessary cloning, the spec was changed to reflect the current behavior. I'm not aware of future changes being worked on.

> So how do they fall back if the user disables javascript?

@hamishwillee I don't think this is a problem because the options values could only ever get out of sync as a result of JavaScript code, so if a user disabled JavaScript then this either wouldn't be a problem, or the site wouldn't work at all. I don't think use (or non-use) of `<selectedcontent>` changes that, though admittedly `<selectedcontent>` is clearly a better choice if you want to build a site that doesn't require JavaScript to work.

philipwalton:
> Is this a specification bug or an implementation bug?

I think this was a specification choice (not bug) to avoid over-cloning. See my [other comment here](https://github.com/mdn/content/pull/39411#discussion_r2100573321) for more context.

files/en-us/web/html/reference/elements/selectedcontent/index.md
hamishwillee: @philipwalton That's my point. People want to build sites to fall back to when the site can work without JavaScript. Maybe that is never the case if you're using a front end framework?

Whether or not chrome planned this and don't consider it a bug, this design is way out of line with the normal expectations of developers. Would be great if Chrome could reconsider options to "fix" this for that reason. But in any case, I agree that this means a callout is warranted.

@chrisdavidmills so with your words, and this updated to a warning.

```suggestion

> [!WARNING]
> In particular, this may cause issues with sites that use popular front-end JavaScript frameworks where the included <options> are dynamically updated after creation. These updates will not be cloned to the <selectedcontent> element
```

files/en-us/web/html/reference/elements/selectedcontent/index.md
```diff
@@ -38,7 +38,11 @@ When creating a [Customizable select element](/en-US/docs/Learn_web_development/
 
 Any subsequent `<select>` content will be included in the drop-down picker.
 
-Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected <code>option</code>, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
+Whenever the `<select>` element's selected `<option>` switches from one option to another, the `<selectedcontent>` element's content is removed and replaced by a new cloned copy of the DOM structure of the newly selected `<option>`, which is created using {{domxref("Node.cloneNode", "cloneNode()")}}.
+Dynamic modifications to the selected `<option>` element's content made after the `<select>` element has been created are not automatically cloned to the `<selectedcontent>` element, and must be manually updated by the developer.
+
+> [!WARNING]
+> In particular, this may cause issues with sites that use popular front-end JavaScript frameworks where {{htmlelement("option")}} elements are dynamically updated after creation, as these updates will not be cloned to the `<selectedcontent>` element.
```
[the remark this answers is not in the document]
chrisdavidmills: This works for me.

estelle (approved):

files/en-us/web/html/reference/elements/selectedcontent/index.md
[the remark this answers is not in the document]
philipwalton: Works for me too, thanks all! (And sorry for the delayed reply.)

PR #39416, opened 2025-05-06
Docs: Clarify setListener source in insertAdjacentElement example

istvan-ujjmeszaros:
The `setListener(tempDiv);` calls in the static JavaScript examples for `Element.insertAdjacentElement()` could be confusing as the `setListener` function is not defined within the MDN page's visible snippets.

Its definition is located in the full interactive demo hosted on GitHub, which is linked from the "Examples" section.

This commit adds an inline comment, such as:
`// Definition in the linked GitHub demo`
next to these calls. This provides immediate context for readers, improving the standalone clarity of the code snippets on the MDN page without requiring them to navigate away to understand the `setListener` reference.


wbamberg (approved):
Thanks! I think we should include this complete demo in the page but I will do this in a follow-up.

PR #39419, opened 2025-05-06
More Error Messages

addsoupbase:
### Description
Just added more related error messages not covered 

### Additional details

Users searching error messages will be able to find the solution easier


Josh-Cena (approved):
Sure, thank you!

PR #39422, opened 2025-05-06
Add warning against using element IDs as global properties.

### LINKED ISSUE #39339: Documentation does not warn against using element IDs as global properties
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id

[Description](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id#description)

The documentation for the `id` attribute describes how "Elements with ID attributes are available as global properties", and gives an example of how it can be used with JavaScript, but there is no mention of how it might be bad practice to reference elements this way.

As answers on Stack Overflow have said (see references), this "feature" came to be because IE started it, and other browsers had to follow for compatibility. This functionality is part of the HTML specification, but even the specification advises against using it:

> As a general rule, relying on this will lead to brittle code. Which IDs end up mapping to this API can vary over time, as new features are added to the Web platform, for example. Instead of this, use `document.getElementById()` or `document.querySelector()`.

I would expect to see some sort of note that advises against using this feature, just as the specification does.

- [HTML Standard - 7.2.2.3 Named access on the Window object](https://html.spec.whatwg.org/multipage/nav-history-apis.html#named-access-on-the-window-object)

Stack Overflow:
- [Answer to "Do DOM tree elements with IDs become global properties?"](https://stackoverflow.com/a/3434388)
- [Answer to "Is there a spec that the id of elements should be made global variable?"](https://stackoverflow.com/a/6381508)

Documenting this sort of behavior without any warnings seems like a mistake to me. It concerns me that [this change](https://github.com/mdn/content/pull/25292) has remained here for two years.

### MDN metadata

<details>

* Folder: `en-us/web/html/reference/global_attributes/id`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/reference/global_attributes/id/index.md
* Last commit: https://github.com/mdn/content/commit/e9b6cd1b7fa8612257b72b2a85a96dd7d45c0200
* Document last modified: 2025-04-10T14:56:07.000Z

</details>

younisdev:
### Description

Add a warning about using element IDs as global properties on `window` object, and recommend best practices for safer DOM access.
### Motivation

Help the readers follow modern standard and best practices when selecting and manipulating element using ID.
### Additional details

None.

Fixes #39339

### DIFF
--- a/files/en-us/web/html/reference/global_attributes/id/index.md
+++ b/files/en-us/web/html/reference/global_attributes/id/index.md
@@ -49,7 +49,7 @@ const content = window.preamble.textContent;
 ```
 
 > [!WARNING]
-> While elements with `id` attributes are treated as global properties on the `window` object, depending on this behavior is dangerous and discouraged. It can lead to unexpected conflicts with some browser existing or future APIs. Use `document.getElementById()` or `document.querySelector()` instead.
+> Relying on this behavior is dangerous and discouraged. It can lead to unexpected conflicts with some existing or future APIs in the browser. For example, if browsers add a new global property called `preamble`, then the same code will no longer be able to access the HTML element. Use `document.getElementById()` or `document.querySelector()` instead.
 >
 > For example, if we had an element with an `id="performance"`, the following code would've worked before the introduction of Performance in modern browsers:
 >
--- a/files/en-us/web/html/reference/global_attributes/id/index.md
+++ b/files/en-us/web/html/reference/global_attributes/id/index.md
@@ -48,6 +48,20 @@ You could access the paragraph element in JavaScript using code like:
 const content = window.preamble.textContent;
 ```
 
+> [!WARNING]
+> While elements with `id` attributes are treated as global properties on the `window` object, depending on this behavior is dangerous and discouraged. It can lead to unexpected conflicts with some browser features or future APIs. Use `document.getElementById()` or `document.querySelector()` instead.
+>
+> For example, if we had an element with an `id="performance"`, the following code would've worked before the introduction of Performance in modern browsers:
+>
+> ```html
+> <h3 id="performance"></h3>
+>
+> <script>
+>     performance.innerHTML = "Well Done!";
+>     performance.style.backgroundColor = "green";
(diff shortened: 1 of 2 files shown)


files/en-us/web/html/reference/global_attributes/id/index.md
```diff
@@ -48,6 +48,20 @@ You could access the paragraph element in JavaScript using code like:
 const content = window.preamble.textContent;
 ```
 
+> [!WARNING]
+> While elements with `id` attributes are treated as global properties on the `window` object, depending on this behavior is dangerous and discouraged. It can lead to unexpected conflicts with some browser existing or future APIs. Use `document.getElementById()` or `document.querySelector()` instead.
```
Josh-Cena, line 52:
```suggestion
> Relying on this behavior is dangerous and discouraged. It can lead to unexpected conflicts with some existing or future APIs in the browser. For example, if browsers add a new global property called `preamble`, then the same code will no longer be able to access the HTML element. Use `document.getElementById()` or `document.querySelector()` instead.
```

files/en-us/web/html/reference/global_attributes/id/index.md
```diff
@@ -48,6 +48,20 @@ You could access the paragraph element in JavaScript using code like:
...
+> For example, if we had an element with an `id="performance"`, the following code would've worked before the introduction of Performance in modern browsers:
+>
+> ```html
+> <h3 id="performance"></h3>
+>
+> <script>
+>   performance.innerHTML = "Well Done!";
+>   performance.style.backgroundColor = "green";
+> </script>
+> ```
```
Josh-Cena, line 63: I suggest removing this because we can directly build upon the `preamble` example

Josh-Cena (approved):
Nice, thank you:)

younisdev:
Appreciate it! Thanks for the review!

PR #39423, opened 2025-05-06
Align aria-tab example with APG

CloudAnalystGo:
Updated the example combining tab with tablist.

### Description

I created a PR [as suggested](https://github.com/mdn/mdn/issues/671#issuecomment-2853960413).

Updated the HTML and JS in the example.

### Motivation

I read the guides on Accessibility and briefed myself on the ARIA: tab page. It is elaborate, instructive and full of potential for many examples to be created.

### Additional details

I verified these changes on [the playground](https://developer.mozilla.org/en-US/play).


CloudAnalystGo (commented):
Fixed linting issues

Josh-Cena (approved):
Thanks. We already have established standard practices in https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/, so I just made our example adhere to that. Your previous example also supported tabbing and shift-tabbing to move between tabs, which is non-standard and even harmful, because this means the user cannot quickly get to the content. Usually moving between dynamic stuff should be a different key.

CloudAnalystGo:
If the user's mouse is disabled, they won't even get to the tabbed content
at all. I have tested that in the previous example. The submitted behavior
allowed access to tabs when no noise is available...

CloudAnalystGo:
"when no mouse" is available (sorry for the autocorrect)

PR #39425, opened 2025-05-07
Replace word `include` with `exist`

zhangenming:
### Description

or some other word? setting/assign/specify ...

### Motivation

I think the word "include" might cause misunderstandings
At first, I thought the `include` meant the `attributeFilter ` array value if include some string value


Josh-Cena (commented):
Thanks for the PR. "Include a property" is a common terminology used across MDN, so it is very weird to replace this incidental usage.

PR #39428, opened 2025-05-07
Fix: image loading race condition in canvas pattern code

hanyujie2002:
- Reordered img.onload handler to be set before img.src
- Prevents potential missed events when images load from cache


Josh-Cena:
There is no race condition. The `img.src = ...` and `img.onload = ...` lines run synchronously; it is not possible for an event handler to be called in between these lines. Event jobs can only be fired after the current job is done. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Execution_model

PR #33039, opened 2024-04-11
feat: Editorial review: Long animation frames API docs

chrisdavidmills:
### Description

https://github.com/mdn/content/pull/32937 contains the engineering technical review for my work on [Long Animation Frames API](https://w3c.github.io/long-animation-frames/) docs, which has been completed and approved. Thank you to @tunetheweb and @noamr for your thorough and detailed review work.

This is a new PR based on the same branch, which is intended to contain the editorial review for the same work.

---

### Background information

The [Long Animation Frames API](https://w3c.github.io/long-animation-frames/) was released in Chrome 123. This PR adds docs for this API. 

Specifically, it:

- [x] Adds a guide to using the API
- [x] Adds reference docs for the `PerformanceLongAnimationFrameTiming` interface.
- [x] Adds reference docs for the `PerformanceScriptTiming` interface.
- [x] Adds entries to existing docs that require them, such as https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/entryType
- [x] Updates GroupData.json so that the relevant new items appear in the Performance API sidebar

PR #39439, opened 2025-05-07
Fix firstUIEventTimestamp definition

tunetheweb:
### Description

Correct error in `firstUIEventTimestamp` definition. It's for when a UI event is processed in this frame, rather than queued in this frame.

### Motivation

A developer (@andydavies) asked for this to be clarified as different to [the definition in the explainer](firstUIEventTimestamp), and [Chrome documentation](https://developer.chrome.com/docs/web-platform/long-animation-frames#frame_timestamps_and_durations)

### Additional details

See above.

Oriignally added in #33039 by @chrisdavidmills after technical review in #32937 but we missed this error during review.


wbamberg (approved):
👍 thank you!

PR #39445, opened 2025-05-07
Add links to CSS sidebar

dipikabh:
### Description

This PR adds a few existing landing pages to the sidebar that are not currently discoverable.
A broader CSS area reorganization is coming soon, but this improvement doesn't need to be delayed for that.

### Motivation

To improve the discoverability of the landing pages we already have.

### Additional details

We're currently lacking the functionality in the sidebar to do the following:
- "At-rules" in the sidebar currently only shows the top-level items.
I've tried using `depth: 2` and adding `css-at-rule-descriptor` to `tags`, but the expected child folders (e.g., for [`@counter-style`](https://github.com/mdn/content/tree/main/files/en-us/web/css/%40counter-style)) don't show up.
And we don't want the `listSubPagesGrouped` functionality here.

- "Properties" in the sidebar (and same applies to pseudo-classes and pseudo-elements) are of type `listSubPagesGrouped`.
As a result, we see `-moz-*` and `-webkit-*` groups but we don't have a handle on these top-level group items to link them to their respective landing pages. ([Firefox (-moz-) vendor-prefixed CSS extensions](https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions) and [WebKit (-webkit-) vendor-prefixed CSS extensions](https://developer.mozilla.org/en-US/docs/Web/CSS/WebKit_Extensions)).


files/sidebars/cssref.yaml
```diff
@@ -282,6 +300,13 @@ sidebar:
       - /Web/CSS/CSS_backgrounds_and_borders/Box-shadow_generator
       - /Web/CSS/CSS_backgrounds_and_borders/Border-image_generator
       - /Web/CSS/CSS_backgrounds_and_borders/Border-radius_generator
+  - title: Values and units
```
bsmth, line 303: I think this should be up under Guides after `Transitions`:

```yaml
  - title: Transitions
    details: closed
    children:
      - /Web/CSS/CSS_transitions/Using_CSS_transitions
  - title: Values and units
    details: closed
    children:
      - /Web/CSS/CSS_Values_and_Units/Value_definition_syntax
      - /Web/CSS/CSS_Values_and_Units/Numeric_data_types
      - /Web/CSS/CSS_Values_and_Units/Textual_data_types
      - /Web/CSS/CSS_Values_and_Units/Using_CSS_math_functions
```

files/sidebars/cssref.yaml
```diff
@@ -163,15 +172,20 @@ sidebar:
     details: closed
     children:
       - /Web/CSS/CSSOM_view/Coordinate_systems
+      - /Web/CSS/CSSOM_view/Viewport_concepts
   - title: Display
```
bsmth, line 176: IMO we should use this:

```yaml
  - type: listSubPages
    path: /Web/CSS/CSS_display
    title: Display
    details: closed
```

I think alpha sorting looks equally as good as the existing order, what do you reckon?

files/sidebars/cssref.yaml
```diff
@@ -187,6 +201,7 @@ sidebar:
     children:
       - /Web/CSS/CSS_fonts/OpenType_fonts_guide
       - /Web/CSS/CSS_fonts/Variable_fonts_guide
+      - /Web/CSS/CSS_fonts/WOFF
```
bsmth, line 204: Same with this entry:

```yaml
  - type: listSubPages
    path: /Web/CSS/CSS_fonts
    title: Fonts
    details: closed
```

It's less code, and nudges towards content grouped together in the right place.

files/sidebars/cssref.yaml
```diff
@@ -206,6 +221,8 @@ sidebar:
     details: closed
```
bsmth, line 221: And this one as well if you like:

```yaml
  - type: listSubPages
    path: /Web/CSS/CSS_images
    title: Images
    details: closed
```

files/sidebars/cssref.yaml
```diff
@@ -238,11 +251,16 @@ sidebar:
   - title: Positioning
     details: closed
     children:
+      - /Web/CSS/CSS_positioned_layout/Stacking_context
```
bsmth, line 254: Same comment about `listSubPages` for this dir and `CSS_scroll_snap`

files/sidebars/cssref.yaml
dipikabh: Yes! Alpha sorting works here. Replaced the manual list with `listSubPages`

files/sidebars/cssref.yaml
dipikabh: Alpha sorting is rendering as:

<kbd></kbd>

I guess we'll keep the manual sorting for this one for now:

<kbd></kbd>

WDYT?

files/sidebars/cssref.yaml
dipikabh: Of course! That was my intention. Thanks for catching. Fixed it in https://github.com/mdn/content/pull/39445/commits/9d35f4fb6b1177c87850b983f69f0c07011ad4cb.

dipikabh: Also, keeping the manual sorting here

dipikabh:
Thanks a lot @bsmth for reviewing. Using auto alpha sorting where possible is a great suggestion.
The position of `Values and units` was an error.

This is ready for you to take another look

files/sidebars/cssref.yaml
bsmth: Yeah I guess the manual ordering looks more "common -> niche" in terms of usage

bsmth (approved):
Super, everything looks good, thank you!

files/sidebars/cssref.yaml
bsmth: Let's ignore this for now

files/sidebars/cssref.yaml
bsmth: LGTM!

files/sidebars/cssref.yaml
bsmth: >keeping the manual sorting here

Sounds good!

files/sidebars/cssref.yaml
dipikabh: I think we'll get another chance to look at this with the big reorg

PR #39449, opened 2025-05-07
CSS: mask-position property description

estelle:
update the description and top syntax


files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+Each `mask-position` value in the comma-separated list of values is matched up with an associated mask layer as defined by the list of `mask-image` values, in order. If the number of values in the two properties differs:
+
+- If `mask-position` has more values than mask-image, the excess values of `mask-position` are not used.
+- If `mask-position` has fewer values than mask-image, the `mask-position` values are repeated.
+
+Each `mask-position` defines the associated mask layer's position relative to the associated {{cssxref("mask-origin")}} value. The values of the `mask-origin` are similarly matched up with the `mask-image` values, in order, with excess values being unused or repeating the values if there is an under-count. This means each mask layer has both an associated `mask-origin` and `mask-position` value. If no `mask-origin` is set, the value defaults to `padding-box` meaning the origin of each `mask-position` is the element's [padding-box](/en-US/docs/Web/CSS/box-edge#padding-box).
```
chrisdavidmills, line 76:
```suggestion
Each `mask-position` defines the associated mask layer's position relative to the associated {{cssxref("mask-origin")}} value. The `mask-origin` property values are similarly matched up with the `mask-image` values, in order, with excess `mask-position` values being unused or `mask-position` values being repeated if they are fewer in number than the `mask-origin` values. Each mask layer, therefore, has an associated `mask-origin` and `mask-position` value.

If no `mask-origin` is set, the value defaults to `padding-box`, meaning the origin of each `mask-position` is the element's [padding-box](/en-US/docs/Web/CSS/box-edge#padding-box).
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+### One-value syntax
+
+If only one value is specified, the second value is assumed to be `center`. If the single value is a `<length>` or `<percentage>`, that value defines the position of the mask along the horizontal axis, with the mask being vertically centered within the origin box. For example, `mask-position: 0%;` is equal to `mask-position: 0% center`.
```
chrisdavidmills, line 80:
```suggestion
If only one `mask-position` value is specified, the second value is assumed to be `center`. If the value is a `<length>` or `<percentage>`, it defines the position of the mask along the horizontal axis, with the mask being vertically centered within the origin box. For example, `mask-position: 0%;` is equal to `mask-position: 0% center`.
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+If you are using a keyword for positioning, the axis of the single value doesn't matter as the other value will resolve to `center`. While the default of `mask-position` is `0% 0%` which equates to `mask-position: top left`, the declaration `mask-position: top;` is equal to `mask-position: top center;`, `mask-position: left;` is equal to `mask-position: center left`, and `mask-position: center;` is equal to `mask-position: center center`.
```
chrisdavidmills, line 82:
```suggestion
If you use a single keyword for positioning, the other value will resolve to `center`. The default of `mask-position` is `0% 0%`, which equates to `mask-position: top left`. However:

- `mask-position: top;` is equivalent to `mask-position: top center;`.
- `mask-position: left;` is equivalent to `mask-position: center left`.
- `mask-position: center;` is equal to `mask-position: center center`.
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+If the value is a {{cssxref("&lt;length&gt;")}} value, it represents the horizontal position as an offset from the left edge of the mask positioning. A positive value represent an offset inward from the left edge of the box container. The position can be set outside of the element's box as a negative value is an outward offset that places the item off from the left edge of the container.
```
chrisdavidmills, line 84:
```suggestion
If the value is a {{cssxref("&lt;length&gt;")}} value, it represents the horizontal position as an offset from the left edge of the mask positioning. A positive value represents an offset inward from the left edge of the box container. The position can be set outside of the element's box using a negative value — this creates an outward offset that places the item outside the container's left edge.
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+#### Percentage values
+
+If the value is a {{cssxref("&lt;percentage&gt;")}}, the value is also the horizontal position value relative to the width of the container, positioned relative to the left edge, but the offset is not from the edge of the mask to the edge of the box, but rather, the mask image dimension is [subtracted from the container's dimension](/en-US/docs/Web/CSS/background-position#regarding_percentages), and then a percentage of the resulting value is used as the direct offset from the box's left edge, which is the same as [percentage values for `background-position`](/en-US/docs/Web/CSS/background-position#regarding_percentages). The equation is:
```
chrisdavidmills, line 88:
```suggestion
A {{cssxref("&lt;percentage&gt;")}} value represents the mask's horizontal position value relative to the width of the container, positioned relative to the left edge. However, the offset is not from the mask edge to the box edge. Instead, the mask image dimension is [subtracted from the container's dimension](/en-US/docs/Web/CSS/background-position#regarding_percentages), and then a percentage of the resulting value is used as the direct offset from the box's left edge, which is the same as [percentage values for `background-position`](/en-US/docs/Web/CSS/background-position#regarding_percentages).

The equation is:
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+`(container dimension - mask dimension) \* (position %) = (dimension offset value)`
```
chrisdavidmills, line 90:
```suggestion
`(container dimension - mask dimension) / position percentage = dimension offset value`
```

? 

The original was a bit weird

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+Given a mask that is `100px` wide and and origin box is `1000px` wide, setting `mask-postion: 10%;`, the equivalent of `10% 50%`, means that the point 10% across and 50% down the mask is to be placed at the point 10% across and 50% down the box container. The mask is vertically centered at `90px` from the left edge (`(1000 - 100) * 10% = 90`). If the left offset had been `0%`, the left edge of mask would be on the left of the container (`(1000 - 100) * 0% = 0`). If the left offset had been `100%`, the right edge of mask would be on the right of the container because the left offset would be the mask-size width away from the right edge (`(1000 - 100) * 100% = 900`).
```
chrisdavidmills, line 92:
```suggestion
Given a `100px`-wide mask and a `1000px`-wide origin box, setting `mask-postion: 10%;` (the equivalent of `10% 50%`) results in the mask being vertically centered at `90px` from the left edge (`(1000 - 100) * 10% = 90`). If the left offset had been `0%`, the mask's left edge would be flush to the left of the container (`(1000 - 100) * 0% = 0`). If the left offset had been `100%`, the mask's right edge would be flush to the right of the container (`(1000 - 100) * 100% = 900`).
```

chrisdavidmills: This paragraph took me a lot of time to unpick and understand. I've removed some of the detail, but I think it is more understandable now, as it is less of an information overload. The remaining text still allows you to understand the equation.

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+### Two-value syntax
+
+The two value `<position>` specifies the position of the mask image inside its mask positioning area with length and percentage values defaulting to being offsets from the `left` and `top`. If the two values are {{cssxref("&lt;length&gt;")}} values, {{cssxref("&lt;percentage&gt;")}} values, or the keyword `center`, the first value represents the horizontal position as an offset from the left edge of the mask positioning area and the second value the vertical position as an offset from it's top edge, with [percentages being offset](#percentage-values) by the masks dimension. If {{cssxref("&lt;percentage&gt;")}} values are specified, the first value is also the horizontal position value relative to the left edge and the second value
```
chrisdavidmills, line 96:
```suggestion
A two-value `<position>` specifies the position of the mask image inside its mask positioning area, with length and percentage values specifying offsets from the `left` and `top` of the area.

If the two values are {{cssxref("&lt;length&gt;")}} values, {{cssxref("&lt;percentage&gt;")}} values, or the keyword `center`, the first value represents the horizontal position as an offset from the left edge of the mask positioning area, and the second value represents the vertical position as an offset from it's top edge, with [percentages being offset](#percentage-values) by the mask's size in that dimension.

In addition, if {{cssxref("&lt;percentage&gt;")}} values are specified, the first value is also the horizontal position value relative to the left edge, and the second value is also the vertical position value relative to the top edge.
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+A pair of axis-specific keywords can be reordered as can an axis-spedific keywork and a length or percentage, but two length or percentage values are not interchangeable.
```
chrisdavidmills, line 98:
```suggestion
A pair of axis-specific keywords can be reordered, as can an axis-specific keyword and a length or percentage, but two length or percentage values are not interchangeable.
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+If one of the two values is `top`, `right`, `bottom`, or `left`, the order of the two values doesn't matter. Any `center` or `<length-percentage>` value in the pair of values will be applied to the other dimension.
+
+### Four-value syntax
+
+The four-value syntax is the same as the two value syntax, except the `<length>` and `<percentage>` values are offsets from the edge specified, not necessary the top or left edge. The four-value syntax allows changing the order and changing the offset edge.
```
chrisdavidmills, line 103:
```suggestion
The four-value syntax is the same as the two-value syntax, except the `<length>` and `<percentage>` values are offsets from the edge specified, not necessarily the top or left edge. The four-value syntax allows changing the order and the offset edge.
```

chrisdavidmills: I find this "changing the order and the offset edge" description and the following example paragraph really confusing. Would it be better to just say that the offset values follow the same order as four-value `margin` or `padding` values, and then give a single example?

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+For example, while `mask-position: 1em 2em` is a `1em` horizontal offset from the left box edge and a `2em` vertical offset from the top edge, `mask-position: top 1em left 2em` changes the order, being a `2em` horizontal offset from the left edge and a `1em` vertical offset from the top edge.
```
chrisdavidmills, line 105: See above comment

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+Usually the four-value syntax is used to offset from the bottom and/or right, with `mask-position:  bottom 10px right 20px` represents a `10px` vertical offset up from the bottom edge and a `20px` horizontal offset leftward from the right edge.
```
chrisdavidmills, line 107:
```suggestion
Usually the four-value syntax is used to offset from the bottom and/or right. For example, `mask-position:  bottom 10px right 20px` creates a `10px` vertical offset up from the bottom edge and a `20px` horizontal offset leftward from the right edge.
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
...
+
+Unlike the `<bg-position>` data type values for {{cssxref("background-position")}}, the `<position>` values for `mask-position` do not allow for a 3-value syntax and do not allow offsetting from `center`. When offsetting the mask from the `bottom` or `right`, the `mask-position` requires all four values (`top` or `bottom`, `left` or `right`, and two length or percentage values) to be declared.
```
chrisdavidmills, line 109:
```suggestion
Unlike the `<bg-position>` data type values for {{cssxref("background-position")}}, the `<position>` values for `mask-position` do not allow for a 3-value syntax and do not allow offsetting from `center`. When offsetting the mask from the `bottom` or `right`, the `mask-position` requires all four values (`top` or `bottom`, `left`, or `right`, and two length or percentage values) to be declared.
```

"and two length or percentage values" is confusing — this suggests that there are six values in total, which is incorrect. Do you mean that two of the values must be lengths of percentages? In which case, I'm still not sure about that. Needs clarification.

chrisdavidmills (changes requested):
@estelle good work. The description section is pretty heavy going, and I've tried to suggest simplifications where possible.

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,56 @@ mask-position: revert-layer;
 mask-position: unset;
 ```
 
-One or more `<position>` values, separated by commas.
-
 ### Values
 
+One or more `<position>` values, separated by commas.
```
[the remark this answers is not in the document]
estelle: good call. added `<position>` to the syntax headers

files/en-us/web/css/mask-position/index.md
estelle: it's a multiplication, not a division.

files/en-us/web/css/mask-position/index.md
estelle: they don't follow the same order. But if you're confused, then others will be too, so I'll fix.

files/en-us/web/css/mask-position/index.md
estelle: > One thing to note is that, unlike the `<bg-position>` data type values for {{cssxref("background-position")}}, the `<position>` values for `mask-position` do not allow for a 3-value syntax and do not allow offsetting from `center`. When offsetting the mask from the `bottom` or `right`, the `mask-position` requires all four values (either `top` or `bottom` as the vertical offset edge along with the vertical length or percentage offset value and either `left` or `right`as the horizontal offset edge along with the horizontal length or percentage offset value) to be declared.

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,68 @@ mask-position: revert-layer;
...
+
+The equation is:
+
+`(container dimension - mask dimension) * position percentage = dimension offset value`
+
+Given a `100px`-wide mask and a `1000px`-wide origin box, setting `mask-postion: 10%;` (the equivalent of `10% 50%`) results in the mask being vertically centered at `90px` from the left edge. The equation is `(1000 - 100) * 10% = 90`. If the left offset had been `0%`, the mask's left edge would be flush to the left of the container (`(1000 - 100) * 0% = 0`).
+
+If the left offset had been `100%`, the mask's right edge would be flush to the right of the container as the left edge of the `100px` wide mask would be `900px` (`(1000 - 100) * 100% = 900`) from the left edge of the container (The `100px` width put `900px` from the left edge, means the right edge would be `1000px` from the left edge, which is the right edge of the container).
```
chrisdavidmills, line 102:
```suggestion
If the left offset had been `100%`, the mask's right edge would be flush to the right of the container as the left edge of the `100px` wide mask would be `900px` (`(1000 - 100) * 100% = 900`) from the left edge of the container (the `100px` mask width plus `900px` distance from the left edge means the right edge would be `1000px` from the left edge, which is the right edge of the container).
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,68 @@ mask-position: revert-layer;
...
+
+The real power of the four-value syntax is that it allows us the define the offset edge. For example, `mask-position:  bottom 10px right 20px` creates a `10px` vertical offset up from the bottom edge and a `20px` horizontal offset leftward from the right edge. Usually the four-value syntax is used to offset from the bottom and/or right. But this syntax is also helpful if you can't remember the offset edge order for the two-value syntax.
+
+One thing to note is that, unlike the `<bg-position>` data type values for {{cssxref("background-position")}}, the `<position>` values for `mask-position` do not allow for a 3-value syntax and do not allow offsetting from `center`. When offsetting the mask from the `bottom` or `right`, the `mask-position` requires all four values (either `top` or `bottom` as the vertical offset edge along with the vertical length or percentage offset value and either `left` or `right`as the horizontal offset edge along with the horizontal length or percentage offset value) to be declared.
```
chrisdavidmills, line 121:
```suggestion
One thing to note is that, unlike the `<bg-position>` data type values for {{cssxref("background-position")}}, the `<position>` values for `mask-position` do not allow for a 3-value syntax and do not allow offsetting from `center`. When offsetting the mask from the `bottom` or `right`, the `mask-position` requires all four values to be declared.
```

I'd just say this: What is going on is fairly obvious from the previous examples, and that parenthetical made this paragraph really hard to read.

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,68 @@ mask-position: revert-layer;
...
+
+A pair of axis-specific keywords can be reordered, as can an axis-specific keyword and a length or percentage, but two length or percentage values are not interchangeable.
+If one of the two values is `top`, `right`, `bottom`, or `left`, the order of the two values doesn't matter. Any `center` or `<length-percentage>` value in the pair of values will be applied to the other dimension.
+
+### Four-value syntax
+
+The four-value syntax is the similar to the two-value syntax. In the two-value syntax, the `<length>` and `<percentage>` values are offsets from the top or left edge. The four-value syntax allows defining the offset edge. For example, `mask-position: 1em 2em` is a `1em` horizontal offset from the left box edge and a `2em` vertical offset from the top edge. In the four-value syntax, this would be `mask-position: left 1em top 2em`. Because we're defining the offset edges when using the four-value syntax, the order isn't important: Setting `mask-position: top 2em left 1em` produces the same result; also a `2em` horizontal offset from the left edge and a `1em` vertical offset from the top edge.
```
chrisdavidmills, line 117: OK, I now understand how these values work. However, I think this paragraph is bit long and waffly, and could be clarified a bit. Maybe say something like:

```suggestion
The four-value syntax consists of two pairs of values, each pair containing a keyword specifying the edge to offset from, and a `<length>` and `<percentage>` value specifying the offset distance. For example, `mask-position: left 1em top 2em` specifies a `1em` horizontal offset from the left box edge and a `2em` vertical offset from the top edge. The two-value equivalent would be `mask-position: 1em 2em`.

Because we're defining the offset edges when using the four-value syntax, the order isn't important: `mask-position: top 2em left 1em` and `mask-position: left 1em top 2em` both produce the same result.
```

files/en-us/web/css/mask-position/index.md
```diff
@@ -38,12 +57,68 @@ mask-position: revert-layer;
...
+
+The real power of the four-value syntax is that it allows us the define the offset edge. For example, `mask-position:  bottom 10px right 20px` creates a `10px` vertical offset up from the bottom edge and a `20px` horizontal offset leftward from the right edge. Usually the four-value syntax is used to offset from the bottom and/or right. But this syntax is also helpful if you can't remember the offset edge order for the two-value syntax.
```
chrisdavidmills, line 119:
```suggestion
The real power of the four-value syntax is that it allows us to specify offset edges other than `left` and `top`. For example, `mask-position:  bottom 10px right 20px` creates a `10px` vertical offset up from the bottom edge and a `20px` horizontal offset leftward from the right edge. Usually, the four-value syntax is used to offset from the bottom and/or right. But this syntax is also helpful if you can't remember the offset edge order for the two-value syntax.
```

This also begs the question, what happens if you try to specify `left` and `right`, or `top` and `bottom`. Is it just invalid?

chrisdavidmills (changes requested):
@estelle A couple more bits for you to look at, but this is getting really close now.

files/en-us/web/css/mask-position/index.md
estelle: this kind of answers the question you had above "This also begs the question, what happens if you try to specify left and right, or top and bottom. Is it just invalid?" That was my way of addressing it, so don't want to delete it (but am open to moving it)

chrisdavidmills: Yes, maybe move it to a separate paragraph, like this:

```suggestion
One thing to note is that, unlike the `<bg-position>` data type values for {{cssxref("background-position")}}, the `<position>` values for `mask-position` do not allow for a 3-value syntax and do not allow offsetting from `center`. When offsetting the mask from the `bottom` or `right`, the `mask-position` requires all four values to be declared.

For the four-value syntax to be valid, it needs to specify either `top` or `bottom` as the vertical offset edge, along with the vertical length or percentage offset value, and either `left` or `right` as the horizontal offset edge, along with the horizontal length or percentage offset value. 
```

chrisdavidmills (approved):
@estelle just one more suggestion, so I'll approve this now. Nice work.

PR #39450, opened 2025-05-08
Remove experimental status for `selectionchange` event on <input> and <textarea> elements

Zuzu-Typ:
### Description
The selectionchange event of input and textarea elements is now supported in both Chrome and Safari (as well as Firefox, which was already supporting it for a while). It should no longer be marked as experimental.

Attention: I am a first time contributor. If there are any other places where this needs to be updated (e.g. for the navigation, where an experimental icon is shown or for the Baseline newly available banner), please let me know.

### Motivation
I've helped update the browser compatibility data and now I would like to help removing the experimental status from the MDN page.

### Additional details
This change allows developers to know they can use the selectionchange event in future web applications.

See
Description | Link
-------------|-----
Related Issue | https://github.com/mdn/browser-compat-data/issues/25672
Related PR | https://github.com/mdn/browser-compat-data/pull/26681
Webkit bug tracker | https://bugs.webkit.org/show_bug.cgi?id=271033
Chrome bug tracker | https://issues.chromium.org/issues/40840956
Automated test cases | https://wpt.fyi/results/selection/textcontrols/selectionchange.html?label=experimental&label=master&aligned
JSFiddle by @caugner | https://jsfiddle.net/jx6gakhb/

### DIFF
--- a/files/en-us/web/api/htmlinputelement/selectionchange_event/index.md
+++ b/files/en-us/web/api/htmlinputelement/selectionchange_event/index.md
@@ -3,12 +3,10 @@ title: "HTMLInputElement: selectionchange event"
 short-title: selectionchange
 slug: Web/API/HTMLInputElement/selectionchange_event
 page-type: web-api-event
-status:
-  - experimental
 browser-compat: api.HTMLInputElement.selectionchange_event
 ---
 
-{{APIRef}}{{SeeCompatTable}}
+{{APIRef}}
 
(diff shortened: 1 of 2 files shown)


hamishwillee (approved):
@Zuzu-Typ Thanks. This is fine, but FYII you don't need to do it - for Web APIs there is tooling that is run "quite often" to fix up the status tagging (usually every few days as far as I know).

Great to be up to date though.

PR #39451, opened 2025-05-08
css: add 'stretch' value examples

OnkarRuikar:


OnkarRuikar:
Firefox needs the `layout.css.stretch-size-keyword.enabled` flag to be enabled on the `about:config` page.

files/en-us/web/css/height/index.md
```diff
@@ -163,6 +163,40 @@ div {
...
+  border: solid;
+}
+
+.child {
+  margin: 10px;
+  background: #0999;
+}
+
+.stretch {
+  height: -webkit-fill-available;
```
estelle, line 191: this example is demonstrating `-webkit-fill-available;`, which is not explained. we should either show it working on its own, or not show it at all. so let's kill line 191.

As we're adding a working example  ( the example with no prefix), we should remove the note, " To check aliases used by browsers for the `stretch` value and its implementation status, see the [Browser compatibility](#browser_compatibility) section."

OnkarRuikar: killed the line 191 and  the comment

OnkarRuikar:
Now test in Firefox with `layout.css.stretch-size-keyword.enabled` set.

estelle (approved):
Thanks! 🎉

PR #39225, opened 2025-04-21
mask-image: examples and description

estelle:
add a bunch of examples
create a description section

PR #39453, opened 2025-05-08
Update the styling of a star mask example

CristinaSargu:
### Description

The preview of the Star Mask example was missing some styles to display properly. I’ve added a few adjustments to ensure the star pattern fits the image correctly without being cut off.

image

### Motivation

To make the information easier to understand.

### DIFF
--- a/files/en-us/web/css/mask-image/index.md
+++ b/files/en-us/web/css/mask-image/index.md
@@ -139,6 +139,8 @@ We use `mask-star.svg` as a mask image on our first image:
 ```css
 img:first-of-type {
   mask-image: url(https://mdn.github.io/shared-assets/images/examples/mask-star.svg);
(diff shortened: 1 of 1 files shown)


bsmth:
Thanks a lot. I think it looks way better in isolation, but it's intentionally missing `mask-size`. You can see the next example uses mask-size for illustration: https://developer.mozilla.org/en-US/docs/Web/CSS/mask-image#multiple_masks

I think it might be better to close for now, because the sequence of examples wouldn't make sense in context with the prose etc. Does that make sense?

Noting this was added in https://github.com/mdn/content/pull/39225.

estelle:
Thank you for the suggestion, but the example intentionally omits the size. It's important that the masks on the left match the SVG on the right. Closing.

PR #39444, opened 2025-05-07
chore: link to a11y pages from parent, sidebar fix

bsmth:
### Description

This PR links to pages when missing from parent. Also fixes code style in sidebar for attributes and roles.

__Relates to:__

- [x] https://github.com/mdn/content/pull/39420

PR #39472, opened 2025-05-09
feat(aria): Add short titles, format titles for roles

bsmth:
### Description

In this PR, I'm making the nav consistent for ARIA attributes and roles. That means:

__Additions:__
- `short-title` for attributes and roles

__Changes:__
- Update titles for attributes to have same format as roles

For example all role pages:

```yaml
title: "ARIA: cell role"
```

become

```diff
title: "ARIA: cell role"
+short-title: cell
```

And all attribute pages:

```yaml
title: aria-activedescendant
```

Become:

```diff
+title: "ARIA: aria-activedescendant attribute"
+short-title: aria-activedescendant
```

The same change is repeated through the whole section.

Follow up from (separated out to reduce noise / complexity of that review):

- [x] https://github.com/mdn/content/pull/39444


dipikabh (approved):
This update makes sense, thanks!

Although, for the `title` should we use the format "ARIA attribute: attribute-name`" instead?
For that reason, perhaps we'd also want to update the existing "ARIA: role-name role" to "ARIA role: role-name".

We have a similar format we use in HTML titles, like here https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/capture: "HTML attribute: capture".

bsmth:
Thank you!

> Although, for the `title` should we use the format "ARIA attribute: attribute-name`" instead?

Good suggestion. Personally, I prefer `ARIA: name attribute` format like `category: name type` - this seems more natural to search for. For ARIA things that would be "button role" or "aria-required attribute". I think the same applies to HTML, like "HTML: autocomplete attribute", what do you think?

PR #39473, opened 2025-05-09
Adding @starting-style to CSS nesting at-rules

acemir:
### Description

Adding @starting-style to the list of "at-rules that can be nested"

### Motivation

On the item 2 of the [Syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style#syntax) section of @starting-style, it says the can be nested within an existing ruleset.


estelle (approved):

PR #39479, opened 2025-05-11
Add an example to search for multi-byte pattern in an array

def00111:
And manipulate the array directly.

### Motivation

There was no example for manipulating the array directly without converting the whole array to a string first.

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
@@ -327,7 +327,9 @@ Array.prototype.indexOfMulti = function (searchElements, fromIndex) {
 
   const initial = i;
   for (
-    let j = 1, m = searchElements.length, n = this.length;
+    let j = 1,
+        m = searchElements.length,
+        n = this.length;
     j < m && i < n;
     j++
   ) {
@@ -355,10 +357,14 @@ function listener(details) {
   };
 
   filter.onstop = (event) => {
-    let i, pos;
-    while ((i = data.indexOfMulti(bytes, pos)) !== -1) {
+    for (
+      let i = data.indexOfMulti(bytes),
+          m = elements.length,
+          n = bytes.length;
+      i >= 0;
+      i = data.indexOfMulti(bytes, i + m + n)
+    ) {
       data.splice(i, 0, ...elements);
-      pos = i + elements.length + bytes.length;
     }
 
     filter.write(new Uint8Array(data));
--- a/files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
@@ -427,7 +427,9 @@ function listener(details) {
       const initial = i;
       let found = false;
       for (let j = 1, l = uint8.length - i; j < l; j++) {
-        if (uint8[++i] == bytes[j] && j == l - 1) {
+        if (uint8[++i] !== bytes[j]) {
+          break;
+        } else if (j === l - 1) {
           found = true;
         }
       }
--- a/files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
@@ -327,9 +327,7 @@ Array.prototype.indexOfMulti = function (searchElements, fromIndex) {
 
   const initial = i;
   for (
-    let j = 1,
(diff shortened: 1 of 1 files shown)


dotproto:
After giving this some thought, I'm not sure this page is the right place for this example. I understand the desire to have a concrete example of working with multi-byte sequences, but our in-page code samples tend to be relatively short and easy to parse. I think [mdn/webextensions-examples](https://github.com/mdn/webextensions-examples) may be a better home for this. While we don't currently have a "cookbook" section, I feel like this might be a good candidate for that type of content.

Tagging @rebloor for a second opinion.

files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
```diff
@@ -315,6 +315,68 @@ browser.webRequest.onBeforeRequest.addListener(
...
+
+    const initial = i;
+    for (let j = 1, l = searchElements.length; j < l; j++) {
+      if (this[++i] !== searchElements[j]) {
+        return this.indexOfMulti(searchElements, initial + 1);
+      }
+    }
+    return initial;
+  },
+});
```
dotproto, line 341: Modifying built-in prototypes like this is generally considered an anti-pattern. It would be better to refactor this so `indexOfMulti` is a standalone function.

```suggestion
function indexOfMulti(arr, searchElements, fromIndex) {
  let i = arr.indexOf(searchElements[0], fromIndex);
  if (searchElements.length === 1 || i === -1) {
    // Not found or no other elements to check
    return i;
  }

  if (i + searchElements.length > arr.length) {
    return -1;
  }

  const initial = i;
  for (let j = 1, l = searchElements.length; j < l; j++) {
    if (arr[++i] !== searchElements[j]) {
      return indexOfMulti(arr, searchElements, initial + 1);
    }
  }
  return initial;
}
```

dotproto: If you'd like to improve this example, I'd suggest revising this function to use the [KMP algorithm](https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm) as this will reduce time spent on partial matches. That change isn't required for this PR, but if you don't want to make that change we should create a tracking issue to improve this in the future.

files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
```diff
@@ -315,6 +315,68 @@ browser.webRequest.onBeforeRequest.addListener(
...
+      i >= 0;
+      i = data.indexOfMulti(bytes, i + m + n)
+    ) {
+      // Insert "WebExtension " at the given index
+      data.splice(i, 0, ...elements);
+    }
+
+    filter.write(new Uint8Array(data));
+    filter.close();
+  };
```
dotproto, line 370: Another suggestion that's out of scope for this PR: progressively apply the transform in `ondata` callbacks rather than buffering the entire file in memory and serially applying the transform at the end. This would improve scenarios where content can progressively render while streaming, such as with HTML responses.

dotproto (changes requested):
@rebloor and I discussed this offline. One of the things he pointed out is that the length of this example isn't as atypical as I originally thought. My view now is that we can probably land this with a few tweaks.

files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
```diff
@@ -315,6 +315,143 @@ browser.webRequest.onBeforeRequest.addListener(
 );
 ```
 
+This example demonstrates, how to search for multi-byte pattern in an array:
+
+```js
+// JavaScript program to search the pattern in given array 
+// using KMP Algorithm
+
+function constructLps(pat, lps) {
```
def00111, line 324: Taken from https://www.geeksforgeeks.org/kmp-algorithm-for-pattern-searching/

files/en-us/mozilla/add-ons/webextensions/api/webrequest/streamfilter/ondata/index.md
```diff
@@ -315,6 +315,143 @@ browser.webRequest.onBeforeRequest.addListener(
 );
 ```
 
+This example demonstrates, how to search for multi-byte pattern in an array:
```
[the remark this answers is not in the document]
rebloor: @def00111 yes, I think this is now getting to the point where we need to move it out of the documentation into a web extensions example.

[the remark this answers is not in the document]
def00111: I opened https://github.com/mdn/webextensions-examples/pull/586

PR #39485, opened 2025-05-12
Deduplicate clipboard event pages

### LINKED ISSUE #31125: Deduplicate documentations for `copy` `cut` `paste` events
### MDN URL

-

duplicate documentations

For BCD, the three events are only defined under `Element`

But for mdn content, the three events has their documentations under `Window` `Document` `Element` `SVGGraphicsElement` and even `HTMLElement`, and they all use the BCD data under `Element`

Is it necessary, and, is some of them has been dumplite yet?

_No response_

_No response_

Josh-Cena:
Fix https://github.com/mdn/content/issues/31125.

Our current structure for these events is very weird:

- `HTMLElement` and `SVGGraphicsElement` already inherit from `Element`, so it doesn't make sense to duplicate these events
- `Document` and `Window` are bubble targets. No other events that bubble to document or window have separate pages.
- BCD only has a single entry, so there's a BCD/content mismatch.

I think it makes perfect sense to align with all other events.


wbamberg (approved):
👍 thank you!

PR #39494, opened 2025-05-13
Correct page title for `Uint8Array.prototype.setFromHex()`

dontwanttothink:
### Description

It looks like `Uint8Array.prototype.setFromHex()` is correct here instead of `Uint8Array.setFromHex()`. This is an instance method.

### Motivation

If the title is incorrect, it can be confusing.

### Additional details

- https://tc39.es/proposal-arraybuffer-base64/spec/#sec-uint8array.prototype.setfromhex


Josh-Cena (approved):
Oops, great catch, thanks :D

PR #39496, opened 2025-05-13
doc: add explanation for multiple `Transfer-Encoding` values

samuel871211:
### Motivation

Currently, the MDN document doesn't explain the meaning of multiple values of `Transfer-Encoding`.

### Additional details

https://datatracker.ietf.org/doc/html/rfc9112#section-6.1-6

```
For example,

Transfer-Encoding: gzip, chunked

indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body.
```

### DIFF
--- a/files/en-us/web/http/reference/headers/transfer-encoding/index.md
+++ b/files/en-us/web/http/reference/headers/transfer-encoding/index.md
@@ -14,11 +14,9 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
 Each segment of a multi-node connection can use different `Transfer-Encoding` values.
 If you want to compress data over the whole connection, use the end-to-end {{HTTPHeader("Content-Encoding")}} header instead.
 
-When present in a message it indicates the encodings applied to the message body.
-While multiple encodings may be specified, and the client must support the `chunked` directive, in practice this header is rarely used in requests (and if used, only with `chunked`).
+In practice this header is rarely used, and in those cases it is almost always used with `chunked`.
 
-When present on a response it indicates the compression used on the message, and/or whether the message has been chunked.
-Note that if the message is chunked, this must be applied last, after any other compression.
+That said, the specification indicates that when present in a message it indicates the compression used on the message in that hop, and/or whether the message has been chunked.
 For example, `Transfer-Encoding: gzip, chunked` indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body.
 
 The header is optional in responses to a {{HTTPMethod("HEAD")}} request as these messages have no body and, therefore, no transfer encoding.
--- a/files/en-us/web/http/reference/headers/transfer-encoding/index.md
+++ b/files/en-us/web/http/reference/headers/transfer-encoding/index.md
@@ -21,7 +21,7 @@ When present on a response it indicates the compression used on the message, and
 Note that if the message is chunked, this must be applied last, after any other compression.
 For example, `Transfer-Encoding: gzip, chunked` indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body.
 
-The response to a {{HTTPMethod("HEAD")}} request has no body and no transfer encoding, so the header is optional.
+The header is optional in responses to a {{HTTPMethod("HEAD")}} request as these messages have no body and, therefore, no transfer encoding.
 When present it indicates the value that would have applied to the corresponding response to a {{HTTPMethod("GET")}} message, if that `GET` request did not include a preferred `Transfer-Encoding`.
 
 > [!WARNING]
--- a/files/en-us/web/http/reference/headers/transfer-encoding/index.md
+++ b/files/en-us/web/http/reference/headers/transfer-encoding/index.md
@@ -14,8 +14,8 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
 Each segment of a multi-node connection can use different `Transfer-Encoding` values.
 If you want to compress data over the whole connection, use the end-to-end {{HTTPHeader("Content-Encoding")}} header instead.
 
-When present in a request it indicates the encodings supported by the client.
-Multiple encodings may be specified, and the client must support the `chunked` directive.
+When present in a request it indicates the encodings applied to the request body.
+While multiple encodings may be specified, and the client must support the `chunked` directive, in practice this header is rarely used in requests (and if used, only with `chunked`).
 
 When present on a response it indicates the compression used on the message, and/or whether the message has been chunked.
 Note that if the message is chunked, this must be applied last, after any other compression.
--- a/files/en-us/web/http/reference/headers/transfer-encoding/index.md
+++ b/files/en-us/web/http/reference/headers/transfer-encoding/index.md
@@ -45,6 +45,7 @@ Transfer-Encoding: deflate
 Transfer-Encoding: gzip
 
 // Several values can be listed, separated by a comma
+// This indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body
 Transfer-Encoding: gzip, chunked
 ```
 
--- a/files/en-us/web/http/reference/headers/transfer-encoding/index.md
+++ b/files/en-us/web/http/reference/headers/transfer-encoding/index.md
@@ -14,7 +14,7 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
 Each segment of a multi-node connection can use different `Transfer-Encoding` values.
 If you want to compress data over the whole connection, use the end-to-end {{HTTPHeader("Content-Encoding")}} header instead.
 
-When present in a request it indicates the encodings applied to the request body.
+When present in a message it indicates the encodings applied to the message body.
 While multiple encodings may be specified, and the client must support the `chunked` directive, in practice this header is rarely used in requests (and if used, only with `chunked`).
 
 When present on a response it indicates the compression used on the message, and/or whether the message has been chunked.
--- a/files/en-us/web/http/reference/headers/transfer-encoding/index.md
+++ b/files/en-us/web/http/reference/headers/transfer-encoding/index.md
@@ -14,7 +14,16 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
 Each segment of a multi-node connection can use different `Transfer-Encoding` values.
 If you want to compress data over the whole connection, use the end-to-end {{HTTPHeader("Content-Encoding")}} header instead.
 
-When present on a response to a {{HTTPMethod("HEAD")}} request that has no body, it indicates the value that would have applied to the corresponding {{HTTPMethod("GET")}} message.
+When present in a request it indicates the encodings supported by the client.
+Multiple encodings may be specified, and the client must support the `chunked` directive.
+
+When present on a response it indicates the compression used on the message, and/or whether the message has been chunked.
+Note that if the message is chunked, this must be applied last, after any other compression. 
+For example, `Transfer-Encoding: gzip, chunked` indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body.
+While multiple compression options may be specified in theory, there is no practical value in applying more than one compression.
+
+The response to a {{HTTPMethod("HEAD")}} request has no body and no transfer encoding, so the header is optional.
+When present it indicates the value that would have applied to the corresponding response to a {{HTTPMethod("GET")}} message, if that `GET` request did not include a preferred `Transfer-Encoding`.
 
 > [!WARNING]
 > HTTP/2 disallows all uses of the `Transfer-Encoding` header other than the HTTP/2 specific value `"trailers"`.
@@ -45,7 +54,6 @@ Transfer-Encoding: deflate
 Transfer-Encoding: gzip
 
 // Several values can be listed, separated by a comma
-// This indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body
 Transfer-Encoding: gzip, chunked
 ```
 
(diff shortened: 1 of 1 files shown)


hamishwillee (commented):
Thanks for this @samuel871211 

I have modified it because the document had bigger problems based on my reading of the spec. In particular, the meaning of the encoding differs depending on whether it is a request or a response.

My read of the spec is that:
- In a request the values indicate the supported encodings that the server is expected to choose from. Essentially the client is saying if it gets a message with these encodings it can decode them. You can specify multiple directives, comma separated. I assume that they are ordered by preference, but I don't see that stated.
- In a response directives indicate the encoding used for a particular message.
  - You could use multiple compression directives, but you never would - as it makes no sense.
  - However you can use any compression with `chunked`. The spec says that you can only chunk once and at the end. So a response with compression and chunking will always look like `gzip, chunked`. It can't look like `chunked, gzip`,.


@bsmth Can you take over review, as I am no longer objective?
I am fairly sure of this from the spec and "similar headers", but not certain.

samuel871211:
> * In a request the values indicate the supported encodings that the server is expected to choose from. Essentially the client is saying if it gets a message with these encodings it can decode them. You can specify multiple directives, comma separated. I assume that they are ordered by preference, but I don't see that stated.

@hamishwillee Thanks for the review. This sounds like the definition of `Accept-Encoding`. Would it be possible to share the documents you read?

hamishwillee:
@samuel871211 Yes, I am very much basing this on `Accept-Encoding` and `Content-Encoding`. I believe this does the same kind of thing, but using the same header for both the request and response.
So in the request we're using `Transfer-Encoding` header to indicate the supported encodings of the client, and in the response we're using it to indicate the actual values.

The spec really doesn't say much about its use in the request other than it can be present in HTTP1/1, so I'm really just inferring what it is for. BUT, we do know that this information is missing so we need to say something.

I'm also not 100% sure this bit is correct:L

> Transfer-Encoding is a [hop-by-hop header](https://pr39496.review.mdn.allizom.net/en-US/docs/Web/HTTP/Reference/Headers#hop-by-hop_headers), that is applied to a message between two nodes, not to a resource itself. Each segment of a multi-node connection can use different Transfer-Encoding values.

I mean I think it is true, but perhaps incomplete. The spec seems to indicate that a hop could also apply additional encoding. The reason I did not try and capture this is that I can't see where you would do this. I.e. if you're already gzip you won't compress. Perhaps you would add chunked though?

Upshot, deferring to @bsmth because he has worked on this recently.

files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -13,7 +13,16 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
 Each segment of a multi-node connection can use different `Transfer-Encoding` values.
 If you want to compress data over the whole connection, use the end-to-end {{HTTPHeader("Content-Encoding")}} header instead.
 
-When present on a response to a {{HTTPMethod("HEAD")}} request that has no body, it indicates the value that would have applied to the corresponding {{HTTPMethod("GET")}} message.
+When present in a request it indicates the encodings supported by the client.
```
bsmth, line 16: Not quite, the `TE` header is supposed to be used for this, see https://datatracker.ietf.org/doc/html/rfc9112#section-7.4

bsmth:
> I mean I think it is true, but perhaps incomplete. The spec seems to indicate that a hop could also apply additional encoding. The reason I did not try and capture this is that I can't see where you would do this. I.e. if you're already gzip you won't compress. Perhaps you would add chunked though?
> 
> Upshot, deferring to @bsmth because he has worked on this recently.

Thanks, Hamish. Looks like there's been lots of confusion about Transfer-Encoding (TE) v Content-Encoding (CE) over the years, and:

>The web has definitely centered around content-encoding rather than t-e,

* https://bugzilla.mozilla.org/show_bug.cgi?id=68517#c22
* https://stackoverflow.com/questions/11641923/transfer-encoding-gzip-vs-content-encoding-gzip - long threaded discussion

Stdlibs [like in Go](https://raindev.io/blog/http-content-and-transfer-encoding/#go-standard-library) basically use CE for compression (end-to-end) and TE to indicate message chunking.

For those reasons, we should make that a bit clear about what's permitted in the spec, and what happens in the real world - and this hop-by-hop note is good, and to point to Content-Encoding and Accept-Encoding for requesting & sending compressed resources.

files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -13,7 +13,16 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
...
+Multiple encodings may be specified, and the client must support the `chunked` directive.
```
hamishwillee, line 17:
```suggestion
When present in a request it indicates the encodings applied to the request body.
While multiple encodings may be specified, and the client must support the `chunked` directive, in practice this header is rarely used in requests (and if used, only with `chunked`).
```

files/en-us/web/http/reference/headers/transfer-encoding/index.md
hamishwillee: Thanks for finding that! I have tried to update this bit in https://github.com/mdn/content/pull/39496/files#r2114994600

hamishwillee:
@bsmth I've included your comment pretty much verbatim. Sufficient?

files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -13,7 +13,16 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
 Each segment of a multi-node connection can use different `Transfer-Encoding` values.
 If you want to compress data over the whole connection, use the end-to-end {{HTTPHeader("Content-Encoding")}} header instead.
 
-When present on a response to a {{HTTPMethod("HEAD")}} request that has no body, it indicates the value that would have applied to the corresponding {{HTTPMethod("GET")}} message.
+When present in a request it indicates the encodings applied to the request body.
```
bsmth, line 16:
```suggestion
In requests, it indicates the encodings applied to the request body.
```

files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -13,7 +13,16 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
...
+While multiple encodings may be specified, and the client must support the `chunked` directive, in practice this header is rarely used in requests (and if used, only with `chunked`).
+
+When present on a response it indicates the compression used on the message, and/or whether the message has been chunked.
```
bsmth, line 19:
```suggestion
In responses, it indicates the compression used on the message, and/or whether the message has been chunked.
```

files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -13,7 +13,16 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
...
+Note that if the message is chunked, this must be applied last, after any other compression.
+For example, `Transfer-Encoding: gzip, chunked` indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body.
+While multiple compression options may be specified in theory, there is no practical value in applying more than one compression.
+
+The response to a {{HTTPMethod("HEAD")}} request has no body and no transfer encoding, so the header is optional.
+When present it indicates the value that would have applied to the corresponding response to a {{HTTPMethod("GET")}} message, if that `GET` request did not include a preferred `Transfer-Encoding`.
```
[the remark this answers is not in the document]
hamishwillee: I agree in principle, but my initial problem with this is that it talked about responses only, so it looked like we had omitted one case. I think this is good but I'll merge and see the shape of it.

files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -14,7 +14,16 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
...
 If you want to compress data over the whole connection, use the end-to-end {{HTTPHeader("Content-Encoding")}} header instead.
 
-When present on a response to a {{HTTPMethod("HEAD")}} request that has no body, it indicates the value that would have applied to the corresponding {{HTTPMethod("GET")}} message.
+When present in a request it indicates the encodings applied to the request body.
+While multiple encodings may be specified, and the client must support the `chunked` directive, in practice this header is rarely used in requests (and if used, only with `chunked`).
+
+When present on a response it indicates the compression used on the message, and/or whether the message has been chunked.
+Note that if the message is chunked, this must be applied last, after any other compression.
+For example, `Transfer-Encoding: gzip, chunked` indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body.
+While multiple compression options may be specified in theory, there is no practical value in applying more than one compression.
```
bsmth: I would get rid of this, at least:

```suggestion
```

bsmth (approved):
I'll leave a +1 so we have this case covered. Here's one last answer from SO that I thought was good:

> Transfer-Encoding is hop-by-hop, while Content-Encoding is end-to-end.
> 
> This means that if there is a proxy involved, anywhere, the proxy will see the TE gzip, unzip it, and not necessarily forward the request as TE gzip.
> 
> So, the choices are
> 
> 1. CE gzip and always know what you will be getting, requiring logic to decompress the response.
> 2. TE gzip and never know what you will be getting requiring logic to decide whether to decompress the response and the logic to decompress it when required.
> 
> The logical choice is CE gzip.

At https://stackapps.com/a/918

files/en-us/web/http/reference/headers/transfer-encoding/index.md
hamishwillee: I'm OK with that. Its true, but if we're only really using with chunked, redundant.

files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -14,7 +14,15 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
...
 
-When present on a response to a {{HTTPMethod("HEAD")}} request that has no body, it indicates the value that would have applied to the corresponding {{HTTPMethod("GET")}} message.
+When present in a request it indicates the encodings applied to the request body.
+While multiple encodings may be specified, and the client must support the `chunked` directive, in practice this header is rarely used in requests (and if used, only with `chunked`).
+
+When present on a response it indicates the compression used on the message, and/or whether the message has been chunked.
+Note that if the message is chunked, this must be applied last, after any other compression.
+For example, `Transfer-Encoding: gzip, chunked` indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body.
+
+The response to a {{HTTPMethod("HEAD")}} request has no body and no transfer encoding, so the header is optional.
```
hamishwillee: Your text tidier
```suggestion
The header is optional in responses to a {{HTTPMethod("HEAD")}} request as these messages have no body and, therefore, no transfer encoding.
```

files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -14,7 +14,15 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
 Each segment of a multi-node connection can use different `Transfer-Encoding` values.
 If you want to compress data over the whole connection, use the end-to-end {{HTTPHeader("Content-Encoding")}} header instead.
 
-When present on a response to a {{HTTPMethod("HEAD")}} request that has no body, it indicates the value that would have applied to the corresponding {{HTTPMethod("GET")}} message.
+When present in a request it indicates the encodings applied to the request body.
```
hamishwillee:
```suggestion
When present in a message it indicates the encodings applied to the message body.
```

files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -14,7 +14,15 @@ The HTTP **`Transfer-Encoding`** {{glossary("request header", "request")}} and {
...
+When present in a message it indicates the encodings applied to the message body.
+While multiple encodings may be specified, and the client must support the `chunked` directive, in practice this header is rarely used in requests (and if used, only with `chunked`).
+
+When present on a response it indicates the compression used on the message, and/or whether the message has been chunked.
+Note that if the message is chunked, this must be applied last, after any other compression.
+For example, `Transfer-Encoding: gzip, chunked` indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body.
+
+The header is optional in responses to a {{HTTPMethod("HEAD")}} request as these messages have no body and, therefore, no transfer encoding.
+When present it indicates the value that would have applied to the corresponding response to a {{HTTPMethod("GET")}} message, if that `GET` request did not include a preferred `Transfer-Encoding`.
```
hamishwillee: OK. I messed it up, but I this fixes it up. The point being to make it clear up front this is rarely used. If you hate that, the line could move to the middle.

```suggestion
In practice this header is rarely used, and in those cases it is almost always used with `chunked`.

That said, the specification indicates that when present in a message it indicates the compression used on the message in that hop, and/or whether the message has been chunked.
For example, `Transfer-Encoding: gzip, chunked` indicates that the content has been compressed using the gzip coding and then chunked using the chunked coding while forming the message body.

The header is optional in responses to a {{HTTPMethod("HEAD")}} request as these messages have no body and, therefore, no transfer encoding.
When present it indicates the value that would have applied to the corresponding response to a {{HTTPMethod("GET")}} message, if that `GET` request did not include a preferred `Transfer-Encoding`.
```

hamishwillee (approved):
Thanks @samuel871211 @bsmth 

I think this is good. In particular it makes this about messages rather than requests or responses (and hence applies to all), and makes it clear that even though the spec allows these things, they don't usually need to happen.

bsmth:
Thanks, both! Looks better 👍🏻

PR #39502, opened 2025-05-13
Clean up #addons:mozilla.org room Matrix link

janbrasna:
### Description

Replaces the room link with its canonical version.

### Motivation

The current "internal" ID doesn't offer a good expectation where does it lead eventually.

### Additional details

Extra metadata hinting at instances being used when creating this room reference are also removed, as they are not relevant for anyone besides the original user creating said link.


willdurand (approved):

bsmth:
Thanks, both! Merging shortly 👍🏻

PR #8862, opened 2021-09-12
Use keydown instead of the deprecated keypress event

deontic:
Make the example code under the heading "Watching for the Enter key" at https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API work with `keydown` instead of the deprecated `keypress` event.
This will potentially prevent beginners or people otherwise learning from the example code from using a deprecated event, which is generally bad practice (it was not stated that the `keypress` event is deprecated within the example code).

Fixes: #None

> What was wrong/why is this fix needed? (quick summary only)

deprecated event, must be replaced

PR #39506, opened 2025-05-13
Update full screen API example code

chuchuva:
per https://mdn.github.io/dom-examples/fullscreen-api/main.js

### DIFF
--- a/files/en-us/web/api/fullscreen_api/index.md
+++ b/files/en-us/web/api/fullscreen_api/index.md
@@ -82,11 +82,15 @@ When the page is loaded, this code is run to set up an event listener to watch f
 
 ```js
 const video = document.getElementById("video");
-document.addEventListener("keypress", function(e) {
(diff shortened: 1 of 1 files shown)


wbamberg (approved):
👍 thank you @chuchuva !

Josh-Cena:
Ummm... The sync should have happened in the other direction. It is almost never the case that dom-examples is ahead of content. `keypress` was replaced with `keydown` in #8862 and arrow functions are obviously preferrred to function expressions. The only change that may be necessary is the use of `video`.

wbamberg:
Ah, good catch. I'll file a PR to revert this if you like.

wbamberg:
https://github.com/mdn/dom-examples/pull/306 is the dom-examples side.

wbamberg:
https://github.com/mdn/content/pull/39525 is the mdn/content side.

PR #39516, opened 2025-05-14
Editorial review: Document CSS if() function

chrisdavidmills:
### Description

Chrome supports the CSS [`if()`](https://drafts.csswg.org/css-values-5/#if-notation) function from version 137. See https://chromestatus.com/feature/6313805904347136.

This PR documents the new function. Specifically, it:

- [x] Adds an `if()` reference page (nearly done, just needs examples)
- [x]  Adds `if()` to the CSS values and units module page
- [x] Adds `if()` to the CSS value functions ref page.


files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,525 @@
...
+The CSS `if()` function provides conditional logic to CSS property values, and works in a similar fashion to JavaScript [`if ... else`](/en-US/docs/Web/JavaScript/Reference/Statements/if...else) statements.
+
+The `if()` function can be used inside the value of any property, and can contain zero or more `<if-condition> : <value>` pairs, followed by zero or one `else : <value>` pair.
+
+The return value is calculated as follows:
+
+1. The `<if-condition>` expressions are evaluated, in the order they appear in the function.
+2. The first `<if-condition>` that evaluates to true has its associated `<value>` returned.
+3. If none of the provided `<if-condition>` expressions evaluate to true (or no `<if-condition> : <value>` pairs are provided), the `else` `<value>` will be returned, if an `else : <value>` pair is included.
+4. If no `else : <value>` pair is included, nothing is returned.
```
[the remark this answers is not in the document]
chrisdavidmills: OK, I've tried changing this sentence to:

> If no `else : <value>` pair is included, {{glossary("guaranteed-invalid")}} is returned.

As we have a nice little glossary entry that explains guaranteed-invalid. Does that work?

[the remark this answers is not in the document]
chrisdavidmills: (in my next commit)

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,525 @@
...
+
+The CSS `if()` function provides conditional logic to CSS property values, and works in a similar fashion to JavaScript [`if ... else`](/en-US/docs/Web/JavaScript/Reference/Statements/if...else) statements.
+
+The `if()` function can be used inside the value of any property, and can contain zero or more `<if-condition> : <value>` pairs, followed by zero or one `else : <value>` pair.
+
+The return value is calculated as follows:
+
+1. The `<if-condition>` expressions are evaluated, in the order they appear in the function.
+2. The first `<if-condition>` that evaluates to true has its associated `<value>` returned.
+3. If none of the provided `<if-condition>` expressions evaluate to true (or no `<if-condition> : <value>` pairs are provided), the `else` `<value>` will be returned, if an `else : <value>` pair is included.
```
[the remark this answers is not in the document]
chrisdavidmills: Thanks, this is really useful. I've updated my description to make it clear that it is possible to put an else: value pair anywhere, but then added a new section to talk about you should mostly just have one at the end, except for in very specific circumstances, which is where your debug example comes in.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,525 @@
...
+```
+
+If the `--scheme` custom property is set to a value of `ice` on the same element, the provided `linear-gradient()` value is returned.
+
+Using style queries inside `if()` statements has an advantage over using style queries via {{cssxref("@container")}} — you can check for a set property value on an element and set a different property on the same element as a result, rather than having to check styles on a container and apply styles to the container's children.
+
+However, you can only set single property values at a time with `if()`, whereas `@container` queries can be used to set whole sets of rules. You can also only check for a single property value at a time with `if()`, and can't use `and`, `or`, or `not`. There is no way to do something like this:
+
+```css
+style(--scheme: dark or --scheme: very-dark): black;
```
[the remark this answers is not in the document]
chrisdavidmills: OK, I've rewritten it, tested and included some and/or/not examples.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,525 @@
...
+media(print): white;
+```
+
+You can also use media features — the following returns a value of `0 auto` if the current viewport width is less than `700px`:
+
+```css
+media(width < 700px): 0 auto;
+```
+
+This is really useful when you need to vary a single property value based on a media query result, however, you can't use complex media queries inside an `if()` statement (multiple queries connected via `and`, `or`, or `not` keywords). For such cases, a regular {{cssxref("@media")}} construct is needed. The two approaches are complementary, and have different uses.
```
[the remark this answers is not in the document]
chrisdavidmills: OK, I've updated this section to suit, and included `and` and `or` examples. However, I had real trouble getting a `not` example to work. I ended up trying many different permutations, for example

```
background-color: if(
  media(not (width < 500px)): blue
);

background-color: if(
  not media(width < 500px): blue
);
``` 

Can you help me find an example that works? Or, is this a bug?

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,525 @@
...
+supports(selector(h2 > p)): blue;
+```
+
+Interestingly, support queries for a property without a value don't work, for example:
+
+```css
+supports(color): blue;
+```
+
+Feature queries are really useful inside `if()` statements when you need to vary a single property value based on support for a particular property: value combination. however, you can't use complex feature queries inside an `if()` statement (multiple queries connected via `and`, `or`, or `not` keywords). For such cases, a regular {{cssxref("@supports")}} construct is needed. The two approaches are complementary, and have different uses.
```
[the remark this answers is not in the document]
chrisdavidmills: Cool, updated, and I've added some examples.

files/en-us/web/css/css_values_and_units/index.md
```diff
@@ -47,7 +48,7 @@ Every CSS declaration consists of a property/value pair. The value can take vari
 - {{cssxref("tan()")}}
 - {{cssxref("url()")}}
 
-Additional functions, including `calc-mix()`, `crossorigin()`, `first-valid()`, `if()`, `integrity()`, `progress()`, `random()`, `random-item()`, `referrerpolicy()`, `sibling-count()`, `sibling-index()`, `src()`, `type()`, and `toggle()`, are defined in the specifications, but not yet implemented in browsers.
+Additional functions, including `calc-mix()`, `crossorigin()`, `first-valid()`, `integrity()`, `progress()`, `random()`, `random-item()`, `referrerpolicy()`, `sibling-count()`, `sibling-index()`, `src()`, `type()`, and `toggle()`, are defined in the specifications, but not yet implemented in browsers.
```
una, line 51: `sibling-count()` and `sibling-index()` are now implemented in Chromium. Not sure if this is worth mentioning/changing

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+```css-nolint
+background-image: if (
+  style(--scheme: ice): linear-gradient(to left, #caf0f8, white, #caf0f8);
+)
+```
+
+If the `--scheme` custom property is set to a value of `ice` on the same element, the provided `linear-gradient()` value is returned.
+
+Using style queries inside `if()` statements has an advantage over using style queries via {{cssxref("@container")}} — you can check for a set property value on an element and set a different property on the same element as a result, rather than having to check styles on a container and apply styles to the container's children.
```
una: Another advantage is you can target the element directly and it doesn't require a parent element

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+When you want to set multiple declarations or rules based on one media query, a regular {{cssxref("@media")}} construct is needed. The two approaches are complementary, and have different uses.
+
+#### Feature queries
+
+[Feature query](/en-US/docs/Web/CSS/CSS_conditional_rules/Using_feature_queries) if conditions can be used to set a value for a property depending on whether the browser supports a particular property value.
+
+For example, the following returns an {{cssxref("color_value/lch()")}} color, if `lch()` colors are supported:
+
+```css-nolint
```
una, line 240: should these example show an else? i.e. what happens if theyre not supported?

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,608 @@
...
+color: if(
+  supports((selector(h2 + p)) or (color: not-a-color)): blue;
+);
+
+color: if(
+  supports(not selector(h2 + p)): blue;
+);
+```
+
+Interestingly, support queries for a property without a value don't work, for example:
```
una: @andruud is this an implementation gap or is it specced to be like this?

files/en-us/web/css/if/index.md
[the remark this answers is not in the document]
andruud: Hmm, this works for me: https://jsbin.com/polupuzelo/edit?html,output

files/en-us/web/css/if/index.md
andruud: It's per spec: you need to give `@supports` a full _declaration_.

files/en-us/web/css/css_values_and_units/index.md
chrisdavidmills: Totally, but we should do that as part of a separate PR. @rachelandrew, can you add these to my work stream please? Cheers.

files/en-us/web/css/if/index.md
andruud: Also, this example is a bit silly. If `color` isn't supported, the whole declaration (containing the `if()`) will be dropped anyway.

andruud:
Looks good. I don't have permission to click "Approve" here, though.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+The CSS `if()` function provides conditional logic to CSS property values, and works in a similar fashion to JavaScript [`if ... else`](/en-US/docs/Web/JavaScript/Reference/Statements/if...else) statements.
+
+The `if()` function can be used inside the value of any property, and can contain zero or more `<if-condition> : <value>` pairs and `else : <value>` pairs.
+
+The return value is calculated as follows:
+
+1. The `<if-condition>` expressions are evaluated, in the order they appear in the function.
+2. The first `<if-condition>` that evaluates to true, or `else` that appears, has its associated `<value>` returned.
+3. If none of the conditions evaluate to true, {{glossary("guaranteed_invalid_value", "guaranteed-invalid")}} is returned, in which case the property retains its initial value.
```
tabatkins: "property retains its initial value" isn't necessarily true, depending on where the if() is used - it can trigger fallback, etc. Maybe just say it's "similar to an failed var() substitution".

files/en-us/web/css/if/index.md
tabatkins: In fact, I think it's worthwhile to make this much more prominent - an if() style() query is *not* a container query, it's querying the style *on the element itself*. We'll likely add container queries later.

chrisdavidmills: I think this is basically what I was trying to say, but you put it much more clearly and concisely ;-)

I've updated the paragraph to:

> Using style queries inside `if()` statements has an advantage over using style queries via {{cssxref("@container")}} — you can target an element with styles directly based on whether a custom property is set on it, rather than having to check set styles on a container parent element.

Does that work for you?

files/en-us/web/css/if/index.md
chrisdavidmills: It is a good idea to show a couple of examples. I have updated it to do so.

I've not added an `else`  to all of them, as I didn't think it was that useful to do so.

files/en-us/web/css/if/index.md
chrisdavidmills: Yeah, I agree. I've tried updating it to:

```
margin: if(
  supports(justify-content): 0;
)
```

You could conceivably implement some kind of default fallback layout using `margin: 0 auto` that is used if flexbox isn't supported, but then unset it if it is supported...maybe? Hrm, it was hard to come up with examples for this stuff.

chrisdavidmills:
> Looks good. I don't have permission to click "Approve" here, though.

@andruud no worries — thanks for the review! I will move this one on to the editorial review stage now.

files/en-us/web/css/if/index.md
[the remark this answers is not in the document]
chrisdavidmills: OK, that is weird — I don't know what was going on when I was trying this out ;-)

I've verified it works, and added an example to the page. Thanks for the help here.

files/en-us/web/css/if/index.md
tabatkins: This still is an invalid `supports()` query; you need a full property (name + value) to query support. This isn't new to `if()`, it's been how `@supports` has worked since the feature was first defined.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+  supports(color: lch(77.7% 0 0)): lch(77.7% 0 0);
+  else: rgb(192, 192, 192);
+)
+```
+
+Selector support queries also work. The following will return a value of `0` if the browser supports the `h2 > p` selector, or `initial` if not:
+
+```css-nolint
+margin-top: if(
+  supports(selector(h2 + p)): 0;
```
tabatkins: `h2 + p` has been part of CSS since CSS1; maybe we want a slightly more realistic example?

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+background-image: if(
+  style(--scheme: ice): linear-gradient(to left, #caf0f8, white, #caf0f8);
+  style(--scheme: fire): linear-gradient(to left, #ffc971, white, #ffc971);
+  else: none
+);
+
+/* Partial value */
+border: 3px solid if(
+  supports(color: lch(77.7% 0 0)): lch(77.7% 0 0);
+  else: #c0c0c0
```
estelle:
```suggestion
  else: #03045e;
);
border: if(
  supports(color: lch(77.7% 0 0)): 3px solid lch(77.7% 0 0);
  else: 3px solid #c0c0c0;
);

/* Multiple tests */
background-image: if(
  style(--scheme: ice): linear-gradient(to left, #caf0f8, white, #caf0f8);
  style(--scheme: fire): linear-gradient(to left, #ffc971, white, #ffc971);
  else: none;
);

/* Partial value */
border: 3px solid if(
  supports(color: lch(77.7% 0 0)): lch(77.7% 0 0);
  else: #c0c0c0;
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+3. If none of the conditions evaluate to true, {{glossary("guaranteed_invalid_value", "guaranteed-invalid")}} is returned, in which case the property retains its initial value.
+
+For example:
+
+```css-nolint
+div {
+  background-image: if(
+    style(--scheme: ice): linear-gradient(to left, #caf0f8, white, #caf0f8);
+    style(--scheme: fire): linear-gradient(to left, #ffc971, white, #ffc971);
+    else: none
```
estelle:
```suggestion
    else: none;
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+div {
+  background-image: if(
+    /* Is this one not working? */
+    style(--scheme: ice): linear-gradient(to left, #caf0f8, white, #caf0f8);
+    /* Not working indicator image */
+    else: url("debug.png");
+    /* Never considered */
+    style(--scheme: fire): linear-gradient(to left, #ffc971, white, #ffc971);
+    /* Never considered */
+    else: none
```
estelle:
```suggestion
    else: none;
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+background-color: if(
+  style((--scheme: dark) or (--scheme: very-dark)): black
+);
+
+background-color: if(
+  style((--scheme: dark) and (--contrast: hi)): black
+);
+
+background-color: if(
+  not style(--scheme: light): black
```
estelle:
```suggestion
  style((--scheme: dark) or (--scheme: very-dark)): black;
);

background-color: if(
  style((--scheme: dark) and (--contrast: hi)): black;
);

background-color: if(
  not style(--scheme: light): black;
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+slug: Web/CSS/if
+page-type: css-function
+status:
+  - experimental
+browser-compat: css.types.if
+---
+
+{{CSSRef}}{{SeeCompatTable}}
+
+The **`if()`** [CSS](/en-US/docs/Web/CSS) [function](/en-US/docs/Web/CSS/CSS_Values_and_Units/CSS_Value_Functions) allows different values to be set for a property depending on the result of a test. The test can be based on a [style query](/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_style_queries), a [media query](/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries), or a [feature query](/en-US/docs/Web/CSS/CSS_conditional_rules/Using_feature_queries).
```
estelle:
```suggestion
The **`if()`** [CSS](/en-US/docs/Web/CSS) [function](/en-US/docs/Web/CSS/CSS_Values_and_Units/CSS_Value_Functions) allows different values to be set for a property depending on the result of a conditional test. The test can be based on a [style query](/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_style_queries), a [media query](/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries), or a [feature query](/en-US/docs/Web/CSS/CSS_conditional_rules/Using_feature_queries).
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+## Syntax
+
+```css-nolint
+/* Style query condition */
+background-color: if(style(--scheme: dark): black);
+padding: if(style(--size: 2xl): 1em; else: 0.25em);
```
estelle:
```suggestion
background-color: if(style(--scheme: dark): black;);
padding: if(style(--size: 2xl): 1em; else: 0.25em;);
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+/* Media query condition */
+background: if(media(print): white; else: black);
+margin: if(media(max-width: 700px): 0 auto ; else: 20px auto 0);
```
estelle:
```suggestion
background: if(media(print): white; else: black;);
margin: if(media(max-width: 700px): 0 auto ; else: 20px auto 0;);
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+browser-compat: css.types.if
+---
+
+{{CSSRef}}{{SeeCompatTable}}
+
+The **`if()`** [CSS](/en-US/docs/Web/CSS) [function](/en-US/docs/Web/CSS/CSS_Values_and_Units/CSS_Value_Functions) allows different values to be set for a property depending on the result of a test. The test can be based on a [style query](/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_style_queries), a [media query](/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries), or a [feature query](/en-US/docs/Web/CSS/CSS_conditional_rules/Using_feature_queries).
+
+## Syntax
+
+```css-nolint
```
estelle, line 16: My preference is for grouping them by syntax rather than by conditional type. Here's my suggestion (https://codepen.io/pen/)

according to the template, "Only include the function itself, not a complete declaration in which it occurs"

```css
/* Single <if-test> */
if(style(--scheme: dark): #eee;)
if(media(print): #000;)
if(media(width > 700px): 0 auto;)
if(supports(color: lch(7.1% 60.23 300.16)): lch(7.1% 60.23 300.16);)

/* <if-test> with else */
if(style(--size: 2xl): 1em; else: 0.25em;)
if(media(print): white; else: black;)
if(media(width < 700px): 0 auto ; else: 20px auto)
if(
  supports(color: lch(7.1% 60.23 300.16)): lch(7.1% 60.23 300.16);
  else: #03045e;
)
if(
  supports(color: lch(77.7% 0 0)): 3px solid lch(77.7% 0 0);
  else: 3px solid #c0c0c0;
)

/* Multiple <if-test>s */
if(
  style(--scheme: ice): linear-gradient(#caf0f8, white, #caf0f8);
  style(--scheme: fire): linear-gradient(#ffc971, white, #ffc971);
  else: none;
)

/* <if-test> within a shorthand */
3px yellow if(
  style(--color: green): dashed;
  style(--color: yellow): inset;
  else: solid;
)
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+### Parameters
+
+The `if()` function's syntax is as follows:
+
+```plain
+if([[<if-condition> : <value>;] | [else : <value>;]]*)
+```
+
+The parameters are:
```
estelle: Based on https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types/CSS_function_page_template

```suggestion
The parameter is a semi-colon–separated list of `<if-branch>`es. Each `<if-branch>` is an `<if-condition>` followed by a colon and a `<value>. 

```plain
<if-branch> = <if-condition> : <value>;
```/endplain

The semi-colon is optional for the last `<if-branch>`. 

- `<if-condition>`

  - : An `<if-test>`s or the `else` keyword.
```
and then include the rest, but idented one level.
not sure if the "The semi-colon is optional for the last `<if-branch>`. " should be here or elsewhere yet (still reading)

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+- `<if-condition>`
```
estelle, line 59:
```suggestion
- `<if-test>`
```

files/en-us/web/css/if/index.md
estelle: if we keep the syntax thing, it should be:

```plain
if([[<if-test> : <value>;] | [else : <value>;]]*)
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+  - : A keyword representing an `<if-condition>` that always evaluates to true.
+
+- `<value>`
+
+  - : A property value.
+
+### Return value
+
+A value, or {{glossary("guaranteed_invalid_value", "guaranteed-invalid")}}, in which case the property retains its initial value.
```
estelle:
```suggestion
A value or {{glossary("guaranteed_invalid_value", "guaranteed-invalid")}}.
```

estelle: if it's in a custom prop, the fallback is used, not initial i think.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+## Description
+
+The CSS `if()` function provides conditional logic to CSS property values, and works in a similar fashion to JavaScript [`if ... else`](/en-US/docs/Web/JavaScript/Reference/Statements/if...else) statements.
+
+The `if()` function can be used inside the value of any property, and can contain zero or more `<if-condition> : <value>` pairs and `else : <value>` pairs.
```
estelle: can it be used in a calc() or other function?

files/en-us/web/css/if/index.md
estelle: "else" is an "if-condition", so we should use "if-test" here and the rest of the doc.

files/en-us/web/css/if/index.md
estelle:
```suggestion
The `if()` function can be used inside the value of any property, and can contain zero or more semi-colon-separated `<if-condition>`s. Each `<if-condition>` is either an `<if-test> : <value>` pair or or an `else : <value>` pair. The semi-colon after the last `<if-condition>` is optional.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+The return value is calculated as follows:
+
+1. The `<if-condition>` expressions are evaluated, in the order they appear in the function.
+2. The first `<if-condition>` that evaluates to true, or `else` that appears, has its associated `<value>` returned.
```
estelle:
```suggestion
2. The first `<if-condition>` that evaluates to true, whether it's an `<if-test>` or an `else`, has its associated `<value>` returned.
```

or

```suggestion
2. The first `<if-condition>` that evaluates to `true` has its associated `<value>` returned.
```

files/en-us/web/css/if/index.md
estelle: what he said.

how about something like this: (the links i added have to be fixed)
```suggestion
3. If no `<if-condition>` evaluates to `true`, the function returns a {{glossary("guaranteed_invalid_value", "&lt;guaranteed-invalid&gt:")}}. This behaves as invalid or `false` if the `if()` function is used in a value statement that has a fallback, such as a {{cssxref("custom-property")}} or [`anchor()`]() function.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+    style(--scheme: fire): linear-gradient(to left, #ffc971, white, #ffc971);
+    else: none
+  );
+}
+```
+
+In this case, we are setting a different {{cssxref("background-image")}} {{cssxref("linear-gradient()")}} value on {{htmlelement("div")}} elements depending on whether a `--scheme` [custom property](/en-US/docs/Web/CSS/--*) is set to `ice` or `fire`. If neither value is set, the `else` value comes into play, and the `background-image` property is set to `none`.
+
+> [!NOTE]
+> Each condition must be separated from its associated value with a colon, and each `<if-condition> : <value>` pair must be separated with a semi-colon.
```
estelle: optionally put the "the last semi-colon is optional" stuff here instead of or in addition to where i suggested it above.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+> [!WARNING]
+> There must be no space between the `if` and the opening parenthesis (`(`). If there is, the whole declaration is invalid.
+
+If a single `<if-condition>` or `<value>` is invalid, it does not invalidate the entire `if()` function; instead, the parser moves on to the next `<if-condition> : <value>` pair.
+
+### Frequency and position of `else : <value>` pairs
+
+It is possible to include as many `else : <value>` pairs as you want inside an `if()` function, in whatever positions you want. However, in most cases it makes sense to use a single `else : <value>` pair at the end of your values list to provide a default value that is always returned if none of the supplied conditions evaluate to true.
```
estelle:
```suggestion
You can include multiple `else : <value>` pairs inside an `if()` function, in any position. However, in most cases, a single `else : <value>` pair at the end of the semi-colon-separated list is used to provide the default value that is always returned if none of the `<if-test>`s evaluate to true.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+If you included it before the `<if-condition> : <value>` pairs, the `else` would evaluate to true, and none of the later pairs would be considered. The following example `if()` always returns `none`:
```
estelle:
```suggestion
If you include an `else : <value>` pair before any `<if-test> : <value>` pairs, the later conditions are not evaluated because the `else` always evaluates to `true`. The following `if()` always returns `none`:
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+```css-nolint
+div {
+  background-image: if(
+    else: none;
+    /* Never considered */
+    style(--scheme: ice): linear-gradient(to left, #caf0f8, white, #caf0f8);
+    /* Never considered */
```
estelle:
```suggestion
    style(--scheme: ice): linear-gradient(#caf0f8, white, #caf0f8);
```
let's move the comments into the text
i also suggested removing the `to left` throughout the doc to make the examples shorter.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+  );
+}
+```
+
+One circumstance in which you might want to put an `else : <value>` in a position other than the end of the values list is when a value is not behaving as expected, and you are trying to debug it. For example:
+
+```css-nolint
+div {
+  background-image: if(
+    /* Is this one not working? */
```
estelle: covert the comments to explanatory text

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+    else: url("debug.png");
+    /* Never considered */
+    style(--scheme: fire): linear-gradient(to left, #ffc971, white, #ffc971);
+    /* Never considered */
+    else: none
+  );
+}
+```
+
+Note that an `if()` function is still valid if it contains only an `else : <value>` pair, or nothing at all:
```
estelle:
```suggestion
Note that an `if()` function is still valid if it contains only an `else : <value>` pair, or nothing at all. The following property values are both valid:
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+background-image: if();
+```
+
+However, these functions are not useful. The {{cssxref("background-color")}} value will always be set to `yellow`, and the `background-image` will retain its initial value. You would be better off setting the first property to just `yellow`, and not setting the second property at all:
+
+```css
+background-color: yellow;
+```
+
+### Types of if condition
```
estelle:
```suggestion
### Types of if-tests
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+```css
+background-color: if(else: yellow);
+background-image: if();
+```
+
+However, these functions are not useful. The {{cssxref("background-color")}} value will always be set to `yellow`, and the `background-image` will retain its initial value. You would be better off setting the first property to just `yellow`, and not setting the second property at all:
+
+```css
+background-color: yellow;
+```
```
estelle:
```suggestion
These functions are not useful. They were included to demonstrate their validity. In this case, the {{cssxref("background-color")}} value will always be set to `yellow` and the `background-image` will be set to its initial value. You would be better off setting the `background-color` directly to `yellow`, and the `background-image` to  `initial` or `none`.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+### Types of if condition
+
+An `<if-condition>` can be one of three different types. This section looks at each on in detail.
```
estelle:
```suggestion
An `<if-test>` accepts one of three query types. This section looks at each one in detail.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+#### Style queries
+
+[Container style query](/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_style_queries) if conditions allow you to test whether a particular property value is set on an element, and apply a value to a different property as a result. We walked through a style query example earlier on; let's look at another example:
```
estelle:
```suggestion
A [container style query](/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_style_queries) `<if-test>` allows you to test whether a particular property value is set on an element, and apply a value to a different property as a result. We walked through several style query examples earlier on; let's look at another example:
```

files/en-us/web/css/if/index.md
estelle: how about `sibling-count()`?

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+);
+
+margin-top: if(
+  supports(not selector(h2 + p)): 0;
+);
+```
+
+Interestingly, support queries for a property without a value don't work, for example:
+
+```css-nolint
```
estelle: add the keyword to make this "bad code"

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+margin: if(
+  supports(justify-content): 0;
+)
+```
+
+Feature queries are really useful inside `if()` statements when you need to vary a single property value based on support for a particular property: value combination.
+
+When you want to set multiple declarations or rules based on one feature query, a regular {{cssxref("@supports")}} construct is needed. The two approaches are complementary, and have different uses.
```
estelle:
```suggestion
Feature queries are really useful inside `if()` statements when you need to vary a single property value based on support for a particular value or a separate property. When you want to set multiple declarations or rules based on one feature query, a regular {{cssxref("@supports")}} construct is better. The two approaches are complementary, and have different uses.
```
pedantic, but "needed" means you can't use this, and you could

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+### Providing fallback values
+
+`if()` statements do not gracefully degrade; explicit fallbacks need to be provided for non-supporting browsers.
```
estelle:
```suggestion
The `if()` statement does not gracefully degrade; an explicit fallback needs to be provided for non-supporting browsers.
```
To not start with code

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+For example, in this case we provide a static {{cssxref("padding")}} value for browsers that don't support `if()`. Browsers that do support `if()` will override the first declaration with the second one, which sets a padding value dynamically depending on whether the `--size: 2xl` custom property is set.
```
estelle:
```suggestion
For example, in this case we provide a static {{cssxref("padding")}} value for browsers that don't support `if()`. Browsers that support `if()` will override the first declaration with the second one, which sets different padding values depending on whether or not the `--size: 2xl` custom property is set.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+```css
+padding: 1em;
+padding: if(style(--size: 2xl): 1em; else: 0.25em);
+```
+
```
estelle, line 284: I would like to add something like this here:
```suggestion
Remember to include the `else` condition. In `if()` supporting browsers, if no `else` value were included and `--size` was not equal to `2xl`, the padding would be set to `initial`.

```

just to bring the point home

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+### Whole and partial values
+
+An `if()` function can be set as the value of any CSS property, but it can also be used to determine parts of property values. For example, the following sets a different {{cssxref("border-color")}} inside a {{cssxref("border")}} shorthand property, depending on whether {{cssxref("color_value/lch()")}} colors are supported:
+
+```css-nolint
+border: if(
+  supports(color: lch(77.7% 0 0)): 3px solid lch(77.7% 0 0);
```
estelle, line 294: If you want to use something that isn't supported in all browsers, there's a list. My "pro-tip" is when i am looking for unsupported properties, I go to https://openwebdocs.github.io/web-docs-backlog/chromium/ and search in-page for CSS.

estelle: do we want to do:

`supports(color: lch(0 0 0))` instead of the used value, here and elsewhere, to make it shorter?

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+  else: 3px solid #c0c0c0;
+);
+```
+
+But you can simplify it like this, using the `if()` function to determine only the color portion:
```
estelle:
```suggestion
You can also use the `if()` function to set just one component in a multi-component shorthand. Here we use the `if()` function to  determine the {{cssxref("border-color")}} component only:
```
getting rid of "simplify". It is a few characters shorter, but I don't think that's not enough of a simplification to use simplify.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+```css-nolint
+border: 3px solid
+  if(
+    supports(color: lch(77.7% 0 0)): 3px solid lch(77.7% 0 0) ; else: #c0c0c0
```
estelle:
```suggestion
    supports(color: lch(77.7% 0 0)): 3px solid lch(77.7% 0 0); else: #c0c0c0;
```

or

```
    supports(color: lch(77.7% 0 0)): 3px solid lch(77.7% 0 0);
    else: #c0c0c0;
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+  );
+```
+
+### Nesting if() functions
+
+Because an `if()` function can take the place of whole or partial property values, it is possible to nest multiple `if()` functions.
+
+For example, in this declaration we are using `if()` to set a `color` property value depending on various conditions. We have an outer `if()` function that returns a particular value depending on whether the `--scheme` custom property is set to `ice` or `fire` (with an `else` value of `black` that is returned if neither of the other conditions return true).
```
estelle:
```suggestion
For example, in this declaration we are using `if()` to set a `color` property value depending on various conditions. We have an outer `if()` function that returns a particular value depending on whether the `--scheme` custom property is set to `ice` or `fire` (with an `else` value of `black` that is returned if neither of the conditions return true).
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+However, the two if condition values are also `if()` functions. These inner `if()` functions return an `lch()` color value if the browser supports `lch()`, or a hex value otherwise.
```
estelle:
```suggestion
If `--scheme` is set to `ice` or `fire`, the color value is also set  with `if()` functions. These inner `if()` functions return an `lch()` color value if the browser supports `lch()`, or a hex value otherwise.
```

estelle: Let's change this example to a media query example. LCH came before `if()`, so there will never be a browser that supports if but not LCH. LCH is baseline new, but will be baseline widely available in 3 months. I assume if will be used often with style and media queries and very rarely with supports.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+```html live-sample___basic
+<section style="--show-apple:true">
+  <article><h2>First article</h2></article>
+  <article><h2>Second article</h2></article>
+</section>
+```
+
+#### CSS
+
+In our CSS, we first target the `<section>` element, laying it out with [flexbox](/en-US/docs/Web/CSS/CSS_flexible_box_layout) and setting a {{cssxref("gap")}} between the two child `<article>` elements. We then use an `if()` function with a media query if condition to set the value of the {{cssxref("flex-direction")}} property to `row` if the document is in landscape orientation, or `column` if it is in portrait orientation. This lays out the `article` elements side-by-side in wide screens, and top-to-bottom in narrow screens.
```
estelle:
```suggestion
In our CSS, we first target the `<section>` element, laying it out with [flexbox](/en-US/docs/Web/CSS/CSS_flexible_box_layout) and setting a {{cssxref("gap")}} between the two child `<article>` elements. We then use an `if()` function with a [media query](/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries) `<if-test>` to set the value of the {{cssxref("flex-direction")}} property to `row` if the document is in landscape orientation, or `column` if it is in portrait orientation. This lays out the `article` elements side-by-side in wide screens, and top-to-bottom in narrow screens.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+  display: flex;
+  gap: 16px;
+  flex-direction: if(
+    media(orientation:landscape): row;
+    else: column;
+  )
+}
+```
+
+Next, we target the `<h2>` element's {{cssxref("::before")}} pseudo-element, setting its {{cssxref("content")}} property to contain an apple emoji., but only if `--show-apple:true` is set (we did this earlier in our HTML). We achieve this using an `if()` function with a style query if condition:
```
estelle:
```suggestion
Next, we target the `<h2>` element's {{cssxref("::before")}} pseudo-element, setting its {{cssxref("content")}} property to an apple emoji, but only if `--show-apple: true` is set (we did this earlier with an inline {{htmlelement("style")}} in our HTML). We achieve this using an `if()` function with a [style query](/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#style_queries_for_custom_properties) `<if-test>`:
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+```css-nolint live-sample___basic
+h2::before {
+  content: if(
+    style(--show-apple:true): "🍎 ";
+  );
+}
+```
+
+Finally, we target the `<h2>` element itself. We use a feature query if condition to test whether the browser supports `lch()` colors, and set the {{cssxref("color")}} property to an `lch()` color if so, or a hex equivalent if not.
```
estelle:
```suggestion
Finally, we target the `<h2>` elements. We use a [feature query](/en-US/docs/Web/CSS/CSS_conditional_rules/Using_feature_queries) if-test to test whether the browser supports `lch()` colors, and set the {{cssxref("color")}} property to an `lch()` color if so, or a hex equivalent if not.
```
again, i recommend changing the LCH to something else

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+#### Result
+
+This demo will render as follows:
+
+{{EmbedLiveSample("basic", "100%", "240")}}
+
+Note how the styling is applied. You can test out the conditional styling pretty easily for at least the first two `if()` queries by modifying the rendered demo using your browser's devtools:
+
+- Try for example removing the `<section>` element's `style` attribute and note how the apple emojis are no longer rendered.
```
estelle:
```suggestion
- Remove the `<section>` element's `style` attribute and note how the apple emojis are no longer rendered.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+- You could also try changing the `height` attribute of the embedding `<iframe>` to say `1200px`, resulting in the document orientation changing from landscape to portrait, and note how the layout changes as a result.
```
estelle:
```suggestion
- Change the `height` attribute of the embedding `<iframe>` to `1200px`. This will change document orientation from landscape to portrait. Note how the layout changes as a result.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
+- [CSS values and units modules](/en-US/docs/Web/CSS/CSS_Values_and_Units)
```
estelle:
```suggestion
- [CSS values and units module](/en-US/docs/Web/CSS/CSS_values_and_units)
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+aside {
+  color: var(--color2);
+  background-color: var(--color1);
+  padding: 20px;
+}
+```
+
+Finally, we use `if()` functions in a couple more places:
+
+- We set our `<h1>` element's {{cssxref("font-size")}} to be `calc(3rem + 2vw)` if the viewport is wider than `700px`, and `3rem` if not. This means that the font sizes dynamically with the viewport on wider screens, but stays the same on narrow screens.
```
estelle:
```suggestion
- We set our `<h1>` element's {{cssxref("font-size")}} to be `calc(3rem + 2vw)` if the viewport is wider than `700px`, and `3rem` otherwise. This means the font size updates dynamically with viewport width changes on wider screens, but stays the same on narrow screens.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+.ice {
+  --scheme: ice;
+}
+
+.fire {
+  --scheme: fire;
+}
+```
+
+Here we can start to see the real power of CSS `if()` functions combined with custom properties. We are using `if()` functions to set our `--color1` and `--color2` custom properties to different color values depending on the value set for the `--scheme` custom property. We then use the `--color1` and `--color2` values in our `<article>` element's {{cssxref("color")}}, {{cssxref("border")}}, and {{cssxref("background-image")}} properties, and our `<aside>` element's `color` and `background-color` properties.
```
estelle:
```suggestion
Here we can start to see the real power of CSS `if()` functions combined with custom properties. We use `if()` functions to set our `--color1` and `--color2` custom properties to different color values depending on the value of the `--scheme` custom property. We then use the `--color1` and `--color2` values in our `<article>` element's {{cssxref("color")}}, {{cssxref("border")}}, and {{cssxref("background-image")}} properties, and our `<aside>` element's `color` and `background-color` properties.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+body {
+  max-width: 700px;
+  margin: if(
+    media(width < 700px): 0 auto;
+    else: 20px auto 0;
+  );
+}
+```
+
+Next up, we set the `--scheme` custom property to a different value on our `<article>` element depending on the `class` name set on it by our JavaScript when a new value is selected in our `<select>` element. You'll see the significance of this in the next CSS block.
```
estelle:
```suggestion
We then set the `--scheme` custom property to match the `<article>` element's `class` name. The class will be by our JavaScript when a new value is selected in our `<select>` element. You'll see the significance of the custom element value in the next CSS block.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,621 @@
...
+}
+```
+
+```css-nolint live-sample___color-scheme
+body {
+  max-width: 700px;
+  margin: if(
+    media(width < 700px): 0 auto;
+    else: 20px auto 0;
+  );
```
estelle: This could also be:

```
  margin: if(
    media(width < 700px): 0;
    else: 20px;
  ) auto 0;
```
This suggestion is way more complex, but if you wanted to give an example of including an `if()` for a single component in the shorthand property, this is an opportunity for that.

estelle (commented):
Looks good.
Two main issues:

1. an `<if-condition>` is either an `<if-test>` or `else`. This means we should be using `<if-test>` when we mean an if-condition that isn't an else. Also need to add the hyphen after the `if-` in all cases.

2. LCH is supported in all browsers, so is supported in all browsers that support `if()`. We should change the LCH to another property that doesn't have better support than if, or change occurrences to media or style queries.

files/en-us/web/css/if/index.md
chrisdavidmills: @tabatkins, OK, fair enough. I'll just delete the example and the preceding line of text.

files/en-us/web/css/if/index.md
chrisdavidmills: I like @estelle's text, and have used it, with a few tweaks, such as fixing up the links.

files/en-us/web/css/if/index.md
chrisdavidmills: It is really hard to come up with a decent example for this. I thought about it a bit more, and eventually decided on using the `:buffering` pseudo-class, which doesn't currently have great cross-browser support:

```css
margin-top: if(
  supports(selector(:buffering)): 1em;
  else: initial;
)
```

I thought maybe you could use `:buffering` to selectively display a "Buffering" message on a media element, and then use the above if() structure to apply some extra space to display such messages in browsers that support the selector. Or something like that?

files/en-us/web/css/if/index.md
chrisdavidmills: Yup, these all look great and are valid. I've used your suggestion; nice work!

files/en-us/web/css/if/index.md
chrisdavidmills: I've updated this bit, as I think it works nicely. Thanks.

I've also gone through the entire article to make sure `<if-test>` is used to describe the if test types, not "if condition" or whatever else.

files/en-us/web/css/if/index.md
chrisdavidmills: Yes. I've tested an example and added it to the "Nesting if() functions" section, as I think it fits well there.

Interestingly enough, it works fine, but the Chrome devtools can't work out how to display the syntax properly. I might file a chrome bug about it.

files/en-us/web/css/if/index.md
chrisdavidmills: I've moved it to here, as I think it works better.

files/en-us/web/css/if/index.md
chrisdavidmills: Done, and done.

Well, I removed all instances of `to left` except the one in the final example, as in that case I wanted the gradient to be horizontal.

files/en-us/web/css/if/index.md
chrisdavidmills: done

files/en-us/web/css/if/index.md
chrisdavidmills: We removed this one altogether, based on @tabatkins's earlier comment.

files/en-us/web/css/if/index.md
chrisdavidmills: Done. Like it.

files/en-us/web/css/if/index.md
chrisdavidmills: Useful pro tip! I don't particularly want to change all those color values.

files/en-us/web/css/if/index.md
chrisdavidmills: Yup, I totally agree. I've replaced the inner `if()`s with `prefers-color-scheme` media query tests, to return appropriate text colors depending on whether the user prefers a light or dark color scheme.

files/en-us/web/css/if/index.md
chrisdavidmills: I'm not sure what though. In this case, I want to include a feature query example because I want to include one of each. But I'm not sure what feature query would be better. Pretty much anything you put will become better supported over time.

files/en-us/web/css/if/index.md
chrisdavidmills: Yup, I like this. Nice to show one of these examples in a live example. I've added it in, and updated the description to suit.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+/* Single <if-test> */
+if(style(--scheme: dark): #eee;)
+if(media(print): #000;)
+if(media(width > 700px): 0 auto;)
+if(supports(color: lch(7.1% 60.23 300.16)): lch(7.1% 60.23 300.16);)
+
+/* <if-test> with else */
+if(style(--size: 2xl): 1em; else: 0.25em;)
+if(media(print): white; else: black;)
+if(media(width < 700px): 0 auto ; else: 20px auto)
```
estelle:
```suggestion
if(media(width < 700px): 0 auto; else: 20px auto)
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+div {
+  background-image: if(
+    style(--scheme: ice): linear-gradient(#caf0f8, white, #caf0f8);
+    style(--scheme: fire): linear-gradient(#ffc971, white, #ffc971);
+    else: none;
+  );
+}
+```
+
+In this case, we are setting a different {{cssxref("background-image")}} {{cssxref("linear-gradient()")}} value on {{htmlelement("div")}} elements depending on whether a `--scheme` [custom property](/en-US/docs/Web/CSS/--*) is set to `ice` or `fire`. If neither value is set, the `else` value comes into play, and the `background-image` property is set to `none`.
```
estelle:
```suggestion
In this case, we are setting a different {{cssxref("linear-gradient()")}} values as the {{cssxref("background-image")}} on {{htmlelement("div")}} elements, depending on whether a `--scheme` [custom property](/en-US/docs/Web/CSS/--*) is set to `ice` or `fire`. If `--scheme` doesn't exist, or it exists and is set to any other value, the `else` value comes into play, and the `background-image` property is set to `none`.
```
First suggestion is just so the two links aren't next to each other without content between them (cognitive: clearer that they are separate links)
second suggestion is to kind of indicate that a non-existent custom prop won't invalidate the function.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+
+> [!NOTE]
+> Each condition must be separated from its associated value with a colon, and each `<if-condition> : <value>` pair must be separated with a semi-colon. The semi-colon is optional for the last `<if-condition> : <value>` pair.
+
+> [!WARNING]
+> There must be no space between the `if` and the opening parenthesis (`(`). If there is, the whole declaration is invalid.
+
+If a single `<if-condition>` or `<value>` is invalid, it does not invalidate the entire `if()` function; instead, the parser moves on to the next `<if-condition> : <value>` pair.
```
estelle:
```suggestion
If a single `<if-condition>` or `<value>` is invalid, it does not invalidate the entire `if()` function; instead, the parser moves on to the next `<if-condition> : <value>` pair. If no `<if-condition>` nor `<value>` valid, the function returns a guaranteed-invalid.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+
+### Frequency and position of `else : <value>` pairs
+
+You can include multiple `else : <value>` pairs inside an `if()` function, in any position. However, in most cases, a single `else : <value>` pair at the end of the semi-colon-separated list is used to provide the default value that is always returned if none of the `<if-test>`s evaluate to true.
+
+If you include an `else : <value>` pair before any `<if-test> : <value>` pairs, the later conditions are not evaluated because the `else` always evaluates to `true`. The following `if()` therefore always returns `none`, and the two `<if-test> : <value>` pairs are never evaluated:
```
estelle:
```suggestion
If you include an `else : <value>` pair before any `<if-test> : <value>` pairs, the later conditions are not evaluated because `else` evaluates to `true`. The following `if()` therefore always returns `none`, and the two `<if-test> : <value>` pairs are never evaluated:
```
If you don't go with this edit, we should still get rid of "the" unless you add "keyword"
should "later" be "latter"?

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+div {
+  background-image: if(
+    else: none;
+    style(--scheme: ice): linear-gradient(#caf0f8, white, #caf0f8);
+    style(--scheme: fire): linear-gradient(#ffc971, white, #ffc971)
+  );
+}
+```
+
+One circumstance in which you might want to put an `else : <value>` in a position other than the end of the values list is when a value is not behaving as expected, and you are trying to debug it. In the following example, we are trying to work out whether the first `<if-test> : <value>` pair is working properly. If it isn't, the `else : <value>` pair returns a value of `url("debug.png")` to display an image indicating that the first `<if-test> : <value>` pair needs fixing. The last two `<if-test> : <value>` pairs are again never evaluated.
```
estelle:
```suggestion
Debugging a value is not behaving as expected is a case in which you might want to put an `else : <value>` in a position other than the end of the values list. In the following example, we are trying to work out whether the first `<if-test> : <value>` pair is working properly. If it isn't, the `else : <value>` pair returns a value of `url("debug.png")` to display an image indicating that the first `<if-test> : <value>` pair needs fixing. The last two `<if-test> : <value>` pairs are again never evaluated.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+);
+
+background-color: if(
+  not style(--scheme: light): black;
+);
+```
+
+A `@container` query does have some advantages — you can only set single property values at a time with `if()` style queries, whereas `@container` queries can be used to conditionally apply whole sets of rules. The two approaches are complementary, and have different uses.
+
+Note that container style queries currently don't support regular CSS properties, just CSS custom properties. So for example, the following won't work:
```
estelle:
```suggestion
Note that container style queries currently don't support regular CSS properties, just CSS custom properties. For example, the following won't work:
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+
+```css-nolint
```
estelle:
```suggestion
```css-nolint example-bad
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+> [!NOTE]
+> Remember to include the `else` condition. In `if()`-supporting browsers, if no `else` value were included and `--size` was not equal to `2xl`, the padding would be set to `initial`.
+
+### Whole and partial values
+
+An `if()` function can be set as the value of any CSS property, but it can also be used to determine parts of property values. For example, the following sets a different {{cssxref("border-color")}} inside a {{cssxref("border")}} shorthand property, depending on whether {{cssxref("color_value/lch()")}} colors are supported:
+
+```css-nolint
+border: if(
+  supports(color: lch(77.7% 0 0)): 3px solid lch(77.7% 0 0);
```
estelle, line 294: Let's try to remove the LCH for supports other than in the features area, where `supports()` has to be used.

```suggestion
An `if()` function can be set as the value of any CSS property, but it can also be used to determine parts of property values. For example, the following sets a different {{cssxref("border-width")}} inside a {{cssxref("border")}} shorthand property, depending on the width of the viewport:

```css-nolint
border: if(
  (width > 620px): 6px solid #c0c0c0;
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+);
+```
+
+You can also use the `if()` function to set just one component in a multi-component shorthand. Here we use the `if()` function to determine the {{cssxref("border-color")}} component only:
+
+```css-nolint
+border: 3px solid
+  if(
+    supports(color: lch(77.7% 0 0)): 3px solid lch(77.7% 0 0); else: #c0c0c0;
+  );
```
estelle:
```suggestion
You can also use the `if()` function to set just one component in a multi-component shorthand. Here we use the `if()` function to determine the {{cssxref("border-width")}} component only:

```css-nolint
border: if((width > 620px): 6px; else: 3px;)  solid #c0c0c0;
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+```
+
+### Nesting if() functions
+
+Because an `if()` function can take the place of whole property values or individual components, it is possible to nest multiple `if()` functions, and nest `if()` functions inside other functions such as {{cssxref("calc()")}}.
```
estelle:
```suggestion
Because an `if()` function can take the place of whole property values or individual components, it is possible to nest `if()` functions within other `if()` functions, and to nest `if()` functions inside other functions such as {{cssxref("calc()")}}.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+
+## Examples
+
+### Basic `if()` usage
+
+In this example, we'll show basic usage of each of the three types of `<if-test>`.
+
+#### HTML
+
+Our HTML features a {{htmlelement("section")}} element with two {{htmlelement("article")}} elements inside it, containing `<h2>` [headings](/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements). The `<section>` has a custom property set on it inside a [`style`](/en-US/docs/Web/HTML/Reference/Global_attributes/style) attribute — `--show-apple:true` — which we use later on to conditionally set a property value.
```
estelle:
```suggestion
Our HTML features a {{htmlelement("section")}} element with two {{htmlelement("article")}} elements inside it, containing `<h2>` [headings](/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements). The `<section>` has a custom property set on it inside its [`style`](/en-US/docs/Web/HTML/Reference/Global_attributes/style) attribute — `--show-apple:true` — which we use later on to conditionally set a property value.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+    color: if(
+    supports(color: lch(29.57% 43.25 344.44)): lch(29.57% 43.25 344.44);
+    else: #792359;
+  )
+}
+```
+
+#### Result
+
+This demo will render as follows:
```
estelle:
```suggestion
```
or "This example renders as follows:"
(no "will")

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+
+{{EmbedLiveSample("basic", "100%", "240")}}
+
+Note how the styling is applied. You can test out the conditional styling pretty easily for at least the first two `if()` queries by modifying the rendered demo using your browser's devtools:
```
estelle:
```suggestion
Note how the styling is applied. Test out the conditional styling for the first two `if()` queries by modifying the rendered demo using your browser's devtools:
```
getting rid of "easy"

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+
+- Remove the `<section>` element's `style` attribute and note how the apple emojis are no longer rendered.
+- Change the `height` attribute of the embedding `<iframe>` to `1200px`. This will change document orientation from landscape to portrait. Note how the layout changes as a result.
```
estelle:
```suggestion
- Change the `height` attribute of the embedding `<iframe>` to `1200px`. This will change the orientation from landscape to portrait. Note how the layout changes as a result.
```

estelle (commented):
heading to the airport, so will finish later.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
+- [CSS values and units module](/en-US/docs/Web/CSS/CSS_values_and_units)
```
estelle:
```suggestion
```
moved down

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+- [Container style queries](/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_style_queries)
+- [Media queries](/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries)
+- [Feature queries](/en-US/docs/Web/CSS/CSS_conditional_rules/Using_feature_queries)
```
estelle, line 609:
```suggestion
- [Feature queries](/en-US/docs/Web/CSS/CSS_conditional_rules/Using_feature_queries)
- [CSS values and units module](/en-US/docs/Web/CSS/CSS_values_and_units)
```
i think modules may go last.

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+```html live-sample___basic
+<section style="--show-apple:true">
+  <article><h2>First article</h2></article>
+  <article><h2>Second article</h2></article>
+</section>
+```
+
+#### CSS
+
+In our CSS, we first target the `<section>` element, laying it out with [flexbox](/en-US/docs/Web/CSS/CSS_flexible_box_layout) and setting a {{cssxref("gap")}} between the two child `<article>` elements. We then use an `if()` function with a [media query](/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries) `<if-test>` to set the value of the {{cssxref("flex-direction")}} property to `row` if the document is in landscape orientation, or `column` if it is in portrait orientation. This lays out the `article` elements side-by-side in wide screens, and top-to-bottom in narrow screens.
```
estelle:
```suggestion
In our CSS, we first target the `<section>` element, laying it out with [flexbox](/en-US/docs/Web/CSS/CSS_flexible_box_layout) and setting a {{cssxref("gap")}} between the two child `<article>` elements. We then use an `if()` function with an [`orientation`](/en-US/docs/Web/CSS/@media/orientation) media query `<if-test>` to set the value of the {{cssxref("flex-direction")}} property to `row` if the document is in landscape orientation, or `column` if it is in portrait orientation. This lays out the `article` elements side-by-side in wide screens, and top-to-bottom in narrow screens.
```
we do have 5 other links to using media queries, so think it might be helpful to link to orientation once
likely we should use "inline axis" and "block direction" instead of side-by-side and top-to-bottom, but i find what you have easier to read.... but if you want to be more pendantic than me ;)

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+{{EmbedLiveSample("basic", "100%", "240")}}
+
+Note how the styling is applied. You can test out the conditional styling pretty easily for at least the first two `if()` queries by modifying the rendered demo using your browser's devtools:
+
+- Remove the `<section>` element's `style` attribute and note how the apple emojis are no longer rendered.
+- Change the `height` attribute of the embedding `<iframe>` to `1200px`. This will change document orientation from landscape to portrait. Note how the layout changes as a result.
+
+### Controlling a color scheme with `if()`
+
+This demo shows how you can start to have some real fun with CSS `if()` functions. Amongst other things, we use `if()` functions to conditionally set the values of a couple of custom properties to control an entire color scheme.
```
estelle: totally optional edit as what you had is correct. my suggestion is to add extra marketing/excitement to the intro, which may be a no-no

```suggestion
This demo shows how you can start to have some real fun with CSS `if()` functions. Amongst other things, we use `if()` functions to conditionally set the values of some custom properties, enabling us to control the entire color scheme!
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+
+#### HTML
+
+Our HTML contains an {{htmlelement("article")}} element with some content inside it — a top-level heading, a couple of {{htmlelement("p")}} elements, and an {{htmlelement("aside")}}. Below that we include a {{htmlelement("form")}} containing a {{htmlelement("select")}} drop-down that allows you to select a color scheme.
```
estelle:
```suggestion
Our HTML contains an {{htmlelement("article")}} element with some content inside it — a top-level heading, a couple of {{htmlelement("p")}} elements, and an {{htmlelement("aside")}}. We also include a {{htmlelement("form")}} containing a {{htmlelement("select")}} drop-down that enables selecting a color scheme.
```
getting rid of "below" per an mdn PR from a few months back

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+
+```html-nolint live-sample___color-scheme
```
estelle, line 437: is the nolint so ipsum isn't spell-checked?

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+    <option value="">Default</option>
+    <option value="ice">Ice</option>
+    <option value="fire">Fire</option>
+  </select>
+</form>
+```
+
+### JavaScript
+
+Next, let's look at our minimal JavaScript. This adds a [`change`](/en-US/docs/Web/API/HTMLElement/change_event) event listener to the `<select>` element. When a new value is selected, our script sets the `<article>` element's [`class`](/en-US/docs/Web/HTML/Reference/Global_attributes/class) attribute to equal that value.
```
estelle:
```suggestion
Our JavaScript adds a [`change`](/en-US/docs/Web/API/HTMLElement/change_event) event listener to the `<select>` element. When a new value is selected, our script sets the `<article>` element's [`class`](/en-US/docs/Web/HTML/Reference/Global_attributes/class) attribute to equal that value.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+const selectElem = document.querySelector("select");
+
+selectElem.addEventListener("change", () => {
+  articleElem.className = selectElem.value;
+});
+```
+
+### CSS
+
+In our CSS, we start by giving the `<body>` element a {{cssxref("max-width")}} of `700px`. We then center it on the screen using `auto` {{cssxref("margin")}} values. However, we use an `if()` function with a media query `<if-test>` to set the top margin component inside the `margin` shorthand to `0` if the viewport width is less than `700px`, and `20px` if it is wider. This means that on wide screens, we get a bit of margin at the top of the content, but it is removed on narrow screens, where it looks a bit weird.
```
estelle:
```suggestion
In our CSS, we give the `<body>` element a {{cssxref("max-width")}} of `700px` and center it using `auto` {{cssxref("margin")}} values. However, we use an `if()` function with a media query `<if-test>` to set the {{cssxref("margin-top")}} component inside the `margin` shorthand to `0` if the viewport's width is less than `700px`, and `20px` if it is wider. This means that on wide screens we get a bit of margin at the top of the content, but this is removed on narrow screens, where it looks a bit weird.
```

files/en-us/web/css/if/index.md
```diff
@@ -0,0 +1,612 @@
...
+.ice {
+  --scheme: ice;
+}
+
+.fire {
+  --scheme: fire;
+}
+```
+
+Here we can start to see the real power of CSS `if()` functions combined with custom properties. We use `if()` functions to set our `--color1` and `--color2` custom properties to different color values depending on the value of the `--scheme` custom property. We then use the `--color1` and `--color2` values in our `<article>` element's {{cssxref("color")}}, {{cssxref("border")}}, and {{cssxref("background-image")}} properties, and our `<aside>` element's `color` and `background-color` properties.
```
estelle:
```suggestion
We can see the real power of CSS `if()` functions when we combine them with custom properties. Here we use `if()` functions to set our `--color1` and `--color2` custom properties to different color values depending on the value of the `--scheme` custom property. We then use the `--color1` and `--color2` values in our `<article>` element's {{cssxref("color")}}, {{cssxref("border")}}, and {{cssxref("background-image")}} properties, and our `<aside>` element's `color` and `background-color` properties.
```

estelle (commented):
a couple of the edits are suggetions to get the reader excited. all suggestions are minor.

files/en-us/web/css/if/index.md
chrisdavidmills: I think it was just because prettier made the code layout a bit weird.

files/en-us/web/css/if/index.md
chrisdavidmills: This was mostly good; I tightened it up a bit to:

> Because an `if()` function can take the place of whole property values or individual components, it is possible to nest `if()` functions within other `if()` functions, and inside other functions such as {{cssxref("calc()")}}.

files/en-us/web/css/if/index.md
chrisdavidmills: I can see what you're trying to do here, but the section doesn't really make sense if you change the affected component. The wording was a little borked anyway, so I ended updating it to:

> An `if()` function can be set as the value of any CSS property, but it can also be used to determine parts of property values. For example, the following sets a different {{cssxref("border-color")}} inside a {{cssxref("border")}} shorthand property, depending on whether {{cssxref("color_value/lch()")}} colors are supported:
>
>```css-nolint
> border: if(
>   supports(color: lch(77.7% 0 0)): 3px solid lch(77.7% 0 0);
>  else: 3px solid #c0c0c0;
> );
>```
>
> However, we could use the `if()` function to determine the {{cssxref("border-color")}} component only:

> ```css-nolint
> border: 3px solid
>   if(
>     supports(color: lch(77.7% 0 0)): lch(77.7% 0 0); else: #c0c0c0;
>   );
> ```

files/en-us/web/css/if/index.md
chrisdavidmills: See below comment. I'd rather use the color

files/en-us/web/css/if/index.md
chrisdavidmills: Good call; added

files/en-us/web/css/if/index.md
chrisdavidmills: I played with this a bit, and ended up with:

> Debugging a value that is not behaving as expected is one case where you might want to put an `else : <value>` in a position other than the end of the value list. In the following example, we are trying to work out whether the first `<if-test> : <value>` pair is working properly. If it isn't, the `else : <value>` pair returns a value of `url("debug.png")` to display an image indicating that the first `<if-test> : <value>` pair needs fixing. The last two `<if-test> : <value>` pairs are again never evaluated.

files/en-us/web/css/if/index.md
chrisdavidmills: I just removed the "the". I think it sounds better and more definite with "always" included. "latter" is not quite right because I want to include all of the values after the `else : <value>` pair; I think "latter" implies just the ones at the end.

I thought about it a bit, and went with "the conditions that follow it":

> If you include an `else : <value>` pair before any `<if-test> : <value>` pairs, the conditions that follow it are not evaluated because `else` always evaluates to `true`. The following `if()` therefore always returns `none`, and the two `<if-test> : <value>` pairs are never evaluated:

WDYT?

files/en-us/web/css/if/index.md
chrisdavidmills: Mostly used, with a couple of tweaks:

> If a single `<if-condition>` or `<value>` is invalid, it does not invalidate the entire `if()` function; instead, the parser moves on to the next `<if-condition> : <value>` pair. If no `<if-condition>` nor `<value>` is valid, the function returns {{glossary("guaranteed_invalid_value", "guaranteed-invalid")}}.

files/en-us/web/css/if/index.md
chrisdavidmills: Sounds reasonable. I used it with a couple of tweaks:

> In this case, we are setting a different {{cssxref("linear-gradient()")}} as the {{cssxref("background-image")}} on {{htmlelement("div")}} elements, depending on whether a `--scheme` [custom property](/en-US/docs/Web/CSS/--*) is set to `ice` or `fire`. If `--scheme` doesn't exist, or it exists and is set to any other value, the `else` value comes into play, and the `background-image` property is set to `none`.

estelle (approved):

PR #39527, opened 2025-05-14
Set Vary header in the Fetch metadata example

wbamberg:
Set the `Vary` header in the response, in the fetch metadata example, as reported in https://front-end.social/@openwebdocs/114506561385552270.

We probably ought to have a proper guide to fetch metadata on MDN.

@jub0bs , thank you for finding this! Would you like to review?

### DIFF
--- a/files/en-us/web/security/attacks/xs-leaks/index.md
+++ b/files/en-us/web/security/attacks/xs-leaks/index.md
@@ -186,6 +186,7 @@ function isAllowed(req) {
 app.get("/admin", (req, res) => {
   if (isAllowed(req)) {
     // Respond with the admin page if the user is admin
+    res.setHeader("Vary", "sec-fetch-site, sec-fetch-mode");
     getAdminPage(req, res);
   } else {
     res.status(404).send("Not found.");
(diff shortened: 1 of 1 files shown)


files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -195,6 +196,8 @@ app.get("/admin", (req, res) => {
 
 Since the attacker's request is cross-site and is not a navigation, then this server always returns an error for it, whether the user is signed in or not.
 
+Note that we also send the {{httpheader("Vary")}} response header. This ensures that if the response is cached, the cached response will only be given to requests with the same fetch metadata header values.
```
wbamberg, line 199:
```suggestion
Note that we also send the {{httpheader("Vary")}} response header. This ensures that if the response is cached, the cached response will only be given to requests with the same values for the Fetch metadata headers we are using.
```

jub0bs:
@wbamberg I believe that the Vary header should be set in both branches (i.e. regardless of whether `isAllowed(req)` is truthy):

```diff
  app.get("/admin", (req, res) => {
+   res.setHeader("Vary", "sec-fetch-site, sec-fetch-mode");
    if (isAllowed(req)) {
```

Some caching intermediaries may indeed cache 4xx responses:

> Most commonly, caches store the successful result of a retrieval request: i.e., a 200 (OK) response to a GET request [...] **However, it is also possible to store redirects, negative results (e.g., 404 (Not Found))** [...]

(from [RFC 9111](https://www.rfc-editor.org/rfc/rfc9111.html#name-overview-of-cache-operation), my emphasis)

files/en-us/web/security/attacks/xs-leaks/index.md
```diff
@@ -186,6 +186,7 @@ function isAllowed(req) {
 app.get("/admin", (req, res) => {
   if (isAllowed(req)) {
     // Respond with the admin page if the user is admin
+    res.setHeader("Vary", "sec-fetch-site, sec-fetch-mode");
```
wbamberg, line 189:
```suggestion
app.get("/admin", (req, res) => {
  res.setHeader("Vary", "sec-fetch-site, sec-fetch-mode");
  if (isAllowed(req)) {
    // Respond with the admin page if the user is admin
```

jub0bs:
LGTM now 👍

chrisdavidmills (approved):
LGTM!

PR #39529, opened 2025-05-15
Update index.md

jabcreations:
Updated that the error key will always trigger the error event handler.


estelle (approved):
Thank you. Also, congratulations on your first merged MDN content PR. Welcome to the team!

PR #39534, opened 2025-05-15
Fix typo

seppewyns:
### Description

Fixes a typo in the definition of a trap on the Proxy reference page ("define" -> "defines").


Josh-Cena (approved):
Nice catch, thanks

PR #39538, opened 2025-05-15
Modify the code example to actually use Promise.try

PMudra:
### Description

The async/await example usage of `Promise.try` did not actually use the function. With this change, the goal of the code example should now be correct.


Josh-Cena:
In async/await, you do not need to use `Promise.try`. You only use `Promise.try` when you have to use promise methods.

PR #39545, opened 2025-05-16
Fix content issues

Josh-Cena:
Fix some references to non-existent images; fix bad macro; upgrade link from HTTP to HTTPS


bsmth (approved):
Great, ty 👍🏻

PR #39547, opened 2025-05-16
fix(workflows): disable cache in privileged workflows

caugner:
### Description

Disables the cache (used by `actions/setup-node`) in privileged workflows.

### Motivation

Limits the impact of less-privileged workflows that may be vulnerable to code injection.


bsmth (approved):
tnx

PR #39556, opened 2025-05-18
doc[iframe/index.md]: change `Window.showModalDialog()` to `HTMLDialogElement.showModal()`

samuel871211:
### Motivation

The link of `Window.showModalDialog()` is currently points to [HTMLDialogElement.showModal()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal)


chrisdavidmills (approved):
Thanks for updating this, @samuel871211. LGTM!

PR #39559, opened 2025-05-18
Specify which credientials header value is default

DanKaplanSES:
- Technically this is redundant because the example that follows mentions it in a comment, but I think the typical reader will expect to find that information here, not in the example.
- I tried to make this change similar in format to https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#including_credentials

### DIFF
--- a/files/en-us/web/api/request/credentials/index.md
+++ b/files/en-us/web/api/request/credentials/index.md
@@ -20,8 +20,9 @@ A string with one of the following values:
 
 - `omit`
   - : Never send credentials in the request or include credentials in the response.
-- `same-origin` (the default)
+- `same-origin`
   - : Only send and include credentials for same-origin requests.
+    This is the default.
 - `include`
   - : Always include credentials, even for cross-origin requests.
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/request/credentials/index.md
```diff
@@ -20,7 +20,7 @@ A string with one of the following values:
 
 - `omit`
   - : Never send credentials in the request or include credentials in the response.
-- `same-origin`
+- `same-origin` (the default)
   - : Only send and include credentials for same-origin requests.
```
hamishwillee, line 24: We really should have an icon for defaults to highlight them better. 
Thanks for this. I've moved this down into text because if you add it like this it changes the implicit url of the item (created by the fact this is a definition list). That makes break any links here, and makes it harder to just guess the link name (`#same_origin`).


```suggestion
- `same-origin`
  - : Only send and include credentials for same-origin requests.
    This is the default.
```

hamishwillee (approved):
Thanks very much!

PR #39560, opened 2025-05-18
Add missing "to" in html basics course debugging walkthrough page

lacrosse1991:
added missing "to"

### Description

There was a missing "to" in this paragraph, before the word "inspect". 

"So how do you find markup errors? Later on we will show you how to find errors in HTML using a tool called the [HTML validator](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Debugging_HTML#html_validation), but first we will show you how inspect your HTML manually using a DOM inspector, and then explore what kinds of markup errors you might be looking for, and how the browser might interpret those."

### Motivation

I would like to make the MDN HTML course easier to understand by addressing any grammatical errors that I might come across. 

### Additional details

N/A

N/A


hamishwillee (approved):
Good catch. Thanks, and welcome to MDN!

PR #39569, opened 2025-05-19
Edit credentials summary similar to spec

### LINKED ISSUE #39561: `fetch()`: What cookies are sent when `"credentials": "include"` and `"cookies"` are set?
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#including_credentials

I've been assuming this is how the `"credentials": "include"` and `"credentials": "same-origin"` work: 

The browser stores a list of cookies, some of them related to credentials, some of them not. When the browser executes a `fetch` with `"credentials": "include"` or `"credentials": "same-origin"`, the browser sends the _credential-related cookies_ to the server behind the scenes. To send all stored cookies, the difference (i.e., `all cookies - credential related cookies`) must be explicitly added to a `cookies` header in the `fetch()` call. 

1. Is this accurate? Or does `"credentials": "include"` and `"credentials": "same-origin"` send all cookies, like this comment says? 
1. If `"credentials": "include"` or `"credentials": "same-origin"` and `cookies` contain overlapping keys, which one is sent to the server? 
    1. If both are sent to the server, which one has precedence, the `credentials` or the `cookies`?

I was reading [this github comment](https://github.com/node-fetch/node-fetch/issues/49#issuecomment-145006792), and couldn't confirm this detail (emphasis mine):

> You should add them manually as headers. You can in fact use the same code for server and for client, because **on client the Cookie headers are ignored and credentials: 'include' works instead**.
> 
> ```
> fetch('/', {
>   credentials: 'include',
>   headers: {'Cookie': 'foo=bar;bar=foo'}
> })
> ```

If this is true, can it be mentioned somewhere in this article?

If I'm requesting information that seems more appropriate for reference documentation, maybe it should go here instead: https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials 

This may be the appropriate content to modify:

> Credentials are cookies, [TLS](https://developer.mozilla.org/en-US/docs/Glossary/TLS) client certificates, or authentication headers containing a username and password.

### MDN metadata

<details>

* Folder: `en-us/web/api/fetch_api/using_fetch`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/fetch_api/using_fetch/index.md
* Last commit: https://github.com/mdn/content/commit/759102220c07fb140b3e06971cd5981d8f0f134f
* Document last modified: 2025-04-28T15:45:41.000Z

</details>

DanKaplanSES:
Spec:

> Credentials are HTTP cookies, TLS client certificates, and authentication entries (for HTTP authentication).

Closes #39561


files/en-us/web/api/fetch_api/using_fetch/index.md
```diff
@@ -205,7 +205,7 @@ See the reference documentation for {{domxref("RequestInit", "", "mode")}} for m
 
 ### Including credentials
 
-Credentials are cookies, {{glossary("TLS")}} client certificates, or authentication headers containing a username and password.
+Credentials are cookies, {{glossary("TLS")}} client certificates, and authentication headers.
 
 To control whether or not the browser sends credentials, as well as whether the browser respects any **`Set-Cookie`** response headers, set the `credentials` option, which can take one of the following three values:
```
wbamberg, line 210: I think if we're going to try to be more explicit here (and I think we should) then we should go all the way. Something like:

```suggestion
In the context of an HTTP request, a credential is an extra piece of data sent along with the request, that the server may use to authenticate the user. In a fetch request, all the following items are considered to be credentials:

- HTTP cookies
- {{glossary("TLS")}} client certificates
- The {{httpheader("Authorization")}} and {{httpheader("Proxy-Authorization")}} headers.

By default, credentials are only included in same-origin requests. To customize this behavior, as well as to control whether the browser respects any **`Set-Cookie`** response headers, set the [`credentials`](/en-US/docs/Web/API/RequestInit#credentials) option, which can take one of the following three values:
```

HOWEVER we also have the problem that the definition of "credentials" is repeated in at least three places: here, the page for [`Request.credentials`](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials), and the [`RequestInit`](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#credentials) page. When the definition is short and a little vague, as it currently is, it's OK to repeat, but if we want to be definitive we should only have one copy, and link from the other two. I'm not quite sure which place should have the definition (except it should not be `Request.credentials`). I feel maybe this guide page is best?

wbamberg: @hamishwillee , I would also value your input here.

files/en-us/web/api/fetch_api/using_fetch/index.md
```diff
@@ -205,9 +205,13 @@ See the reference documentation for {{domxref("RequestInit", "", "mode")}} for m
 
 ### Including credentials
 
-Credentials are cookies, {{glossary("TLS")}} client certificates, or authentication headers containing a username and password.
+In the context of an HTTP request, a credential is an extra piece of data sent along with the request, that the server may use to authenticate the user. All the following items are considered to be credentials:
```
DanKaplanSES: @wbamberg This is a huge improvement. I also think the definitive guide should be in this file because it's easy to digest and has a high likelihood of being applicable to your first usage.

I had some edits to suggest, but I'm mostly interested in your thoughts about this one:


```suggestion
In the context of the Fetch API, a credential is an extra piece of data sent along with the request that the server may use to authenticate the user. All the following items are considered to be credentials:
```

I'm assuming that this definition is specific to the `fetch` API, not all of HTTP. e.g., a `curl` command with a `credentials` header wouldn't behave the way this section describes, would it?

files/en-us/web/api/fetch_api/using_fetch/index.md
```diff
@@ -205,9 +205,13 @@ See the reference documentation for {{domxref("RequestInit", "", "mode")}} for m
...
 
-To control whether or not the browser sends credentials, as well as whether the browser respects any **`Set-Cookie`** response headers, set the `credentials` option, which can take one of the following three values:
+- HTTP cookies
+- {{glossary("TLS")}} client certificates
+- The {{httpheader("Authorization")}} and {{httpheader("Proxy-Authorization")}} headers.
+
+By default, credentials are only included in same-origin requests. To customize this behavior, as well as to control whether the browser respects any **`Set-Cookie`** response headers, set the [`credentials`](/en-US/docs/Web/API/RequestInit#credentials) option, which can take one of the following three values:
```
DanKaplanSES, line 214: I think the `Set-Cookie` response is handled differently for each of these bullet points. If I'm understanding that correctly, it may make sense to add the specifics in each one.

files/en-us/web/api/fetch_api/using_fetch/index.md
wbamberg: Sure, I'm happy with this, although "fetch" here should not get code formatting because it isn't an identifier in this context.

files/en-us/web/api/fetch_api/using_fetch/index.md
wbamberg: I suppose "include credentials in the response" in bullet 1 and "send **and include** credentials" in bullet 2 are clumsy ways of saying "include the `Set-Cookie` response header" (since it doesn't make sense to include, say, TLS client certificate in a response. Similarly in the later bullets about "the response, **including credentials**, will be delivered to the caller".

It would be better to be clear that this is (AFAICT) specifically about `Set-Cookie`.

It's a very complicated corner of the fetch API, and one that I tried to improve when I rewrote this doc, but there's definitely still room for improvement!

files/en-us/web/api/fetch_api/using_fetch/index.md
DanKaplanSES: @wbamberg I'm not sure what you mean by "an identifier in the context." Perhaps you could word it a different way?

FWIW, the reason I used code formatting was to emphasize it applies to the fetch API, not HTTP in general, and I think the word "fetch" is used colloquially to mean, "A client initiates a call to a server." I didn't want a reader to mistakenly interpret it that way.

wbamberg: I mean, there's no code entity which we'd refer to as `fetch`. There's a `fetch()` method, and we would use code formatting it we wanted to say something like "When you call `fetch()`...". When we talk about "the fetch API" we're talking about an abstraction. In the rest of this guide, and in the overview page, we actually capitalize it, like "the Fetch API".

If you wanted, you could say "In the context of the Fetch API, ...".

wbamberg (approved):
I think this is an improvement so I'm going to merge it.

https://github.com/mdn/content/pull/39569/files#r2099031115 suggests that there's some leftovers wrt `Set-Cookie` but let's do that in a follow-up if necessary.

PR #39576, opened 2025-05-19
Clarify use of prefers-reduced-motion: reduce in CSS example

### LINKED ISSUE #39543: prefers-reduced-motion media query check should use prefers-reduced-motion: reduce?
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion#toning_down_the_animation_scaling

Toning down the animation scaling

I noticed that on https://web.dev/articles/prefers-reduced-motion the example uses `prefers-reduced-motion: reduce` while on the MDN page it only uses `prefers-reduce-motion`. I think that might potentially be a mistake?

```
.animation {
  animation: pulse 1s linear infinite both;
  background-color: purple;
}

/* Tone down the animation to avoid vestibular motion triggers. */
@media (prefers-reduced-motion: reduce) {
  .animation {
    animation: dissolve 4s linear infinite both;
    background-color: green;
    text-decoration: overline;
  }
}

```

https://web.dev/articles/prefers-reduced-motion

_No response_

farizadam:
### Description

This PR updates the @media rule from prefers-reduced-motion to prefers-reduced-motion: reduce in the CSS example to explicitly match the widely used syntax and align with best practices.

### Motivation

The update addresses potential confusion for developers. Although @media (prefers-reduced-motion) works as a boolean, using the explicit : reduce value improves clarity and aligns with other documentation like [web.dev](https://web.dev/articles/prefers-reduced-motion).

Fixes #39543


chrisdavidmills (changes requested):
Hi @farizadam, and thank you for your contribution to MDN! I think the change you have made here is good, however, I wonder if you should go slightly further? The syntax example at the top of the page still shows

```
@media (prefers-reduced-motion) {
```

Maybe you could also change that to be consistent? 

Then, I noticed the place where it says "This keyword value evaluates as true." That makes sense to more experienced developers, but I thought less experienced developers might benefit from a more explicit description, along the lines of "The `reduce` keyword value evaluates as true, therefore `@media (prefers-reduced-motion)` is equivalent to `@media (prefers-reduced-motion: reduce)`."

What do you think?

farizadam:
@chrisdavidmills I've updated the PR to address both of your suggestions:

1. Changed the syntax example at the top to use `prefers-reduced-motion: reduce`
2. Added the explicit explanation about the equivalence between `@media (prefers-reduced-motion)` and `@media (prefers-reduced-motion: reduce)`

Thank you for your review feedback!

chrisdavidmills (approved):
Perfect, nice one @farizadam! I just made a minor formatting fix, but this looks ready to merge now.

Thanks again for your contribution!

PR #39578, opened 2025-05-20
add Forbids  value `auto`

zhangenming:
### Description

according to https://drafts.csswg.org/css-grid/#line-placement:~:text=In%20all%20the%20above%20productions%2C%20the%20%3Ccustom%2Dident%3E%20additionally%20excludes%20the%20keywords%20span%20and%20auto.


Josh-Cena:
You probably need to update the property pages too, like `grid-row-start`

Josh-Cena (approved):
Nice, thank you

PR #39581, opened 2025-05-20
Keep the sentence clear and direct.

ajuchacko:


Josh-Cena:
Hi, they mean the same thing and I don't think one is clearer than the other.

PR #39582, opened 2025-05-20
Update "Specifying a uniform font size for h1"

zcorpan:
See https://github.com/whatwg/html/pull/11102

### Description

Update wording to reflect that the spec has been changed.

### Motivation

Preparing to land the spec change. (This can be merged after the spec change has landed.)


chrisdavidmills (approved):
@zcorpan usually, we don't like putting time-dependent information in MDN content. But in this case, I don't see a better way of conveying what has happened.

I do think we should add a data point to the BCD to show what browsers have made this change, and when, otherwise people reading "browsers that implement the old context-dependent default style" might think "well, which browsers need this?"

Issue #37761, opened 2025-01-22
SVG: Contents of the style element needs CDATA encapsulation


TriMoon:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/style

### What specific section or headline is this issue about?

Description and Example

### What information was incorrect, unhelpful, or incomplete?

```html
<style>
...
</style>
```

### What did you expect to see?

```html
<style><![CDATA[
...
]]</style>
```

### Do you have any supporting links, references, or citations?

https://www.w3.org/TR/SVG11/styling.html#StyleElementExample

### Do you have anything more you want to share?

Without this `<![CDATA[`...`]]` encapsulation, your style rules won't be able to use the `<`, `>`, `&`, and `"` characters which are used for **selectors**, strings or content inside strings !
What's worse is that, without this encapsulation, your SVG will be rejected by FireFox's XML-parser (with error), maybe other browsers too have not tried.

Perhaps Firefox and/or some specification should explicitly allow usage of those characters inside the contents of the SVG `<style>` tag without the need to "escape" them using char-entities like `&amp;` for `&`...

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/svg/element/style`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/style
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/svg/element/style/index.md
* Last commit: https://github.com/mdn/content/commit/bc780f1de369fc7f14c73bbdccdeeb8e375cc93e
* Document last modified: 2024-12-22T22:57:14.000Z

</details>

Josh-Cena:
This is tricky. Our examples are always in HTML, and the `<svg>` is just an HTML exotic element. We don't show SVG as XML, and we can't because our live sample system doesn't support embedding XML.

TriMoon:
I'm unsure if you understood the problem properly...
You don't need to embed the SVG on the example page, but you need to add what i wrote to the example text for it to work properly as a standalone SVG when the user copy&pastes that text in a fresh file...

Josh-Cena:
CDATA is not a thing in HTML. If you use CDATA in an embedded HTML it will be treated as a comment.

TriMoon:
I have no words left to make my point, because you are looking at it from a different angle as i am...
Your angle could be solved by escaping etc but anyhow, i gave my feedback 😉

Josh-Cena:
[...]

PR #39598, opened 2025-05-21
Remove unnecessary pseudo-namespace declarations.

meiqinyan:
Given that the example is apparently embedded in HTML5, the `xmlns` attributes are useless.

### DIFF
--- a/files/en-us/web/svg/reference/element/stop/index.md
+++ b/files/en-us/web/svg/reference/element/stop/index.md
@@ -19,10 +19,7 @@ svg {
 ```
 
 ```html
-<svg
-  viewBox="0 0 10 10"
-  xmlns="http://www.w3.org/2000/svg"
-  xmlns:xlink="http://www.w3.org/1999/xlink">
+<svg viewBox="0 0 10 10">
   <defs>
     <linearGradient id="myGradient" gradientTransform="rotate(90)">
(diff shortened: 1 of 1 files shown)


pepelsbey:
We show SVG examples without any HTML context (even if we use HTML language tag), just as they are. Without the HTML context, SVG files will break (won’t render at all). It’s just safer to have `xmlns` attributes for all our readers’ use cases.

meiqinyan:
Readers should be familiar with their use cases and not cargo-cult based on what examples online show. Perhaps it would be more useful instead to add a note alongside the code, linking to a complete explanation of the requirements of XML namespacing outside of HTML5.

For these snippets, the namespace declarations (and XML declarations and document type declarations, for that matter) are simply visual noise that detract from the examples being demonstrated.

Josh-Cena:
FYI, we'd like to remove the HTML embedding altogether and make sure that all SVG examples are well-formed standalone XML: https://github.com/mdn/content/issues/37761

dipikabh:
Thanks @pepelsbey and @Josh-Cena for the added context.

@meiqinyan I'm going to close this PR for now. The need for cleanup in all code snippets (on this and other similar pages) will depend on what we're able to support on the platform eventually. Thanks for highlighting the issue though!

PR #39610, opened 2025-05-22
Remove empty learn modules

chrisdavidmills:
### Description

When I reorganized the MDN learn area, I included the [Transform and animate CSS](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Transform_animate) and [Security and privacy](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Security_privacy) modules to make the structure as consistent as possible with the [MDN Curriculum](https://developer.mozilla.org/en-US/curriculum/).

However, they are not real modules — they are just landing page skeletons — so they provide no value over the equivalent curriculum pages.

We have therefore decided to remove them for now.


files/en-us/_redirects.txt
```diff
@@ -5825,6 +5825,8 @@
 /en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills:_HTML5_controls	/en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills/Input_types
 /en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills:_Other_controls	/en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills/Other_controls
 /en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills:_Styling_basics	/en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills/Styling_basics
+/en-US/docs/Learn_web_development/Extensions/Security_privacy	/en-US/docs/Learn_web_development/Extensions
```
bsmth, line 5828: What do you think about redirecting to https://developer.mozilla.org/en-US/docs/Web/Security instead? We have traffic to the `/Learn_web_development/Extensions/Security_privacy` URL, so pointing to "next best thing" for the security topic is better for intent

chrisdavidmills: That makes sense. Can I just edit the URL in the redirect manually, or will that break things?

bsmth: Not recommended to do it manually due to checks built in to the tooling (de-dupe, etc.), you should be able to run:

```bash
yarn content add-redirect "/en-US/docs/Learn_web_development/Extensions/Security_privacy" "/en-US/docs/Web/Security"
```

chrisdavidmills: Ah, I see, and that overwrites the previous redirect behavior. I've run the command and committed the change.

files/en-us/_redirects.txt
```diff
@@ -5825,6 +5825,8 @@
 /en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills:_HTML5_controls	/en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills/Input_types
 /en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills:_Other_controls	/en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills/Other_controls
 /en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills:_Styling_basics	/en-US/docs/Learn_web_development/Extensions/Forms/Test_your_skills/Styling_basics
+/en-US/docs/Learn_web_development/Extensions/Security_privacy	/en-US/docs/Web/Security
+/en-US/docs/Learn_web_development/Extensions/Transform_animate	/en-US/docs/Learn_web_development/Extensions
```
bsmth, line 5829: Same comment about `/en-US/docs/Web/CSS/CSS_transforms/Using_CSS_transforms`, sorry missed it the first time

bsmth (approved):
One redirect to update, then good to go, tnx 👍🏻

files/en-us/_redirects.txt
chrisdavidmills: I'm not convinced that this redirect would be as useful as the module was intended to be about transitions and animations as well, not just transforms.

PR #39623, opened 2025-05-22
Reverting release note branches

Rumyra:
### Description

We're reverting release note branches while we re-consider the workflow

This pr adds 139 as beta and 140 as nightly - there will be a follow up pr next week to add 141


bsmth (approved):
Looks good

PR #39625, opened 2025-05-22
Fix some lint reports

Josh-Cena:
I am starting to run ESLint on our code. This PR fixes reports of some rules that reported <10 times. These rules and configurations aim for more concise, more secure, and more idiomatic code. Obviously some stylistic choices may be subjective, but generally the intention should be clear.


bsmth (approved):
Thanks, all looking good 👍🏻

PR #39626, opened 2025-05-22
Add the default value indicator in prose

dipikabh:
### Description

As pointed out by @caugner, the [content-visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility#values) property page doesn't explicitly mention in prose what the default value is. This PR adds that sentence.

I looked at a few more pages randomly and added the similar "default value" statement in the "Values" section.

### Motivation

To provide default value info alongside the value's description in addition to the info being present in the Formal definition section.

### DIFF
--- a/files/en-us/web/css/grid-column-end/index.md
+++ b/files/en-us/web/css/grid-column-end/index.md
@@ -90,7 +90,7 @@ grid-column-end: unset;
   - : Contributes nothing to the grid item's placement, indicating auto-placement, an automatic span, or a default span of `1`. This is the default value.
 - `<custom-ident>`
 
-  - : Contributes the first line to the grid item's placement if there is a named line with the name '\<custom-ident>-end'.
+  - : Contributes the first line to the grid item's placement if there is a named line with the name `<custom-ident>-end`.
 
     > [!NOTE]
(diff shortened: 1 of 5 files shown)


files/en-us/web/css/grid-column-end/index.md
```diff
@@ -87,10 +87,10 @@ grid-column-end: unset;
 ### Values
 
 - `auto`
-  - : Is a keyword indicating that the property contributes nothing to the grid item's placement, indicating auto-placement, an automatic span, or a default span of `1`.
+  - : Contributes nothing to the grid item's placement, indicating auto-placement, an automatic span, or a default span of `1`. This is the default value.
 - `<custom-ident>`
 
-  - : If there is a named line with the name '\<custom-ident>-end', it contributes the first such line to the grid item's placement.
+  - : Contributes the first line to the grid item's placement if there is a named line with the name '\<custom-ident>-end'.
```
estelle, line 93: while this is better than the original, so approvable, i don't really understand it. 
```suggestion
  - : Contributes the first line to the grid item's placement if there is a named line with the name `<custom-ident>-end`.
```

estelle (approved):
a few comments, but better than the original, so approving. If you want a re-review of updates, let me know.

dipikabh:
Thanks a lot for the review!

Issue #39611, opened 2025-05-22
`<li>` missing inside `<ul>`


page200:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Lists

### What information was incorrect, unhelpful, or incomplete?

That page contains

```html-nolint
<ul>
  milk
  eggs
  bread
  hummus
</ul>
```

That is not the intended code. `<li>` are missing.

### What did you expect to see?

No intermediate steps during which the code is wrong.

### Do you have anything more you want to share?

Some people don't read the entire page. The wrong code (without red markings or so) might be misleading for those who don't read the text that implies that it's wrong.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/learn_web_development/core/structuring_content/lists`
* MDN URL: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Lists
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/learn_web_development/core/structuring_content/lists/index.md
* Last commit: https://github.com/mdn/content/commit/48d220a8cffdfd5f088f8ca89724a9a92e34d8c0
* Document last modified: 2025-04-11T19:13:37.000Z

</details>

caugner:
That's intentional: The code block without `<li>`s is immediately followed by the code block with `<li>`s.

The paragraph between them says:

> The last step is to wrap each list item in a [`<li>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/li) (list item) element:

https://github.com/mdn/content/blob/f752b3500ba7057c244ba0ef6d5fd6ce0655d764/files/en-us/learn_web_development/core/structuring_content/lists/index.md#L36-L67

page200:
Some people don't read the entire page. The wrong code (without red markings or so) might be misleading for those who don't read the text that implies that it's wrong.

caugner:
Fixed via https://github.com/mdn/content/pull/39631.

PR #39631, opened 2025-05-22
Add prose to clarify that the invalid HTML is intentional

dipikabh:
### Description

This PR updates the [Unordered lists](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Lists#unordered_lists) section in the Learn area to address the concern raised in https://github.com/mdn/content/issues/39611.

It clarifies that the first example is intentionally invalid and that's because it is part of a step-by-step explanation to show how to create unordered lists.

### Motivation

To avoid confusion about the code being incorrect and to help readers who might just be skimming the page.


wbamberg:
Sorry to drive-by and feel free to ignore, but wouldn't it be OK just to have one step here? It doesn't seem that complicated a thing, to have to break it up (and then have to add the extra scaffolding of explaining that the middle step in invalid).

Josh-Cena:
Also does this intend to fix https://github.com/mdn/content/issues/39611?

bsmth (approved):
Nice thanks!

dipikabh:
Thanks everyone for your input.
I agree - this can be simplified a lot. And in this instance, we don't really need to show the intermediate step (the invalid HTML).
I've updated the prose: retained a bit of the steps phrasing and removed the incomplete HTML.

PR #39635, opened 2025-05-23
FF140 Relnote - TaskSignal.any()

hamishwillee:
FF140 supports [`TaskSignal.any()`](https://developer.mozilla.org/en-US/docs/Web/API/TaskSignal/any_static) in nightly in https://bugzilla.mozilla.org/show_bug.cgi?id=1964407

This adds a release note entry (in the experimental section)

Related work can be tracked in #39627


pepelsbey (approved):
Looks good! Thank you :)

Issue #26883, opened 2023-05-21
<link> to stylesheet won't work because of missing xmlns


Rolf-B:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_and_CSS

### What specific section or headline is this issue about?

First example

### What information was incorrect, unhelpful, or incomplete?

`<link>` needs `xmlns="http://www.w3.org/1999/xhtml" to work`, since `<link>` is not a SVG element. The existing xmlns:xlink in the example is not sufficent.

### What did you expect to see?

Either `xmlns:xhtml="http://www.w3.org/1999/xhtml"` in the SVG element and <xhtml:link...> to reference the stylesheet, or a local namespace declaration like this:

`<link xmlns="http://www.w3.org/1999/xhtml" rel="stylesheet" href="style8.css" type="text/css" />`

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/svg/tutorial/svg_and_css`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_and_CSS
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/svg/tutorial/svg_and_css/index.md
* Last commit: https://github.com/mdn/content/commit/b4f998244660723175f8e06b5d77f68cfb1d1f1a
* Document last modified: 2023-03-06T23:38:10.000Z

</details>

PR #39636, opened 2025-05-23
FF139 SVG discard element and api removed - expr feature too

hamishwillee:
FF139 (uplifted from FF140) removes experimental support for the SVG [`<discard>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/discard) element added behind pref in FF136, and the [`SVGDiscardElement`](https://developer.mozilla.org/en-US/docs/Web/API/SVGDiscardElement) interface added to release in FF136. 
I've removed the experimental feature entirely as it never released.

I've removed the `<discard>` element with direct as discussed in #26883 and all links.

The  `SVGDiscardElement` is reverted to experimental, but may be being deleted too (depends on BCD). This can merge and I'll do that later if needed.

Associated docs work can be tracked in #39618


bsmth (approved):
Thanks!

PR #39638, opened 2025-05-23
FF140 Relnote: Serializes < and > in attributes

hamishwillee:
FF140 ships support for escaping `<` and `>` to `&lt;` and `&gt;` in attributes when serializing HTML in https://bugzilla.mozilla.org/show_bug.cgi?id=1962084. This affects all the obvious methods like innerHTML, outerHTML, getHTML.

This adds a release note for the feature.

Related docs work can be tracked in #39628


pepelsbey (approved):
Looks good! Thank you :)

PR #39639, opened 2025-05-23
FF140 escaping `<` and `>` to `&lt;` and `&gt;` in attributes when serializing HTML

hamishwillee:
FF140 ships support for escaping `<` and `>` to `&lt;` and `&gt;` in attributes when serializing HTML in https://bugzilla.mozilla.org/show_bug.cgi?id=1962084. This affects all the obvious methods like innerHTML, outerHTML, getHTML etc.

This PR removes the info in the experimental features page (the addition to Relnote is in a separate PR).

I also added a note to the end of each of the affected methods pointing down to the browser compatibility for this feature. Note that the feature is non standard, but is expected to be standardized soon.

Related docs work can be tracked in #39628

### DIFF
--- a/files/en-us/web/api/element/gethtml/index.md
+++ b/files/en-us/web/api/element/gethtml/index.md
@@ -15,9 +15,8 @@ The options can be used to include nested shadow roots that have been set as {{d
 
 Without arguments, child nodes that are shadow roots are not serialized, and this method behaves in the same way as reading the value of {{domxref("Element.innerHTML")}}.
 
-Note that some browsers serialize `<` and `>` in attributes as `&lt;` and `&gt;` in the returned HTML (see [Browser compatibility](#browser_compatibility)).
-This prevents certain exploits where code becomes executable when serialized and then deserialized into HTML.
-
+Note that some browsers serialize the `<` and `>` characters as `&lt;` and `&gt;` when they appear in attribute values (see [Browser compatibility](#browser_compatibility)).
+This is to prevent a potential security vulnerability ([mutation XSS](https://research.securitum.com/dompurify-bypass-using-mxss/)) in which an attacker can craft input that bypasses a [sanitization function](/en-US/docs/Web/Security/Attacks/XSS#sanitization), enabling a cross-site scripting (XSS) attack.
 ## Syntax
 
 ```js-nolint
--- a/files/en-us/web/api/element/gethtml/index.md
+++ b/files/en-us/web/api/element/gethtml/index.md
@@ -15,6 +15,9 @@ The options can be used to include nested shadow roots that have been set as {{d
 
 Without arguments, child nodes that are shadow roots are not serialized, and this method behaves in the same way as reading the value of {{domxref("Element.innerHTML")}}.
 
+Note that some browsers serialize `<` and `>` in attributes as `&lt;` and `&gt;` in the returned HTML (see [Browser compatibility](#browser_compatibility)).
+This prevents certain exploits where code becomes executable when serialized and then deserialized into HTML.
+
 ## Syntax
 
 ```js-nolint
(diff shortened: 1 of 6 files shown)


files/en-us/web/api/element/outerhtml/index.md
```diff
@@ -9,32 +9,27 @@ browser-compat: api.Element.outerHTML
...
-string.
+DOM interface gets the serialized HTML fragment describing the element including its descendants.
+It can also be set to replace the element with nodes parsed from the given string.
 
-To only obtain the HTML representation of the contents of an element, or to replace the
-contents of an element, use the {{domxref("Element.innerHTML", "innerHTML")}} property
-instead.
+To only obtain the HTML representation of the contents of an element, or to replace the contents of an element, use the {{domxref("Element.innerHTML", "innerHTML")}} property instead.
+
+Note that some browsers serialize `<` and `>` in attributes as `&lt;` and `&gt;` when reading the HTML (see [Browser compatibility](#browser_compatibility)).
```
hamishwillee, line 17: This is the only actual change in this page.

files/en-us/web/api/shadowroot/innerhtml/index.md
```diff
@@ -8,9 +8,10 @@ browser-compat: api.ShadowRoot.innerHTML
 
 {{APIRef("Shadow DOM")}}
 
-The **`innerHTML`** property of the {{domxref("ShadowRoot")}}
-interface sets or returns a reference to the DOM tree inside the
-`ShadowRoot`.
+The **`innerHTML`** property of the {{domxref("ShadowRoot")}} interface sets gets or sets the HTML markup to the DOM tree inside the `ShadowRoot`.
```
hamishwillee, line 11: The text here seemed completely wrong - it isn't a reference. There is a lot more that can be said. 

I don't want to say it in this PR because that would be out of scope for this task. I do expect to be back here though, as I am currently doing work on the HTLM sanitization API which tanjentially affects this. I will also be looking at it as I add docs on the injection sinks for TrustedTypes.

files/en-us/web/api/element/gethtml/index.md
```diff
@@ -15,6 +15,9 @@ The options can be used to include nested shadow roots that have been set as {{d
 
 Without arguments, child nodes that are shadow roots are not serialized, and this method behaves in the same way as reading the value of {{domxref("Element.innerHTML")}}.
 
+Note that some browsers serialize `<` and `>` in attributes as `&lt;` and `&gt;` in the returned HTML (see [Browser compatibility](#browser_compatibility)).
+This prevents certain exploits where code becomes executable when serialized and then deserialized into HTML.
+
```
wbamberg, line 20: I think this could be clearer. There are some some details, like we should say "attribute values" not just "attributes", and AFAIK we usually say "parsed", not "deserialized", when we're talking about turning a string of HTML into DOM objects.

But regarding the exploit, there are a couple of things we might want to mention here:

(1) what is the exploit trying to accomplish
(2) how the exploit accomplishes it

For (1), AIUI we could say something like "bypass a sanitizer" and we could link to https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/XSS#sanitization to explain what sanitization means here.

For (2), what's written here is a bit vague tbh and from it, I would not be able to understand what's happening. For that, we'd need to link to something like https://research.securitum.com/dompurify-bypass-using-mxss/, which does explain it (although it's hard going). But I'm not sure it's worth it - I think it is OK to omit the "how".

So all that said, I'd suggest something like:

```suggestion
Note that some browsers serialize the `<` and `>` characters as `&lt;` and `&gt;` when they appear in attribute values (see [Browser compatibility](#browser_compatibility)). This is to prevent a potential security vulnerability in which an attacker can craft input that bypasses a [sanitization function](/en-US/docs/Web/Security/Attacks/XSS#sanitization), enabling a cross-site scripting (XSS) attack.
```

wbamberg: Yes, I think this is fine. I like linking to the blog post if we're going to mention mutation XSS, so people have somewhere to find out what it is.

pepelsbey (approved):
Looks good! Thank you :)

PR #39642, opened 2025-05-23
Editorial review: Document canvas context lang attribute

chrisdavidmills:
### Description

Chrome 136 supports the lang attribute for on- and off-screen canvas 2D drawing contexts: see https://chromestatus.com/feature/5066778773028864.

This PR adds documentation for the new attribute.


files/en-us/web/api/canvasrenderingcontext2d/lang/index.md
```diff
@@ -0,0 +1,190 @@
...
+
+The example is rendered as follows:
+
+{{ EmbedLiveSample('canvas-l10n', "100%", 220) }}
+
+Try changing the document language using the `<select>` element. When the language is set to English, the font will be rendered with the "fi" ligature. However, when it is set to Turkish, the font will be rendered without the "fi" ligature, because that locale doesn't include it.
+
+### Language support for offscreen canvases
+
+This example is the same as the previous example, except that the font is rendered to a {{domxref("OffscreenCanvasRenderingContext2D")}} then the resulting bitmap is transferred to the on-screen `<canvas>` to display. The `lang` attribute can be used to provide localization support to the off-screen canvas in just the same way as the regular on-screen `<canvas>`.
```
schenney-chromium, line 108: The `inherit` value behaves differently for Offscreen in that it is not dynamic. The inherited value is set when the offscreen is created, and then remains fixed for the lifetime of the offscreen. The spec text is `An OffscreenCanvas object has an internal inherited language and inherited direction set when the OffscreenCanvas s created.`. I think it's important to call this out in a paragraph following this one.

On the demo the best approach would be to set the `lang` attribute explicitly to the selected language. If fact, I think that's the only way to have the language change on the offscreen after creation.

schenney-chromium (commented):
Thanks so much for adding this.

The behavior of `lang = inherit` for the offscreen case is not super intuitive due to technical limitations with offscreen canvas. Any alternative would be even less intuitive because the behavior would depend on _how_ the offscreen was created and whether it was used in a worker.

schenney-chromium (commented):
LGTM. Thanks again. Not sure if an MDN reviewer needs to also review, but I would recommend it as I am inexperienced with MDN style and conventions (at least from an authoring perspective).

wbamberg:
I do have some more detailed comments on this PR, that I'll get to, but my initial comment is that this seems to be burying the lede. As I understand things, the key message, and the reason this property exists, is like this:

- you sometimes need to set a language for a canvas, so it knows how to render (for example) language-dependent font features
- an on-screen canvas is always associated with a particular `<canvas>` element, so whenever you render content using it, it can derive the language from the value of the `<canvas>` element's `lang` attribute (which might be set on the `<canvas>` element itself or inherited from an ancestor).
- however, an on-screen canvas renders the content _before_ it's associated with a `<canvas>` element, so it can't use the `lang` attribute of the `<canvas>` element for language-dependent rendering.

The `lang` property addresses this issue, allowing you to set a language directly on a canvas rendering context, whether you are using an offscreen canvas or not.

Is this correct? If so, I think we should say this, which we don't currently. There is some stuff like this, after "Due to technical limitations...", but these are presented as limitations of the `inherit` value, not of the Canvas API as a whole.

But I'm wondering my understanding is correct, because if it is, some other things are wrong. For example:

> For off-screen canvases, the lang value is inherited when the associated [OffscreenCanvasRenderingContext2D](https://pr39642.review.mdn.allizom.net/en-US/docs/Web/API/OffscreenCanvasRenderingContext2D) object is first created, and then fixed for the lifetime of the [OffscreenCanvas](https://pr39642.review.mdn.allizom.net/en-US/docs/Web/API/OffscreenCanvas).

...this presents the issue as just one of dynamic updates, but if I'm right, the `lang` value is not inherited from the `lang` attribute of the `<canvas>` _at all_ for offscreen canvases. It seems to be inherited from the `lang` attribute of the document (i.e. the `<html>` element) - which makes sense, because the offscreen rendering context has no idea which `<canvas>` it will end up in.

chrisdavidmills:
> I do have some more detailed comments on this PR, that I'll get to, but my initial comment is that this seems to be burying the lede.

I hear you, @wbamberg. I've just pushed a commit that adds the desired explanation to a new description section, along with the `inherit` behavior and limitations stuff.

This will probably require a bit more wrangling, as it is somewhat fiddly to make sense of, but I think we're getting close.

files/en-us/web/api/canvasrenderingcontext2d/lang/index.md
```diff
@@ -0,0 +1,208 @@
...
+browser-compat: api.CanvasRenderingContext2D.lang
+---
+
+{{APIRef}}
+
+The **`CanvasRenderingContext2D.lang`** property of the Canvas 2D API gets or sets the language of the canvas drawing context.
+
+## Value
+
+A [BCP 47 language tag](https://datatracker.ietf.org/doc/html/rfc5646) representing the language of the canvas context, or `inherit`, in which case the language is inherited from the [`lang`](/en-US/docs/Web/HTML/Reference/Global_attributes/lang) attribute of the originating {{HTMLElement("canvas")}} element or the nearest available ancestor with an explicit `lang` set.
```
wbamberg: AFAICT there are three sorts of values: a BCP 47 tag, the string value `"inherit"`, or the empty string. IMO it would be worth bulleting them to make them scannable, e.g.

```
One of the following values:
- A BCP 47 language tag.
- The string value `"inherit"`.
- An empty string.
```

I don't know how an empty string is interpreted.

wbamberg: The page for `lang` has a nice intro to BCP language tag syntax: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang#language_tag_syntax. What would be great IMO would be to make a new glossary entry "BCP 47 language tag", and move that into there, and link to it from both these pages. I don't think just linking people to the RFC is terribly helpful.

files/en-us/web/api/canvasrenderingcontext2d/lang/index.md
```diff
@@ -0,0 +1,208 @@
...
+
+The default value is `inherit`.
+
+## Description
+
+Sometimes, it is necessary to set a language for a canvas rendering context so that it knows how to render language-dependent features such as some fonts. An on-screen canvas context (`CanvasRenderingContext2D`) is always associated with a particular `<canvas>` element, so whenever you render content using it, it can derive the language from the value of the `<canvas>` element's `lang` attribute.
```
wbamberg, line 27:
```suggestion
Sometimes, you need to set a language for a canvas rendering context so that it knows how to render language-dependent features: for example, some fonts render certain characters differently in different languages. An on-screen canvas context (`CanvasRenderingContext2D`) is always associated with a particular `<canvas>` element, so whenever you render content using it, it can derive the language from the value of the `<canvas>` element's `lang` attribute.
```

(a bit pedantic, but "you need to" is shorter and more direct than "it is necessary to". I don't actually like my other suggestion much, but "language-dependent features such as some fonts" isn't really right - a font isn't a language-dependent feature, it might implement some though.)

files/en-us/web/api/canvasrenderingcontext2d/lang/index.md
```diff
@@ -0,0 +1,208 @@
...
+
+// Set context language
+ctx.lang = "en";
+// Logs "en"
+console.log(ctx.lang);
+```
+
+### Demonstrating canvas context localization support
+
+In this example, we render a text string to a 2D canvas context in a particular font that has language-dependant ligatures. We allow the canvas context's language to be adjusted so you can see the difference in rendering.
```
wbamberg:
```suggestion
In this example, we render a text string to a 2D canvas context in a particular font that has language-dependent ligatures. We allow the canvas context's language to be adjusted so you can see the difference in rendering.
```

files/en-us/web/api/canvasrenderingcontext2d/lang/index.md
```diff
@@ -0,0 +1,208 @@
...
+
+In the JavaScript, we first grab references to the `<canvas>` element, its `CanvasRenderingContext2D`, and the `<select>` element, then load the language-dependant font using the [CSS Font Loading API](/en-US/docs/Web/API/CSS_Font_Loading_API). Once the font is loaded, we run an `init()` function. This function defines another function — `drawText()`, which draws some text to the canvas context that uses the loaded font, adds a [`change`](/en-US/docs/Web/API/HTMLElement/change_event) [event listener](/en-US/docs/Web/API/EventTarget/addEventListener) to the `<select>` element, then calls `drawText()` so that the text is immediately drawn to the canvas when the page first loads.
+
+```js live-example___canvas-l10n
+const canvasElem = document.querySelector("canvas");
+const ctx = canvasElem.getContext("2d");
+
+const selectElem = document.querySelector("select");
+
+let latoMediumFontFace = new FontFace(
```
wbamberg:
```suggestion
const latoMediumFontFace = new FontFace(
```

files/en-us/web/api/canvasrenderingcontext2d/lang/index.md
```diff
@@ -0,0 +1,208 @@
...
+```js live-example___offscreen-l10n
+const canvasElem = document.querySelector("canvas");
+const ctx = canvasElem.getContext("bitmaprenderer");
+
+const offscreen = new OffscreenCanvas(canvasElem.width, canvasElem.height);
+const offscreen_ctx = offscreen.getContext("2d");
+
+const selectElem = document.querySelector("select");
+
+let latoMediumFontFace = new FontFace(
```
wbamberg:
```suggestion
const latoMediumFontFace = new FontFace(
```

wbamberg (commented):
Thanks for the updates Chris. I tried to improve the stuff on `inherit`, but I wasn't really able to, so I gave up. Like you say, this is tricky to explain.

files/en-us/web/api/canvasrenderingcontext2d/lang/index.md
chrisdavidmills: I totally agree with the first change. I think the second change was OK, but I've tweaked it a bit to "...some fonts have certain characters rendered differently in different languages."

i.e., it is not the fonts themselves that do the rendering.

files/en-us/web/api/canvasrenderingcontext2d/lang/index.md
chrisdavidmills: I've made the suggested change to the "Value" section. I had a bit of a play around with the empty string value, but found it difficult to see what effect it had, if any. Where did you read about this possible value?

chrisdavidmills: I think the separate glossary page for "BCP 47 language tag" is a really good idea; I've referred to it quite a few times in my last few PRs, and you are right, the RFc is not very helpful. I'll work on that now.

wbamberg: > I've made the suggested change to the "Value" section. I had a bit of a play around with the empty string value, but found it difficult to see what effect it had, if any. Where did you read about this possible value?

https://html.spec.whatwg.org/multipage/canvas.html#concept-canvastextdrawingstyles-language

> Objects that implement the [CanvasTextDrawingStyles](https://html.spec.whatwg.org/multipage/canvas.html#canvastextdrawingstyles) interface have an associated language value used to localize font rendering. Valid values are a BCP 47 language tag, the empty string, or "inherit" where the language comes from the [canvas](https://html.spec.whatwg.org/multipage/canvas.html#the-canvas-element) element's language, or the associated [document element](https://dom.spec.whatwg.org/#document-element) when there is no [canvas](https://html.spec.whatwg.org/multipage/canvas.html#the-canvas-element) element. Initially, the [language](https://html.spec.whatwg.org/multipage/canvas.html#concept-canvastextdrawingstyles-language) must be "[inherit](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-lang-inherit)".

I don't know, maybe it's fine to omit this. I just saw it in the spec and wondered.

chrisdavidmills: OK, on further playing, it looks like the empty string can be used to specify that you don't want the canvas to have a language (it also stops it from inheriting a language from the HTML document). I've updated the description in light of this.

I've also added the BCP 47 glossary entry.

files/en-us/glossary/bcp_47_language_tag/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+- The HTML [`lang`](/en-US/docs/Web/HTML/Reference/Global_attributes/lang) attribute
+- The HTML {{htmlelement("track")}} element
+- The {{jsxref("Temporal")}} object
+- The {{domxref("CanvasRenderingContext2D.lang")}} property
+- The [Translator and Language Detector APIs](/en-US/docs/Web/API/Translator_and_Language_Detector_APIs)
+
+## BCP 47 syntax
+
+The full BCP 47 syntax is in-depth enough to mark extremely specific language dialects, but most usage is much simpler.
```
wbamberg:
```suggestion
The full BCP 47 syntax is specified in {{rfc("5646")}}. It is capable of identifying extremely specific language dialects, but most usage is much simpler.
```

(mostly, I think it is worth pointing to the definitive syntax prominently, even if most people will be able to use the more informal description here)

wbamberg (approved):
OK I have a comment to add a more prominent link to the RFC for the BCP 47 page, but I'm approving this anyway so you can merge with that change, if you accept it.

Thanks Chris! The glossary page looks very useful. Turns out there are lots of MDN pages that link to the BCP RFC, but I won't ask you to update them all 🤣.

files/en-us/glossary/bcp_47_language_tag/index.md
chrisdavidmills: Super, yup, like it!

chrisdavidmills:
> OK I have a comment to add a more prominent link to the RFC for the BCP 47 page, but I'm approving this anyway so you can merge with that change, if you accept it.
> 
> Thanks Chris! The glossary page looks very useful. Turns out there are lots of MDN pages that link to the BCP RFC, but I won't ask you to update them all 🤣.

Yay, thanks @wbamberg! 

I was definitely thinking of updating my pages that link to the RFC to link to this page instead; doing the whole lot might be a good joint project.

PR #39644, opened 2025-05-23
Lint: use parseInt with a radix

Josh-Cena:
`parseInt` tries to guess the radix if it's not explicitly provided, which often leads to unintended behavior. A radix of `10` should always be explicitly given.

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/api/menus/overridecontext/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/api/menus/overridecontext/index.md
@@ -50,7 +50,7 @@ document.addEventListener(
       // set the context to "opening a tab context menu".
       browser.menus.overrideContext({
         context: "tab",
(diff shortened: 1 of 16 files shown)


files/en-us/mozilla/add-ons/webextensions/api/menus/overridecontext/index.md
```diff
@@ -50,7 +50,7 @@ document.addEventListener(
       // set the context to "opening a tab context menu".
       browser.menus.overrideContext({
         context: "tab",
-        tabId: parseInt(foo.dataset.tabId),
+        tabId: parseInt(foo.dataset.tabId, 10),
```
bsmth, line 53: TIL, thanks: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#radix

bsmth (approved):
LGTM, thanks 👍🏻

PR #39645, opened 2025-05-23
Lint: more concise regex syntax

Josh-Cena:
Use `\d` instead of `[0-9]`; make sure `m` and `s` regexes actually use the feature enabled by the flag (otherwise the flag is redundant)

### DIFF
--- a/files/en-us/learn_web_development/extensions/forms/form_validation/index.md
+++ b/files/en-us/learn_web_development/extensions/forms/form_validation/index.md
@@ -788,7 +788,7 @@ const email = document.getElementById("mail");
 const error = document.getElementById("error");
 
 // Regular expression for email validation as per HTML specification
-const emailRegExp = /^[\w.!#$%&'*+/=?^`{|}~-]+@[a-z0-9-]+(?:\.[a-z0-9-]+)*$/i;
+const emailRegExp = /^[\w.!#$%&'*+/=?^`{|}~-]+@[a-z\d-]+(?:\.[a-z\d-]+)*$/i;
 
 // Check if the email is valid
 const isValidEmail = () => {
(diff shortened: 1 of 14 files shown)


chrisdavidmills (approved):
LGTM; cheers @Josh-Cena!

files/en-us/learn_web_development/extensions/forms/form_validation/index.md
```diff
@@ -788,7 +788,7 @@ const email = document.getElementById("mail");
 const error = document.getElementById("error");
 
 // Regular expression for email validation as per HTML specification
-const emailRegExp = /^[\w.!#$%&'*+/=?^`{|}~-]+@[a-z0-9-]+(?:\.[a-z0-9-]+)*$/i;
+const emailRegExp = /^[\w.!#$%&'*+/=?^`{|}~-]+@[a-z\d-]+(?:\.[a-z\d-]+)*$/i;
```
bsmth, line 791: Just a note here that the line above talks about a regex in the HTML spec

Josh-Cena: I checked; I don't think there's a regex verbatim. Rather, HTML uses BNF.

ltrebing:
I think that the regular expression in https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/email#basic_validation should either be reverted to the old version (which was [the “official” expression from the HTML spec](https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address) except for one unnecessary `\` character), or there should be a link to the source with an explanation stating that the shorter expression is (supposedly) equivalent.

Josh-Cena:
Oh I didn't realize there's an official version. In that case yes we should restore it to the official version.

bsmth:
> Oh I didn't realize there's an official version. In that case yes we should restore it to the official version.

I think the spec version says it's "an implementation" - but I guess it's 'the HTML spec version', so, also +1 on replacing it.

PR #39646, opened 2025-05-23
Lint: improve import/label/comment formatting

Josh-Cena:
Fix a few cosmetic things that Prettier can't handle itself.

- The `a: b` syntax is misparsed by Prettier as a label, and therefore adds a `;` instead of the correct `,`
- There should be a new line after import/require statements.
- Imports should appear at the top of the file.
- Comments should have padding spaces: `// a` instead of `//a`.


bsmth (approved):
LGTM, thanks 👍🏻

PR #39647, opened 2025-05-23
Lint: use concise arrows

Josh-Cena:
Our style guide explicitly requires the use of concise arrows `(a) => b` over `(a) => { return b; }`.

### DIFF
--- a/files/en-us/web/api/cache/index.md
+++ b/files/en-us/web/api/cache/index.md
@@ -83,86 +83,85 @@ self.addEventListener("fetch", (event) => {
   console.log("Handling fetch event for", event.request.url);
 
   event.respondWith(
-    caches.open(CURRENT_CACHES.font).then((cache) => {
-      return cache
-        .match(event.request)
-        .then((response) => {
-          if (response) {
-            // If there is an entry in the cache for event.request,
-            // then response will be defined and we can just return it.
-            // Note that in this example, only font resources are cached.
-            console.log(" Found response in cache:", response);
+    caches
+      .open(CURRENT_CACHES.font)
+      .then((cache) => cache.match(event.request))
+      .then((response) => {
+        if (response) {
+          // If there is an entry in the cache for event.request,
+          // then response will be defined and we can just return it.
+          // Note that in this example, only font resources are cached.
+          console.log(" Found response in cache:", response);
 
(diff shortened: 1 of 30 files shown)


files/en-us/web/api/cache/index.md
```diff
@@ -83,86 +83,85 @@ self.addEventListener("fetch", (event) => {
...
+            // (i.e. HTTP status code of 4xx or 5xx).
+            // We also only want to cache responses that correspond
+            // to fonts, i.e. have a Content-Type response header that
+            // starts with "font/".
+            // Note that for opaque filtered responses
+            // https://fetch.spec.whatwg.org/#concept-filtered-response-opaque
+            // we can't access to the response headers, so this check will
+            // always fail and the font won't be cached.
+            // All of the Google Web Fonts are served from a domain that
+            // supports CORS, so that isn't an issue here.
```
bsmth, line 134: Unrelated to the changes, but I think we should move this out to prose, I find it hard to read both the code and the comments in this form.

bsmth (approved):
All good except for https://pr39647.review.mdn.allizom.net/en-US/docs/Web/API/Cache#examples, but we can defer if you want to do this separately

files/en-us/web/api/cache/index.md
Josh-Cena: The code and the comments are quite intertwined. There still doesn't exist a way to conveniently annotate code with lots of prose—we would have to split this up into many pieces, which I also hate because that's not lintable. Alternatively we can just use a numbered list of remarks after the code, but I'm not sure if that's more readable either.

bsmth: >conveniently annotate code with lots of prose

Maybe we can track as a feature request. Markdown has footnotes, I wonder if we could use that in a nice way like:

````md
```js
if (response) {
  // [^1]
  console.log("Found response in cache: ", response);
  return response;
}
```

[^1]: If there is an entry in the cache for event.request, 
  then response will be defined and we can just return it. 
  In this example, only font resources are cached.
````

Josh-Cena: This would be great! Not sure how it should render, though, I was imagining a two-column layout.

PR #39648, opened 2025-05-23
feat(http): Improve header titles, short-titles

bsmth:
### Description

In this PR, I'm adding more content/context to page titles and using current `titles` -> `short-title`s. 

__Additions:__
- `short-title` for HTTP headers and methods

__Changes:__
- Update `title` to say if it's a header, a Permissions-Policy or CSP directive, and that `CONNECT` etc. all refers to request methods.

For example all header pages:

```yaml
title: Accept-CH
```

become

```diff
-title: Accept-CH
+title: Accept-CH header
+short-title: Accept-CH
```

And all CSP/Permissions Policy directives:

```yaml
title: "CSP: script-src"
```

Become:

```diff
-title: "CSP: script-src"
+title: "Content-Security-Policy: script-src directive"
+short-title: script-src
```

The same change is repeated through the whole section.

- [x] https://github.com/mdn/content/pull/39472


dipikabh (approved):
Neat! The update looks good, thanks.

Also leaving this food for thought here: Should we also add "HTTP" to the header titles? Might help with searchability. As in:
```yaml
title: Accept-CH HTTP header
```

bsmth:
> Neat! The update looks good, thanks.
> 
> Also leaving this food for thought here: Should we also add "HTTP" to the header titles? Might help with searchability. As in:
> 
> ```yaml
> title: Accept-CH HTTP header
> ```

Thank you! I thought about this, but rari is already adding HTTP in the page `<title>`, so I avoided for now. We do this for HTML attributes, though: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete

PR #39652, opened 2025-05-23
Add docs for Speculation Rules features: target_hint and clear-site-data

tunetheweb:
### Description

Chrome 138 will add two new features for Speculation Rules:
- [`target_hint`](https://chromestatus.com/feature/5162540351094784)
- [`clear-site-data`](https://chromestatus.com/feature/4755745652080640)

### Motivation

New features

### Additional details

**`target_hint`**

Chrome Status link: https://chromestatus.com/feature/5162540351094784
Explainer link: https://github.com/WICG/nav-speculation/blob/main/triggers.md#window-name-targeting-hints

**`Clear-site-data`**

Chrome Status link: https://chromestatus.com/feature/4755745652080640
Explainer link: https://github.com/WICG/nav-speculation/blob/main/clear-site-data-integration.md

BCD PR: https://github.com/mdn/browser-compat-data/pull/26893


files/en-us/web/http/reference/headers/clear-site-data/index.md
```diff
@@ -54,7 +54,21 @@ Clear-Site-Data: "*"
...
+
+  - : The server signals that the client should reload all browsing contexts for the origin of the response ({{domxref("Location.reload")}}).
+
+- `"prefetchCache"`
+
+  - : Used to evict {{ domxref("Speculation Rules API", "speculation rules") }} prefetches that are scoped to the referrer origin.
+
+- `"prerenderCache"`
+
+  - : Used to cancel {{ domxref("Speculation Rules API", "speculation rules") }} prerenders that are scoped to the referrer origin.
```
hamishwillee, line 70: What is the difference between cancel and evict used for `prefetchCache` - i.e. what specific difference in behaviour are you trying to capture here?

hamishwillee: ALSO This will otherwise render as code

```suggestion
  - : Used to cancel {{domxref("Speculation Rules API", "speculation rules","", "nocode")}} prerenders that are scoped to the referrer origin.
```

files/en-us/web/http/reference/headers/clear-site-data/index.md
```diff
@@ -67,19 +81,17 @@ Clear-Site-Data: "*"
...
-  - : The server signals that the client should reload all browsing contexts for the origin of the response ({{domxref("Location.reload")}}).
 - `"*"` (wildcard)
   - : The server signals that the client should clear all types of data for the origin of the response. If more data types are added in future versions of this header, they will also be covered by it.
 
 ## Examples
 
 ### Sign out of a website
 
-If a user signs out of your website or service, you might want to remove locally stored data. To do this, add the `Clear-Site-Data` header to the page that confirms the logging out from the site has been accomplished successfully (`https://example.com/logout`, for example):
+If a user signs out of your website or service, you might want to remove locally stored data and speculated navigations. To do this, add the `Clear-Site-Data` header to the page that confirms the logging out from the site has been accomplished successfully (`https://example.com/logout`, for example):
```
hamishwillee, line 91: I feel that perhaps "speculated navigations" is still not well enough understood to be "bare text" like this. Maybe a link, or maybe something like.

```suggestion
If a user signs out of your website or service, you might want to remove locally stored data, including any prefetched or prerendered content for {{domxref("Speculation Rules API", "speculated navigations","", "nocode")}}.
To do this, add the `Clear-Site-Data` header to the page that confirms the logging out from the site has been accomplished successfully (`https://example.com/logout`, for example):
```

files/en-us/web/http/reference/headers/clear-site-data/index.md
```diff
@@ -90,6 +102,20 @@ If this header is delivered with the response at `https://example.com/clear-cook
...
 
+### Clearing speculations
+
+If this header is delivered with the response at `https://example.com/change-state.json`, all prerendered speculations on the same domain `https://example.com` and any subdomains (like `https://stage.example.com`, etc.), will be cleared out.
+
+```http
+Clear-Site-Data: "prerenderCache"
+```
+
+To clear both prefetch and prerender speculations, both `prefetchCache` and `prerenderCache` must be sent:
```
hamishwillee, line 113: IMO This OK because we have immediately above linked to the API.

files/en-us/web/html/reference/elements/script/type/speculationrules/index.md
```diff
@@ -175,6 +175,10 @@ Each object can contain the following properties:
 
   - : A string used to identify a rule or ruleset. This will be included in the {{HTTPHeader("Sec-Speculation-Tags")}} request header for any speculations covered by that rule.
 
+- `"target_hint"` {{experimental_inline}}
+
+  - : A string indicating where the page expects the prerendered content to be activated. For example, `"target_hint": "_blank"` or `"target_hint": "_self"` (the default if not specified). Check [Browser compatibility](#browser_compatibility) for values beyond these. This directive is not supported for `prefetch` speculations.
```
hamishwillee, line 180: We don't normally forshadow that there might be extensions in future versions - otherwise we'd have the line everywhere.

```suggestion
  - : A string indicating where the page expects the prerendered content to be activated.
    This is either `"target_hint": "_blank"` or `"target_hint": "_self"` (the default if not specified). 
  
    The directive not supported for prefetch speculations.
     
```

files/en-us/web/html/reference/elements/script/type/speculationrules/index.md
```diff
@@ -554,6 +558,46 @@ Or to identify individual rules:
 
 See {{HTTPHeader("Sec-Speculation-Tags")}} for more examples.
 
+### `target_hint` example
+
+A `target_hint` can be included to indicate:
```
hamishwillee, line 563:
```suggestion
A `target_hint` can be included to indicate the target window in which matching prerender speculations will be opened:
```

hamishwillee (commented):
Thanks @tunetheweb - looks pretty clear to me. I've added some suggestions for discussion. I'd like to see the BCD go in before we push this in if possible.

files/en-us/web/http/reference/headers/clear-site-data/index.md
tunetheweb: Oh that's an interesting nuance I didn't consider when I took this wording from the explainer: https://github.com/WICG/nav-speculation/blob/main/clear-site-data-integration.md#implementation

Technically, the `prefetchCache` directive instructs the browser to evict it from the prefetch cache (so it can't be used for a subsequent navigation).
Whereas `prerenderCache` instructs the browser to cancel any prerender that exists (which is kinda like a new, hidden, background tab ready to be brought to the foreground).

It's a subtle nuance, but I think the current wording is actually correct. Open to ideas of better ways to express this though?

Will accept your other edit in the meantime and reopen this until we close this discussion.

files/en-us/web/http/reference/headers/clear-site-data/index.md
```diff
@@ -90,6 +102,20 @@ If this header is delivered with the response at `https://example.com/clear-cook
 Clear-Site-Data: "cookies"
 ```
 
+### Clearing speculations
+
+If this header is delivered with the response at `https://example.com/change-state.json`, all prerendered speculations on the same domain `https://example.com` and any subdomains (like `https://stage.example.com`, etc.), will be cleared out.
```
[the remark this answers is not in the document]
tunetheweb: To me a cache suggests something that is stored and delivered unaltered and will be the same each time (i.e. it's the static code that's cached). A prerendered page is subtley different in that it's a (potentially fully) rendered, including JavaScript execution and so might include dynamic content (e.g. prerendering a "top headlines" page which might change). So think "cache" is a little confusing — despite that being in the directive name — and think "prerendered content" actually fits better. WDYT?

Other than that, the other changes LGTM.

tunetheweb:
> Thanks @tunetheweb - looks pretty clear to me. I've added some suggestions for discussion. I'd like to see the BCD go in before we push this in if possible.

SGTM. I opened both together so I’d know the anchor links when adding to BCD but yeah let’s wait until both are ready to merge. Plus a few things still open in this PR to resolve.

files/en-us/web/http/reference/headers/clear-site-data/index.md
```diff
@@ -54,7 +54,21 @@ Clear-Site-Data: "*"
...
+
+  - : The server signals that the client should reload all browsing contexts for the origin of the response ({{domxref("Location.reload")}}).
+
+- `"prefetchCache"`
+
+  - : Used to evict {{domxref("Speculation Rules API", "speculation rules", "", "nocode")}} prefetches that are scoped to the referrer origin.
+
+- `"prerenderCache"`
+
+  - : Used to cancel {{domxref("Speculation Rules API", "speculation rules","", "nocode")}} prerenders that are scoped to the referrer origin.
```
hamishwillee, line 70: Follows on from https://github.com/mdn/content/pull/39652#discussion_r2106553808

@tunetheweb Thanks

- prefetch "evict" tells me that you are clearing prefetch cache entries for the referrer origin. From this I don't know if ongoing prefetch operations are canceled.
- prerenderCache "cancel" means to me that you are cancelling a prerender operation. It does not (absolutely) tell me that you've cleared an existing prerender.

Given that both of these are called "caches" I would use the word "clear" for both, possibly with cancel if that is happening. The detail of entries vs hidden tabs is no necessary at this level and only the confusion of cancelling ongoing opreations vs just deleting the existing entries matters.

So both might be something like:

> Used to clear {{domxref("Speculation Rules API", "speculation rules", "", "nocode")}} [prefetches|prerenders] that are scoped to the referrer origin, cancelling any ongoing operations.

files/en-us/web/http/reference/headers/clear-site-data/index.md
hamishwillee: I don't think you need a detailed example, but perhaps _after_ the code below (just before the specifications) you might want to explain that you can do either, and why. I'm thinking something like

> You can also clear either prefetched or prerendered content separately.
> Often you will need to clear prerendered content, but will be able to keep prefetched content for when the page is displayed or for future prerenders.
> The reason for this is that prerendered pages are fully evaluated when they are created and hence are more likely to become stale, while prefetched content is not evaluated until the page is rendered.

files/en-us/web/api/speculation_rules_api/index.md
```diff
@@ -306,6 +308,8 @@ User-specific state problems can occur for other user settings, for example lang
 
 The best mitigation for these cases, and indeed any time when content can get out of sync with the server, is for pages to refresh themselves as needed. For example, a server might use the [Broadcast Channel API](/en-US/docs/Web/API/Broadcast_Channel_API), or another mechanism such as {{domxref("Window/fetch", "fetch()")}} or a {{domxref("WebSocket")}}. Pages can then update themselves appropriately, including speculatively loaded pages that have not yet been activated.
 
+Stale speculations can be cleared using the {{ httpheader("Clear-Site-Data") }} response header with the {{ httpheader("Clear-Site-Data#prefetchCache", `prefetchCache`) }} or {{ httpheader("Clear-Site-Data#prerenderCache", `prerenderCache`) }} values (or both) as appropriate where keeping them fresh is not possible. This header can be returned on any same-site HTTP request (such as an `/api/add-to-cart` API call).
```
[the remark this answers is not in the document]
hamishwillee: You mean keep the prefetch part as discussed in https://github.com/mdn/content/pull/39652/files#r2108088582 ?
That is worth doing. 

The main point though I was trying to make here was to show the concrete case where all those prerendered pages might be "invisibly displaying the number of items in the cart". it just feels more "real".
I'll keep an eye out for your changes.

files/en-us/web/http/reference/headers/clear-site-data/index.md
tunetheweb: Done.

files/en-us/web/api/speculation_rules_api/index.md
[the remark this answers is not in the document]
tunetheweb: I meant the other changes you suggested here.

I've just realise the add to cart example follows on from the previous example. So don't think the middle paragraph is needed as it's repetitive so went with this:

```suggestion
Where refreshes are not possible, speculations can be cleared using the {{ httpheader("Clear-Site-Data") }} response header with the {{ httpheader("Clear-Site-Data#prefetchCache", `prefetchCache`) }} or {{ httpheader("Clear-Site-Data#prerenderCache", `prerenderCache`) }} values (or both) as appropriate.

The header can be returned on any same-site HTTP request (such as an `/api/add-to-cart` API call).
```

files/en-us/web/http/reference/headers/clear-site-data/index.md
tunetheweb: Tried a few things, but think the fuller example actually is the clearest so went with that.

files/en-us/web/http/reference/headers/clear-site-data/index.md
```diff
@@ -54,7 +54,21 @@ Clear-Site-Data: "*"
     > In browsers that support the `"clientHints"` data type, client hints are also cleared when the `"cache"`, `"cookies"`, or `"*"` types are specified. `"clientHints"` is therefore only needed when none of those other types are specified.
 
 - `"cookies"`
+
   - : The server signals that the client should remove all cookies for the origin of the response URL. HTTP authentication credentials are also cleared out. This affects the entire registered domain, including subdomains. So `https://example.com` as well as `https://stage.example.com`, will have cookies cleared.
+
+- `"executionContexts"` {{Experimental_Inline}}
+
+  - : The server signals that the client should reload all browsing contexts for the origin of the response ({{domxref("Location.reload")}}).
```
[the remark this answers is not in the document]
tunetheweb: Actually I've changed my mind on this. `executionContexts` is still in [the spec](https://www.w3.org/TR/clear-site-data/), even if no browser has implemented it yet. So think it is worth keeping. It's not so much deprecated as just not implemented yet. So let's leave it (but move it to be in alphabetical order).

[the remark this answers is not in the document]
hamishwillee: FWIW the MDN rule for this is to no document things that don't exist in any browser. Given this is moved though, I'm going to pretend I didn't see it.

files/en-us/web/http/reference/headers/clear-site-data/index.md
```diff
@@ -90,6 +103,28 @@ If this header is delivered with the response at `https://example.com/clear-cook
...
+
+```http
+Clear-Site-Data: "prefetchCache", "prerenderCache"
+```
+
+There are cases where clearing one or the other, or both, is appropriate.
+
+For example, a client-side rendered application that pulls in data from JavaScript might use `prerenderCache` on state change to discard the prerendered pages, but keep the prefetched HTML to use when the page is rendered (or prerendered again).
+
+On the other hand, if the prefetched HTML document contains stale data but the corresponding prerendered page is set up to refresh the data when it is displayed, you may not need to use `prerenderCache` but you probably will want to use the `prefetchCache` directive: so that the stale HTML isn't used in a future prerender.
```
hamishwillee, line 124: FYI I did a minor subedit around this bit to refer to prefetched HTML document rather than just "HTML document". I know they are the same thing, but mentioning "prefetch" here makes directly associates this HTML document in my mind with the relevant directive.

hamishwillee (approved):
Thanks very much @tunetheweb 

The last round of changes you did look good. I was interested to see the change regarding targets in `where` not being needed. Make sense.

FYI I work Mon, Tue, Friday on this stuff. Sometimes I might take longer than you expect to rereview. Sorry!

PR #39661, opened 2025-05-24
Lint: prefer for...of loops over plain for loops

Josh-Cena:
It is prescribed by our style guide that `for...of` loops should be preferred over `for` loops in all cases. In the past we had done a great deal of work to refactor almost everything but it seems we left some behind.


bsmth (approved):
Great, tnx 👍🏻

PR #39677, opened 2025-05-26
Make String.p.pad{End,Start} articles consistent

### LINKED ISSUE #39671: Inconsistency: padStart and padEnd
There is a minor inconsistency in the wording between:

- [String.prototype.padStart()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart)
- [String.prototype.padEnd()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd)

Primarily, the use of **"space"** is more meaningful and preferred.

- The default value is the unicode "space" character (U+0020).
- The default value for this parameter is " " (U+0020).

Additionally, these 2 methods are almost identical, except for the start & end references.
IMHO, the text for these two documents should also be identical, except for the start & end references.

Josh-Cena:
Fix https://github.com/mdn/content/issues/39671

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/string/padend/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/string/padend/index.md
@@ -7,9 +7,7 @@ browser-compat: javascript.builtins.String.padEnd
 
 {{JSRef}}
 
-The **`padEnd()`** method of {{jsxref("String")}} values pads this string with a given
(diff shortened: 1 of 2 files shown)


chrisdavidmills (approved):
LGTM

PR #39680, opened 2025-05-26
FF140 Relnote/Expr Feat: Dialog closeBy attribute

hamishwillee:
FF140 enables the [`HTMLDialogElement.closedBy`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/closedBy) property and `<dialog>` [`closedby`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby) attribute in Nightly (pref `dom.dialog.light-dismiss.enabled` ) in https://bugzilla.mozilla.org/show_bug.cgi?id=1964077

This adds an experimental feature update and corresponding note in the release notes.

Related docs work can be tracked in #39621

### DIFF
--- a/files/en-us/mozilla/firefox/experimental_features/index.md
+++ b/files/en-us/mozilla/firefox/experimental_features/index.md
@@ -1100,7 +1100,7 @@ The [HTML Sanitizer API](/en-US/docs/Web/API/HTML_Sanitizer_API) allow developer
   </tbody>
 </table>
 
-### Choose allowed mechanisms for closing dialogs
+### `closedBy` attribute for `<dialog>`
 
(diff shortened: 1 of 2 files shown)


pepelsbey (commented):
Looks good! Thank you :) Though I’d suggest a different title for the feature: it describes what was implemented, and the description follows. Also, some stray backticks.

hamishwillee:
> Looks good! Thank you :) Though I’d suggest a different title for the feature: it describes what was implemented, and the description follows. Also, some stray backticks.

I chose to do it the way I did because we do both the element attribute and javascript object property.
Historically the approach I have used is more common in experimental features: describing the overall feature purpose rather than the specific functionality. 
This makes sense to me since it is more useful for a reader who doesn't know what the attributes do already.

But it isn't worth the time to argue this, so I accepted your changes

FYI, I also rebased because for some reason this was re-adding the an entry to the experimental features for HTML input sanitization. Really weird.

Thanks for the review, ready for you to look gain.

pepelsbey (approved):
Perfect! Thank you :)

PR #39681, opened 2025-05-26
fix: typo in options_ui deprecation warning

Ma-ve:
### Description

Fixes a typo in the deprecation warning for `browser_style`

### Motivation

It's sloppy and an easy fix


rebloor (approved):
Thanks for fixing @Ma-ve

PR #39573, opened 2025-05-19
Update index.md for better description for the coordinates.

EbereUzodufa:
### Description
Update the description for matrix(a, b, c, d, tx, ty) to show that it also stands for matrix(scaleX(), skewY(), skewX(), scaleY(), translateX(), translateY()), which is easier to understand as a developer.

### Motivation
Clarity. I am working with SVG at the moment, and I had to check different docs to understand the meaning of `matrix(a, b, c, d, tx, ty)`. Outside AI, MDN is most developers' first search place.

PR #39684, opened 2025-05-26
Update index.md

HomyeeKing:
revert https://github.com/mdn/content/pull/39573


chrisdavidmills (approved):
@HomyeeKing Sorry about adding in an error in the first place, that was my bad. And I see that the correct equivalent is mentioned at the bottom of the Values section anyway, so we can definitely get rid of this.

Thanks for the fix.

PR #39690, opened 2025-05-26
Array.fromAsync() mapFn does not always have its input awaited

### LINKED ISSUE #39688: Array.fromAsync mapFn does not await input item if it came from async iterable
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync

Using mapFn

Hi, I’m the champion of the Array.fromAsync proposal. While reviewing MDN’s documentation, I found a small error in the “[Using mapFn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync#using_mapfn)” section. It says:

> Both the input and output of mapFn are awaited internally by Array.fromAsync().

This is incorrect, or, rather, it is correct only when the input data structure given to Array.fromAsync is either a sync iterable or non-iterable array-like object. It is not true when the input data structure is an async iterable. This is designed to match the current behavior of `for await`.

It is also worth calling out that this means that:
```js
Array.fromAsync(input)
```
…is not equivalent to:
```js
Array.fromAsync(input, x => x)
```
…at least when `input` is an async iterable.

For example:

```js
function createAsyncIter () {
  let i = 0;
  return {
    [Symbol.asyncIterator]() {
      return {
        async next() {
          if (i > 2) return { done: true };
          i++;
          return { value: Promise.resolve(i), done: false }
        }
      }
    }
  };
}

// This prints `[Promise.resolve(1), Promise.resolve(2), Promise.resolve(3)]`:
console.log(await Array.fromAsync(createAsyncIter()));

// This prints `[1, 2, 3]`:
console.log(await Array.fromAsync(createAsyncIter(), x => x));
```

1. The incorrect sentence should be updated to state that the input items to `mapFn` are *not* awaited before being given to `mapFn` when the input is an async iterable.
2. The corrected sentence should make an analogy to `for await`.
3. It should also be noted that `Array.fromAsync(input)` and `Array.fromAsync(input, x => x)` are not equivalent, when `input` is an async iterable.
4. The example I gave above probably should be added to the documentation too.

I have [updated the proposal explainer to clarify this](https://github.com/tc39/proposal-array-from-async/blob/main/README.md#mapping-function). See also tc39/proposal-array-from-async#19.

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/javascript/reference/global_objects/array/fromasync`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/global_objects/array/fromasync/index.md
* Last commit: https://github.com/mdn/content/commit/364a4d02b10854ab7cef4ff4b0ec3616d4e1c8ab
* Document last modified: 2025-05-23T04:35:27.000Z

</details>

Josh-Cena:
Fix https://github.com/mdn/content/issues/39688.

@js-choi

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/array/fromasync/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/array/fromasync/index.md
@@ -12,19 +12,19 @@ The **`Array.fromAsync()`** static method creates a new, shallow-copied `Array` 
 ## Syntax
 
 ```js-nolint
-Array.fromAsync(arrayLike)
-Array.fromAsync(arrayLike, mapFn)
-Array.fromAsync(arrayLike, mapFn, thisArg)
+Array.fromAsync(items)
+Array.fromAsync(items, mapFn)
+Array.fromAsync(items, mapFn, thisArg)
 ```
 
 ### Parameters
 
-- `arrayLike`
+- `items`
   - : An async iterable, iterable, or array-like object to convert to an array.
 - `mapFn` {{optional_inline}}
   - : A function to call on every element of the array. If provided, every value to be added to the array is first passed through this function, and `mapFn`'s return value is added to the array instead (after being [awaited](/en-US/docs/Web/JavaScript/Reference/Operators/await)). The function is called with the following arguments:
     - `element`
-      - : The current element being processed in the array. Because all elements are first [awaited](/en-US/docs/Web/JavaScript/Reference/Operators/await), this value will never be a [thenable](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables).
+      - : The current element being processed in the array. If `items` is a sync iterable or array-like object, then all elements are first [awaited](/en-US/docs/Web/JavaScript/Reference/Operators/await), and `element` will never be a [thenable](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables). If `items` is an async iterable, then each yielded value is passed as-is.
     - `index`
       - : The index of the current element being processed in the array.
 - `thisArg` {{optional_inline}}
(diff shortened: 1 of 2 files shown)


js-choi:
Looks good to me, thank you.

files/en-us/web/javascript/reference/global_objects/array/fromasync/index.md
```diff
@@ -70,6 +79,31 @@ Array.fromAsync(asyncIterable).then((array) => console.log(array));
...
+    },
+  };
+}
+
+Array.fromAsync(createAsyncIter()).then((array) => console.log(array));
+// (3) [Promise, Promise, Promise]
+```
+
+> [!NOTE]
+> In practice, you would rarely encounter an async iterable that yields promises, because if you implement it using an [async generator function](/en-US/docs/Web/JavaScript/Reference/Statements/async_function*), then the [`yield`](/en-US/docs/Web/JavaScript/Reference/Operators/yield) expression automatically unwraps promises.
```
chrisdavidmills, line 105:
```suggestion
> In practice, you will rarely encounter an async iterable that yields promises, because if you implement it using an [async generator function](/en-US/docs/Web/JavaScript/Reference/Statements/async_function*), then the [`yield`](/en-US/docs/Web/JavaScript/Reference/Operators/yield) expression automatically unwraps promises.
```

chrisdavidmills (approved):
Reads pretty nicely, great work @Josh-Cena. I've included one small nitpick suggestion, but I'm approving it so you can merge at will.

PR #39709, opened 2025-05-27
Lint: replace string concatenation with template strings, replace unary plus with Number()

Josh-Cena:
Our style guide requires using template literals instead of string concatenation. This PR fixes all remaining cases.


files/en-us/web/api/document_object_model/examples/index.md
```diff
@@ -13,143 +13,111 @@ This chapter provides some longer examples of web and XML development using the
...
-    <p id="pid" onclick="window.location.href = 'http://www.cnn.com/';">
-      linker
-    </p>
-
-    <form>
-      <p><input value="rec" type="button" onclick="changeText();" /></p>
-    </form>
-  </body>
-</html>
+<form>
```
bsmth, line 101: Just noticed this is missing, although maybe this is better?

```suggestion
<p id="pid">
  Some text content
</p>
<form>
```

bsmth (commented):
Everything looks good to me except https://pr39709.review.mdn.allizom.net/en-US/docs/Web/API/Document_Object_Model/Examples which has some minor things to fix. Small tidy and we're good to go

bsmth (approved):
Thanks, I think everything's good now 👍🏻

PR #39712, opened 2025-05-27
chore(fx): pages missing text before first h2

bsmth:
### Description

These pages are missing text before the first h2


dipikabh (approved):
Thanks!

PR #39713, opened 2025-05-27
Correct variables in console.log

mikoMK:
The second and third `console.log()` in the first example on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant/from#examples use `instant` instead of `instant2`/`instant3`


Josh-Cena (approved):
Great catch. Thank you

PR #39729, opened 2025-05-29
Internalize role="timer" codepen demo

Josh-Cena:
This example is straightforward. I copied it over, modernized the code, and added some explanations.

### DIFF
--- a/files/en-us/web/accessibility/aria/reference/roles/timer_role/index.md
+++ b/files/en-us/web/accessibility/aria/reference/roles/timer_role/index.md
@@ -35,6 +35,64 @@ Along with [`alert`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_ro
 
 If a time limit needs to be in place, for example, for security reasons, the user should have the option to turn it off or extend it. This restriction does not apply if the time limit is due to a live event, such as an auction or a game, or if the time to complete the form is essential for a valid submission.
 
+## Examples
+
+### A basic timer
+
+This example has a timer that counts down from 30 seconds to 0 seconds. The whole time display region has `role="time"`, and also [`aria-atomic`](/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-atomic) to indicate that the region should be announced as a whole, not just the changed regions. Due to the implicit `aria-live="off"`, changes are not announced by default; we manually set it to `"assertive"` when the timer reaches 10 seconds remaining so it gets announced once.
+
+```html
+<div id="countdown" role="timer" aria-atomic="true">
+  <span id="number">30</span> seconds left!
+</div>
+```
+
+```css
+html {
+  font-size: 50px;
+  text-align: center;
+  margin-top: 1em;
+  font-family: sans-serif;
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Thanks!

files/en-us/web/accessibility/aria/reference/roles/timer_role/index.md
```diff
@@ -35,6 +35,64 @@ Along with [`alert`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_ro
 
 If a time limit needs to be in place, for example, for security reasons, the user should have the option to turn it off or extend it. This restriction does not apply if the time limit is due to a live event, such as an auction or a game, or if the time to complete the form is essential for a valid submission.
 
+## Examples
+
+### A basic timer
+
+This example has a timer that counts down from 30 seconds to 0 seconds. The whole time display region has `role="time"`, and also [`aria-atomic`](/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-atomic) to indicate that the region should be announced as a whole, not just the changed regions. Due to the implicit `aria-live="off"`, changes are not announced by default; we manually set it to `"assertive"` when the timer reaches 10 seconds remaining so it gets announced once.
```
scottaohara, line 42:
```suggestion
This example has a timer that counts down from 30 seconds to 0 seconds. The whole time display region has `role="timer"`, and also [`aria-atomic`](/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-atomic) to indicate that the region should be announced as a whole, not just the changed regions. Due to the implicit `aria-live="off"`, changes are not announced by default; we manually set it to `"assertive"` when the timer reaches 10 seconds remaining so it gets announced once.
```

files/en-us/web/accessibility/aria/reference/roles/timer_role/index.md
```diff
@@ -35,6 +35,64 @@ Along with [`alert`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_ro
...
+  const timeNow = new Date().getTime();
+  const elapsedTime = Math.floor((timeNow - startTime) / 1000);
+  let newNumber = 30 - elapsedTime;
+
+  if (newNumber >= 0) {
+    numElement.textContent = newNumber;
+  }
+
+  if (newNumber === 10) {
+    liveRegion.setAttribute("aria-live", "assertive");
```
scottaohara, line 82: Why aria live assertive and role=alert which is already a slightly more powerful aria-live assertive implicitly ?

scottaohara:
@Josh-Cena of course this gets merged as I find a typo and have a question in my review…

Josh-Cena:
Oh oops. Would send a follow up.

> Why aria live assertive and role=alert which is already a slightly more powerful aria-live assertive implicitly ?

I'm not sure; this is already in the example I copied over. Which would you reckon as redundant? Happy to make modifications.

scottaohara:
Aria-live isn’t needed if the element is also being changed to a role=alert

PR #39732, opened 2025-05-29
Editorial review: Update docs for Chrome 137 WebGPU additions

chrisdavidmills:
---

### Description

Chrome 137 supports the following WebGPU updates:

- When using [copyBufferToBuffer()](https://developer.mozilla.org/en-US/docs/Web/API/GPUCommandEncoder/copyBufferToBuffer), you can omit the offsets and the `size` parameter when copying the entire buffer. See https://developer.chrome.com/blog/new-in-webgpu-137?hl=en#buffers_copy_without_specifying_offsets_and_size.
- When creating a `GPUBindGroup` using [`createBindGroup()`](https://developer.mozilla.org/en-US/docs/Web/API/GPUDevice/createBindGroup), you can specify a `GPUTextureView` in place of a `GPUExternalTexture`. See https://developer.chrome.com/blog/new-in-webgpu-137?hl=en#use_texture_view_for_externaltexture_binding

This PR updates the docs to include details of both.

### DIFF
--- a/files/en-us/web/api/gpucommandencoder/copybuffertobuffer/index.md
+++ b/files/en-us/web/api/gpucommandencoder/copybuffertobuffer/index.md
@@ -53,14 +53,14 @@ The following criteria must be met when calling **`copyBufferToBuffer()`**, othe
 
 ## Examples
 
-In our [basic compute demo](https://mdn.github.io/dom-examples/webgpu-compute-demo/), we use `copyBufferToBuffer()` to copy the contents of our `output` buffer to the `stagingBuffer`.
+In our [basic compute demo](https://mdn.github.io/dom-examples/webgpu-compute-demo/), we use `copyBufferToBuffer()` to copy the contents of our `outputBuffer` to the `stagingBuffer`.
 
 ```js
 // …
 
 // Create an output buffer to read GPU calculations to, and a staging buffer to be mapped for JavaScript access
 
-const output = device.createBuffer({
+const outputBuffer = device.createBuffer({
   size: BUFFER_SIZE,
   usage: GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC,
 });
@@ -79,7 +79,7 @@ const commandEncoder = device.createCommandEncoder();
 
 // Copy output buffer to staging buffer
 commandEncoder.copyBufferToBuffer(
-  output,
+  outputBuffer,
   0, // Source offset
   stagingBuffer,
   0, // Destination offset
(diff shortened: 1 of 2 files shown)


files/en-us/web/api/gpucommandencoder/copybuffertobuffer/index.md
```diff
@@ -82,6 +86,9 @@ commandEncoder.copyBufferToBuffer(
   BUFFER_SIZE,
 );
 
+// Since we are copying the entire buffer, this can be shortened to
+// commandEncoder.copyBufferToBuffer(output, stagingBuffer);
```
beaufortfrancois: Nit: rename `output` to `outputBuffer` for consistency with `stagingBuffer`

beaufortfrancois (commented):
LGTM with nit

files/en-us/web/api/gpucommandencoder/copybuffertobuffer/index.md
chrisdavidmills: Done!

chrisdavidmills:
Do you mind giving this one an editorial review, @hamishwillee? It's a really tiny one.

files/en-us/web/api/gpucommandencoder/copybuffertobuffer/index.md
```diff
@@ -16,22 +16,26 @@ The **`copyBufferToBuffer()`** method of the
 ## Syntax
 
 ```js-nolint
+copyBufferToBuffer(source, destination)
```
hamishwillee, line 19: The spec says description does not indicate that size is optional in https://gpuweb.github.io/gpuweb/#commands-buffer-copies , but does indicate it in the IDL https://gpuweb.github.io/gpuweb/#gpucommandencoder

So, I **think** that there are two options, one for copying the whole buffer from 0, and one for copying part of the buffer from 0. I think the note below is still fine though, because it refers to copying the whole buffer.
Might be worth some comment on this case with size though.

```suggestion
copyBufferToBuffer(source, destination)
copyBufferToBuffer(source, destination, size)
```

hamishwillee (commented):
The docs if fine modulo any changes coming out of https://github.com/mdn/content/pull/39732/files#r2122324375

This looks like a BCD change right? I.e. previously this was not supported?
Can you link the BCD entry?

chrisdavidmills:
> This looks like a BCD change right? I.e. previously this was not supported? Can you link the BCD entry?

Yes. BCD entry is at https://github.com/mdn/browser-compat-data/pull/26935.

files/en-us/web/api/gpucommandencoder/copybuffertobuffer/index.md
chrisdavidmills: Yeah, good call. I had grokked this, but I wasn't sure it needed to be said. I think you are right, though; it's good to be explicit about all possible cases.

I've added the case to the syntax and added an explanation to the note.

I also tested this case to make sure it works.

hamishwillee (approved):
Looks good thanks!

PR #39676, opened 2025-05-26
FF140 Relnote: Cookie Store API

hamishwillee:
FF140 reenables the cookie store API in https://bugzilla.mozilla.org/show_bug.cgi?id=1958875

This adds the release note that was reverted in FF136 in https://github.com/mdn/content/pull/38867 (with minor modifications).

It also makes "tweaks" to the API overview.

Related docs work can be tracked in #39612

PR #39734, opened 2025-05-30
FF140 Relnote - CookieStore API update

hamishwillee:
This fixes the FF140 Cookie Store API release note updated in https://github.com/mdn/content/pull/39676.

BCD shows that the whole API is supported (the missing bits were implemented in the short window while this was unpublished and not tagged as dev-docs-needed). 

This updates the docs to match. Note that I believe the comment about the returned properties to still be true, but I'm querying that in https://bugzilla.mozilla.org/show_bug.cgi?id=1958875#c15

Related docs work can be tracked in #39612

### DIFF
--- a/files/en-us/web/api/extendablecookiechangeevent/index.md
+++ b/files/en-us/web/api/extendablecookiechangeevent/index.md
@@ -11,14 +11,11 @@ The **`ExtendableCookieChangeEvent`** interface of the {{domxref("Cookie Store A
 
 Cookie changes that cause the `ExtendableCookieChangeEvent` to be dispatched are:
 
-- A cookie is newly created and not immediately removed.
+- A cookie is newly created and not immediately removed, or if a cookies is replaced.
   In this case `type` is "changed".
 - A cookie is newly created and immediately removed.
   In this case `type` is "deleted"
 - A cookie is removed. In this case `type` is "deleted".
-
-> [!NOTE]
-> A cookie that is replaced due to the insertion of another cookie with the same name, domain, and path, is ignored and does not trigger a change event.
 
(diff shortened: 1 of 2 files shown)


files/en-us/web/api/extendablecookiechangeevent/index.md
```diff
@@ -7,17 +7,16 @@ browser-compat: api.ExtendableCookieChangeEvent
...
-- A cookie is newly created and not immediately removed. In this case `type` is "changed".
-- A cookie is newly created and immediately removed. In this case `type` is "deleted"
+- A cookie is newly created and not immediately removed, or if a cookies is replaced.
+  In this case `type` is "changed".
+- A cookie is newly created and immediately removed.
+  In this case `type` is "deleted"
 - A cookie is removed. In this case `type` is "deleted".
 
-> [!NOTE]
-> A cookie that is replaced due to the insertion of another cookie with the same name, domain, and path, is ignored and does not trigger a change event.
```
hamishwillee, line 19: According to the FF developer this is incorrect "any change is detected" https://bugzilla.mozilla.org/show_bug.cgi?id=1958875#c16.

This is in line with what the spec says, which is that any script visible change causes the event to be fired. 
So my take is that a replacement is script visible. However a create and immediate delete is perhaps not.

I have added the words "- A cookie is newly created and not immediately removed_, or if a cookies is replaced._". Above as well.

I haven't been able to get code working to test this unfortunately. I've spent the time I allocated myself to "extra due diligence" and so will trust the developer.

hamishwillee:
@pepelsbey Can you please review.

sideshowbarker (approved):

PR #39736, opened 2025-05-30
Update attacks nav

wbamberg:
This PR does some plumbing and updates around the new "Attacks" pages that we have:
- delete https://developer.mozilla.org/en-US/docs/Web/Security/Types_of_attacks and redirect to https://developer.mozilla.org/en-US/docs/Web/Security/Attacks
- delete the old pages on clickjacking and CSRF, and redirect to the new pages
- rewrite the glossary pages for CSRF and XSS
- update various other links to point to the new attacks pages.

The only bit of "Types of attacks" we're losing is the bit on session fixation. I wasn't too sure what to do about this. I don't think the existing text there is particularly helpful, but I'm not ready to write something about this yet. For now I've linked to OWASP instead for session fixation.


files/en-us/glossary/cross-site_scripting/index.md
```diff
@@ -6,13 +6,17 @@ page-type: glossary-definition
...
+A **cross-site scripting** (XSS) attack is one in which an attacker is able to get a target site to execute malicious code as though it was part of the website. The code can then do anything that the site's own code can do. For example, the attacker could:
 
-These attacks succeed if the Web app does not employ enough validation or encoding. The user's browser cannot detect the malicious script is untrustworthy, and so gives it access to any cookies, session tokens, or other sensitive site-specific information, or lets the malicious script rewrite the {{glossary("HTML")}} content.
+- Access and modify all the content of the site's loaded pages, and any content in local storage
+- Make HTTP requests with the user's credentials, enabling them to impersonate the user or access sensitive data
+
+All XSS attacks depend on a website doing two things:
+
+1. Accepting some input that could have been crafted by an attacker
+2. Including this input in a page without sanitizing it: that is, without ensuring that it won't be executable as JavaScript.
```
hamishwillee: Inconsistent full stops in bullets. The rules are "do for sentences, don't for phrases, make sure that all items in a list are either sentences or phrases. https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide#lists

```suggestion
2. Including this input in a page without sanitizing it: that is, without ensuring that it won't be executable as JavaScript
```

hamishwillee: PS But you are expected to capitalize the phrase (as done), which I like, but is not technically proper :-)

files/en-us/glossary/csrf/index.md
```diff
@@ -1,24 +1,22 @@
...
 
-This can be done, for example, by including malicious parameters in a {{glossary("URL")}} behind a link that purports to go somewhere else:
+A CSRF attack is possible if a website:
 
-```html
-<img src="https://www.example.com/index.php?action=delete&id=123" />
-```
+- Uses HTTP requests to change some state on the server.
+- Uses only cookies to validate that the request came from an authenticated user.
+- Uses only parameters in the request that an attacker can predict.
```
hamishwillee: And under rules as in previous note, presumably

```suggestion
- Uses HTTP requests to change some state on the server
- Uses only cookies to validate that the request came from an authenticated user
- Uses only parameters in the request that an attacker can predict
```

files/en-us/web/http/guides/cookies/index.md
```diff
@@ -219,14 +219,14 @@ If no `SameSite` attribute is set, the cookie is treated as `Lax` by default.
 
 Because of the design of the cookie mechanism, a server can't confirm that a cookie was set from a secure origin or even tell _where_ a cookie was originally set.
 
-A vulnerable application on a subdomain can set a cookie with the `Domain` attribute, which gives access to that cookie on all other subdomains. This mechanism can be abused in a _session fixation_ attack. See [session fixation](/en-US/docs/Web/Security/Types_of_attacks#session_fixation) for primary mitigation methods.
+A vulnerable application on a subdomain can set a cookie with the `Domain` attribute, which gives access to that cookie on all other subdomains. This mechanism can be abused in a [session fixation](https://owasp.org/www-community/attacks/Session_fixation) attack.
```
hamishwillee: Maybe vulnerable can be assumed

```suggestion
An application on a subdomain can set a cookie with the `Domain` attribute, which gives access to that cookie on all other subdomains. This mechanism can be abused in a [session fixation](https://owasp.org/www-community/attacks/Session_fixation) attack.
```

hamishwillee (approved):
@wbamberg I think this is good and have approved.

BUT, I'd like to ask that you consider (_even though you don't like it_) pulling the existing (more or less) session fixation docs into an attack doc under the Security/Attacks page, and cross link that. Then add an issue to address that page later.

Reasoning 
- Security/Attacks is a pretty good summation of the main attacks (as I understand them) omitting only session fixation. This feels like an omission, at least in the attacks page
- I don't want this to fall off the list of things to appear on that page.
- The existing text isn't really clear, but I get the gist. We're no worse off having it there.

An alternative would be a paragraph in the Attacks page with an unliked heading.

Something to think about

wbamberg:
> @wbamberg I think this is good and have approved.
> 
> BUT, I'd like to ask that you consider (_even though you don't like it_) pulling the existing (more or less) session fixation docs into an attack doc under the Security/Attacks page, and cross link that. Then add an issue to address that page later.
> 

Yeah I did waver about doing something like that but I worried about the amount of extra work it would be to get that into a state that I was happy with. Unless I really do just copy it as-is, I think it's a lot. And then I'd really have to park this PR until all that is resolved.

> Reasoning
> 
> * Security/Attacks is a pretty good summation of the main attacks (as I understand them) omitting only session fixation. This feels like an omission, at least in the attacks page
> 

Well... maybe? Only session fixation? What about say brute-forcing passwords, or phishing, or intercepting OTP over SMS? I'd expect to include those in a guide on authentication, and (I think) I'd expect to include session fixation in a guide on session management. I'm not sure why I would expect that, rather than include it in "attacks".

> * I don't want this to fall off the list of things to appear on that page.
> 
> * The existing text isn't really clear, but I get the gist. We're no worse off having it there.
> 
> 
> An alternative would be a paragraph in the Attacks page with an unliked heading.
> 
> Something to think about

hamishwillee:
@wbamberg I was proposing a "copy as is" - not "try to make it good" (hence the need to create an issue to follow up).

You make a good point that there are lots of other alternatives - I hadn't thought of any of those. Personally I'd quite like it if those were on the page too - and if it is too much work to create pages, then unlinked headings could still be useful. 
But as you say at the top of the page "common attacks" not "exhaustive list"

It is not a tragedy if we lose the session fixation stuff or these other things - I'm  just saying what I would do. I approved it because I don't feel very strongly about this point, but I do feel strongly that this PR is overall really good.

wbamberg:
OK, then here's what I'd like to do.

1. Merge this as is.
2. Right away file a new PR to resurrect the "session fixation" as a separate attack. I'll try to keep my urge to rewrite it under control.
3. File an issue to find a permanent home and form for a session fixation article.

That way I get to merge this tonight, and I know it won't block on anything else that comes up, but we're sure not to lose the session fixation stuff.

PR #39739, opened 2025-05-31
New guide: mask shorthand

estelle:
i am working on a much deeper guide about all the properties, so that link is red in this PR, but since this is done and that is taking a while, PRing this now.


files/en-us/web/css/css_masking/index.md
```diff
@@ -68,6 +68,10 @@ While masking provides more control and options, clipping can perform better if
 
   - : Introduction to masking in CSS, the various mask image types, and the effects of luminance and alpha-transparency in masking.
 
+- [Declaring multiple masks](/en-US/docs/Web/CSS/CSS_masking/Multiple_masks)
+
+  - : Understanding mask layers and declaring multiple mask images.
```
dipikabh, line 73: To make this parallel to the previous two guide descriptions
```suggestion
  - : Introduction to mask layers and how to declare multiple mask images.
```

files/en-us/web/css/css_masking/multiple_masks/index.md
```diff
@@ -0,0 +1,181 @@
...
+
+When including multiple mask images, you can define the way the [mask layers are composited](#the_mask-composite_property), or combined. While each mask image is [positioned](/en-US/docs/Web/CSS/CSS_masking/Mask_properties#the_mask-position_property) relative to an [origin](/en-US/docs/Web/CSS/CSS_masking/Mask_properties#the_mask-origin_property) box, and can be [sized](/en-US/docs/Web/CSS/CSS_masking/Mask_properties#the_mask-size_property), [repeated](/en-US/docs/Web/CSS/CSS_masking/Mask_properties#the_mask-repeat_property), and [clipped](/en-US/docs/Web/CSS/CSS_masking/Mask_properties#the_mask-clip_property), those features are described in the [masking properties guide](/en-US/docs/Web/CSS/CSS_masking/Mask_properties). This guide is focused on the syntax for multiple mask images.
+
+## CSS masks
```
dipikabh, line 13:
```suggestion
## Understanding masks in CSS
```

files/en-us/web/css/css_masking/multiple_masks/index.md
```diff
@@ -0,0 +1,181 @@
...
+
+The syntax is:
+
+`<image> <position> / <size> <repeat> <origin> <clip> <composite> <mode>`
+
+For the most part, the order of the properties is flexible, but there are a few quirks and exceptions.
+
+The `mask-origin` and `mask-clip` values are similar. Both accept [`<geometry-box>`](/en-US/docs/Web/CSS/box-edge#geometry-box) values, with `mask-clip` also supporting `no-clip`. If one `<geometry-box>` value and the `no-clip` keyword are present then the `<geometry-box>` sets `mask-origin` value and `mask-clip` is set to `no-clip`. In this case the order doesn't matter. If only one `<geometry-box>` value is present (and there is no `no-clip`), but the `mask-origin` and `mask-clip` components are set to that value. As there is only one value, no order matters here either. The order matters only if two `<geometry-box>` values are present; in this case, the first `<geometry-box>` sets the `mask-origin` component and the second `<geometry-box>` value is the `mask-clip` component.
```
dipikabh, line 152:
```suggestion
For the most part, the order of the properties is flexible, but there are a few quirks and exceptions:

- The `mask-origin` and `mask-clip` values are similar. Both accept [`<geometry-box>`](/en-US/docs/Web/CSS/box-edge#geometry-box) values, with `mask-clip` also supporting `no-clip`.
  - If one `<geometry-box>` value and the `no-clip` keyword are present, then the `<geometry-box>` sets `mask-origin` value and `mask-clip` is set to `no-clip`. In this case, the order doesn't matter.
  - If only one `<geometry-box>` value is present (and there is no `no-clip`), but the `mask-origin` and `mask-clip` components are set to that value. As there is only one value, no order matters here either.
- The order matters only if two `<geometry-box>` values are present; in this case, the first `<geometry-box>` sets the `mask-origin` component and the second `<geometry-box>` value is the `mask-clip` component.
```

dipikabh (changes requested):
Thanks for adding this guide, @estelle!

I've done an initial readthrough and have a few requests before I send a more detailed review:
- Could you do a quick proofread to fix some obvious typos, like "Each mask layers consists of...", "...as it's value", "...creates a mask layers", "impoorted" in a link text, missing comma after "In this guide". "The minimum required..." should probably be "The minimum requirement".
- The first sentence is quite long and could be split up for better readability and clarity.
- I really like the masquerade ball analogy, but I wonder if "masquerade balls" is understood universally. Do you think we can use another example?
- The page is quite long. Can we consider trimming (removing) parts that are covered in other guides; for example:
  - Could we remove the explanation of what masks are and the analogy to the intro guide and link to it from here?
  - The first four paragraphs can be compressed to briefly set the context and then jump into the main topic of declaring multiple masks. I feel we're not getting to it quickly enough.
- The section headings can be more descriptive; for example, "Overview of the `mask-image` property" and "Declaring mask images using the `mask` shorthand".
- The "Layer syntax" section feels like the core of this guide. Should that be at the `##` level? And possibly renamed to "Understanding the mask layer syntax"? 
- Also wondering about structure: should we introduce the layer syntax before covering all the individual component properties, and then explain the `mask` constituent properties in relation to that syntax?
- The "Component properties" section need not follow a definition list format. This could be converted to a bulleted list and maybe include just the salient points.
- Should a link to this guide also be added on the related property pages, the ones that are listed in the "Layer syntax" section?

PR #39742, opened 2025-06-01
Fix/css color syntax

tarak6984:
---


Josh-Cena:
I fail to see how this PR changes anything in the direction of an improvement. You changed one example to demonstrate what it's not designed to demonstrate, and changed the syntax section to stray away from our page template.

PR #39743, opened 2025-06-01
docs: enhance HSL color function documentation

tarak6984:
- Expanded the 'Legacy versus modern syntax' section with clear examples
- Added side-by-side comparison of modern and legacy syntax
- Included a 'Key differences' section highlighting the main changes
- Added a recommendation to use modern syntax for new projects

---


Josh-Cena:
You are sticking substantial explanation within examples, which is an antipattern in itself. What you are writing here is already sufficiently covered in the description above, especially in the syntax section. Your addition also strongly gives off LLM vibes; I think you should be aware that all the work you submit should be typed by yourself.

PR #39748, opened 2025-06-02
fix: typo in how_long_is_too_long

ujjwal313:
### Description

Fixes a typo in "How Long is too Long" section in Web Performance.


Josh-Cena (commented):
The original is correct. "Find something" is followed by an adjective, like "find this movie exciting". "Speedy" is an adjective.

PR #39758, opened 2025-06-03
fix: change format of title to match other element pages

ghost:
---

### Description
I changed the formatting of the title to match the other element pages and wrapped it in double quotes.

### Motivation
To make it follow the standard set by the other element pages, it caught me off guard.


Josh-Cena:
Hi, none of the MathML elements use this format; you are confusing MathML with HTML. If you want to propose this as the new convention, you would need to write up a proposal first.

PR #39759, opened 2025-06-03
Editorial review: Add docs for view-transition-name: match-element

chrisdavidmills:
---

### Description

Chrome 137 supports the `view-transition-name` property's `match-element` value. See https://chromestatus.com/feature/5098745710247936.

This PR:

- Adds information on the new value to the `view-transition-name` reference page and rewrites the page to provide better structure.
- Updates the relevant "Using..." guide to add appropriate mentions of the new property value, and improves the structure.
- Updates the API landing page to link to my new `match-element` demo, which is being added to `dom-examples` via https://github.com/mdn/dom-examples/pull/311.

### DIFF
--- a/files/en-us/web/api/view_transition_api/index.md
+++ b/files/en-us/web/api/view_transition_api/index.md
@@ -65,7 +65,7 @@ See [Using the View Transition API](/en-US/docs/Web/API/View_Transition_API/Usin
 ### Properties
 
 - {{cssxref("view-transition-name")}}
-  - : Provides the selected element with a separate identifying name and causes it to participate in a separate view transition from the root view transition — or no view transition if the `none` value is specified.
+  - : Provides the selected element with a separate identifying name and causes it to participate in a separate animation from the root snapshot — or no animation if the `none` value is specified.
 - {{cssxref("view-transition-class")}}
   - : Provides an additional method of styling to selected elements that have a `view-transition-name`.
 
--- a/files/en-us/web/api/view_transition_api/index.md
+++ b/files/en-us/web/api/view_transition_api/index.md
@@ -86,8 +86,9 @@ See [Using the View Transition API](/en-US/docs/Web/API/View_Transition_API/Usin
 
 - [Basic View Transitions SPA demo](https://mdn.github.io/dom-examples/view-transitions/spa/): A basic image gallery demo with view transitions, featuring separate animations between old and new images, and old and new captions.
 - [Basic View Transitions MPA demo](https://mdn.github.io/dom-examples/view-transitions/mpa/): A sample two-page site that demonstrates usage of cross-document (MPA) view transitions, providing a custom "swipe up" transition when the two pages are navigated between.
+- [View Transitions match-element demo](https://mdn.github.io/dom-examples/view-transitions/match-element/): An SPA featuring animated list items, which demonstrates usage of the `view-transition-name` property's `match-element` value.
 - [HTTP 203 playlist](https://http203-playlist.netlify.app/): A video player demo app that features several different SPA view transitions, many of which are explained in [Smooth transitions with the View Transition API](https://developer.chrome.com/docs/web-platform/view-transitions/).
-- [List of Chrome DevRel team members](https://view-transitions.chrome.dev/profiles/mpa/): A basic team profile pages app that demonstrates how to use the {{domxref("Window.pagereveal_event", "pagereveal")}} and {{domxref("Window.pageswap_event", "pageswap")}} events to customize the outgoing and inbound animations of a cross-document view transition based on the "from" and "to" URLs.
+- [Chrome DevRel view transitions demos](https://view-transitions.chrome.dev/): A series of View Transition API demos.
 
 ## Specifications
 
--- a/files/en-us/web/api/view_transition_api/index.md
+++ b/files/en-us/web/api/view_transition_api/index.md
@@ -65,9 +65,9 @@ See [Using the View Transition API](/en-US/docs/Web/API/View_Transition_API/Usin
 ### Properties
 
 - {{cssxref("view-transition-name")}}
-  - : Provides the selected element with a separate identifying name and causes it to participate in a separate animation from the root snapshot — or no animation if the `none` value is specified.
+  - : Specifies the view transition snapshot group that selected elements will participate in, which enables an element to be animated separately from the rest of the page during a view transition.
 - {{cssxref("view-transition-class")}}
-  - : Provides an additional method of styling to selected elements that have a `view-transition-name`.
+  - : Provides an additional method of styling selected elements that have a `view-transition-name`.
 
 ### Pseudo-elements
 
--- a/files/en-us/web/api/view_transition_api/index.md
+++ b/files/en-us/web/api/view_transition_api/index.md
@@ -65,7 +65,7 @@ See [Using the View Transition API](/en-US/docs/Web/API/View_Transition_API/Usin
 ### Properties
 
 - {{cssxref("view-transition-name")}}
-  - : Specifies the view transition snapshot group that selected elements will participate in, which enables an element to be animated separately from the rest of the page during a view transition.
+  - : Specifies the view transition snapshot that selected elements will participate in, which enables an element to be animated separately from the rest of the page during a view transition.
 - {{cssxref("view-transition-class")}}
   - : Provides an additional method of styling selected elements that have a `view-transition-name`.
 
--- a/files/en-us/web/api/view_transition_api/using/index.md
+++ b/files/en-us/web/api/view_transition_api/using/index.md
@@ -17,7 +17,7 @@ Let's walk through the process by which a view transition works:
    - In the case of cross-document transitions (MPAs), a view transition is triggered by initiating navigation to a new document. Both the current and destination documents of the navigation need to be on the same origin, and opt-in to the view transition by including a {{cssxref("@view-transition")}} at rule in their CSS with a `navigation` descriptor of `auto`.
      > [!NOTE]
      > An active view transition has an associated {{domxref("ViewTransition")}} instance (for example, returned by `startViewTransition()` in the case of same-document (SPA) transitions). The `ViewTransition` object includes several promises, allowing you to run code in response to different parts of the view transition process being reached. See [Controlling view transitions with JavaScript](#controlling_view_transitions_with_javascript) for more information.
-2. On the current (old page) view, the API captures snapshots of elements that have a {{cssxref("view-transition-name")}} declared on them.
+2. On the current (old page) view, the API captures static image **snapshots** of elements that have a {{cssxref("view-transition-name")}} declared on them.
 3. The view change occurs:
 
    - In the case of same-document transitions (SPAs), the callback passed to `startViewTransition()` is invoked, which causes the DOM to change.
@@ -26,7 +26,7 @@ Let's walk through the process by which a view transition works:
 
    - In the case of cross-document transitions (MPAs), the navigation occurs between the current and destination documents.
 
-4. The API captures snapshots from the new view as a live representation.
+4. The API captures "live" snapshots (meaning, interactive DOM regions) from the new view.
 
    At this point, the view transition is about to run, and the {{domxref("ViewTransition.ready")}} promise fulfills, allowing you to respond by running a custom JavaScript animation instead of the default, for example.
 
@@ -35,6 +35,14 @@ Let's walk through the process by which a view transition works:
 
 > [!NOTE]
 > If the document's [page visibility state](/en-US/docs/Web/API/Page_Visibility_API) is `hidden` (for example if the document is obscured by a window, the browser is minimized, or another browser tab is active) during a {{domxref("Document.startViewTransition()", "document.startViewTransition()")}} call, the view transition is skipped entirely.
+
+### An aside on snapshots
+
+It is worth noting that when talking about view transitions, we commonly use the term snapshot to refer to a part of the page that has a `view-transition-name` declared on it and will therefore be animated separately from other parts of the page with different `view-transition-name` values set on them. As you've read above, the process of animating a snapshot via a view transition actually involves two separate snapshots of the old and new UI states, but we use snapshot to refer to the whole page area for simplicity.
+
+The snapshot of the old UI state is a static image, so that the user can't interact with it as it animates "out".
+
+The snapshot of the new UI state is an interactive DOM region, so that the user can start to interact with the new content as it animates "in".
 
 ### The view transition pseudo-element tree
 
@@ -48,15 +56,12 @@ To handle creating the outbound and inbound transition animations, the API const
       └─ ::view-transition-new(root)
 ```
 
-> [!NOTE]
-> A {{cssxref("::view-transition-group()")}} subtree is created for every captured `view-transition-name`.
-
 In the case of same-document transitions (SPAs), the pseudo-element tree is made available in the document. In the case of cross-document transitions (MPAs), the pseudo-element tree is made available in the destination document only.
 
 The most interesting parts of the tree structure are as follows:
 
-- {{cssxref("::view-transition")}} is the root of view transitions overlay, which contains all view transition snapshot groups and sits over the top of all other page content.
-- A {{cssxref("::view-transition-group()")}} acts as a container for each view transition snapshot group. The `root` argument specifies the default snapshot group — the view transition animation will apply to the snapshot whose `view-transition-name` is `root`. By default, this is the {{cssxref(":root")}} element, because the default browser styles define this:
+- {{cssxref("::view-transition")}} is the root of view transitions overlay, which contains all view transition groups and sits over the top of all other page content.
+- A {{cssxref("::view-transition-group()")}} acts as a container for each view transition snapshot. The `root` argument specifies the default snapshot — the view transition animation will apply to the snapshot whose `view-transition-name` is `root`. By default, this is a snapshot of the {{cssxref(":root")}} element, because the default browser styles define this:
 
   ```css
   :root {
--- a/files/en-us/web/api/view_transition_api/using/index.md
+++ b/files/en-us/web/api/view_transition_api/using/index.md
@@ -66,7 +66,7 @@ The most interesting parts of the tree structure are as follows:
 
   Be aware however that page authors can change this by unsetting the above, and setting `view-transition-name: root` on a different element.
 
-- {{cssxref("::view-transition-old()")}} targets the static snapshot of the old page element, and {{cssxref("::view-transition-new()")}} targets the live snapshot of the new page element. Both of these render as replaced content, in the same manner as an {{htmlelement("img")}} or {{htmlelement("video")}}, meaning that they can be styled with handy properties like {{cssxref("object-fit")}} and {{cssxref("object-position")}}.
+- {{cssxref("::view-transition-old()")}} targets the static snapshot of the old page element, and {{cssxref("::view-transition-new()")}} targets the live snapshot of the new page element. Both of these render as replaced content, in the same manner as an {{htmlelement("img")}} or {{htmlelement("video")}}, meaning that they can be styled with properties like {{cssxref("object-fit")}} and {{cssxref("object-position")}}.
 
 > [!NOTE]
 > It is possible to target different DOM elements with different custom view transition animations by setting a different {{cssxref("view-transition-name")}} on each one. In such cases, a `::view-transition-group()` is created for each one. See [Different animations for different elements](#different_animations_for_different_elements) for an example.
@@ -156,7 +156,7 @@ It is recommended that you target the `::view-transition-group()` with such styl
 > [!NOTE]
 > This is also a good option for safeguarding your code — `::view-transition-group()` also animates and you could end up with different durations for the `group`/`image-pair` pseudo-elements versus the `old` and `new` pseudo-elements.
 
-In the case of cross-document (MPA) transitions, the pseudo-elements need to be included in the destination document only for the view transition to work. If you want to use the view transition in both directions, you'll need to include it in both, of course.
+In the case of cross-document (MPA) transitions, the pseudo-elements need to be included in the destination document only for the view transition to work. If you want to use the view transition in both directions, you'll need to include it in both.
 
 Our [View Transitions MPA demo](https://mdn.github.io/dom-examples/view-transitions/mpa/) includes the above CSS, but takes the customization a step further, defining custom animations and applying them to the `::view-transition-old(root)` and `::view-transition-new(root)` pseudo-elements. The result is that the default cross-fade transition is swapped out for a "swipe up" transition when navigation occurs:
 
@@ -274,7 +274,7 @@ The `view-transition-name` property can take a unique {{cssxref("custom-ident")}
 
 It can also take keyword values of:
 
-- `none`: Causes the element to not participate in any view transitions.
+- `none`: Causes the element to not participate in a view transitions.
 - `match-element`: Automatically sets unique `view-transition-name` values on all selected elements.
 
 ### Taking advantage of the default animation styles
--- a/files/en-us/web/api/view_transition_api/using/index.md
+++ b/files/en-us/web/api/view_transition_api/using/index.md
@@ -80,7 +80,11 @@ This section illustrates how to create a basic view transition, in both the SPA 
 
 ### Basic SPA view transition
 
-As an example, an SPA may include functionality to fetch new content and update the DOM in response to an event of some kind, such as a navigation link being clicked or an update being pushed from the server. In our [View Transitions SPA demo](https://mdn.github.io/dom-examples/view-transitions/spa/) we've simplified this to a `displayNewImage()` function that shows a new full-size image based on the thumbnail that was clicked. We've encapsulated this inside an `updateView()` function that only calls the View Transition API if the browser supports it:
+An SPA may include functionality to fetch new content and update the DOM in response to an event of some kind, such as a navigation link being clicked or an update being pushed from the server.
+
+Our [View Transitions SPA demo](https://mdn.github.io/dom-examples/view-transitions/spa/) is a basic image gallery. We have a series of {{htmlelement("a")}} elements that contain thumbnail {{htmlelement("img")}} elements, dynamically generated using JavaScript. We also have a {{htmlelement("figure")}} element containing a {{htmlelement("figcaption")}} and an `<img>`, which displays the full-size gallery images.
+
+When a thumbnail is clicked, the `displayNewImage()` function is run via {{domxref("Document.startViewTransition()")}}, which causes the full-size image and its associated caption to be displayed inside the `<figure>`. We've encapsulated this inside an `updateView()` function that only calls the View Transition API if the browser supports it:
 
 ```js
 function updateView(event) {
@@ -216,11 +220,6 @@ With this CSS applied, the generated pseudo-element tree will now look like this
 
 The existence of the second set of pseudo-elements allows separate view transition styling to be applied just to the `<figcaption>`. The different old and new view captures are handled separately from one another.
 
-> [!NOTE]
-> The value of `view-transition-name` is a unique {{cssxref("custom-ident")}}; it can be any identifier that wouldn't be misinterpreted as a keyword. The keyword `none` is not a valid name, as that value means the element will not participate in any view transitions. Also avoid `auto`, as it's being discussed as a way to [determine `view-transition-name` automatically](https://drafts.csswg.org/css-view-transitions-2/#auto-vt-name).
->
-> `view-transition-name` values must also be unique. If two rendered elements have the same `view-transition-name` at the same time, {{domxref("ViewTransition.ready")}} will reject and the transition will be skipped.
-
 The following code applies a custom animation just to the `<figcaption>`:
 
 ```css
@@ -268,6 +267,15 @@ Here we've created a custom CSS animation and applied it to the `::view-transiti
 >   animation-duration: 2s;
 > }
 > ```
+
+### Valid `view-transition-name` values
+
+The `view-transition-name` property can take a unique {{cssxref("custom-ident")}} value, which can be any identifier that wouldn't be misinterpreted as a keyword. The value of `view-transition-name` for each rendered element must be unique. If two rendered elements have the same `view-transition-name` at the same time, {{domxref("ViewTransition.ready")}} will reject and the transition will be skipped.
+
+It can also take keyword values of:
+
+- `none`: Causes the element to not participate in any view transitions.
+- `match-element`: Automatically sets unique `view-transition-name` values on all selected elements.
 
 ### Taking advantage of the default animation styles
 
--- a/files/en-us/web/api/view_transition_api/using/index.md
+++ b/files/en-us/web/api/view_transition_api/using/index.md
@@ -38,7 +38,7 @@ Let's walk through the process by which a view transition works:
 
 ### An aside on snapshots
 
-It is worth noting that when talking about view transitions, we commonly use the term snapshot to refer to a part of the page that has a `view-transition-name` declared on it and will therefore be animated separately from other parts of the page with different `view-transition-name` values set on them. As you've read above, the process of animating a snapshot via a view transition actually involves two separate snapshots of the old and new UI states, but we use snapshot to refer to the whole page area for simplicity.
+It is worth noting that when talking about view transitions, we commonly use the term _snapshot_ to refer to a part of the page that has a `view-transition-name` declared on it. These sections will be animated separately from other parts of the page with different `view-transition-name` values set on them. While the process of animating a snapshot via a view transition actually involves two separate snapshots—one of the old and one of the new UI states—we use snapshot to refer to the whole page area for simplicity.
 
 The snapshot of the old UI state is a static image, so that the user can't interact with it as it animates "out".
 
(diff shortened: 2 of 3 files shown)


files/en-us/web/css/view-transition-name/index.md
```diff
@@ -7,7 +7,7 @@ browser-compat: css.properties.view-transition-name
 
 {{CSSRef}}
 
-The **`view-transition-name`** [CSS](/en-US/docs/Web/CSS) property provides the selected element with a distinct identifying name (a {{cssxref("custom-ident")}}) and causes it to participate in a separate [view transition](/en-US/docs/Web/API/View_Transition_API) from the root view transition — or no view transition if the `none` value is specified.
+The **`view-transition-name`** [CSS](/en-US/docs/Web/CSS) property specifies the [view transition](/en-US/docs/Web/API/View_Transition_API) group that selected elements will appear in, which enables an element to be animated using a separate view transition to the default (root) transition, or not animated at all.
```
bramus: This is not exactly correct:

- The `view-transition-name` specifies that it will be snapshotted (the word “group” is a bit vague/confusing – the “participate in” from before is more clear).
- At all times, there is only 1 single View Transition happening. The VT contains multiple snapshotted elements.

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,17 +18,87 @@ view-transition-name: figure-caption;
...
+view-transition-name: match-element;
 ```
 
 ### Values
 
 - {{cssxref("custom-ident")}}
-  - : An identifying name that causes the selected element to participate in a separate [view transition](/en-US/docs/Web/API/View_Transition_API) from the root view transition. The identifier must be unique. If two rendered elements have the same `view-transition-name` at the same time, {{domxref("ViewTransition.ready")}} will reject and the transition will be skipped.
-    > [!NOTE]
-    > The `<custom-ident>` cannot be `auto`.
+  - : An identifying name that causes the selected element to participate in a separate [view transition](/en-US/docs/Web/API/View_Transition_API) from the root view transition. The `<custom-ident>` cannot be `auto`, `match-element`, or `none`.
```
bramus: Same remark as before: it’s not a separate VT, but 1 VT with multiple snapshotted elements.

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,17 +18,87 @@ view-transition-name: figure-caption;
...
+- `match-element`
+  - : The selected element is automatically given an unique identifying name based on the type of element, which enables it to be transitioned separately from all other elements on the page. The name is an internal token representing the element, which is not visible to the web document.
 - `none`
   - : The selected element will not participate in a view transition.
 
+> [!NOTE]
+> The specification also mentions an `auto` keyword value, which isn't yet supported in any browser.
```
bramus: Safari 18.2+ supports `auto`. The property value is being disputed at the CSSWG, though.

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,17 +18,87 @@ view-transition-name: figure-caption;
...
+::view-transition-old(figure-caption) {
+  animation: 0.25s linear both shrink-x;
+}
+
+::view-transition-new(figure-caption) {
+  animation: 0.25s 0.25s linear both grow-x;
+}
+```
+
+If you don't want an element to be animated during a view transition, you can specify a `view-transition-name` value of `none`:
```
bramus: I believe this is more correct:

```suggestion
If you don't want an element to be snapshotted separately, you can specify a `view-transition-name` value of `none`:
```

This because the element can still be snapshotted as part of a parent element with a `view-transition-name`

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,17 +18,87 @@ view-transition-name: figure-caption;
...
+
+```css
+li:nth-child(1) { view-transition-name: item1; }
+li:nth-child(2) { view-transition-name: item2; }
+li:nth-child(3) { view-transition-name: item3; }
+li:nth-child(4) { view-transition-name: item4; }
+...
+```
+
+To get around this problem, you can use the `match-element` value, which causes the browser to give each selected element a unique internal `view-transition-name` based on its type:
```
bramus: What do you mean by “based on its type”?

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,17 +18,87 @@ view-transition-name: figure-caption;
...
+
+```css
+li {
+  view-transition-name: match-element;
+}
+```
+
+Because of technical limitations, `match-element` can only be used for single-page view transitions. The auto-generated internal identifiers are not transferrable across different documents.
```
bramus: Suggestion to be more explicit and state the “technical limitations”:

```suggestion
Because `match-element` assigns automatic `view-transition-name` values based on element identity, it can only be used for single-page view transitions. The auto-generated internal identifiers are not transferrable across different elements or documents.
```

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: OK, thanks for the commentary here. I'm pretty sure I've heard other people referring to "groups" or "snapshot groups" to describe the different snapshots, but I've updated the language through to stay away from this terminology as much as possible. Let me know if you still have concerns.

I've also updated the wording to make sure I'm not talking about multiple view transitions.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: Updated

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: I've decided to remove the mention of `auto` for now in that case. I'll revisit it later on, and document it separately.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: OK. Updated. I've also updated the description of `none` to mention this caveat.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: I was kind of guessing, as "identity" sounds vague to me; I decided maybe it means based on type (as in, type selector), but it sounds like I'm going down the wrong path here. I've updated the wording, in both places it appeared.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: Updated; thanks!

files/en-us/web/api/view_transition_api/index.md
```diff
@@ -65,7 +65,7 @@ See [Using the View Transition API](/en-US/docs/Web/API/View_Transition_API/Usin
 ### Properties
 
 - {{cssxref("view-transition-name")}}
-  - : Provides the selected element with a separate identifying name and causes it to participate in a separate view transition from the root view transition — or no view transition if the `none` value is specified.
+  - : Provides the selected element with a separate identifying name and causes it to participate in a separate animation from the root snapshot — or no animation if the `none` value is specified.
```
bramus: When set to `none` it does not mean it won’t animate, but it means that it won’t be snapshotted individually. It could still be captured (and animated)  as part of a parent element with a `view-transition-name`.

files/en-us/web/api/view_transition_api/using/index.md
```diff
@@ -269,6 +268,15 @@ Here we've created a custom CSS animation and applied it to the `::view-transiti
 > }
 > ```
 
+### Valid `view-transition-name` values
+
+The `view-transition-name` property can take a unique {{cssxref("custom-ident")}} value, which can be any identifier that wouldn't be misinterpreted as a keyword. The value of `view-transition-name` for each rendered element must be unique. If two rendered elements have the same `view-transition-name` at the same time, {{domxref("ViewTransition.ready")}} will reject and the transition will be skipped.
+
+It can also take keyword values of:
+
+- `none`: Causes the element to not participate in a view transitions.
```
bramus:
```suggestion
- `none`: The element will not be snapshotted individually as part of a view transition.
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
...
-    > [!NOTE]
-    > The `<custom-ident>` cannot be `auto`.
+  - : An identifying name that causes the selected element to participate in a separate snapshot from the root snapshot. The `<custom-ident>` cannot be `auto`, `match-element`, or `none`.
+- `match-element`
+  - : The selected element is automatically given an unique identifying name based on its identity, which enables it to be animated separately from all other elements on the page. The name is an internal token representing the element, which is not visible to the web document.
 - `none`
-  - : The selected element will not participate in a view transition.
+  - : The selected element will not participate in a view transition, unless it has a parent element with a `view-transition-name` set, in which case it would be snapshotted as part of that element.
```
bramus:
```suggestion
  - : The selected element will not participate in a view transition, unless it has a parent element with a `view-transition-name` set, in which case it would be snapshotted as part of that parent element.
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
...
+
+To get around this problem, you can use the `match-element` value, which causes the browser to give each selected element a unique internal `view-transition-name`:
+
+```css
+li {
+  view-transition-name: match-element;
+}
+```
+
+Because `match-element` assigns automatic `view-transition-name` values based on element identity, it can only be used for single-page view transitions. The auto-generated internal identifiers are not transferrable across different elements or documents.
```
bramus:
```suggestion
Because `match-element` assigns automatic `view-transition-name` values based on element identity, it can only be used for same-document view transitions. The auto-generated internal identifiers are not transferrable across different elements or documents.
```

bramus (commented):
Some final nits

files/en-us/web/api/view_transition_api/index.md
chrisdavidmills: OK, I think I finally get this. Thanks for your patience ;-)

I've updated this to:

> Specifies the view transition snapshot group that selected elements will participate in, which enables an element to be animated separately from the rest of the page during a view transition.

This reads the same as on the property page now.

files/en-us/web/api/view_transition_api/using/index.md
chrisdavidmills: I updated this to:

> Causes the element to not participate in a separate snapshot, unless it has a parent element with a `view-transition-name` set, in which case it will be snapshotted as part of that element.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: I don't think this is necessary; we say "parent" earlier in the sentence.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: updated!

chrisdavidmills:
Thanks for the review, @bramus! I'm now moving this to the editorial review stage.

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
...
+view-transition-name: match-element;
 ```
 
 ### Values
 
 - {{cssxref("custom-ident")}}
-  - : An identifying name that causes the selected element to participate in a separate [view transition](/en-US/docs/Web/API/View_Transition_API) from the root view transition. The identifier must be unique. If two rendered elements have the same `view-transition-name` at the same time, {{domxref("ViewTransition.ready")}} will reject and the transition will be skipped.
-    > [!NOTE]
-    > The `<custom-ident>` cannot be `auto`.
+  - : An identifying name that causes the selected element to participate in a separate snapshot from the root snapshot. The `<custom-ident>` cannot be `auto`, `match-element`, or `none`.
```
estelle:
```suggestion
  - : An identifying name that causes the selected element to participate in a separate snapshot from the root snapshot. The `<custom-ident>` cannot be `auto`, `match-element`, `none`, or a [CSS-wide keyword](/en-US/docs/Web/CSS/CSS_Values_and_Units/CSS_data_types#css-wide_keywords) value.
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
 
 /* Keyword value */
 view-transition-name: none;
+view-transition-name: match-element;
```
estelle, line 21:
```suggestion
view-transition-name: match-element;

/* Global values */
view-transition-name: inherit;
view-transition-name: initial;
view-transition-name: revert;
view-transition-name: revert-layer;
view-transition-name: unset;
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
...
 
 ### Values
 
 - {{cssxref("custom-ident")}}
-  - : An identifying name that causes the selected element to participate in a separate [view transition](/en-US/docs/Web/API/View_Transition_API) from the root view transition. The identifier must be unique. If two rendered elements have the same `view-transition-name` at the same time, {{domxref("ViewTransition.ready")}} will reject and the transition will be skipped.
-    > [!NOTE]
-    > The `<custom-ident>` cannot be `auto`.
+  - : An identifying name that causes the selected element to participate in a separate snapshot from the root snapshot. The `<custom-ident>` cannot be `auto`, `match-element`, or `none`.
+- `match-element`
+  - : The selected element is automatically given an unique identifying name based on its identity, which enables it to be snapshotted separately from all other elements on the page. The name is an internal token representing the element, which is not visible to the web document.
```
estelle:
```suggestion
  - : The selected element is automatically given a unique identifying name. This name enables the element to be snapshotted separately from all other elements on the page and is not visible to the web document.
```
I didn't understand this, and it seems to be more information than necessary; so shrank it down to what I think is correct

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
...
 - `none`
-  - : The selected element will not participate in a view transition.
+  - : The selected element will not participate in a separate snapshot, unless it has a parent element with a `view-transition-name` set, in which case it will be snapshotted as part of that element.
+
+## Description
+
+By default, when a view transition is applied to a web app, all changes to the UI that occur during that transition are snapshotted and animated together, which is the default — or `root` — snapshot (see [The view transition pseudo-element tree](/en-US/docs/Web/API/View_Transition_API/Using#the_view_transition_pseudo-element_tree)). By default, this animation is a smooth cross-fade, which can be seen in action in our [View Transitions SPA demo](https://mdn.github.io/dom-examples/view-transitions/spa/).
```
estelle:
```suggestion
By default, when a view transition is applied to a web app, all changes to the UI that occur during that transition are snapshotted and animated together. This is the default — or `root` — snapshot (see [The view transition pseudo-element tree](/en-US/docs/Web/API/View_Transition_API/Using#the_view_transition_pseudo-element_tree)). By default, this animation is a smooth cross-fade, which can be seen in action in the [View Transitions SPA demo](https://mdn.github.io/dom-examples/view-transitions/spa/).
```

estelle: not sure what `root` means? Is that a custom ident, or `::root`, or something else? If it is just the root element, remove the code ticks (here and in the next paragraph)

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
...
+
+If you want certain elements to be animated differently from the `root` snapshot during the view transition, you can do so by giving them a different `view-transition-name`, for example:
+
+```css
+figcaption {
+  view-transition-name: figure-caption;
+}
+```
+
+You can then specify what animation you want for the before and after snapshots using the relevant view transition pseudo-elements — {{cssxref("::view-transition-old()")}} and {{cssxref("::view-transition-new()")}}. For example:
```
estelle:
```suggestion
You can then specify which animations you want for the before and after snapshots using the relevant view transition pseudo-elements — {{cssxref("::view-transition-old()")}} and {{cssxref("::view-transition-new()")}}. For example:
```

a before and after, so two animations at least, please it can be multiple for each

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -7,7 +7,7 @@ browser-compat: css.properties.view-transition-name
 
 {{CSSRef}}
 
-The **`view-transition-name`** [CSS](/en-US/docs/Web/CSS) property provides the selected element with a distinct identifying name (a {{cssxref("custom-ident")}}) and causes it to participate in a separate [view transition](/en-US/docs/Web/API/View_Transition_API) from the root view transition — or no view transition if the `none` value is specified.
+The **`view-transition-name`** [CSS](/en-US/docs/Web/CSS) property specifies the [view transition](/en-US/docs/Web/API/View_Transition_API) snapshot group that selected elements will participate in, which enables an element to be animated separately from the rest of the page during a view transition.
```
estelle: what is a "snapshot group", or "snapshot"

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
...
+::view-transition-old(figure-caption) {
+  animation: 0.25s linear both shrink-x;
+}
+
+::view-transition-new(figure-caption) {
+  animation: 0.25s 0.25s linear both grow-x;
+}
+```
+
+If you don't want an element to be snapshotted separately, you can specify a `view-transition-name` value of `none`:
```
estelle, line 64: it's not clear to me what snapshot means. This may need to be updated (or this may be ok if it's clear above)

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
...
+<ul>
+  <li>Item 1</li>
+  <li>Item 2</li>
+  <li>Item 3</li>
+  <li>Item 4</li>
+  ...
+</ul>
+```
+
+The `view-transition-name` `<custom-ident>` must be unique for each rendered element taking part in the view transition. If two rendered elements have the same `view-transition-name` at the same time, {{domxref("ViewTransition.ready")}} will reject and the transition will be skipped. Giving each one a unique name can be inconvenient, especially as the number of elements gets larger:
```
estelle:
```suggestion
The `view-transition-name` `<custom-ident>` must be unique for each rendered element taking part in the view transition. If two rendered elements have the same `view-transition-name` at the same time, {{domxref("ViewTransition.ready")}} will reject and the transition will be skipped. Sometimes you may want to animate multiple UI elements separately in a view transition. Giving each one a unique name can be inconvenient. especially as the number of elements gets larger. For example, if you have a list of elements on a page and want to rearrange them in some way, giving each list item a unique `view-transition-name` could lead to excessive declarations:
```
Suggest moving the intro sentence from above into this paragraph, which means the `<ul>` should be moved down to if you go with this edit.

suggestion: 

```html
<ul>
  <li>Item 1</li>
  <li>Item 2</li>
...
  <li>Item 499</li>
  <li>Item 500</li>
</ul>
```
if you want (makes it very clear that it's very long, realizing you did the same thing but with more subtlety.

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -18,16 +18,83 @@ view-transition-name: figure-caption;
...
+
+To get around this problem, you can use the `match-element` value, which causes the browser to give each selected element a unique internal `view-transition-name`:
+
+```css
+li {
+  view-transition-name: match-element;
+}
+```
+
+Because `match-element` assigns automatic `view-transition-name` values based on element identity, it can only be used for same-document view transitions. The auto-generated internal identifiers are not transferrable across different elements or documents.
```
estelle, line 122: I think you wrote everything here that was in the value section, but in case... put anything i yeeted up there down here.

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -39,12 +106,190 @@ view-transition-name: none;
 
 ## Examples
 
+### Basic `view-transition-name` usage
+
+Our [View Transitions SPA demo](https://mdn.github.io/dom-examples/view-transitions/spa/), is a basic image gallery. See [Basic SPA view transition](/en-US/docs/Web/API/View_Transition_API/Using#basic_spa_view_transition) for a more detailed explanation of how this demo works.
```
estelle:
```suggestion
This example comes from the [View Transitions SPA demo](https://mdn.github.io/dom-examples/view-transitions/spa/), which is a basic image gallery. The [Basic SPA view transition](/en-US/docs/Web/API/View_Transition_API/Using#basic_spa_view_transition) provides a more detailed explanation of how this demo works.
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -39,12 +106,190 @@ view-transition-name: none;
...
+
+Most of the UI changes are animated using the `root` transition snapshot. However, the `<figcaption>` is given a `view-transition-name` of `figure-caption` to allow it to be animated differently from the rest of the page:
```
estelle, line 138: root here too

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -39,12 +106,190 @@ view-transition-name: none;
...
+::view-transition-old(figure-caption) {
+  animation: 0.25s linear both shrink-x;
+}
+
+::view-transition-new(figure-caption) {
+  animation: 0.25s 0.25s linear both grow-x;
+}
+```
+
+Here we've created a custom CSS animation and applied it to the `::view-transition-old(figure-caption)` and `::view-transition-new(figure-caption)` pseudo-elements. We've also added several other styles to both to keep them in the same place and stop the default styling from interfering with our custom animations.
```
estelle:
```suggestion
We create a custom CSS animation and apply it to the `::view-transition-old(figure-caption)` and `::view-transition-new(figure-caption)` pseudo-elements. We also apply other styles to keep them both in the same place and to stop the default styling from interfering with our custom animations.
```
make current tense, rather than past, and move this up to precede the code

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -39,12 +106,190 @@ view-transition-name: none;
...
+
+### Using the `match-element` value
+
+Our [View Transitions match-element demo](https://mdn.github.io/dom-examples/view-transitions/match-element/) contains a list of content items in a sidebar with a large main content space beside it. The headings inside the list items can be clicked to cause them to animate into the main content space and show all their content.
```
estelle:
```suggestion
The [View Transitions match-element demo](https://mdn.github.io/dom-examples/view-transitions/match-element/) contains a list of content items in a sidebar with a large main content space beside it. The headings inside the list items can be clicked, which will cause them to animate into the main content space and show all their content.
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -39,12 +106,190 @@ view-transition-name: none;
...
+
+#### HTML
+
+The markup features a {{htmlelement("main")}} element, which contains an [unordered list](/en-US/docs/Web/HTML/Reference/Elements/ul) and an {{htmlelement("article")}} element. Each child {{htmlelement("li")}} element inside the list contains text content, which includes an {{htmlelement("a")}} element inside a [heading](/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements).
```
estelle:
```suggestion
The {{htmlelement("main")}} element contains an [unordered list](/en-US/docs/Web/HTML/Reference/Elements/ul) and an {{htmlelement("article")}} element. The multiple child {{htmlelement("li")}} elements inside the list each contain an {{htmlelement("a")}} element inside a [heading](/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements).
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -39,12 +106,190 @@ view-transition-name: none;
...
+.active-item {
+  position: absolute;
+  z-index: 1;
+  translate: 310px;
+  width: calc(100% - 310px);
+  height: 100%;
+}
+```
+
+By default, all the rendered elements involved in the view transition will be animated together, in one cross-fade. However, in this case, we don't want this — we want to give each list item an individual movement animation. We can achieve this by applying `view-transition-name: match-element` to the list items:
```
estelle:
```suggestion
By default, all the rendered elements involved in the view transition are animated together in one cross-fade. However, in this case, we don't want this — we want to give each list item an individual movement animation. We can achieve this by applying `view-transition-name: match-element` to every list item:
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -39,12 +106,190 @@ view-transition-name: none;
...
+
+```css
+.match-element-applied li {
+  view-transition-name: match-element;
+}
+```
+
+The `match-element-applied` class is applied to the `<main>` element. You can remove this class by unchecking the checkbox in the bottom-right corner of the UI. This allows you to compare the singular cross-fade animation you get without `view-transition-name: match-element` to the individual movement animations you get when it is applied.
```
estelle:
```suggestion
The `match-element-applied` class is applied to the `<main>` element. When looking at the live demo, you can remove this class by unchecking the checkbox in the bottom-right corner of the UI. This allows you to compare the singular cross-fade animation you get without `view-transition-name: match-element` to the individual movement animations you get when it is applied.
```

estelle: link "live demo"

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -39,12 +106,190 @@ view-transition-name: none;
...
+
+We also apply a specific {{cssxref("animation-duration")}} to all view transition groups (signified by the `*` identifier) using the {{cssxref("::view-transition-group()")}} pseudo-element, and give all the old and new snapshots a {{cssxref("height")}} of `100%` to work around differences in their aspect ratios and make the animations look smoother:
```
estelle:
```suggestion
We also use the {{cssxref("::view-transition-group()")}} pseudo-element to apply an {{cssxref("animation-duration")}} to all the view transition groups (signified by the `*` identifier) and give all the old and new snapshots a {{cssxref("height")}} of `100%` to work around differences in their aspect ratios and make the animations look smoother:
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -39,12 +106,190 @@ view-transition-name: none;
...
+  } else {
+    // Run updateActiveItem() via startViewTransition()
+    const transition = document.startViewTransition(() =>
+      updateActiveItem(event),
+    );
+  }
+});
+```
+
+Running the `updateActiveItem()` function via the `startViewTransition()` function causes the UI changes to be animated smoothly.
```
estelle:
```suggestion
Running the `updateActiveItem()` function via the `startViewTransition()` function causes the UI changes to animate smoothly.
```

estelle (commented):
Most of the suggested edits are minor, but there are a couple of areas that could benefit from a bit more explanation

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: The `root` custom ident is put in place by the browser to provide a default view transition snapshot of the whole page. It is equivalent to setting `view-transition-name: root` on the `::root` element. This is explained in the linked "Using..." guide section.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: This should probably be "snapshot" rather than "snapshot group". When a part of the rendered DOM has a `view-transition-name` set on it, and a JS function is run that updates that part of the DOM (as a callback function inside a startViewTransition() call), a snapshot is taken of it. Each snapshot is actually two separate representations of the same UI region:

- A static image of the old UI state (so it is not interactive as it animates out)
- A live representation of the new UI state (so it can be interacted with as it animates in)

These two can then be animated separately. The default animation for the root snapshot is a cross fade - old UI representation fades out, and new UI representation fades in.

This is explained in the "Using..." guide.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: See above

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: I've gone through the "Using..." guide and made some clarifications and added some explanatory text to make all of this clearer.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: I ended up putting the bit about all `view-transition-name`s needing to be unique at the end of the previous section, as really that's relevant to the whole description, not just this section.

I ended up adding "Item 99" to the code snippets, as I liked the idea of making it more obvious how significant the problem could be.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: As I read it now, I think we're all good.

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: It's an actual `view-transition-name` set by the browser, so I think it is OK to have it in code?

files/en-us/web/css/view-transition-name/index.md
chrisdavidmills: Good call. Added!

files/en-us/web/api/view_transition_api/using/index.md
```diff
@@ -36,6 +36,14 @@ Let's walk through the process by which a view transition works:
 > [!NOTE]
 > If the document's [page visibility state](/en-US/docs/Web/API/Page_Visibility_API) is `hidden` (for example if the document is obscured by a window, the browser is minimized, or another browser tab is active) during a {{domxref("Document.startViewTransition()", "document.startViewTransition()")}} call, the view transition is skipped entirely.
 
+### An aside on snapshots
+
+It is worth noting that when talking about view transitions, we commonly use the term snapshot to refer to a part of the page that has a `view-transition-name` declared on it and will therefore be animated separately from other parts of the page with different `view-transition-name` values set on them. As you've read above, the process of animating a snapshot via a view transition actually involves two separate snapshots of the old and new UI states, but we use snapshot to refer to the whole page area for simplicity.
```
estelle:
```suggestion
It is worth noting that when talking about view transitions, we commonly use the term _snapshot_ to refer to a part of the page that has a `view-transition-name` declared on it. These sections will be animated separately from other parts of the page with different `view-transition-name` values set on them. While the process of animating a snapshot via a view transition actually involves two separate snapshots—one of the old and one of the new UI states—we use snapshot to refer to the whole page area for simplicity.
```

files/en-us/web/api/view_transition_api/using/index.md
```diff
@@ -48,15 +56,12 @@ To handle creating the outbound and inbound transition animations, the API const
...
-> [!NOTE]
-> A {{cssxref("::view-transition-group()")}} subtree is created for every captured `view-transition-name`.
-
 In the case of same-document transitions (SPAs), the pseudo-element tree is made available in the document. In the case of cross-document transitions (MPAs), the pseudo-element tree is made available in the destination document only.
 
 The most interesting parts of the tree structure are as follows:
 
-- {{cssxref("::view-transition")}} is the root of view transitions overlay, which contains all view transition snapshot groups and sits over the top of all other page content.
-- A {{cssxref("::view-transition-group()")}} acts as a container for each view transition snapshot group. The `root` argument specifies the default snapshot group — the view transition animation will apply to the snapshot whose `view-transition-name` is `root`. By default, this is the {{cssxref(":root")}} element, because the default browser styles define this:
+- {{cssxref("::view-transition")}} is the root of view transitions overlay, which contains all view transition groups and sits over the top of all other page content.
```
estelle:
```suggestion
- {{cssxref("::view-transition")}} is the root of the view transitions overlay, which contains all view transition groups and sits above all other page content.
```

files/en-us/web/css/view-transition-name/index.md
```diff
@@ -69,6 +69,8 @@ If you don't want an element to be snapshotted separately, you can specify a `vi
 }
 ```
 
+The `view-transition-name` `<custom-ident>` must be unique for each rendered element taking part in the view transition. If two rendered elements have the same `view-transition-name` at the same time, {{domxref("ViewTransition.ready")}} will reject and the transition will be skipped.
```
estelle: the word "reject" seems incorrect. Maybe:

```suggestion
The `view-transition-name` `<custom-ident>` must be unique for each rendered element taking part in the view transition. If two rendered elements have the same `view-transition-name` at the same time, the {{domxref("ViewTransition.ready")}} {{JSxRef("Promise")}} will reject and the transition will be skipped.
```
in the API docs, the original likely makes sense, but in the CSS docs, explaining a bit should help

estelle (approved):
just minor suggestions. approving to not hold you up.

chrisdavidmills:
@estelle super; thank you so much!

PR #39768, opened 2025-06-03
Fixed Capitalisation in "closedBy" attribute

JB-ea:
The "B" was lowercase

---


Josh-Cena:
HTML attributes are always lowercase. Perhaps you are looking for https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/closedBy instead.

JB-ea:
Oh my bad, Sorry for wasting your time.

PR #39771, opened 2025-06-03
Replace deprecated substr with slice

karelbilek:
string.substr() is deprecated; string.slice() is not

---

### Description

This replaces substr, which is deprecated, with slice, which is not deprecated.

### Motivation

Substr is deprecated, slice is not deprecated.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

### DIFF
--- a/files/en-us/web/api/fetch_api/using_fetch/index.md
+++ b/files/en-us/web/api/fetch_api/using_fetch/index.md
@@ -503,7 +503,7 @@ async function* makeTextFileLineIterator(fileURL) {
     const result = newline.exec(chunk);
     if (!result) {
       if (readerDone) break;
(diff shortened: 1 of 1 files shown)


wbamberg (approved):
👍 thank you @karelbilek !

PR #39773, opened 2025-06-03
Add missing argument for the fromOffsetsToSelection function

rubywwwilde:
I tried to use the provided code, but it throws an error. According to source code, editor element should be passed as the third  argument: https://github.com/mdn/dom-examples/blob/3f316b6be2e29596b1dfb2e82e358167cef6c372/edit-context/html-editor/editor.js#L106

---

### Description

I tried to use the provided code, but it throws an error. According to source code, editor element should be passed as the third  argument

### Motivation

This would allow them use examples without debugging them

### Additional details

https://github.com/mdn/dom-examples/blob/3f316b6be2e29596b1dfb2e82e358167cef6c372/edit-context/html-editor/editor.js#L106


wbamberg (approved):
👍 thank you!

PR #39775, opened 2025-06-03
Add and update Japanese translations for the CSS sidebar

mfuji09:
---

### Description

Add and update Japanese translations for the CSS sidebar

### Motivation

I want to show the sidebar in Japanese.

### Additional details

The translate word for "Containment" has been changed because the container query has been included and the old translate word isn't fit for the module's functionary.


dipikabh (approved):
Thanks a lot!

PR #39520, opened 2025-05-14
39303 ship temporal ff release

dletorey:
### Description

- Added Firefox 139 release note for Temporal API
- Added partial implementation note for Temporal API on experimental features

### Motivation

- Working on [issue #39303](https://github.com/mdn/content/issues/39303)

### Related issues and pull requests

- [Content PR](https://github.com/mdn/content/pull/39471)
- [Automate BCD PR](https://github.com/mdn/browser-compat-data/pull/26753)
- [exceptions BCD PR](https://github.com/mdn/browser-compat-data/pull/26758)

PR #39776, opened 2025-06-03
chore(fx): Temporal no longer experimental in Fx

bsmth:
### Motivation

Removing shipped features from experimental page.

- [x] https://github.com/mdn/content/pull/39520


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #39778, opened 2025-06-03
Fix security sidebar: remove "types of attacks" entry

wbamberg:
https://github.com/mdn/content/pull/39736 deleted "Types of attacks", this PR deletes the corresponding sidebar entry.


dipikabh (approved):
Thanks!

I thought I'd see this is in the current sidebar, but this entry is already not there or at least I couldn't find it.

PR #39779, opened 2025-06-03
Fix uncapitalized letter S in start of paragraph

helpimnotdrowning:
Found while browsing the Elements list

---

### Description

Fix uncapitalized letter S in start of paragraph

### Motivation

Found while browsing the Elements list a few minutes ago


estelle (approved):
Thanks! 🎉

PR #39787, opened 2025-06-04
Remove redundant aria-live changes in timer example

Josh-Cena:
Address review in https://github.com/mdn/content/pull/39729

### DIFF
--- a/files/en-us/web/accessibility/aria/reference/roles/timer_role/index.md
+++ b/files/en-us/web/accessibility/aria/reference/roles/timer_role/index.md
@@ -39,7 +39,7 @@ If a time limit needs to be in place, for example, for security reasons, the use
 
 ### A basic timer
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/accessibility/aria/reference/roles/timer_role/index.md
```diff
@@ -80,10 +80,8 @@ function decrement() {
 
   if (newNumber === 10) {
     liveRegion.setAttribute("aria-live", "assertive");
-    liveRegion.setAttribute("role", "alert");
```
scottaohara, line 83: Per my comments in the original pr, I’d have kept the role switching rather then the aria live attribute changes. 

The alert role has better support for announcing upon its injection than the aria-live attribute

scottaohara (approved):
Thanks for making the updates

bsmth:
Thanks, both 👍🏻

PR #39790, opened 2025-06-04
Add <noscript> scripting enabled behaviour note

chronoDave:
### Description

The [behaviour]((https://html.spec.whatwg.org/multipage/scripting.html#the-noscript-element:the-head-element-7)) of the `<noscript>` element when scripting is enabled is undocumented.

`<noscript>` behaves differently to other elements when scripting is enabled, where it returns a [`Text node`](https://developer.mozilla.org/en-US/docs/Web/API/Text) instead of [`Element nodes`](https://developer.mozilla.org/en-US/docs/Web/API/Element).

### Motivation

```TS
const noscript = document.querySelector('noscript');
noscript.replaceWith(...noscript.children);
```

I was expecting the noscript element to be replaced with it's children but instead found the noscript element removed. Looking at the [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/noscript) I could not find any mention of this behaviour and had to look for it in the [HTML Standard](https://html.spec.whatwg.org/multipage/scripting.html#the-noscript-element:the-head-element-7).

For that reason I think it's worth adding a note that describes this behaviour.

### Additional details

 - [\<noscript\>: The Noscript element (MDN)](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/noscript)
 - [4.12.2 The noscript element (HTML Standard)](https://html.spec.whatwg.org/multipage/scripting.html#the-noscript-element)

### DIFF
--- a/files/en-us/web/html/reference/elements/noscript/index.md
+++ b/files/en-us/web/html/reference/elements/noscript/index.md
@@ -37,8 +37,8 @@ Rocks!
 
 The `<noscript>` element represents its children differently depending on if scripting is enabled:
 
- - If scripting is disabled the `<noscript>` element represents its children as [HTML content](/en-US/docs/Web/API/HTMLElement).
- - If scripting is enabled, the `<noscript>` element represents its children as [text](/en-US/docs/Web/API/Text).
+- If scripting is disabled the `<noscript>` element represents its children as [HTML content](/en-US/docs/Web/API/HTMLElement).
+- If scripting is enabled, the `<noscript>` element represents its children as [text](/en-US/docs/Web/API/Text).
 
 ## Technical summary
 
--- a/files/en-us/web/html/reference/elements/noscript/index.md
+++ b/files/en-us/web/html/reference/elements/noscript/index.md
@@ -27,6 +27,9 @@ This element only includes the [global attributes](/en-US/docs/Web/HTML/Referenc
 
 Rocks!
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


estelle:
I agree that this page needs more information, but am not convinced that the note suggested is the best explanation of the behavior you are seeking to point out to the reader. Maybe instead adding something like:

```md
If scripting is disabled, the element represents its content. If within the `<head>`, the elements contents will be parsed as regular HTML meta data. If within the `<body>`,  the contents of the `<noscript>` element will be rendered as regular HTML content. 
```
then an explanation of how the browser handles the content in the head and body

chronoDave:
I moved the note into a separate section "Usage notes" similiar to [\<source\>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/source), the note did look out of place.

I don't think it's neccessary to describe head / body interaction as that's already covered in the [technical summary](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/noscript#technical_summary) under "Permitted content".

Not entirely sure if "HTML content" and "text" should be links, please let me know what you think.

hamishwillee (dismissed):
Thanks, and welcome to MDN! 

EDIT, I've removed my initial approval - not because I think this is bad or anything, but because I note that Estelle is looking at this and may have issues she wants to address.

estelle (approved):
Thank you!

Congratulations on your first merged MDN content PR, and sorry for the delay. 

Welcome to the team!

PR #39792, opened 2025-06-04
Add note about `beforeinstallprompt` event not fired if PWA already installed

Zwyx:
---

### Description

Add note about `beforeinstallprompt` event not fired if the PWA is already installed, so users don't think using `getInstalledRelatedApps` is required.

### Motivation

I wasted time digging into `getInstalledRelatedApps` and was disappointed that it wasn't supported by Chrome on desktop, only to realise thanks to https://web.dev/learn/pwa/installation-prompt that we actually don't need it :-)

### Additional details

https://web.dev/learn/pwa/installation-prompt

### DIFF
--- a/files/en-us/web/progressive_web_apps/how_to/trigger_install_prompt/index.md
+++ b/files/en-us/web/progressive_web_apps/how_to/trigger_install_prompt/index.md
@@ -52,7 +52,7 @@ The event handler here does three things:
 
 Note that the event will not fire if:
 
-- The PWA is already installed (valid only for desktop and WebAPK on Android).
+- The PWA is already installed.
 - The app does not pass the [PWA installation criteria](/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable#installability).
 - The PWA is not installable on the current device (for example, because of a lack of permissions).
 
--- a/files/en-us/web/progressive_web_apps/how_to/trigger_install_prompt/index.md
+++ b/files/en-us/web/progressive_web_apps/how_to/trigger_install_prompt/index.md
@@ -52,7 +52,7 @@ The event handler here does three things:
 
 Note that the event will not fire if:
 
-- The PWA is aleardy installed (valid only for desktop and WebAPK on Android).
+- The PWA is already installed (valid only for desktop and WebAPK on Android).
 - The app does not pass the PWA installation criteria.
 - The PWA is not installable on the current device (for example, because of a lack of permissions).
 
--- a/files/en-us/web/progressive_web_apps/how_to/trigger_install_prompt/index.md
+++ b/files/en-us/web/progressive_web_apps/how_to/trigger_install_prompt/index.md
@@ -50,6 +50,12 @@ The event handler here does three things:
 - Take a reference to the event object that's passed into the handler. This is an instance of {{domxref("BeforeInstallPromptEvent")}}, and is what will enable us to prompt the user to install the app.
 - Reveal our in-app install UI by removing the `hidden` attribute on the button.
 
+Note that the event will not fire if:
+
+- The PWA is aleardy installed (valid only for desktop and WebAPK on Android).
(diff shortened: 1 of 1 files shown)


files/en-us/web/progressive_web_apps/how_to/trigger_install_prompt/index.md
```diff
@@ -50,6 +50,12 @@ The event handler here does three things:
 - Take a reference to the event object that's passed into the handler. This is an instance of {{domxref("BeforeInstallPromptEvent")}}, and is what will enable us to prompt the user to install the app.
 - Reveal our in-app install UI by removing the `hidden` attribute on the button.
 
+Note that the event will not fire if:
+
+- The PWA is aleardy installed (valid only for desktop and WebAPK on Android).
```
chrisdavidmills:
```suggestion
- The PWA is already installed (valid only for desktop and WebAPK on Android).
```

Also, I'm not sure what "valid only for desktop and WebAPK on Android" means — does it mean it will only be blocked from firing if the app is already installed, on desktop and Android browsers? Or does it mean that it will only be blocked from firing if the app is already installed, if it was installed on desktop, or installed from a PWA APK file on Android?

files/en-us/web/progressive_web_apps/how_to/trigger_install_prompt/index.md
```diff
@@ -50,6 +50,12 @@ The event handler here does three things:
...
+- The app does not pass the PWA installation criteria.
```
chrisdavidmills:
```suggestion
- The app does not pass the [PWA installation criteria](/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable#installability).
```

Let's link to our installation criteria section.

chrisdavidmills (changes requested):
Thanks for the addition, @Zwyx — this is useful. I've got a couple of review comments for you to consider.

Zwyx:
Yes the `valid only for desktop and WebAPK on Android` isn't clear indeed. It's mentioned at https://web.dev/learn/pwa/installation-prompt 

I have tested, on both desktop and Android, that the event is not fired if the app is installed.

So I think this implies that the event might still be fired on iOS, but I'm not sure and I can't test it. So I didn't rephrase this part.

If you or someone else is able to test it on iOS, it would be very interesting and it's super easy: just install https://mlearn-pwa-web-app-install-prompt.glitch.me (which is the demo app from https://web.dev/learn/pwa/installation-prompt).

chrisdavidmills:
> Yes the `valid only for desktop and WebAPK on Android` isn't clear indeed. It's mentioned at https://web.dev/learn/pwa/installation-prompt
> 
> I have tested, on both desktop and Android, that the event is not fired if the app is installed.
> 
> So I think this implies that the event might still be fired on iOS, but I'm not sure and I can't test it. So I didn't rephrase this part.
> 
> If you or someone else is able to test it on iOS, it would be very interesting and it's super easy: just install https://mlearn-pwa-web-app-install-prompt.glitch.me (which is the demo app from https://web.dev/learn/pwa/installation-prompt).

I've just tested it on Chrome iOS on my daughter's phone, and the `beforeinstallprompt` event isn't supported there at all...which I guess makes sense as it will be running on WebKit in the background?

So, I'm not 100% sure what to do here. I'd be inclined to just remove that line, as it is confusing, and the line is true for both Chrome desktop and Android from your tests?

Zwyx:
Thank you for testing it :-) But yes indeed, it must be Safari because of the [Apple browser ban](https://open-web-advocacy.org/apple-browser-ban/). Some countries have started to force Apple to stop doing this though, so some people should be able to have actually Chrome on iOS now.

But yes, we can remove the line :+1: I've just done it.

chrisdavidmills (approved):
Cool, thanks @Zwyx! Let's get this merged.

PR #39795, opened 2025-06-04
Fixed view-transition-class exampled

BeBoRE:
---

### Description

As stated here https://developer.mozilla.org/en-US/docs/Web/CSS/::view-transition-old#pt-class-selector to select the view-transition-class it needs to be preceded by a dot.

### Motivation

I tried to use the example and it did not work.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/view-transition-class


estelle (approved):

PR #39798, opened 2025-06-04
Clarify guide around nonces with inline scripts or styles

robhudson:
---

### Description

Updated the CSP guide to clarify that nonces are only intended or useful for inline scripts and styles.

### Motivation

CSP nonces don't impact scripts or styles that that have a source specified. So this guide was a bit misleading.

### DIFF
--- a/files/en-us/web/http/guides/csp/index.md
+++ b/files/en-us/web/http/guides/csp/index.md
@@ -160,7 +160,7 @@ Content-Security-Policy:
   script-src 'nonce-416d1177-4d12-4e3b-b7c9-f6c409789fb8'
 ```
 
-The server then includes this value as the value of the `nonce` attribute of all the `<script>` and/or `<style>` tags that they intend to include in the document.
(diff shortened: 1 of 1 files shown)


hamishwillee (commented):
Thanks @robhudson , I am not sure this is correct, or at least it adds a different set of confusions.  My understanding is that you can quite happily use a nonce for an external script, and you might do so if you don't want to explicitly whitelist the location of the script for any reason. I think it also lets you allow dynamic addition of scripts from trusted third party scripts.

However my understanding is that you're right that you're far less likely to do this for an external script. The main benefit of the nonce to allow inline scripts to run when needed, that would otherwise be blocked by your CSP - which largely exists to prevent inline scripts running.

I think this goes too far. It effectively says you can't use nonces on external scripts. 

However I'm not certain so I'm inviting @wbamberg to have a look since he was the last person to work with this text.

robhudson:
Yes, you're right. I think I mentally separated the two in my mind... inlined scripts can be allowed via nonce, linked scripts can be allowed by adding to the `script-src`. The spec reads as if a nonce on a linked script shouldn't work but I just did a test across various browsers and a nonce does allow it to load a linked script. So I'll close this. Apologies for the confusion. (I did test this before filing this PR but had a bug in my test case I discovered. Oops)

hamishwillee:
Thanks @robhudson - absolutely fine!

PR #39799, opened 2025-06-04
Correct documentation around non-shared views to `Atomics.notify()`

Renegade334:
### Description

Unlike the wait methods, `Atomics.notify()` does not throw when passed a view on a non-shared array buffer. The MDN documentation mostly alludes to this behaviour correctly, but incorrectly states that it will throw a `TypeError` if passed such an argument.

### Additional details

https://tc39.es/ecma262/#sec-atomics.notify


Josh-Cena (approved):
Oh, good catch, thank you

Issue #16120, opened 2022-05-17
Remove jsfiddle/jsbin/codepen links


wbamberg:
We'd prefer not to have JSFiddle links on MDN because it's hard for us to maintain this content.

Here's a list of all the places we're linking to JSFidddles (I think). Ideally, we would replace there with live samples. Note that many of them are in the Games area - I'm not sure if it's worth investing the time to remove these ones..

- [x] https://developer.mozilla.org/en-US/docs/games/techniques/2d_collision_detection
- [x] https://developer.mozilla.org/en-US/docs/games/techniques/3d_on_the_web/building_up_a_basic_demo_with_aframe
- [x] https://developer.mozilla.org/en-US/docs/games/techniques/3d_on_the_web/building_up_a_basic_demo_with_babylon.js
- [x] https://developer.mozilla.org/en-US/docs/games/techniques/3d_on_the_web/building_up_a_basic_demo_with_playcanvas/engine
- [x] https://developer.mozilla.org/en-US/docs/games/techniques/3d_on_the_web/building_up_a_basic_demo_with_three.js
- [x] https://developer.mozilla.org/en-US/docs/games/techniques/audio_for_web_games
- [ ] https://developer.mozilla.org/en-US/docs/games/tutorials/2d_breakout_game_phaser/animations_and_tweens
- [ ] https://developer.mozilla.org/en-US/docs/games/tutorials/2d_breakout_game_phaser/bounce_off_the_walls
- [ ] https://developer.mozilla.org/en-US/docs/games/tutorials/2d_breakout_game_phaser/build_the_brick_field
- [ ] https://developer.mozilla.org/en-US/docs/games/tutorials/2d_breakout_game_phaser/buttons
- [ ] https://developer.mozilla.org/en-US/docs/games/tutorials/2d_breakout_game_phaser/collision_detection
- [ ] https://developer.mozilla.org/en-US/docs/games/tutorials/2d_breakout_game_phaser/extra_lives
- [ ] https://developer.mozilla.org/en-US/docs/games/tutorials/2d_breakout_game_phaser/game_over
- [ ] https://developer.mozilla.org/en-US/docs/games/tutorials/2d_breakout_game_phaser/initialize_the_framework
- [ ] https://developer.mozilla.org/en-US/docs/games/tutorials/2d_breakout_game_phaser/load_the_assets_and_print_them_on_screen
[...]

PR #39802, opened 2025-06-05
Remove external script dependency & jsfiddle in 2D collision detection

Josh-Cena:
Remove existing examples, which use an obscure library, and replace them with the jsfiddle ones, modernized and completely refactored.

Part of https://github.com/mdn/content/issues/16120

### DIFF
--- a/files/en-us/games/techniques/2d_collision_detection/index.md
+++ b/files/en-us/games/techniques/2d_collision_detection/index.md
@@ -49,15 +49,18 @@ const collider = {
   },
 };
 
+const container = document.getElementById("container");
+
 class BaseEntity {
   ref;
   position;
-  constructor(ref) {
-    this.ref = ref;
-    this.position = {
-      x: parseInt(ref.style.left, 10),
-      y: parseInt(ref.style.top, 10),
-    };
+  constructor(position) {
+    this.position = position;
+    this.ref = document.createElement("div");
+    this.ref.classList.add("entity");
+    this.ref.style.left = `${this.position.x}px`;
+    this.ref.style.top = `${this.position.y}px`;
+    container.appendChild(this.ref);
   }
   shiftPosition(dx, dy) {
     this.position.x += dx;
@@ -79,8 +82,6 @@ class BaseEntity {
     throw new Error("isCollidingWith must be implemented in subclasses");
   }
 }
-
-const container = document.getElementById("container");
 
 document.addEventListener("keydown", (e) => {
   e.preventDefault();
@@ -143,15 +144,18 @@ const collider = {
   },
 };
 
+const container = document.getElementById("container");
+
 class BaseEntity {
   ref;
   position;
-  constructor(ref) {
-    this.ref = ref;
-    this.position = {
-      x: parseInt(ref.style.left, 10),
-      y: parseInt(ref.style.top, 10),
-    };
+  constructor(position) {
+    this.position = position;
+    this.ref = document.createElement("div");
+    this.ref.classList.add("entity");
+    this.ref.style.left = `${this.position.x}px`;
+    this.ref.style.top = `${this.position.y}px`;
+    container.appendChild(this.ref);
   }
   shiftPosition(dx, dy) {
     this.position.x += dx;
@@ -173,8 +177,6 @@ class BaseEntity {
     throw new Error("isCollidingWith must be implemented in subclasses");
   }
 }
-
-const container = document.getElementById("container");
 
 document.addEventListener("keydown", (e) => {
   e.preventDefault();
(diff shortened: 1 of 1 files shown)


files/en-us/games/techniques/2d_collision_detection/index.md
```diff
@@ -8,129 +8,364 @@ page-type: guide
 
 Algorithms to detect collision in 2D games depend on the type of shapes that can collide (e.g., Rectangle to Rectangle, Rectangle to Circle, Circle to Circle). Generally you will have a simple generic shape that covers the entity known as a "hitbox" so even though collision may not be pixel perfect, it will look good enough and be performant across multiple entities. This article provides a review of the most common techniques used to provide collision detection in 2D games.
 
+## Engine code
```
bsmth, line 11: This is pretty good. I have two suggestions:

1. Mention that you can use the arrow keys to move the entity around in each example
2. We can hack around the code duplication using multiple live-sample tags

For example:

````md
## Engine code

The demos in this page don't rely on any external library, so we implement all the orchestration ourselves, which includes rendering, handling user input, and invoking behaviors of each entity. The code is shown below (it won't be repeated for each example):

```html live-sample___bounding-box live-sample___circle-collision
<div id="container"></div>
```

```css live-sample___bounding-box live-sample___circle-collision
.entity {
  display: inline-block;
  position: absolute;
  height: 20px;
  width: 20px;
  background-color: blue;
}

.movable {
  left: 50px;
  top: 50px;
  background-color: red;
}

.collision-state {
  background-color: green !important;
}
```

```js live-sample___bounding-box live-sample___circle-collision
const collider = {
  moveableEntity: null,
  staticEntities: [],
  checkCollision() {
    // Important: the isCollidingWith method is what we are implementing
    const isColliding = this.staticEntities.some((staticEntity) =>
      this.moveableEntity.isCollidingWith(staticEntity),
    );
    this.moveableEntity.setCollisionState(isColliding);
  },
};

const container = document.getElementById("container");

class BaseEntity {
  ref;
  position;
  constructor(position) {
    this.position = position;
    this.ref = document.createElement("div");
    this.ref.classList.add("entity");
    this.ref.style.left = `${this.position.x}px`;
    this.ref.style.top = `${this.position.y}px`;
    container.appendChild(this.ref);
  }
  shiftPosition(dx, dy) {
    this.position.x += dx;
    this.position.y += dy;
    this.redraw();
  }
  redraw() {
    this.ref.style.left = `${this.position.x}px`;
    this.ref.style.top = `${this.position.y}px`;
  }
  setCollisionState(isColliding) {
    if (isColliding && !this.ref.classList.contains("collision-state")) {
      this.ref.classList.add("collision-state");
    } else if (!isColliding) {
      this.ref.classList.remove("collision-state");
    }
  }
  isCollidingWith(other) {
    throw new Error("isCollidingWith must be implemented in subclasses");
  }
}

document.addEventListener("keydown", (e) => {
  e.preventDefault();
  switch (e.key) {
    case "ArrowLeft":
      collider.moveableEntity.shiftPosition(-5, 0);
      break;
    case "ArrowUp":
      collider.moveableEntity.shiftPosition(0, -5);
      break;
    case "ArrowRight":
      collider.moveableEntity.shiftPosition(5, 0);
      break;
    case "ArrowDown":
      collider.moveableEntity.shiftPosition(0, 5);
      break;
  }
  collider.checkCollision();
});
```

## Axis-aligned bounding box

One of the simpler forms of collision detection is between two rectangles that are axis aligned — meaning no rotation. The algorithm works by ensuring there is no gap between any of the 4 sides of the rectangles. Any gap means a collision does not exist.

```js live-sample___bounding-box
class BoxEntity extends BaseEntity {
  width = 20;
  height = 20;

  isCollidingWith(other) {
    return (
      this.position.x < other.position.x + other.width &&
      this.position.x + this.width > other.position.x &&
      this.position.y < other.position.y + other.height &&
      this.position.y + this.height > other.position.y
    );
  }
}

for (let i = 0; i < 100; i++) {
  collider.staticEntities.push(
    new BoxEntity({
      x: Math.floor(Math.random() * 500),
      y: Math.floor(Math.random() * 500),
    }),
  );
}

const moveableEntity = new BoxEntity({ x: 500, y: 500 });
moveableEntity.ref.classList.add("movable");
collider.moveableEntity = moveableEntity;
```

Use keyboard arrow keys to move the entity around in the box:

{{EmbedLiveSample("bounding-box", "", 550)}}

## Circle collision

Another simple shape for collision detection is between two circles. This algorithm works by taking the center points of the two circles and ensuring the distance between the center points are less than the two radii added together.

```css live-sample___circle-collision
.entity {
  border-radius: 50%;
}
```

> [!NOTE]
> The circles' `x` and `y` coordinates refer to their top left corners, so we need to add the radius to compare their centers.

```js live-sample___circle-collision
class CircleEntity extends BaseEntity {
  radius = 10;

  isCollidingWith(other) {
    const dx =
      this.position.x + this.radius - (other.position.x + other.radius);
    const dy =
      this.position.y + this.radius - (other.position.y + other.radius);
    const distance = Math.sqrt(dx * dx + dy * dy);
    return distance < this.radius + other.radius;
  }
}
```

```js hidden live-sample___circle-collision
for (let i = 0; i < 100; i++) {
  collider.staticEntities.push(
    new CircleEntity({
      x: Math.floor(Math.random() * 500),
      y: Math.floor(Math.random() * 500),
    }),
  );
}

const moveableEntity = new CircleEntity({ x: 500, y: 500 });
moveableEntity.ref.classList.add("movable");
collider.moveableEntity = moveableEntity;
```

Use keyboard arrow keys to move the entity around in the box:

{{EmbedLiveSample("circle-collision", "", 550)}}
````

bsmth (approved):
Good to go once you had a look at the comment. Whether you apply the `live-sample` suggestion or not is up to you, you may not like it.

files/en-us/games/techniques/2d_collision_detection/index.md
Josh-Cena: Nice! Didn't think of it

bsmth: >multiple live-sample tags

TBH, I'm not even sure it's intended to work like that, but it's good to know for cases like this

PR #39808, opened 2025-06-05
fix(html): Add the qualifier to global attribute page titles

dipikabh:
### Description

We have the qualifier "HTML attribute:" on attribute pages ([example](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/accept)), but it is missing from titles for global attributes. 
This PR adds "HTML global attribute:" to the `title` key and repurposes the existing title as `short-title`.

### Motivation

To ensure consistency in title format in the same technology area. This also benefits SEO.

### DIFF
--- a/files/en-us/web/html/reference/global_attributes/accesskey/index.md
+++ b/files/en-us/web/html/reference/global_attributes/accesskey/index.md
@@ -1,5 +1,5 @@
 ---
-title: "HTML global attribute: accesskey"
+title: accesskey global attribute
 short-title: accesskey
 slug: Web/HTML/Reference/Global_attributes/accesskey
 page-type: html-attribute
--- a/files/en-us/web/html/reference/global_attributes/accesskey/index.md
+++ b/files/en-us/web/html/reference/global_attributes/accesskey/index.md
@@ -1,5 +1,6 @@
 ---
-title: accesskey
+title: "HTML global attribute: accesskey"
+short-title: accesskey
 slug: Web/HTML/Reference/Global_attributes/accesskey
 page-type: html-attribute
 browser-compat: html.global_attributes.accesskey
--- a/files/en-us/web/html/reference/global_attributes/accesskey/index.md
+++ b/files/en-us/web/html/reference/global_attributes/accesskey/index.md
@@ -1,5 +1,5 @@
 ---
-title: accesskey global attribute
+title: HTML accesskey global attribute
 short-title: accesskey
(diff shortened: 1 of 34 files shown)


chrisdavidmills (approved):
LGTM! Nice change, @dipikabh, this will improve human comprehension as well as SEO.

chrisdavidmills:
> LGTM! Nice change, @dipikabh, this will improve human comprehension as well as SEO.

@dipikabh oops, I've just noticed that this is a draft. Let me know when you are ready for me to take another look.

dipikabh:
Sorry @chrisdavidmills - I hit Draft as fast as I could so you wouldn't have to review again.
You were too quick! :)

Essentially, I put the PR on hold to push out a different change. @bsmth pointed out (and I agree) that the updated front matter makes the meta `<title>` on the page to look like:

`<title>HTML global attribute: exportparts - HTML: HyperText Markup Language | MDN</title>`

That is what's happening for attribute pages, like for https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/accept, it is:

`<title>HTML attribute: accept - HTML: HyperText Markup Language | MDN</title>`

To avoid that "HTML...HTML" repetition, I've dropped the technology prefix. In this new commit, I've updated the titles to follow this pattern:

```diff
- title: HTML global attribute: exportparts
+ title: exportparts global attribute
```

This now renders as:

`<title>exportparts global attribute - HTML: HyperText Markup Language | MDN</title>`

This style is slightly different from our existing attribute and element titles, but we're going to watch out for any negative SEO before making changes on other pages.

This is ready for your review now, thank you!

chrisdavidmills (approved):
@dipikabh this is a bit of a shame, as it makes the actual page headings a bit less strong, but I can understand and appreciate the compromise you've made here.

I think it still makes the headings better.

bsmth (approved):
Nice 👍🏻 All looks good to me

dipikabh:
Thank you both, @chrisdavidmills and @bsmth! 

@bsmth and I talked a bit more about this and for now, going with this format:

```
title: HTML accesskey global attribute
```
I'll merge shortly.

This change should also help with quick spotting and disambiguation in our Quick search (instead of having users pick out the correct page, the one they want, by looking at the slugs):

| Before | After |
| --- | --- |
| <kbd></kbd> | <kbd></kbd> |
| | |
| <kbd></kbd> | <kbd></kbd> |
| | |
| <kbd></kbd> | <kbd></kbd> |

PR #39815, opened 2025-06-06
Initial content for Integrity-Policy

### LINKED ISSUE #676: Integrity Policy
### Proposal

https://github.com/w3c/webappsec-subresource-integrity/pull/133 introduces an `Integrity-Policy` header.

Its goal is to enable developers to ensure all the scripts in their page are loaded with [Subresource-Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity), and block any scripts without integrity. It also provide developers with appropriate reports in case of blocking (or in report-only mode).

### Browser support

Chromium 138 (stable at end of June 2025).

### Tasks

- [ ]

### Dependencies

- [ ]

### Additional information

_No response_

Yes

yoavweiss:
---

### Description

This documents the Integrity-Policy headers.

### Motivation

I recently added Integrity-Policy to the platform, and I'd like to document it so developers can use it.

### Additional details

https://w3c.github.io/webappsec-subresource-integrity/#integrity-policy-section
https://chromestatus.com/feature/5178394056327168
https://github.com/WebKit/standards-positions/issues/458
https://github.com/WebKit/WebKit/pull/45773

We need to file an equivalent BCD PR that would update the fact that this is shipped in Chromium 138. (It also landed in WebKit ToT, but not yet in an STP)

Fixes https://github.com/mdn/mdn/issues/676


files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,25 @@ Browsers handle SRI by doing the following:
 
 2. If the script or stylesheet doesn't match its associated `integrity` value, the browser must refuse to execute the script or apply the stylesheet, and must instead return a network error indicating that fetching of that script or stylesheet failed.
 
+## Integrity policy
+
+The `Integrity-Policy` and `Integrity-Policy-Report-Only` HTTP headers enable a document to enforce a policy regarding the integrity metadata requirements on all the subresources it loads of certain types of subresources. (currently scripts)
```
chrisdavidmills:
```suggestion
The `Integrity-Policy` and `Integrity-Policy-Report-Only` HTTP headers enable a document to enforce a policy regarding the integrity metadata requirements on loaded subresources of specified types (currently scripts only).
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,25 @@ Browsers handle SRI by doing the following:
...
+
+When specified, the browser would not allow requests without an `integrity` attribute to be made. (or simply report about them if the policy is specified using the `Integrity-Policy-Report-Only` header)
```
chrisdavidmills:
```suggestion
When an `Integrity-Policy` header is set, the browser will not allow requests without an `integrity` attribute to be made. The `Integrity-Policy-Report-Only` header reports which requests would not be allowed if an `Integrity-Policy` header were set.
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,25 @@ Browsers handle SRI by doing the following:
...
+
+The headers are defined as a [structured field dictionary](https://www.rfc-editor.org/rfc/rfc8941.html#name-dictionaries) with the following keys:
```
chrisdavidmills:
```suggestion
The header values are defined as [structured field dictionaries](https://www.rfc-editor.org/rfc/rfc8941.html#name-dictionaries) with the following keys:
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,25 @@ Browsers handle SRI by doing the following:
...
+
+- `blocked-destinations` - defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is "script".
```
chrisdavidmills:
```suggestion
- `blocked-destinations`
  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is "script".
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,25 @@ Browsers handle SRI by doing the following:
...
+- `sources` - defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of integrity sources. The default and only currently supported value is "inline".
```
chrisdavidmills:
```suggestion
- `sources`
  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of integrity sources. The default and only currently supported value is "inline".
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,25 @@ Browsers handle SRI by doing the following:
...
+- `endpoints` - defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [reporting endpoints](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#endpoint).
```
chrisdavidmills:
```suggestion
- `endpoints`
  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [reporting endpoints](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#endpoint).
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,25 @@ Browsers handle SRI by doing the following:
...
+
+In case a request is blocked by integrity policy, a [Reporting API](/en-US/docs/Web/API/Reporting_API) violation report is created with of the "integrity-violation" type. It contains:
```
chrisdavidmills:
```suggestion
In cases where a request is blocked by an integrity policy, a [Reporting API](/en-US/docs/Web/API/Reporting_API) violation report is created with a type of "integrity-violation". It contains:
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,25 @@ Browsers handle SRI by doing the following:
...
+
+- `documentURL` - the URL of the document.
```
chrisdavidmills: Again, reformat this as a definition list, like I did for the previous one.

chrisdavidmills (changes requested):
Thanks for the addition, @yoavweiss!

What is here looks like a good start, however you also need to add reference pages for the two new headers, and link to them from this document. The HTTP header reference pages don't necessarily need to be long; see https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept for an example.

files/en-us/web/security/subresource_integrity/index.md
yoavweiss: I guess I need to clarify this.

When an `Integrity-Policy` header is specified, requests without an integrity attribute are blocked. (Also, no-cors requests. I need to add that part)
If a reporting endpoint is specified in `Integrity-Policy`, we'd get reports on top of blocking.

The `Integrity-Policy-Report-Only` header does not block requests. It just reports requests that would have been blocked if the policy was enforced.

chrisdavidmills: Cool, makes sense - yes, definitely update the text to make sure this is stated.

yoavweiss: done!

files/en-us/web/security/subresource_integrity/index.md
yoavweiss: done!

yoavweiss:
Added header docs

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Syntax
+
+```http
+Integrity-Policy-Report-Only: blocked-destinations=(<destination>),sources=(<source>),endpoints=(<endpoint>)
+```
+
+where `<destination>` is a [request destination](/en-US/docs/Web/API/Request/destination) value, `<source>` is an integrity metadata source, and `<endpoint>` is a [reporting endpoint](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#endpoint).
+
+The header values are defined as [structured field dictionaries](https://www.rfc-editor.org/rfc/rfc8941.html#name-dictionaries) with the following keys:
```
yoavweiss: This is copied over from the overall Integrity Policy section. I'm not sure which of these syntax descriptions we want (or do we want both??). Let me know.

caugner:
@yoavweiss It looks like you ran `npm i` or similar (this repo still uses yarn v1). Can you remove the `package-lock.json`, and revert the `yarn.lock` changes?

yoavweiss:
> @yoavweiss It looks like you ran `npm i` or similar (this repo still uses yarn v1). Can you remove the `package-lock.json`, and revert the `yarn.lock` changes?

done

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,33 @@ Browsers handle SRI by doing the following:
 
 2. If the script or stylesheet doesn't match its associated `integrity` value, the browser must refuse to execute the script or apply the stylesheet, and must instead return a network error indicating that fetching of that script or stylesheet failed.
 
+## Integrity policy
+
+The `Integrity-Policy` and `Integrity-Policy-Report-Only` HTTP headers enable a document to enforce a policy regarding the integrity metadata requirements on loaded subresources of specified types (currently scripts only).
```
chrisdavidmills:
```suggestion
The {{httpheader("Integrity-Policy")}} and {{httpheader("Integrity-Policy-Report-Only")}} HTTP headers enable a document to enforce a policy regarding the integrity metadata requirements on loaded subresources of specified types (currently scripts only).
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,33 @@ Browsers handle SRI by doing the following:
...
+
+When an {{HTTP("Integrity-Policy")}} header is specified, the browser would block requests with [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode or without an `integrity` attribute from being made.
```
chrisdavidmills:
```suggestion
When an `Integrity-Policy` header is specified, the browser would block requests with [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode or without an `integrity` attribute from being made.
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,33 @@ Browsers handle SRI by doing the following:
...
+When an {{HTTP("Integrity-Policy-Report-Only")}} header is specified, such requests would be made, but reported on to the specified reporting endpoint.
```
chrisdavidmills:
```suggestion
When an `Integrity-Policy-Report-Only` header is specified, such requests would be made, but reported to the specified reporting endpoint.
```

chrisdavidmills: Removing these two macro calls as we now link to the header pages in the first sentence, and they were named wrongly, so causing errors to appear in the page (the wrong name might have been in one of my suggestions; that being the case, sorry!).

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,33 @@ Browsers handle SRI by doing the following:
...
+
+The header values are defined as [structured field dictionaries](https://www.rfc-editor.org/rfc/rfc8941.html#name-dictionaries) with the following keys:
+
+- `blocked-destinations`
+  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is "script".
```
chrisdavidmills:
```suggestion
  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `"script"`.
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,33 @@ Browsers handle SRI by doing the following:
...
+- `sources`
+  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of integrity sources. The default and only currently supported value is "inline".
```
chrisdavidmills:
```suggestion
  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of integrity sources. The default and only currently supported value is `"inline"`.
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +128,33 @@ Browsers handle SRI by doing the following:
...
+- `endpoints`
+  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [reporting endpoints](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#endpoint).
+
+In cases where a request is blocked by an integrity policy, a [Reporting API](/en-US/docs/Web/API/Reporting_API) violation report is created with a type of "integrity-violation". It contains:
```
chrisdavidmills:
```suggestion
In cases where a request is blocked by an integrity policy, a [Reporting API](/en-US/docs/Web/API/Reporting_API) violation report is created with a type of `"integrity-violation"`. It contains:
```

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,58 @@
...
+slug: Web/HTTP/Reference/Headers/Integrity-Policy-Report-Only
+page-type: http-header
+browser-compat: http.headers.IntegrityPolicyReportOnly
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Integrity-Policy-Report-Only`** response header allows website administrators to ensure that all resources the user agent loads (of a certain type) have [Subresource Integrity](/en-US/docs/Web/Security/Subresource_Integrity) guarantees.
+
+It does so by _reporting_ on requests (of certain [request destinations](/en-US/docs/Web/API/Request/destination)) without integrity metadata or in [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode that are being made.
```
chrisdavidmills:
```suggestion
It does so by _reporting_ on requests (of certain [request destinations](/en-US/docs/Web/API/Request/destination)) without integrity metadata or in [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode.
```

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,58 @@
...
+
+This helps make developers aware of content manipulation of fetched subresources, and helps them to eventually deploy {{HTTPHeader("Integrity-Policy")}} without undesired breakage.
```
chrisdavidmills:
```suggestion
This helps make developers aware of content manipulation of fetched subresources, and allows them to test [integrity policies](/en-US/docs/Web/Security/Subresource_Integrity#integrity_policy) and eventually deploy an {{HTTPHeader("Integrity-Policy")}} header without breakage.
```

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,58 @@
...
+  </tbody>
+</table>
+
+## Syntax
+
+```http
+Integrity-Policy-Report-Only: blocked-destinations=(<destination>),sources=(<source>),endpoints=(<endpoint>)
+```
+
+where `<destination>` is a [request destination](/en-US/docs/Web/API/Request/destination) value, `<source>` is an integrity metadata source, and `<endpoint>` is a [reporting endpoint](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#endpoint).
```
chrisdavidmills: I'm not convinced that we need this line, as the values are explained in the below bullets too. Remove?

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,58 @@
...
+
+The header values are defined as [structured field dictionaries](https://www.rfc-editor.org/rfc/rfc8941.html#name-dictionaries) with the following keys:
+
+- `blocked-destinations`
+  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is "script".
```
chrisdavidmills:
```suggestion
  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `"script"`.
```

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,58 @@
...
+- `sources`
+  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of integrity sources. The default and only currently supported value is "inline".
```
chrisdavidmills:
```suggestion
  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of integrity sources. The default and only currently supported value is `"inline"`.
```

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
chrisdavidmills: I think having it in multiple places is fine.

files/en-us/web/http/reference/headers/integrity-policy/index.md
```diff
@@ -0,0 +1,57 @@
...
+  </tbody>
+</table>
+
+## Syntax
+
+```http
+Integrity-Policy: blocked-destinations=(<destination>),sources=(<source>),endpoints=(<endpoint>)
+```
+
+where `<destination>` is a [request destination](/en-US/docs/Web/API/Request/destination) value, `<source>` is an integrity metadata source, and `<endpoint>` is a [reporting endpoint](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#endpoint).
```
chrisdavidmills: Again, I'm not convinced we need this line.

files/en-us/web/http/reference/headers/integrity-policy/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+The header values are defined as [structured field dictionaries](https://www.rfc-editor.org/rfc/rfc8941.html#name-dictionaries) with the following keys:
+
+- `blocked-destinations`
+  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is "script".
```
chrisdavidmills:
```suggestion
  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `"script"`.
```

files/en-us/web/http/reference/headers/integrity-policy/index.md
```diff
@@ -0,0 +1,57 @@
...
+- `sources`
+  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of integrity sources. The default and only currently supported value is "inline".
```
chrisdavidmills:
```suggestion
  - : Defines an [inner list](https://www.rfc-editor.org/rfc/rfc8941.html#name-inner-lists) of integrity sources. The default and only currently supported value is `"inline"`.
```

chrisdavidmills (changes requested):
@yoavweiss some more comments for you, but nothing major. We are very close now. Nice work on the header pages!

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
hamishwillee: This should either explain what an inner list is or link to a definition on MDN. The information in the spec at the end of this link is unhelpful.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,58 @@
...
+- `endpoints`
```
hamishwillee: Is this the same as `report-to` in other APIs?

hamishwillee (commented):
Does this use the [Reporting_API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API)?

If so, should link/use that doc too.

Whether or not it does what does the report look like? 

Is there a BCD entry on the way?

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: Is there MDN info on structured fields?

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: It is. It defines a reporting endpoint.

yoavweiss:
> Does this use the [Reporting_API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API)?
> 
> If so, should link/use that doc too.

It is, and it should. I'll add something.

> Whether or not it does what does the report look like?

I'll try to copy other examples for representing the report body (e.g. CSP?)

> Is there a BCD entry on the way?

I need to add one, yes.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
chrisdavidmills: It doesn't, no, but I'm not really sure where we'd add a general reference for it. A glossary entry would perhaps be a bit weird.

What is a good example of an HTTP header that uses the same syntax structure that we have already documented?

We could possibly improve things by not linking to the spec definitions and instead adding an "Examples" section to each reference page below the "Syntax" section, which shows what an example header looks like.

yoavweiss: Well, [Reporting-Endpoints](https://www.w3.org/TR/reporting-1/#reporting-endpoints) uses a dictionary, but not one with inner-lists.

[Signature-Input](https://datatracker.ietf.org/doc/html/rfc9421#name-the-signature-input-http-fi) is another, but I'm not sure this one is web-exposed quite-yet.

hamishwillee: > We could possibly improve things by not linking to the spec definitions and instead adding an "Examples" section to each reference page below the "Syntax" section, 

I quite like that idea. Why don't we start with doing a first example here.

hamishwillee:
> > Does this use the [Reporting_API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API)?
> > If so, should link/use that doc too.
> 
> It is, and it should. I'll add something.
> 
> > Whether or not it does what does the report look like?
> 
> I'll try to copy other examples for representing the report body (e.g. CSP?)

All sounds very good. As we're in discussion on the COEP/COOP headers we can probably follow similar patterns.

I don't want to tread on Chris toes while you two work out what you want here, so can you ping me when you think this is done and I can do a more thorough review.
It is worth me waiting because that COOP/COEP stuff will refresh me on how this kind of thing is supposed to work.

files/en-us/web/http/reference/headers/integrity-policy/index.md
yoavweiss: Removed

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: Removed

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
hamishwillee: Or we might include the example right here alongside the parameter.

chrisdavidmills: Yeah, let's do that. 

> An example header might look like this:
>
> ```
> Sample header here
> ```

chrisdavidmills: And then remove the spec links. At that point, I think I'm pretty happy with this.

yoavweiss: Done! PTAL

hamishwillee:
@yoavweiss I'm back Friday and will pick up my bit of the review.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
chrisdavidmills: OK, cool. Having an example available definitely helps.

A few more points:

- Having looked at this on the page, I think it would be good to have the example underneath an "##Examples" heading in each case, to separate it from the syntax.
- The spec links are still there, on all three pages. Didn't we agree to remove them?
- In the examples, you've not included a `sources` key. Wouldn't it be good to include that?
- This also raises the questions of whether one or more of the keys are optional. If that is the case, the optional ones need to be marked with an `{{optional_inline}}` macro on the `<dt>` of the description list. So for example:

```
`sources` {{optional_inline}}
   Defines a list of integrity sources. The default and only currently supported value is "inline".
```

yoavweiss: `sources` is currently optional, and its default value is the same as its only possible value: "inline". Let me know if you think we should add it to the example, but I'd expect developers to care about it only once we have more than a single source of integrity metadata

chrisdavidmills: I think this is OK, provided it is marked as optional as detailed above.

hamishwillee:
@yoavweiss FYI only at some point you added a yarn.lock and package, which is not a good idea. I've rebased to remove those so we can cleanly rebase update when needed.

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -118,6 +118,14 @@ You can use the following {{HTMLElement("script")}} element to tell a browser th
 > [!NOTE]
 > For more details on the purpose of the `crossorigin` attribute, see [CORS settings attributes](/en-US/docs/Web/HTML/Reference/Attributes/crossorigin).
 
+### Integrity enforcement with the `Integrity-Policy` header
+
+You can add the following header to your document in order to ensure that all the external scripts it loads are loaded with integrity (and aren't loaded with [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode)
```
hamishwillee: Usually a good idea to cross link on first use.

```suggestion
You can add the {{httpheader("Integrity-Policy")}} header to your document in order to ensure that all the external scripts it loads are loaded with integrity (and aren't loaded with [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode)
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -128,6 +136,34 @@ Browsers handle SRI by doing the following:
...
+
+The header values are defined as structured field dictionaries with the following keys:
+
+- `blocked-destinations`
+  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `"script"`.
+- `sources` {{optional_inline}}
+  - : Defines a list of integrity sources. The default and only currently supported value is `"inline"`.
+- `endpoints` {{optional_inline}}
+  - : Defines a list of [reporting endpoints](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#endpoint).
+- `blocked-destinations`
```
hamishwillee: Duplicate. Breaks rendering.
```suggestion
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -100,6 +100,43 @@ For subresource-integrity verification of a resource served from an origin other
...
+- `documentURL`
+  - : The URL of the document.
+- `blockedURL`
+  - : The URL of the blocked subresource.
+- `destination`
+  - : The request destination.
+- `reportOnly`
+  - : Whether the report is a result of a report-only policy. When false, that means that the subresource was blocked from loading.
+
 ## Examples
```
hamishwillee, line 148: I have pushed this Examples section to the end. Either you interleave examples in the explanation of you have it at the end. But generally you don't have an examples half way though a guide and then add additional content. I can explain why if you need me to.

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -100,6 +100,43 @@ For subresource-integrity verification of a resource served from an origin other
...
+
+The {{httpheader("Integrity-Policy")}} and {{httpheader("Integrity-Policy-Report-Only")}} HTTP headers enable a document to enforce a policy regarding the integrity metadata requirements on loaded subresources of specified types (currently scripts only).
+
+When an `Integrity-Policy` header is specified, the browser would block requests with [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode or without an `integrity` attribute from being made.
+When an `Integrity-Policy-Report-Only` header is specified, such requests would be made, but reported to the specified reporting endpoint.
+
+The header values are defined as structured field dictionaries with the following keys:
+
+- `blocked-destinations`
+  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `"script"`.
```
hamishwillee: Currently is not helpful here - as the content dates we don't know if it is true.
```suggestion
  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only allowed value is `"script"`.
```

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,62 @@
...
+title: Integrity-Policy-Report-Only header
+short-title: Integrity-Policy-Report-Only
+slug: Web/HTTP/Reference/Headers/Integrity-Policy-Report-Only
+page-type: http-header
+browser-compat: http.headers.IntegrityPolicyReportOnly
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Integrity-Policy-Report-Only`** response header allows website administrators to ensure that all resources the user agent loads (of a certain type) have [Subresource Integrity](/en-US/docs/Web/Security/Subresource_Integrity) guarantees.
```
hamishwillee: @yoavweiss Assuming this is consistent with the other headers that use the reporting API there are a couple of nuances that aren't quite captured by this and the other header doc.

1. It isn't that one header blocks and the other reports. It is that one header blocks and report, while the other just reports. 
2. The reporting header is normally used in a rollout phase before full enforcment
3. NOrmally reports are only sent in both cases if a reporting endpoint is set. The endpoint mapping also has to be set in a `Reporting-Endpoint` header, so that should be mentioned. If there is no endpoint or no mapping then this will fail.

hamishwillee: Further, I presume this policy applies to the particular document, the script it loads, and any included scripts?

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+The header values are defined as structured field dictionaries with the following keys:
+
+- `blocked-destinations`
+  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `"script"`.
+- `sources` {{optional_inline}}
+  - : Defines a list of integrity sources. The default and only currently supported value is `"inline"`.
+- `endpoints` {{optional_inline}}
+  - : Defines a list of [reporting endpoints](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#endpoint).
```
hamishwillee:
```suggestion
  - : Defines a list of [reporting endpoints](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#endpoint).
    Reports are sent to all valid endpoints.
```

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,62 @@
...
+
+An example header might look like:
+
+```http
+Integrity-Policy-Report-Only: blocked-destinations=(script), endpoints=(integrity-endpoint, general-endpoint)
```
hamishwillee: 2. This example has a both an `integrity-endpoint` and  `general-endpoint`. The implication here is that this we are specifying one endpoint for integrity and one for other stuff. That is not the case in my understanding.  The reporting-endpoint header might specify reporting endpoints for multiple things (including multiple endpoints for integrity, and some for CSP), but what is in this list is where reports for these particular violations are sent. So you would normally have only one, or one and a backup server. Am I making the problem here clear?

```suggestion
Integrity-Policy-Report-Only: blocked-destinations=(script), endpoints=(integrity-endpoint, some-other-integrity-endpoint)
```

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,62 @@
...
+```
```
hamishwillee: 1. Would be good to see an example with `sources` being specified, or note why you haven't done so.
2. HTTP examples are normally under their own section `## Examples` and ideally have a third level header as well that refers to the specific thing shown.
3. In HTTP we generally have an example showing the header "in context". I.e. showing a message with the header set, and other relevant headers. In this case that might be Reporting-Endpoint. They might also show a response - not sure, but I presume you might show the cors enabling headers.

This makes things explanatory and visual, and allows for easier future extension.

I don't want to write your example, but the framework of it might look like this (with appropriate linking)

````suggestion
## Examples

### Reporting when scripts lack integrity metadata

This example shows a request that reports when any {{htmlelement("script")}} (or `HTMLScriptElement`) in the resource does not specify a hash that matches the script downloaded from the server, or if the resource is requested `no-cors`.

Note that the `integrity-endpoint` used in `Integrity-Policy-Report-Only` is define in the `Reporting-Endpoints` header.

```http
ALL THE OTHER HEADERS AND GUFF that are very relevant
Integrity-Policy-Report-Only: blocked-destinations=(script), endpoints=(integrity-endpoint, backup-integrity-endpoint)
```
```

The response might look like this, and note these points of interest.

```http
IF NEEDED :-)
```

````

File: files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md [hunk 33]
```diff
@@ -0,0 +1,62 @@
...
+```http
+Integrity-Policy-Report-Only: blocked-destinations=(<destination>),sources=(<source>),endpoints=(<endpoint>)
+```
+
+The header values are defined as structured field dictionaries with the following keys:
+
+- `blocked-destinations`
+  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `"script"`.
+- `sources` {{optional_inline}}
+  - : Defines a list of integrity sources. The default and only currently supported value is `"inline"`.
```
hamishwillee: What does it mean to set or not-set `inline`?

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,62 @@
...
+## Syntax
+
+```http
+Integrity-Policy-Report-Only: blocked-destinations=(<destination>),sources=(<source>),endpoints=(<endpoint>)
+```
+
+The header values are defined as structured field dictionaries with the following keys:
+
+- `blocked-destinations`
+  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `"script"`.
```
hamishwillee: For all case, I _think_ we should remove the string marker. In HTTP all values are strings. However in some cases there are directive values where the string requires `"`. So if you put the `"` it is ambiguous as to whether a developer should do so.


```suggestion
  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `script`.
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -100,6 +100,43 @@ For subresource-integrity verification of a resource served from an origin other
...
+
+   For subresource-integrity verification of a resource served from an origin other than the document in which it's embedded, browsers additionally check the resource using [Cross-Origin Resource Sharing (CORS)](/en-US/docs/Web/HTTP/Guides/CORS), to ensure the origin serving the resource allows it to be shared with the requesting origin.
+
+2. If the script or stylesheet doesn't match its associated `integrity` value, the browser must refuse to execute the script or apply the stylesheet, and must instead return a network error indicating that fetching of that script or stylesheet failed.
+
+## Integrity policy
+
+The {{httpheader("Integrity-Policy")}} and {{httpheader("Integrity-Policy-Report-Only")}} HTTP headers enable a document to enforce a policy regarding the integrity metadata requirements on loaded subresources of specified types (currently scripts only).
+
+When an `Integrity-Policy` header is specified, the browser would block requests with [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode or without an `integrity` attribute from being made.
```
hamishwillee:
```suggestion
When an `Integrity-Policy` header is specified, the browser blocks requests with [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode or without an `integrity` attribute from being made, and will also report violations if a valid reporting endpoint is specified.
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -100,6 +100,43 @@ For subresource-integrity verification of a resource served from an origin other
...
+When an `Integrity-Policy-Report-Only` header is specified, such requests would be made, but reported to the specified reporting endpoint.
```
hamishwillee:
```suggestion
When an `Integrity-Policy-Report-Only` header is specified the browser allows requests that violate the policy, but will report violations to the reporting endpoint (if a valid reporting endpoint is specified).
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -118,15 +155,13 @@ You can use the following {{HTMLElement("script")}} element to tell a browser th
 > [!NOTE]
 > For more details on the purpose of the `crossorigin` attribute, see [CORS settings attributes](/en-US/docs/Web/HTML/Reference/Attributes/crossorigin).
 
-## How browsers handle Subresource Integrity
+### Integrity enforcement with the `Integrity-Policy` header
```
hamishwillee, line 166: Once the example is extended as commented above, cross link to it, or perhaps update this with similar. It is helpful to have the context IMO as it also shows the reporting endpoints.

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -100,6 +100,43 @@ For subresource-integrity verification of a resource served from an origin other
...
+In cases where a request is blocked by an integrity policy, a [Reporting API](/en-US/docs/Web/API/Reporting_API) violation report is created with a type of `"integrity-violation"`. It contains:
+
+- `documentURL`
+  - : The URL of the document.
+- `blockedURL`
+  - : The URL of the blocked subresource.
+- `destination`
+  - : The request destination.
+- `reportOnly`
+  - : Whether the report is a result of a report-only policy. When false, that means that the subresource was blocked from loading.
```
hamishwillee: This is strictly accurate and useful. The problem with having it here is that this is guide and the reporting interface should be reference and have a browser compatibility data to allow us to reflect changes.

This information should be documented in the Reporting API as a dictionary `IntegrityViolationReportBody`. The browser compatibility for that should be created as outlined here: https://github.com/mdn/browser-compat-data/issues/27021#issuecomment-2958619985 (you need to do it here because BCD does not accept dictionaries.

Then in this point, I think you would do something like this:

````suggestion
In cases where a request is blocked by an integrity policy, a [Reporting API](/en-US/docs/Web/API/Reporting_API) violation report is created with a type of `integrity-violation` and a body of type {{domxref("IntegrityViolationReportBody")}} that includes information such as the URL of the document and the blocked resource.

A typical resource might look like this

```json
{
Show a real example
}
```
````

File: files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md (hunk 33 above)
Comment (yoavweiss) [reply files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md]: Currently, there's no meaningful difference. The intent is to add more integrity metadata `sources` in the future, at which point `sources=(inline, manifest)` would be different than e.g. `sources=(manifest)`

Comment (hamishwillee) [reply files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md]: That's cool. Worth saying that, and whether there is a default assumed? -i.e. should you put this in now preemtively.

File: files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md (hunk 32 above)
Comment (yoavweiss) [reply files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md]: Done!

File: files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md (hunk 33 above)
Comment (yoavweiss) [reply files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md]: Added some words to that effect

File: files/en-us/web/security/subresource_integrity/index.md [hunk 39]
```diff
@@ -114,28 +114,27 @@ Browsers handle SRI by doing the following:
...
-When an `Integrity-Policy-Report-Only` header is specified, such requests would be made, but reported to the specified reporting endpoint.
+When an `Integrity-Policy` header is specified, the browser blocks requests with [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode or without an `integrity` attribute from being made, and will also report violations if a valid reporting endpoint is specified.
+When an `Integrity-Policy-Report-Only` header is specified the browser allows requests that violate the policy, but will report violations to the reporting endpoint (if a valid reporting endpoint is specified).
 
 The header values are defined as structured field dictionaries with the following keys:
 
 - `blocked-destinations`
-  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only currently supported value is `"script"`.
+  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only allowed value is `"script"`.
```
hamishwillee:
```suggestion
  - : Defines a list of [request destinations](/en-US/docs/Web/API/Request/destination) to be blocked. The only allowed value is `script`.
```

files/en-us/web/security/subresource_integrity/index.md
```diff
@@ -114,28 +114,27 @@ Browsers handle SRI by doing the following:
...
-  - : The URL of the blocked subresource.
-- `destination`
-  - : The request destination.
-- `reportOnly`
-  - : Whether the report is a result of a report-only policy. When false, that means that the subresource was blocked from loading.
+A typical resource might look like this
+
+```json
+{
+Show a real example
```
hamishwillee: Obviously you will extend this :-)

yoavweiss: yeah, `git push` failed.. :)

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: > * It isn't that one header blocks and the other reports. It is that one header blocks and report, while the other just reports.

This should be clarified already



> * The reporting header is normally used in a rollout phase before full enforcment
> * NOrmally reports are only sent in both cases if a reporting endpoint is set. The endpoint mapping also has to be set in a `Reporting-Endpoint` header, so that should be mentioned. If there is no endpoint or no mapping then this will fail.

I'll clarify those points.

> Further, I presume this policy applies to the particular document, the script it loads, and any included scripts?

The policy applies to a document and is being enforced on all the scripts it loads.

files/en-us/web/security/subresource_integrity/index.md
yoavweiss: makes sense!

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: done

hamishwillee:
@yoavweiss Thanks for this. Good changes. I made a number of updates which I will try explain inline in https://github.com/mdn/content/pull/39815/commits/fb00c7a7c23684e52c16e5b3d23e91052f544157

The main point is to better cross link to the reporting API - this isn't just for sending reports to the reporting-endpoint it can also be used for local reporting.

Are you still intending to add a dictionary in the Web API as well and cross link? That is where you would describe the format of the report that you have shown?

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,105 @@
...
+browser-compat: http.headers.IntegrityPolicyReportOnly
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Integrity-Policy-Report-Only`** response header allows website administrators to report on resources that the user agent loads that would violate [Subresource Integrity](/en-US/docs/Web/Security/Subresource_Integrity) guarantees if the integrity policy was enforced (using the {{HTTPHeader("Integrity-Policy")}} header).
+
+Reports may be generated for requests on specified [request destinations](/en-US/docs/Web/API/Request/destination) that omit integrity metadata, or that are made in [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode.
+For reports to be sent to a reporting endpoint, the `Integrity-Policy-Report-Only` header must specify a valid reporting endpoint name that matches an endpoint declared using the {{HTTPHeader("Reporting-Endpoints")}} header.
+Reports are generated using the [Reporting API](/en-US/docs/Web/API/Reporting_API), and may also be observed in the page for which the integrity policy is being enforced.
```
hamishwillee:
```suggestion
Reports are generated using the [Reporting API](/en-US/docs/Web/API/Reporting_API), and may also be observed in the page for which the integrity policy is being reported.
```

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,105 @@
...
+browser-compat: http.headers.IntegrityPolicyReportOnly
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Integrity-Policy-Report-Only`** response header allows website administrators to report on resources that the user agent loads that would violate [Subresource Integrity](/en-US/docs/Web/Security/Subresource_Integrity) guarantees if the integrity policy was enforced (using the {{HTTPHeader("Integrity-Policy")}} header).
+
+Reports may be generated for requests on specified [request destinations](/en-US/docs/Web/API/Request/destination) that omit integrity metadata, or that are made in [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode.
+For reports to be sent to a reporting endpoint, the `Integrity-Policy-Report-Only` header must specify a valid reporting endpoint name that matches an endpoint declared using the {{HTTPHeader("Reporting-Endpoints")}} header.
+Reports are generated using the [Reporting API](/en-US/docs/Web/API/Reporting_API), and may also be observed in the page for which the integrity policy is being reported.
```
hamishwillee: FYI this is for better cross linking to the reporting API and to make it clear that:
- the reporting endpoint is required to send reports that way,
- but you can also observe reports in the page itself.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,105 @@
...
+title: Integrity-Policy-Report-Only header
+short-title: Integrity-Policy-Report-Only
+slug: Web/HTTP/Reference/Headers/Integrity-Policy-Report-Only
+page-type: http-header
+browser-compat: http.headers.IntegrityPolicyReportOnly
+---
+
+{{HTTPSidebar}}
+
+The HTTP **`Integrity-Policy-Report-Only`** response header allows website administrators to report on resources that the user agent loads that would violate [Subresource Integrity](/en-US/docs/Web/Security/Subresource_Integrity) guarantees if the integrity policy was enforced (using the {{HTTPHeader("Integrity-Policy")}} header).
```
hamishwillee, line 11: FYI this and the last sentence more clearly indicate that this is a tool for rolling out a policy. First you roll this out and make sure that all instances of your metadata are correct. Then you start enforcing - at which point you can spot future changes that break your guarantees, or attempts to bypass the guarantees by modifying the source.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,105 @@
...
+
+The header values are defined as structured field dictionaries with the following keys:
+
+- `blocked-destinations`
+
+  - : A list of [request destinations](/en-US/docs/Web/API/Request/destination) that must include valid integrity metadata.
+    Allowed values are:
+
+    - `script`
+      - : Script resources.
```
hamishwillee: It's a little "to my taste" but even for single options having it split out makes future extension easier.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,105 @@
...
+
+  - : A list of integrity sources that must include integrity metadata.
+    Allowed values are:
+
+    - `inline`
+
+      - : Inline sources, such as scripts and styles, must include integrity metadata.
+        This is the default.
+
+        As this is the default and only value, omitting `sources` is equivalent to specifying `sources=(inline)`.
```
hamishwillee: @yoavweiss Can you please check this. 

Is it that this will violate on all inline things that have integrity metadata, such as scripts AND styles as I have said, or just the inline cases that match our blocked destinations? I.e. you might guess that this has to match the blocked sources but I don't think it does?

Further, what I'm trying to do is make it clear that right now you can omit this but if we later added a new source `fred` then doing `sources=(fred)` would mean that you are _NOT_ enforcing the policy on `inline` resources.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,105 @@
...
+This example shows a document that reports when any {{htmlelement("script")}} (or `HTMLScriptElement`) does not specify an `integrity` attribute, or when a script resource is requested in [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode.
+
+Note that the `integrity-endpoint` used in `Integrity-Policy-Report-Only` is defined in the {{httpheader("Reporting-Endpoints")}} header.
+
+```http
+Reporting-Endpoints: integrity-endpoint=https://example.com/integrity, backup-integrity-endpoint=https://report-provider.example/integrity
+Integrity-Policy-Report-Only: blocked-destinations=(script), endpoints=(integrity-endpoint, backup-integrity-endpoint)
+```
+
+The [report payload](/en-US/docs/Web/API/Reporting_API#reporting_server_endpoints) might look like this.
```
hamishwillee, line 73: I think this is the best link for payload.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -0,0 +1,105 @@
...
+  "url": "https://example.com",
+  "body": {
+    "documentURL": "https://example.com",
+    "blockedURL": "https://example.com/main.js",
+    "destination": "script",
+    "reportOnly": false
+  }
+}
+```
+
```
hamishwillee, line 87: IF/WHEN you update the reporting API we might link to the format of that payload below here.

files/en-us/web/http/reference/headers/integrity-policy/index.md
```diff
@@ -0,0 +1,105 @@
+---
```
hamishwillee, line 1: FYI the changes I made here "more or less" mirror the changes in the report-only doc. So the questions are the same too :-)

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: > Further, what I'm trying to do is make it clear that right now you can omit this but if we later added a new source `fred` then doing `sources=(fred)` would mean that you are _NOT_ enforcing the policy on `inline` resources.

Yeah, that is correct!



> Is it that this will violate on all inline things that have integrity metadata, such as scripts AND styles as I have said, or just the inline cases that match our blocked destinations?

Just the blocked destinations.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: Makes sense. We can also add a link to https://developer.mozilla.org/en-US/docs/Web/API/ReportingObserver

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: Makes sense!

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: Yeah, not sure there's a better alternative

files/en-us/web/http/reference/headers/integrity-policy/index.md
yoavweiss: I think I answered/changed both. Let me know if I missed anything

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -12,7 +12,7 @@ The HTTP **`Integrity-Policy-Report-Only`** response header allows website admin
 
 Reports may be generated for requests on specified [request destinations](/en-US/docs/Web/API/Request/destination) that omit integrity metadata, or that are made in [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode.
 For reports to be sent to a reporting endpoint, the `Integrity-Policy-Report-Only` header must specify a valid reporting endpoint name that matches an endpoint declared using the {{HTTPHeader("Reporting-Endpoints")}} header.
-Reports are generated using the [Reporting API](/en-US/docs/Web/API/Reporting_API), and may also be observed in the page for which the integrity policy is being reported.
+Reports are generated using the [Reporting API](/en-US/docs/Web/API/Reporting_API), and may also be observed in the page for which the integrity policy is being enforced, using a [ReportingObserver](/en-US/docs/Web/API/ReportingObserver).
```
hamishwillee: This is incorrect - we're not enforcing the policy in the `Integrity-Policy-Report-Only` (i.e. I changed that deliberately).
Some report types include a `disposition` field to let you know if the report is enforced or reporting.

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
```diff
@@ -52,7 +52,7 @@ The header values are defined as structured field dictionaries with the followin
 
     - `inline`
 
-      - : Inline sources, such as scripts and styles, must include integrity metadata.
+      - : The integrity metadata source is inline to the content, such as the [integrity attribute](/en-US/docs/Web/API/HTMLScriptElement/integrity).
```
hamishwillee, line 51: Thanks!

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
yoavweiss: Apologies, copy/paste error :/

files/en-us/web/http/reference/headers/integrity-policy/index.md
hamishwillee: @yoavweiss Except the most important one 

> Are you still intending to add a dictionary in the Web API as well and cross link? That is where you would describe the format of the report that you have shown?

files/en-us/web/http/reference/headers/integrity-policy-report-only/index.md
hamishwillee:
```suggestion
Reports are generated using the [Reporting API](/en-US/docs/Web/API/Reporting_API), and may also be observed in the page for which the integrity policy is being reported, using a [ReportingObserver](/en-US/docs/Web/API/ReportingObserver).
```

yoavweiss: Fixed

hamishwillee (changes requested):
This all looks good except for the missing `IntegrityViolationReportBody`, which should be a peer of https://developer.mozilla.org/en-US/docs/Web/API/CSPViolationReportBody

Let me know if you're planning to do this. If not, I could take a shot at it.

yoavweiss:
> This all looks good except for the missing `IntegrityViolationReportBody`, which should be a peer of https://developer.mozilla.org/en-US/docs/Web/API/CSPViolationReportBody
> 
> Let me know if you're planning to do this. If not, I could take a shot at it.

I'm actually not sure we want to document it as an interface, as there are (vague) plans to turn all these interfaces to dicts: https://github.com/w3c/reporting/pull/284

hamishwillee:
> > This all looks good except for the missing `IntegrityViolationReportBody`, which should be a peer of https://developer.mozilla.org/en-US/docs/Web/API/CSPViolationReportBody
> > Let me know if you're planning to do this. If not, I could take a shot at it.
> 
> I'm actually not sure we want to document it as an interface, as there are (vague) plans to turn all these interfaces to dicts: [w3c/reporting#284](https://github.com/w3c/reporting/pull/284)

The change to dictionaries happened and is currently causing some conflicts with the IDL. 

I'm happy for you to document it as either as an interface, or as a dictionary. The only difference is that if you do it as a dictionary you would omit `to_json` and not derive from ReportBody.
The reason both are OK is that the interface is a nothing convenience.

yoavweiss:
> I'm happy for you to document it as either as an interface, or as a dictionary.

If we're documenting it as a dictionary, won't it make the most sense to document it inline (either with the headers, or the broader feature), give that it won't have an explicit name?

hamishwillee:
> If we're documenting it as a dictionary, won't it make the most sense to document it inline (either with the headers, or the broader feature), give that it won't have an explicit name?

The MDN ideal is indeed to do it inline, but not if that would make the docs unusable, which this would.

The dictionaries will still have an explicit name in the spec. All the other ReportBody derived things that are visible in the ReportingObserver are keeping their dictionary name too,  and MDN will likely continue to use those. 

If it is documented in the header, then in theory you should duplicate that in both the enforce and report-to header for this particular header. Then we have to do that for all the other report bodies as well.

Then we have to do the same thing for all the objects to reflect their use in the API.
But where would we do that inline?
The "right" place is the callback to the observer constructor, but if we do this inline, we'd have to do it for every single body we have now and in future.

Upshot, is that this is a case where it makes sense to use dictionaries, because doing it inline would be a lot of duplication, and it would be very hard for users to work what do expect for the different kinds of reports.

Does that make sense?

The possible exception is  COEP and COOP since these don't have a name and aren't reported through the API (apparently). I'm still tempted to create a body for them but add a note they can ONLY be used through the reporting server endpoint.

yoavweiss:
> Let me know if you're planning to do this. If not, I could take a shot at it.

@hamishwillee maybe it's best if you give it a go, as I'm being distracted by the next shiny object..

hamishwillee:
> @hamishwillee maybe it's best if you give it a go, as I'm being distracted by the next shiny object..

I'll give it a go. Thank you for this. Looking good. Friday or Monday.

hamishwillee:
FYI only, I reduced this to only a few commits. The merge history made it hard for me to update and track what was going on. Some loss of "who did what" at commit level sorry.

The reviewdog comments seem spurious/wrong. I am checking.

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+
+- `blockedURL` {{ReadOnlyInline}}
+  - : A string representing the URL of the resource blocked by an enforced integrity policy (or just reported for a [reportOnly](#reportonly) policy).
+- `documentURL` {{ReadOnlyInline}}
+  - : A string representing the URL of the document that is attempting to load the resource.
+- `destination` {{ReadOnlyInline}}
+  - : A string indicating the [`Request.destination`](/en-US/docs/Web/API/Request/destination#script) of the resource that was blocked.
+    This can currently only be: `"script"`.
+- `reportOnly` {{ReadOnlyInline}}
+  - : A boolean: `false` if the policy was enforced, and `true` if the violation was only reported.
```
hamishwillee: Note, there didn't seem much else to say here, so decided not to create sub documents for each of the dictionary properties.

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
+---
```
hamishwillee, line 1: @chrisdavidmills and  @yoavweiss  I've finished the draft update - ready for you to look.

Chris, Yoav updated the headers to my satisfaction (and yours presumably). However the reporting also works via the reporting API so I have added this dictionary that defines the report. This follows the the same model as the other reporting API interfaces. I have cross linked this to the reporting API docs, and also into the header docs so that it can be the information is available from either direction.

The only real difference is that this is a dictionary, while the other report body pages refer to interfaces. The reason is that all the reports are changing to be dictionaries in the spec if they are used by the reporting API.

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+title: IntegrityViolationReportBody
+slug: Web/API/IntegrityViolationReportBody
+page-type: web-api-interface
+status:
+  - experimental
+browser-compat: http.headers.Integrity-Policy
```
hamishwillee, line 7: This may change to some other compat or an "as-well" compat. But right now there is no BCD for the body format, so using the HTTP header for this (and it will match). I'm in discussion with BCD team.

files/en-us/web/api/reporting_api/index.md
```diff
@@ -103,10 +110,15 @@ These HTTP response headers define the endpoints where reports are sent.
...
+Report endpoints can be set for the following reports using the {{CSP("report-to")}} directive on the corresponding headers:
 
 - CSP Violations
-  - : {{CSP("report-to")}} on {{HTTPHeader("Content-Security-Policy")}} or {{HTTPHeader("Content-Security-Policy-Report-Only")}}.
+  - : {{HTTPHeader("Content-Security-Policy")}} or {{HTTPHeader("Content-Security-Policy-Report-Only")}}.
+
+Report endpoints can be set for the following reports using the [`endpoints`](/en-US/docs/Web/HTTP/Reference/Headers/Integrity-Policy#endpoints) field in a structured dictionary on the corresponding headers:
+
+- Integrity-Policy Violations
+  - : {{httpheader("Integrity-Policy")}} or {{httpheader("Integrity-Policy-Report-Only")}}
```
chrisdavidmills:
```suggestion
  - : {{httpheader("Integrity-Policy")}} or {{httpheader("Integrity-Policy-Report-Only")}}.
```

files/en-us/web/api/reporting_api/index.md
```diff
@@ -103,10 +110,15 @@ These HTTP response headers define the endpoints where reports are sent.
...
-Report endpoints can be set for the following reports using the `report-to` directive on the corresponding headers:
+Report endpoints can be set for the following reports using the {{CSP("report-to")}} directive on the corresponding headers:
 
 - CSP Violations
-  - : {{CSP("report-to")}} on {{HTTPHeader("Content-Security-Policy")}} or {{HTTPHeader("Content-Security-Policy-Report-Only")}}.
+  - : {{HTTPHeader("Content-Security-Policy")}} or {{HTTPHeader("Content-Security-Policy-Report-Only")}}.
+
+Report endpoints can be set for the following reports using the [`endpoints`](/en-US/docs/Web/HTTP/Reference/Headers/Integrity-Policy#endpoints) field in a structured dictionary on the corresponding headers:
+
+- Integrity-Policy Violations
```
chrisdavidmills: Seems a bit weird that "Violations" is capitalized here, but I'm guessing you are just doing it to keep consistent with the existing content. Not a massive issue.

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Instance properties
+
+- `blockedURL` {{ReadOnlyInline}}
+  - : A string representing the URL of the resource blocked by an enforced integrity policy (or just reported for a [reportOnly](#reportonly) policy).
+- `documentURL` {{ReadOnlyInline}}
+  - : A string representing the URL of the document that is attempting to load the resource.
+- `destination` {{ReadOnlyInline}}
+  - : A string indicating the [`Request.destination`](/en-US/docs/Web/API/Request/destination#script) of the resource that was blocked.
+    This can currently only be: `"script"`.
```
chrisdavidmills:
```suggestion
    This can currently only be `"script"`.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+- `reportOnly` {{ReadOnlyInline}}
+  - : A boolean: `false` if the policy was enforced, and `true` if the violation was only reported.
+
+## Description
+
+Integrity Policy violations are reported when a document attempts to load a resource that does not meet the [Subresource Integrity](/en-US/docs/Web/Security/Subresource_Integrity) guarantees of a policy set using either {{httpheader("Integrity-Policy")}} or {{httpheader("Integrity-Policy-Report-Only")}} HTTP headers.
```
chrisdavidmills:
```suggestion
Integrity Policy violations are reported when a document attempts to load a resource that does not meet the [Subresource Integrity](/en-US/docs/Web/Security/Subresource_Integrity) guarantees of a policy set using either the {{httpheader("Integrity-Policy")}} or {{httpheader("Integrity-Policy-Report-Only")}} HTTP headers.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+
+Specifically, a report is sent when a document attempts to load a {{htmlelement("script")}} resource (or other [request destination](/en-US/docs/Web/API/Request/destination) listed in the policy) that does not have valid integrity metadata, or to make a request in [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode.
+
+Violation reports maybe obtained in a violating document using a {{domxref("ReportingObserver")}} callback (defined in the {{domxref("ReportingObserver/ReportingObserver","ReportingObserver()")}} constructor), filtering on report objects that have a `type` of `"integrity-violation"`.
```
chrisdavidmills:
```suggestion
Violation reports may be obtained in a violating document using a {{domxref("ReportingObserver")}} callback (defined in the {{domxref("ReportingObserver/ReportingObserver","ReportingObserver()")}} constructor), filtering on report objects that have a `type` of `"integrity-violation"`.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+
+Violation reports may also be sent as a JSON objects in a POST to the [`endpoints`](/en-US/docs/Web/HTTP/Reference/Headers/Integrity-Policy#endpoints) specified in the {{httpheader("Integrity-Policy")}} and {{httpheader("Integrity-Policy-Report-Only")}} headers.
```
chrisdavidmills:
```suggestion
Violation reports may also be sent as JSON objects in `POST` requests to the [`endpoints`](/en-US/docs/Web/HTTP/Reference/Headers/Integrity-Policy#endpoints) specified in the {{httpheader("Integrity-Policy")}} and {{httpheader("Integrity-Policy-Report-Only")}} headers.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+The JSON report objects are a serialization of the reports returned in the {{domxref("ReportingObserver")}}, and therefore also have a `type` of `"integrity-violation"`, and a `body` property that is a serialization of this interface.
```
chrisdavidmills:
```suggestion
The JSON report objects are a serialization of the reports returned in the {{domxref("ReportingObserver")}}, and therefore also have a `type` of `"integrity-violation"`, and a `body` property that is a serialization of this object.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+Note that endpoint values set in the policy must map to identifiers set using the {{HTTPHeader("Reporting-Endpoints")}} header.
+
+## Examples
+
+### Reporting using the API
+
+This example shows how you can obtain Integrity Policy violation reports using a {{domxref("ReportingObserver")}}.
+
+First we set a page's integrity policy using the {{httpheader("Integrity-Policy")}}.
+The policy below reports and blocks resource loading of any {{htmlelement("script")}} or {{domxref("HTMLScriptElement")}} that does not specify an `integrity` attribute, or when a script resource is requested in [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode.
```
chrisdavidmills:
```suggestion
The policy below reports and blocks resource loading of any {{htmlelement("script")}} element or {{domxref("HTMLScriptElement")}} object that does not specify an `integrity` attribute, or when a script resource is requested in [no-cors](/en-US/docs/Web/API/Request/mode#no-cors) mode.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+Note that for this example we're only interested in reporting the violations using the API, so we're omitting the reporting endpoints:
+
+```http
+Integrity-Policy: blocked-destinations=(script)
+```
+
+Next assume that our page includes the following element to load a script.
```
chrisdavidmills:
```suggestion
Next, we'll assume that our page includes the following element to load a script.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+Because we want to trigger a violation, it omits the `integrity` attribute used to check the script matches our expected version.
+We could also/alternatively omit the `cross-origin` attribute so the request is sent in `no-cors` mode.
```
chrisdavidmills:
```suggestion
We could also omit the `cross-origin` attribute so the request is sent in `no-cors` mode.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+> A script that complies with the policy might look like this:
+>
+> ```html
+> <script
+>   src="https://example.com/example-framework.js"
+>   integrity="sha384-oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC"
+>   crossorigin="anonymous"></script>
+> ```
+
+In order to get violations within the page, we construct a new {{domxref("ReportingObserver")}} object to listen for reports with the type `"integrity-violation"`, passing a callback that will receive and log the reports.
```
chrisdavidmills:
```suggestion
To observe violations within the page, we construct a new {{domxref("ReportingObserver")}} object to listen for reports with the type `"integrity-violation"`, passing a callback that will receive and log the reports.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+  {
+    types: ["integrity-violation"],
+    buffered: true,
+  },
+);
+
+observer.observe();
+```
+
+Above we log the each violation report object and a JSON-string version of the object, which might look similar to the object below.
```
chrisdavidmills:
```suggestion
Above, we log each violation report object and a JSON-string version of the object, which might look similar to the object below.
```

files/en-us/web/api/integrityviolationreportbody/index.md
```diff
@@ -0,0 +1,168 @@
...
+}
+```
+
+### Sending a report to a reporting endpoint
+
+Configuring a web page to send an Integrity Policy violation report to a [reporting server endpoint](/en-US/docs/Web/API/Reporting_API#reporting_server_endpoints) is very similar to the previous example.
+
+The main difference is that we need to specify one or more reporting endpoints where we want the reports to be sent, using the {{httpheader("Reporting-Endpoints")}} response header, and then reference these in the `endpoints` field when setting the policy.
+
+You can see this below, where we first define two endpoints `integrity-endpoint` and `backup-integrity-endpoint`, and then reference them in our policy:
```
chrisdavidmills:
```suggestion
You can see this below, where we first define two endpoints — `integrity-endpoint` and `backup-integrity-endpoint` — and then reference them in our policy:
```

chrisdavidmills (approved):
@hamishwillee I've looked over your new bits, and made some language suggestions, but nothing major. Good job.

files/en-us/web/api/reporting_api/index.md
hamishwillee: Cheers. I think I'll fix up both.
```suggestion
- Integrity-Policy violations
```

files/en-us/web/api/reporting_api/index.md
```diff
@@ -103,10 +110,15 @@ These HTTP response headers define the endpoints where reports are sent.
 - {{HTTPHeader("Report-To")}} {{deprecated_inline}}
   - : No longer part of the Reporting API but still supported by some browsers. This sets the name and URL of reporting endpoint groups, which may be used with a number of HTTP headers especially for [Network Error Logging](/en-US/docs/Web/HTTP/Guides/Network_Error_Logging) that has not yet been updated to support `Reporting-Endpoints`. Other Reporting API reports should use `Reporting-Endpoints` instead for better future support.
 
-Report endpoints can be set for the following reports using the `report-to` directive on the corresponding headers:
+Report endpoints can be set for the following reports using the {{CSP("report-to")}} directive on the corresponding headers:
 
 - CSP Violations
```
hamishwillee:
```suggestion
- CSP violations
```

hamishwillee (approved):
Thanks for the review @chrisdavidmills - looks good @yoavweiss - merging!

Issue #36615, opened 2024-11-01
Investigate removing all w3.org/TR URLs


Josh-Cena:
We nearly never quote W3C TRs, as they are snapshots but we are more interested in the latest drafts.

~~I've counted the following (using `grep -ho 'https://www.w3.org/TR/[^ #)>]*' **/*.md | sort | uniq -c | sort -nr | awk '{print "- [ ] " $2 " (" $1 ")"}'`)~~

[Please see comment below for how to find a list of links]

Josh-Cena:
(Updating) The following specs don't have editor draft versions (or already are):

- https://www.w3.org/TR/adapt-tools/
- https://www.w3.org/TR/arabic-math/
- https://www.w3.org/TR/ATAG20/
- https://www.w3.org/TR/css/
- https://www.w3.org/TR/emma/
- https://www.w3.org/TR/html-longdesc/
- https://www.w3.org/TR/jsgf/
- https://www.w3.org/TR/ldp/
- https://www.w3.org/TR/media-frags/
- https://www.w3.org/TR/rdf-primer/
- https://www.w3.org/TR/semantic-interpretation/
- https://www.w3.org/TR/SMIL/
- https://www.w3.org/TR/smil-animation/
- https://www.w3.org/TR/soap12-part1/
- https://www.w3.org/TR/speech-synthesis/
- https://www.w3.org/TR/UAAG20/
- https://www.w3.org/TR/UNDERSTANDING-WCAG20/
- https://www.w3.org/TR/WCAG20-TECHS/
- https://www.w3.org/TR/xml/
- https://www.w3.org/TR/xml-names/
- https://www.w3.org/TR/xml-stylesheet/
- https://www.w3.org/TR/2000/NOTE-SOAP-20000508/ (https://www.w3.org/TR/SOAP redirects to this as well)

The following links, plus all remaining yeared w3.org links, refer to historical references.

- https://www.w3.org/TR/CSS1/
- https://www.w3.org/TR/DOM-Level-1/
- https://www.w3.org/TR/DOM-Level-2-Core/
- https://www.w3.org/TR/DOM-Level-2-Style/
- https://www.w3.org/TR/DOM-Level-2-Traversal-Range/
- https://www.w3.org/TR/MathML3/

PR #39818, opened 2025-06-06
Replace w3.org/TR links: WCAG, ARIA, SVG

Josh-Cena:
This is a big part of https://github.com/mdn/content/issues/36615. I'm replacing all specs related to a11y/SVG.

In passing, I'm also replacing all WCAG spec links with WCAG 2.2. This sometimes leaves things inconsistent because we also have `https://www.w3.org/WAI/WCAG21/` links remaining, but 2.1 and 2.2 aren't really different.


pepelsbey (approved):
Looks good! Thank you :)

PR #39819, opened 2025-06-06
Internalize JSFiddle in Matrix math for the web

Josh-Cena:
Incorporated all examples about transform matrices, which also make the screenshots unnecessary. The bulk of the code is already on the page anyway; just some clever hacks I just learned about reusing code blocks between live samples are necessary.

Also removed another JSFiddle that demonstrates feature testing, because we already have BCD coverage for this and feature testing doesn't pertain to WebGL in particular.


files/en-us/web/api/webgl_api/matrix_math_for_the_web/index.md
```diff
@@ -18,7 +18,8 @@ Let's begin by considering the **identity matrix**. This is a special transforma
 
 The identity matrix looks like this in JavaScript:
 
-```js-nolint
+```js live-sample___translation_matrix_ex live-sample___scale_matrix_ex live-sample___rotation_matrix_ex live-sample___matrix_composition_ex
```
bsmth: I like how we can reuse some boilerplate like this. Really minor: I wonder if it's worth removing the tags for this block, then using a hidden block with the same contents as a template / starter for the rest of the page. Rationale is we get:

image

A "play" button with some content in there that hasn't been introduced yet.

bsmth (approved):
Examples look great, thanks a lot 👍🏻

files/en-us/web/api/webgl_api/matrix_math_for_the_web/index.md
Josh-Cena: Good idea, I removed the only dependency on the identity matrix instead, which should also be a bit more performant

Josh-Cena: Do you think we should do what you said for the other utilities or just this one?

bsmth: I think this one is the most important, the others are a bit more involved, so when you click play there, you see a bigger example. I'm fine with the current state of the branch, this is looking good

PR #39838, opened 2025-06-07
Remove feature detection code for HTML in XHR

Josh-Cena:
This feature is universally supported since 2013; there's no need to test it. This removes two JSFiddle links as well (https://github.com/mdn/content/issues/16120)


wbamberg (approved):
👍

PR #39841, opened 2025-06-07
Remove codepen in web audio user control

Josh-Cena:
This example doesn't really illustrate anything and doesn't pertain to web audio in particular. Also rewrote the text a bit to promote HTML rather than ARIA.

https://github.com/mdn/content/issues/16120


bsmth (approved):
Super, thanks 👍🏻

PR #39844, opened 2025-06-07
Internalize ReadableStream.cancel example

Josh-Cena:
This example is already on the page, so just need to make it a live sample

https://github.com/mdn/content/issues/16120


bsmth (approved):
+1. It's an awesome example

PR #39846, opened 2025-06-07
fixed broken DOMActivate example

addsoupbase:
Fixed broken, overly complicated example, as well as other stuff like 'see also'.

### Description

fixed example and correctly updated information for the `DOMActivate` event


wbamberg (commented):
In general this looks great, but the example doesn't work for me in Firefox (it works in Chrome and Safari).

Josh-Cena (approved):
Thanks! This is already a great improvement, and I'm merging it just because it's removing this example that I'm long having a headache for. I've changed the wording to make it less assertive about the behavior, because I believe the spec actually requires `detail` to be 0, and non-zero values are non-standard.

PR #39849, opened 2025-06-07
Fix error in "Modify an element's style" sample

elisee:
Hi,

[The code sample in "Modify an element's style"](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Using_dynamic_styling_information#modify_an_elements_style) was not working correctly (throwing an error) when clicking on the "Reset background color" button.

Changing `elem` to `p1` fixes it.


Josh-Cena (approved):
Good catch. Thanks

PR #39860, opened 2025-06-08
Fix grammar

zakingslayerv22:
I think "a" is missing. It could also be reworded as "...this results in a..."

---


estelle (approved):
Thanks! 🎉

PR #39863, opened 2025-06-09
FF140: MutationEvent removal

hamishwillee:
FF140 removes the [`MutationEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MutationEvent) and associated events such as `DOMAttrModified`, `DOMCharacterDataModified` ... in https://bugzilla.mozilla.org/show_bug.cgi?id=1963043

This adds a removal in the release notes and removes the entry from  the experimental features page (added when it was first put "on the path for removal".

Related docs work can be tracked in https://github.com/mdn/content/issues/39825


dipikabh (approved):
Thank you!

PR #39869, opened 2025-06-09
Remove active learning from HTML video and audio

chrisdavidmills:
### Description

As part of a larger project to make the challenges/reader exercises in MDN learn more consistent, this PR removes the old active learning text from the [HTML video and audio](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/HTML_video_and_audio) learning task.

It also adds a solution to the task, hidden inside a summary/detail structure.

### DIFF
--- a/files/en-us/learn_web_development/core/structuring_content/html_video_and_audio/index.md
+++ b/files/en-us/learn_web_development/core/structuring_content/html_video_and_audio/index.md
@@ -107,7 +107,7 @@ An audio player will tend to play an audio track directly, e.g., an MP3 or Ogg f
 > [!NOTE]
 > Several popular formats, such as MP3 and MP4/H.264, are excellent but are encumbered by patents; that is, there are patents covering some or all of the technology that they're based upon. In the United States, patents covered MP3 until 2017, and H.264 is encumbered by patents through at least 2027.
 >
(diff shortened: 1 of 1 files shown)


files/en-us/learn_web_development/core/structuring_content/html_video_and_audio/index.md
```diff
@@ -277,21 +277,48 @@ For more details, including on how to add labels please read [Adding captions an
...
+For this task, we'd (ideally) like you to go out into the world and record some of your own video and audio — most phones these days allow you to record audio and video very easily and, provided you can transfer it on to your computer, you can use it. You may have to do some conversion to end up with a WebM and MP4 in the case of video, and an MP3 and Ogg in the case of audio, but there are enough programs out there to allow you to do this without too much trouble, such as [Miro Video Converter](http://www.mirovideoconverter.com/) and [Audacity](https://sourceforge.net/projects/audacity/). We'd like you to have a go!
```
bsmth: Miro converter looks a little wobbly now (HTTP only). What about this one?

```suggestion
For this task, why not go out into the world and record some of your own video and audio? If you have a phone, use that to record audio and video, transfer it to your computer, and try it out. You may have to do some conversion to end up with a WebM and MP4 in the case of video, and an MP3 and Ogg in the case of audio, but there are enough programs and tools out there to allow you to do this without too much trouble, such as [CloudConvert](https://cloudconvert.com/mp4-converter) (online) and [Audacity](https://sourceforge.net/projects/audacity/) (desktop application). We'd like you to have a go!
```

bsmth (approved):
Another small comment for prose. If you've other web-based converters that are good, feel free to change the cloudconvert URL to something else.

After that, good to go!

PR #39876, opened 2025-06-09
fix outdated Safari compatibility note for `appearance: none`

mohammad-kenaan:
This PR updates the compatibility information for the CSS property `appearance: none`.

Previously, the MDN note incorrectly stated that Safari does not support `appearance: none`. However, this limitation only applied to older versions of Safari. Modern Safari versions (from Safari 16 onwards) support it properly.

This correction improves the accuracy of the documentation and ensures developers are not misled by outdated notes.

### Description

Clarified that Safari 16+ removed styling restrictions on , allowing appearance: none; for flexible styling.

### Motivation

The previous documentation was outdated regarding Safari's behavior. This update improves accuracy and helps developers style search inputs consistently.

### Additional details

[Safari 16 Release Notes](https://developer.apple.com/documentation/safari-release-notes/safari-16-release-notes)


chrisdavidmills (commented):
Hi @mohammad-kenaan, and thanks for your concern.

However, this isn't saying that Safari doesn't support `appearance: none`. This is saying that setting `appearance: none` is useful because it allows you to get around the restrictions on styling search inputs in Safari.

Also, the release notes you link to don't say that `appearance: none` is supported as of Safari 16. It says "Removed most non-standard CSS appearance values."

So, it sounds like there is a misunderstanding here, and that you are conflating something unrelated with the main point of the PR.

If we could find some solid evidence of search inputs being fully stylable as of a certain version of Safari, that would warrant a change.

mohammad-kenaan:
Thanks for the clarification, Chris!

You're absolutely right that the current note focuses on styling restrictions in Safari.

However, I’d like to clarify that the specific example mentioned — being unable to style `font-size` on `<input type="search">` — no longer applies in Safari 16 and newer. I tested this directly on Safari 18.5 on macOS and was able to set `font-size` without needing `appearance: none;`.

Would it make sense to update the note to specify that this restriction applies only to older versions of Safari?

This way, developers won’t be misled into thinking they still *need* `appearance: none;` for such basic styling.

chrisdavidmills:
> Would it make sense to update the note to specify that this restriction applies only to older versions of Safari?
> 
> This way, developers won’t be misled into thinking they still _need_ `appearance: none;` for such basic styling.

@mohammad-kenaan Right, I see what you mean now. I don't think we need such a large section to cover this issue anymore, now it has been fixed in Safari for a few versions. What do you think about cutting the content of this section down to something like this:

> The `appearance: none;` value used to be particularly useful for consistently styling [`<input type="search">`](/en-US/docs/Web/HTML/Reference/Elements/input/search) elements.  Without it, Safari didn't allow {{cssxref("height")}} or {{cssxref("font-size")}} values to be set on them. However, this is no longer the case in Safari 16 and later. You may still want to target `input[type="search"]` explicitly with `appearance: none;` if your browser support matrix includes Safari versions older than 16.
>
> In search inputs, the "x" delete button, which appears when the value is not null, disappears when the input loses focus in Edge and Chrome, but stays put in Safari. To remove via CSS, you can use this following rule:
> 
> ```css
> input[type="search"]:not(:focus, :active)::-webkit-search-cancel-button {
>   display: none;
> }
> ```

?

mohammad-kenaan:
What a great idea. I'll update the assignment right away.

files/en-us/learn_web_development/extensions/forms/advanced_form_styling/index.md
```diff
@@ -100,27 +100,16 @@ In most cases, the effect is to remove the stylized border, which makes CSS styl
 
 ### Taming search boxes
 
-[`<input type="search">`](/en-US/docs/Web/HTML/Reference/Elements/input/search) is basically just a text input, so why is `appearance: none;` useful here? The answer is that Safari search boxes have some styling restrictions — you can't adjust their `height` or `font-size` freely, for example.
+[`<input type="search">`](/en-US/docs/Web/HTML/Reference/Elements/input/search) is basically just a text input, so why is `appearance: none;` useful here? The `appearance: none;` value used to be particularly useful for consistently styling input type="search" elements. Without it, Safari didn't allow {{cssxref("height")}} or {{cssxref("font-size")}} values to be set on them. However, this is no longer the case in Safari 16 and later. You may still want to target `input[type="search"]` explicitly with `appearance: none;` if your browser support matrix includes Safari versions older than 16.
```
chrisdavidmills:
```suggestion
 The `appearance: none;` value used to be particularly useful for consistently styling [`<input type="search">`](/en-US/docs/Web/HTML/Reference/Elements/input/search) elements. Without it, Safari didn't allow {{cssxref("height")}} or {{cssxref("font-size")}} values to be set on them. However, this is no longer the case in Safari 16 and later. You may still want to target `input[type="search"]` explicitly with `appearance: none;` if your browser support matrix includes Safari versions older than 16.
```

chrisdavidmills: I don't think that first line fits well anymore.

files/en-us/learn_web_development/extensions/forms/advanced_form_styling/index.md
```diff
@@ -100,27 +100,16 @@ In most cases, the effect is to remove the stylized border, which makes CSS styl
 
 ### Taming search boxes
```
chrisdavidmills:
```suggestion
### Search boxes and `appearance`
```

chrisdavidmills: We're not really doing as much taming any more ;-)

chrisdavidmills (changes requested):
@mohammad-kenaan couple more bits for you; very close now!

chrisdavidmills (approved):
Awesome, thanks again @mohammad-kenaan! I think this is ready to merge now.

(I also ran the markdown through Grammarly and made some general improvements to the grammar. I hope you don't mind)

PR #39877, opened 2025-06-09
Update examples to use "use strict";

def00111:


Josh-Cena:
Hi :) MDN examples do not use `"use strict"`. They are intended as parts of a bigger program that will be copied anywhere, not necessarily at the top. You are responsible for any ambient setup like making them strict scripts, making them modules, adding shebangs, etc. I know we have a few outliers here and there, but this is the general principle.

PR #39878, opened 2025-06-09
Fix issue 39784: nonce and hash values may use characters from either…

### LINKED ISSUE #39784: Possibly wrong information about CSP-nonce
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#nonce-nonce_value

'nonce-<nonce_value>'

> This value consists of the string `nonce-` followed by a [base64-encoded](https://developer.mozilla.org/en-US/docs/Glossary/Base64) string.

This statement seems incorrect, and is immediately contradicted by the example below it, which uses a nonce that's not valid base64. I think the writer got confused with the CSP hashes, which MUST be base64 strings. AFAIK the nonce is not constrained to base64, as the example shows. There probably are constraints on the nonce, but I don't know them in detail. I can imagine that using e.g. emoji in a nonce might cause issues.

This value consists of the string `nonce-` followed by a string.

_No response_

_No response_

wbamberg:
Fixes https://github.com/mdn/content/issues/39784.


hamishwillee (approved):
Thanks. LGTM.

PR #39880, opened 2025-06-10
COEP and COOP reporting documentation

### LINKED ISSUE #39814: COOP and COEP have no documentation of the `report-to` parameter
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy

_No response_

I'd expect the COOP and COEP pages to include the [`report-to` header parameter](https://html.spec.whatwg.org/#coop-report-to), that enables them to define reporting endpoints.

I expected to find documentation on how a developer can define a reporting endpoint for COOP or COEP.

https://github.com/camillelamy/explainers/blob/main/coop_reporting.md
https://html.spec.whatwg.org/#coop-report-to
https://html.spec.whatwg.org/#coep-report-to

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/http/reference/headers/cross-origin-embedder-policy`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/reference/headers/cross-origin-embedder-policy/index.md

</details>

hamishwillee:
COOP and COEP use the reporting API to report violations. This adds docs.

- [ ] `COEPViolationReportBody` - report structure for COEP violations 
- [ ] `COOPViolationReportBody` - report structure for COOP violations 
- [ ] Update Reporting API docs everywhere to include links to these
- [ ] `Cross-Origin-Embedder-Policy-Report-Only` - add report-to
- [ ] `Cross-Origin-Embedder-Policy` - add report-to
- [ ] `Cross-Origin-Opener-Policy` - add report-to
- [ ] `Cross-Origin-Opener-Policy-Report-Only` - add report-to

Fixes #39814

### DIFF
--- a/files/en-us/web/api/coepviolationreportbody/blockedurl/index.md
+++ b/files/en-us/web/api/coepviolationreportbody/blockedurl/index.md
@@ -1,0 +1,42 @@
+---
+title: "COEPViolationReportBody: blockedURL property"
+short-title: blockedURL
+slug: Web/API/COEPViolationReportBody/blockedURL
+page-type: web-api-instance-property
+browser-compat: api.COEPViolationReportBody.blockedURL
+---
+
+{{APIRef("Reporting API")}}
+
+The **`blockedURL`** read-only property of the {{domxref("COEPViolationReportBody")}} interface returns the URL of the resource that triggered the violation.
+
+## Value
+
+An string containing a URL value.
+
+## Examples
+
+### Get the URL of the resource that triggered the report
+
+In this example we create a new {{domxref("ReportingObserver")}} to observe COEP violation reports, then log the value of `blockedURL` to the console.
+
+```js
+const options = {
+  types: ["coep"],
+  buffered: true,
+};
+
+const observer = new ReportingObserver((reports, observer) => {
+  const firstReport = reports[0];
+  console.log(firstReport.type); // coep
+  console.log(firstReport.body.blockedURL); // the URL of the violating resource
+}, options);
+```
+
+## Specifications
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
--- a/files/en-us/web/api/coepviolationreportbody/index.md
+++ b/files/en-us/web/api/coepviolationreportbody/index.md
@@ -1,0 +1,95 @@
+---
+title: COEPViolationReportBody
+slug: Web/API/COEPViolationReportBody
+page-type: web-api-interface
+browser-compat: api.COEPViolationReportBody
+---
+
+{{APIRef("Reporting API")}}
+
+The `COEPViolationReportBody` dictionary represents the {{domxref("Report.body","body")}} of an [Cross-Origin-Embedder-Policy (COEP)](/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy) policy violation report.
+
+> [!NOTE]
+> This object does not derive from {{domxref("ReportBody")}} (unlike other {{domxref("Report.body")}} values).
+
+## Instance properties
+
+- {{domxref("COEPViolationReportBody.type")}} {{ReadOnlyInline}}
+  - : A string representing whether the violation was caused by loading a child resource or a dedicated worker.
+- {{domxref("COEPViolationReportBody.blockedURL")}} {{ReadOnlyInline}}
+  - : A string containing the URL of the blocked resource.
+- {{domxref("COEPViolationReportBody.disposition")}} {{ReadOnlyInline}}
+  - : A string representing whether the violation blocked resource loading or was only reported on.
+
+<!--
+- {{domxref("COEPViolationReportBody.destination")}} {{ReadOnlyInline}}
+  - : A string indicating the parent navigable for which loading was was blocked: `"iframe"`.
+-->
+
+## Description
+
+COEP policy violations are reported using the [Reporting API](/en-US/docs/Web/API/Reporting_API) generic reporting framework.
+
+The reports are enabled by setting the `report-to` parameter with a valid reporting endpoint name on the {{httpheader("Cross-Origin-Embedder-Policy")}} and/or {{httpheader("Cross-Origin-Embedder-Policy-Report-Only")}} headers used to set the policy.
+Valid endpoint names are defined using the {{httpheader("Reporting-Endpoints")}} header.
+
+Violation reports are generated whenever a policy set by {{httpheader("Cross-Origin-Embedder-Policy")}} or {{httpheader("Cross-Origin-Embedder-Policy-Report-Only")}} blocks or would block the loading of a resource, respectively.
+
+A report is a {{domxref("Report")}} instance which has the {{domxref("Report.type","type")}} of `coep` and a {{domxref("Report.body","body")}} property that is an object of this type.
+Reports can be returned via the {{domxref("ReportingObserver")}} interface or serialized and sent in a `POST` to the reporting endpoint.
+
+There are two may causes of violations, which are indicated by the value of the {{domxref("Report.type","type")}} property.
+The first is a `navigation` violation, which is caused when an {{htmlelement("iframe")}} or other child browsing context attempts to load a new document or other resource that is not compatible with the embedder policy of its parent for cross origin isolation.
+The second is a `"worker initialization"` violation, which is when a page attempts to load a dedicated worker with an embedder policy that is not compatible with the page policy for cross origin isolation.
+
+In both cases "incompatible" means that the parent embedder policy requires cross origin isolation but the embedder policy of the child resource does not provide those that isolation.
+If the resource isn't loaded into a child context (such as a frame), or if the parent does not require isolation, then the embedder policy of the loaded resource doesn't matter.
+
+## Examples
+
+### COEP report
+
+In this example we create a new {{domxref("ReportingObserver")}} to observe coep reports, then log the first report to the console.
+
+```js
+const options = {
+  types: ["coep"],
+  buffered: true,
+};
+
+const observer = new ReportingObserver((reports, observer) => {
+  const firstReport = reports[0];
+  console.log(firstReport.type); // coep
+  console.log(firstReport);
+}, options);
+```
+
+The logged report object for a COEP violation from loading an iframe might look like this:
+
+```json
+[
+  {
+    "type": "coep",
+    "url": "https://url-where-report-sent",
+    "body": {
+      "type": "navigation",
+      "blockedURL": "https://url-of-frame-that-was-blocked",
+      "disposition": "enforce"
+    }
+  }
+]
+```
+
+## Specifications
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
+- {{httpheader("Reporting-Endpoints")}}
+- [Reporting API](/en-US/docs/Web/API/Reporting_API)
+- [The Reporting API](https://developer.chrome.com/docs/capabilities/web-apis/reporting-api)
--- a/files/en-us/web/api/coepviolationreportbody/type/index.md
+++ b/files/en-us/web/api/coepviolationreportbody/type/index.md
@@ -1,0 +1,47 @@
+---
+title: "COEPViolationReportBody: type property"
+short-title: type
+slug: Web/API/COEPViolationReportBody/type
+page-type: web-api-instance-property
+browser-compat: api.COEPViolationReportBody.type
+---
+
+{{APIRef("Reporting API")}}{{AvailableInWorkers}}
+
+The **`type`** read-only property of the {{domxref("COEPViolationReportBody")}} dictionary returns a value that indicates whether the violation was caused by attempting to load a child resource or a dedicated worker.
+
+## Value
+
+A string that can have one of the following values:
+
+- `"navigation"`
+  - : The violation was caused by an {{htmlelement("iframe")}} or other child browsing context attempting to load a new document.
+- `"worker initialization"`
+  - : The violation was caused by a page attempting to load a dedicated worker.
+
+## Examples
+
+### Get the type of a COEP violation report
+
+In this example we create a new {{domxref("ReportingObserver")}} to observe COEP violation reports, then log the value of `type` to the console.
+
+```js
+const options = {
+  types: ["coep"],
+  buffered: true,
+};
+
+const observer = new ReportingObserver((reports, observer) => {
+  const firstReport = reports[0];
+  console.log(firstReport.type); // coep
+  console.log(firstReport.body.type); // navigation or worker initialization
+}, options);
+```
+
+## Specifications
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
--- a/files/en-us/web/api/coepviolationreportbody/disposition/index.md
+++ b/files/en-us/web/api/coepviolationreportbody/disposition/index.md
@@ -1,0 +1,49 @@
+---
+title: "COEPViolationReportBody: disposition property"
+short-title: disposition
+slug: Web/API/COEPViolationReportBody/disposition
+page-type: web-api-instance-property
+browser-compat: api.COEPViolationReportBody.disposition
+---
+
+{{APIRef("Reporting API")}}
+
+The **`disposition`** read-only property of the {{domxref("COEPViolationReportBody")}} dictionary represents whether the report is for a violation that was blocked or only reported.
+
+## Value
+
+A string that can have one of the following values:
+
+- `"enforce"`
+  - : The violation caused loading of the embedded resource to be blocked.
+    This is set for violations of policies set with {{httpheader("Cross-Origin-Embedder-Policy")}}.
+- `"reporting"`
+  - : The violation was reported without blocking the resource from loading.
+    This is set for violations of policies set with {{httpheader("Cross-Origin-Embedder-Policy-Report-Only")}}.
+
+## Examples
+
+### Get the disposition of a report
+
+In this example we create a new {{domxref("ReportingObserver")}} to observe COEP violation reports, then log the value of `disposition` to the console.
+
+```js
+const options = {
+  types: ["coep"],
+  buffered: true,
+};
+
+const observer = new ReportingObserver((reports, observer) => {
+  const firstReport = reports[0];
+  console.log(firstReport.type); // coep
+  console.log(firstReport.body.disposition);
+}, options);
+```
+
+## Specifications
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
--- a/files/jsondata/GroupData.json
+++ b/files/jsondata/GroupData.json
@@ -1374,6 +1374,8 @@
       "overview": ["Reporting API"],
       "guides": [],
       "interfaces": [
+        "COEPViolationReportBody",
+        "COOPViolationReportBody",
         "CSPViolationReportBody",
         "DeprecationReportBody",
         "InterventionReportBody",


files/en-us/web/api/coepviolationreportbody/index.md
```diff
@@ -0,0 +1,95 @@
...
+  - : A string representing whether the violation was caused by loading a child resource or a dedicated worker.
+- {{domxref("COEPViolationReportBody.blockedURL")}} {{ReadOnlyInline}}
+  - : A string containing the URL of the blocked resource.
+- {{domxref("COEPViolationReportBody.disposition")}} {{ReadOnlyInline}}
+  - : A string representing whether the violation blocked resource loading or was only reported on.
+
+<!--
+- {{domxref("COEPViolationReportBody.destination")}} {{ReadOnlyInline}}
+  - : A string indicating the parent navigable for which loading was was blocked: `"iframe"`.
+-->
```
hamishwillee: When I run https://wpt.live/html/cross-origin-embedder-policy/reporting-navigation.https.html I get
-  additional property `destination` on the report. This isn't in the spec. 

@yoavweiss Do you know the story here? Normally we'd document the property here and in browser compatibility data. The question really is whether this is "coming in to the spec" or "didn't make it into the spec". Do we know the history?

files/en-us/web/api/coepviolationreportbody/index.md
```diff
@@ -0,0 +1,95 @@
...
+{{APIRef("Reporting API")}}
+
+The `COEPViolationReportBody` dictionary represents the {{domxref("Report.body","body")}} of an [Cross-Origin-Embedder-Policy (COEP)](/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy) policy violation report.
+
+> [!NOTE]
+> This object does not derive from {{domxref("ReportBody")}} (unlike other {{domxref("Report.body")}} values).
+
+## Instance properties
+
+- {{domxref("COEPViolationReportBody.type")}} {{ReadOnlyInline}}
```
hamishwillee: When I run https://wpt.live/html/cross-origin-embedder-policy/reporting-navigation.https.html I get another value for this than the values in the spec: `corp`. 

@yoavweiss Do you know the story here? Is this value deprecated, going/coming?
Are there other values we know about?

files/en-us/web/api/coepviolationreportbody/index.md
```diff
@@ -0,0 +1,95 @@
...
+page-type: web-api-interface
+browser-compat: api.COEPViolationReportBody
+---
+
+{{APIRef("Reporting API")}}
+
+The `COEPViolationReportBody` dictionary represents the {{domxref("Report.body","body")}} of an [Cross-Origin-Embedder-Policy (COEP)](/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy) policy violation report.
+
+> [!NOTE]
+> This object does not derive from {{domxref("ReportBody")}} (unlike other {{domxref("Report.body")}} values).
```
hamishwillee: This is necessary because until now we could say that a body was derived from `ReportBody`. But according to https://github.com/whatwg/html/issues/11365 even if this is true, that will likely be an abstract dictionary.
Further, this object is not named in the spec. I have chosen this name.

FYI if `Report` also ends up being an abstract base dictionary we might even end up defining this as a "report" rather than a "reportbody". Too early to say.

files/en-us/web/api/coepviolationreportbody/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+Violation reports are generated whenever a policy set by {{httpheader("Cross-Origin-Embedder-Policy")}} or {{httpheader("Cross-Origin-Embedder-Policy-Report-Only")}} blocks or would block the loading of a resource, respectively.
+
+A report is a {{domxref("Report")}} instance which has the {{domxref("Report.type","type")}} of `coep` and a {{domxref("Report.body","body")}} property that is an object of this type.
+Reports can be returned via the {{domxref("ReportingObserver")}} interface or serialized and sent in a `POST` to the reporting endpoint.
+
+There are two may causes of violations, which are indicated by the value of the {{domxref("Report.type","type")}} property.
+The first is a `navigation` violation, which is caused when an {{htmlelement("iframe")}} or other child browsing context attempts to load a new document or other resource that is not compatible with the embedder policy of its parent for cross origin isolation.
+The second is a `"worker initialization"` violation, which is when a page attempts to load a dedicated worker with an embedder policy that is not compatible with the page policy for cross origin isolation.
```
hamishwillee: This is how I interpret the spec in https://html.spec.whatwg.org/multipage/browsers.html#embedder-policy-checks

Essentially COEP allows you to specify that there must be a resource loading policy or CORS on the thing to be embedded, and whatever it nests. 
Ultimately the intent is to ensure that if you care about cross-origin isolation, you'll be able to set a policy that enforces it for emedded resources.

So a violation occurs if the thing that is loading requires cross-origin isolation and the thing to be loaded does not guarantee that. 

@yoavweiss Can you sanity check this?

files/en-us/web/api/coopviolationreportbody/columnnumber/index.md
```diff
@@ -0,0 +1,27 @@
...
+
+{{APIRef("Reporting API")}}
+
+The **`columnNumber`** read-only property of the {{domxref("COOPViolationReportBody")}} interface indicates the column number in the source file that triggered the [Cross-Origin-Opener-Policy (COEP)](/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-OPener-Policy) violation.
+
+This property is most useful alongside {{domxref("COOPViolationReportBody.sourceFile")}} and {{domxref("COOPViolationReportBody.lineNumber")}}, as it provides the location of the column in that file and line that resulted in a violation.
+
+## Value
+
+An integer containing the column number that triggered the violation.
```
hamishwillee: For line, column, source file - Need to 
- Add example
- Add info about what type (s) reports it can be returned in. I think they appear in "access-to-opener" reports (all of them)

files/en-us/web/api/coepviolationreportbody/index.md
yoavweiss: I don't, but @camillelamy might

files/en-us/web/api/coepviolationreportbody/index.md
yoavweiss: Again, @camillelamy may know

files/en-us/web/api/coepviolationreportbody/index.md
yoavweiss: Both Report and ReportBody are now specified as dictionaries: https://www.w3.org/TR/reporting-1/#interface-reporting-observer

But this is somewhat aspirational and isn't how they're implemented yet: https://issues.chromium.org/issues/424210695

files/en-us/web/api/coepviolationreportbody/index.md
yoavweiss: Apologies but I don't have a lot of context on this. Again @camillelamy is the authority here.

files/en-us/web/api/coepviolationreportbody/index.md
hamishwillee: Thanks. My current plan is to mark this as non-standard in the browser compatibility data. 

I will remove this from here unless @camillelamy is able to provide information - because there is no way for me to be canonical about it.
But since it is in compatibility data it is clear we haven't just forgotten about it.

files/en-us/web/api/coepviolationreportbody/index.md
hamishwillee: Cool. Again, it's non-standard. In this case we might just not mention it at all unless we can get some information.

files/en-us/web/api/coepviolationreportbody/index.md
hamishwillee: @yoavweiss Thanks. This is going to make things complicated.

TLDR; I'm probably going to continue documenting this as `COEPViolationReportBody` and  `COOPViolationReportBody` for now while we wait for the dust to settle. But will discuss in the writing team.

---

I'm trying to work out what compatibility data should do in https://github.com/mdn/browser-compat-data/pull/27047#issuecomment-2975088114 - it is messy because BCD doesn't document dictionaries, but on the other hand, they can only remove things 2 years after they are removed from the web platform. It is going to be confusing for all concerned.

For docs I'm not entirely certain what will happen. Generally Docs don't document dictionaries either, but if we do, we tend to document only the top level dictionaries.
Where I see us going in the longer term is having dictionaries such as `CSPViolationReport`, `COEPViolationReport`.
We could do this for the COEP/COOP reports because the spec doesn't indicate any derivation for them at all - it just refers to "an object".

files/en-us/web/api/coepviolationreportbody/index.md
hamishwillee: @camillelamy Can you please advise, or suggest someone else who might help. If not I guess I'll have to start adding issues to the spec.

files/en-us/web/api/coopviolationreportbody/type/index.md
```diff
@@ -0,0 +1,121 @@
...
+browser-compat: api.COOPViolationReportBody.type
+---
+
+{{APIRef("Reporting API")}}
+
+The **`type`** read-only property of the {{domxref("COOPViolationReportBody")}} dictionary returns a value that indicates whether the violation is reported by the opened or opener document, and whether it was triggered by a navigation or an attempt to access an opened document that is blocked by the policy.
+
+The fields present in the report depend on the type.
+
+## Value
```
hamishwillee: Hi @yoavweiss 

A violation is triggered in a navigation if the policy causes the opened doc to be opened in a new BCG.
The violation report might have a body type of `navigation-to-response` or `navigation-from-response`, which are from the perspective of the opened and opener document.

What I think happens here is that if a resource with a COOP header and a reporting endpoint is opened and this causes a BCG switch then it sends a `navigation-to-response`.
If that same resource is the opener then it sends a `navigation-from-response`.

YOu don't both reports right - unless both opener and opened have a BCG (but they would then send just their type of message).

files/en-us/web/api/coopviolationreportbody/disposition/index.md
```diff
@@ -0,0 +1,49 @@
+---
```
hamishwillee: @yoavweiss

Do you happen to know the how HTTP enforcement and report-only headers interwork?

I **think** that 
- you can have `Cross-Origin-Opener-Policy` and/or `Cross-Origin-Opener-Policy-Report-Only` headers on any document.
- The report(s) sent are from the perspective of the document with the policy and reporting endpoint.
- Violations are determined by comparing that documents policy (which may be enforced _or_ report only or both) and the **enforced policy of the _other document_** (you're always comparing the reporting/enforce policy in the current document to the actual policy of the other document).

So say I have an opener with a `Cross-Origin-Opener-Policy: same-site` that includes a reporting endpoint, and there is a navigation to another document that in this case has no policy (so a policy of `unsafe-none` and no reporting endpoint)
- The policy will be evaluated and there is a violation because the opener is opened in a new BCD.
- As per https://github.com/mdn/content/pull/39880/files#r2151347130 I think a report will be sent to the reporting endpoint of the opener of type `navigation-from-response`. 
- No report is sent for the opened.

Now consider the case where the opener still sets `Cross-Origin-Opener-Policy: same-site` but the opened sets `Cross-Origin-Opener-Policy-Report-Only: same-site-no-opener` and `Cross-Origin-Opener-Policy-Report-Only: unsafe-none` 
- There is a policy violation because the opened is opened in a new BCG.
-  I think a report will be sent to the reporting endpoint of the opener of type `navigation-from-response`. 
-  I think a two reports will be sent to the reporting endpoint of the "opened" document of type `navigation-from-response`. 
  - The first will be of disposition: enforced, and effectivePolicy `unsafe-none`
  - The second will be of disposition: reporting, and effectivePolicy `same-site-no-opener`
  
  
I think that is what the spec says, and it makes sense to me.

files/en-us/web/api/coepviolationreportbody/index.md
camillelamy: I think Chrome has another potential value which is worker, when the issue is coming from a worker initialization being blocked. That said, I think what we're doing with COEP on workers is non optimal, so I wanted to raise the issue and modify that. That would remove the need for this field. So I think it is fine to mark it as non standard.

files/en-us/web/api/coepviolationreportbody/index.md
camillelamy: Ok, so it seems in Chrome we have three values `corp`, `navigation` and `worker initialization`. The comment above about worker initialization applies. However, I think it makes sense to distinguish between corp and navigation, and I am surprised it is not in the spec. `corp` is used when a subresource is blocked. `navigation` when a navigation is blocked. And `worker initialization` when a web worker script is blocked from loading (but I think this one should go away and the script should never be blocked but inherit COEP from its creator).

files/en-us/web/api/coepviolationreportbody/index.md
camillelamy: So the issue here is that technically COEP applies even if the page does not have cross-origin isolation. I think it'd be better to be very precise about the checks happening:

- A navigation violation happens when an iframe embedded in a document with either COEP require-corp or COEP credentialless loads a document that:
  - Has neither COEP require-corp or COEP credentialless.
  - Or is cross-origin with the embedder (the one with COEP reporting), and the document does not have a CORP header that allows embedding in the parent.
-  A worker initialization violation happens when a dedicated worker created by a document with either COEP require-corp or COEP credentialless tries to load a worker script with neither COEP require-corp or COEP credentialless headers. Again, I plan to file a spec bug to change this behavior so the COEP is inherited from teh creator and this kind of violation never happens.

files/en-us/web/api/coepviolationreportbody/index.md
camillelamy: Ok I looked it up some more, and in Chrome destination is the type of resource being loaded, including iframes. But for subresources this can be different. I was not involved in the COEP spec nor in the implementation of COEP reporting in Chrome, so I am not sure of the intent of either here.

files/en-us/web/api/coepviolationreportbody/index.md
hamishwillee: @camillelamy Thank you very much - on quick scan I _think_ that is sufficient for me to continue.

I've been sidetracked but I will be coming back to this. 
Assuming I have further questions, would it be possible for you to you invite the person who wrote the chrome implementation to be involved so I'm not hassling you? If for nothing else, I'd like to ping them to see if the `corp` could be added to the spec. Otherwise we're more likely to end up with divergent implementations.

files/en-us/web/api/coopviolationreportbody/disposition/index.md
yoavweiss: I'll again have to call on @camillelamy for help here, as I'm not sure what the answer here should be.

files/en-us/web/api/coepviolationreportbody/index.md
camillelamy: Unfortunately, the person left the company. So I am pretty much the best person to answer your questions for Chrome, but my understanding is sometimes limited.

files/en-us/web/api/coopviolationreportbody/disposition/index.md
camillelamy: So I wrote the spec for reporting in COOP and am much more familiar with the implementation, so this one I can answer.

For COOP, we try to see if a browsing context group would happen were we to enforce all the policies we're asked about.

In practice, we separate enforce/report-only mode (which can have different reporting endpoints).

For enforcement, we only compare with the other document enforced policy: we are reporting that a BCG switch did happen after all.

For report-only, this is a bit more complex. First, we check if the other document report-only policy would ensure that there would be no BCG swap. If that's the case, we won't send a report. Our rational is that developers might want to deploy COOP on several documents in their origin in report-only mode first, and we should not warn them about issues that would not happen if they switch all of their report-only policies to enforcement.

If the report-only don't match, then we check what would happen if we enforced the report-only policy of the document we're navigating to compared with the enforced policy of the current document. We also compare the report-only policy of the current document with the enforced policy of the document we're navigating to. If any of those would trigger a BCG swap, then we note that the documents would ned up in two different browsing contexts.  We keep track of this information even if the documents navigate again, this is important for the access reporting. When the BCG switch would be triggered, we also send a report to the relevant endpoints of both the current document and the document we're navigating to. If there are no endpoints, then we don't send a response.

In your second example, you listed two report only policies so I am going to assume you meant unsafe-none as the enforced policy and same-origin-no-opener as the report-only one. In that case, you would get:

- You get a `navigation-from-response` report on the endpoint of the opener, with disposition enforced and policy same-origin.
- You get a `navigation-to-response` report on the enforced endpoint of the openee with disposition enforced and policy unsafe-none.
- You get a `navigation-to-response` report on the report-only endpoint of the openee with disposition report-only and policy same-origin-no-opener.

Now if the openee report-only policy was `same-origin`, and it was same-origin with the opener, you would not get the third report-only report. But you would still get the first two. And if both documents were same-origin and had report-only `same-origin` COOP, you would not get any report at all. If only one of them had the report-only COOP, then you would get reports.

This gets a bit more complex for access reporting, as we're keeping track of whether any report-only COOP would have triggered a BCG swap. Then when an access happen, we check if the two documents would be in different BCGs if COOP had been enforced. But the BCG swap might have been triggered during a previous navigation. To give you an example:

- You document has COOP report-only `same-origin`. It opens a popup with a cross-origin document. We record that all documents in the popup would be in a different BCG from the documents in the opener page if COOP was enforced. Accesses to the popup/from the popup are reported (provided they involve a same-origin document).
- The popup navigates to another origin. There is no new BCG switch due to COOP, but we would still report accesses because the pages would still be in different BCGs.
- The opener navigates to a non-COOP page. This would also trigger a BCG switch if COOP was enforced. We record this. However, we no longer have a reporting endpoint for COOP, so we don't send reports anymore.

hamishwillee:
FYI, I have not forgotten this. It's just not as high on the priority list as I would like. Won't be this month.

files/en-us/web/api/coepviolationreportbody/blockedurl/index.md
```diff
@@ -0,0 +1,42 @@
+---
```
hamishwillee: @camillelamy I'm hoping to get back to this next month.

When I last looked at this it [sounded like](https://github.com/mdn/browser-compat-data/issues/27021#issuecomment-2998926966) COEP and COOP violations would only be reported to server endpoints, and not be visible to reporting observers via the Web API. However looking in the spec for COEP it explicitly states that this will be visible:

> The "coep" report type is a [report type](https://w3c.github.io/reporting/#report-type) whose value is "coep". It is [visible to ReportingObservers](https://w3c.github.io/reporting/#visible-to-reportingobservers).

Can you confirm that it is visible and that this is also intended for COOP?

hamishwillee: @camillelamy Could you please confirm this? Is the report also visible via the API/to reporting observers?

dipikabh:
@hamishwillee removing myself for now, please add me back if and when this needs an edit review.

files/en-us/web/api/coepviolationreportbody/blockedurl/index.md
camillelamy: COEP reports are visible to reporting observers.

COOP property access reports are also visible to reporting observers. The navigation ones are not.

hamishwillee: Great. I have sidestepped this for now by splitting this particular doc into only COEP. Will do COOP later.

files/en-us/web/api/coepviolationreportbody/destination/index.md
```diff
@@ -0,0 +1,47 @@
...
+page-type: web-api-instance-property
+status:
+  - non-standard
+browser-compat: api.COEPViolationReportBody.destination
+spec-urls:
+  - https://html.spec.whatwg.org/multipage/browsers.html#embedder-policy-checks
+  - https://html.spec.whatwg.org/multipage/browsers.html#coep
+---
+
+{{APIRef("Reporting API")}} {{non-standard_header}}
```
hamishwillee: Note, non-standard because it isn't in the spec.

files/en-us/web/api/coepviolationreportbody/destination/index.md
```diff
@@ -0,0 +1,47 @@
...
+title: "COEPViolationReportBody: destination property"
+short-title: destination
+slug: Web/API/COEPViolationReportBody/destination
+page-type: web-api-instance-property
+status:
+  - non-standard
+browser-compat: api.COEPViolationReportBody.destination
+spec-urls:
+  - https://html.spec.whatwg.org/multipage/browsers.html#embedder-policy-checks
+  - https://html.spec.whatwg.org/multipage/browsers.html#coep
```
hamishwillee: Note, spec URLS here and in other docs until I can get BCD to provide something better.

files/en-us/web/api/coepviolationreportbody/type/index.md
```diff
@@ -0,0 +1,55 @@
...
+
+A string that can have one of the following values:
+
+- `"corp"`
+  - : The violation was triggered by a document with {{httpheader("Cross-Origin-Embedder-Policy")}} set to [`require-corp`](/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy#require-corp) attempting to load a cross-origin sub-resource that does not explicitly allow itself to be embedded (by setting an appropriate {{httpheader("Cross-Origin-Resource-Policy")}}).
+- `"navigation"`
+  - : The violation was triggered by a document attempting to load an {{htmlelement("iframe")}} (of any origin) that does not set a {{httpheader("Cross-Origin-Resource-Policy")}} that allows it to be loaded into the document's origin.
+    Note that a violation can occur even if the `<iframe>` source is same-origin with the document.
+- `"worker initialization"`
+  - : The violation was caused by a page attempting to load a dedicated worker that does not set a {{httpheader("Cross-Origin-Resource-Policy")}} that allows it to be loaded into the document's origin.
```
hamishwillee: @camillelamy I believe this to be true, but would appreciate it if you can confirm.

This comes from https://github.com/mdn/content/pull/39880#discussion_r2156572060

The reason I ask is that you were talking about making a spec update but I suspect it hasn't happened yet, and even if it has, might not have rolled out to implementations.

FWIW the reporting API indicates that the report should be a dictionary derived from the ReportBody dictionary. The spec would be easier to read if this was the case, and each of the properties had a normative explanation.

files/en-us/web/api/coepviolationreportbody/index.md
```diff
@@ -0,0 +1,188 @@
+---
+title: COEPViolationReportBody
```
hamishwillee: FYI According to the current reporting API this should be a dictionary derived from `ReportBody` dictionary. 
Previously `ReportBody` was an object with no properties and a `toJson` method, but that has been removed from the spec. However no implementations have yet changed.
That makes all the docs complicated, because we should be able to get rid of the `ReportBody` abstraction.

Howver the spec does not define _this_ object as a named dictionary - its just an object. I have therefore defined this as a dictionary with a useful name.

Note also that MDN doesn't document dictionaries generally, but it makes sense to do so and continue to do so for this case.

files/en-us/web/api/report/body/index.md
```diff
@@ -8,19 +8,18 @@ browser-compat: api.Report.body
...
-{{domxref("DeprecationReportBody")}}, {{domxref("InterventionReportBody")}}, or
-{{domxref("CSPViolationReportBody")}}.
-These all inherit from the base `ReportBody` class — study their reference
-pages for more information on what the particular report body types contain.
+An object.
+
+The properties on the object depend on the {{domxref("Report/type", "type")}} of report.
+A mapping between the `type` and report body object can be found in [Report types](/en-US/docs/Web/API/Reporting_API#report_types).
+
+Note that the specification defines this property as a {{domxref("ReportBody")}}-derived dictionary, where `ReportBody` has no properties or methods.
```
hamishwillee: FYI see my comment here https://github.com/mdn/content/pull/39880/changes#r2815443920

This isn't at all useful, but we can't get rid of it yet - unless reviewer would like me to do so :-)

files/en-us/web/api/report/type/index.md
```diff
@@ -8,14 +8,16 @@ browser-compat: api.Report.type
...
+The `type` determines the object used to populate the {{domxref("Report.body","body")}} property of the `report`.
 
 ## Value
 
-A string representing the type of the report. Currently the available types are
-`deprecation`, `intervention`, and `crash`.
+A string representing the type of the report.
+Allowed types include: `coep`, `csp-violation`, `deprecation`, `intervention`, `integrity-violation`.
+
+See [Report types](/en-US/docs/Web/API/Reporting_API#report_types) in _Reporting API_ for more information.
```
hamishwillee: FYI, it is going to be hard to be sure this is all in sync over time. I have therefore cross linked to report types, which is a useful list and mapping of report types to the body type.  We'll try keep them in sync, but if not at least there is a master list.

files/en-us/web/http/reference/headers/cross-origin-embedder-policy-report-only/index.md
```diff
@@ -0,0 +1,84 @@
...
+The HTTP **`Cross-Origin-Embedder-Policy-Report-Only`** (COEP) {{Glossary("response header")}} configures the current document's _report-only_ policy for loading and embedding cross-origin resources that are requested in `no-cors` mode.
+
+The header allows website administrators to report on resources that would be blocked by COEP, without preventing them from being loaded.
+This allows for a softer rollout of enforcement.
+
+> [!NOTE]
+> See {{HTTPHeader("Cross-Origin-Embedder-Policy")}} for more complete information and examples.
+>
+> The headers are the same except `Cross-Origin-Embedder-Policy` blocks resources from loading is needed for a document to be [cross-origin isolated](/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy#cross-origin_isolation).
+> In addition, [Violation reports](/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy#violation_reports) from `Cross-Origin-Embedder-Policy-Report-Only` have a {{domxref("COEPViolationReportBody/disposition", "disposition")}} of `"reporting"` instead of `enforce`.
```
hamishwillee: This header is almost the same as the `Cross-Origin-Embedder-Policy` . Rather than duplicate everything I have kept the syntax and directives, but removed the Description and examples.

This section notes the differences and points you to `Cross-Origin-Embedder-Policy`

files/en-us/web/http/reference/headers/cross-origin-embedder-policy/index.md
```diff
@@ -7,18 +7,9 @@ browser-compat: http.headers.Cross-Origin-Embedder-Policy
 sidebar: http
 ---
 
-The HTTP **`Cross-Origin-Embedder-Policy`** (COEP) {{Glossary("response header")}} configures the current document's policy for loading and embedding cross-origin resources.
+The HTTP **`Cross-Origin-Embedder-Policy`** (COEP) {{Glossary("response header")}} configures the current document's policy for loading and embedding cross-origin resources that are requested in `no-cors` mode.
```
hamishwillee, line 10: FYI This has had significant changes.

For the new features I have added a section on violation reporting and an example of just server-side reporting.  This links to the reporting API docs - there is some duplication, but I think worth it. 

More significantly I have slightly reframed the way this is presented.

This had somewhat mixed the disucssion of CEOP and CORS, which are the mechanisms for a resource to define whether it can be embedded for no-cors and cors requests, respectively.
These are both part of ensuring cross-origin isolation but they are aren't otherwise "linked"

The way it was though, it read like if you have a CORS error and you have set the COEP header, you'd get a COEP voilation. Actually you get a CORS violation, which isn't reported except via console.

Now tis is more clear.

hamishwillee:
@sideshowbarker @yoavweiss @camillelamy This documentation on COEP and COEP reporting is ready for review. 
I'd appreciate some sanity checking if you have time.

Sorry it took so long! I plan on looking at it again on Friday as doubtless there are some typos etc, but the broad story should be right. Note that this borrows from the structure of the other docs - the HTTP headers dicuss the reports, but the detail on how to get them and what the properties are is documented as part of the Reporting API in the relevant `body` object.

files/en-us/web/api/coepviolationreportbody/index.md
hamishwillee: Note, I am going to make these all top level report dictionaries as suggested by BCD (and that is in line with MDN behaviour). Work on going.

files/en-us/web/api/coepviolationreport/index.md
```diff
@@ -0,0 +1,209 @@
...
+
+```http
+Cross-Origin-Embedder-Policy: require-corp
+```
+
+This header enforces that all resources must be served with the {{HTTPHeader("Cross-Origin-Resource-Policy")}} header and a value of `cross-origin` in order to be loaded into the document's origin (`https://example.com`).
+Provided the server hosting `some-image.png` doesn't set the header we don't need to do anything else to trigger a COEP violation.
+
+To observe violations within the page, we construct a new {{domxref("ReportingObserver")}} object to listen for reports with the type `"coep"`, passing a callback that will receive and log the reports.
+This code needs to be loaded before the script that causes the violation, in the same page:
```
hamishwillee:
```suggestion
This code needs to be loaded before the script that causes the violation:
```

files/en-us/web/api/coepviolationreport/index.md
```diff
@@ -0,0 +1,209 @@
...
+  "type": "coep",
+  "url": "https://url-of-page-attempting-to-load-resource-in-violation",
+  "": {
+    "type": "corp",
+    "blockedURL": "https://url-of-blocked-resource",
+    "destination": "image",
+    "disposition": "enforce"
+  }
+}
+```
```
hamishwillee, line 104:
````suggestion
The structure of a typical report is shown below.
Note that we can see the URL of both the page that had its policy violated (`url`) and the resource that was blocked from loading (`body.blockedURL`).
We can also see that the report was triggered by a `corp` violation, and from the `body.disposition` that it was enforced (and not just reported).

```json
{
  "type": "coep",
  "url": "https://url-of-page-attempting-to-load-resource-in-violation",
  "body": {
    "type": "corp",
    "blockedURL": "https://url-of-blocked-resource",
    "destination": "image",
    "disposition": "enforce"
  }
}
```

COEP violation reports can also be sent as a JSON object in a `POST` to a [reporting server endpoint](/en-US/docs/Web/API/Reporting_API#reporting_server_endpoints).
The reporting server endpoint name is specified in the [`report-to`](/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy#report-to_endpoint_name) parameter when setting {{httpheader("Cross-Origin-Embedder-Policy")}} or {{httpheader("Cross-Origin-Embedder-Policy-Report-Only")}}.
Valid endpoint names and their mapping to a particular URL are defined using the {{httpheader("Reporting-Endpoints")}} header.

The structure of the server report is almost exactly the same as `COEPViolationReport`, except that it additionally includes `age` and `user_agent` fields.

```json
{
  "age": 412139,
  "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
  "type": "coep",
  "url": "https://url-of-page-attempting-to-load-resource-in-violation",
  "body": {
    "type": "corp",
    "blockedURL": "https://url-of-blocked-resource",
    "destination": "image",
    "disposition": "enforce"
  }
}
```
````

File: files/en-us/web/api/coepviolationreport/index.md [hunk 19]
```diff
@@ -0,0 +1,187 @@
...
+- `url`
+  - : A string representing the URL of the document that generated the report.
+
+## Description
+
+A document's policies for loading and embedding cross-origin resources that are requested in `no-cors` mode are configured and enforced using the {{httpheader("Cross-Origin-Embedder-Policy")}} HTTP header, and may also be reported but not enforced using the {{httpheader("Cross-Origin-Embedder-Policy-Report-Only")}} header.
+
+COEP policy violations may be reported whenever a policy set by those headers blocks (or would block) the loading of a resource.
+
+You can monitor for COEP violation reports within the page in which they are triggered using the [Reporting API](/en-US/docs/Web/API/Reporting_API).
```
hamishwillee:
```suggestion
You can monitor for COEP violation reports within the page that sets the policy using the [Reporting API](/en-US/docs/Web/API/Reporting_API).
```

files/en-us/web/http/reference/headers/cross-origin-embedder-policy/index.md
```diff
@@ -55,8 +52,123 @@ Cross-Origin-Embedder-Policy: unsafe-none | require-corp | credentialless
...
+      type: "navigation",
+      blockedURL: "https://url-of-resource-that-was-blocked",
+      destination: "image",
+      disposition: "enforce",
+    },
+  },
+];
+```
+
+For more information and examples see {{domxref("COEPViolationReport")}} and [Reporting API](/en-US/docs/Web/API/Reporting_API).
```
hamishwillee:
````suggestion
### Violation reports

Violations of a COEP may be sent as JSON objects in `POST` requests to a remote server, or observed in the violating page as using the [Reporting API](/en-US/docs/Web/API/Reporting_API).

To observe violations using the [Reporting API](/en-US/docs/Web/API/Reporting_API) you create a {{domxref("ReportingObserver")}} to observe {{domxref("COEPViolationReport")}} instances that have the `type` of `"coep"`.
The format of the `COEPViolationReport` report is shown below:

```js
[
  {
    type: "coep",
    url: "https://url-of-document-that-generated-report",
    body: {
      type: "navigation",
      blockedURL: "https://url-of-resource-that-was-blocked",
      destination: "image",
      disposition: "enforce",
    },
  },
];
```

To send violation reports to a remote server, you specify an endpoint name in the COEP header's [`report-to`](#report-to_endpoint_name) parameter, and provide a mapping from this name to a server address in the {{HTTPHeader("Reporting-Endpoints")}} header.
The format of the JSON report object a serialized version of `COEPViolationReport` that additionally includes `age` and `user_agent` properties:

```json
[
  {
    "age": 967132,
    "body": {
      "blockedURL": "https://url-of-resource-that-was-blocked",
      "destination": "image",
      "disposition": "enforce",
      "type": "corp"
    },
    "type": "coep",
    "url": "https://url-of-document-that-generated-report",
    "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36"
  }
]
```

For more information and examples see {{domxref("COEPViolationReport")}} and [Reporting API](/en-US/docs/Web/API/Reporting_API).
````

**Comment by hamishwillee** [comment]
@sideshowbarker FYI I've started the work turning the other interfaces into dictionaries in #43195.

There have been corresponding changes here for consistency. Specifically, most of the reporting is covered in the report object such as `COEPViolationReport` - there is boilerplate on the headers to reference this and only simple examples. This is to reduce duplication but still make reporting obvious from the headers.

File: files/en-us/web/http/reference/headers/cross-origin-embedder-policy/index.md [hunk 21]
```diff
@@ -32,21 +23,27 @@ The policy applies to loaded resources, and resources in {{htmlelement("iframe")
...
+  - : A document can only load resources requested in `no-cors` mode from the same origin, or resources that have explicitly set the {{HTTPHeader("Cross-Origin-Resource-Policy")}} header to a value that allows it to be embedded.
 
     Cross-origin resource loading will be blocked by COEP unless:
     - The resource is requested in `no-cors` mode and the response includes a {{HTTPHeader("Cross-Origin-Resource-Policy")}} header that allows it to be loaded into the document origin.
-    - The resource is requested in `cors` mode and the resource supports and is permitted by CORS.
-      This can be done, for example, in HTML using the [`crossorigin`](/en-US/docs/Web/HTML/Reference/Attributes/crossorigin) attribute, or in JavaScript by making a request with [`{mode="cors"}`](/en-US/docs/Web/API/RequestInit#cors).
+    - The resource is requested in `cors` mode, for example, in HTML using the [`crossorigin`](/en-US/docs/Web/HTML/Reference/Attributes/crossorigin) attribute, or in JavaScript by making a request with [`{mode="cors"}`](/en-US/docs/Web/API/RequestInit#cors).
```
sideshowbarker:
```suggestion
    - The resource is requested in `cors` mode; for example, in HTML using the [`crossorigin`](/en-US/docs/Web/HTML/Reference/Attributes/crossorigin) attribute, or in JavaScript by making a request with [`{mode="cors"}`](/en-US/docs/Web/API/RequestInit#cors).
```

files/en-us/web/http/reference/headers/cross-origin-embedder-policy/index.md
```diff
@@ -55,8 +52,83 @@ Cross-Origin-Embedder-Policy: unsafe-none | require-corp | credentialless
...
+If neither of these policies are set, then by default, resources can be loaded or embedded into a document as though they had a CORP value of `cross-origin` (meaning that they _can_ be loaded cross origin).
+
+The **`Cross-Origin-Embedder-Policy`** allows you to require that CORP headers be set, in responses to `no-cors` requests, in order to load cross-site resources into the current document.
+You can also set the policy to keep the default behavior, or to allow the resources to be loaded, but strip any credentials that might otherwise be sent.
+The policy applies to loaded resources, and resources in {{htmlelement("iframe")}}s and nested frames.
+
+> [!NOTE]
+> The `Cross-Origin-Embedder-Policy` doesn't override or affect the embedding behavior for a resource for which CORP or CORS has been set.
+> If CORP restricts a resource to being embedded only `same-origin`, it won't be loaded cross-origin into a resource irrespective of the COEP value.
```
sideshowbarker:
```suggestion
> If CORP restricts a resource to being embedded only `same-origin`, it won't be loaded cross-origin into a resource — irrespective of the COEP value.
```

files/en-us/web/http/reference/headers/cross-origin-embedder-policy/index.md
```diff
@@ -55,8 +52,83 @@ Cross-Origin-Embedder-Policy: unsafe-none | require-corp | credentialless
...
+
+### Cross-origin isolation
+
+COEP and CORS together ensure that the browser process only contains resources that have consented to be shared or contain no private data.
```
sideshowbarker:
```suggestion
COEP and CORS together ensure that the browser process only contains resources that have consented to be shared, or that contain no private data.
```

files/en-us/web/http/reference/headers/cross-origin-embedder-policy/index.md
```diff
@@ -55,8 +52,83 @@ Cross-Origin-Embedder-Policy: unsafe-none | require-corp | credentialless
...
+The response header for the document sets the `Cross-Origin-Embedder-Policy` and {{httpheader("Reporting-Endpoints")}} headers as shown below.
+Since the `require-corp` directive is set, all cross-origin resources requested in `no-cors` mode must be served with the CORP header.
+The `report-to` parameter specifies the name `"coep-endpoint"` as the name of the endpoint where reports should be sent, and `Reporting-Endpoints` specifies how that name maps to a particular URL.
+
+```http
+Reporting-Endpoints: coep-endpoint="https://some-example.com/coep"
+Cross-Origin-Embedder-Policy: require-corp; report-to="coep-endpoint"
+```
+
+In order for the `some-image.png` to be loaded without triggering a violation it would need to set {{HTTPHeader("Cross-Origin-Resource-Policy")}} to `cross-origin`.
```
sideshowbarker:
```suggestion
In order for the `some-image.png` to be loaded without triggering a violation, it would need to set {{HTTPHeader("Cross-Origin-Resource-Policy")}} to `cross-origin`.
```

files/en-us/web/http/reference/headers/cross-origin-embedder-policy/index.md
```diff
@@ -55,8 +52,83 @@ Cross-Origin-Embedder-Policy: unsafe-none | require-corp | credentialless
...
+If we omit the header or don't include it as `cross-origin` a violation will occur.
```
sideshowbarker:
```suggestion
If we omit the header or don't include it as `cross-origin`, a violation will occur.
```

files/en-us/web/http/reference/headers/cross-origin-embedder-policy/index.md
```diff
@@ -55,8 +52,83 @@ Cross-Origin-Embedder-Policy: unsafe-none | require-corp | credentialless
...
+      "type": "corp"
+    },
+    "type": "coep",
+    "url": "https://example.com",
+    "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36"
+  }
+]
+```
+
+The `type` of the report is `coep`, the `url` is the document in which the violation occurred.
```
sideshowbarker:
```suggestion
The `type` of the report is `coep`, and the `url` is the document in which the violation occurred.
```

files/en-us/web/http/reference/headers/cross-origin-embedder-policy-report-only/index.md
```diff
@@ -0,0 +1,85 @@
...
+sidebar: http
+---
+
+The HTTP **`Cross-Origin-Embedder-Policy-Report-Only`** (COEP) {{Glossary("response header")}} configures the current document's _report-only_ policy for loading and embedding cross-origin resources that are requested in `no-cors` mode.
+
+The header allows website administrators to report on resources that would be blocked by {{HTTPHeader("Cross-Origin-Embedder-Policy")}}, without preventing them from being loaded.
+This allows for a softer rollout of enforcement.
+
+Violations of the policy may be reported using the [Reporting API](/en-US/docs/Web/API/Reporting_API).
+Reports can be observed in the page for which the policy is being set using a [`ReportingObserver`](/en-US/docs/Web/API/ReportingObserver), and sent to server endpoints defined in a {{HTTPHeader("Reporting-Endpoints")}} HTTP response header and selected using the [`report-to`](#report-to_endpoint_name) parameter.
```
sideshowbarker:
```suggestion
Reports can be observed in the page for which the policy is being set using a [`ReportingObserver`](/en-US/docs/Web/API/ReportingObserver), and sent to server endpoints defined in a {{HTTPHeader("Reporting-Endpoints")}} HTTP response header, and selected using the [`report-to`](#report-to_endpoint_name) parameter.
```

files/en-us/web/http/reference/headers/cross-origin-embedder-policy-report-only/index.md
```diff
@@ -0,0 +1,85 @@
...
+- `unsafe-none`
+  - : Allows the document to load cross-origin resources requested in `no-cors` mode **without** giving explicit permission through the {{HTTPHeader("Cross-Origin-Resource-Policy")}} header.
+    This is the default value.
+
+- `require-corp`
+  - : A document can only load resources requested in `no-cors` mode from the same origin, or resources that have explicitly set the {{HTTPHeader("Cross-Origin-Resource-Policy")}} header to a value that allows it to be embedded.
+
+    Cross-origin resource loading will be blocked by COEP unless:
+    - The resource is requested in `no-cors` mode and the response includes a {{HTTPHeader("Cross-Origin-Resource-Policy")}} header that allows it to be loaded into the document origin.
+    - The resource is requested in `cors` mode, for example, in HTML using the [`crossorigin`](/en-US/docs/Web/HTML/Reference/Attributes/crossorigin) attribute, or in JavaScript by making a request with [`{mode="cors"}`](/en-US/docs/Web/API/RequestInit#cors).
```
sideshowbarker:
```suggestion
    - The resource is requested in `cors` mode; for example, in HTML using the [`crossorigin`](/en-US/docs/Web/HTML/Reference/Attributes/crossorigin) attribute, or in JavaScript by making a request with [`{mode="cors"}`](/en-US/docs/Web/API/RequestInit#cors).
```

files/en-us/web/api/coepviolationreport/index.md
```diff
@@ -0,0 +1,216 @@
...
+  - : The body of the report, containing more information about the violation.
+    This is an object with the following properties:
+    - `type`
+      - : A string representing the cause of the violation that triggered the report.
+        This can have one of the following values:
+        - `"corp"`
+          - : A document with {{httpheader("Cross-Origin-Embedder-Policy")}} set to [`require-corp`](/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Embedder-Policy#require-corp) attempted to load a cross-origin sub-resource that does not explicitly allow itself to be embedded (by setting an appropriate {{httpheader("Cross-Origin-Resource-Policy")}}).
+        - `"navigation"`
+          - : An {{htmlelement("iframe")}} in a document that has either COEP `require-corp` or `credentialless` directives loads a document that:
+            - Has neither COEP `require-corp` nor `credentialless` directive
```
sideshowbarker:
```suggestion
            - Has neither the COEP `require-corp` nor `credentialless` directive
```

files/en-us/web/api/coepviolationreport/index.md
```diff
@@ -0,0 +1,216 @@
...
+        This can have one of the following values:
+        - `"enforce"`
+          - : The violation caused loading of the embedded resource to be blocked.
+            This is set for violations of policies set with {{httpheader("Cross-Origin-Embedder-Policy")}}.
+        - `"reporting"`
+          - : The violation was reported without blocking the resource from loading.
+            This is set for violations of policies set with {{httpheader("Cross-Origin-Embedder-Policy-Report-Only")}}.
+
+- `type`
+  - : The string `"coep"` indicating that this is a COEP violation report.
```
sideshowbarker:
```suggestion
  - : The string `"coep"`, indicating that this is a COEP violation report.
```

sideshowbarker (approved):

camillelamy:
> @sideshowbarker @yoavweiss @camillelamy This documentation on COEP and COEP reporting is ready for review. I'd appreciate some sanity checking if you have time.
> 
> Sorry it took so long! I plan on looking at it again on Friday as doubtless there are some typos etc, but the broad story should be right. Note that this borrows from the structure of the other docs - the HTTP headers dicuss the reports, but the detail on how to get them and what the properties are is documented as part of the Reporting API in the relevant `body` object.

Thanks! I had a look and it looks good.

hamishwillee:
Thanks @sideshowbarker (and @camillelamy ) for review - I have accepted all suggestions.

PR #39882, opened 2025-06-10
Remove duplicate word on NodeList page

caugner:
### Description

Removes a duplicate word on `NodeList` page.

### Motivation

Noticed while looking at https://github.com/mdn/translated-content-de/issues/94.


caugner (commented):
Self-approval.

PR #39884, opened 2025-06-10
Fix cspell timeout

Josh-Cena:
This regex is causing superlinear move; fixed it to what I believe to be the correct logic.

I chose to not have any special treatment for `\\`, because backslashes are not escape characters in HTML.


bsmth (approved):
LGTM, tested using:

```
yarn lint:typos-group-by-file
```

PR #39896, opened 2025-06-11
Add scrim-inline to list of allowed elements

chrisdavidmills:
### Description

This PR adds "scrim-inline" to our markdownlint list of allowed elements. This will needed in the near future when we look to experiment with embedded scrims for purposes such as learning assessments.


fiji-flo (approved):
👍

PR #39757, opened 2025-06-03
Remove inline script/style (API A–F)

Josh-Cena:
See https://github.com/mdn/content/pull/39744.

PR #39909, opened 2025-06-12
Fix element ID

KenStoneBlue:
…ml canvas element id

The canvas element id is 'tutorial'.
But the simple example is doing a get element by id of 'canvas' Changing the get element by id to 'tutorial' will allow the simple example to run and draw the two intersecting rectangles.

### Description

The canvas element id is 'tutorial'.
But the simple example is doing a get element by id of 'canvas' Changing the get element by id to 'tutorial' will allow the simple example to run and draw the two intersecting rectangles.

### Motivation

Currently the simple example does not run because the canvas element id does not match the javascript simple example.

### DIFF
--- a/files/en-us/web/api/canvas_api/tutorial/basic_usage/index.md
+++ b/files/en-us/web/api/canvas_api/tutorial/basic_usage/index.md
@@ -131,7 +131,7 @@ canvas {
 
 ```js
 function draw() {
-  const canvas = document.getElementById("canvas");
+  const canvas = document.getElementById("tutorial");
   if (canvas.getContext) {
     const ctx = canvas.getContext("2d");
 
--- a/files/en-us/web/api/canvas_api/tutorial/basic_usage/index.md
+++ b/files/en-us/web/api/canvas_api/tutorial/basic_usage/index.md
@@ -11,7 +11,7 @@ Let's start this tutorial by looking at the {{HTMLElement("canvas")}} {{Glossary
 ## The `<canvas>` element
 
 ```html
-<canvas id="tutorial" width="150" height="150"></canvas>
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for filing! This is a regression introduced in https://github.com/mdn/content/pull/39757, which changed the `id` to `"tutorial"` but didn't update the JavaScript. 

The only issue I have with this PR is that this page is part of a series, which follow the same pattern (see e.g. the example in https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Applying_styles_and_colors#a_fillstyle_example) and it would be good for them all to use the same `id` value of `"tutorial"`.

I don't know if you want to extend this PR to update all the other pages in this tutorial If you don't I'll just merge this and take care of the others :).

Josh-Cena:
Apparently, this is the only page that uses `id="tutorial"` while all other pages use `canvas`... Much easier to be updated in the other direction

KenStoneBlue:
Looks good to me thanks Josh and wbamberg

wbamberg:
> Apparently, this is the only page that uses `id="tutorial"` while all other pages use `canvas`... Much easier to be updated in the other direction

The reason I didn't suggest that was that https://github.com/mdn/content/pull/39757 changed the `id` from `"canvas"` to `"tutorial"` and I had assumed there was a reason behind that. If there isn't, I agree, easier to revert this change.

Josh-Cena:
The reason was because the init template before that used `id="tutorial"` so I copied that as the source of truth... but turned out that was actually the odd one.

wbamberg (approved):
👍 thank you!

PR #39919, opened 2025-06-12
Use WebP icons in web app manifest

fulldecent:
### Description

This updates the web app manifest from recommending PNG images to instead recommend WebP images.

### Motivation

The [MDN image file type and format guide](https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Image_types#webp_image) recommends WebP over PNG.

### Additional details

N/A

### DIFF
--- a/files/en-us/web/progressive_web_apps/manifest/index.md
+++ b/files/en-us/web/progressive_web_apps/manifest/index.md
@@ -31,14 +31,14 @@ All members are optional in the specification, but some applications require som
   "name": "MDN Web Docs",
   "icons": [
     {
-      "src": "/favicon-192x192.png",
+      "src": "/favicon-192x192.webp",
       "sizes": "192x192",
-      "type": "image/png"
+      "type": "image/webp"
     },
     {
-      "src": "/favicon-512x512.png",
+      "src": "/favicon-512x512.webp",
       "sizes": "512x512",
-      "type": "image/png"
+      "type": "image/webp"
     }
   ],
   "start_url": ".",


hamishwillee (commented):
Thanks @fulldecent .

I don't think we make this recommendation for manifests because I am fairly sure that while most browsers support WebP in content, there isn't as broad support for webp in favicons.

Note that It depends both on support in browsers for this format in a manifest, but also in the desktop and mobile OS shell where this might appear.
Support also has to have both wide coverage in terms of OS and browser and also some depth (i.e. for some releases) or this wouldn't make sense.

Unfortunately I don't know because MDN does not keep this kind of data.

@dipikabh might know if this is a reasonable recommendation.

fulldecent:
This is perhaps the most cited source for which browsers support WebP: https://caniuse.com/webp

And this PR does not recommend changing the favicon.ico file. But nowadays many websites are using WebP favicons for the `<link rel="icon" ...` on their site. (Source: [GitHub search](https://github.com/search?q=rel%3D%22icon%22+%22image%2Fwebp%22&type=code))

W3C does have a [WebP in their manifest file](https://github.com/w3c/manifest/blob/8ff396e78636b370a4137f2f490738812ae6348e/explainer.md?plain=1#L55).

hamishwillee:
Thanks @fulldecent . The caniuse means nothing much - that's browser not OS support, and also pretty much "can display in content".

The W3C explainer is perhaps a _little_ more compelling. Not much though, because explainers have nowhere near the rigor of specifications.

dipikabh:
Agree with your take, @hamishwillee, and the need to be cautious regarding webp support.

WebP is definitely a modern format, but favicons are a special case.
We still recommend ico, gif, and png for favicons ([in guide](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Webpage_metadata#adding_custom_icons_to_your_site), [in glossary](https://developer.mozilla.org/en-US/docs/Glossary/Favicon)). So in this case, I'd suggest not to switch to webp for the favicon.

Also, as a side note, the best practice when using webp would be to also specify a fallback image format.

hamishwillee:
Thanks @dipikabh 
It's a bit of a chicken and egg thing. I suspect the favicon will never need to replace .ico because in a 16x16 icon there is no compression benefit to webp (or PNG). But as the icons get bigger there might be benefit. If all your app imagery is in webp it is going to be tempting to keep your icons in the same format.

So at some point we should verify and possibly update the glossary and guide recommendations. For now though, I think manifest usage is more risky than any other usage.

> Also, as a side note, the best practice when using webp would be to also specify a fallback image format.

I'd be more fine with this if it had a fallback.

@fulldecent I'm going to close this as "too early". Happy for it to be reopened with some verification about the platforms on which it is known to work and a fallback.
Thanks again for contributing - really appreciate you going to the effort of not just noting the possible shortcoming, but adding a fix in a PR.

PR #39920, opened 2025-06-12
Fix content issues

Josh-Cena:
Fix recently introduced content issues


pepelsbey (approved):
Good catch! Thank you :)

PR #39924, opened 2025-06-13
fix: remove web audio tools page from sidebar

Nirajn2311:
### Description

Remove web audio tools from sidebar as the page has been removed and the sidebar shows a dummy item in place of it under the Guides section.

### Additional details

MDN page link - https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API


Josh-Cena (approved):
Good catch. Thanks

PR #39804, opened 2025-06-05
feat(html): Move meta element name attrs out into separate pages

bsmth:
### Description

This PR makes several changes to `<meta name=` docs.

### Motivation

The standard metadata names page has a lot of information that's hard to read. This PR tidies up the definition lists by bringing the types out into subpages.

### Additional details

__Additions:__
* new page for `html/reference/attributes/content`
* new pages for `meta name` types (mostly extracted from [html/reference/elements/meta/name](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name))
  * `html/reference/elements/meta/name/color-scheme`
  * `html/reference/elements/meta/name/referrer`
  * `html/reference/elements/meta/name/robots`
  * `html/reference/elements/meta/name/theme-color`
  * `html/reference/elements/meta/name/viewport`

__Modifications:__
* Short titles for `script/type` pages
* link to script types and meta names in sidebars

PR #39929, opened 2025-06-13
chore(html): Add minor clarification for media query example

bsmth:
### Description

I noticed we're talking about media queries on this page before we show an example without a media query.

### Motivation

Clarity about the basic example without any media queries. This should help readers understand the general purpose first before applying any additional logic. It's a small, but important change for comprehension, IMO.

- [x] https://github.com/mdn/content/pull/39804


dipikabh (approved):
Thanks 👍

PR #39941, opened 2025-06-14
Fix the incorrect meaning of the `title`

liuxuelian:


chrisdavidmills (approved):
@liuxuelian good spot; thank you for this fix!

I tested it in three browsers, and sure enough, you are correct.

PR #39942, opened 2025-06-14
Update index.md (Permitted Content) for address element

dwilbourne:
added aside element as illegal descendant of address element in the Permitted Content

### Additional details

https://www.w3.org/TR/2011/WD-html-markup-20110525/aside.html#:~:text=Additional%20constraints%20and%20admonitions,tag%20and%20an%20end%20tag.


Josh-Cena:
This is already included in the "sectioning content" list.

Just FYI, the [WHATWG spec](https://html.spec.whatwg.org/multipage/sections.html#the-address-element) is the canonical reference.

PR #39944, opened 2025-06-14
Update index.md

asummers1:
### Description

I added additional information concerning function behavior based on reading the DOM specification.
### Motivation

Users who are reading the docs might assume that elements are returned in a specific order (e.g. the order of the elements in the DOM).

### Additional details

https://dom.spec.whatwg.org/#ref-for-dom-document-getelementsbyclassname%E2%91%A0


Josh-Cena:
They _are_ in tree order. Please see https://dom.spec.whatwg.org/#concept-collection:

> The view is linear. In the absence of specific requirements to the contrary, the nodes within the [collection](https://dom.spec.whatwg.org/#concept-collection) must be sorted in [tree order](https://dom.spec.whatwg.org/#concept-tree-order).

PR #39947, opened 2025-06-15
Omitting roundingInterval

tim-we:
### Description

Improved the description of the `roundingInterval` argument of the CSS `round` function. I explained what happens when you omit the value.

### Motivation

In the syntax definition the argument is optional so I was wondering what would happen. Had to look at the spec to find out.

### Additional details

Source (from the spec):
> If the [type](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-type) of A matches [<number>](https://drafts.csswg.org/css-values/#number-value), then B may be omitted, and defaults to 1; omitting B is otherwise invalid.

https://drafts.csswg.org/css-values/#round-func

### DIFF
--- a/files/en-us/web/css/round/index.md
+++ b/files/en-us/web/css/round/index.md
@@ -48,7 +48,7 @@ The `valueToRound` is rounded according to the rounding strategy, to the nearest
 
 - `roundingInterval`
   - : The rounding interval.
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
Hello @tim-we, and thank you for your contribution to MDN!

I've reviewed your addition and the specification wording, and I've conducted some testing with `round()`. What you say looks correct.

Let's get it merged.

PR #39949, opened 2025-06-15
Fix PublicKeyCredentialCreationOptions.pubKeyCredParams.alg

lennartkloock:
### Description

This PR fixes the description for `PublicKeyCredentialCreationOptions.pubKeyCredParams.alg`.
The number `-8` does not identify the `Ed25519` algorithm but `EdDSA` (see the linked IANA document).
`Ed25519` has number `-19`.

### Motivation

This change was made to clarify which algorithm is referred to in the section.

### Additional details

- https://www.iana.org/assignments/cose/cose.xhtml#algorithms

### DIFF
--- a/files/en-us/web/api/publickeycredentialcreationoptions/index.md
+++ b/files/en-us/web/api/publickeycredentialcreationoptions/index.md
@@ -125,7 +125,7 @@ The **`PublicKeyCredentialCreationOptions`** dictionary represents the object pa
 
       - : A number that is equal to a [COSE Algorithm Identifier](https://www.iana.org/assignments/cose/cose.xhtml#algorithms), representing the cryptographic algorithm to use for this credential type. It is recommended that relying parties that wish to support a wide range of authenticators should include at least the following values in the provided choices:
 
-        - `-8`: Ed25519
+        - `-8`: EdDSA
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):
Nice catch. I think you also caught a bug in the spec — because it seems the reason we have it wrong in MDN is that we based the docs on what the spec has at https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-pubkeycredparams

So please also consider raising a spec bug at https://github.com/w3c/webauthn/issues/new/choose

sideshowbarker:
Lennart, thanks much for catching this and fixing it, and congrats on landing your first docs change here — welcome aboard 🎉

lennartkloock:
> Nice catch. I think you also caught a bug in the spec — because it seems the reason we have it wrong in MDN is that we based the docs on what the spec has at https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-pubkeycredparams
> 
> So please also consider raising a spec bug at https://github.com/w3c/webauthn/issues/new/choose

Thank you! I just opened a PR on their repo. https://github.com/w3c/webauthn/pull/2304

PR #39959, opened 2025-06-17
Update index.md

trrapp12:
adds some clarifying comments about reading reduce examples in the wild when they are written in ES6 syntax

### Description

Adds an example of a reducer with an implicit and explicit return to demonstrate what a reduce function may look like in the wild

### Motivation

It took me a long time to differentiate .map() and .forEach() from a reducer function because everything was written in arrow functions, so it took me forever to catch on that I actually needed to return something from my reducer EVERYTIME, instead of what I was thinking it did (create side effects within the reducer like a map or forEach)

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/array/reduce/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/array/reduce/index.md
@@ -62,13 +62,29 @@ The value that results from running the "reducer" callback function to completio
 
 ## Description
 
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Hi, this is reference docs, which assumes you already know everything about JavaScript, and certainly is not the place to learn about function syntax. If you aren't familiar with concise arrow functions, please read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions or https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions instead.

trrapp12:
That is a ridiculous and naive statement.  Javascript beginners come to MDN all the time to try and learn.  This sounds like the voice of someone who is  a little full of themselves and has forgotten what it was like to have to learn on your own.

Josh-Cena:
Sure, but the reference docs is not the place to "try and learn". We have the "tutorials and guides" section for that purpose.

PR #39962, opened 2025-06-17
New guide: Mask properties

estelle:
Ready for review!


files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
+---
+title: CSS mask properties
+slug: Web/CSS/CSS_masking/Mask_properties
+page-type: guide
+---
+
+{{CSSRef}}
+
+CSS masking is a technique that enables you to define visible portions of an element by applying a mask, which selectively reveals or hides parts of the element based on the alpha channels, and optionally colors, of applied mask image or images.
```
chrisdavidmills:
```suggestion
CSS masking is a technique that enables you to define visible portions of an element by applying a mask, which selectively reveals or hides parts of the element based on the alpha channels and optionally colors of the applied mask images.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+The [introductory guide to masking](/en-US/docs/Web/CSS/CSS_masking) introduces the different types of mask images and their modes. The guide to [declaring multiple masks](/en-US/docs/Web/CSS/CSS_masking/Multiple_masks) discusses the [mask layers](/en-US/docs/Web/CSS/CSS_masking/Multiple_masks#understanding_mask_layers) and the {{cssxref("mask")}} shorthand property, providing a brief introduction to the shorthand's component properties. In this guide we explore these component properties in greater deal and look at how these properties interact and how, in cases where multiple mask images are declared, the [mask layers are composited](#the_mask-composite_property), or combined.
```
chrisdavidmills:
```suggestion
The [introductory guide to masking](/en-US/docs/Web/CSS/CSS_masking) introduces the different types of mask images and their modes. The guide to [declaring multiple masks](/en-US/docs/Web/CSS/CSS_masking/Multiple_masks) discusses the [mask layers](/en-US/docs/Web/CSS/CSS_masking/Multiple_masks#understanding_mask_layers) and the {{cssxref("mask")}} shorthand property, providing a brief introduction to the shorthand's component properties. In this guide, we explore these component properties in greater detail and look at how they interact. We also explain how, in cases where multiple mask images are declared, the [mask layers are composited](#the_mask-composite_property), or combined.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+CSS masks are composed of one or more mask layers, with a mask layer created for every value in the comma separated list of `mask` or `mask-image` values, whether the value is an image or mask source or the keyword `none`. Every {{cssxref("mask-image")}} is [positioned](#the_mask-position_property) relative to an [origin](#the_mask-origin_property) box. The mask images can be [sized](#the_mask-size_property), [repeated](#the_mask-repeat_property), and [clipped](#the_mask-clip_property), then composited together with previous layers to create the final visual mask on the element.
```
chrisdavidmills:
```suggestion
CSS masks are composed of one or more mask layers, with a mask layer created for every value in the comma-separated list of `mask` or `mask-image` values, whether the values are images, mask sources, or the keyword `none`. Every {{cssxref("mask-image")}} is [positioned](#the_mask-position_property) relative to an [origin](#the_mask-origin_property) box. The mask images can be [sized](#the_mask-size_property), [repeated](#the_mask-repeat_property), and [clipped](#the_mask-clip_property), then composited together with previous layers to create the final visual mask on the element.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+## The `mask-image` property
+
+The minimum required to create a mask is a {{cssxref("mask-image")}} value with a value other than `none`.
```
chrisdavidmills:
```suggestion
The minimum requirement to create a mask is a {{cssxref("mask-image")}} property set to a value other than `none`.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+The keyword `none` within a list of mask sources creates a mask layer. However, if `none` is the only value of the `mask-image` property, no masking occurs.
+
+The mask image can be a [CSS gradient](/en-US/docs/Web/CSS/CSS_images/Using_CSS_gradients), an [imported image](/en-US/docs/Web/CSS/CSS_masking/Masking#with_impoorted_images) (such as a PNG, SVG, etc.), or an SVG {{svgelment("mask")}} element.
```
chrisdavidmills:
```suggestion
The mask image can be a [CSS gradient](/en-US/docs/Web/CSS/CSS_images/Using_CSS_gradients), an [imported image](/en-US/docs/Web/CSS/CSS_masking/Masking#with_impoorted_images) (such as a PNG, SVG, etc.), or an SVG {{svgelement("mask")}} element.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+```css
+.masked-element {
+  mask-image:
+    url(alphaImage.png), linear-gradient(to right, black, transparent),
+    radial-gradient(circle, white 50%, transparent 75%), none, url(#svg-mask);
+}
+```
+
+This declaration creates five mask layers, applying four mask images to the `.masked-element` element.
```
chrisdavidmills:
```suggestion
Because one of the mask images is specified as `none`, only four mask images are applied to the `.masked-element` element.
```

chrisdavidmills: Otherwise it sounds too repetitious.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+### The importance of `none`
+
+The `none` layer has no visual effect, but still serves an important purpose. This fourth layer in our five layers will match the fourth value of any another comma-separated `mask-*` property values. Each value in a comma-separated list of `mask-*` values applies to a separate mask layer.
```
chrisdavidmills:
```suggestion
The `none` layer has no visual effect, but still serves an important purpose. This fourth layer in our five-layer structure will match the fourth value of any other comma-separated `mask-*` property values. Each value in a comma-separated list of `mask-*` values applies to a separate mask layer.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+As previously mentioned, the number of layers is determined by the number of comma-separated values in the {{cssxref("mask-image")}} property value, even if a value is `none`. Each `mask-*` value is matched up with the `mask-image` values, in order. If the number of values in a `mask-*` property differs from the number of mask layers, any excess values are ignored, or, if the property has fewer values than the number of mask layers, the values are repeated.
+
+If a `mask-*` property has a single value, this value applies to all the layers. If we have five values, the fourth value applies to the `none` layer, with the last value being applied to the `<mask>` source layer. If there are two comma-separated values, the first value will apply to all the odd layers only, including that `<mask>` source layer.
```
chrisdavidmills: I'd show a couple of actual code examples interspersed with explanatory sentences. As it is, this is not that easy to visualize.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+Learn more about [mask layers and the `none` keyword](/en-US/docs/Web/CSS/CSS_masking/Multiple_masks#mask_layers_and_the_none_keyword).
+
+## The `mask-mode` property
+
+The {{cssxref("mask-mode")}} property can be used to set the mode of each mask layer to either `alpha` or `luminance`, or allow it to default to the source's mode by setting the value to `match-source`, which is the default. While most `mask-*` properties have an analogous `background-*` property, such as the `mask-image`to the {{cssxref("background-image")}} property, `mask-mode` property (and [the `mask-composite` property](#the_mask-composite_property)) has no property in the {{cssxref("background")}} shorthand that is analogous.
```
chrisdavidmills:
```suggestion
The {{cssxref("mask-mode")}} property can be used to set the mode of each mask layer to either `alpha` or `luminance`, or allow it to default to the source's mode by setting the value to `match-source`, which is the default. While most `mask-*` properties have an analogous `background-*` property (`mask-image` is analogous to the {{cssxref("background-image")}} property, for example), `mask-mode` and [`mask-composite`](#the_mask-composite_property) have no analogous {{cssxref("background")}} property.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+### Mask types: `alpha` and `luminance`
+
+Each mask is either an `alpha` or a `luminance` mask.
+
+With alpha masks, the alpha-transparency of each mask pixel is important. Wherever the mask is opaque the corresponding parts of the element will be visible. Wherever the mask is transparent, the corresponding parts of the element will be hidden. Wherever the mask is semi-opaque the element will be equally semi-opaque. The color of the mask doesn't matter; just the alpha-transparency of the colors.
```
chrisdavidmills:
```suggestion
With `alpha` masks, the alpha-transparency of each mask pixel is important. Wherever the mask is opaque, the corresponding parts of the element will be visible. Wherever the mask is transparent, the corresponding parts of the element will be hidden. Wherever the mask is semi-opaque, the element will be equally semi-opaque. The color of the mask doesn't matter, just the alpha-transparency of the colors.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+With `luminance` masks, both the [brightness of mask's colors](/en-US/docs/Web/CSS/CSS_masking/Masking#alpha_transparency_versus_luminance) and the alpha channel determine the opaqueness of the masked areas.
+
+> [!NOTE]
+> All examples with be using the following image as the underlying element upon which masks will be applied:
```
chrisdavidmills:
```suggestion
> All subsequent examples will be using the following image as a `background-image` on an element to which masks will be applied:
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+### The `mask-mode` default value: `match-source`
+
+The default value of the `mask-mode` property is `match-source`. This value sets the `mask-mode` to match the mask's mode type. The `match-source` value resolves to `alpha` for every mask except masks where the mask source is an SVG {{svgelement("mask")}} element. In the case of `<mask>`, the `match-source` resolves to the value of the `<mask>` element's {{cssxref("mask-type")}} property. If the `<mask>` element (not the "masked element") doesn't have the `mask-type` property defined, this property defaults to the value of the SVG {{svgAttr("mask-type")}} attribute, if present. If that is omitted as well, the `match-source` value will resolve to `luminance`.
```
chrisdavidmills: This is a lot to follow for one paragraph. I wonder if you can break down the steps by which the `mask-mode` of a SVG `<mask>` element mask source is resolved into some bullet points?

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+Continuing with the `masked-element` example, if we don't explicitly set the `mask-mode` property, it will default to `match-source` for each layer, as if we had set the following:
```
chrisdavidmills, line 161: Some of this stuff seems kind of obvious, as you've already said that the default value is `match-source`.

Also, I've looked down the page and seen that you've done a similar, incredibly in-depth, analysis of each property and how it matches with the mask layers, and what the shorthand equivalent is. It does come across rather long-winded. Couldn't you cut down the verbosity a bit by making the example in each section a bit shorter, and then having the in-depth analysis and shorthand equivalent stuff at the end, in one section, all together?

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+The fourth mask layer has `none` declared, meaning this layer creates no masking effect. The `.masked-element` sets `mask-mode: match-source;`. Had this instead been a comma-separated list of values, the fourth value would have applied to this `none` layer, allowing the fifth value to apply to the fifth layer.
+
+The fifth mask layer is comprised of an SVG {{svgelement("mask")}} element that has `svg-mask` as it's [`id`](/en-US/docs/Web/HTML/Reference/Global_attributes/id). While default mask mode of the other layers is `alpha`, the default [mask type of SVG `<mask>` elements](/en-US/docs/Web/CSS/CSS_masking/Masking#svg_mask_as_mask_source) the `mask-type` value, or, if not set, the `mask-type` attribute. If that isn't defined either, the value defaults `luminance`. In other words, the masking effect of the `<mask>` is determined by both the brightness and transparency of the `<mask>` element's colors.
```
chrisdavidmills:
```suggestion
The fifth mask layer is comprised of an SVG {{svgelement("mask")}} element that has `svg-mask` as its [`id`](/en-US/docs/Web/HTML/Reference/Global_attributes/id). While the default mask mode of the other layers is `alpha`, the default [mask type of SVG `<mask>` elements](/en-US/docs/Web/CSS/CSS_masking/Masking#svg_mask_as_mask_source) is the `mask-type` value, or, if not set, the `mask-type` attribute. If that isn't defined either, the value defaults to `luminance`. In other words, the masking effect of the `<mask>` is determined by both the brightness and transparency of the `<mask>` element's colors.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+  mask:
+    url(alphaImage.png) match-source,
+    linear-gradient(to right, black, transparent) match-source,
+    radial-gradient(circle, white 50%, transparent 75%) match-source,
+    none match-source,
+    url(#svg-mask) match-source;
+}
+```
+
+The first mask layer, `url(alphaImage.png)`, references an image. As this isn't a `<mask>` element within an `<svg>`, the `mask-mode` resolves to `alpha`, with the opaque parts of this image making the corresponding parts of the element visible, while the transparent or semi-transparent parts are invisible or partially visible.
```
chrisdavidmills, line 185: I think some of this stuff could be cut down a bit. You've already explained how they resolve above.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+    linear-gradient(to right, black, transparent) alpha,
+    radial-gradient(circle, white 50%, transparent 75%) alpha,
+    none match-source,
+    url(#svg-mask) luminance;
+}
+```
+
+## The `mask-position` property
+
+Analogous to the {{cssxref("background-position")}} property, the {{cssxref("mask-position")}} property sets the initial position of the mask image relative to the mask layer's origin box, defined by [the `mask-origin` property](#the_mask-origin_property). Just like `background-position`, the value is one, two, or four {{cssxref("&lt;position&gt;")}} values, which defines one to two relative or absolute position offsets.
```
chrisdavidmills:
```suggestion
Analogous to the {{cssxref("background-position")}} property, the {{cssxref("mask-position")}} property sets the initial position of the mask image relative to the mask layer's origin box, defined by [the `mask-origin` property](#the_mask-origin_property). Just like `background-position`, the value is one, two, or four {{cssxref("&lt;position&gt;")}} values, which define one to two relative or absolute position offsets.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+The syntax follows the [`background-position`'s `<position>` syntax](/en-US/docs/Web/CSS/background-position#position).
+
+If only one keyword value is present, it specifies the mask origin edge against which the mask will be placed or the center in the case of `center`, with the other dimension being `center`. If the single value is a {{cssxref("&lt;length-percentage&gt;")}}, this specifies the X coordinate relative to the left edge of the mask origin, with the Y coordinate set to `50%`.
+
+If two values are present, if both are keyword values, the order doesn't matter, but the value can not include two vertical or two horizontal axis values; `left right` and `top bottom` are both invalid. If one value is a keyword, and the other is a `<length-percentage>`, the order again doesn't matter. If the keyword is `left` or `right`, this defines the X coordinate relative to the left edge and the value defines the Y coordinate relative to the top edge. Similarly, if `top` or `bottom` is included, this defines the Y coordinate positioning the element against the top or bottom edge respectively, with the other value defining the X value relative to the left edge of the mask origin box.
```
chrisdavidmills: Start off by saying what the two values represent.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+}
+```
+
+## The `mask-position` property
+
+Analogous to the {{cssxref("background-position")}} property, the {{cssxref("mask-position")}} property sets the initial position of the mask image relative to the mask layer's origin box, defined by [the `mask-origin` property](#the_mask-origin_property). Just like `background-position`, the value is one, two, or four {{cssxref("&lt;position&gt;")}} values, which defines one to two relative or absolute position offsets.
+
+The syntax follows the [`background-position`'s `<position>` syntax](/en-US/docs/Web/CSS/background-position#position).
+
+If only one keyword value is present, it specifies the mask origin edge against which the mask will be placed or the center in the case of `center`, with the other dimension being `center`. If the single value is a {{cssxref("&lt;length-percentage&gt;")}}, this specifies the X coordinate relative to the left edge of the mask origin, with the Y coordinate set to `50%`.
```
chrisdavidmills: I wonder whether it is worth repeating all this stuff, when you explain it in detail on the reference page anyway. Maybe include a more abbreviated version here?

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+With the two value syntax, if one is `center` and the other is a `<length-percentage>`, or if both are `<length-percentage>` values, the order _does_ matter; with the first value defining the horizontal positioning and the second value defining the vertical position.
+
+The four value syntax includes two {{cssxref("length-percentage")}} offsets each preceded by the origin side for that offset. In the two `<length-percentage>` syntax, the origin sides are not included and the position originates from the top left of the mask origin box. For example, `10px 20px` omits the origin sides, but is the equivalent of `left 10px top 20px`. In this case, the offset sides aren't required, but the four value `mask-position` enables offsetting the mask from other edge combinations, such as `left 10px bottom 20px`. In the four-syntax value, the edge keyword must precede the offset value, but the order of the X and Y coordinate values don't matter, as the offset edge is defined by the keyword preceding it rather than the declaration order.
```
chrisdavidmills: Explain this as two length/offset pairs?

chrisdavidmills: This paragraph is really hard to parse.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+When offsetting using percentage values, the mask's dimension is subtracted from the element's dimension, just as is done with [percentage offsets with `background-position`](/en-US/docs/Web/CSS/background-position#regarding_percentages).
+
+The `mask-position` property defines only the initial position of the mask image. By "initial position", if the [mask repeats](#the_mask-repeat_property), the browser places the first mask image in the position defined by the `mask-position` property, thus defining the placement of the mask repetitions.
```
chrisdavidmills: You could probably remove most of the content above this point in the section, and point to the reference page instead. What follows this is concise and useful.

chrisdavidmills:
```suggestion
The `mask-position` property defines the initial position of the mask image. By "initial position", if the [mask repeats](#the_mask-repeat_property), the browser places the first mask image in the position defined by the `mask-position` property, thus defining the placement of the mask repetitions.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+}
+.twoValue {
+  mask-position: -20px -10px;
+}
+.fourValue {
+  mask-position: right -20px bottom -10px;
+}
+```
+
+The `mask-position` defines the position of the first mask image's placement:
```
chrisdavidmills: You need to explain which of the above rules relates to which image in the rendered example. As-is, this is confusing.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+```css live-sample___position_no-repeat
+img {
+  mask-repeat: no-repeat;
```
chrisdavidmills, line 311: Include this in the above `img` rule, along with the `mask-image`?

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+}
+```
+
+{{EmbedLiveSample("position_no-repeat", "", "250px")}}
+
+By default, the images are then repeated:
```
chrisdavidmills: This is also confusing. Given the above code, the images would not be repeated. I think you should say in a note somewhere that the images are repeated by default, hence us needing to set `mask-repeat: no-repeat;`

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+    linear-gradient(to right, black, transparent) 0% 0% match-source,
+    radial-gradient(circle, white 50%, transparent 75%) 0% 0% match-source,
+    none 0% 0% match-source,
+    url(#svg-mask) 0% 0% match-source;
+}
+```
+
+## The `mask-origin` property
+
+When an element has padding, border, or both, the {{cssxref("mask-origin")}} property defines which of these box edge values acts as the mask origin box for that mask layer. In the [`mask-position`](#the_mask-position_property) masking example, the position defined was relative to the content-box, padding-box, and border-box, as the `<img>` had no border or padding set, so these three [`<coord-box>`](/en-US/docs/Web/CSS/box-edge#values) values were all same (the `<div>` container did have a `1px` border).
```
chrisdavidmills: "the position defined was relative to the content-box, padding-box, and border-box" — surely they are not positioned relative to all three simultaneously? They would just be relative to the default origin? This is very confusing.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+Analogous to the {{cssxref("background-origin")}} property, but with a different initial value and SVG-only values, the {{cssxref("mask-origin")}} property specifies the _mask positioning area_, which is the mask origin box area within which a mask image is positioned. HTML elements can have masks contained within their content border box, padding box, or content box. The `mask-origin` property sets the origin of a mask, determining the origin of the `mask-position` property, also known as the _mask positioning area_. For example, if the `mask-position` is `top left`, is that relative to the border's outer edge, the padding's outer edge, or the content's outer edge?
```
chrisdavidmills: This paragraph repeats the same information twice and is therefore confusing and hard to follow. You should cut it down to one explanation.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+}
+legend {
+  align-self: baseline;
+}
+label {
+  display: block;
+}
+```
+
+In this example, the `mask-position` places the initial mask in the top left corner of the `<img>` element that has a large border and padding, with a green background color to enable seeing the star masking on the padding area. Change the value of the `mask-origin` property to observe the difference the property can make.
```
chrisdavidmills:
```suggestion
In this example, the `mask-position` places the initial mask in the top left corner of the `<img>` element that has a large border and padding, with a green background color to enable seeing the star masking on the padding area.
```

You effectively say the same thing just above the live example.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+:has(#origin_content-box:checked) img {
+  mask-origin: content-box;
+}
+```
+
+Change the value of the `mask-origin` property by changing the selected radio button, looking at the top left point of the mask top left star as you do so.
+
+{{EmbedLiveSample("origin", "", "350px")}}
+
+The default value is `border-box`. With this value, the initial mask is placed at the borders top left edge and is not clipped. When the initial mask is placed at the outer or inner edge of the padding, there is room above it and to the left; these repeating masks are clipped.
```
chrisdavidmills:
```suggestion
The default value is `border-box`. With this value, the initial mask is placed at the border's top-left edge and is not clipped. When the initial mask is placed at the outer or inner edge of the padding, there is room above it and to the left; these repeating masks are clipped.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+By default, HTML elements have masks their masks positioned relative to the `border-box`, which is the outer edge of the border. Continuing with the `masked-element` example, if we don't explicitly set the `mask-origin` property, it will default to `border-box` for each layer, as if we had set the following:
```
chrisdavidmills: You already said "The default value is `border-box`" earlier on. This feels unnecessarily repetitive.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+}
+```
+
+For SVG elements, which don't have the associated CSS layout boxes, a mask can be contained inside the SVG element's fill, stroke, or view box.
+
+## The `mask-clip` property
+
+The {{cssxref("mask-clip")}} property determines the area of the element that will be affected by a mask, effectively clipping the element at the defined box edge. It is analogous to the {{cssxref("background-clip")}} property, but with some different values.
+
+Because the `mask-clip` property accepts all the `mask-origin` values and both have the same `border-box` default value, the two properties may seem similar, but they serve very different purposes. While `mask-origin` determines where the first mask image will be positioned, the `mask-clip` property defines which area of the original element will be painted with the painted content clipped to this area. It is important to understand them both, as if the mask-origin causes the mask-position to place the mask image outside the clipping area, the mask will be clipped.
```
chrisdavidmills:
```suggestion
Because the `mask-clip` property accepts all the `mask-origin` values and both have the same `border-box` default value, the two properties may seem similar, but they serve very different purposes. While `mask-origin` determines where a mask image will be positioned, the `mask-clip` property causes the original element to have its content clipped to the specified box. It is important to understand them both, as if the mask-origin causes the mask-position to place the mask image outside the clipping area, the mask will be clipped.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+The `mask-clip` property accepts all the `mask-origin` values, as well as it's own `no-clip` value. The `no-clip` value sets the painted content to not be clipped, meaning the painted content is not restricted and isn't clipped by default. You can still cause the mask image to be clipped by positioning it outside of the border content area using `mask-position` values that are less than zero or resolve to greater than 100%.
```
chrisdavidmills:
```suggestion
The `mask-clip` property accepts all the `mask-origin` values, as well as its own `no-clip` value. The `no-clip` value sets the painted content not to be clipped. You can still cause the mask image to be clipped by positioning it outside of the border content area using `mask-position` values that are less than zero or resolve to greater than 100%.
```

chrisdavidmills: The removed line was repetitious and made it sound like the default value is `no-clip`, which isn't correct.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+Setting the `mask-clip` and `mask-origin` to different values can cause the mask layer image to be clipped. For example, If an element's has a border and padding, if `mask-clip` is set to `content-box` while `mask-origin` is set to `border-box`, and the `mask-position` is set to the `top left` edge, the mask layer image will be clipped at the top-left edge.
+
+We can add clipping options to the previous example, both to demonstrate both the different non-svg `mask-clip` values and how the different `mask-origin` values are impacted by these values.
```
chrisdavidmills:
```suggestion
The next example adds clipping options to the previous example to demonstrate the different non-SVG `mask-clip` values and to show how they impact the different `mask-origin` values.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+  mask-clip: padding-box;
+}
+:has(#clip_content-box:checked) img {
+  mask-clip: content-box;
+}
+```
+
+{{EmbedLiveSample("clip", "", "350px")}}
+
+The first mask is placed at the top left edge of the mask origin container, then repeated. If the origin box is the border-box while the clipping region is the content-box, the top and left areas of the mask origin container will be clipped. Generally, you will want the mask-clip to be the same as the mask origin.
```
chrisdavidmills:
```suggestion
The first mask is placed at the top-left edge of the mask origin container, then repeated. If the origin box is the `border-box` and the clipping region is the `content-box`, the top and left areas of the mask origin container will be clipped. Generally, you will want the `mask-clip` to be the same as the `mask-origin`.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+    none 0% 0% border-box border-box match-source,
+    url(#svg-mask) 0% 0% border-box border-box match-source;
+}
+```
+
+In the `mask` shorthand, if only one [`<geometry-box>`](/en-US/docs/Web/CSS/clip-path#geometry-box) value is given, it sets both the `mask-origin` and `mask-clip` property values. If two `<geometry-box>` values are present, the first defines the `mask-origin` and the second defines the `mask-clip`.
+
+For mask layer images that do not reference an SVG {{svgelement("mask")}} element, the `mask-clip` property defines whether the mask painting area, or the area affected by the mask, is the border, padding, or content box. The painted content of the element will be restricted to this area.
+
+When the mask layer's {{cssxref("mask-image")}} source is a `<mask>`, the `mask-clip` property has no affect. Rather, the `<mask>` element's {{svgAttr("x")}}, {{svgAttr("y")}}, {{svgAttr("width")}}, {{svgAttr("height")}}, and {{svgAttr("maskUnits")}} attributes determine the mask painting area.
```
chrisdavidmills:
```suggestion
When the mask layer's {{cssxref("mask-image")}} source is a `<mask>`, the `mask-clip` property has no effect. Rather, the `<mask>` element's {{svgAttr("x")}}, {{svgAttr("y")}}, {{svgAttr("width")}}, {{svgAttr("height")}}, and {{svgAttr("maskUnits")}} attributes determine the mask painting area.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+## The `mask-size` property
+
+The {{cssxref("mask-size")}} property is used to size mask layers. This property is analogous to the {{cssxref("background-size")}} property, taking the same values. When sizing your masks, remember that areas of the element not covered by the mask images are hidden.
+
+There are three ways to declare a `mask-size`: the `cover` or `contain` keyword, a length, percentage, or `auto`, or two values that are a combination of lengths, percentages, or `auto`. The mask image can be left to its natural size, stretched, or constrained to fit the available space.
```
chrisdavidmills: This would be better as bullet points for quick reference.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+If the `mask-size` is set to `contain` or `cover`, the mask image is rendered by preserving its {{glossary("aspect ratio")}} at the largest size contained within or covering the mask positioning area. In these cases, the mask image will grow or shrink so that the width is either the same width as the mask positioning area or the height of the image is as tall as the mask positioning area's height. With `contain`, the image is resized until one dimension is the same size as the corresponding mask positioning area's dimension without overflowing the area. With `cover` the image is resized until the second dimensions is the same size as the corresponding mask positioning area's dimension, overflowing the other dimension if the aspect ratio of the element and mask aren't the same.
```
chrisdavidmills: This paragraph is a bit unwieldy and hard to follow. Maybe include a separate paragraph for each keyword, and try to simplify?

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+With `cover`, `contain`, and `<percentage>` values, the size is relative to the origin box. This means, if you set the `mask-size` to `cover`, `contain`, or `100%`, whether the size will be the dimension of the content box, padding box, or border box, depends on the value of the [`mask-origin` property](#the_mask-origin_property).
+
+In our star-mask and flag image example, the aspect ratio of both the mask image and the `<img>` are 1:1, meaning in this case `cover`, `contain`, and `100%` will all produce the same sized mask:
```
chrisdavidmills: Is this the correct example? This is comparing `mask-origin` values, not `mask-size` values. I'd expect to see a `mask-size` value here.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+}
+```
+
+Change the value of the `mask-origin` property to see how the different values impact the mask size:
+
+{{EmbedLiveSample("size", "", "350px")}}
+
+This example included one `<percentage>` value. When one `<length-percentage>` value is specified, it defines only the mask width, with the height defaulting to `auto`, which maintains the aspect ratio. When two values are specified, the first defines the mask width and the second defines its height.
+
+The default value of `mask-size` is `auto`, rendering the mask at its {{glossary("intrinsic size")}}, the size at which the mask would be displayed if no CSS were applied. The underlying {{glossary("aspect ratio")}} of the mask image is maintained except if you declare two length-percentage values that distorts the mask image. This is because, if both components of `mask-size` are specified and are not `auto`, the mask image renders at the specified size.
```
chrisdavidmills:
```suggestion
The default value of `mask-size` is `auto`, rendering the mask at its {{glossary("intrinsic size")}}, the size at which the mask would be displayed if no CSS were applied. The underlying {{glossary("aspect ratio")}} of the mask image is maintained if you set a single `<length-percentage>` value, or two values in the same ratio as the aspect ratio. If you declare two values not in the same ratio as the aspect ratio, the mask image is distorted.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+Like with all longhand components of shorthand property, if the {{cssxref("mask")}} shorthand property is set and the value of the `mask-size` property is not defined within any mask layer, the `mask-size` value is reset to its initial value of `auto` for those mask layers.
+
+If the image has no intrinsic proportion, such as with [CSS gradients](/en-US/docs/Web/CSS/gradient) which have no intrinsic dimensions or proportion, the default `auto` is the entirety of the mask positioning area as set by [the `mask-origin` property](#the_mask-origin_property).
```
chrisdavidmills:
```suggestion
If the image has no intrinsic proportion, for example in the case of a [CSS gradient](/en-US/docs/Web/CSS/gradient), the default `auto` is the entirety of the mask positioning area as set by [the `mask-origin` property](#the_mask-origin_property).
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+    radial-gradient(circle, white 50%, transparent 75%) 0% 0% / auto border-box
+      border-box match-source,
+    none 0% 0% / auto border-box border-box match-source,
+    url(#svg-mask) 0% 0% / auto border-box border-box match-source;
+}
+```
+
+## The `mask-repeat` property
+
+Only in [the `mask-position` example](#the_mask-position_property), where we set `mask-repeat: none;`, was there only a single, non-repeating mask. In all the other examples, the star mask has been repeated along the X and Y axis. Even in the `mask-size` example, if you set the `mask-origin` to any value other than `content-box`, the mask repeated. This is the default.
```
chrisdavidmills: It seems weird to start with this paragraph. Start with what the property does. I'm not convinced we really need this recap.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+The {{cssxref("mask-repeat")}} property defines how mask images are repeated, or tiled, after the initial mask image has been sized and positioned. The first (and possibly only) mask-image repetition is positioned by [the `mask-position` property](#the_mask-position_property) and sized by [the `mask-size` property](#the_mask-size_property). The positions of the repeated masks are based on this initial mask instance. The `mask-repeat` property defines if and how that mask image is repeated along the horizontal axis, the vertical axis, both axes. The `mask-repeat` property is analogous to the {{cssxref("background-repeat")}} property, accepting the same [`<repeat-style>`](/en-US/docs/Web/CSS/mask-repeat#values) values.
```
chrisdavidmills: I'm finding this really hard to understand. Don't `mask-repeat` values match up to the `mask-image` values (repeated or ignored, depending on how many there are) in the same way as the other `mask-*` values? This explanation sounds like there is something special going on, which I'm not convinced there is. This could be simplified, surely?

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+    radial-gradient(circle, white 50%, transparent 75%) 0% 0% / auto repeat
+      border-box border-box match-source,
+    none 0% 0% / auto repeat border-box border-box match-source,
+    url(#svg-mask) 0% 0% / auto repeat border-box border-box match-source;
+}
+```
+
+## The `mask-composite` property
+
+The {{cssxref("mask")}} shorthand includes the {{cssxref("mask-composite")}} property which, in multiple mask declarations. This property defines how the multiple masks interact with each other, or are combined, in creating the final mask effect. Each value in the comma-separated list of values determines whether the browser should `add`, `subtract`, `intersect` or `exclude` the associated mask layer to or from the mask layers below it. Similar to `mask-mode`, and the other `mask-*` properties, there is no property in the {{cssxref("background")}} shorthand that is analogous.
```
chrisdavidmills:
```suggestion
The {{cssxref("mask")}} shorthand includes the {{cssxref("mask-composite")}} property, which defines how multiple masks are combined to create the final mask effect. Each value in the comma-separated list of values determines whether the browser should `add`, `subtract`, `intersect`, or `exclude` the associated mask layer from or to the mask layers below it. Similar to `mask-mode`, and the other `mask-*` properties, there is no property in the {{cssxref("background")}} shorthand that is analogous.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+    alt="Pride flag" />
+</div>
+<div class="exclude">
+  <img
+    src="https://mdn.github.io/shared-assets/images/examples/progress-pride-flag.jpg"
+    alt="Pride flag" />
+</div>
+```
+
+The `mask-composite` property is only relevant in cases that have two or more mask layers. This reads "mask layers", not "mask images, because if `none` is included, a transparent black mask is composited!
```
chrisdavidmills:
```suggestion
The `mask-composite` property is only relevant in cases with two or more mask layers. This doesn't include `mask-image: none` values, which result in transparent black masks being composited.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+body {
+  flex-flow: row wrap;
+}
+```
+
+{{EmbedLiveSample("composite", "", "600px")}}
+
+The semi-transparent star mask is added to, subtracted from, intersected with, or excluded from the striped mask, depending on the `mask-composite` value.
+
+If we inverse the mask layers, for some values we get very different results:
```
chrisdavidmills:
```suggestion
If we reverse the order of the mask layers, we can get very different results:
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+      #f00 0 20px,
+      #f005 20px 40px,
+      transparent 40px 60px
+    );
+}
+```
+
+{{EmbedLiveSample("composite2", "", "350px")}}
+
+In the first example, the star is subtracted from the stripes. In the second, the stripes are subtracted from the star.
```
chrisdavidmills:
```suggestion
In the first example, the stars are subtracted from the stripes. In the second, the stripes are subtracted from the stars.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,805 @@
...
+
+Like all the other `mask` component properties, `mask-composite` takes a comma-separated list of values. Because the property effects how masks are combined, this property is only relevant for multiple mask layers and the number of used values is one less than the number of mask layers.
+
+The order of operations is that the last pair of masks are composited first. Then the previous mask image is composited with the product of the already composited mask images that come after it.
```
chrisdavidmills:
```suggestion
The last pair of masks is composited first. The previous mask image is then composited with the previous composition.
```

chrisdavidmills (changes requested):
@estelle Some feedback for you. There's a lot of great information here, but it seems very dense and quite repetitive in some places. As an overarching comment, I wonder if you need to include quite as much detail in this guide, because a lot of it is also said in the reference pages for these properties. You could maybe cut down on the verbiage here, for an easier-flowing guide, and link to the ref pages for the full detail.

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: kept the commas. they're important to the meaning.

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: The edit isn't accurate. Updated with example demonstrating how none impacts intersect

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: broke it into two paragraphs. Most people don't get how repeat actually works (based on the placement of the first image)

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: merged the default part into the next paragraph.

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: added:

> This example demonstrates how, when `mask-size` is set to `cover`, `contain` or a `<percentage>` value, the actual size of the mask may differ depending on the value of the `mask-origin` property:

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: created 3 paragraphs

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: ok

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: removed the first sentence

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: updated

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: whittled it down

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: that was supposed to be hidden. Updated

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: clearer now that the previous example's css is hidden

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: added the class to the example

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: keeping it this way as i assume most people will be skipping around; more people doing that than reading through the whole thing.

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: clarified and added a paragraph break.

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: it's hard to understand so I want to keep the examples

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: added an example

estelle (commented):
updated per the review, thanks!

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+    url(alphaImage.png), linear-gradient(to right, black, transparent),
+    radial-gradient(circle, white 50%, transparent 75%), none, url(#svg-mask);
+}
+```
+
+Because one of the mask images is specified as `none`, only four mask images are applied to the `.masked-element` element, while creating five mask layers.
+
+### The importance of `none`
+
+The `none` layer generally has no visual effect (see the [`mask-composite` property](#the-mask-composite-property) for how it impact the applied mask), but as each value in a comma-separated list of `mask-*` values applies to a separate mask layer, the `none` value serves an important purpose even when it doesn't change the composited mask.
```
chrisdavidmills:
```suggestion
The `none` layer generally has no visual effect (see the [`mask-composite` property](#the-mask-composite-property) for how it impacts the applied mask), but as each value in a comma-separated list of `mask-*` values applies to a separate mask layer, the `none` value serves an important purpose even when it doesn't change the composited mask.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+### Mask types: `alpha` and `luminance`
+
+Each mask is either an `alpha` or a `luminance` mask.
+
+With `alpha` masks, the alpha-transparency of each mask pixel is important. Wherever the mask is opaque, the corresponding parts of the element will be visible. Wherever the mask is transparent, the corresponding parts of the element will be hidden. Wherever the mask is semi-opaque, the element will be equally semi-opaque. The color of the mask doesn't matter, just the alpha-transparency of the colors.
+
+With `luminance` masks, both the [brightness of mask's colors](/en-US/docs/Web/CSS/CSS_masking/Masking#alpha_transparency_versus_luminance) and the alpha channel determine the opaqueness of the masked areas.
+
+> [!NOTE]
+> All subsequent examples will be using the following image as a `background-image` on an element to which masks will be applied:
```
chrisdavidmills:
```suggestion
> All subsequent examples use the following image as a `background-image` on an element to which masks will be applied:
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+.alpha img {
+  mask-mode: alpha;
+}
+
+.luminance img {
+  mask-mode: luminance;
+}
+```
+
+{{EmbedLiveSample("mode", "", "250px")}}
```
chrisdavidmills: Could do with being a little bit taller, to avoid a vertical scrollbar (I can see one on Firefox, at least)

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+
+### The `mask-mode` default value: `match-source`
+
+The default value of the `mask-mode` property is `match-source`. This value sets the `mask-mode` to match the mask's mode type. The `match-source` value resolves to `alpha` for every mask except masks where the mask source is an SVG {{svgelement("mask")}} element.
+
+In the case of the SVG `<mask>` element as the mask source, the `match-source` resolves to the value of the `<mask>` element's {{cssxref("mask-type")}} property. If the `<mask>` element (not the "masked element") doesn't have the CSS `mask-type` property defined, this property defaults to the value of the SVG {{svgAttr("mask-type")}} attribute, if present. If that is omitted as well, the `match-source` value will resolve to `luminance`.
```
chrisdavidmills:
```suggestion
When an SVG `<mask>` element is used as the mask source, the `match-source` value resolves to the value of the `<mask>` element's {{cssxref("mask-type")}} property. If the `<mask>` element (not the "masked element") doesn't have the CSS `mask-type` property defined, this property defaults to the value of the SVG {{svgAttr("mask-type")}} attribute, if present. If that is omitted as well, the `match-source` value will resolve to `luminance`.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+    none match-source,
+    url(#svg-mask) match-source;
+}
+```
+
+The first mask layer, `url(alphaImage.png)`, references an image. As this isn't a `<mask>` element within an `<svg>`, the `mask-mode` resolves to `alpha`, with the opaque parts of this image making the corresponding parts of the element visible, while the transparent or semi-transparent parts are invisible or partially visible.
+
+The `linear-gradient(to right, black, transparent)` is the second mask layer and `radial-gradient(circle, white 50%, transparent 75%)` is the third. Again, these aren't `<mask>` elements, so the `match-source` value resolves to `alpha`. The masking effect of these layers is determined by the [opaqueness of the gradient mask](/en-US/docs/Web/CSS/CSS_masking/Masking#opaqueness_versus_transparency) by default.
+
+The fourth mask layer has `none` declared, meaning this layer creates no masking effect. The `.masked-element` sets `mask-mode: match-source;`. Had this instead been a comma-separated list of values, the fourth value would have applied to this `none` layer, allowing the fifth value to apply to the fifth layer.
```
chrisdavidmills: Where you say "Had this instead been a comma-separated list of values...", what is "this"? This last sentence is really confusing. Is it needed?

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+
+If only one {{cssxref("&lt;length-percentage&gt;")}} value is specified, this specifies the X coordinate relative to the left edge of the mask origin, with the Y coordinate set to `50%`.
+
+If two keyword values are specified, the order of the value doesn't matter, but the value can not include two vertical or two horizontal axis values; `left right` and `top bottom` are both invalid.
+
+### Two-value syntax
+
+If two values are present, including one keyword and one `<length-percentage>` value, the order only matter if the keyword is `center`. If the keyword is `left` or `right`, this defines the X coordinate relative to the left edge and the value defines the Y coordinate relative to the top edge. Similarly, if `top` or `bottom` is included, this defines the Y coordinate positioning the element against the top or bottom edge respectively, with the other value defining the X value relative to the left edge of the mask origin box. When one is `center` and the other is a `<length-percentage>`, the first value defines the horizontal position and the second value the vertical.
```
chrisdavidmills:
```suggestion
If two values are present, including one keyword and one `<length-percentage>` value, the order only matters if the keyword is `center`:

- If the keyword is `left` or `right`, it defines the X coordinate relative to the left edge, and the value defines the Y coordinate relative to the top edge.
- Similarly, a `top` or `bottom` keyword defines the Y coordinate positioning the element against the top or bottom edge, respectively, with the other value defining the X value relative to the left edge of the mask origin box.
- When one value is the `center` keyword and the other is a `<length-percentage>`, the first value defines the horizontal position and the second value defines the vertical position.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+
+If two values are present and both are `<length-percentage>` values, the order again matters; the first value defines the horizontal positioning as an offset from the left edge of the mask-positioning area while the second value defines the vertical position as an offset from the top edge of the mask-positioning area.
+
+### Four-value syntax
+
+Mask positions can also be relative to corners other than the top left. The four value syntax enables offsetting the mask from any edge. The value includes two {{cssxref("length-percentage")}} offsets each preceded by the origin side for that offset. Whether you declare the horizontal or vertical pair first doesn't matter, but you must declare the origin-side keyword (`left`, `right`, `top`, `bottom`, `x-start`, `x-end`, `y-start`, `y-end`, `block-start`, `block-end`, `inline-start`, or `inline-end`) before the offset `<length-percentage>` in each pair, and the two origin-sides can't be from the same axes.
```
chrisdavidmills:
```suggestion
Mask positions can also be relative to corners other than the top left. The four-value syntax enables offsetting the mask from any corner. The value includes two {{cssxref("length-percentage")}} offsets, each preceded by the origin side for that offset. Whether you declare the horizontal or vertical pair first doesn't matter, but you must declare the origin-side keyword (`left`, `right`, `top`, `bottom`, `x-start`, `x-end`, `y-start`, `y-end`, `block-start`, `block-end`, `inline-start`, or `inline-end`) before the offset `<length-percentage>` in each pair, and the two origin-sides can't be from the same axis.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+
+In the two `<length-percentage>` syntax, the origin sides are `top` and `left`, in that order. For example, `mask-position: 10px 20px` is the equivalent of `mask-position: left 10px top 20px`. When offsetting from the top and left, the offset sides aren't required, but the order matters. With the four value syntax, you can use `mask-position` to offset the mask image from any edge combinations, such as `left 10px bottom 20px`, and the order of the sides doesn't matter, as the offset edge is defined by the keyword preceding it rather than the declaration order.
+
+When offsetting using percentage values, the mask's dimension is subtracted from the element's dimension, just as is done with [percentage offsets with `background-position`](/en-US/docs/Web/CSS/background-position#regarding_percentages).
```
chrisdavidmills, line 247: Maybe add an subheading above this paragraph, to separate out the example from the 1/2/4 value descriptions? Maybe "Placement of repeated masks", or similar?

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+
+The `mask-position` property defines the initial position of the mask image. By "initial position", if the [mask repeats](#the_mask-repeat_property), the browser places the first mask image in the position defined by the `mask-position` property, thus defining the placement of the mask repetitions.
+
+For example, for the first image, we set the position to `bottom right`, meaning the first mask will be placed at the bottom right edge of the mask origin box. Because mask images repeat by default, the repeating masks will be positioned against the first placed mask's top and left sides.
```
chrisdavidmills:
```suggestion
In this example, we set the position of the first image to `bottom right`, meaning the first mask will be placed at the bottom right edge of the mask origin box. Because mask images repeat by default, the repeating masks will be positioned against the first placed mask's top and left sides.
```

chrisdavidmills: Maybe also add a description of where you are placing the second and third images? It seems a bit weird to talk about the first one only.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+}
+```
+
+## The `mask-origin` property
+
+When an element has padding, a border, or both, the {{cssxref("mask-origin")}} property defines which of these box edge values acts as the mask origin box, or the _mask positioning area_, within which a mask image is positioned, for that layer. The `mask-origin` property is analogous to the {{cssxref("background-origin")}} property, but with a different initial value and SVG-only values.
+
+HTML elements can have masks contained within their content border box, padding box, or content box. For example, if the `mask-position` is `top left`, is that relative to the border's outer edge, the padding's outer edge, or the content's outer edge?
+
+In the [`mask-position`](#the_mask-position_property) masking example, the position defined was relative to the content-box, padding-box, and border-box, as the `<img>` had no border or padding set, so these three [`<coord-box>`](/en-US/docs/Web/CSS/box-edge#values) values were all same (the `<div>` container did have a `1px` border).
```
chrisdavidmills:
```suggestion
In the [`mask-position`](#the_mask-position_property) masking example, the position defined was relative to the border-box (the default behavior), although it is worth noting that the `<img>` had no border or padding set, therefore the content-box, padding-box, and border-box origins would all be the same in this case.
```

Would something like this be better? It still sounds really confusing and incorrect to say that the position defined is relative to all three origins simultaneously. Even though I kind of understand what you mean now.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+}
+:has(#origin_padding-box:checked) img {
+  mask-origin: padding-box;
+}
+:has(#origin_content-box:checked) img {
+  mask-origin: content-box;
+}
+```
+
+Change the value of the `mask-origin` property by changing the selected radio button, looking at the top left point of the mask top left star as you do so.
```
chrisdavidmills:
```suggestion
Change the value of the `mask-origin` property by changing the selected radio button, looking at the position of the top left star mask as you do so.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+}
+```
+
+For SVG elements, which don't have the associated CSS layout boxes, a mask can be contained inside the SVG element's fill, stroke, or view box.
+
+## The `mask-clip` property
+
+The {{cssxref("mask-clip")}} property determines the area of the element that will be affected by a mask, effectively clipping the element at the defined box edge. It is analogous to the {{cssxref("background-clip")}} property, but with some different values.
+
+Because the `mask-clip` property accepts all the `mask-origin` values and both have the same `border-box` default value, the two properties may seem similar, but they serve very different purposes. While `mask-origin` determines where a mask image will be positioned, the `mask-clip` property causes the original element to have its content clipped to the specified box. It is important to understand them both, as if the mask-origin causes the mask-position to place the mask image outside the clipping area, the mask will be clipped.
```
chrisdavidmills:
```suggestion
Because the `mask-clip` property accepts all the `mask-origin` values and both have the same `border-box` default value, the two properties may seem similar, but they serve very different purposes. While `mask-origin` determines where a mask image will be positioned, the `mask-clip` property causes the original element to have its content clipped to the specified box. It is important to understand them both: if the `mask-origin` causes the `mask-position` to place the mask image outside the clipping area, the mask will be clipped.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+
+The `mask-clip` property accepts all the `mask-origin` values, as well as its own `no-clip` value. The `no-clip` value sets the painted content not to be clipped. You can still cause the mask image to be clipped by positioning it outside of the border content area using `mask-position` values that are less than zero or resolve to greater than 100%.
+
+Setting the `mask-clip` and `mask-origin` to different values can cause the mask layer image to be clipped. For example, If an element's has a border and padding, if `mask-clip` is set to `content-box` while `mask-origin` is set to `border-box`, and the `mask-position` is set to the `top left` edge, the mask layer image will be clipped at the top-left edge.
```
chrisdavidmills:
```suggestion
Setting the `mask-clip` and `mask-origin` to different values can cause the mask layer image to be clipped. For example, if an element with a border and padding has `mask-clip` set to `content-box` and `mask-origin` set to `border-box`, and the `mask-position` is set to the `top left` edge, the mask layer image will be clipped at the top-left edge.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+
+The next example adds clipping options to the previous example to demonstrate the different non-SVG `mask-clip` values and to show how they impact the different `mask-origin` values.
```
chrisdavidmills, line 459: In this example, `mask-clip` values of `border-box` and `no-clip` always result in the same effect, for all `mask-origin` values.

Is there a case where `mask-clip` values of `border-box` and `no-clip` will result in different effects? It feels to me like this should be talked about a bit.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+
+The mask image can be left to its natural size, stretched, or constrained to fit the available space. The {{glossary("aspect ratio")}} of the mask image is maintained by default, but declaring two `<length-percentage>` values can distort the mask image if the ratio of the two values is not the same as the original image (`mask-repeat: round` is the other property/value pair that may distort the mask image).
+
+If the `mask-size` is set to `contain`, the mask image will be the largest size it can be while being fully contained within the mask positioning area; the mask image is resized until one dimension is the same size as the corresponding mask positioning area's dimension without overflowing the area.
+
+When set to `cover`, the mask image is resized until the second dimension is the same size as the corresponding mask positioning area's dimension, overflowing the other dimension if the aspect ratio of the element and mask aren't the same. With `cover`, the mask image will be the smallest size it can be to fully cover the entire mask-positioning area, with the mask being clipped if the mask's aspect ratio differs from the aspect ratio of the mask-positioning area.
```
chrisdavidmills:
```suggestion
If the `mask-size` is set to `contain`, the mask image will be the largest size it can be while being fully contained within the mask positioning area. In this case, the mask image is not clipped, but rather, it is fully contained.

When set to `cover`, the mask image will be the smallest size it can be to fully cover the entire mask-positioning area, with the mask being clipped if the mask's aspect ratio differs from the aspect ratio of the mask-positioning area.
```

chrisdavidmills: This makes sense to me. The bits I've suggested removing just seemed waffly and confusing. Maybe also add in a diagram to illustrate the values visually?

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+    radial-gradient(circle, white 50%, transparent 75%) 0% 0% / auto border-box
+      border-box match-source,
+    none 0% 0% / auto border-box border-box match-source,
+    url(#svg-mask) 0% 0% / auto border-box border-box match-source;
+}
+```
+
+## The `mask-repeat` property
+
+The {{cssxref("mask-repeat")}} property defines how mask images are repeated, or tiled, after the initial mask image has been sized and positioned. The `mask-repeat` property defines if and how that mask image is repeated along the horizontal axis, the vertical axis, both axes. In most of the the previous examples, you may have noticed the star mask repeated along the X and Y axis. This is because `repeat` is the default.
```
chrisdavidmills:
```suggestion
The {{cssxref("mask-repeat")}} property defines how mask images are repeated, or tiled, after the initial mask image has been sized and positioned. The `mask-repeat` property defines if and how that mask image is repeated along the horizontal and vertical axes. In most of the previous examples, you may have noticed the star mask repeated along the X and Y axes. This is because `repeat` is the default value.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,869 @@
...
+body {
+  flex-flow: row wrap;
+}
+```
+
+{{EmbedLiveSample("composite", "", "600px")}}
+
+The semi-transparent star mask is added to, subtracted from, intersected with, or excluded from the striped mask, depending on the `mask-composite` value.
+
+The `mask-composite` property is only relevant in cases that have two or more mask layers. This reads "mask layers", not "mask images, because if `none` is included, the transparent black mask is composited. A `none` can have a profound effect of masking in the case of `subtract` and `intersect`. For example, if the `mask-mode` resolves to `luminance`, subtracting a black mask will remove the entire mask (the element will be hidden). Similarly, if the `none` is the last layer with `mask-composite: intersect` set for that layer, the entire element will be hidden. Here, we add a third layer, with `none`, to the previous example:
```
chrisdavidmills:
```suggestion
The `mask-composite` property is only relevant in cases with two or more mask layers. This reads "mask layers", not "mask images" because, if `none` is included, the transparent black mask is composited. A `none` value can have a profound effect on masking in the case of `subtract` and `intersect`. For example, if the `mask-mode` resolves to `luminance`, subtracting a black mask will remove the entire mask (the element will be hidden). Similarly, if `none` is the last layer with `mask-composite: intersect` set for that layer, the entire element will be hidden. Here, we add a third layer, with `none`, to the previous example:
```

chrisdavidmills (changes requested):
@estelle this is still pretty heavy going, but it is much better now. I've got a few more comments for you to consider.

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,873 @@
...
+.alpha img {
+  mask-mode: alpha;
+}
+
+.luminance img {
+  mask-mode: luminance;
+}
+```
+
+{{EmbedLiveSample("mode", "", "250px")}}
```
estelle:
```suggestion
{{EmbedLiveSample("mode", "", "270px")}}
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,873 @@
...
+    none match-source,
+    url(#svg-mask) match-source;
+}
+```
+
+The first mask layer, `url(alphaImage.png)`, references an image. As this isn't a `<mask>` element within an `<svg>`, the `mask-mode` resolves to `alpha`, with the opaque parts of this image making the corresponding parts of the element visible, while the transparent or semi-transparent parts are invisible or partially visible.
+
+The `linear-gradient(to right, black, transparent)` is the second mask layer and `radial-gradient(circle, white 50%, transparent 75%)` is the third. Again, these aren't `<mask>` elements, so the `match-source` value resolves to `alpha`. The masking effect of these layers is determined by the [opaqueness of the gradient mask](/en-US/docs/Web/CSS/CSS_masking/Masking#opaqueness_versus_transparency) by default.
+
+The fourth mask layer has `none` declared, meaning this layer creates no masking effect. The `.masked-element` sets `mask-mode: match-source;`. Had this instead been a comma-separated list of values, the fourth value would have applied to this `none` layer, allowing the fifth value to apply to the fifth layer.
```
estelle:
```suggestion
The fourth mask layer has `none` declared, meaning the mask for this layer is a transparent black. The `.masked-element` sets `mask-mode: match-source;`. Had the `masked-mode` instead been a comma-separated list of five different values, the fourth value would have applied to this `none` layer, allowing the fifth value to apply to the fifth layer.
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,873 @@
...
+
+In the two `<length-percentage>` syntax, the origin sides are `top` and `left`, in that order. For example, `mask-position: 10px 20px` is the equivalent of `mask-position: left 10px top 20px`. When offsetting from the top and left, the offset sides aren't required, but the order matters. With the four value syntax, you can use `mask-position` to offset the mask image from any edge combinations, such as `left 10px bottom 20px`, and the order of the sides doesn't matter, as the offset edge is defined by the keyword preceding it rather than the declaration order.
+
+When offsetting using percentage values, the mask's dimension is subtracted from the element's dimension, just as is done with [percentage offsets with `background-position`](/en-US/docs/Web/CSS/background-position#regarding_percentages).
```
estelle, line 247:
```suggestion
### Percentage values

When offsetting using percentage values, the mask's dimension is subtracted from the element's dimension, just as is done with [percentage offsets with `background-position`](/en-US/docs/Web/CSS/background-position#regarding_percentages).
```

files/en-us/web/css/css_masking/mask_properties/index.md
```diff
@@ -0,0 +1,875 @@
...
+
+In the two `<length-percentage>` syntax, the origin sides are `top` and `left`, in that order. For example, `mask-position: 10px 20px` is the equivalent of `mask-position: left 10px top 20px`. When offsetting from the top and left, the offset sides aren't required, but the order matters. With the four value syntax, you can use `mask-position` to offset the mask image from any edge combinations, such as `left 10px bottom 20px`, and the order of the sides doesn't matter, as the offset edge is defined by the keyword preceding it rather than the declaration order.
+
+### Percentage values
+
+When offsetting using percentage values, the mask's dimension is subtracted from the element's dimension, just as is done with [percentage offsets with `background-position`](/en-US/docs/Web/CSS/background-position#regarding_percentages).
+
+The `mask-position` property defines the initial position of the mask image. By "initial position", if the [mask repeats](#the_mask-repeat_property), the browser places the first mask image in the position defined by the `mask-position` property, thus defining the placement of the mask repetitions.
```
estelle, line 251:
```suggestion
### Positioning repeating mask images

The `mask-position` property defines the initial position of the mask image. By "initial position", if the [mask repeats](#the_mask-repeat_property), the browser places the first mask image in the position defined by the `mask-position` property, thus defining the placement of the mask repetitions.
```

files/en-us/web/css/css_masking/mask_properties/index.md
estelle: The `border-box` value doesn't clip in this case. The main question is does the `border-box` value ever result in clipping. I haven't found an example of it clipping, but the spec says it's clipped to the border-box, which is the outer edge. so i can't think of an example, and worndered the same as i wrote this, but don't have an example to give and don't want to get into the weeds and be wrong.

chrisdavidmills (approved):
Awesome, let's get this merged. Nice job, @estelle.

PR #39964, opened 2025-06-17
Add information about SVG transform support on <svg> root

chrisdavidmills:
### Description

Chrome 137+ supports setting an SVG `transform` attribute on the root `<svg>` element. See https://chromestatus.com/feature/6129368025530368 for details.

This PR updates the `transform` attribute page to mention this, and to clean up the structure a bit.

### DIFF
--- a/files/en-us/web/svg/reference/attribute/transform/index.md
+++ b/files/en-us/web/svg/reference/attribute/transform/index.md
@@ -17,7 +17,6 @@ In SVG 2, you can use the `transform` attribute on any element, including the {{
 Note that using `transform` on the `<svg>` root is a newer feature, and you should check [browser compatibility](#browser_compatibility) for support.
 Using `transform` on the `<svg>` root is convenient for applying transforms to an entire SVG image without the need for extra wrapper elements or CSS workarounds.
 
-
 In SVG 1.1, only these 16 elements were allowed to have a `transform` applied: {{SVGElement('a')}}, {{SVGElement('circle')}}, {{SVGElement('clipPath')}}, {{SVGElement('defs')}}, {{SVGElement('ellipse')}}, {{SVGElement('foreignObject')}}, {{SVGElement('g')}}, {{SVGElement('image')}}, {{SVGElement('line')}}, {{SVGElement('path')}}, {{SVGElement('polygon')}}, {{SVGElement('polyline')}}, {{SVGElement('rect')}}, {{SVGElement('switch')}}, {{SVGElement('text')}}, and {{SVGElement('use')}}.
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/svg/reference/attribute/transform/index.md
```diff
@@ -11,7 +11,38 @@ The **`transform`** attribute defines a list of transform definitions that are a
 > [!NOTE]
 > As a presentation attribute, `transform` also has a CSS property counterpart: {{cssxref("transform")}}. When both are specified, the CSS property takes priority. However, be aware that there are some differences in syntax between the CSS property and the attribute. See the documentation for the CSS property {{cssxref('transform')}} for the specific syntax to use in that case.
```
hamishwillee: I think this is a little redundant.

```suggestion
> As a presentation attribute, `transform` also has a CSS property counterpart: {{cssxref("transform")}}. When both are specified, the CSS property takes priority. Note that there are some differences in syntax between the CSS property and the attribute!
```

files/en-us/web/svg/reference/attribute/transform/index.md
```diff
@@ -11,7 +11,38 @@ The **`transform`** attribute defines a list of transform definitions that are a
...
 
-You can use this attribute with any SVG element.
+## Elements
+
+In SVG 2, you can use the `transform` attribute on any SVG element inside the {{SVGElement('svg')}} root. In addition, some browsers support using `transform` on the `<svg>` root itself, which is convenient for applying transforms to an entire SVG image without the need for extra wrapper elements or CSS workarounds.
```
hamishwillee: I'm worried this won't get updated when supported by all browser. Which is annoying, because I like the way you have worded this. 

We document the spec, so perhaps you might consider this. 

```suggestion
In SVG 2, you can use the `transform` attribute on any element, including the {{SVGElement('svg')}} root.
Note that using `transform` on the `<svg>` root is a newer feature, and you should check [browser compatibility](#browser_compatibility) for support.
Using `transform` on the `<svg>` root is convenient for applying transforms to an entire SVG image without the need for extra wrapper elements or CSS workarounds.

```

hamishwillee (approved):
@chrisdavidmills Definitely an improvement. One nit https://github.com/mdn/content/pull/39964/files#r2157795404

As this is an update the only requirement for approval is that it is an improvement, which it is. But if you could find it in your heart to add an example of SVG transform in the root, I think it would be "a very good thing".

chrisdavidmills:
> @chrisdavidmills Definitely an improvement. One nit https://github.com/mdn/content/pull/39964/files#r2157795404
> 
> As this is an update the only requirement for approval is that it is an improvement, which it is. But if you could find it in your heart to add an example of SVG transform in the root, I think it would be "a very good thing".

That is a good idea. Done!

chrisdavidmills:
> @chrisdavidmills Definitely an improvement. One nit https://github.com/mdn/content/pull/39964/files#r2157795404
> 
> As this is an update the only requirement for approval is that it is an improvement, which it is. But if you could find it in your heart to add an example of SVG transform in the root, I think it would be "a very good thing".

That is a good idea. Done!

PR #39971, opened 2025-06-17
Fix uncaught ReferenceError in example

aabounegm:
### Description

The `event` parameter was previously not accepted by the event handler, but used in its body, leading to a reference error.

### Motivation

For the interactive playground to work

### Additional details

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey


bsmth (approved):
Nice catch, thank you very much! Well done on your first contribution 🎉

PR #39980, opened 2025-06-18
Editorial review: Add docs for MediaTrackSettings.screenPixelRatio

chrisdavidmills:
### Description

Chrome 136+ supports the `MediaTrackSettings.screenPixelRatio` property, on desktop only. See https://chromestatus.com/feature/6738236535472128 for details.

This PR adds a ref page for the new property.

### DIFF
--- a/files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
+++ b/files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
@@ -8,33 +8,40 @@ browser-compat: api.MediaTrackSettings.screenPixelRatio
 
 {{APIRef("Media Capture and Streams")}}
 
-The {{domxref("MediaTrackSettings")}} dictionary's **`screenPixelRatio`** property is a number representing the ratio of the size of a CSS pixel on the capturing screen by the vertical size of a pixel on the captured display surface. It cannot be used as a constraint or capability.
+The {{domxref("MediaTrackSettings")}} dictionary's **`screenPixelRatio`** property is a number representing the ratio of the physical size of a pixel on the captured display surface (displayed at its physical resolution) to the logical size of a CSS pixel on the capturing screen (displayed at its logical resolution). It cannot be used as a constraint or capability.
 
 This property allows applications using the [Screen Capture API](/en-US/docs/Web/API/Screen_Capture_API) to save resources by sending the video of a screen capture at its logical, or device independent, resolution.
 
 ## Value
 
-A number representing the screeen pixel ratio.
+A number representing the screen pixel ratio.
 
-This is calculated by dividing the size of a {{glossary("CSS pixel")}} at a page zoom of `1.0` and using a scale factor of `1.0` on the capturing screen by the vertical size of a pixel from the [display surface](/en-US/docs/Web/API/MediaTrackConstraints/displaySurface).
+This is calculated by dividing the size of a {{glossary("CSS pixel")}} at a page zoom of `1.0` and using a scale factor of `1.0` on the capturing screen by the vertical size of a pixel from the captured [display surface](/en-US/docs/Web/API/MediaTrackConstraints/displaySurface).
 
 ## Description
 
-It is common for a screen to have zoom applied via the operating system (OS), for example by the user zooming in. The resolution before applying the zoom is called the **logical resolution**, and the resolution after the zoom is applied is called the **physical resolution**.
+It is common for a screen to have zoom applied via the operating system (OS), for example when the display is a high-resolution display, and you want the graphics to be shown at the same physical size as they would on a standard resolution display. The resolution before applying the zoom is called the **logical resolution**, and the resolution after the zoom is applied is called the **physical resolution**.
 
-A video-conferencing app can save bandwidth and CPU when the captured screen is zoomed in on the sender side by:
+If the sender's captured screen is zoomed in, then the physical resolution is greater than the logical resolution, and a video-conferencing app can therefore save bandwidth and CPU by:
 
 1. Removing the zoom applied to the captured display surface by the OS.
 2. Sending the video of the screen capture at the logical resolution.
 3. Reapplying the zoom after receiving it on the remote client to size it back up to its physical resolution.
 
-The `screenPixelRatio` property describes the ratio of the physical size of a pixel to its logical size, and therefore enables the application to work out how much of a zoom has been applied, and then constrain the video to the logical size.
+The `screenPixelRatio` property describes the ratio of the physical size of a pixel to the logical size of a CSS pixel, and therefore enables the application to work out how much of a zoom factor has been applied, and then constrain the video to the logical size.
+
+For example:
+
+- If the captured display surface is being displayed on a standard resolution screen where physical pixel dimensions are about the same as CSS pixel dimensions, `screenPixelRatio` will return a value of `1`.
+- If, however, the captured display surface is being displayed on a high-dpi resolution screen where physical pixel dimensions are about double the CSS pixel dimensions, `screenPixelRatio` will return a value of `2`.
 
 ## Examples
 
 ### Basic `screenPixelRatio` usage
 
-The following snippet shows how an application may use `screenPixelRatio` to constrain a {{domxref("MediaStreamTrack")}}'s' [`width`](/en-US/docs/Web/API/MediaTrackConstraints/width) and [`height`](/en-US/docs/Web/API/MediaTrackConstraints/height) to the logical resolution when scaling exceeds an acceptable threshold (specified by an application-defined constant).
+In this example, the application defines a constant `RESOLUTION_LIMIT`, which represents the scaling factor beyond which which the sending application should send video at the logical resolution rather than the physical resolution.
+
+When `screenPixelRatio` exceeds this limit, the application uses the `screenPixelRatio` value to calculate the logical resolution from the physical resolution, and then constrains the captured {{domxref("MediaStreamTrack")}} to the logical resolution.
 
 ```js
 const RESOLUTION_LIMIT = 1.5;
@@ -47,11 +54,11 @@ async function startCapture() {
   const settings = track.getSettings();
   const capabilities = track.getCapabilities();
 
-  if (track.screenPixelRatio > RESOLUTION_LIMIT) {
-    let physicalWidth = capabilities.width.max;
-    let physicalHeight = capabilities.height.max;
-    let logicalWidth = physicalWidth / settings.screenPixelRatio;
-    let logicalHeight = physicalHeight / settings.screenPixelRatio;
+  if (settings.screenPixelRatio > RESOLUTION_LIMIT) {
+    const physicalWidth = capabilities.width.max;
+    const physicalHeight = capabilities.height.max;
+    const logicalWidth = physicalWidth / settings.screenPixelRatio;
+    const logicalHeight = physicalHeight / settings.screenPixelRatio;
     await track.applyConstraints({
       width: logicalWidth,
       height: logicalHeight,
--- a/files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
+++ b/files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
@@ -1,0 +1,76 @@
+---
+title: "MediaTrackSettings: screenPixelRatio property"
+short-title: screenPixelRatio
+slug: Web/API/MediaTrackSettings/screenPixelRatio
+page-type: web-api-instance-property
+browser-compat: api.MediaTrackSettings.screenPixelRatio
+---
+
+{{APIRef("Media Capture and Streams")}}
+
+The {{domxref("MediaTrackSettings")}} dictionary's **`screenPixelRatio`** property is a double-precision floating point value indicating the ratio of the size of a CSS pixel on the capturing screen by the vertical size of a pixel on the captured display surface. It cannot be used as a constraint or capability.
+
+This property allows applications using the [Screen Capture API](/en-US/docs/Web/API/Screen_Capture_API) to save resources by sending the video of a screen capture using their logical, or device independent, resolution.
+
+## Value
+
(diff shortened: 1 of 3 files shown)


guidou:
lgtm % link to the spec

guidou:
spec link:  https://w3c.github.io/mediacapture-screen-share/#dfn-screenpixelratio

chrisdavidmills:
> spec link: https://w3c.github.io/mediacapture-screen-share/#dfn-screenpixelratio

I added spec/browser compat data for this feature in https://github.com/mdn/browser-compat-data/pull/27092, so that data should appear as soon as we resolve what to do with the issue highlighted in the description. 

I think in the meantime this PR can be put forward to the editorial review stage. Thanks for the tech review, @guidou!

wbamberg (commented):
I found this page quite hard to understand.

There's a very technical description of the property, repeated word for word, twice:

> a double-precision floating point value indicating the ratio of the size of a CSS pixel on the capturing screen by the vertical size of a pixel on the captured display surface

The "value" has an additional sentence which is a direct quote from the spec.

The description seems to be taken from the chromestatus entry, and I kind of understand it, on its own, but there's nothing there that connects the general explanation with how exactly I could use this property to meet the use case described.

It also has hedging terms like "The logical resolution is **normally** lower than the physical resolution". What is "normally" doing here? When is it, when isn't it? And how is this reflected by the screenPixelRatio property? That is, how can examining the value of screenPixelRatio tell me whether it is or not? The definition of this property doesn't mention physical resolution at all.

My understanding of this property is something like:

> If a user has the display zoomed in, then the physical resolution is greater than the logical resolution. In a case like this, it's a more efficient use of bandwidth for the sender to capture the video at the logical resolution and send the video at that resolution, and for the recipient to then reapply the zoom when they receive the video stream.
>
> The `screenPixelRatio` property describes the ratio of the physical size of a pixel to its logical size, and therefore enables the application to work out how much of a zoom has been applied, and then constrain the video to the logical size.

Is that about right? If so I think it's important to say this. It's important to connect the technical definition of the property with the use case given in the description.

chrisdavidmills:
> I found this page quite hard to understand.

@wbamberg thanks for your thoughtful comment here. I agree that the wording is a bit complicated. I've pushed another commit that simplifies and de-duplicates, and also takes your wording into account.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
```diff
@@ -0,0 +1,76 @@
...
+title: "MediaTrackSettings: screenPixelRatio property"
+short-title: screenPixelRatio
+slug: Web/API/MediaTrackSettings/screenPixelRatio
+page-type: web-api-instance-property
+browser-compat: api.MediaTrackSettings.screenPixelRatio
+---
+
+{{APIRef("Media Capture and Streams")}}
+
+The {{domxref("MediaTrackSettings")}} dictionary's **`screenPixelRatio`** property is a number representing the ratio of the size of a CSS pixel on the capturing screen by the vertical size of a pixel on the captured display surface. It cannot be used as a constraint or capability.
```
wbamberg: This first sentence is quite complex, but AIUI "by" is the word used to connect the two values being compared in the ratio, right? Why "by"? Usually I'd expect the connector in a ratio to be "to". And AIUI the value is positive if the physical size is greater than the logical size, so wouldn't that make it "the ratio of [physical size] to [logical size]", i.e. the physical would come first?

See e.g. https://en.wikipedia.org/wiki/Ratio:

> if there are eight oranges and six lemons in a bowl of fruit, then the ratio of oranges to lemons is eight to six (that is, 8:6, which is equivalent to the ratio 4:3). Similarly, the ratio of lemons to oranges is 6:8 (or 3:4) and the ratio of oranges to the total amount of fruit is 8:14 (or 4:7). 

I also think it would be helpful to clarify the connection between "CSS pixel" and "logical size", and "vertical size of a pixel" and "physical size". This intro and the value talks only about the first set of terms, and the description talks only about the second set. I assume they are the same(?) but as a reader I should not have to make that assumption.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+A number representing the screeen pixel ratio.
+
+This is calculated by dividing the size of a {{glossary("CSS pixel")}} at a page zoom of `1.0` and using a scale factor of `1.0` on the capturing screen by the vertical size of a pixel from the [display surface](/en-US/docs/Web/API/MediaTrackConstraints/displaySurface).
+
+## Description
+
+It is common for a screen to have zoom applied via the operating system (OS), for example by the user zooming in. The resolution before applying the zoom is called the **logical resolution**, and the resolution after the zoom is applied is called the **physical resolution**.
+
+A video-conferencing app can save bandwidth and CPU when the captured screen is zoomed in on the sender side by:
```
wbamberg:
```suggestion

If the screen in the sender is zoomed in, then the physical resolution is greater than the logical resolution, and a video-conferencing app can therefore save bandwidth and CPU by:
```

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. Removing the zoom applied to the captured display surface by the OS.
+2. Sending the video of the screen capture at the logical resolution.
+3. Reapplying the zoom after receiving it on the remote client to size it back up to its physical resolution.
```
wbamberg, line 29: Sorry if this is a stupid q, but how does the receiver side know whether/how much to reapply the zoom?

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
wbamberg: Is the underlying issue here only about zoom levels? If so, it would be nice if this concept were introduced here, and the connection made between zoom levels and different values for this property.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+The `screenPixelRatio` property describes the ratio of the physical size of a pixel to its logical size, and therefore enables the application to work out how much of a zoom has been applied, and then constrain the video to the logical size.
+
+## Examples
+
+### Basic `screenPixelRatio` usage
+
+The following snippet shows how an application may use `screenPixelRatio` to constrain a {{domxref("MediaStreamTrack")}}'s' [`width`](/en-US/docs/Web/API/MediaTrackConstraints/width) and [`height`](/en-US/docs/Web/API/MediaTrackConstraints/height) to the logical resolution when scaling exceeds an acceptable threshold (specified by an application-defined constant).
```
wbamberg: This could also use some more explanation IMO.

We could say something like:

> In this example, the application defines a constant `RESOLUTION_LIMIT`, which represents the scaling factor beyond which which the sending application should send video at the logical resolution rather than the physical resolution.
>
> When `screenPixelRatio` exceeds this limit, the application uses the `screenPixelRatio` value to calculate the logical resolution from the physical resolution, and then constrains video capture to the logical resolution.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+async function startCapture() {
+  const stream = await navigator.mediaDevices.getDisplayMedia({
+    video: true,
+  });
+  const track = stream.getVideoTracks()[0];
+  const settings = track.getSettings();
+  const capabilities = track.getCapabilities();
+
+  if (track.screenPixelRatio > RESOLUTION_LIMIT) {
```
wbamberg:
```suggestion
  if (settings.screenPixelRatio > RESOLUTION_LIMIT) {
```

I think?

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
```diff
@@ -0,0 +1,76 @@
...
+    let physicalWidth = capabilities.width.max;
+    let physicalHeight = capabilities.height.max;
+    let logicalWidth = physicalWidth / settings.screenPixelRatio;
+    let logicalHeight = physicalHeight / settings.screenPixelRatio;
```
wbamberg: Can all be `const`.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+{{APIRef("Media Capture and Streams")}}
+
+The {{domxref("MediaTrackSettings")}} dictionary's **`screenPixelRatio`** property is a number representing the ratio of the size of a CSS pixel on the capturing screen by the vertical size of a pixel on the captured display surface. It cannot be used as a constraint or capability.
+
+This property allows applications using the [Screen Capture API](/en-US/docs/Web/API/Screen_Capture_API) to save resources by sending the video of a screen capture at its logical, or device independent, resolution.
+
+## Value
+
+A number representing the screeen pixel ratio.
```
wbamberg:
```suggestion
A number representing the screen pixel ratio.
```

(but also this doesn't really tell me anything. It's called `screenPixelRatio`, so I can guess this it is the "screen pixel ratio". But it doesn't tell me what that means.)

files/en-us/web/api/screen_capture_api/index.md
```diff
@@ -77,6 +77,8 @@ The Screen Capture API adds properties to the following dictionaries defined by
   - : A boolean value, which is `true` if the video being captured doesn't directly correspond to a single onscreen display area.
 - {{domxref("MediaTrackSettings.suppressLocalAudioPlayback")}}
   - : A boolean value, which is `true` if the audio being captured is not played out of the user's local speakers.
+- {{domxref("MediaTrackSettings.screenPixelRatio")}}
+  - : A double-precision floating point value indicating the ratio of the size of a CSS pixel on the capturing screen by the vertical size of a pixel on the captured display surface. It cannot be used as a constraint or capability.
```
wbamberg: In the page you've replaced "double-precision floating point value" with "number" (which I think is correct) but this isn't updated. Also I guess any other change to the page intro ought to be reflected here.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
chrisdavidmills: Right, OK, you've made some good points here. I've adjusted the wording as suggested, and made the terminology more consistent.

I've also tested it slightly more thoroughly, adjusted the description of why the difference in ratio occurs, and added some example value explanations.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
chrisdavidmills: used, but tweaked a little bit.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
chrisdavidmills: You'd send the `screenPixelRatio` value over with the captured stream, and then zoom the video by that amount on the receiving end.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
chrisdavidmills: Sure, added, but tweaked.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
chrisdavidmills: Yup, good spot. Updated.

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
chrisdavidmills: Good call; updated

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
chrisdavidmills: Typo fixed.

I don't really know what else you want me to write here. The rest of the page, in general, explains what this means.

Should I update it to just "A number"?

files/en-us/web/api/screen_capture_api/index.md
chrisdavidmills: Darnit, I forgot about this one. Updated!

files/en-us/web/api/mediatracksettings/screenpixelratio/index.md
wbamberg: Fair enough, it's fine.

wbamberg (approved):
Thanks for your work on this Chris, I'm approving.

But, what do you want to do about BCD? At the moment the spec and BCD sections look pretty sad.

I read https://github.com/mdn/browser-compat-data/pull/27092 and https://github.com/mdn/browser-compat-data/issues/27119.

It is tricky because yes, BCD doesn't represent dictionaries, and this makes sense really because there's no compat story about "just data" - data doesn't do anything, and when nothing is done, there's no compat story.

What I would do here is:
- not include BCD for this page
- add a subfeature to [`getSettings`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/getSettings) that captures the fact that the returned object includes `screenPixelRatio`, and tie the BCD to that. I don't know if there's a convention for adding subfeatures like "the returned object has property `foo` set".

chrisdavidmills:
Thanks a lot @wbamberg, these are good suggestions

> What I would do here is:
> 
>     * not include BCD for this page

I've removed the BCD section for now, and added a direct `spec-urls` entry to the frontmatter so that the spec table shows up.

> 
>     * add a subfeature to [`getSettings`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/getSettings) that captures the fact that the returned object includes `screenPixelRatio`, and tie the BCD to that. I don't know if there's a convention for adding subfeatures like "the returned object has property `foo` set".

I'll investigate doing this as my next port of call, and see what the BCD folk say about it.

PR #39982, opened 2025-06-18
Editorial review: Document the Captured Surface Control API

chrisdavidmills:
### Description

Chrome 136+ supports the [Captured Surface Control](https://w3c.github.io/mediacapture-surface-control/) API, another extension to the [Screen Capture API](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Capture_API). See https://chromestatus.com/feature/5092615678066688 for details of the addition.

This PR documents the new features, providing reference docs and a guide.


files/en-us/web/api/capturecontroller/forwardwheel/index.md
```diff
@@ -0,0 +1,85 @@
...
+## Syntax
+
+```js-nolint
+forwardWheel(element)
+```
+
+### Parameters
+
+- `element`
+  - : A reference to the element whose captured `wheel` events you want to forward to the associated captured display surface.
```
eladalon1983: The `wheel` events of that element are not the thing that is "captured". Rather, the wheel events are forwarded to the "captured" surface. (Same comment throughout.)

files/en-us/web/api/capturecontroller/forwardwheel/index.md
```diff
@@ -0,0 +1,85 @@
...
+
+A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
+
+### Exceptions
+
+- `InvalidStateError` {{domxref("DOMException")}}
+  - : Thrown when:
+    - The capturing {{domxref("MediaStream")}} returned by the originating {{domxref("MediaDevices.getDisplayMedia()")}} call is no longer capturing, for example because the associated {{domxref("MediaStreamTrack")}} objects have had {{domxref("MediaStreamTrack.stop", "stop()")}} called on them.
+    - The application is capturing itself.
+    - an attempt was made to invoke `forwardWheel()` without transient activation.
```
eladalon1983: It is permissible to call `forwardWheel()` without transient activation, **if** the relevant permission is already `"granted"`.

files/en-us/web/api/capturecontroller/forwardwheel/index.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Examples
+
+### Basic `forwardWheel()` usage
+
+In our live demo, explained in [Using the Captured Surface Control API](/en-US/docs/Web/API/Screen_Capture_API/Captured_Surface_Control), we call a function called `startForwarding()` after the capturing `getDisplayMedia()` promise fulfills:
+
+```js
+videoElem.srcObject =
+  await navigator.mediaDevices.getDisplayMedia(displayMediaOptions);
```
eladalon1983: You can omit `displayMediaOptions`.

files/en-us/web/api/capturecontroller/getsupportedzoomlevels/index.md
```diff
@@ -0,0 +1,84 @@
...
+An array of numbers representing the different zoom percentages that the captured display surface supports.
+
+### Exceptions
+
+- `InvalidStateError` {{domxref("DOMException")}}
+  - : The capturing {{domxref("MediaStream")}} returned by the originating {{domxref("MediaDevices.getDisplayMedia()")}} call is no longer capturing, for example because the associated {{domxref("MediaStreamTrack")}} objects have had {{domxref("MediaStreamTrack.stop", "stop()")}} called on them.
+
+## Examples
+
+### Basic `getSupportedZoomLevels()` usage
```
eladalon1983, line 38: The section title says `getSupportedZoomLevels()`, but the first API surface discussed is actually the event handler for `"zoomlevelchange"`.

files/en-us/web/api/capturecontroller/getsupportedzoomlevels/index.md
```diff
@@ -0,0 +1,84 @@
...
+function updateZoomButtonState() {
+  decBtn.disabled = false;
+  incBtn.disabled = false;
+  if (controller.zoomLevel === zoomLevels[0]) {
+    decBtn.disabled = true;
+  } else if (controller.zoomLevel === zoomLevels[zoomLevels.length - 1]) {
+    incBtn.disabled = true;
+  }
+}
+```
```
eladalon1983, line 66: You'd might want to mention that `decreaseZoomLevel()` and `increaseZoomLevel()` should still be called from a try-block, because the zoom level could be changed asynchronously by an entity other than the application; for example, by the user's direct interaction with the user agent,

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
```diff
@@ -0,0 +1,62 @@
...
+### Exceptions
+
+- `NotAllowedError` {{domxref("DOMException")}}
+  - : The operation is disallowed by a {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy).
+
+## Examples
+
+### Basic `resetZoomLevel()` usage
+
+The following snippet adds an event listener to a button so that when it is clicked, the `resetZoom()` function is called. This in turn calls the `resetZoomLevel()` method, resetting the captured surface's zoom level.
```
eladalon1983: You probably want to explain what "resetting" means in this context. Namely, that it sets the zoom level to 100.

files/en-us/web/api/capturecontroller/zoomlevel/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+A number representing the current zoom percentage of the captured display surface.
+
+## Examples
+
+### Basic `zoomLevel` usage
+
+When a captured display surface's zoom level changes, a {{domxref("CaptureController.zoomlevelchange_event", "zoomlevelchange")}} event fires on the controller, which can be used to run an event handler such as the following. This writes the updated zoom percentage to an output element of some kind.
+
+```js
```
eladalon1983, line 25: This section is quite similar to another snippet above which also used the same event handler. (But maybe I am being confused by multiple documents being updated concurrently with roughly the same information.)

files/en-us/web/api/screen_capture_api/captured_surface_control/index.md
```diff
@@ -0,0 +1,294 @@
...
+}
+```
+
+## Finished demo
+
+The finished demo renders like this:
+
+{{EmbedLiveSample("surface-control-demo", , "500px", , , , "captured-surface-control")}}
+
+EDITORIAL: The demo works, but when you scroll the captured display surface, it crashes the browser tab.
```
eladalon1983: Really?! Have you filed a Chromium bug? (If so, please assign it to me.)

files/en-us/web/api/capturecontroller/forwardwheel/index.md
chrisdavidmills: Ah, good catch. I've checked through the docs and edited them to make sure I'm not referring to captured wheel events here, or anywhere else.

files/en-us/web/api/capturecontroller/forwardwheel/index.md
chrisdavidmills: I've added something to the page to cover this:

> The `forwardWheel` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation), in which case the user is asked for permission to scroll the captured page. Specifically, the only events that can successfully invoke it are `click` and `input`. If the relevant permission is already `"granted"`, transient activation is not needed.

chrisdavidmills: I wonder if there is anything we need to say about this API in the Permissions API docs. Permissions API integration is not mentioned anywhere in the spec; just Permissions-Policy.

files/en-us/web/api/capturecontroller/forwardwheel/index.md
chrisdavidmills: I've thought about this and elected to include the following at the start of each brief code snippet:

```js
// Create controller and start capture
const controller = new CaptureController();
videoElem.srcObject = await navigator.mediaDevices.getDisplayMedia({
  controller,
});

// ...
```

Just so that we show the context for the `controller` object then being used.

files/en-us/web/api/capturecontroller/getsupportedzoomlevels/index.md
chrisdavidmills: Good point. I've cut this down to just focus on the specific bit that uses `getSupportedZoomLevels()`.

files/en-us/web/api/capturecontroller/getsupportedzoomlevels/index.md
chrisdavidmills: Good call. I've updated the relevant functions in the guide to use try/catch, and added a note highlighting why you should do this. I've also added the code blocks and note to the reference pages where appropriate (so, this one, and the `decreaseZoomLevel()` and `increaseZoomLevel()` pages).

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
chrisdavidmills: Done. Added to this page, and the guide.

files/en-us/web/api/capturecontroller/zoomlevel/index.md
chrisdavidmills: Yeah, it is used quite a lot. I've updated the wording on this page to at least make it more relevant to the `zoomLevel` property.

files/en-us/web/api/screen_capture_api/captured_surface_control/index.md
chrisdavidmills: Done: https://issues.chromium.org/issues/427295493.

I haven't assigned it to you because I don't think I have the permissions to do so. Mentioning @guidou so that it is handled while you are OOO.

hamishwillee:
Ping me when you're ready for editorial review. Remember Will is away next week.

chrisdavidmills:
> Ping me when you're ready for editorial review. Remember Will is away next week.

Ah, thanks @hamishwillee. I think you can start the editorial review. Maybe leave it till Friday, in case @guidou has any technical comments?

hamishwillee:
Probably Monday :-)

files/en-us/web/http/reference/headers/permissions-policy/index.md
```diff
@@ -94,6 +94,9 @@ You can specify
 - {{httpheader('Permissions-Policy/camera', 'camera')}} {{experimental_inline}}
   - : Controls whether the current document is allowed to use video input devices. When this policy is disabled, the {{jsxref("Promise")}} returned by {{domxref("MediaDevices.getUserMedia", "getUserMedia()")}} will reject with a `NotAllowedError` {{DOMxRef("DOMException")}}.
 
+- {{HTTPHeader('Permissions-Policy/captured-surface-control', 'captured-surface-control')}} {{experimental_inline}}
+  - : Controls whether or not the document is permitted to use the [Captured Surface Control API](/en-US/docs/Web/API/Screen_Capture_API/Captured_Surface_Control). When this policy is disabled, the promise returned by the API's main methods will reject with a `NotAllowedError` {{DOMxRef("DOMException")}} if permission is not obtained.
```
hamishwillee: This sentence does not make sense.

> When this policy is disabled, the promise returned by the API's main methods will reject with a `NotAllowedError` {{DOMxRef("DOMException")}} if permission is not obtained.

The previous document also misuses the term "disabled", and I may well have written that. I.e. you can't "disable a policy". You can only set the scope to which the permission is granted - that's not disabling.

```suggestion
  - : Controls whether or not the document is permitted to use the [Captured Surface Control API](/en-US/docs/Web/API/Screen_Capture_API/Captured_Surface_Control).
    The promise returned by the API's main methods will reject with a `NotAllowedError` {{DOMxRef("DOMException")}} if the permission is not allowed.
```

files/en-us/web/http/reference/headers/permissions-policy/index.md
```diff
@@ -94,6 +94,9 @@ You can specify
 - {{httpheader('Permissions-Policy/camera', 'camera')}} {{experimental_inline}}
   - : Controls whether the current document is allowed to use video input devices. When this policy is disabled, the {{jsxref("Promise")}} returned by {{domxref("MediaDevices.getUserMedia", "getUserMedia()")}} will reject with a `NotAllowedError` {{DOMxRef("DOMException")}}.
```
hamishwillee: For same reasons as https://github.com/mdn/content/pull/39982/files#r2188763920
```suggestion
  - : Controls whether the current document is allowed to use video input devices.
    The {{jsxref("Promise")}} returned by {{domxref("MediaDevices.getUserMedia", "getUserMedia()")}} will reject with a `NotAllowedError` {{DOMxRef("DOMException")}} if the permission is not allowed.
```

files/en-us/web/http/reference/headers/permissions-policy/captured-surface-control/index.md
```diff
@@ -0,0 +1,43 @@
...
+status:
+  - experimental
+browser-compat: http.headers.Permissions-Policy.captured-surface-control
+---
+
+{{HTTPSidebar}}{{SeeCompatTable}}
+
+The HTTP {{HTTPHeader("Permissions-Policy")}} header `captured-surface-control` directive controls whether or not the document is permitted to use the [Captured Surface Control API](/en-US/docs/Web/API/Screen_Capture_API/Captured_Surface_Control). Specifically, the {{domxref("CaptureController.forwardWheel", "forwardWheel()")}}, {{domxref("CaptureController.increaseZoomLevel", "increaseZoomLevel()")}}, {{domxref("CaptureController.decreaseZoomLevel", "decreaseZoomLevel()")}}, and {{domxref("CaptureController.resetZoomLevel", "resetZoomLevel()")}} methods are controlled by this directive.
+
+If `captured-surface-control` is disabled in a document, the above methods will throw a `NotAllowedError` exception.
```
hamishwillee: Disabled isn't quite right as per https://github.com/mdn/content/pull/39982/files#r2188763920

It is more that the permission  has not been allowed for the document.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
```diff
@@ -0,0 +1,78 @@
...
+page-type: web-api-instance-method
+status:
+  - experimental
+browser-compat: api.CaptureController.decreaseZoomLevel
+---
+
+{{APIRef("Screen Capture API")}}{{SeeCompatTable}}{{SecureContext_Header}}
+
+The {{domxref("CaptureController")}} interface's **`decreaseZoomLevel()`** method decreases the captured display surface's zoom percentage to the next level down.
+
```
hamishwillee, line 16: 1. The spec also says this requires transient activation. We should say this in all cases:

   > It follows from these steps that [increaseZoomLevel](https://w3c.github.io/mediacapture-surface-control/#dom-capturecontroller-increasezoomlevel)(), [decreaseZoomLevel](https://w3c.github.io/mediacapture-surface-control/#dom-capturecontroller-decreasezoomlevel)() and [resetZoomLevel](https://w3c.github.io/mediacapture-surface-control/#dom-capturecontroller-resetzoomlevel)() are only callable with [transient activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation), because [permitted event types for zoom-setting](https://w3c.github.io/mediacapture-surface-control/#dfn-permitted-event-types-for-zoom-setting) only contains [event types](https://dom.spec.whatwg.org/#dom-event-type) that confer this activation.

2. The spec says that this requires the permission-policy permission, which you have captured in `NotAllowedError`. But if you're going to refer to transient activation you might as well also say this here.

3. You need more than just the `captured-surface-control` permission-policy to be allowed in your document. This is what the specs refer to as a powerful feature so you also need the `captured-surface-control` permission as per the permissions API. They are different things. This is what causes the prompt when you first try to access any of the APIs.

   The interrelationship between these two things is confusing because they share the same name but they do different things. The permission-policy gates access via the server. The permission API gates via the user giving permission. If the permission policy is not granted, the user does not get asked. The permission API query for support aggregates both - you'll get denied if the permission-policy header doesn't allow the document permission.
   
This all implies a bunch of changes to this PR in a number of places. This is what I have done in a number of similar APIs - using the clipboard API as an example
- Create a security considerations section in the API Overview like this: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations
- Create an API specific section in each affected API like this https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read#security_considerations
- Add a mention of the permission to the Permission API https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API#permission-aware_apis and make sure the BCD includes this permission https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API#api.permissions
- Note in your exceptions that it is more than just the permission-policy 

That's a bit of churn across all the docs, so I won't detail review until you've thought about this.

EDIT as per https://github.com/mdn/content/pull/39982/files#r2188788630 I somewhat prefer what you have done `forwarWheel` where the  security considerations are in the top description rather than their own section. But IMO we should still have the overview security considerations and we need the information done for all methods and consistent.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
```diff
@@ -0,0 +1,78 @@
...
+### Return value
+
+A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
+
+### Exceptions
+
+- `InvalidStateError` {{domxref("DOMException")}}
+  - : The captured display surface is already at its minimum supported zoom level.
+- `NotAllowedError` {{domxref("DOMException")}}
+  - : The operation is disallowed by a {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy).
```
hamishwillee: Disallowed by permission policy and not granted by user (permission API). Also requires transient activation. Unlike the forward wheel case I am not sure whether granted permission means transient activation not required. I think both might be required.

https://github.com/mdn/content/pull/39982/files#r2188783496

We should check these are all consistent.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
```diff
@@ -0,0 +1,78 @@
...
+  controller,
+});
+
+// ...
+
+decBtn.addEventListener("click", decreaseZoom);
+
+async function decreaseZoom() {
+  try {
+    controller.decreaseZoomLevel();
```
hamishwillee: Should we be awaiting?

hamishwillee: PS Can you check the other cases. I am pretty sure we should be, but it is the kind of thing I don't normally notice

files/en-us/web/api/capturecontroller/forwardwheel/index.md
```diff
@@ -0,0 +1,88 @@
...
+status:
+  - experimental
+browser-compat: api.CaptureController.forwardWheel
+---
+
+{{APIRef("Screen Capture API")}}{{SeeCompatTable}}{{SecureContext_Header}}
+
+The {{domxref("CaptureController")}} interface's **`forwardWheel()`** method starts forwarding {{domxref("Element.wheel_event", "wheel")}} events fired on the referenced element to the viewport of an associated captured display surface.
+
+The `forwardWheel` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation), in which case the user is asked for permission to scroll the captured page. Specifically, the only events that can successfully invoke it are `click` and `input`. If the relevant permission is already `"granted"`, transient activation is not needed.
```
hamishwillee: Ah, and here I see that you do know about permissions etc :-). Presumably you just missed it for decreaseZoom level. 

I quite like having all this information here. I'd argue that since we don't have a standard for security considerations it is better to have this information here for the API reference. BUT you should still have the overview page section on this.
Which approach you use up to you, but be consistent.

See https://github.com/mdn/content/pull/39982/files#r2188783496

files/en-us/web/api/capturecontroller/forwardwheel/index.md
```diff
@@ -0,0 +1,88 @@
...
+```js
+async function startForwarding() {
+  try {
+    await controller.forwardWheel(videoElem);
+  } catch (e) {
+    console.log(e);
+  }
+}
+```
+
```
hamishwillee, line 75: Up to you, but should we link to the wheel events too?

files/en-us/web/api/capturecontroller/getsupportedzoomlevels/index.md
```diff
@@ -0,0 +1,78 @@
...
+None.
+
+### Return value
+
+An array of numbers representing the different zoom percentages that the captured display surface supports.
+
+### Exceptions
+
+- `InvalidStateError` {{domxref("DOMException")}}
+  - : The capturing {{domxref("MediaStream")}} returned by the originating {{domxref("MediaDevices.getDisplayMedia()")}} call is no longer capturing, for example because the associated {{domxref("MediaStreamTrack")}} objects have had {{domxref("MediaStreamTrack.stop", "stop()")}} called on them.
```
hamishwillee, line 32: Spec also says 

If surfaceType is not a [supported display surface type](https://w3c.github.io/mediacapture-surface-control/#dfn-supported-display-surface-type), [throw](https://webidl.spec.whatwg.org/#dfn-throw) a "[NotSupportedError](https://webidl.spec.whatwg.org/#notsupportederror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException).

files/en-us/web/api/capturecontroller/increasezoomlevel/index.md
```diff
@@ -0,0 +1,78 @@
...
+### Return value
+
+A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
+
+### Exceptions
+
+- `InvalidStateError` {{domxref("DOMException")}}
+  - : The captured display surface is already at its maximum supported zoom level.
+- `NotAllowedError` {{domxref("DOMException")}}
+  - : The operation is disallowed by a {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy).
```
hamishwillee: All same comments as decreaseZoomLevel

files/en-us/web/api/capturecontroller/increasezoomlevel/index.md
```diff
@@ -0,0 +1,78 @@
...
+  controller,
+});
+
+// ...
+
+incBtn.addEventListener("click", increaseZoom);
+
+async function increaseZoom() {
+  try {
+    controller.increaseZoomLevel();
```
hamishwillee: await>

files/en-us/web/api/capturecontroller/index.md
```diff
@@ -18,11 +18,31 @@ A `CaptureController` object is associated with a capture session by passing it
...
 
+## Instance properties
+
+- {{ domxref("CaptureController.zoomLevel", "zoomLevel") }} {{Experimental_Inline}}
+  - : The captured display surface's current zoom percentage.
+
 ## Instance methods
 
+- {{ domxref("CaptureController.decreaseZoomLevel", "decreaseZoomLevel()") }} {{Experimental_Inline}}
+  - : Decreases the captured display surface's zoom percentage to the next level down.
```
hamishwillee: It is not a percentage. Well it is, but the spec goes to some effort to refer to it as some kind of browser/specific arbitrary scale. Maybe we should do the same?

```suggestion
  - : Decreases the captured display surface's zoom level by one increment.
```

files/en-us/web/api/capturecontroller/index.md
```diff
@@ -18,11 +18,31 @@ A `CaptureController` object is associated with a capture session by passing it
...
+- {{ domxref("CaptureController.forwardWheel", "forwardWheel()") }} {{Experimental_Inline}}
+  - : Starts forwarding {{domxref("Element.wheel_event", "wheel")}} events fired on the referenced element to the viewport of an associated captured display surface.
+- {{ domxref("CaptureController.getSupportedZoomLevels", "getSupportedZoomLevels()") }} {{Experimental_Inline}}
+  - : Returns the different zoom percentages that the captured display surface supports.
```
hamishwillee: As above, prefer levels over percentages, and this grammar more readable.

```suggestion
  - : Returns the different zoom levels supported by the captured display surface.
```

files/en-us/web/api/capturecontroller/index.md
```diff
@@ -18,11 +18,31 @@ A `CaptureController` object is associated with a capture session by passing it
...
+- {{ domxref("CaptureController.increaseZoomLevel", "increaseZoomLevel()") }} {{Experimental_Inline}}
+  - : Increases the captured display surface's zoom percentage to the next level up.
```
hamishwillee: If the decreateZoom... gets updated, this will have to be updated too.

files/en-us/web/api/capturecontroller/index.md
```diff
@@ -18,11 +18,31 @@ A `CaptureController` object is associated with a capture session by passing it
...
+- {{ domxref("CaptureController.resetZoomLevel", "resetZoomLevel()") }} {{Experimental_Inline}}
+  - : Resets the captured display surface's zoom percentage to its initial level, which is `100`.
```
hamishwillee:
```suggestion
  - : Resets the captured display surface's zoom to its initial level, which is `100`.
```

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+None.
+
+### Return value
+
+A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
+
+### Exceptions
+
+- `NotAllowedError` {{domxref("DOMException")}}
```
hamishwillee, line 35: All same comments for permission etc

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
```diff
@@ -0,0 +1,70 @@
...
+videoElem.srcObject = await navigator.mediaDevices.getDisplayMedia({
+  controller,
+});
+
+// ...
+
+resetBtn.addEventListener("click", resetZoom);
+
+async function resetZoom() {
+  controller.resetZoomLevel();
```
hamishwillee: await?

files/en-us/web/api/capturecontroller/zoomlevel/index.md
```diff
@@ -0,0 +1,55 @@
...
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.CaptureController.zoomLevel
+---
+
+{{APIRef("Screen Capture API")}}{{SeeCompatTable}}
+
+The **`zoomLevel`** read-only property of the
+{{domxref("CaptureController")}} interface returns the captured display surface's current zoom percentage.
```
hamishwillee: No arbitrary breaks. Actually I prefer breaks on full stops, but that's just me :-)

```suggestion
The **`zoomLevel`** read-only property of the {{domxref("CaptureController")}} interface returns the captured display surface's current zoom level.
```

files/en-us/web/api/capturecontroller/zoomlevel/index.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Value
+
+A number representing the current zoom percentage of the captured display surface.
```
hamishwillee: Everywhere, I still prefer level. But up to you.

```suggestion
A number representing the current zoom level of the captured display surface.
```

files/en-us/web/api/capturecontroller/index.md
```diff
@@ -18,11 +18,31 @@ A `CaptureController` object is associated with a capture session by passing it
 - {{ domxref("CaptureController.CaptureController", "CaptureController()") }} {{Experimental_Inline}}
```
hamishwillee, line 18: The description above is VERY abstract. Do you think we might explain what a capture session actually is here, or is used for?

files/en-us/web/api/mediadevices/getdisplaymedia/index.md
```diff
@@ -28,7 +28,7 @@ getDisplayMedia(options)
...
     - `video` {{optional_inline}}
-      - : A boolean or a {{domxref("MediaTrackConstraints")}} instance; the default value is `true`. If this option is omitted or set to `true`, the stream will contain a video track A value of `true` indicates that the returned {{domxref("MediaStream")}} will contain a video track. Since `getDisplayMedia()` requires a video track, if this option is set to `false` the promise will reject with a `TypeError`.
+      - : A boolean or a {{domxref("MediaTrackConstraints")}} instance; the default value is `true`. If this option is omitted or set to `true`, the returned {{domxref("MediaStream")}} will contain a video track. Since `getDisplayMedia()` requires a video track, if this option is set to `false` the promise will reject with a `TypeError`.
```
hamishwillee, line 31: So just to be clear, this has to be `true` for this method to resolve rather than reject. My interpretation of the `audio` below then is that if that is true then the source has both video and audio. But you can't have just audio on the surface. Right?

files/en-us/web/api/screen_capture_api/index.md
```diff
@@ -27,17 +28,27 @@ captureStream =
...
 
-The {{jsxref("Promise")}} returned by `getDisplayMedia()` resolves to a {{domxref("MediaStream")}} which streams the captured media.
+The {{jsxref("Promise")}} returned by `getDisplayMedia()` resolves to a {{domxref("MediaStream")}} which streams the captured display surface.
 
 See the article [Using the Screen Capture API](/en-US/docs/Web/API/Screen_Capture_API/Using_Screen_Capture) for a more in-depth look at how to use the API to capture screen contents as a stream.
 
-The Screen Capture API also has features that limit the part of the screen captured in the stream:
+### Screen capture extensions
+
+The Screen Capture API also has features that extend the capabilities available to the Screen Capture API:
```
hamishwillee: This feels like a tautology. Even if it isn't, can we reword?

files/en-us/web/api/screen_capture_api/index.md
```diff
@@ -89,13 +100,15 @@ The Screen Capture API adds properties to the following dictionaries defined by
 
 ## Permissions Policy validation
```
hamishwillee: "Validation" seems wrong. I would probably make this security considerations and link to or duplicate the overview security considerations. Permission policy is just part of it.

files/en-us/web/api/screen_capture_api/index.md
```diff
@@ -89,13 +100,15 @@ The Screen Capture API adds properties to the following dictionaries defined by
...
 
-{{Glossary("User agent", "User agents")}} that support [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) (either using the HTTP {{HTTPHeader("Permissions-Policy")}} header or the {{HTMLElement("iframe")}} attribute [`allow`](/en-US/docs/Web/HTML/Reference/Elements/iframe#allow)) can specify a desire to use the Screen Capture API using the directive `display-capture`:
+{{Glossary("User agent", "User agents")}} that support [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) (either using the HTTP {{HTTPHeader("Permissions-Policy")}} header or the {{HTMLElement("iframe")}} attribute [`allow`](/en-US/docs/Web/HTML/Reference/Elements/iframe#allow)) can specify a desire to use the Screen Capture API using the directive {{HTTPHeader("Permissions-Policy/display-capture", "display-capture")}}:
```
hamishwillee: This isn't right. It says the user agent can specify a desire to use the XXX API. It is the site that expresses a desire to allow the use the API in certain origins.

files/en-us/web/api/screen_capture_api/index.md
```diff
@@ -89,13 +100,15 @@ The Screen Capture API adds properties to the following dictionaries defined by
...
 
 ```html
 <iframe allow="display-capture" src="/some-other-document.html">…</iframe>
 ```
 
-The default allowlist is `self`, which lets any content within the same origin use Screen Capture.
+You can also specify a desire to use the [Captured Surface Control API](/en-US/docs/Web/API/Screen_Capture_API/Captured_Surface_Control) via the {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} directive. Specifically, the {{domxref("CaptureController.forwardWheel", "forwardWheel()")}}, {{domxref("CaptureController.increaseZoomLevel", "increaseZoomLevel()")}}, {{domxref("CaptureController.decreaseZoomLevel", "decreaseZoomLevel()")}}, and {{domxref("CaptureController.resetZoomLevel", "resetZoomLevel()")}} methods are controlled by this directive.
```
hamishwillee: Maybe "A site can also specify". I don't really like the "desire" wording, but it isn't wrong.

files/en-us/web/http/reference/headers/permissions-policy/index.md
chrisdavidmills: Good point, that man ;-)

files/en-us/web/http/reference/headers/permissions-policy/captured-surface-control/index.md
chrisdavidmills: Got it. I've updated the wording to use the same suggestion you made on the main `Permissions-Policy` doc.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
chrisdavidmills: Thanks for the helping me to figure out what is needed here. The relationship between Permissions API and `Permissions-Policy` is a bit fiddly to understand, but you've helped me a lot in that regard.

I think I've done everything you asked for:

- Adding the "Security considerations" section to the main API landing page.
- Making sure the security information is added and consistent on each of the relevant method pages.
- Adding the mention to the Permissions API page.
- Adding BCD: see https://github.com/mdn/browser-compat-data/pull/27245

Let me know if you think it needs anything else.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
chrisdavidmills: I've added consistent exception entries for Permissions-Policy header, Permissions API permission, and transient activation, to all relevant method pages.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
chrisdavidmills: The code works without `await` in these cases; afaiui, an async feature called from inside an async function without await will still work, but run in a synchronous manner, provided it doesn't rely on the async nature to work. Or something like that.

I've updated these instances to include `await`, and it still works just fine, but I think including `await` is better, as it makes the methods' async nature clearer.

files/en-us/web/api/capturecontroller/forwardwheel/index.md
chrisdavidmills: See earlier comments!

files/en-us/web/api/capturecontroller/forwardwheel/index.md
chrisdavidmills: I linked to the `wheel` event page at the top of the article, but I've also added a sentence below this last code block in the example, which explains the outcome and includes another link to the `wheel` event page:

> This results in the {{domxref("Element.wheel_event", "wheel")}} events fired on the referenced element being forwarded to the captured display surface, allowing the capturing app to scroll it.

files/en-us/web/api/capturecontroller/getsupportedzoomlevels/index.md
chrisdavidmills: Good spot. I covered this in the `forwardWheel()` page with the following:

> - `NotSupportedError` {{domxref("DOMException")}}
>   - : The surface type being captured is not a browser tab. 

So I've added it to this page too.

files/en-us/web/api/capturecontroller/increasezoomlevel/index.md
chrisdavidmills: All dealt with, I think.

files/en-us/web/api/capturecontroller/increasezoomlevel/index.md
chrisdavidmills: added

files/en-us/web/api/capturecontroller/index.md
chrisdavidmills: updated

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
chrisdavidmills: Yup, all added

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
chrisdavidmills: added

files/en-us/web/api/capturecontroller/index.md
chrisdavidmills: This is how I always describe constructors, and noone ever seems to complain ;-)

I do explain what a capture session is and how the `CaptureController` relates to it in the page intro, in a roundabout way.

HOWEVER, on reflection, this is just using an extra term that isn't really needed and creates more confusion. It isn't used anywhere in the regular Screen Capture API docs, except for where I've added it.

I've updated this text, and related text elsewhere, to just use "captured display surface" instead.

files/en-us/web/api/mediadevices/getdisplaymedia/index.md
chrisdavidmills: Yes, correct.

files/en-us/web/api/screen_capture_api/index.md
chrisdavidmills: Bah, not my finest moment, that. I've updated it to:

> The Screen Capture API has additional features that extend its capabilities:

files/en-us/web/api/screen_capture_api/index.md
chrisdavidmills: > I would probably make this security considerations

This is exactly what I did, so I think the overall structure is OK, even if some of the details need tweaking?

files/en-us/web/api/screen_capture_api/index.md
chrisdavidmills: Ah yes, I noticed that. I updated "User agents" to "Websites".

files/en-us/web/api/screen_capture_api/index.md
chrisdavidmills: Updated to "A site..."

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
hamishwillee: Yeah. I tend to think that if you're not going to await it then you don't need to wrap it in another function - i.e. it is pointless to do so unless you are awaiting.

files/en-us/web/api/capturecontroller/forwardwheel/index.md
hamishwillee: Reading through my review - I was so repetitive. Sorry!

files/en-us/web/api/capturecontroller/forwardwheel/index.md
hamishwillee: That's good

files/en-us/web/api/mediadevices/getdisplaymedia/index.md
hamishwillee: Thanks very much. It makes sense, was just unexpected to me.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
```diff
@@ -10,7 +10,9 @@ browser-compat: api.CaptureController.decreaseZoomLevel
 
 {{APIRef("Screen Capture API")}}{{SeeCompatTable}}{{SecureContext_Header}}
 
-The {{domxref("CaptureController")}} interface's **`decreaseZoomLevel()`** method decreases the captured display surface's zoom percentage to the next level down.
+The {{domxref("CaptureController")}} interface's **`decreaseZoomLevel()`** method decreases the captured display surface's zoom level by one increment.
+
+The `decreaseZoomLevel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation), in which case the user is asked for permission to zoom the captured page. If the relevant permission is already `"granted"`, transient activation and permission gathering is not needed.
```
hamishwillee: Just to be clear:

> If the relevant permission is already `"granted"`, transient activation and permission gathering is not needed.

I know this to be true for the forwardWheel based on the way the spec flows. However if you look at the spec for this method and for the increase method (https://w3c.github.io/mediacapture-surface-control/#subroutine-setting-the-zoom-level) it seems that the transient test happens well before the permission check - which implies that it is always run.

I could be wrong - we should ask the spec authors. Note also https://github.com/mdn/content/pull/39982/files#r2191243157 where I expand a little on what I mean by this.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
```diff
@@ -29,9 +31,13 @@ A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
 ### Exceptions
 
 - `InvalidStateError` {{domxref("DOMException")}}
-  - : The captured display surface is already at its minimum supported zoom level.
+  - : Thrown when
```
hamishwillee:
```suggestion
  - : Thrown when:
```

files/en-us/web/api/capturecontroller/forwardwheel/index.md
```diff
@@ -35,11 +35,11 @@ A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
...
     - The application is capturing itself.
-    - An attempt was made to invoke `forwardWheel()` without transient activation, when permission to use it has not already been granted by the user.
+    - An attempt is made to invoke `forwardWheel()` without transient activation, when permission to use it has not already been granted by the user.
 - `NotAllowedError` {{domxref("DOMException")}}
   - : Thrown when:
-    - The operation is disallowed by a {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy).
-    - Permission to forward wheel events was explicitly denied by the user.
+    - A {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) is set that does not permit the page to use the Captured Surface Control API.
```
hamishwillee: If permission-policy is enforced by a browser then the policy does not need to be explicitly set - the default is almost always `self`. 

I'm not sure it is worth being quite so pedantic, in particular as it would need to be changed in a few places, but it might be better to say something like.

```suggestion
    - The page {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) does not permit the page to use the Captured Surface Control API.
```

files/en-us/web/api/capturecontroller/increasezoomlevel/index.md
```diff
@@ -10,7 +10,9 @@ browser-compat: api.CaptureController.increaseZoomLevel
 
 {{APIRef("Screen Capture API")}}{{SeeCompatTable}}{{SecureContext_Header}}
 
-The {{domxref("CaptureController")}} interface's **`increaseZoomLevel()`** method increases the captured display surface's zoom percentage to the next level up.
+The {{domxref("CaptureController")}} interface's **`increaseZoomLevel()`** method increases the captured display surface's zoom level by one increment.
+
+The `increaseZoomLevel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation), in which case the user is asked for permission to zoom the captured page. If the relevant permission is already `"granted"`, transient activation and permission gathering is not needed.
```
hamishwillee: See https://github.com/mdn/content/pull/39982/files/14f73dc48b5ea9137f68f047098304b136be4ae4..9117b62158edd0456caa75569ba062f49073bd6d#r2191213343

files/en-us/web/api/capturecontroller/index.md
```diff
@@ -9,9 +9,9 @@ browser-compat: api.CaptureController
 
 {{APIRef("Screen Capture API")}}{{SeeCompatTable}}{{SecureContext_Header}}
 
-The **`CaptureController`** interface provides methods that can be used to further manipulate a capture session separate from its initiation via {{domxref("MediaDevices.getDisplayMedia()")}}.
+The **`CaptureController`** interface provides methods that can be used to further manipulate a captured display surface (captured via {{domxref("MediaDevices.getDisplayMedia()")}})
 
-A `CaptureController` object is associated with a capture session by passing it into a {{domxref("MediaDevices.getDisplayMedia", "getDisplayMedia()")}} call as the value of the options object's `controller` property.
+A `CaptureController` object is associated with a captured display surface by passing it into a `getDisplayMedia()` call as the value of the options object's `controller` property.
```
hamishwillee, line 14: Yes, this is much better.

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
```diff
@@ -10,7 +10,9 @@ browser-compat: api.CaptureController.resetZoomLevel
 
 {{APIRef("Screen Capture API")}}{{SeeCompatTable}}{{SecureContext_Header}}
 
-The {{domxref("CaptureController")}} interface's **`resetZoomLevel()`** method resets the captured display surface's zoom percentage to its initial level, which is `100`.
+The {{domxref("CaptureController")}} interface's **`resetZoomLevel()`** method resets the captured display surface's zoom to its initial level, which is `100`.
+
+The `resetZoomLevel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation), in which case the user is asked for permission to zoom the captured page. If the relevant permission is already `"granted"`, transient activation and permission gathering is not needed.
```
hamishwillee: See https://github.com/mdn/content/pull/39982/files/14f73dc48b5ea9137f68f047098304b136be4ae4..9117b62158edd0456caa75569ba062f49073bd6d#r2191213343

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
```diff
@@ -28,8 +30,12 @@ A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
 
 ### Exceptions
 
+- `InvalidStateError` {{domxref("DOMException")}}
+  - : An attempt is made to invoke `forwardWheel()` without transient activation, when permission to use it has not already been granted by the user.
```
hamishwillee: Copy paste error. This is not forwardWheel

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
```diff
@@ -28,8 +30,12 @@ A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
...
 - `NotAllowedError` {{domxref("DOMException")}}
-  - : The operation is disallowed by a {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy).
+  - : Thrown when:
+    - A {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) is set that does not permit the page to use the Captured Surface Control API.
```
hamishwillee: Same comment as above about setting.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
```diff
@@ -29,9 +31,13 @@ A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
...
+    - The captured display surface is already at its minimum supported zoom level.
+    - An attempt is made to invoke `forwardWheel()` without transient activation, when permission to use it has not already been granted by the user.
```
hamishwillee: Copy paste error. This is not forwardWheel doc.

files/en-us/web/api/capturecontroller/increasezoomlevel/index.md
```diff
@@ -29,9 +31,13 @@ A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
 ### Exceptions
 
 - `InvalidStateError` {{domxref("DOMException")}}
-  - : The captured display surface is already at its maximum supported zoom level.
+  - : Thrown when:
+    - The captured display surface is already at its maximum supported zoom level.
+    - An attempt is made to invoke `forwardWheel()` without transient activation, when permission to use it has not already been granted by the user.
```
hamishwillee: Copy paste error. This is not forwardWheel. Also, as per my other note, I'm not sure it is true for methods other than forward wheel https://github.com/mdn/content/pull/39982#discussion_r2191213343

files/en-us/web/api/capturecontroller/forwardwheel/index.md
```diff
@@ -35,11 +35,11 @@ A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
   - : Thrown when:
     - The capturing {{domxref("MediaStream")}} returned by the originating {{domxref("MediaDevices.getDisplayMedia()")}} call is no longer capturing, for example because the associated {{domxref("MediaStreamTrack")}} objects have had {{domxref("MediaStreamTrack.stop", "stop()")}} called on them.
     - The application is capturing itself.
-    - An attempt was made to invoke `forwardWheel()` without transient activation, when permission to use it has not already been granted by the user.
+    - An attempt is made to invoke `forwardWheel()` without transient activation, when permission to use it has not already been granted by the user.
```
hamishwillee: For some reason the "already" doesn't feel right. Maybe

```suggestion
    - An attempt is made to invoke `forwardWheel()` without transient activation, when permission to use it has not been granted by the user.
```

files/en-us/web/api/screen_capture_api/index.md
hamishwillee: I think structure is good.

files/en-us/web/api/capturecontroller/index.md
hamishwillee: I like the change.

files/en-us/web/api/screen_capture_api/captured_surface_control/index.md
```diff
@@ -0,0 +1,303 @@
...
+<iframe allow="captured-surface-control" src="/some-other-document.html"
+  >…</iframe
+>
+```
+
+Specifically, the {{domxref("CaptureController.forwardWheel", "forwardWheel()")}}, {{domxref("CaptureController.increaseZoomLevel", "increaseZoomLevel()")}}, {{domxref("CaptureController.decreaseZoomLevel", "decreaseZoomLevel()")}}, and {{domxref("CaptureController.resetZoomLevel", "resetZoomLevel()")}} methods are controlled by this directive.
+
+The default allowlist for `captured-surface-control` is `self`, which lets any content within the same origin use Captured Surface Control.
+
+## App HTML
```
hamishwillee: We should also say a few words about the broader security considerations here. You might want to link to the other section. Here's a quick suggestion that adds a note about the user permission too at the end

````suggestion
## A note on permissions

A website can control access to the Captured Surface Control API using the {{HTTPHeader("Permissions-Policy")}} {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} directive, or the equivalent {{HTMLElement("iframe")}} [`allow`](/en-US/docs/Web/HTML/Reference/Elements/iframe#allow)) attribute value:

```html
<iframe allow="captured-surface-control" src="/some-other-document.html"
  >…</iframe
>
```

Specifically, the {{domxref("CaptureController.forwardWheel", "forwardWheel()")}}, {{domxref("CaptureController.increaseZoomLevel", "increaseZoomLevel()")}}, {{domxref("CaptureController.decreaseZoomLevel", "decreaseZoomLevel()")}}, and {{domxref("CaptureController.resetZoomLevel", "resetZoomLevel()")}} methods are controlled by this directive.

The default allowlist for `captured-surface-control` is `self`, which lets any content within the same origin use Captured Surface Control.

In addition, if permission is granted by the policy, the user must also grant permission.
This can either be explicit permission, granted by responding to a prompt, or implicit permission granted by interacting with a control that calls one of the methods (transient activation) when user permission has not been explicitly denied.

## App HTML
````

File: files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md (hunk 12 above)
Comment (hamishwillee) [reply files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md:16]: Looks great structurally. Just nit picking now.

File: files/en-us/web/api/screen_capture_api/index.md [hunk 46]
```diff
@@ -87,15 +98,21 @@ The Screen Capture API adds properties to the following dictionaries defined by
...
+
+The default allowlist for both directives is `self`, which lets any content within the same origin use Screen Capture.
+
+These features are [powerful features](/en-US/docs/Web/Security#secure_contexts_and_feature_permissions), meaning that making sure they are not disallowed via a `Permissions-Policy` is not the only consideration. If they are not disallowed via a `Permissions-Policy`, the user will still be asked for permission to use the listed features via a browser prompt, which uses the [Permissions API](/en-US/docs/Web/API/Permissions_API).
+
+In addition, the specification requires that a user has recently interacted with the page to use these features — this means that [transient activation](/en-US/docs/Glossary/Transient_activation) is required, that is, the features must be invoked inside an appropriate event handler function. For example, the user must click a button to use page zooming functionality.
```
hamishwillee: VERY pedantically I think that it is not correct for the forwardwheel.
As per https://github.com/mdn/content/pull/39982/files#r2191236227 the comment I added was:

> In addition, if permission is granted by the policy, the user must also grant permission.
> This can either be explicit permission, granted by responding to a prompt, or implicit permission granted by interacting with a control that calls one of the methods (transient activation) when user permission has not been explicitly denied.

Usually if you grant explicit permission then I would not expect a requirement for transient activation, and that seems to be the case for the forward wheel. I.e. you grant blanket permission, you deny blanket permission, or you don't do either, but you can grant implicit permission if the permission API prompt is pending (via transient activation).

But as per my comment in https://github.com/mdn/content/pull/39982/files#r2191213343 this looks correct for the zoom methods.

So I think this last bit might need to change. I think we need to confirm with the spec authors in case I am just way off base.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
chrisdavidmills: The trouble is, the spec author is on holiday for a month. I'd rather publish it with a tiny detail that might not be entirely correct than leave it for a month.

I've done a bit more testing, and I think you are right that transient activation is always needed for the zoom methods, but not `forwardWheel()`. Also, the permission checks all seem to happen at the point where screen capture is attempted (i.e. the `getDisplayMedia()` call). This all considered, I've rewritten the info at the top of the zoom methods in this form:

> The `decreaseZoomLevel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation). In addition, the user is asked for permission to share tabs when screen capture is first attempted; if the user grants permission, this also includes permission to scroll and zoom tabs. If the relevant permission is already `"granted"`, permission gathering is not needed.

And the one at the top of the `forwardWheel()` like this:

> The `forwardWheel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation). Specifically, the only events that can successfully invoke it are `click` and `input`. In addition, the user is asked for permission to share tabs when screen capture is first attempted; if the user grants permission, this also includes permission to scroll and zoom tabs. If the relevant permission is already `"granted"`, transient activation and permission gathering are not needed.

These reflect my experience while testing them.

files/en-us/web/api/capturecontroller/forwardwheel/index.md
chrisdavidmills: It's all good - I think this is an important distinction. I've updated this on all relevant pages.

files/en-us/web/api/capturecontroller/increasezoomlevel/index.md
chrisdavidmills: See earlier comment

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
chrisdavidmills: See earlier comment

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
chrisdavidmills: fixed

files/en-us/web/api/capturecontroller/resetzoomlevel/index.md
chrisdavidmills: updated

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
chrisdavidmills: fixed

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
chrisdavidmills: I also updated the exceptions sections to suit.

files/en-us/web/api/capturecontroller/increasezoomlevel/index.md
chrisdavidmills: updated

files/en-us/web/api/screen_capture_api/captured_surface_control/index.md
chrisdavidmills: Your wording largely sounded good here. I've mostly used it, with some tweaks. I've also added a link, as suggested.

files/en-us/web/api/screen_capture_api/index.md
chrisdavidmills: Have a look at the update I've made, and see what you think. I'm sure I was told this at some point.

I mean, if you look at the developer.chrome.com article, it doesn't mention transient activation at all.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
hamishwillee: 1. The two proposed statements don't reflect "I've done a bit more testing, and I think you are right that transient activation is always needed for the zoom methods, but not forwardWheel(). "

An accurate statement based on that ^^^ would be:

> The `decreaseZoomLevel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation). In addition, the user is asked for permission to share tabs when screen capture is first attempted; if the user denies permission the zoom level cannot be changed even with transient activation.

I.e. Explicit denial prevents access whether or not you have permission. But if you have permission, or permission is pending, then transient activation will work.

And the one at the top of the `forwardWheel()` like this:

> The `forwardWheel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation). Specifically, the only events that can successfully invoke it are `click` and `input`. In addition, the user is asked for permission to share tabs when screen capture is first attempted; if the user grants permission, this also includes permission to scroll tabs. If the relevant permission is already `"granted"`, transient activation is not needed.

i.e. Explicit permission means no transient activation. Explicit denial means no access under any conditions. Permission pending means transient activation.

I also removed "and permission gathering a" from the end of that, as not needed.


2. Yes, it is better to get this in, as long as you are willing to track possible fixing.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
```diff
@@ -12,7 +12,7 @@ browser-compat: api.CaptureController.decreaseZoomLevel
 
 The {{domxref("CaptureController")}} interface's **`decreaseZoomLevel()`** method decreases the captured display surface's zoom level by one increment.
 
-The `decreaseZoomLevel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation), in which case the user is asked for permission to zoom the captured page. If the relevant permission is already `"granted"`, transient activation and permission gathering is not needed.
+The `decreaseZoomLevel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation). In addition, the user is asked for permission to share tabs when screen capture is first attempted; if the user grants permission, this also includes permission to scroll and zoom tabs. If the relevant permission is already `"granted"`, permission gathering is not needed.
```
hamishwillee: See https://github.com/mdn/content/pull/39982#discussion_r2199240786 - in this case permission is entirely about denying access - you still need transient activation to use the feature. Also I don't think the last sentence is needed.

It might be I was confused by your intent. In which case this is slightly wrong, but yes, not so bad that we couldn't merge and come back to fix. 

I won't comment elsewhere but same logic applies.

files/en-us/web/api/capturecontroller/forwardwheel/index.md
```diff
@@ -12,7 +12,7 @@ browser-compat: api.CaptureController.forwardWheel
 
 The {{domxref("CaptureController")}} interface's **`forwardWheel()`** method starts forwarding {{domxref("Element.wheel_event", "wheel")}} events fired on the referenced element to the viewport of an associated captured display surface.
 
-The `forwardWheel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation), in which case the user is asked for permission to scroll the captured page. Specifically, the only events that can successfully invoke it are `click` and `input`. If the relevant permission is already `"granted"`, transient activation and permission gathering is not needed.
+The `forwardWheel()` method must be invoked via [transient activation](/en-US/docs/Glossary/Transient_activation). Specifically, the only events that can successfully invoke it are `click` and `input`. In addition, the user is asked for permission to share tabs when screen capture is first attempted; if the user grants permission, this also includes permission to scroll and zoom tabs. If the relevant permission is already `"granted"`, transient activation and permission gathering are not needed.
```
hamishwillee: See https://github.com/mdn/content/pull/39982#discussion_r2199240786 - in this case granted permission removes need for transient activation, denial removes access. Not sure the last sentence needed.

I won't comment elsewhere but same logic applies

files/en-us/web/api/capturecontroller/forwardwheel/index.md
```diff
@@ -35,11 +35,11 @@ A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
...
     - The application is capturing itself.
-    - An attempt is made to invoke `forwardWheel()` without transient activation, when permission to use it has not already been granted by the user.
+    - An attempt is made to invoke `forwardWheel()` without transient activation, when permission to use it has not been granted by the user.
 - `NotAllowedError` {{domxref("DOMException")}}
   - : Thrown when:
-    - A {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) is set that does not permit the page to use the Captured Surface Control API.
-    - Permission to forward wheel events is explicitly denied by the user.
+    - The page's {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} [Permissions Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) does not permit the page to use the Captured Surface Control API.
+    - Permission to capture the display surface is explicitly denied by the user.
```
hamishwillee, line 42: These changes all great.

files/en-us/web/api/screen_capture_api/index.md
hamishwillee: See https://github.com/mdn/content/pull/39982#discussion_r2199240786

As I said their, I don't think your text quite matches your intent based on testing. But I'm not sure whether we misunderstand each other, and I wouldn't block if you want to merge as long as you are willing to continue to chase this.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
chrisdavidmills: OK, so I think it is clear from this exchange that I find this stuff really bloody confusing ;-)

Thanks for helping me out with the wording. I've updated all those pages as required.

Yup, I'm happy to track potential fixes.

files/en-us/web/api/screen_capture_api/index.md
chrisdavidmills: I've updated it to:

> In addition, the specification requires that a user has recently interacted with the page to use these features — this means that [transient activation](/en-US/docs/Glossary/Transient_activation) is required. See the individual method pages for more details.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
chrisdavidmills: I've used the wording you suggested earlier for this section.

files/en-us/web/api/capturecontroller/forwardwheel/index.md
chrisdavidmills: Again, I've used the wording you suggested.

files/en-us/web/api/capturecontroller/decreasezoomlevel/index.md
hamishwillee: You're not alone! This is my third rodeo on this and I do get it. BUT, that doesn't mean that the spec authors haven't opted for something slightly different in this case. 

I'll look again on Monday.

files/en-us/web/api/screen_capture_api/captured_surface_control/index.md
```diff
@@ -0,0 +1,307 @@
...
+<iframe allow="captured-surface-control" src="/some-other-document.html">
+  ...
+</iframe>
+```
+
+Specifically, the {{domxref("CaptureController.forwardWheel", "forwardWheel()")}}, {{domxref("CaptureController.increaseZoomLevel", "increaseZoomLevel()")}}, {{domxref("CaptureController.decreaseZoomLevel", "decreaseZoomLevel()")}}, and {{domxref("CaptureController.resetZoomLevel", "resetZoomLevel()")}} methods are controlled by this directive.
+
+The default allowlist for `captured-surface-control` is `self`, which lets any content within the same origin use Captured Surface Control.
+
+In addition, if permission is granted by the policy, the user must also grant permission. This can either be explicit permission, granted by responding to a prompt, or implicit permission granted by interacting with a control that calls one of the methods ([transient activation](/en-US/docs/Glossary/Transient_activation)) when user permission has not been explicitly denied.
```
hamishwillee: A bit of a nit, but I wanted to make it clear that the policy gates the ability of a user to offer permission

```suggestion
If permission is allowed by the website policy, the user may then grant (or deny) permission to access the controlled APIs. This can either be explicit permission, granted by responding to a prompt, or implicit permission granted by interacting with a control that calls one of the methods ([transient activation](/en-US/docs/Glossary/Transient_activation)) when user permission has not been explicitly denied.
```

files/en-us/web/api/screen_capture_api/index.md
```diff
@@ -87,15 +98,21 @@ The Screen Capture API adds properties to the following dictionaries defined by
...
 
 ```html
 <iframe allow="display-capture" src="/some-other-document.html">…</iframe>
 ```
 
-The default allowlist is `self`, which lets any content within the same origin use Screen Capture.
+A site can also specify a desire to use the [Captured Surface Control API](/en-US/docs/Web/API/Screen_Capture_API/Captured_Surface_Control) via the {{HTTPHeader("Permissions-Policy/captured-surface-control", "captured-surface-control")}} directive. Specifically, the {{domxref("CaptureController.forwardWheel", "forwardWheel()")}}, {{domxref("CaptureController.increaseZoomLevel", "increaseZoomLevel()")}}, {{domxref("CaptureController.decreaseZoomLevel", "decreaseZoomLevel()")}}, and {{domxref("CaptureController.resetZoomLevel", "resetZoomLevel()")}} methods are controlled by this directive.
+
+The default allowlist for both directives is `self`, which lets any content within the same origin use Screen Capture.
```
hamishwillee:
```suggestion
The default allowlist for both directives is `self`, which permits any content within the same origin use Screen Capture.
```

files/en-us/web/api/screen_capture_api/index.md
```diff
@@ -87,15 +98,21 @@ The Screen Capture API adds properties to the following dictionaries defined by
...
+
+These features are [powerful features](/en-US/docs/Web/Security#secure_contexts_and_feature_permissions), meaning that making sure they are not disallowed via a `Permissions-Policy` is not the only consideration. If they are not disallowed via a `Permissions-Policy`, the user will still be asked for permission to use the listed features via a browser prompt, which uses the [Permissions API](/en-US/docs/Web/API/Permissions_API).
```
hamishwillee: I think this is slightly better. 

```suggestion
These methods are considered [powerful features](/en-US/docs/Web/Security#secure_contexts_and_feature_permissions), which means that even if permission is allowed via a `Permissions-Policy`, the user will still be prompted for permission to use them. The [Permissions API](/en-US/docs/Web/API/Permissions_API) can be used to query the aggregate permission (from both the website and the user) for using the listed features.
```

hamishwillee (approved):
@chrisdavidmills I think this looks good - certainly it is accurate as I read the spec. Hopefully the spec authors will confirm when they return.

I have added 3 nit suggestions. Approving so you are not blocked if you disagree with them.

PR #39985, opened 2025-06-18
Document subclassing Event as an alternative to CustomEvent

tbroyer:
### Description

Document subclassing Event as an alternative to CustomEvent

Also remove "the old-fashioned way" section, as the Event constructor has been available across browsers for nearly 10 years now.

### Motivation

Many people don't realize that it's possible to extend the Event interface, which generally makes for a better API (removes the `.detail` intermediate, can possibly use `instanceof`). The pages that deal with "custom events" in the MDN don't mention that possibility.


files/en-us/web/api/customevent/index.md
```diff
@@ -7,7 +7,7 @@ browser-compat: api.CustomEvent
 
 {{APIRef("DOM")}}{{AvailableInWorkers}}
 
-The **`CustomEvent`** interface represents events initialized by an application for any purpose.
+The **`CustomEvent`** interface represents events initialized by an application for any purpose. As an alternative, it is also possible to [subclass the `Event` interface](/en-US/docs/Web/Events/Creating_and_triggering_events#adding_custom_data_–_subclassing_event).
```
wbamberg:
```suggestion
The **`CustomEvent`** interface can be used to attach custom data to an event generated by an application.

As an alternative to `CustomEvent`, you can [subclass the `Event` interface](/en-US/docs/Web/Events/Creating_and_triggering_events#adding_custom_data_–_subclassing_event) to add custom data and behavior.
```

(I think it's better to have the first paragraph just as an overview of what _this interface_ does. And some editorial changes too.)

wbamberg (commented):
Thanks for your PR!

I just had editorial comments on the `CustomEvent` page.

I do have issues with https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events but I think most of them are pre-existing and your change hasn't made them any worse.

wbamberg (approved):
👍 thank you!

PR #39987, opened 2025-06-18
Elaborated on use of Service Workers on http://localhost

evgenyfadeev:
### Description

Mentioned that there is an exception allowing use of service workers on http://localhost and added a note explaining that in more detail.

### Motivation

The document is mentioning that Service Workers only run in secure contexts, but did not mention how to develop service workers on localhost, which by default does not come with an SSL Certificate.

### Additional details

none

none

### DIFF
--- a/files/en-us/web/api/service_worker_api/index.md
+++ b/files/en-us/web/api/service_worker_api/index.md
@@ -17,7 +17,10 @@ Service workers run in a worker context: they therefore have no DOM access and r
 
 Service workers can't import JavaScript modules dynamically, and [`import()`](/en-US/docs/Web/JavaScript/Reference/Operators/import#browser_compatibility) will throw an error if it is called in a service worker global scope. Static imports using the [`import`](/en-US/docs/Web/JavaScript/Reference/Statements/import) statement are allowed.
 
-Service workers only run over HTTPS, for security reasons. Most significantly, HTTP connections are susceptible to malicious code injection by {{Glossary("MitM", "man in the middle")}} attacks, and such attacks could be worse if allowed access to these powerful APIs. In Firefox, service worker APIs are also hidden and cannot be used when the user is in [private browsing mode](https://support.mozilla.org/en-US/kb/private-browsing-use-firefox-without-history).
+Service workers only run over HTTPS for security reasons (with the exception of `http://localhost`, see the note below). Most significantly, HTTP connections are susceptible to malicious code injection by {{Glossary("MitM", "man in the middle")}} attacks, and such attacks could be worse if allowed access to these powerful APIs. In Firefox, service worker APIs are also hidden and cannot be used when the user is in [private browsing mode](https://support.mozilla.org/en-US/kb/private-browsing-use-firefox-without-history).
+
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/service_worker_api/index.md
```diff
@@ -17,7 +17,10 @@ Service workers run in a worker context: they therefore have no DOM access and r
...
 
-Service workers only run over HTTPS, for security reasons. Most significantly, HTTP connections are susceptible to malicious code injection by {{Glossary("MitM", "man in the middle")}} attacks, and such attacks could be worse if allowed access to these powerful APIs. In Firefox, service worker APIs are also hidden and cannot be used when the user is in [private browsing mode](https://support.mozilla.org/en-US/kb/private-browsing-use-firefox-without-history).
+Service workers only run over HTTPS for security reasons (with the exception of `http://localhost`, see the note below). Most significantly, HTTP connections are susceptible to malicious code injection by {{Glossary("MitM", "man in the middle")}} attacks, and such attacks could be worse if allowed access to these powerful APIs. In Firefox, service worker APIs are also hidden and cannot be used when the user is in [private browsing mode](https://support.mozilla.org/en-US/kb/private-browsing-use-firefox-without-history).
+
+> [!NOTE]
+> To facilitate local development, most modern browsers treat `http://localhost` as a secure context, thus allowing for the use of service workers without an SSL certificate.
```
wbamberg:
```suggestion
Service workers are only available in [secure contexts](/en-US/docs/Web/Security/Secure_Contexts): this means that their document is served over HTTPS, although browsers also treat `http://localhost` as a secure context, to facilitate local development. HTTP connections are susceptible to malicious code injection by {{Glossary("MitM", "man in the middle")}} attacks, and such attacks could be worse if allowed access to these powerful APIs.
```

wbamberg (commented):
Thanks for your PR!

I'm not quite sure we need to note the localhost exception here, since it's really just part of the definition on "secure context". But if we do, I think it's better to integrate it into the paragraph than have another blue box (we already have quite the pile of these here).

Also the bit about private browsing seems to be no longer true: https://bugzilla.mozilla.org/show_bug.cgi?id=1320796.

files/en-us/web/api/service_worker_api/index.md
evgenyfadeev: I thought it would help the reader who is not aware that http://localhost is considered secure. Naively one might consider otherwise since it's not https. Perhaps this could be abbreviated too. Up to you ultimately, of course.

wbamberg: My suggestion does keep the bit about localhost, so I'm not sure if your comment here is a pushback to the suggestion.

That aside though: there are lots of APIs in the web platform that require a secure context. Rather than define "secure context" in every place, it's better to define it in one place and consistently link there. On MDN that place is https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts, which is why I've linked to that page in my suggestion.

evgenyfadeev: makes sense, imho a brief mention with a link to a reference would be useful. It's late here and I might have misread your earlier comment. If it's still relevant tomorrow, i'll modify the PR in line with your suggestion.

wbamberg (approved):
Thanks @evgenyfadeev !

PR #39992, opened 2025-06-19
Fix bug in the second code example

MuhammedSaleh:
In the second code example, the call to Atomics.compareExchange method used undeclared variables and wrong numbers of arguments.


Josh-Cena (approved):
Nice catch. Thank you!

PR #39995, opened 2025-06-19
docs: Add warning for Array.prototype.toLocaleString behavior (#39988)

### LINKED ISSUE #39988: Add a prominent warning under `Array#toLocaleString` that it probably doesn't do what you expect
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toLocaleString

From the current article:

> Note that the method itself does not consume the two parameters — it only passes them to the toLocaleString() of each element. The choice of the separator string depends on the host's current locale, not the locales parameter.

This feels like it should be more prominent and formatted as a warning with a recommendation to use `Intl.ListFormat` instead, as it's wildly different from what I'd expect `Array#toLocaleString` to do (which is to behave roughly like `Intl.ListFormat`, possibly with `{ style: 'narrow' }`).

For example:

```js
const nums = [8888, 9999]
nums.toLocaleString('zh')
// expected: "8888、9999"
// actual: "8,888,9,999"
new Intl.ListFormat('zh', { style: 'narrow' }).format(nums.map(String))
// gives expected result "8888、9999"
```

Shrinivassab:
### Description

This PR adds a prominent warning under `Array.prototype.toLocaleString` to clarify its behavior regarding the `locales` parameter and separator usage. It also recommends using `Intl.ListFormat` for list-style formatting.

### Motivation

Developers often misinterpret `toLocaleString` as behaving like `Intl.ListFormat`, expecting it to localize list separators. This change highlights the distinction and prevents confusion by linking to the correct alternative.

### Additional details

- Fixes #39988
- Added example comparing the output of `toLocaleString()` and `Intl.ListFormat`.

Fixes #39988

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/array/tolocalestring/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/array/tolocalestring/index.md
@@ -44,26 +44,27 @@ A string representing the elements of the array.
 
 ## Description
 
-The `Array.prototype.toLocaleString` method traverses its content, calling the `toLocaleString` method of every element with the `locales` and `options` parameters provided, and concatenates them with an implementation-defined separator (such as a comma ","). 
+The `Array.prototype.toLocaleString` method traverses its content, calling the `toLocaleString` method of every element with the `locales` and `options` parameters provided, and concatenates them with an implementation-defined separator (such as a comma ",").
 
-> **⚠️ Warning:**  
-> The `toLocaleString()` method **does not** control the separator used between array elements using the `locales` or `options` arguments. It only passes these arguments to the `toLocaleString()` of each element. The actual separator (usually a comma) depends solely on the host's current locale, not the `locales` parameter.  
-> 
-> If you're expecting localized list formatting (e.g., language-specific conjunctions), consider using [`Intl.ListFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat) instead.  
-> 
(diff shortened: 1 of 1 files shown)


files/en-us/web/javascript/reference/global_objects/array/tolocalestring/index.md
```diff
@@ -44,7 +44,10 @@ A string representing the elements of the array.
 
 ## Description
 
-The `Array.prototype.toLocaleString` method traverses its content, calling the `toLocaleString` method of every element with the `locales` and `options` parameters provided, and concatenates them with an implementation-defined separator (such as a comma ","). Note that the method itself does not consume the two parameters — it only passes them to the `toLocaleString()` of each element. The choice of the separator string depends on the host's current locale, not the `locales` parameter.
+The `Array.prototype.toLocaleString` method traverses its content, calling the `toLocaleString` method of every element with the `locales` and `options` parameters provided, and concatenates them with an implementation-defined separator (such as a comma ",").
+
+> [!WARNING]
```
Josh-Cena:
```suggestion
> [!NOTE]
```

Josh-Cena (approved):
Thanks

We have a specific note syntax. Please see https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#notes_warnings_and_callouts.

Also make sure that notes are brief. Avoid code blocks. Put examples in the examples section.

PR #39998, opened 2025-06-19
Fixed improperly referenced sample

TheyCallMeFatbird:
### Description

I fixed a improperly made and referenced EmbedLiveSample with proper syntax.

### Motivation

I'm making these changes because it's important for readers to see how such result could have an impact on what they're making.


Josh-Cena:
Do you actually see a problem with the current live sample?

`EmbedLiveSample` takes the ID of a heading. It can normalize non-IDs, but most of our `EmbedLiveSample` calls use IDs, not heading text.

PR #40005, opened 2025-06-20
Fix typos in th element page

tim-we:
### Description

Fixed two typos on the `<th>` element MDN page:
- ~speech reader~ -> screen reader
- ~will extends~ -> will extend

### Motivation

I want correctness ;)


chrisdavidmills (approved):
LGTM!

Thanks @tim-we, we appreciate the correctness ;-)

PR #40013, opened 2025-06-21
typo

yg-i:
The property's name is 'mode' not 'options.mode'


hamishwillee:
Thanks, but this fix is incorrect.

You are right that the property is `mode`, but the section you have pointed to is referring to how this **readonly** property is set. That is via the `attachShadow()` method, which does take an options option.

Josh-Cena:
@hamishwillee the current text is confusing though. It says "the `options.mode` property of the object passed to..." which implies that you are passing an object like `{ options: { mode: "open" } }`. It should either remove `options.` or say "the `options.mode` argument passed to..."

hamishwillee (approved):
Thanks @yg-i  and @Josh-Cena 

You're right - sorry!

PR #40025, opened 2025-06-22
Fixing explanations of onsuccess/onerror listeners for IndexedDB

afonyaa:
### Description

The paragraph that aimed to be fixed contained 2 issues:
1. Logical error: there is no 'success'/'error' events on  DOM event type property, according to official specification https://html.spec.whatwg.org/multipage/indices.html#events-2

2. The explanation text was just hard to read and focus, so it is a bit simplified and more concised

### Motivation

I made this change, so the developers do not be confused with unexpected 'error' / 'success' DOM event types

### Additional details
https://html.spec.whatwg.org/multipage/indices.html#events-2

### DIFF
--- a/files/en-us/web/api/indexeddb_api/using_indexeddb/index.md
+++ b/files/en-us/web/api/indexeddb_api/using_indexeddb/index.md
@@ -60,7 +60,7 @@ request.onsuccess = (event) => {
 };
 ```
 
-Which of the two functions, `onsuccess()` or `onerror()`, gets called? If everything succeeds, the `onsuccess()` function is triggered, receiving the success event as its argument, where the target property is the `request` object. Otherwise, if there was any problem, the `onerror()` function is triggered with the error event as its argument.
+If the request succeeds, the {{domxref("IDBRequest.success_event", "success")}} event is fired, and the function assigned to `onsuccess` is called. If the request fails,  the {{domxref("IDBRequest.error_event", "error")}} event is fired, and the function assigned to `onerror` is called.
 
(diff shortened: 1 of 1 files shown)


wbamberg:
> ### Description
> 
> The paragraph that aimed to be fixed contained 2 issues:
> 
> 1. Logical error: there is no 'success'/'error' events on  DOM event type property

The text doesn't say that though, it says that the `success` event is an `Event` whose [`type`](https://developer.mozilla.org/en-US/docs/Web/API/Event/type) is `"success"`, which is correct:

> To fire a success event at a request, run these steps:
>
>    Let event be the result of [creating an event](https://dom.spec.whatwg.org/#concept-event-create) using [Event](https://dom.spec.whatwg.org/#event).
>
>    Set event’s [type](https://dom.spec.whatwg.org/#dom-event-type) attribute to "success".

(from https://w3c.github.io/IndexedDB/#fire-success-event)

However I do agree that the existing text is confusing and wrong in other ways (specifically, there is no `onsuccess()` function: `onsuccess` is an attribute whose value is an anonymous function).

files/en-us/web/api/indexeddb_api/using_indexeddb/index.md
```diff
@@ -60,7 +60,7 @@ request.onsuccess = (event) => {
 };
 ```
 
-Which of the two functions, `onsuccess()` or `onerror()`, gets called? If everything succeeds, a success event (that is, a DOM event whose `type` property is set to `"success"`) is fired with `request` as its `target`. Once it is fired, the `onsuccess()` function on `request` is triggered with the success event as its argument. Otherwise, if there was any problem, an error event (that is, a DOM event whose `type` property is set to `"error"`) is fired at `request`. This triggers the `onerror()` function with the error event as its argument.
+Which of the two functions, `onsuccess()` or `onerror()`, gets called? If everything succeeds, the `onsuccess()` function is triggered, receiving the success event as its argument, where the target property is the `request` object. Otherwise, if there was any problem, the `onerror()` function is triggered with the error event as its argument.
```
wbamberg:
```suggestion
If the request succeeds, the {{domxref("IDBRequest.success_event", "success")}} event is fired, and the function assigned to `onsuccess` is called. If the request fails,  the {{domxref("IDBRequest.error_event", "error")}} event is fired, and the function assigned to `onerror` is called.
```

wbamberg (approved):
👍 thank you!

PR #40042, opened 2025-06-23
Fourth set of CSS active learning updates

chrisdavidmills:
### Description

As part of a larger project to make the challenges/reader exercises in MDN learn more consistent, this PR:

- Makes general wording improvements and makes the interactive reader exercises more obvious in the RWD/media queries articles.
- Removes "Active learning" wording from media queries.
- Removes the multi-col "test your skills", as this is no longer part of the core curriculum flow.


files/en-us/learn_web_development/core/css_layout/media_queries/index.md
```diff
@@ -483,13 +498,15 @@ article {
...
+If you view the layout in Responsive Design Mode in DevTools, or narrow your browser window down to a mobile-like width, you'll see that it works pretty well as a straightforward mobile view of the site.
 
 {{EmbedLiveSample("walkthrough", "", "600px")}}
 
-From this point, start to drag the Responsive Design Mode view wider until you can see that the line lengths are becoming quite long, and we have space for the navigation to display in a horizontal line. This is where we'll add our first media query. We'll use ems, as this will mean that if the user has increased their text size, the breakpoint will happen at a similar line-length but wider viewport, than someone with a smaller text size.
+### Creating a two-column layout for medium widths
 
-Add the following to your CSS:
+Drag the window wider until you can see that the line lengths are becoming quite long; at this point you've got space for the navigation to display in a horizontal line. This is where we'll add our first media query. We'll use ems, as this will mean that if the user has increased their text size, the breakpoint will happen at a similar line-length but wider viewport, than someone with a smaller text size.
```
bsmth: Linking to https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Values_and_units#lengths might also be interesting

```suggestion
Drag the window wider until you can see that the line lengths are becoming quite long; at this point you've got space for the navigation to display in a horizontal line. This is where we'll add our first media query. We'll use `em` units, as this will mean that if the user has increased their text size, the breakpoint will happen at a similar line-length but wider viewport, than someone with a smaller text size.
```

files/en-us/learn_web_development/core/css_layout/responsive_design/index.md
```diff
@@ -467,21 +471,23 @@ If you look at the HTML source of a responsive page, you will usually see the fo
 <meta name="viewport" content="width=device-width,initial-scale=1" />
 ```
 
-This [viewport](/en-US/docs/Web/HTML/Guides/Viewport_meta_element) meta tag tells mobile browsers that they should set the width of the viewport to the device width, and scale the document to 100% of its intended size, which shows the document at the mobile-optimized size that you intended.
+This [viewport](/en-US/docs/Web/HTML/Guides/Viewport_meta_element) meta tag tells mobile browsers that they should set the width of the viewport to the device's width, and scale the document to 100% of its intended size, which shows the document at the mobile-optimized size that you intended.
```
bsmth: We also have https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/viewport btw, I think the dedicated page is relatively new

files/en-us/learn_web_development/core/css_layout/media_queries/index.md
```diff
@@ -534,29 +553,25 @@ Add the following CSS to your styles:
 }
 ```
 
-If you look at the result at different widths you can see how the design responds and works as a single column, two columns, or three columns, depending on the available width. This is a basic example of a mobile first responsive design.
+That's the example finished. If you look at the result at different widths you can see how the design responds and works as a single column, two columns, or three columns, depending on the available width. This is a basic example of a mobile-first responsive design.
 
-## The viewport meta tag
+### viewport meta
 
 If you look at the HTML source in the above example, you'll see the following element included in the head of the document:
```
bsmth, line 560: Same comment here, we can link to https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/viewport, too if you like.

bsmth (approved):
Couple of small suggestions and xrefs to references, but otherwise +1 from me, tnx!

files/en-us/learn_web_development/core/css_layout/responsive_design/index.md
chrisdavidmills: Ah, that's good to know. I've linked to the ref page at the top and the guide at the bottom, as I think both are quite useful.

files/en-us/learn_web_development/core/css_layout/media_queries/index.md
chrisdavidmills: I've linked this to the ref page instead.

PR #40043, opened 2025-06-23
feat(html): Add meta http-equiv attr page

bsmth:
### Description

__additions:__
- Dedicated page for the `http-equiv` attr.
- Sidebar entry

__Changes:__
- updates links from the DL `meta#http-equiv` to the page instead: `meta/http-equiv`

### Motivation

There's a lot to say for this, there's too much info in the single DL, it warrants a separate page, exposing compat data.

- [x] https://github.com/mdn/content/pull/39804


files/en-us/mozilla/firefox/releases/55/index.md
```diff
@@ -143,7 +143,7 @@ Firefox 55 was released on August 8, 2017. This article lists key changes that a
   `<div xml:base="https://example.com/" style="background:url(picture.jpg)"></div>` ([Firefox bug 1350521](https://bugzil.la/1350521)).
 
 - The {{htmlelement("style")}} element's [`scoped`](/en-US/docs/Web/HTML/Reference/Elements/style#scoped) attribute has been hidden behind a pref (`layout.css.scoped-style.enabled`) in content documents in Firefox 55+, as no other browsers support it.
-- Support for the obscure `MSThemeCompatible` value of the {{htmlelement("meta")}} element's [`http-equiv`](/en-US/docs/Web/HTML/Reference/Elements/meta#http-equiv) attribute has been removed from Gecko. No other modern browsers support it, and it was causing compatibility problems ([Firefox bug 966240](https://bugzil.la/966240)).
+- Support for the obscure `MSThemeCompatible` value of the {{htmlelement("meta")}} element's [`http-equiv`](/en-US/docs/Web/HTML/Reference/Elements/meta/http-equiv) attribute has been removed from Gecko. No other modern browsers support it, and it was causing compatibility problems ([Firefox bug 966240](https://bugzil.la/966240)).
```
hamishwillee: Fly by - either remove or link to glossary for Gecko. It's not well known.

```suggestion
- Support for the obscure `MSThemeCompatible` value of the {{htmlelement("meta")}} element's [`http-equiv`](/en-US/docs/Web/HTML/Reference/Elements/meta/http-equiv) attribute has been removed from Firefox. No other modern browsers support it, and it was causing compatibility problems ([Firefox bug 966240](https://bugzil.la/966240)).
```

files/en-us/web/html/reference/elements/meta/index.md
```diff
@@ -21,49 +21,21 @@ The type of metadata provided by the `<meta>` element can be one of the followin
...
-        > - [MDN Understanding WCAG, Guideline 3.2 explanations](/en-US/docs/Web/Accessibility/Guides/Understanding_WCAG/Understandable#guideline_3.2_—_predictable_make_web_pages_appear_and_operate_in_predictable_ways)
-        > - [Understanding Success Criterion 2.2.1 | W3C Understanding WCAG 2.0](https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits-required-behaviors.html)
-        > - [Understanding Success Criterion 2.2.4 | W3C Understanding WCAG 2.0](https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits-postponed.html)
-        > - [Understanding Success Criterion 3.2.5 | W3C Understanding WCAG 2.0](https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-no-extreme-changes-context.html)
-
+- [`http-equiv`](/en-US/docs/Web/HTML/Reference/Elements/meta/http-equiv)
+  - : Defines a pragma directive, which are instructions for the browser for processing the document.
+    The attribute's name is short for `http-equivalent` because the allowed values are names of equivalent HTTP headers.
```
hamishwillee, line 33: The first sentence doesn't mean anything to me, and none of this says what it actually does. I think pragma doesn't mean much to developers who only work in JavaScript.

Can we do (something like):

```suggestion
  - : The `http-equiv` (short for `http-equivalent`) is used to specify a set of HTTP headers that the browser will apply to the document, as though they had been included in the response that returned the page.
```

files/en-us/web/html/reference/elements/meta/http-equiv/index.md
```diff
@@ -0,0 +1,109 @@
...
+title: <meta> http-equiv attribute
+short-title: <meta> http-equiv
+slug: Web/HTML/Reference/Elements/meta/http-equiv
+page-type: html-attribute
+browser-compat: html.elements.meta.http-equiv
+---
+
+{{HTMLSidebar}}
+
+The **`http-equiv`** attribute of the {{htmlelement("meta")}} element allows you to provide pragma directives in document metadata, which are instructions for the browser that could otherwise be given by an HTTP header.
```
hamishwillee: Pragmas don't exist in JavaScript so there is a reasonable argument to saying that "pragma directive" needs to be  explained. I think it could be omitted entirely. I used one approach here https://github.com/mdn/content/pull/40043/files#r2162735420 which could be copied. Or following this pattern

```suggestion
The **`http-equiv`** attribute of the {{htmlelement("meta")}} element allows you to provide processing instructions for the browser that could otherwise be given in an HTTP header.
```

files/en-us/web/html/reference/elements/meta/http-equiv/index.md
```diff
@@ -0,0 +1,109 @@
...
+When a `<meta>` element has a `http-equiv` attribute, a [`content`](/en-US/docs/Web/HTML/Reference/Attributes/content) attribute defines the corresponding `http-equiv` value.
+For example, the following `<meta>` tag tells the browser to refresh the page after 5 minutes:
+
+```html
+<meta http-equiv="Refresh" content="300" />
+```
+
+## Value
+
+The attribute's name is short for `http-equivalent` because the allowed values are names of equivalent HTTP headers.
```
hamishwillee, line 23: I would move this back up into the top section.

files/en-us/web/html/reference/elements/meta/http-equiv/index.md
```diff
@@ -0,0 +1,109 @@
...
+The HTML specification defines the following set of `http-equiv` values:
```
hamishwillee: The attribute's name is `http-equiv` because the allowed values are names of equivalent HTTP headers - the "short for" thing doesn't follow. I'd mention the "short for" up in the top line of the description.

But this is an important point - only a subset of headers are allowed. The fact that this is what the spec says is implicit for anything not marked non-standard. How about

```suggestion
Only a subset of the HTTP headers are supported as `http-equiv` values.
These include:
```

hamishwillee (commented):
Looks great. A few nitish things

files/en-us/web/html/reference/elements/meta/index.md
```diff
@@ -21,49 +21,21 @@ The type of metadata provided by the `<meta>` element can be one of the followin
 This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Global_attributes).
 
 > [!NOTE]
-> The [`name`](/en-US/docs/Web/HTML/Reference/Elements/meta/name) attribute has a specific meaning for the `<meta>` element. The [`itemprop`](/en-US/docs/Web/HTML/Reference/Global_attributes/itemprop) attribute must not be set on a `<meta>` element that includes a [`name`](/en-US/docs/Web/HTML/Reference/Elements/meta/name), [`http-equiv`](#http-equiv), or [`charset`](#charset) attribute.
+> The [`name`](#name) attribute has a specific meaning for the `<meta>` element.
+> The [`itemprop`](/en-US/docs/Web/HTML/Reference/Global_attributes/itemprop) attribute must not be set on a `<meta>` element that includes a [`name`](#name), [`http-equiv`](#http-equiv), or [`charset`](#charset) attribute.
```
Josh-Cena: I would appreciate if it looks like this

```suggestion
> The [`name`](/en-US/docs/Web/HTML/Reference/Elements/meta/name) attribute has a specific meaning for the `<meta>` element.
> The [`itemprop`](/en-US/docs/Web/HTML/Reference/Global_attributes/itemprop) attribute must not be set on a `<meta>` element that includes a [`name`](/en-US/docs/Web/HTML/Reference/Elements/meta/name), [`http-equiv`](/en-US/docs/Web/HTML/Reference/Elements/meta/http-equiv), or [`charset`](#charset) attribute.
```

files/en-us/web/html/reference/elements/meta/http-equiv/index.md
bsmth: I like it, thanks, we mention in the next sentence that it's document-level

files/en-us/web/html/reference/elements/meta/http-equiv/index.md
bsmth: I think it's a little looser, so the HTML spec has some values with defined behavior, but I think it's possible to simulate many others.

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-DNS-Prefetch-Control#turning_on_and_off_prefetching

Reporting and even client hints:

```html
<meta
      http-equiv="Reporting-Endpoints"
      content="csp-endpoint='https://example.com/csp-reports'"  …

<meta http-equiv="Accept-CH"…
```

I think implementations are really flex with what they will process, it may be possible to set nearly everything in this way.

files/en-us/web/html/reference/elements/meta/index.md
bsmth: Sure 👍🏻

files/en-us/web/html/reference/elements/meta/http-equiv/index.md
bsmth: But we should be careful, though:

https://almanac.httparchive.org/en/2022/security#preventing-attacks-using-meta

And https://www.htmhell.dev/adventcalendar/2023/7/

>One of the issues with adding security headers using <meta> tag is developers often assume that any of the HTTP security response headers can be added through this method. This is not true and can often lead to a false sense of security since any invalid security header will be ignored by the browser. It is important to remember that even though http-equiv helps in defining HTTP response headers, it is an [enumerated attribute](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute). Hence, it can only help in delivering a fixed set of [HTTP header](https://html.spec.whatwg.org/multipage/semantics.html#pragma-directives) and any other HTTP header that is set using this method will be ignored.

hamishwillee: @bsmth Thanks - interesting!

Agreed, that does make it relevant to mention the specification rather than "allowed headers".
But we should add a note capturing that browsers may offer non-standard support additional headers, and the risks associated with using these rather than the actual headers.

I also still agree with my comment here:

> The attribute's name is http-equiv because the allowed values are names of equivalent HTTP headers - the "short for" thing doesn't follow. I'd mention the "short for" up in the top line of the description.

files/en-us/web/html/reference/elements/meta/http-equiv/index.md
```diff
@@ -0,0 +1,107 @@
...
+page-type: html-attribute
+browser-compat: html.elements.meta.http-equiv
+---
+
+{{HTMLSidebar}}
+
+The **`http-equiv`** attribute of the {{htmlelement("meta")}} element allows you to provide processing instructions for the browser that could otherwise be given in an HTTP header.
+The metadata is _document-level metadata_ that applies to the whole page.
+
+When a `<meta>` element has a `http-equiv` attribute, a [`content`](/en-US/docs/Web/HTML/Reference/Attributes/content) attribute defines the corresponding `http-equiv` value.
```
Josh-Cena:
```suggestion
When a `<meta>` element has an `http-equiv` attribute, a [`content`](/en-US/docs/Web/HTML/Reference/Attributes/content) attribute defines the corresponding `http-equiv` value.
```

files/en-us/web/html/reference/elements/meta/index.md
bsmth:
```suggestion
  - : The `http-equiv` (short for 'HTTP equivalent') attribute tells the browser to behave as if the response that returned the page included certain HTTP headers.
  The values you can provide match the names of HTTP headers such as {{httpheader("Refresh")}} and {{httpheader("Content-Security-Policy")}}.
```

I had another go at this, what do you think?

files/en-us/web/html/reference/elements/meta/http-equiv/index.md
```diff
@@ -0,0 +1,107 @@
...
+title: <meta> http-equiv attribute
+short-title: <meta> http-equiv
+slug: Web/HTML/Reference/Elements/meta/http-equiv
+page-type: html-attribute
+browser-compat: html.elements.meta.http-equiv
+---
+
+{{HTMLSidebar}}
+
+The **`http-equiv`** attribute of the {{htmlelement("meta")}} element allows you to provide processing instructions for the browser that could otherwise be given in an HTTP header.
```
bsmth:
```suggestion
The **`http-equiv`** attribute of the {{htmlelement("meta")}} element allows you to provide processing instructions for the browser as if the response that returned the document included certain HTTP headers.
```

files/en-us/web/html/reference/elements/meta/http-equiv/index.md
```diff
@@ -0,0 +1,114 @@
...
+- `x-ua-compatible` {{deprecated_inline}}
+  - : Used by legacy versions of the now-retired {{glossary("Microsoft Internet Explorer")}} so that it more closely followed specified behavior.
+    If specified, the `content` attribute must have the value `"IE=edge"`.
+    User agents now ignore this pragma.
+    The name derives from the `X-UA-Compatible` HTTP header.
+
+> [!WARNING]
+> Unrecognized headers or invalid values are ignored by the browser.
+> **Do not set other security headers** using `<meta http-equiv=`, as this can lead to a false sense of security.
+> Some browsers may process headers that are not listed above, but this behavior varies across implementations.
```
hamishwillee: It's a nit, but at the point you get to the second line you don't know that you can process other headers (third line), so it doesn't make sense.

Shifting a little

```suggestion
> Some browsers process additional headers that are not listed above.
> Since unrecognized headers or invalid values are ignored, this can lead to inconsistent behavior across browser implementations.
> In particular, **Do not set other security headers** using `<meta http-equiv=`, as this can lead to a false sense of security!
```

hamishwillee (approved):
Thanks @bsmth . I fixed a nit to my taste and this looks good to merge.

PR #40050, opened 2025-06-23
Editorial review: Add details of Chrome support for H265 in WebRTC

chrisdavidmills:
### Description

Chrome 136+ supports the HEVC/H265 codec in WebRTC; see https://chromestatus.com/feature/5153479456456704.

This PR adds details of this to https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/WebRTC_codecs.

I searched around MDN, but didn't really see anywhere else where I should mention this.


hamishwillee (changes requested):
@chrisdavidmills This is exactly the right place to update. One concern though. See https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Video_codecs#hevc_h.265

image

I don't think we should just change that to "Yes" without some research as to whether this is a typo or something special is going on with Chrome.

chrisdavidmills:
> @chrisdavidmills This is exactly the right place to update. One concern though. See https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Video_codecs#hevc_h.265
> 
> image
> 
> I don't think we should just change that to "Yes" without some research as to whether this is a typo or something special is going on with Chrome.

Ah, I missed this table entry. Doesn't that just mean that support for it being used in WebRTC wasn't there before, but now it is so that we can toggle it to "yes"? I don't know what kind of special circumstances you might be referring to, but then again, I know very little about WebRTC.

I've asked the Chrome folks to look at this one in case of any specific information I need to include.

hamishwillee:
> Doesn't that just mean that support for it being used in WebRTC wasn't there before, but now it is so that we can toggle it to "yes"? I do

It doesn't say "works with" it says "compatible with". So asking the Chrome people what it takes to make a format  compatible or incompatible with WebRTC and if there is such a distinction, how they got around it in this case is the right thing to do.

I suspect it does mean compatible with, but if that proves wrong we can change the docs. I know someone we can ask in FF land if you need me to.

hamishwillee:
@chrisdavidmills FYI this one fell off both our radars. Is it still relevant? I think it is waiting on you.

chrisdavidmills:
> @chrisdavidmills FYI this one fell off both our radars. Is it still relevant? I think it is waiting on you.

@hamishwillee I asked the Chrome folks, and no one responded. So we are stuck in limbo. Can you ask your contacts about this? I'd still be inclined to just change it to "Yes", as the left column says "compatible".

hamishwillee:
@chrisdavidmills I can't remember who my contact was but I've done a little more research on this.

WebRTC defines a baseline codec set, which defines the set of recommended codecs. These were (anecdotally) selected to be patent/royalty free, and did not include H265. I don't believe any codec was specifically disallowed.
So "compatible" in the table likely meant some combination of "not recommended by the spec" and "no one supports this".

The logical way to address this is to add a compatibility table (values obviously wrong). 

image

chrisdavidmills:
> The logical way to address this is to add a compatibility table (values obviously wrong).

@hamishwillee OK, I've researched support values and added a compat table as requested.

hamishwillee (approved):
Thanks @chrisdavidmills - looks great.

PR #40056, opened 2025-06-24
chore: replace old note card style with GFM alerts syntax

yin1999:
### Description

With the [Prettier's bug fixed](https://prettier.io/blog/2025/06/23/3.6.0#change-16596), we're cleared to use GFM alerts everywhere in the docs!

Note: I've also removed the following lines in `files/en-us/mdn/writing_guidelines/howto/markdown_in_mdn/index.md`:

```md
> [!WARNING]
> Currently, due to a [Prettier bug](https://github.com/prettier/prettier/issues/15479), the GFM alert syntax cannot be used if the first character of a note or warning is a formatting symbol, such as a backquote, asterisk, square bracket or curly bracket. In this case, use the old syntax `> **Note:**` instead. Writers are not required to rephrase the content to work around the formatter.

<!-- another part -->

> [!WARNING]
> In the older MDN syntax, the type was localized and added to the first paragraph in bold text, i.e., `**Note:** Foo bar` instead of `[!NOTE] ⏎ Foo bar`.
>
> The older syntax is still supported for migration purposes. Avoid using it in new documentation.
```

Related PR: #40048


Josh-Cena (approved):
Woohooo!

Issue #1866, opened 2021-01-28
Issue with "Separate sites for mobile and desktop": … "Responsive Design" link at the bottom does not link to the correct page.


jacobtrebil:
MDN URL: https://developer.mozilla.org/en-US/docs/Web/Guide/Mobile/Separate_sites

#### Did you test this? If so, how?

<details>
<summary>MDN Content page report details</summary>

* Folder: `en-us/web/guide/mobile/separate_sites`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/Guide/Mobile/Separate_sites
* GitHub URL: https://github.com/mdn/content/blob/master/files/en-us/web/guide/mobile/separate_sites/index.html
* Last commit: https://github.com/mdn/content/commit/452985968235edef6a702b5d5c9c35edab9b6265
* Document last modified: 2021-01-21T16:21:14.000Z

</details>

chrisdavidmills:
This was kinda strange; thanks for reporting it, @jacobtrebil !

We now have a pretty good beginner's guide to responsive design, so I've linked it to that instead. See https://github.com/mdn/content/commit/4ad95e38da46ff56f0dcc42477efa11d253b0569 for the change.

PR #40065, opened 2025-06-24
Clarify script-closable conditions for Window.close() (Fixes #39154)

### LINKED ISSUE #39154: window/close: clarify when a window is opened by script
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Window/close

### Issue

Browsers allow windows to be script-closable when they were opened by web content rather than script and we are working on updating the spec (https://github.com/whatwg/html/pull/11208). MDN currently states they are script-closable when opened using `window.open()`, but link navigations or form submissions can also create script-closable windows.

Windows that are not opened by web content include those opened using the browser UI, as well as `Ctrl+Click`, `Shift+Click` or middle clicks on links. This is not part of the HTML spec, but I tested it on Chrome, Firefox and Safari.

In addition, the session history check might not be too relevant to web developers (https://github.com/whatwg/html/issues/1866#issuecomment-2788880617), we could put less emphasis on it. For me, this check was difficult to understand given there is no reference for `session history`. We could point to [History: length](https://developer.mozilla.org/en-US/docs/Web/API/History/length), I believe it's the same concept. Though the pages on `History` also don't make it completely clear what falls under session history. For example, navigations of embedded iframes will increase the history length of the top level page.

`window.close()` can also be blocked through an [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) handler.

### See also

https://github.com/whatwg/html/pull/11208
https://github.com/whatwg/html/issues/1866
https://bugzilla.mozilla.org/show_bug.cgi?id=1316857

Shrinivassab:
### Description

Added detailed information and examples for the `window.close()` method, clarifying its behavior when called from different contexts (e.g., opened vs. not opened windows).

### Motivation

The previous documentation was unclear about when `window.close()` is allowed to close the current tab or window. These changes help developers understand browser restrictions and avoid unexpected behavior in their applications.

### Additional details

Updated with references to the latest HTML Living Standard and linked to relevant browser compatibility notes.

Fix #39154


files/en-us/web/api/window/close/index.md
```diff
@@ -55,6 +62,16 @@ function closeOpenedWindow() {
 }
 ```
 
+## Script-closable behavior by context
```
Josh-Cena: This duplicates your explanation above and not really helpful for me.

files/en-us/web/api/window/close/index.md
```diff
@@ -11,15 +11,22 @@ browser-compat: api.Window.close
...
+Windows opened by browser UI actions — such as right-click → Open in new tab, Ctrl+Click, Shift+Click, or middle-click — are not script-closable. Calling `close()` on such windows will result in an error message like:
+
+> Scripts may not close windows that were not opened by script.
 
 Note also that `close()` has no effect when called on {{domxref("Window")}}
 objects returned by
 [`HTMLIFrameElement.contentWindow`](/en-US/docs/Web/API/HTMLIFrameElement/contentWindow).
 
+> [!NOTE]
+> Some documentation and specifications mention session history as a factor (e.g., top-level windows with only one history entry). However, this can be confusing and inconsistent across browsers. You can inspect `window.history.length` to understand the session state, but this value may also be affected by iframe activity or the new tab page.
```
Josh-Cena: This means nothing to me; what kind of inconsistency is there? What specifications demand this? Be more specific and helpful

files/en-us/web/api/window/close/index.md
Shrinivassab: Sure, Will remove it, in next commit. Thanks for the comment.

files/en-us/web/api/window/close/index.md
Shrinivassab: Thanks for the feedback!

Updated Note: - Please share your thought? 
"Although it's sometimes claimed that script-closable windows have a session history length of 1, this is not a reliable or standardized rule. Browser behavior varies, and the [WHATWG spec discussion](https://github.com/whatwg/html/issues/1866) notes the confusion around this. In most cases, script-closability depends more on how the window was opened (e.g., via `window.open()` or web-initiated link), rather than its history."

The note was originally meant to address the mention of “top-level windows with a single history entry,” which existed in the previous version of the docs. There’s no strong spec text behind it, and it leads to confusion, especially since history.length can vary between browsers and is influenced by things like iframe navigation.

To clarify this, I have now removed vague wording and referenced [WHATWG issue #1866](https://github.com/whatwg/html/issues/1866), which discusses that this rule isn’t reliable. The updated note now focuses more clearly on the practical developer takeaway: script-closability depends more on how the window was opened, not the session history.

If you feel the revised note still adds confusion, I am happy to remove it entirely, I only retained it to preserve and clarify the previous statement.

Josh-Cena: > The note was originally meant to address the mention of “top-level windows with a single history entry,” which existed in the previous version of the docs. There’s no strong spec text behind it, and it leads to confusion, especially since history.length can vary between browsers and is influenced by things like iframe navigation.

If there's no spec backing it, it sounds more like a documentation bug or outdated non-standard implementation. It would be great if you can collect evidence that it did function that way, but otherwise I wouldn't worry about just removing any mentions of this.

Shrinivassab: Done. Remote Note block and pushed the changes. 
Screenshot 2025-06-25 180628
Thanks for your feedback.

Josh-Cena:
@vinhill Does this look good to you?

files/en-us/web/api/window/close/index.md
```diff
@@ -11,10 +11,12 @@ browser-compat: api.Window.close
...
-This method can only be called on windows that were opened by a script using the
-{{domxref("Window.open()")}} method, or on top-level windows that have a single history entry. If the window doesn't match these requirements, an error
-similar to this one appears in the console:
-`Scripts may not close windows that were not opened by script.`
+This method can only be successfully called on windows that are considered _script-closable_. This generally includes:
+
+- Windows opened using {{domxref("Window.open()")}}
+- Windows opened via web content, such as links (`<a target="_blank">`) or forms (`<form target="_blank">`), without user modifier actions
+
+Windows opened by browser UI actions — such as right-click → Open in new tab, Ctrl+Click, Shift+Click, or middle-click — are not script-closable. Calling `close()` on such windows will result in an error message like: `Scripts may not close windows that were not opened by script.`
```
vinhill:
```suggestion
Windows opened by browser UI actions — such as right-click → Open in new tab, Ctrl+Click, Shift+Click, or middle-click — are not script-closable. Calling `close()` on such windows might result in a console warning like: `Scripts may not close windows that were not opened by script.`
```
The spec doesn't mandate any indication of a failure. From a quick test, Chromium, Safari and Firefox show differently worded console warnings.

files/en-us/web/api/window/close/index.md
```diff
@@ -11,10 +11,12 @@ browser-compat: api.Window.close
 The **`Window.close()`** method closes the current window, or
 the window on which it was called.
 
-This method can only be called on windows that were opened by a script using the
-{{domxref("Window.open()")}} method, or on top-level windows that have a single history entry. If the window doesn't match these requirements, an error
-similar to this one appears in the console:
-`Scripts may not close windows that were not opened by script.`
+This method can only be successfully called on windows that are considered _script-closable_. This generally includes:
```
vinhill:
```suggestion
Windows are _script-closable_ if they were created by web content. This generally includes:
```
I would prefer MDN to use the phrase [created-by-web-content](https://html.spec.whatwg.org/#is-created-by-web-content) given this is what browser code and the specification uses.

Josh-Cena: I'm slightly confused because isn't "created by web content" one of the sufficient conditions of "script-closable"? (Granted I don't know what "session history entries" entails if not `history.length`)

files/en-us/web/api/window/close/index.md
vinhill: > Windows opened by browser UI actions [...] are not script-closable

Unfortunately, it depends :).

If `history.length <= 1`, the window is script-closable even if not created by web content. In addition, these browser UI actions are not part of the specification and whether such windows are script-closable is thus up to implementers. I.e.
- right-click -> open in new tab -> window.close() works
- right-click -> open in new tab -> navigate once -> window.close() likely fails.

I think this goes into details that users are generally not interested in. But we could state something like
> Windows opened by browser UI actions ... are often considered not opened by web content. They are only script closable as long as their history length is one.

files/en-us/web/api/window/close/index.md
vinhill: You're right. I just think "created by web content" is the important condition and a brief summary for the two items mentioned. The history check is rarely relevant.

Per my understanding, `session history entry size` and `history.length` is pretty much equal.

files/en-us/web/api/window/close/index.md
Rajkrithika: Thanks for your comment. Just to confirm, should we note that Firefox/Chrome/Safari warnings differ, or is the generalized ‘might’ sufficient?

Shrinivassab: Thanks for your comment. Just to confirm, should we note that Firefox/Chrome/Safari warnings differ, or is the generalized ‘might’ sufficient?

files/en-us/web/api/window/close/index.md
Shrinivassab: @vinhill @Josh-Cena Thanks for the discussion. I have updated the text to use "created by web content" per the specification language.

files/en-us/web/api/window/close/index.md
Shrinivassab: @vinhill Thanks for catching this edge case. What do you think about this updated text:

"Windows opened by browser UI actions (like right-click → new tab) are often not script-closable. They may only be closed if they haven not been navigated (history length remains 1). Calling `close()` otherwise typically shows a console warning."

This keeps the common case clear while acknowledging the exception. Please, share your thought.

Shrinivassab:
Thank you @vinhill  for your suggestion. I have incorporated the previous feedback. Please let me know if there's anything else that needs to be updated or improved regarding this issue.

files/en-us/web/api/window/close/index.md
```diff
@@ -11,10 +11,12 @@ browser-compat: api.Window.close
...
-This method can only be called on windows that were opened by a script using the
-{{domxref("Window.open()")}} method, or on top-level windows that have a single history entry. If the window doesn't match these requirements, an error
-similar to this one appears in the console:
-`Scripts may not close windows that were not opened by script.`
+Windows are _script-closable_ if they were created by web content. This generally includes:
+
+- Windows opened using {{domxref("Window.open()")}}
+- Windows opened via web content, such as links (`<a target="_blank">`) or forms (`<form target="_blank">`), without user modifier actions
+
+Windows opened by browser UI actions (like right-click → new tab) are often not script-closable. They may only be closed if they have not been navigated (history length remains 1). Calling `close()` otherwise typically shows a console warning.
```
Josh-Cena: Why remove other actions such as Ctrl+Click? I found them to be helpful. Otherwise this LGTM.

Shrinivassab: @Josh-Cena Thanks for the review. I thought as @vinhill approved the current wording, I kept it generic since "browser UI actions" already encompasses all modifier-based interactions. But if you prefer I can add `Ctrl+Click` also. 

`Windows opened by browser UI actions (like right-click → new tab, Ctrl+Click, Shift+Click) are often not script-closable. They may only be closed if they have not been navigated (history length remains 1). Calling `close()` otherwise typically shows a console warning.`

Please share your thought? Thanks in advance...

Josh-Cena: I would prefer the previous more exhaustive examples: `— such as right-click → Open in new tab, Ctrl+Click, Shift+Click, or middle-click —`

Shrinivassab: Sure. Will update it based on your suggestion.

vinhill:
Thanks, looks good to me!

Josh-Cena (approved):
Thanks! LGTM from an editorial perspective.

PR #40080, opened 2025-06-25
New glossary for invariant

Josh-Cena:
I'm constantly reminded that "invariant" is a jargon that not everyone understands. Perhaps a glossary entry would help.


bsmth (approved):
brilliant addition, well done 👍🏻

PR #40081, opened 2025-06-25
Editorial review: Add information on missing MP4/HEVC codec parameter values

chrisdavidmills:
### Description

Chrome 136 adds support to the MediaRecorder API for HEVC encoding, introducing the hvc1.* codec string, and adds new codecs (hev1.* and avc3.*) supporting variable resolution video in MP4. See https://chromestatus.com/feature/6375884229181440 for more details.

This PR attempts to add some information about the new codecs to the https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/codecs_parameter page. The relevant MediaRecorder features all point to this page for the actual codec details, but to make sure the support is clear, I've also added compat data entries for them: https://github.com/mdn/browser-compat-data/pull/27142.


hamishwillee (commented):
Ping me when this is done. FWIW, I like where it is heading.

chrisdavidmills:
> Ping me when this is done. FWIW, I like where it is heading.

@hamishwillee OK, this one has had a technical review and been approved (over email, rather than on GitHub). It is ready for you to look at now. Cheers mate!

files/en-us/web/media/guides/formats/codecs_parameter/index.md
```diff
@@ -898,6 +902,50 @@ The Audio Object Types are defined in ISO/IEC 14496-3 subpart 1, section 1.5.1.
   </tbody>
 </table>
 
+### HEVC: MP4, Quicktime, Matroska
+
+The [High Efficiency Video Coding](https://en.wikipedia.org/wiki/High_Efficiency_Video_Coding) format, also known as H.265 and MPEG-H Part 2, can be included in the [MP4](/en-US/docs/Web/Media/Guides/Formats/Containers#mpeg-4_mp4) (`video/mp4`), [Quicktime](/en-US/docs/Web/Media/Guides/Formats/Containers#quicktime) (`video/quicktime`), and [Matroska](https://en.wikipedia.org/wiki/Matroska) (`video/matroska`) containers.
```
hamishwillee: Should it be "codec"? 

```suggestion
The [High Efficiency Video Coding](https://en.wikipedia.org/wiki/High_Efficiency_Video_Coding) codec, also known as H.265 and MPEG-H Part 2, can be included in the [MP4](/en-US/docs/Web/Media/Guides/Formats/Containers#mpeg-4_mp4) (`video/mp4`), [Quicktime](/en-US/docs/Web/Media/Guides/Formats/Containers#quicktime) (`video/quicktime`), and [Matroska](https://en.wikipedia.org/wiki/Matroska) (`video/matroska`) containers.
```

hamishwillee (commented):
@chrisdavidmills Thanks for this - looks great. 

1. I think we need some cross linking to https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Video_codecs#hevc_h.265. You could do this instead of the wikipedia link - pushing that to a see also.

2. The BCD update is also relevant. That probably should go under each section in https://developer.mozilla.org//en-US/docs/Web/Media/Guides/Formats/Video_codecs and not this. Perhaps a problem for another day?

chrisdavidmills:
> @chrisdavidmills Thanks for this - looks great.
> 
>     1. I think we need some cross linking to https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Video_codecs#hevc_h.265. You could do this instead of the wikipedia link - pushing that to a see also.

Agreed. I've updated it just as you suggested.

>     2. The BCD update is also relevant. That probably should go under each section in https://developer.mozilla.org//en-US/docs/Web/Media/Guides/Formats/Video_codecs and not this. Perhaps a problem for another day?

The whole "how to represent media support" conversation is a hot one on BCD right now, and I don't think a conclusion has been reached yet. It would be better to leave that for now, IMO.

hamishwillee (approved):
It would be great if we could also have a link back from https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Video_codecs#hevc_h.265. There is a link to the container section from this doc which then links to this section, but it is a bit of a round tip.

But I don't want to block on that so merging and up to you if you create that link in a post process. Thanks!

PR #40082, opened 2025-06-25
Fix Summarizer.format enum description

Spacefish:
Description of the enum values was mixed arround


chrisdavidmills (approved):
Good spot, thanks for the fix, @Spacefish!

PR #40088, opened 2025-06-25
fix(css/display): clarify box suppression comment

### LINKED ISSUE #40076: Ambiguous comment for display: none and display: contents
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/display

Syntax

In the "Values" section of the display property documentation, the values are grouped by category using CSS comments. The comment for the group containing display: none and display: contents is potentially misleading.

/* box generation */
display: none;
display: contents;

This comment, /* box generation */, seems to suggest that these values are for generating a box, when in fact their primary purpose is to suppress or prevent the element's principal box from being generated. This can cause confusion for learners.

The comment should be changed to more accurately reflect the function of display: none and display: contents.

Suggestion 1: /* Not to generate box */
Suggestion 2: /* Box Suppression */

Either of these suggestions would provide more clarity than the current comment. 

_No response_

_No response_

Shrinivassab:
### Description

Updated a comment in the CSS `display` property syntax section from `/* box generation */` to `/* Box suppression */` to align with CSS specification terminology.

### Motivation

This change improves accuracy and clarity by reflecting the intended behavior of `display: none` and `display: contents`, which suppress box generation rather than initiate it. This aligns the documentation with current CSS specs and helps developers better understand box model behavior.

### Additional details
See [CSS Display Module Level 3](https://drafts.csswg.org/css-display/#box-suppress) for reference on "box suppression".

Fixes #40076

### DIFF
--- a/files/en-us/web/css/display/index.md
+++ b/files/en-us/web/css/display/index.md
@@ -87,7 +87,7 @@ display: grid;
 display: inline-grid;
 display: flow-root;
 
-/* box generation */
+/* Box suppression */
 display: none;
 display: contents;
 


Josh-Cena (approved):
Thanks

PR #40089, opened 2025-06-26
Additions to "Content with images must be labelled"

mrec:
### Description

Mention that `canvas` needs `aria-label` rather than `alt`, and mention the use of `alt=""` to signal that an image is purely decorative.

### Motivation

I'm new to `canvas` and was confused when  the accessibility checker complained but the linked section only mentioned `alt`.

The existing text mentions that a label is needed for contentful images, but doesn't say how to stop the checker complaining for images that aren't contentful.

### Additional details

N/A

N/A

### DIFF
--- a/files/en-us/web/accessibility/guides/understanding_wcag/text_labels_and_names/index.md
+++ b/files/en-us/web/accessibility/guides/understanding_wcag/text_labels_and_names/index.md
@@ -228,7 +228,7 @@ As a best practice, also provide a {{htmlelement("title")}} for the document tha
 
 ## Content with images must be labeled
 
-Provide descriptive text for all contentful (that is, non-decorative) images and image-like elements. This includes SVG images, {{htmlelement("img")}}, {{htmlelement("canvas")}}, {{htmlelement("map")}}, and {{htmlelement("area")}} elements, as well as {{htmlelement("input")}} elements where `type=image` and {{htmlelement("object")}} elements where the `type` starts with `image/`. The typical way to do this is with the `alt` attribute, but for the {{htmlelement("canvas")}} element the `aria-label` attribute is needed instead. Be sure that the description conveys what is shown in the image.
+Provide descriptive text for all contentful (that is, non-decorative) images and image-like elements. This includes SVG images, {{htmlelement("img")}}, {{htmlelement("canvas")}}, {{htmlelement("map")}}, and {{htmlelement("area")}} elements, as well as {{htmlelement("input")}} elements where `type=image` and {{htmlelement("object")}} elements where the `type` starts with `image/`. The typical way to do this is with the `alt` attribute, but for elements that don't allow the `alt` attribute, such as {{htmlelement("canvas")}}, use [`role="img"`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/img_role) and [`aria-label`](/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) instead. Be sure that the description conveys what is shown in the image.
 
 For `alt` attributes on images which _are_ purely decorative, an empty value can be used to signal to accessibility tools that the element should be ignored.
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/accessibility/guides/understanding_wcag/text_labels_and_names/index.md
```diff
@@ -228,7 +228,9 @@ As a best practice, also provide a {{htmlelement("title")}} for the document tha
 
 ## Content with images must be labeled
 
-Provide descriptive text for all contentful (that is, non-decorative) images and image-like elements. This includes SVG images, {{htmlelement("img")}}, {{htmlelement("canvas")}}, {{htmlelement("map")}}, and {{htmlelement("area")}} elements, as well as {{htmlelement("input")}} elements where `type=image` and {{htmlelement("object")}} elements where the `type` starts with `image/`. The typical way to do this is with the `alt` attribute. Be sure that the description conveys what is shown in the image.
+Provide descriptive text for all contentful (that is, non-decorative) images and image-like elements. This includes SVG images, {{htmlelement("img")}}, {{htmlelement("canvas")}}, {{htmlelement("map")}}, and {{htmlelement("area")}} elements, as well as {{htmlelement("input")}} elements where `type=image` and {{htmlelement("object")}} elements where the `type` starts with `image/`. The typical way to do this is with the `alt` attribute, but for the {{htmlelement("canvas")}} element the `aria-label` attribute is needed instead. Be sure that the description conveys what is shown in the image.
```
bsmth: Seems like `role="img"` is also recommended along with `aria-label`, at least here:

* https://pauljadam.com/demos/canvas.html

Although not in other places:

* https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/canvas#alternative_content_2

mrec: Not sure about this. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/img_role describes the `img` role as identifying **multiple** elements that should be presented as a single image; that doesn't really apply here.

ARIA themselves still don't seem entirely sure; they're discussing it but without any great urgency: https://github.com/w3c/aria/issues/927

Josh-Cena:
@scottaohara I believe this is still waiting on your review. Thanks

files/en-us/web/accessibility/guides/understanding_wcag/text_labels_and_names/index.md
scottaohara: this isn't just `canvas` though - `map`, and `object` also don't use `alt`.

Maybe instead it's mentioned that for the HTML elements which don't allow the `alt` attribute, ARIA attributes such as `aria-label` paired with a `role` that allows naming, would be the best way to ensure these other elements can be given alternative text?

scottaohara (changes requested):

files/en-us/web/accessibility/guides/understanding_wcag/text_labels_and_names/index.md
Josh-Cena:
```suggestion
Provide descriptive text for all contentful (that is, non-decorative) images and image-like elements. This includes SVG images, {{htmlelement("img")}}, {{htmlelement("canvas")}}, {{htmlelement("map")}}, and {{htmlelement("area")}} elements, as well as {{htmlelement("input")}} elements where `type=image` and {{htmlelement("object")}} elements where the `type` starts with `image/`. The typical way to do this is with the `alt` attribute, but for elements that don't allow the `alt` attribute, such as {{htmlelement("canvas")}}, use [`role="img"`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/img_role) and [`aria-label`](/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) instead. Be sure that the description conveys what is shown in the image.
```

scottaohara (approved):

PR #40092, opened 2025-06-26
Ensure that syntax sections start with a code block

Josh-Cena:
My checker parses all code blocks and lints them. However, most syntax blocks are not actually meant to be linted. To work around this, I added a rule that says "if the syntax section immediately starts with a code block, treat that code as pseudocode and do not lint it". It works really well, and there are only a few outliers where there's preceding text.

I checked through them, and found out that most of them is better off removing the introductory text. This results in a terser syntax section, and overall adheres to the general editorial conventions better.

### DIFF
--- a/files/en-us/web/http/reference/headers/sec-websocket-protocol/index.md
+++ b/files/en-us/web/http/reference/headers/sec-websocket-protocol/index.md
@@ -46,7 +46,7 @@ Sec-WebSocket-Protocol: <sub-protocols>
   - : A comma-separated list of sub-protocol names, in the order of preference.
     The sub-protocols may be selected from the [IANA WebSocket Subprotocol Name Registry](https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name), or may be a custom name jointly understood by the client and the server.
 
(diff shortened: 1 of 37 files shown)


files/en-us/web/http/reference/headers/sec-websocket-protocol/index.md
```diff
@@ -36,24 +36,18 @@ The sub-protocol selected by the server is made available to the web application
...
-Sec-WebSocket-Protocol: <selected-sub-protocol>
-```
-
 ## Directives
 
 - `<sub-protocols>`
   - : A comma-separated list of sub-protocol names, in the order of preference.
     The sub-protocols may be selected from the [IANA WebSocket Subprotocol Name Registry](https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name), or may be a custom name jointly understood by the client and the server.
 
+    As a response header, this should be a single sub-protocol selected by the server.
```
bsmth:
```suggestion
    As a response header, this is a single sub-protocol that the server selected.
```

bsmth (approved):
Looking good, tnx 👍🏻

PR #40097, opened 2025-06-26
Fix: remove incorrect "form" attribute

sb3nder:
### Description

Removed the incorrect `form` attribute on the `<meter>` element.

### Motivation

The `form` attribute on `<meter>` is not not mentioned in the spec and, to my knowledge, is not supported in any current browser implementation.

[HTML Living Standard – meter element](https://html.spec.whatwg.org/multipage/form-elements.html#the-meter-element)

### Additional details

The only references to the `form` attribute on `<meter>` I could find are from W3C:

[www.w3.org | 4.10.17 The meter element — HTML5: Edition for Web Authors](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-meter-element.html)

The page list `form` and says:

> The form attribute is used to explicitly associate the meter element with its form owner.

But it does not list `name`.

On the w3schools website there is an example that uses both `form` and `name`:

`<meter id="anna" form="form1" name="anna" min="0" low="40" high="90" max="100" value="95"></meter>`

[www.w3schools.com | HTML meter form Attribute](https://www.w3schools.com/tags/att_meter_form.asp)

The same page indicates the form attribute is **Not supported** in all browsers.

My guess is this originated from a typo or misinterpretation in earlier W3C drafts.


estelle (approved):
Thank you. IIRC, the meter value is not submitted with the form, so the form attribute wouldn't be necessary. It was included in the spec about years ago, but this page wasn't updated, so thank you.

Congratulations on your first merged MDN content PR. Welcome to the team!

PR #40100, opened 2025-06-27
fix(text/javascript): Clarify charset validity in HTTP vs HTML contex…

### LINKED ISSUE #39505: content/files/en-us/web/http/guides/mime_types/index.md
https://github.com/mdn/content/blob/main/files/en-us/web/http/guides/mime_types/index.md, rendered as https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types, is incorrect in part [textjavascript](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types#textjavascript):

"Per the [IANA Media Types registry](https://www.iana.org/assignments/media-types/media-types.xhtml#text), [RFC 9239](https://www.rfc-editor.org/rfc/rfc9239.html), and the [HTML specification](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:text/javascript), JavaScript content should always be served using the MIME type text/javascript. No other MIME types are considered valid for JavaScript, and using any MIME type other than text/javascript may result in scripts that do not load or run.

You may find some JavaScript content incorrectly served with a charset parameter as part of the MIME type — as an attempt to specify the character set for the script content. That charset parameter isn't valid for JavaScript content, and in most cases will result in a script failing to load."

The three links above (in the respective order):
- [IANA Media Types registry](https://www.iana.org/assignments/media-types/media-types.xhtml#text) > the only occurrence of `text/javascript` links to https://www.iana.org/assignments/media-types/text/javascript, which reads: "Optional parameters:  charset."
- [RFC 9239](https://www.rfc-editor.org/rfc/rfc9239.html) > "The charset parameter is only **used when processing a Script** goal source; Module goal sources MUST always be processed as UTF-8." (bold formatting is mine)
- [HTML specification](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:text/javascript) > (the rest of this list item is one consecutive quote) > For external JavaScript resources, MIME type parameters in `Content-Type` headers are generally ignored. (In some cases the `charset` parameter has an effect.) However, for the script element's type attribute they are significant; it uses the JavaScript MIME type essence match concept.
  
  For example, scripts with their **type attribute** set to "text/javascript; charset=utf-8" will not be evaluated, even though that is a valid JavaScript MIME type when parsed.
- that means "type attribute" of a `<script type="...">` tag in HTML source. Specifically, "their [type](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-type)" links to HTML definition of `type` attribute of `<script>` element. NOT MIME as servbed by web server (or as specified in `data:` URLs).

I'm willing to provide a pull request.

Shrinivassab:
### Description
Updated text/javascript MIME type docs to clarify that charset is valid in HTTP headers.

### Motivation
Fixes incorrect documentation that claimed charset is always invalid. 

Fixes #39505

### DIFF
--- a/files/en-us/web/http/guides/mime_types/index.md
+++ b/files/en-us/web/http/guides/mime_types/index.md
@@ -117,12 +117,6 @@ There are two multipart types:
 
 ## Important MIME types for Web developers
 
-### MIME type parameters in HTML
-
-In HTML, the `type` attribute for elements like `<script>` and `<style>` only supports the **MIME type essence** (e.g., `text/javascript`). Including parameters such as `charset=utf-8` will cause the resource to be ignored.
-
-In contrast, parameters in HTTP `Content-Type` headers (e.g., `Content-Type: text/javascript; charset=utf-8`) are technically allowed. However, modern browsers assume UTF-8 by default and typically ignore the parameter, making it unnecessary in most cases.
-
 ### application/octet-stream
 
 This is the default for binary files. As it means _unknown binary_ file, browsers usually don't execute it, or even ask if it should be executed. They treat it as if the {{HTTPHeader("Content-Disposition")}} header was set to `attachment`, and propose a "Save As" dialog.
@@ -152,11 +146,16 @@ All HTML content should be served with this type. Alternative MIME types for XHT
 
 ### text/javascript
 
-Per the [IANA Media Types registry](https://www.iana.org/assignments/media-types/media-types.xhtml#text), [RFC 9239](https://www.rfc-editor.org/rfc/rfc9239.html), and the [HTML specification](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:text/javascript), JavaScript content should always be served using the MIME type `text/javascript`.
-No other MIME types are considered valid for JavaScript, and using any MIME type other than `text/javascript` may result in scripts that do not load or run.
+JavaScript content should always be served with the MIME type `text/javascript`.
+No other MIME types are considered valid for JavaScript, including the [legacy JavaScript types](#legacy_javascript_mime_types) listed below.
+Using any MIME type other than `text/javascript` may result in scripts that do not load or run.
 
-You may find some JavaScript content incorrectly served with a `charset` parameter as part of the MIME type — as an attempt to specify the character set for the script content.
-That `charset` parameter isn't valid for JavaScript content, and in most cases will result in a script failing to load.
+Note that in HTML the [`type`](/en-US/docs/Web/HTML/Reference/Elements/script/type) attribute for {{htmlelement("script")}} elements may only contain the **MIME type essence**: `text/javascript`.
+Including the parameters such as `charset=utf-8` will prevent the resource from being requested.
+In contrast, when using the HTTP {{httpheader("Content-Type")}} header you may optionally specify the `charset` parameter (browsers assume {{glossary("UTF-8")}} by default).
+When requesting a [JavaScript module](/en-US/docs/Web/JavaScript/Guide/Modules) via HTTP, if you do set the `charset` parameter then the header _must_ specify UTF-8 as shown: `Content-Type: text/javascript; charset=utf-8`.
+
+For more information see: [IANA Media Types registry](https://www.iana.org/assignments/media-types/media-types.xhtml#text), [RFC 9239](https://www.rfc-editor.org/rfc/rfc9239.html), and the [HTML specification](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:text/javascript).
 
 #### Legacy JavaScript MIME types
 
--- a/files/en-us/web/http/guides/mime_types/index.md
+++ b/files/en-us/web/http/guides/mime_types/index.md
@@ -123,7 +123,6 @@ In HTML, the `type` attribute for elements like `<script>` and `<style>` only su
 
 In contrast, parameters in HTTP `Content-Type` headers (e.g., `Content-Type: text/javascript; charset=utf-8`) are technically allowed. However, modern browsers assume UTF-8 by default and typically ignore the parameter, making it unnecessary in most cases.
 
-
 ### application/octet-stream
 
 This is the default for binary files. As it means _unknown binary_ file, browsers usually don't execute it, or even ask if it should be executed. They treat it as if the {{HTTPHeader("Content-Disposition")}} header was set to `attachment`, and propose a "Save As" dialog.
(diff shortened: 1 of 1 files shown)


hamishwillee (commented):
Thanks! The point of the original was pretty clear: the only allowed mime type is `text/javascript` - you shouldn't serve all those other types listed below in the section "Legacy JavaScript MIME types".

That is completely true ^^^, because `charset` is not a mime type - its an optional parameter to the MIME type. Splitting hairs of course, but to my mind the charset issue is a really minor side point to the mime type, so we should keep that first and forefront.

The way I would handle this is to treat it as such. Start with the original text, then have a comment about where you can and cannot use the `charset` parameter with the mime type.

Does that make sense?

@Josh-Cena FYI - you had a comment on the linked issue.

Josh-Cena:
Yeah this PR shifts the focus to the wrong part. I would simply remove the second paragraph, and mention "MIME type essence" somewhere else other than under `text/javascript`, because it applies to `<style>` as well. It appears that the second paragraph is straight incorrect because it implies `charset` is invalid in `Content-Type` which I don't think is the case (I could be wrong).

Shrinivassab:
Thank you @Josh-Cena and @hamishwillee for the helpful feedback. I have now made the following updates:

1. Focused the `text/javascript` section solely on the valid MIME type as originally intended.
2. Removed the charset explanation from that section.
3. Added a new general section about MIME type parameters in HTML, which applies to both `<script>` and `<style>` elements.

Let me know if -- 
1. The new section's placement or wording should be adjusted
2. or You would prefer to omit it entirely

Thanks again for guiding this, I really appreciate your time and support! -

Josh-Cena:
I will let Hamish continue with the review, but (a) this is not the right place to put it because we are listing MIME types under this section (b) you still didn't remove the incorrect paragraph in the original report.

hamishwillee:
Thanks. I have subedit this to my taste - essentially rewording so that the essence stuff is part of the JS section, where the error was reported. 

The previous section added wasn't in the right place IMO (it was about how mime types are used, not "an important MIME type". It also generalized to style types what is about script types - this might apply to styles, but if so, I'd add it to that section.

@Josh-Cena You might want to sanity check this too. If not, I will do so on Monday when I have objectivity again.

files/en-us/web/http/guides/mime_types/index.md
```diff
@@ -146,16 +146,16 @@ All HTML content should be served with this type. Alternative MIME types for XHT
 
 ### text/javascript
 
-Per the [IANA Media Types registry](https://www.iana.org/assignments/media-types/media-types.xhtml#text), [RFC 9239](https://www.rfc-editor.org/rfc/rfc9239.html), and the [HTML specification](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:text/javascript), JavaScript content should always be served using the MIME type `text/javascript`.
-No other MIME types are considered valid for JavaScript, and using any MIME type other than `text/javascript` may result in scripts that do not load or run.
+JavaScript content should always be served with the MIME type `text/javascript`.
+No other MIME types are considered valid for JavaScript, including the [legacy JavaScript types](#legacy_javascript_mime_types) listed below.
+Using any MIME type other than `text/javascript` may result in scripts that do not load or run.
```
Josh-Cena:
```suggestion
For historic reasons, browsers may support some [legacy JavaScript types](#legacy_javascript_mime_types) listed below, but you should not assume scripts served with any MIME type other than `text/javascript` will always load or run.
```

files/en-us/web/http/guides/mime_types/index.md
```diff
@@ -146,16 +146,16 @@ All HTML content should be served with this type. Alternative MIME types for XHT
...
 
-You may find some JavaScript content incorrectly served with a `charset` parameter as part of the MIME type — as an attempt to specify the character set for the script content.
-That `charset` parameter isn't valid for JavaScript content, and in most cases will result in a script failing to load.
+Note that in HTML the [`type`](/en-US/docs/Web/HTML/Reference/Elements/script/type) attribute for {{htmlelement("script")}} elements may only contain the **MIME type essence**: `text/javascript`.
+Including the parameters such as `charset=utf-8` will prevent the resource from being requested.
+In contrast, when using the HTTP {{httpheader("Content-Type")}} header you may optionally specify the `charset` parameter (browsers assume {{glossary("UTF-8")}} by default).
+When requesting a [JavaScript module](/en-US/docs/Web/JavaScript/Guide/Modules) via HTTP, if you do set the `charset` parameter then the header _must_ specify UTF-8 as shown: `Content-Type: text/javascript; charset=utf-8`.
```
Josh-Cena:
```suggestion
Note that in HTML the [`type`](/en-US/docs/Web/HTML/Reference/Elements/script/type) attribute for {{htmlelement("script")}} elements may only contain the **JavaScript MIME type essence**: `text/javascript`.
Including any parameter, such as `charset=utf-8`, is the same as setting the `type` to any other MIME type: the script content is treated as a data block and is not executed as JavaScript.
(Note that setting the `type` to a JavaScript MIME type is a deprecated feature itself: you should omit the `type` in this case.)
In contrast, when using the HTTP {{httpheader("Content-Type")}} header you may optionally specify the `charset` parameter as usual.
```

I don't think we shold go in detail about charset requirements, because other types have charset requirements as well but we don't talk about them.

Shrinivassab: Understood, I will follow the same  approach to stay consistent.

hamishwillee: > I don't think we shold go in detail about charset requirements, because other types have charset requirements as well but we don't talk about them.

@Josh-Cena I disagree with the sentiment: that's like saying  "We did it wrong everywhere else, so let's do it wrong here too". Being right has to start somewhere.

The question here for me is "should users know this if using the MIME type via HTTP". I think the fact that the charset has a mandatory format for modules is important.
If that isn't important, then yes, we might accept this change. Is it?

Josh-Cena: I'm happy to point out, say, that HTML and JSON must be UTF-8 as well, but I think in general the MIME type is tangential to its charset. The latter is more about the specific technology handling it and should be documented within each technical reference. YMMV; I can also see an argument otherwise.

hamishwillee: Definitely arguable, but you've convinced me that on balance this is not the right place. Thanks.

hamishwillee (approved):
Thanks @Shrinivassab @Josh-Cena !

PR #40106, opened 2025-06-27
fix: Prettier 3.6.2 compat bug with MD031

bsmth:
### Description

Bump Prettier and format non-compliant files.

### Motivation

There was a compat bug with Prettier v3.6.1 and markdownlint causing dueling linters.

### Related

- https://github.com/DavidAnson/markdownlint/blob/main/doc/md031.md
- https://redirect.github.com/prettier/prettier/issues/17652
- https://github.com/prettier/prettier/releases/tag/3.6.2

### DIFF
--- a/files/en-us/learn_web_development/core/frameworks_libraries/svelte_stores/index.md
+++ b/files/en-us/learn_web_development/core/frameworks_libraries/svelte_stores/index.md
@@ -577,6 +577,7 @@ Moreover, because web storage only supports saving string values, we will have t
      };
    };
    ```
+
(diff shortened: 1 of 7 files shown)


Josh-Cena:
We should keep using exact versions

bsmth:
> We should keep using exact versions

Remove the `^` range in af524c0f1ddd3187404f59aa2cef295b31ce5924 but I guess we need to run `yarn upgrade` or pin the others separately

Josh-Cena (approved):

PR #40001, opened 2025-06-20
docs: Add warning and clarify named form access via document.formName(#39891)

Shrinivassab:
### Description

Documented named form access on `document` object and discouraged its usage.

### Motivation

Clarifies confusing or undocumented behavior; resolves #39891.

### Related issues and pull requests

#39986

PR #40109, opened 2025-06-27
fix(document.forms): Convert text to warning

dipikabh:
### Description

This is a follow-up to https://github.com/mdn/content/pull/40001.

@Josh-Cena @hamishwillee can you take look?
I think with the phrasing "dangerous and discouraged" in the para, we should ideally change the text into a "Warning" callout.

I've also rephrased the text to follow this format: "though `<this is possible>`, don't do it `<because>`. always `<do this>`."

### DIFF
--- a/files/en-us/web/api/document/forms/index.md
+++ b/files/en-us/web/api/document/forms/index.md
@@ -13,11 +13,12 @@ The **`forms`** read-only property of the {{domxref("Document")}} interface retu
 > [!NOTE]
 > Similarly, you can access a list of a form's component user input elements using the {{domxref("HTMLFormElement.elements")}} property.
 
+You can also access named `<form>` elements as properties of the `document` object.
+For example, `document["login-form"]` and `document.forms["login-form"]` can both be used to access the form named `login-form`.
+
 > [!WARNING]
-> Though you can access named `<form>` elements as properties of the `document` object itself, relying on this behavior is dangerous and discouraged.
-> For example, `document["login-form"]` and `document.forms["login-form"]` can both access the form named `login-form`.
-> However, using `document["login-form"]` can lead to unexpected conflicts with existing or future APIs in the browser.
-> For example, if a browser introduces a built-in `document["login-form"]` property in the future, your code will no longer be able to access your form element.
+> Relying on the `document["form-name"]` pattern is dangerous and discouraged because it can lead to unexpected conflicts with existing or future APIs in the browser.
+> For example, if a browser introduces a built-in `document["login-form"]` property in the future, your code may no longer be able to access the form element.
 > To avoid such conflicts, always use `document.forms` to access named forms.
 
 ## Value
--- a/files/en-us/web/api/document/forms/index.md
+++ b/files/en-us/web/api/document/forms/index.md
@@ -15,7 +15,7 @@ The **`forms`** read-only property of the {{domxref("Document")}} interface retu
 
 > [!WARNING]
 > Though you can access named `<form>` elements as properties of the `document` object itself, relying on this behavior is dangerous and discouraged.
-> For example, `document["login-form"]` and `document.forms["login-form"]` can both access the form named `login-form`. 
+> For example, `document["login-form"]` and `document.forms["login-form"]` can both access the form named `login-form`.
 > However, using `document["login-form"]` can lead to unexpected conflicts with existing or future APIs in the browser.
 > For example, if a browser introduces a built-in `document["login-form"]` property in the future, your code will no longer be able to access your form element.
 > To avoid such conflicts, always use `document.forms` to access named forms.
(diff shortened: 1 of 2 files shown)


Josh-Cena (commented):
I don't think you should make this whole thing a warning. Put the descriptive thing in prose, and then add a prescriptive warning. Something like this:

```md
Named `<form>` elements are also exposed as properties of the `document` object itself.
For example, `document["login-form"]` and `document.forms["login-form"]` can both access the form named `login-form`.

> [!WARNING]
> Relying on this behavior is dangerous and discouraged.
> It can lead to unexpected conflicts with some existing or future APIs in the browser.
> For example, if browsers introduce a new `document` property with the same name as your form, then the same code will no longer be able to access the form element. Always use `document.forms` instead.
```

This aligns with https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id#description

files/en-us/web/api/document/forms/index.md
```diff
@@ -13,10 +13,12 @@ The **`forms`** read-only property of the {{domxref("Document")}} interface retu
...
-Named `<form>` elements are also exposed as properties of the `document` object itself.
-For example, `document["login-form"]` and `document.forms["login-form"]` can both access the form named `login-form`. Relying on this behavior is dangerous and discouraged.
-It can lead to unexpected conflicts with some existing or future APIs in the browser.
-For example, if browsers introduce a new `document` property with the same name as your form, then the same code will no longer be able to access the form element. Always use `document.forms` instead.
+> [!WARNING]
+> Though you can access named `<form>` elements as properties of the `document` object itself, relying on this behavior is dangerous and discouraged.
+> For example, `document["login-form"]` and `document.forms["login-form"]` can both access the form named `login-form`.
+> However, using `document["login-form"]` can lead to unexpected conflicts with existing or future APIs in the browser.
+> For example, if a browser introduces a built-in `document["login-form"]` property in the future, your code will no longer be able to access your form element.
+> To avoid such conflicts, always use `document.forms` to access named forms.
```
dipikabh: - Sure, I've separated the factual description and warning bits.
- I'd also like to retain active voice as far as possible, so "You can also access..." instead of "...are also exposed..."
 
```suggestion
You can also access named `<form>` elements as properties of the `document` object.
For example, `document["login-form"]` and `document.forms["login-form"]` can both be used to access the form named `login-form`.

> [!WARNING]
> Relying on the `document["form-name"]` pattern is dangerous and discouraged because it can lead to unexpected conflicts with existing or future APIs in the browser.
> For example, if a browser introduces a built-in `document["login-form"]` property in the future, your code may no longer be able to access the form element.
> Always use `document.forms` to access named forms.
```
@Josh-Cena wdyt?

Josh-Cena: Let's keep the note's wording exactly consistent with that of `id`. You can choose to change either. Otherwise lgtm

hamishwillee: What he said ^^^. I do like the new structure/wording.

files/en-us/web/html/reference/global_attributes/id/index.md
```diff
@@ -47,20 +47,22 @@ Also, not all valid ID attribute values are valid JavaScript identifiers. For ex
...
-You could access the paragraph element in JavaScript using code like:
+You can access this paragraph element in JavaScript using the following code:
 
 ```js
 const content = window.preamble.textContent;
 ```
 
 > [!WARNING]
-> Relying on this behavior is dangerous and discouraged. It can lead to unexpected conflicts with some existing or future APIs in the browser. For example, if browsers introduce a new global property named `preamble`, then the same code will no longer be able to access the HTML element. Use `document.getElementById()` or `document.querySelector()` instead.
+> Relying on the `window["id-value"]` or `window.idValue` pattern is dangerous and discouraged because it can lead to unexpected conflicts with existing or future APIs in the browser.
```
dipikabh, line 63: I especially want to spell out what "this behavior" is for better clarity

files/en-us/web/api/document/forms/index.md
dipikabh: Sure, I've now updated the warning text on the `id` page as well

hamishwillee (approved):
LGTM. Approving but not merging to give @Josh-Cena a chance to look.

Josh-Cena (approved):

files/en-us/web/html/reference/global_attributes/id/index.md
```diff
@@ -47,20 +47,22 @@ Also, not all valid ID attribute values are valid JavaScript identifiers. For ex
...
+> For example, if a browser introduces a built-in global property named `preamble` in the future, your code may no longer be able to access the HTML element.
+> To avoid such conflicts, always use `document.getElementById()` or `document.querySelector()` to access elements by ID.
```
estelle:
```suggestion
> To avoid such conflicts, always use the {{domxref("Document.getElementById()")}} or {domxref("Document.querySelector()")}} method to access elements by ID.
```
first mention of this method i think

dipikabh: Nice, added link for `querySelector()` as well even though it has been linked before but would be better to provide both links here

dipikabh:
Thank you all for your inputs!

PR #40112, opened 2025-06-27
Improve wording for Promise resolution with thenables

rovazh:
### Description

Clarified the behavior of Promise resolution when passed a thenable value. The description now specifies that if calling the then method throws **an error**, the current promise is rejected with **that error**.

### Motivation

This improves the readability and precision of the documentation by making the error-handling behavior of Promises more explicit. Adding "an error" helps reinforce that the rejection is based on thrown exceptions.


Josh-Cena:
"Throws" is the standard verbiage, just like a function can "return" and we don't necessarily spell out that it "returns a value" because that's simply irrelevant.

PR #40118, opened 2025-06-28
fix: correct prime example

jasonren0403:
### Description

2 is prime number too, we should split first if condition to make sure the implementation is completely correct

### Motivation

make sure example is perfect


Josh-Cena (approved):
Nice, thanks! I made the other examples use the exact same implementation. In addition I added a note about its poor performance. Also I had to remove the extra example you added because this isn't the focus of the example anyway, and I don't want to add one to every page.

PR #40119, opened 2025-06-28
Private property -> private element

### LINKED ISSUE #39984: no such thing as "private properties", they are "elements"
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

"Private properties"

Properties are enumerable, introspectable, reflectable, proxyable. Fields, however, are not.

Thus, "private properties" is *not a concept that exists in JavaScript*, and it causes confusion to describe the `#foo` feature as "properties".

This page should be retitled "Private fields" and the verbiage updated to avoid this confusion.

"Private fields"

_No response_

_No response_

Josh-Cena:
Fix https://github.com/mdn/content/issues/39984

@ljharb


ljharb (commented):
LGTM, altho this is a bit of a rubberstamp

Josh-Cena:
(Note that the only substantial change I made is this:)

> [!NOTE]
> On MDN, we avoid using the term "private property". A [property](/en-US/docs/Glossary/Property/JavaScript) in JavaScript has a string or symbol key, and has attributes like `writable`, `enumerable`, and `configurable`, but private elements have none. While private elements are accessed with the familiar dot notation, they cannot be [proxied](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy), enumerated, deleted, or interacted with using any {{jsxref("Object")}} method.

ljharb:
Love it. Thanks!

pepelsbey (approved):
Looks good! Thank you :)

PR #40133, opened 2025-06-30
Fixed incorrect reference from "current" to "previous" document on unload

DerGernTod:
### Description

Fixed incorrect reference from "current" to "previous" document on unload.

### Motivation

I had to look up the spec as to why our validations against navigation timings fails so often since we implemented it based on MDN. The "current" document is a bit of an ambiguous description regarding the unload event (the description below correctly mensions "previous").

### Additional details

[PerformanceNavigationTiming unloadeventstart spec](https://w3c.github.io/navigation-timing/#dom-performancenavigationtiming-unloadeventstart)


sideshowbarker (approved):

sideshowbarker:
Gernot, thanks for catching this and fixing it, and congrats on landing your first docs change here — welcome aboard 🎉

PR #40135, opened 2025-06-30
feat(scripts): add `start:fred` script

argl:
### Description

Preview of using our new frontend `fred` in content, enabling fred-based rendering locally.

### Motivation

Transition form yari to fred-

### Additional details

run `yarn start:fred` to try it

This is just a preliminary release, to start collecting issues and  regressions. None of the usual writer's tools work yet, but previewing content should work as expected.


package.json
```diff
@@ -37,13 +37,15 @@
...
     "start": "yarn -s info:rari && yarn up-to-date-check && env-cmd --silent cross-env CONTENT_ROOT=files REACT_APP_DISABLE_AUTH=true BUILD_OUT_ROOT=build rari-server",
+    "start:fred": "yarn -s info:rari && yarn up-to-date-check && env-cmd --silent cross-env CONTENT_ROOT=files BUILD_OUT_ROOT=build NODE_ENV=production fred-server",
     "up-to-date-check": "node scripts/up-to-date-check.js",
     "jest": "cross-env NODE_OPTIONS=--experimental-vm-modules jest",
     "test:front-matter-linter": "yarn jest tests"
   },
   "dependencies": {
     "@apideck/better-ajv-errors": "^0.3.6",
     "@caporal/core": "^2.0.7",
+    "@mdn/fred": "^0.0.1",
```
bsmth: Might be better to remove the `^` caret range and pin to current patch version `0.0.1` instead.

argl: Good point.

argl:
There is an issue with the node version, that need fixes in various places to be done correctly. Stay tuned.

caugner (approved):

PR #40146, opened 2025-07-01
OPTIONS Request "May" have body

samuel871211:
### Description

Based on [RFC9110#section-9.3.7](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.7)

```
A client that generates an OPTIONS request containing content MUST send a valid [Content-Type](https://www.rfc-editor.org/rfc/rfc9110.html#field.content-type) header field describing the representation media type. Note that this specification does not define any use for such content.
```

Which means OPTIONS Request "May" have body.

### DIFF
--- a/files/en-us/web/http/reference/methods/options/index.md
+++ b/files/en-us/web/http/reference/methods/options/index.md
@@ -16,7 +16,7 @@ A client can specify a URL with this method, or an asterisk (`*`) to refer to th
   <tbody>
     <tr>
       <th scope="row">Request has body</th>
-      <td>No</td>
+      <td>May</td>
     </tr>
     <tr>
       <th scope="row">Successful response has body</th>
--- a/files/en-us/web/http/reference/methods/options/index.md
+++ b/files/en-us/web/http/reference/methods/options/index.md
@@ -41,7 +41,8 @@ A client can specify a URL with this method, or an asterisk (`*`) to refer to th
   </tbody>
 </table>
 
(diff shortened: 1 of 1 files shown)


bsmth:
Thanks! Do you want to add an asterisk (`May*`) with a note stating that although this is not forbidden, there is no specified behavior, so implementations may vary (i.e., only do this if you know that a server supports this), and include a proper `Content-Type` at least?

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Language for the `Yes*` usage.

samuel871211:
@bsmth Good idea! Let me work on this

samuel871211:
@bsmth Since I am not a native English speaker nor a professional writer, if the sentence is not good enough in grammar or fluency, please feel free to fix this PR or give any feedback.

files/en-us/web/http/reference/methods/options/index.md
```diff
@@ -41,6 +41,8 @@ A client can specify a URL with this method, or an asterisk (`*`) to refer to th
   </tbody>
 </table>
 
+\* Although the [RFC 9110, section 9.3.7: OPTIONS](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.7) does not define any semantics for an `OPTIONS` request body, it is technically allowed. Clients may include a request body with `OPTIONS` as long as they provide a valid {{HTTPHeader("Content-Type")}} header, but behavior is implementation-dependent — use it only when you know the server explicitly supports it.
```
bsmth:
```suggestion
\* Although an `OPTIONS` message with a request body is technically allowed, it has no defined semantics.
You may include a body in an `OPTIONS` message as long as you provide a valid {{HTTPHeader("Content-Type")}} header, and when you know the server expects it, as behavior is implementation-specific.
```

bsmth: Just a small tweak here. We can drop the spec link in this case, and slight reword.

samuel871211:
LGTM

bsmth (approved):
Great, thanks a lot 👍🏻

PR #40147, opened 2025-07-01
TRACE is not allowed in fetch

samuel871211:
### Description

According to [fetch.spec.whatwg.org/#forbidden-method](https://fetch.spec.whatwg.org/#forbidden-method), we can't set `TRACE` method in `fetch`

### DIFF
--- a/files/en-us/web/http/reference/methods/trace/index.md
+++ b/files/en-us/web/http/reference/methods/trace/index.md
@@ -99,8 +99,7 @@ Accept: */*
 
 ## Browser compatibility
 
-The browser doesn't use the `TRACE` method for user-initiated actions, so "browser compatibility" doesn't apply.
-Developers can't set this request method using [`fetch()`](/en-US/docs/Web/API/Window/fetch) because it's a [forbidden method](https://fetch.spec.whatwg.org/#forbidden-method).
+There's no browser compat
 
 ## See also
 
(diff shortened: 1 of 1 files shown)


Josh-Cena:
ok but then this sentence is redundant. It was saying that "although browsers don't natively send it, you can programmatically send it in a browser". Now we are saying that it can only be used in non-browser environments altogether, so we can simply remove this remark and just say "there's no browser compat" as in the first sentence.

samuel871211:
@Josh-Cena Fixed, thank you

files/en-us/web/http/reference/methods/trace/index.md
```diff
@@ -99,8 +99,7 @@ Accept: */*
 
 ## Browser compatibility
 
-The browser doesn't use the `TRACE` method for user-initiated actions, so "browser compatibility" doesn't apply.
-Developers can set this request method using [`fetch()`](/en-US/docs/Web/API/Window/fetch).
+There's no browser compat
```
Josh-Cena:
```suggestion
The browser doesn't use the `TRACE` method for user-initiated actions, so "browser compatibility" doesn't apply.
```

samuel871211:
LGTM

bsmth (approved):
Great, thanks 👍🏻

PR #40150, opened 2025-07-01
Modernize CSS: use media range syntax

Josh-Cena:
The range syntax is almost always better than using `min-` or `max-`:

- It allows specifying exclusive ranges, and inclusive ranges are more obvious
- It's shorter and more straightforward to read
- It avoids the pitfall where if you have two queries like `min-width: 600px` and `max-width: 599px`, neither query would apply if the device is 599.5px wide. In fact we have quite a few cases like this!


bsmth (approved):
LGTM, thanks 👍🏻

PR #21696, opened 2022-10-20
Title update: remove single colon from title

estelle:
Changed the title for these four as double colon notation is the correct way to do it and has been supported for about 20 years. The text includes a comment about the history of single colon notation. It does not need to be part of the title.

PR #40151, opened 2025-07-01
Modernize CSS: use double colon for all pseudoelements

Josh-Cena:
The single-colon versions are regarded as legacy, and these things are conceptually elements, not classes, so we should use the double colon version.


bsmth (approved):
Great, tnx. See also https://github.com/mdn/content/pull/21696

PR #40152, opened 2025-07-01
HTML best practices

Josh-Cena:
- Make sure `html` elements have `lang="en-US"`
- `doctype` should be lowercase
- Remove duplicate attributes


bsmth (approved):
Lovely, thank you

PR #40158, opened 2025-07-01
Fix content issues

Josh-Cena:
- Fix redirected link
- Fix invalid CSS syntax
- Fix incorrect internal link
- Fix incorrect slug, incorrect BCD key


bsmth (approved):
Great, thanks a lot 👍🏻

caugner:
@Josh-Cena Do you know what caused those issues? Shouldn't they have been detected by lint/test.

Josh-Cena:
None of these is covered by any lint. The only one that might be is the invalid BCD key, but people tend to ignore these flaws.

PR #40159, opened 2025-07-01
Technical review: Add doc for CSS progress() function

chrisdavidmills:
### Description

Chrome 138 adds the CSS [`progress()`](https://drafts.csswg.org/css-values-5/#progress) function — see https://chromestatus.com/feature/5112558941634560.

This PR adds docs for the `progress()` function.

Note: The final demo code snippets are out of sync with the version available in the `dom-examples` repo. You can find the latest updates added in this PR: https://github.com/mdn/dom-examples/pull/314. I'll keep it open until the reviews are done, in case there are further changes to make.


files/en-us/web/css/progress/index.md
```diff
@@ -0,0 +1,372 @@
...
+Respectively, these are:
+
+- Progress
+  - : The value to calculate the position of relative to the other two values.
+- Progress start
+  - : The lower progress bound.
+- Progress end
+  - : The upper progress bound.
+
+The parameters can be mathematical expressions or simple values. The values (or sum results) can be any {{cssxref("&lt;number>")}}, {{cssxref("&lt;dimension>")}}, or {{cssxref("&lt;percentage>")}} value. They can be of different units, but they must all be of the same type, or else the function is invalid.
```
pepelsbey: I think it's important enough to show a less obvious failing example below and link to it from here (or just do it inline) to attract more attention to this limitation.

files/en-us/web/css/progress/index.md
```diff
@@ -0,0 +1,372 @@
...
+```css
+opacity: progress(5, 0, 10);
+```
+
+In this case, the computed value of {{cssxref("opacity")}} would be `0.5`, as 5 is mid-way between `0` and `10`.
+
+However, because `progress()` only ever returns a unitless value between `0` and `1`, it is common to combine it with another math function such as {{cssxref("calc()")}} to output the value and units you want. The following example calculates a container {{cssxref("width")}} as a percentage between a minimum width of `320px` and a maximum width of `1200px`. The `calc()` function is used to multiply the `progress()` return value by `100` and set it to having percentage units.
+
+```css
+width: calc(progress(var(--container-width), 320px, 1200px) * 100%);
```
pepelsbey: This example would benefit from the computed value in a comment. Sort of a console log. Something like "the computed value will alternate between x% and y%, depending on the container with".

files/en-us/web/css/progress/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+{{EmbedLiveSample("basic", "100%", "150")}}
+
+The width of the `<div>` is `75%` of the `<section>` width, as the `min-width` is `400px`, the `max-width` is `700px`, and the `width` is `600px`, which is `75%` of the distance between the previous two values.
+
+### Resize effects on a container
+
+This example shows off some more involved uses of the `progress()` function, resulting in some fun effects as the browser window is resized.
+
+This example works much better when rendered full-size in a desktop browser tab. Therefore, we have not rendered it in an embedded live sample in this page. Instead, you can find it running live at [CSS `progress` function demo](https://mdn.github.io/dom-examples/css-progress/) (also see the [source code](https://github.com/mdn/dom-examples/tree/main/css-progress)).
```
pepelsbey:
```suggestion
This example works much better when rendered full-size in a desktop browser tab. Therefore, we have not rendered it in an embedded live sample in this page. Instead, you can find it running live at [CSS `progress()` function demo](https://mdn.github.io/dom-examples/css-progress/) (also see the [source code](https://github.com/mdn/dom-examples/tree/main/css-progress)).
```

files/en-us/web/css/progress/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+Open the live example in a separate tab, and try increasing and decreasing the browser window width to see the effect. Keep this open so you can refer back to it as you read the explanation below.
+
+#### HTML
+
+Our HTML features an {{htmlelement("article")}} element that contains the rest of our content, and two {{htmlelement("section")}} elements — one to hang a background image off, and the other one to contain our content. The `content` `<section>` also contains a `progress` `<div>` representing a width progress bar, the same as the one in our previous demo. We have omitted the rest of the content for brevity.
```
pepelsbey: It would be much easier to read it like that:

```suggestion
Our HTML features an {{htmlelement("article")}} element that contains the rest of our content, and two {{htmlelement("section")}} elements — one to hang a background image off, and the other one to contain our content. The `<section class="content">` also contains a `<div class"progress">` representing a width progress bar, the same as the one in our previous demo. We have omitted the rest of the content for brevity.
```

pepelsbey (commented):
Looks good! A few suggestions below 🙂

files/en-us/web/css/progress/index.md
chrisdavidmills: In my next commit, I've expanded this bit to its own subsection inside the "Description" section, including more explanation and some examples. Let me know what you think.

files/en-us/web/css/progress/index.md
chrisdavidmills: This example was not very good, because we were actually multiplying the `progress()` result by `100%`, which means its width is relative to its own parent element's width, so kinda pointless. 

I've updated it so that it is instead relative to the viewport width, and included an example calculation of the computed value. Let me know what you think.

files/en-us/web/css/progress/index.md
chrisdavidmills: Updated, ta.

files/en-us/web/css/progress/index.md
chrisdavidmills: Good call; updated.

pepelsbey (approved):
Perfect! Thank you 🙂

PR #28602, opened 2023-08-17
Correcting technical summary table for <summary>

Brian-Pob:
### Description

Correcting the implicit ARIA role in the technical summary table for the `summary` element.

### Motivation

To ensure the MDN docs follow the relevant HTML specs.

### Additional details

User [scottaohara](https://github.com/scottaohara) is an editor for the HTML AAM spec and the ARIA in HTML spec. He states in the linked issue that the table should:
1. Be updated to reference what the ARIA in HTML spec says
2. Include the ARIA in HTML spec in the Specifications section

### Related issues and pull requests

Fixes #28446

PR #40165, opened 2025-07-01
Remove Summary a11y warning

leekeh:
### Description

Remove outdated warning on accessibility roles in the `<summary>` element.

### Motivation

The implicit role of the element was changed from `button` to `no corresponding role` in the aria specifications at the end of 2022:

- https://github.com/w3c/html-aria/pull/434
- https://github.com/w3c/html-aria/pull/435

Seems like the `implicit role` was corrected on mdn in 2023 (https://github.com/mdn/content/pull/28602), but this specific warning was forgotten.

Although the specifications still contain [a warning](https://w3c.github.io/html-aria/#el-summary) that browsers will still use a button role, I tested this and found it to not be the case at least in Chromium, Firefox and Safari. So perhaps the specs also need to be updated.

Relates to #28602

### DIFF
--- a/files/en-us/web/html/reference/elements/summary/index.md
+++ b/files/en-us/web/html/reference/elements/summary/index.md
@@ -114,9 +114,6 @@ You can use heading elements in `<summary>`, like this:
 
 This currently has some spacing issues that could be addressed using CSS.
 
-> [!WARNING]
-> Because the `<summary>` element has a default role of [button](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role) (which strips all roles from child elements), this example will not work for users of assistive technologies such as screen readers. The `<h4>` will have its role removed and thus will not be treated as a heading for these users.
-
 ### HTML in summaries
 
 This example adds some semantics to the `<summary>` element to indicate the label as important:
--- a/files/en-us/web/html/reference/elements/summary/index.md
+++ b/files/en-us/web/html/reference/elements/summary/index.md
@@ -114,6 +114,9 @@ You can use heading elements in `<summary>`, like this:
 
 This currently has some spacing issues that could be addressed using CSS.
 
+> [!WARNING]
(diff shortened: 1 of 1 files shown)


files/en-us/web/html/reference/elements/summary/index.md
```diff
@@ -114,9 +114,6 @@ You can use heading elements in `<summary>`, like this:
 
 This currently has some spacing issues that could be addressed using CSS.
 
-> [!WARNING]
-> Because the `<summary>` element has a default role of [button](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role) (which strips all roles from child elements), this example will not work for users of assistive technologies such as screen readers. The `<h4>` will have its role removed and thus will not be treated as a heading for these users.
```
dipikabh: I would suggest updating it as proposed in https://github.com/mdn/content/pull/28602#issuecomment-1688190245 (slightly edited):

```
> [!WARNING]
> The role assigned to the `<summary>` element varies across browsers. Some still assign it a default [`button`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role) role, which removes all roles from its children. This inconsistency can cause issues for users of assistive technologies such as screen readers (`<h4>` in the previous example will have its role removed and will not be treated as a heading for these users). You should test your `<summary>` implementation on multiple platforms to ensure there's consistent accessibility support.
```

dipikabh (changes requested):
@leekeh Thanks a lot for following up on that earlier PR.
Indeed, it got merged without an update to the warning message.

You mention you've tested the current behavior, but I think we should retain the warning for good measure.

files/en-us/web/html/reference/elements/summary/index.md
leekeh: I'm curious for which user agents this is still the case, is this still a concern?

dipikabh: We don't have an exhaustive list of how all the different combinations of browsers and assistive technologies are handling this. So I'd say it's worth keeping this cautionary note around, at least until the spec is updated as well.

dipikabh: @leekeh would you be able to come back to this PR soonish? Would be great to have this published.

dipikabh:
Hi @leekeh, I've gone ahead and applied the edited warning text to keep things moving.
If you have any concerns or follow-ups, please feel free to comment here or open a new PR.

Thanks again for your contribution!!

dipikabh (approved):

PR #40139, opened 2025-06-30
feat: PoC move APIRef macro into front matter

bsmth:
### Description

This PR is a demo of moving the `{{apiref(.*)}}` macro into front matter.

__Changes:__
* Move `APIRef` into front matter
* Update front matter schema to allow for both strings or a list of items which accept string args

### Motivation

I have made the replacements across the entire `files/en-us/web/api/**/*` section for `APIRef`, which touches 7,000+ files. This PR is a small example on a subset of files so we can show that it works as expected with different parameter formats (with/without whitespace, without params).

For the full `APIRef` changes, see:

https://github.com/mdn/content/compare/main...bsmth:content:api-sidebar?expand=1

### Related issues and pull requests

- [x] https://github.com/mdn/rari/pull/237 (landed in [rari v0.1.41](https://github.com/mdn/rari/releases/tag/v0.1.41))

PR #40168, opened 2025-07-02
feat: Move defaultapisidebar into front matter

bsmth:
### Description

Moving the `{{DefaultAPISidebar(.*)}}` macro into front matter.

__Changes:__
* Move `DefaultAPISidebar` into front matter
* Update front matter schema to allow for this usage

### Motivation

Before looking at `APIRef`, we may want to move this macro first, which only touches 360 files.
The `APIRef` macro is used in 7,000+ files, so we can start small. 

For the full `APIRef` changes, see:

https://github.com/mdn/content/compare/main...bsmth:content:api-sidebar?expand=1

- [ ] https://github.com/mdn/content/pull/40139
- [x] https://github.com/mdn/rari/pull/237 (landed in [rari v0.1.41](https://github.com/mdn/rari/releases/tag/v0.1.41))


Josh-Cena:
I think @wbamberg has a really good point that the group name ought to be a separate metadata so it can have a shared schema between `apiref` and `defaultapisidebar` (i.e. a top level `group: Attribution Reporting API` front matter). This also allows easy cross-checking and validation. Has this alternative been considered? (We'll keep the capability of passing arguments to sidebars in case the need arises)

fiji-flo:
@Josh-Cena I'm with you, I think.

So this would be for Web/API/Console_API:
```
---
title: Console API
slug: Web/API/Console_API
page-type: web-api-overview
browser-compat: api.console
group: Console API
sidebar: DefaultAPISidebar
---
```

And for e.g. Web/API/console/log_static:
```
---
title: "console: log() static method"
short-title: log()
slug: Web/API/console/log_static
page-type: web-api-static-method
browser-compat: api.console.log_static
group: Console API
sidebar: APIRef
---
```

That would be my preferred way, too. My only concern is redundancy with web features / bcd. Maybe we can fully derive this out of BCD? Should we look at [`GroupData.json`](https://github.com/mdn/content/blob/main/files/jsondata/GroupData.json) and use bcd keys as index?

caugner:
If that frontmatter key is intended for the API sidebars only, may I suggest naming it `api-group`, `api` or `web-api`?
(I understand that `group` comes from `GroupData.json`, but IMO that's not the most meaningful name.)

Josh-Cena:
CSS modules could also benefit from this, I believe, since the IA is quite similar. I'm indifferent otherwise

caugner:
> CSS modules could also benefit from this, I believe, since the IA is quite similar. I'm indifferent otherwise

Even more so a reason to use distinct front matter keys, to avoid referencing a CSS group in the API context and vice versa. 🙂

caugner:
> Maybe we can fully derive this out of BCD? Should we look at [`GroupData.json`](https://github.com/mdn/content/blob/main/files/jsondata/GroupData.json?rgh-link-date=2025-07-02T12%3A40%3A52Z) and use bcd keys as index?

Currently, this cannot be derived from BCD.

bsmth:
Shall I set this to a draft status in the meantime?

Josh-Cena:
If everyone's onboard with this idea, I think we can move directly to the new solution. Otherwise if we still need discussion and decision making, happy for this to move forward as-is.

bsmth:
Tallying it up looks like Claas is pro (thumbs up), Florian is pro, and so are Will and Josh. I'm neutral on it, so I think it's better to close these and wait for `group` front matter support as creating the content PR is the easy part.

PR #40175, opened 2025-07-02
Browsers downsize the first fitting image in srcset

zakingslayerv22:
It informs that the browser will choose the first image that is bigger than the chosen slot size but it did not mention that the browser would scale it down to the required size.

### DIFF
--- a/files/en-us/web/html/guides/responsive_images/index.md
+++ b/files/en-us/web/html/guides/responsive_images/index.md
@@ -83,7 +83,7 @@ So, with these attributes in place, the browser will:
 1. Look at screen size, pixel density, zoom level, screen orientation, and network speed.
 2. Work out which media condition in the `sizes` list is the first one to be true.
 3. Look at the slot size given to that media query.
-4. Load the image referenced in the `srcset` list that has the same size as the slot. If there isn't an exact match for the display size, the browser will choose the first image that is bigger than the chosen slot size — and scale it down to fit.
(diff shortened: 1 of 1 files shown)


files/en-us/web/html/guides/responsive_images/index.md
```diff
@@ -83,7 +83,7 @@ So, with these attributes in place, the browser will:
 1. Look at screen size, pixel density, zoom level, screen orientation, and network speed.
 2. Work out which media condition in the `sizes` list is the first one to be true.
 3. Look at the slot size given to that media query.
-4. Load the image referenced in the `srcset` list that has the same size as the slot or, if there isn't one, the first image that is bigger than the chosen slot size.
+4. Load the image referenced in the `srcset` list that has the same size as the slot. If there isn't an exact match for the display size, the browser will choose the first image that is bigger than the chosen slot size — and scale it down to fit.
```
dipikabh:
```suggestion
4. Load the image referenced in the `srcset` list that has the same size as the slot. If there isn't an exact match for the display size, the browser will choose the first image that is bigger than the chosen slot size and scale it down to fit.
```

dipikabh (changes requested):

files/en-us/web/html/guides/responsive_images/index.md
dipikabh: Committing this suggestion since it's a minor fix

dipikabh (approved):
Thanks for submitting this update @zakingslayerv22 👍

PR #40177, opened 2025-07-02
Replace AudioContext.setSinkId glitch links

chrisdavidmills:
### Description

The MDN AudioContext.setSinkId documentation [see this, and similar pages](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/setSinkId) has several links to [this demo hosted on Glitch](https://set-sink-id.glitch.me/).

Because Glitch is going away, I've filed this PR to update all the remaining links (I removed a few) to point to the same demo on `dom-examples`.


bsmth (approved):
Great, tnx 👍🏻

PR #40179, opened 2025-07-03
Modernize CSS: consistently use rgb/hsl instead of rgba/hsla

Josh-Cena:
Our own user-facing documentation says:

> Note: The rgba() functional notation is an alias for rgb(). They are exactly equivalent. It is recommended to use rgb().

For reasons of consistency and terseness, we should fix the remaining few cases of `rgba`/`hsla`.


bsmth (approved):
LGTM, thanks 👍🏻

PR #40180, opened 2025-07-03
Remove redundant `1` after var names like `var1` where no `var2` exists

### LINKED ISSUE #40178: Weird variable naming like `var1` in examples where there's no `var2` to contrast with
### MDN URL

various e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertySymbols

_No response_

Grepping with `(const|let) \w+1 =\D+;\n+[^\d\n]+\n` to try and find examples (regex could be improved to reduce false positives/negatives).

Originally spotted on [`Object.getOwnPropertySymbols()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertySymbols) page, example from there:

```js
const object1 = {};
const a = Symbol("a");
const b = Symbol.for("b");

object1[a] = "localSymbol";
object1[b] = "globalSymbol";

const objectSymbols = Object.getOwnPropertySymbols(object1);

console.log(objectSymbols.length);
// Expected output: 2
```

Variable named something like `obj`, `object`, `target` etc. that doesn't imply there's an `object2` to compare with.

_No response_

_No response_

lionel-rowe:
### Description

Fixes https://github.com/mdn/content/issues/40178.

It's quite a big line diff, but the vast majority of the changes with `git diff --word-diff-regex="."` are just `... varName[-1-] ...`.

### Motivation

It's confusing as you're looking around for a contrastive `var2` variable or something else that would indicate the importance of the number `1`, but there often isn't anything.

### Additional details

N/A

Fixes https://github.com/mdn/content/issues/40178


Josh-Cena:
Thanks, it will take me a while to get to it (mainly because I need to check if the variables are referenced elsewhere), but I will eventually...

Josh-Cena (approved):
Looks awesome, thank you so much! I really appreciate your attention to detail and fixing everything en masse. Thanks for all the time you put into this.

PR #40184, opened 2025-07-03
feat: Improvements to HOL blocking article

bsmth:
### Description

The HOL blocking glossary entry is relatively new, but there's some small inaccuracies in there.

It's good to note that this happens over single TCP connections, as there's typically 6 parallel H1 connections that browsers make to origins by default.

I've also updated the slow start glossary to be a little more readable / have less redundant information.

### Motivation

Adding some more specifics for how HOL blocking happens in HTTP and the ways the different protocols try to mitigate it.


files/en-us/glossary/head_of_line_blocking/index.md
```diff
@@ -8,7 +9,16 @@ page-type: glossary-definition
...
+In HTTP/1.1, HOL blocking occurs when a client sends multiple requests to a {{glossary("server")}} over a single, persistent TCP connection.
+A response can be delayed due to network congestion, {{glossary("TCP slow start")}}, problems in transit, or other reasons.
+HTTP/1.1 requests are sent sequentially, so a delay in fully receiving a response blocks the next request-response exchanges.
+A mechanism called _HTTP pipelining_ tried to work around this, where multiple requests were sent off by a client without waiting for any responses.
+Pipelining proved tricky to implement in reality, so this mechanism is rarely used, if ever, and most browsers no longer support it.
+
+HTTP/2 addresses some problems in HTTP/1.1 through request _multiplexing_.
```
bsmth:
```suggestion
HTTP/2 addresses the HOL blocking problems in HTTP/1.1 through request _multiplexing_.
```

files/en-us/glossary/head_of_line_blocking/index.md
```diff
@@ -8,7 +9,16 @@ page-type: glossary-definition
 
 In computer networking, **head-of-line blocking** (_HOL blocking_) refers to a performance bottleneck that occurs when a queue of packets is held up by the first packet in the queue, even though other packets in the queue could be processed.
 
-In HTTP/1.1, HOL blocking can occur when a client sends multiple requests to a {{glossary("server")}} without waiting for the responses. The server processes the requests in order, but if the response to the first request is delayed, the responses to subsequent requests are also delayed. HTTP/2 addresses this issue through request multiplexing, eliminating HOL blocking in the application layer, but it still exists at the transport ({{glossary("TCP")}}) layer.
+In HTTP/1.1, HOL blocking occurs when a client sends multiple requests to a {{glossary("server")}} over a single, persistent TCP connection.
```
hamishwillee, line 12: So it _always_ occurs if you sent multiple requests in this case?

files/en-us/glossary/head_of_line_blocking/index.md
```diff
@@ -8,7 +9,16 @@ page-type: glossary-definition
...
+A response can be delayed due to network congestion, {{glossary("TCP slow start")}}, problems in transit, or other reasons.
+HTTP/1.1 requests are sent sequentially, so a delay in fully receiving a response blocks the next request-response exchanges.
+A mechanism called _HTTP pipelining_ tried to work around this, where multiple requests were sent off by a client without waiting for any responses.
+Pipelining proved tricky to implement in reality, so this mechanism is rarely used, if ever, and most browsers no longer support it.
+
+HTTP/2 addresses the HOL blocking problems in HTTP/1.1 through request _multiplexing_.
+Multiplexing means a single TCP connection can be used to interleave requests and responses in numbered streams.
+A client can send many requests over a single connection without waiting for earlier responses.
+This eliminates HOL blocking in the protocol layer, but it still exists at the transport ({{glossary("TCP")}}) layer.
```
hamishwillee: The first paragraph, in particular felt clunky. How about this (or some variant)

```suggestion
In HTTP/1.1, HOL blocking occurs when a client sends multiple requests to a {{glossary("server")}} over a single, persistent TCP connection, and one of the responses is delayed for any reason — such as network congestion, {{glossary("TCP slow start")}}, or, problems in transit.
HTTP/1.1 requests are sent sequentially, so a delay in fully receiving a response blocks the next request-response exchanges.

A mechanism called _HTTP pipelining_ tried to work around this, where multiple requests were sent off by a client without waiting for any responses.
Pipelining proved tricky to implement in reality, so this mechanism is rarely used, if ever, and most browsers no longer support it.

HTTP/2 addresses the HOL blocking problems in HTTP/1.1 through request _multiplexing_.
Multiplexing allows a single TCP connection to interleave requests and responses in numbered streams.
A client can send many requests over a single connection without waiting for earlier responses.
Note that while HOL has been fixed in HTTP, it is still a problem at the transport ({{glossary("TCP")}}) layer level.
```

files/en-us/glossary/tcp_slow_start/index.md
```diff
@@ -6,15 +6,16 @@ page-type: glossary-definition
 
 {{GlossarySidebar}}
 
-{{glossary('TCP')}} slow start helps buildup transmission speeds to the network's capabilities. It does this without initially knowing what those capabilities are and without creating congestion. {{glossary('TCP')}} slow start is an algorithm used to detect the available bandwidth for packet transmission, and balances the speed of a network connection. It prevents the appearance of network congestion whose capabilities are initially unknown, and slowly increases the volume of information diffused until the network's maximum capacity is found.
+{{glossary('TCP')}} slow start is an algorithm that detects the available network bandwidth for packet transmission and balances transmission speed according to the network's capabilities.
+The slow start mechanism prevents congestion when a network's capabilities are initially unknown, and slowly increases the volume of information diffused until the network's maximum capacity is found.
```
hamishwillee: Is diffused the right word?

files/en-us/glossary/tcp_slow_start/index.md
```diff
@@ -6,15 +6,16 @@ page-type: glossary-definition
...
 
-To implement TCP slow start, the congestion window (_cwnd_) sets an upper limit on the amount of data a source can transmit over the network before receiving an acknowledgment (ACK). The slow start threshold (_ssthresh_) determines the (de)activation of slow start. When a new connection is made, cwnd is initialized to one TCP data or acknowledgment packet, and waits for an acknowledgement, or ACK. When that ACK is received, the congestion window is incremented until the _cwnd_ is greater than _ssthresh_. Slow start also terminates when congestion is experienced.
+To implement TCP slow start, the congestion window (_cwnd_) sets an upper limit on the amount of data a source can transmit over the network before receiving an acknowledgment (ACK). The slow start threshold (_ssthresh_) determines the (de)activation of slow start. When a new connection is made, cwnd is initialized to one TCP data or acknowledgment packet, and waits for an ACK. When that ACK is received, the congestion window is incremented until the _cwnd_ is greater than _ssthresh_. Slow start also terminates when congestion is experienced.
```
hamishwillee: Do we need this level of detail?

I'm not sure about the algorithm. I think what happens is cwnd tells us how long before we expect a response. If we send data for more than cwnd without a response that indicates congestion. So what we're doing is increasing the amount we either reach or desired rate as set by the ssthresh, or we don't get a response, indicating congestion. 

If so, is this right?

```suggestion
To implement TCP slow start, the congestion window (_cwnd_) sets an upper limit on the amount of data that a source can transmit over the network before receiving an acknowledgment (ACK): if a response is not received in this window then the network is considered to be congested.
The slow start threshold (_ssthresh_) determines when slow start is activated and deactivated.
When a new connection is made, cwnd is initialized to one TCP data or acknowledgment packet, and waits for an ACK. When that ACK is received, the congestion window is incremented and the next packet is sent. The process continues until either the _cwnd_ is greater than _ssthresh_ or a packet is sent without a response in the congestion window, after which slow start is terminated.
```

If not, then we need to rewrite anyway, because it doesnt make sense.

files/en-us/glossary/tcp_slow_start/index.md
```diff
@@ -6,15 +6,16 @@ page-type: glossary-definition
...
 
 ## Congestion control
 
-Congestion itself is a state that happens within a network layer when the message traffic is too busy it slows the network response time. The server sends data in TCP packets, the user's client then confirms delivery by returning acknowledgements, or ACKs. The connection has a limited capacity depending on hardware and network conditions. If the server sends too many packets too quickly, they will be dropped. Meaning, there will be no acknowledgement. The server registers this as missing ACKs. Congestion control algorithms use this flow of sent packets and ACKs to determine a send rate.
+Congestion itself is a state that happens within a network layer when the message traffic is too busy it slows the network response time. The server sends data in TCP packets, the user's client then confirms delivery by returning ACKs. The connection has a limited capacity depending on hardware and network conditions. If the server sends too many packets too quickly, they will be dropped (meaning, there will be no acknowledgement). The server registers this as missing ACKs. Congestion control algorithms use this flow of sent packets and ACKs to determine a send rate.
```
hamishwillee: Rewrite. In particular the first sentence.

files/en-us/glossary/tcp_slow_start/index.md
bsmth: I read this on an explainer and thought it sounded nice, but we can just say "sent and received" :)

```suggestion
The slow start mechanism prevents congestion when a network's capabilities are initially unknown, and slowly increases the volume of information sent and received until the network's maximum capacity is found.
```

files/en-us/glossary/head_of_line_blocking/index.md
```diff
@@ -1,13 +1,23 @@
...
+short-title: HOL blocking
 slug: Glossary/Head_of_line_blocking
 page-type: glossary-definition
 sidebar: glossarysidebar
 ---
 
 In computer networking, **head-of-line blocking** (_HOL blocking_) refers to a performance bottleneck that occurs when a queue of packets is held up by the first packet in the queue, even though other packets in the queue could be processed.
 
-In HTTP/1.1, HOL blocking can occur when a client sends multiple requests to a {{glossary("server")}} without waiting for the responses. The server processes the requests in order, but if the response to the first request is delayed, the responses to subsequent requests are also delayed. HTTP/2 addresses this issue through request multiplexing, eliminating HOL blocking in the application layer, but it still exists at the transport ({{glossary("TCP")}}) layer.
+In HTTP/1.1, HOL blocking occurs when a client sends multiple requests to a {{glossary("server")}} over a single, persistent TCP connection, and one of the responses is delayed for any reason — such as network congestion, {{glossary("TCP slow start")}}, or, problems in transit.
```
bsmth:
```suggestion
In HTTP/1.1, HOL blocking occurs when a client sends multiple requests to a {{glossary("server")}} per TCP connection, and a response is delayed for any reason — such as network congestion, {{glossary("TCP slow start")}}, or, problems in transit.
```

The browser will use about 6 TCP connections per origin, so we want to say when one of these gets used for multiple requests, basically.

bsmth: Maybe there's a better reword of that

files/en-us/glossary/head_of_line_blocking/index.md
```diff
@@ -1,13 +1,23 @@
...
+HTTP/1.1 requests are sent sequentially, so a delay in fully receiving a response blocks the next request-response exchanges.
```
bsmth:
```suggestion
HTTP/1.1 requests are sent sequentially per TCP connection, so a delay in receiving a response blocks the next request-response exchange.
```

files/en-us/glossary/head_of_line_blocking/index.md
```diff
@@ -1,13 +1,23 @@
...
+
+A mechanism called _HTTP pipelining_ tried to work around this, where multiple requests were sent off by a client without waiting for any responses.
+Pipelining proved tricky to implement in reality, so this mechanism is rarely used, if ever, and most browsers no longer support it.
+
+HTTP/2 addresses the HOL blocking problems in HTTP/1.1 through request _multiplexing_.
+Multiplexing allows a single TCP connection to interleave requests and responses in numbered streams.
+A client can send many requests over a single connection without waiting for earlier responses.
+Note that while HOL has been fixed in HTTP, it is still a problem at the transport ({{glossary("TCP")}}) layer level.
```
bsmth:
```suggestion
Note that while HOL blocking has been fixed in HTTP/2, it is still a problem at the transport ({{glossary("TCP")}}) layer level.
```

files/en-us/glossary/tcp_slow_start/index.md
```diff
@@ -5,15 +5,16 @@ page-type: glossary-definition
...
```
bsmth: Maybe this is better? We're just short of a new glossary on "congestion avoidance"

```suggestion
Network congestion is what happens when the amount of data sent exceeds the network's capacity (the maximum amount of information that can be transferred at any time).
This can occur when a node on the network is sent more data than it can process, there's a sudden burst of data across the network, a deliberate flooding of information from participants, among other reasons.
When a network is congested, machines on that network experience degraded quality of service, like slow data transfers and dropped packets. 
The slow start algorithm is one major part of TCP congestion control, which helps prevent networks from becoming overwhelmed by traffic from senders.
```

files/en-us/glossary/head_of_line_blocking/index.md
hamishwillee: I'm prefer the original, but how about a compromise (the per TCP does not work for me).

```suggestion
In HTTP/1.1, HOL blocking occurs when a client sends multiple requests to a {{glossary("server")}} over a particular TCP connection, and a response on that connection is delayed for any reason — such as network congestion, {{glossary("TCP slow start")}}, or, problems in transit.
```

files/en-us/glossary/tcp_slow_start/index.md
```diff
@@ -5,15 +5,16 @@ page-type: glossary-definition
...
 
-To implement TCP slow start, the congestion window (_cwnd_) sets an upper limit on the amount of data a source can transmit over the network before receiving an acknowledgment (ACK). The slow start threshold (_ssthresh_) determines the (de)activation of slow start. When a new connection is made, cwnd is initialized to one TCP data or acknowledgment packet, and waits for an acknowledgement, or ACK. When that ACK is received, the congestion window is incremented until the _cwnd_ is greater than _ssthresh_. Slow start also terminates when congestion is experienced.
+To implement TCP slow start, the congestion window (_cwnd_) sets an upper limit on the amount of data a source can transmit over the network before receiving an acknowledgment (ACK). The slow start threshold (_ssthresh_) determines the (de)activation of slow start. When a new connection is made, cwnd is initialized to one TCP data or acknowledgment packet, and waits for an ACK. When that ACK is received, the congestion window is incremented until the _cwnd_ is greater than _ssthresh_. Slow start also terminates when congestion is experienced.
 
 ## Congestion control
```
hamishwillee:
```suggestion
## Congestion avoidance
```

files/en-us/glossary/tcp_slow_start/index.md
hamishwillee: Much better. 

Here's a tweak on your version for consideration. In particular the second sentence was unnecessary/unwieldy.

```suggestion
Network congestion occurs when more data is sent than the network is able to transfer between nodes at a particular time.
When a network is congested, machines on the network experience degraded quality of service, including slow data transfers and dropped packets.
The slow start algorithm is one major part of TCP congestion control, as it ensures that new connections progressively increase data rates and that those rates do not exceed configured congestion limits.
```

files/en-us/glossary/head_of_line_blocking/index.md
bsmth: Good, let's go with that 👍🏻

files/en-us/glossary/tcp_slow_start/index.md
bsmth: nice, tnx 👍🏻

files/en-us/glossary/tcp_slow_start/index.md
bsmth: I found these useful:

* https://www.researchgate.net/figure/TCP-Slow-Start-and-Congestion-Avoidance-phase_fig1_332436793
* https://www.noction.com/blog/tcp-transmission-control-protocol-congestion-control


So during slow start:
* `cwnd` grows exponentially for each ACK
* When we hit `ssthresh`, congestion avoidance begins and `cwnd` grows linearly instead
* If there's packet loss (timeout or duplicates), `ssthresh` set to `cwnd`/2 and `cwnd` is reset
* slow start begins again


I think we can say something like this for the:

```suggestion
During TCP slow start, the congestion window (_cwnd_) sets an upper limit on the amount of data that a source can send over the network before receiving an acknowledgment (ACK).
If a response is not received in this window then the network is considered to be congested.
The slow start threshold (_ssthresh_) determines when slow start is activated and deactivated.
When a new connection is made, the cwnd is initialized, and when an ACK is received, the congestion window is increased and the next packet is sent.
The process continues until either the _cwnd_ is greater than _ssthresh_ or a packet is sent without a response in the congestion window, after which slow start is terminated and congestion avoidance begins.
```

Apparently cubic is the default on a lot of systems since quite a while:

https://en.wikipedia.org/wiki/CUBIC_TCP

pepelsbey (approved):
Looks good to me! I don't have a strong opinion on the level of details. Thanks! 🙃

files/en-us/glossary/tcp_slow_start/index.md
hamishwillee: The problem is that the last sentence might be true, but it isn't helpful without all that information you provided above. Reading it as it is written you'd jump forward to the section on collision avoidance and assume that this text was wrong - because the following sections says that slow start is one of the tools you can use for congestion avoidance.

I think the problem is ... the whole thing. We're trying to spot fix when we need to start again. Slow start is part of either "the" or "A" congestion control algorithm. It really doesn't make sense unless we talk about the whole thing.

Perhaps we can structure  this https://www.researchgate.net/figure/TCP-Slow-Start-and-Congestion-Avoidance-phase_fig1_332436793 and even add a similar diagram?

hamishwillee: Do you think we might do something like this for the whole thing (i.e. replacing all the current text).
I mean really, how much of our audience needs to know more than this. Though the diagram would be nice.

----


**TCP slow start** is the first phase of many common {{glossary('TCP')}} congestion control algorithms.
It is designed to rapidly determine a network's capacity at the beginning of a connection, after a period of inactivity, or after detecting congestion, without itself causing congestion.

When using the algorithm, data rates are set using a congestion window (_cwnd_).
This window defines the upper limit on how much data can be sent in which acknowledgment (ACK) packets are still being reliably received.
If responses are not received with the current cwnd, then the network is considered to be congested.

In the slow start phase the cwnd is initially set to a low value and increased exponentially until a configured slow start threshold (_ssthresh_) is reached.
In the second phase, "congestion avoidance", the cwnd is ramped up linearly until congestion is detected.
If at any point congestion is detected, _ssthresh_ is set to half the level at which congestion was detected and the slow start phase is restarted.

bsmth: I like this direction, so yes, I think you're right to go with your first impression to remove the specifics cwnd etc. I like the intro sentences we have, though, so we could cut even more details to summarize with a variation of this:

>{{glossary('TCP')}} slow start is an algorithm that detects the available network bandwidth for packet transfer and balances transmission speed according to the network's capabilities.
The slow start mechanism prevents congestion when a network's capabilities are initially unknown, and slowly increases the volume of information sent until the network's maximum capacity is found.
>
>Network congestion occurs when more data is sent than the network is able to transfer between nodes at any given time.
When a network is congested, machines on the network experience degraded quality of service, including slow data transfers and dropped packets.
The slow start algorithm is a major part of TCP congestion control, as it ensures that new connections progressively increase data rates and that those rates don't overwhelm other nodes.

Some points to keep, IMO:

* "prevents congestion when a network's capabilities are initially unknown"
* "slowly increases the volume of information sent until the network's maximum capacity is found"
* congestion effects are "degraded quality of service, including slow data transfers and dropped packets"
* "slow start algorithm is a major part of TCP congestion control,"

hamishwillee: @bsmth I think this makes sense, or at least is a better baseline for future discussion. I have merged this in verbatim and if I have further comments I'll add them.

files/en-us/glossary/tcp_slow_start/index.md
```diff
@@ -5,15 +5,15 @@ page-type: glossary-definition
 sidebar: glossarysidebar
 ---
 
-{{glossary('TCP')}} slow start helps buildup transmission speeds to the network's capabilities. It does this without initially knowing what those capabilities are and without creating congestion. {{glossary('TCP')}} slow start is an algorithm used to detect the available bandwidth for packet transmission, and balances the speed of a network connection. It prevents the appearance of network congestion whose capabilities are initially unknown, and slowly increases the volume of information diffused until the network's maximum capacity is found.
+{{glossary('TCP')}} slow start is an algorithm that detects the available network bandwidth for packet transfer and balances transmission speed according to the network's capabilities.
```
hamishwillee: Propose discover rather than detect - it's a process.

```suggestion
{{glossary('TCP')}} slow start is an algorithm that helps discover the available network bandwidth for packet transfer, and balance transmission speed according to the network's capabilities.
```

files/en-us/glossary/tcp_slow_start/index.md
```diff
@@ -5,15 +5,15 @@ page-type: glossary-definition
...
+The slow start mechanism prevents congestion when a network's capabilities are initially unknown, and slowly increases the volume of information sent until the network's maximum capacity is found.
```
hamishwillee: While it's called slow start, it actually isn't slow. Also, this is where I like my sentence on where this is useful - it is not just on when they are initially unknown - if you get congestion then what knew is no longer valid.

```suggestion
The mechanism rapidly increases the volume of information sent from a very low level to a threshold level. If congestion is detected the threshold is reset to a much reduced level and the slow start process starts again.
This prevents congestion when a network's capabilities are unknown or exceeded — at the beginning of a connection, after a period of inactivity, or after detecting congestion — without itself causing congestion.
```

files/en-us/glossary/tcp_slow_start/index.md
bsmth: This looks great now, Hamish. I think the back and forth was worth it for the results 👏🏻

bsmth:
Thanks @hamishwillee for the thoughtful reviews, I'll merge this one 🎉

files/en-us/glossary/tcp_slow_start/index.md
hamishwillee: Phew. Thanks for your patience.

PR #40186, opened 2025-07-03
Remove stream visualizer glitch demo links

chrisdavidmills:
### Description

The MDN [Streams API documentation](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) has several pages that link to [this demo hosted on Glitch](https://whatwg-stream-visualizer.glitch.me/).

Because Glitch is going away, I've decided to remove these links. I don't see this demo as being vital to the documentation, plus it is fairly complicated, and I don't think we should take on the role of maintaining it.

If it gets moved somewhere else, we can always link to the new location in the future.


bsmth (approved):
Thank you!

PR #40207, opened 2025-07-04
remove fetch request streams glitch demo links

chrisdavidmills:
### Description

The MDN [Encoding API documentation](https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API) has a couple of pages that link to [this demo hosted on Glitch](https://fetch-request-stream.glitch.me/).

Because Glitch is going away, and this demo no longer works, I am removing these links from MDN. I've added a link to the DCC article that explains some of this stuff, as it is pretty useful.

This PR updates all the remaining links to point to the new location


bsmth (approved):
Nice one!

PR #40209, opened 2025-07-04
Scrimba survey

argl:
### Description

Adds markup to the embedded scrims in the learning area to show the mini-survey.

### Motivation

We want to know what people think about scrims as learning challenges.


pepelsbey (approved):
Looks good! Thank you :)

PR #40210, opened 2025-07-04
Remove dnd demo glitch link

chrisdavidmills:
### Description

This PR removes a glitch link from our Drag and Drop API docs. We link to enough other demos for it not to be essential, plus, I'm not sure about moving it to dom-examples. It seems to originate from https://css-tricks.com/creating-a-parking-game-with-the-html-drag-and-drop-api/, but the version in that post doesn't seem to work.


bsmth (approved):
Tnx

PR #40211, opened 2025-07-04
Update file system API glitch demo links

chrisdavidmills:
### Description

Our [File System API docs](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API) have several Glitch demo links in them.

Because Glitch is going away, I have moved these to the dom-examples repo: see https://github.com/mdn/dom-examples/pull/323

This PR updates the glitch links to the new locations, and updates the links to the FileSystemObserver demo, which got moved during the dom-examples update.


dipikabh (approved):
Looks good, thank you!

PR #40216, opened 2025-07-05
CSS best practice: always specify generic font family as fallback

Josh-Cena:
We should always specify a generic fallback so the displayed font is not Times New Roman. Even when the font-face is explicitly declared in the code, it may fail to load.


bsmth (approved):
Thank you

PR #40217, opened 2025-07-05
fix: set the condition descriptions to `ul` format

yin1999:
### Description

set the condition descriptions to `ul` format


estelle (approved):
Thank you! 🎉

PR #40218, opened 2025-07-05
doc[http/guides/authentication/index.md]: Removed outdated description about firefox

samuel871211:
### Description

Test Locally with firefox 140.0.2, firefox does not show any warning message.

https://github.com/user-attachments/assets/2c284b21-d218-4999-b225-99ac272267cd

```
Firefox will warn the user with a prompt "You are about to log in to the site www.example.com with the username username, but the website does not require authentication. This may be an attempt to trick you."
```

This seems outdated, but I can't find the exact version firefox changed this behavior.

### DIFF
--- a/files/en-us/web/http/guides/authentication/index.md
+++ b/files/en-us/web/http/guides/authentication/index.md
@@ -153,8 +153,7 @@ Many clients also let you avoid the login prompt by using an encoded URL contain
 https://username:password@www.example.com/
 ```
 
-**The use of these URLs is deprecated**.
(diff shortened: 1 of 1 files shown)


hamishwillee (approved):
I can't find the bug in which it changed either (the string is still in the firefox repo here https://searchfox.org/mozilla-central/source/netwerk/locales/en-US/necko.properties)

IMO this is a blatant security breach, and I can see the URL is not sent for FF and Chrome in the browser and wireshark, so I'm pretty sure this is correct.

I have edited this a bit, because the part before the bit you changed was hinting you could still do this. At some point we probably should delete the whole section, but probably useful to have a period where we note that it can't be done anymore.


Thank you.

PR #40221, opened 2025-07-06
doc[http/guides/cors/index.md]: fix typo

samuel871211:
### Description

Before
but the `Cookie` header in the response is sent by `bar.other`

After
but the `Set-Cookie` header in the response is sent by `bar.other`


hamishwillee (approved):
Thank you!

PR #40233, opened 2025-07-07
Remove Houdini Glitch links

chrisdavidmills:
### Description

Because Glitch is going away imminently, I'm using this PR to remove the CSS Houdini Glitch resources. They are not essential to the docs, and we can always put them back at a later date if they get moved somewhere else.


bsmth (approved):
Great, tnx 👍🏻

PR #40234, opened 2025-07-07
Handle various Chrome demo glitch links

chrisdavidmills:
### Description

This PR handles several links to Chrome dev rel demos that were on Glitch. Because Glitch is going away imminently, we need to do something with the links to avoid 404s on the site.

I've decided to start batching up a few at a time to cut down on the number of PRs.

Specifically, this PR does the following:

- Background Fetch demo - I've instead linked to the article where the demo comes from, which contains a bunch more useful code snippets. They'll hopefully fix the demo link soon. I didn't just remove this one because the demo seems quite important/valuable.
- getDisplayMedia() Screen Sharing controls demo - this one has been moved, so the link is updated to point to the new location.
- Web OTP demo - this one has been moved, so the link is updated to point to the new location.
- Element capture/Region capture demos. I just removed these ones - we already have full demos, and we already link to the article that they originate from.

### DIFF
--- a/files/en-us/web/api/background_fetch_api/index.md
+++ b/files/en-us/web/api/background_fetch_api/index.md
@@ -86,7 +86,7 @@ navigator.serviceWorker.ready.then(async (swReg) => {
 });
 ```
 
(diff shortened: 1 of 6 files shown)


bsmth (approved):
Good, thanks 👍🏻

PR #40235, opened 2025-07-07
Handle a couple more glitch mentions

chrisdavidmills:
### Description

This PR handles a couple more Glitch mentions that I found:

- Window Management API - I've removed the link to the glitch demo here, as there are plenty of other demos linked to
- WebExtensions, `web_accessible_resources` manifest field - these instances are example URLs, which could be basically anything, so I changed them to something else suitable.


bsmth (approved):
Thanks!

PR #40237, opened 2025-07-07
feat: swap out Glitch URL for Render alternative

bsmth:
### Description

As a stopgap measure, I'm deploying the webrtc example on Render.

- https://github.com/bsmth/examples/tree/main/webrtc-from-chat

### Motivation

Glitch are shutting down their services, so these need a new home.

### Additional details

We should move these under the MDN GitHub org and use other hosting, but this keeps the demo alive during the Glitch switchover.


chrisdavidmills (approved):
The changeover LGTM! Thanks @bsmth.

PR #40238, opened 2025-07-07
chore: Remove TODO note from webxr guide

bsmth:
### Description

I encountered a strange-looking TODO that's made it into the content. I presume it's a malformed HTML comment, but I think we're better off without it on the page.

### Motivation

Keeping quality standards high.


Josh-Cena (approved):
This note is probably from the wiki era, so someone _did_ intend for it to be rendered so readers could fix  the alluded "problem". We don't know what it is though.

bsmth:
Yeah, it's hard to tell what's wrong here. Thanks anyway 👍🏻

PR #40246, opened 2025-07-08
Fix all broken anchors in FF release notes

Josh-Cena:
I'm trying to fix all broken links. Anchors are significantly easier to fix than broken links, so this does the easier part.


bsmth (approved):
Looks good to me, thanks!

PR #40250, opened 2025-07-08
Remove last Glitch links

chrisdavidmills:
### Description

Since Glitch is going away, we need to remove/replace/update all the links we have to Glitch-hosted demos.

This PR removes all the remaining Glitch links we had at the end of the project. Most of these links have just been deleted, as they pointed to demos that we didn't have the time to host somewhere else. Where possible, I've tried to link to other articles that provide code examples.

We have made a list of all such pages and demos, in the hope that we can add links back to ported versions in the future.


bsmth (approved):
Thanks a lot, I think this is a good call for the moment.

PR #40251, opened 2025-07-08
fix(breadcrumb): double reference in breadcrumbs

Rumyra:
### Description

Adds a short title to the index of the web folder - this should be picked up and breadcrumbs will then read 'Web > Technology > Reference" instead of "Reference > Technology > Reference"


bsmth (approved):
Super, thanks!

PR #40263, opened 2025-07-08
Clarify `execCommand` valid use cases without alternatives.

### LINKED ISSUE #40245: Amend deprecation notice for `document.execCommand` to indicate that there are still valid use cases given that there's still no viable alternative
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand

Top

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

That's a bit misleading, especially the "Be aware that this feature may cease to work at any time" part (removing `execCommand` will never happen as it would be massively web-breaking).

Perhaps a custom warning instead, noting that behavior may differ somewhat between browsers, but also noting that there are still valid use cases given that there's still no viable alternative.

[javascript - execCommand() is now obsolete, what's the alternative? - Stack Overflow](https://stackoverflow.com/questions/70831583/execcommand-is-now-obsolete-whats-the-alternative)

> Year 2022–2025 answer: The `execCommand()` is officially obsolete/deprecated but there's no alternative. So if you must have rich text support, you have to keep using `execCommand()` and figure out yourself what actually works with browsers that you want to support.
>
> Note that real world user agents (browsers such as Chrome, Firefox and Safari) cannot drop support for `execCommand()` because so many services require support for it.
>
> ...
>
> All the current standardization efforts (Input Events 2, Clipboard API) fail to even try to cover the features that execCommand() actually does (for example, undo/redo, actually changing content within the selection range).
>
> ...
>
> For background information: [medium.com/content-uneditable/…](https://medium.com/content-uneditable/fixing-contenteditable-1a9a5073c35d)

_No response_

younisdev:
### Description

Clarifies  that`execCommand`  still has valid use cases without viable alternatives currently available.
### Motivation

These changes inform readers about the lack of current alternatives for some `execCommand` commands.
### Additional details

None.

Fixes #40245

### DIFF
--- a/files/en-us/web/api/document/execcommand/index.md
+++ b/files/en-us/web/api/document/execcommand/index.md
@@ -10,6 +10,9 @@ browser-compat: api.Document.execCommand
 
 {{ApiRef("DOM")}}{{deprecated_header}}
 
+> [!NOTE]
+> Although the **`document.execCommand`** method is deprecated and no longer recommended for use in new projects, there are still some valid use cases that have not yet been replaced, where it remains the only practical option. For example, implementing undo/redo, changing content within the selection range, or applying rich text formatting currently have no alternatives. Developers should use it carefully and ensure compatibility across major browsers.
+
 The **`execCommand`** method implements multiple different commands. Some of them provide access to the clipboard, while others are for editing [form inputs](/en-US/docs/Web/HTML/Reference/Elements/input), [`contenteditable`](/en-US/docs/Web/HTML/Reference/Global_attributes/contenteditable) elements or entire documents (when switched to [design mode](/en-US/docs/Web/API/Document/designMode)).
 
 To access the clipboard, the newer [Clipboard API](/en-US/docs/Web/API/Clipboard_API) is recommended over `execCommand()`. However, there is no replacement for the editing commands: unlike direct DOM manipulation, modifications performed by `execCommand()` preserve the undo buffer (edit history).
--- a/files/en-us/web/api/document/execcommand/index.md
+++ b/files/en-us/web/api/document/execcommand/index.md
@@ -11,11 +11,11 @@ browser-compat: api.Document.execCommand
 {{ApiRef("DOM")}}{{deprecated_header}}
 
 > [!NOTE]
-> Although the `execCommand` method is deprecated, there are still some valid use cases that do not yet have viable alternatives, such as implementing undo/redo, changing content within the selection range, or applying rich-text formatting. Developers should use it carefully and test to ensure cross-browser compatibility.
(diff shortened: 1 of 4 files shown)


younisdev:
Hey! Just to clarify, I accidentally opened the previous PR from my main branch, that's why i closed it and made this one from the correct branch. Thanks!

files/en-us/web/api/document/execcommand/index.md
```diff
@@ -10,6 +10,9 @@ browser-compat: api.Document.execCommand
 
 {{ApiRef("DOM")}}{{deprecated_header}}
 
+> [!NOTE]
+> Although the `execCommand` method is deprecated and no longer recommended for use in new projects, there are still some valid use cases that have not yet been replaced, where it remains the only practical option. For example, implementing undo/redo, changing content within the selection range, or applying rich text formatting currently have no alternatives. Developers should use it carefully and ensure compatibility across major browsers.
```
lionel-rowe: > "Although the `execCommand` method is ... no longer recommended for use in new projects, there are still some valid use cases ... where it remains the only practical option"

This doesn't make sense to me. Either it shouldn't be used in new projects or there are valid use cases, both can't be true at the same time.

```suggestion
> Although the `execCommand` method is deprecated, there are still some valid use cases that do not yet have viable alternatives, such as implementing undo/redo, changing content within the selection range, or applying rich-text formatting. Developers should use it carefully and test to ensure cross-browser compatibility.
```

Josh-Cena: Both can be true. You can pretend that this feature simply cannot be implemented on the web—for example if the web MIDI API doesn't exist then you simply cannot operate MIDI in a  web app.

lionel-rowe: > You can pretend that this feature simply cannot be implemented on the web

So there are valid use cases but only in existing projects, new projects should simply give up on trying to interoperate with the browser's undo stack etc? I don't follow this logic.

Josh-Cena: Yeah exactly. Pretend it doesn't exist. Lots of features can be deprecated without replacement, which means this capability is removed.

Josh-Cena (approved):
I've applied @lionel-rowe's suggestion, made some changes to deduplicate, and propagated this note to other related pages. I also filed https://github.com/mdn/browser-compat-data/issues/27456 to get the status right. Thanks both!

PR #40264, opened 2025-07-08
Fixing confusion around id versus name

getify:
### Description

Clarifying the roles of `id` (aka "username") vs `name` (aka "display name") in this credential type, aligning them to typical expectations for username/password credentials.

### Motivation

The existing documentation (and sample code) implies confusing/mixed things about the roles of `id` versus `name` in the creation and storage of this credential type. Some parts of this page were accurate (to spec) but other places implied an opposite relationship that contradicts the spec (and browser behavior).

I misunderstood after reading this page, and only figured it out after live testing -- including actually using `navigator.credentials.store()` to prompt to store the credential in the browser -- and reading the specification.

So I want to fix the page so other's aren't misled like I was. :)

### Additional details

n/a

n/a

### DIFF
--- a/files/en-us/web/api/passwordcredentialinit/index.md
+++ b/files/en-us/web/api/passwordcredentialinit/index.md
@@ -29,9 +29,9 @@ The `origin` property is set to the origin of the document the {{domxref("HTMLFo
 - `iconURL` {{optional_inline}}
   - : A string representing the URL of an icon or avatar to be associated with the credential.
 - `id`
-  - : A string representing a unique ID for the credential (e.g., the _username_ in a typical username/password credential pair).
+  - : A string representing the username portion of the username/password combination.
 - `name` {{optional_inline}}
-  - : A string representing the credential's display name (aka, nickname or label; may not be used by the browser).
+  - : A string representing a human-understandable name associated with the credential, intended to help the user select this credential in a user interface.
 - `origin`
   - : A string representing the credential's origin. {{domxref("PasswordCredential")}} objects are origin-bound, which means that they will only be usable on the specified origin they were intended to be used on.
 - `password`
--- a/files/en-us/web/api/passwordcredentialinit/index.md
+++ b/files/en-us/web/api/passwordcredentialinit/index.md
@@ -29,9 +29,9 @@ The `origin` property is set to the origin of the document the {{domxref("HTMLFo
 - `iconURL` {{optional_inline}}
   - : A string representing the URL of an icon or avatar to be associated with the credential.
 - `id`
-  - : A string representing a unique ID for the credential.
+  - : A string representing a unique ID for the credential (e.g., the _username_ in a typical username/password credential pair).
 - `name` {{optional_inline}}
-  - : A string representing the credential username.
+  - : A string representing the credential's display name (aka, nickname or label; may not be used by the browser).
 - `origin`
   - : A string representing the credential's origin. {{domxref("PasswordCredential")}} objects are origin-bound, which means that they will only be usable on the specified origin they were intended to be used on.
 - `password`
@@ -45,8 +45,8 @@ This example constructs an object literal to initialize a password credential.
 
 ```js
 const credInit = {
-  id: "1234",
-  name: "Serpentina",
+  id: "serp1234",  // "username" in a typical username/password pair
+  name: "Serpentina",  // display name for credential
   origin: "https://example.org",
   password: "the last visible dog",
 };
@@ -59,6 +59,8 @@ makeCredential.addEventListener("click", async () => {
   });
   console.log(cred.name);
   // Serpentina
+  console.log(cred.id);
+  // serp1234
   console.log(cred.password);
   // the last visible dog
 });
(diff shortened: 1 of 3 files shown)


getify:
I should also mention that I would suggest this page should include the same "Browser Compatibility" table as on the [PasswordCredential](https://developer.mozilla.org/en-US/docs/Web/API/PasswordCredential) page, as these two pages are inherently tied/related.

I didn't include that in the PR (mostly because I don't know how), but it also might imply other higher-level workflow issues (updates, etc) which I'm not privy to.

getify:
ping @wbamberg

files/en-us/web/api/passwordcredentialinit/index.md
```diff
@@ -29,9 +29,9 @@ The `origin` property is set to the origin of the document the {{domxref("HTMLFo
 - `iconURL` {{optional_inline}}
   - : A string representing the URL of an icon or avatar to be associated with the credential.
 - `id`
-  - : A string representing a unique ID for the credential.
+  - : A string representing a unique ID for the credential (e.g., the _username_ in a typical username/password credential pair).
```
wbamberg: Can we just say something like:

```suggestion
  - : A string representing the username portion of the username/password combination.
```

? This aligns reasonably well with https://developer.mozilla.org/en-US/docs/Web/API/Credential/id :

> This might be a GUID, **username**, or email address, or some other value, depending on the type of credential.

files/en-us/web/api/passwordcredentialinit/index.md
```diff
@@ -29,9 +29,9 @@ The `origin` property is set to the origin of the document the {{domxref("HTMLFo
...
 - `name` {{optional_inline}}
-  - : A string representing the credential username.
+  - : A string representing the credential's display name (aka, nickname or label; may not be used by the browser).
```
wbamberg:
```suggestion
  - : A string representing a human-understandable name associated with the credential, intended to help the user select this credential in a user interface.
```

Or something like that? Based more or less on the [spec language](https://w3c.github.io/webappsec-credential-management/#dom-credentialuserdata-name):

> A name associated with the credential, intended as a human-understandable public name for display in a [credential chooser](https://w3c.github.io/webappsec-credential-management/#credential-chooser).

wbamberg (commented):
Thank you (and sorry for the delay, I was on vacation).

It looks like https://developer.mozilla.org/en-US/docs/Web/API/PasswordCredential/PasswordCredential has a similar problem, and the example in https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/create#creating_a_password_credential needs a similar update to the example you've updated here. Would you mind taking care of these, too?

Josh-Cena:
@getify?

Josh-Cena:
Closing for lack of response

getify:
here is a response: that sucks.

Josh-Cena:
Are you willing to address the reviews above? If so I can reopen the issue. You've been notified twice but no replies were received, and we can't keep an issue open forever.

getify:
here's my perspective:

* the review comments you left were not *objectionable* to me -- meaning I don't *disagree* with them -- but... they're also not how *I think* they should have been worded. So I was neutral on your review comments.

    I left them alone as I didn't want to cause further debate based on my disagreement. I assumed that if your comments weren't just suggestions but something you felt strongly was better, you would land those changes with additional commits on this PR.
    
    In retrospect, I should have just said as much, rather than leaving no comments and leaving the PR in limbo. Sorry for the ambiguity.
    
* the request for addressing similar changes to another page was not unreasonable, but... by the time you had made that request, I had moved on from this task and was quite busy with other things. as such, I didn't take back up the work.

    I again assumed, if my schedule stayed busy and I was unable to pick back up that work, you (or someone else) would prioritize it and land those commits (in this PR or another).
    
    Again, I guess I wish I had just said as much. Sorry for the ambiguity. But I wasn't sure if I was going to get free time "soon" or not, so I was just sort of leaving things for others to pick up where I had left off.

----

All that said, my frustrated comment earlier this evening was a reaction to the choice to close the PR rather than to either land it as-is (and queue up any desired additional work in follow-up issues/PRs) OR to simply make whatever desired changes were appropriate as additional commits and then land it.

Closing an old issue which gets no responses is understandable. But IMO, closing a PR signals that the PR as it stands is unacceptable/deficient. If that's how you felt, so be it. But based on what you'd said, I thought what was in the PR already held enough merit that it deserved to be (at least part of) what landed.

wbamberg:
> I assumed that if your comments weren't just suggestions but something you felt strongly was better, you would land those changes with additional commits on this PR.

I have got abuse from people when I did that to their PRs. So I don't do it any more. I agree that this was a worthwhile PR, and if you don't feel strongly enough about wording to want to argue I am happy to finish it off (although I have been told that the passwords bit of the CM API is not much used, so it doesn't seem that high a priority. But maybe I was misinformed).

getify:
> I have got abuse from people when I did that to their PRs

Well, I'm sorry that kind of thing happens. And I'm sorry if my snarky comment contributed to that. Again, I should have just commented a long while back that my priorities had shifted and I didn't have the bandwidth to jump back on it. Apologies for that omission.

>  if you don't feel strongly enough about wording

I don't, I leave it to your best judgement whichever version of wording you think is best. This was really just a quick "best effort" attempt on my part to help others not get tripped up on what took me several extra hours of fiddling to work out.

Josh-Cena:
Also, FWIW: it's not uncommon for projects to require all review comments to be addressed by the author. Firefox does this for example: the reviewer is definitely not going to spend time updating the PR. If I am slightly more familiar with this topic I will very happily address Will's reviews for you, but I don't know about web auth at all. Thanks for the updates.

wbamberg (approved):
👍 thank you for the fixes!

PR #40265, opened 2025-07-08
Fix <pre> typo

Quetute:
`<pre>` First example from Apollinaire poem is missing a J. 
See source here https://www.paris-a-nu.fr/wp-content/uploads/2018/04/guillaume_apollinaire_calligramme.jpg


Josh-Cena (approved):
Good catch, thank you

PR #40268, opened 2025-07-09
doc[glossary/site/index.md]: Improve wording for same-site example

samuel871211:
### Description

Reworded explanation to use "despite" instead of "because" to more accurately reflect that different ports do not affect same-site status.

Before:

These are the same site because the port is not relevant:

- `https://example.com:8080`
- `https://example.com`

After:

These are the same site despite using different ports:

- `https://example.com:8080`
- `https://example.com`

### DIFF
--- a/files/en-us/glossary/site/index.md
+++ b/files/en-us/glossary/site/index.md
@@ -26,7 +26,7 @@ These are the same site because the registrable domain of `mozilla.org` is the s
 - `https://developer.mozilla.org/en-US/docs/`
 - `https://support.mozilla.org/en-US/`
 
-These are the same site because the port is not relevant:
+These are the same site despite using different ports:
 
 - `https://example.com:8080`
 - `https://example.com`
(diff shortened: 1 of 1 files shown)


bsmth:
Just a side note @samuel871211 - please try to fill out the pull request template to help reviewers. Adding in some details to "Description" and "Motivation" can really make a big difference, and helps things go faster for everyone. Thanks a lot!

files/en-us/glossary/site/index.md
```diff
@@ -26,7 +26,7 @@ These are the same site because the registrable domain of `mozilla.org` is the s
 - `https://developer.mozilla.org/en-US/docs/`
 - `https://support.mozilla.org/en-US/`
 
-These are the same site because the port is not relevant:
+These are the same site despite using different ports:
```
dipikabh: With "despite", it feels like we're shifting the focus to the difference rather than explaining why they're considered the same. It's doesn't read like a strong improvement.
Instead we could be more direct (and emphasize that "different ports do not affect")
```suggestion
These are considered the same site because the port number is ignored when determining the site:
```

dipikabh (changes requested):

dipikabh (approved):
Thanks!

PR #40274, opened 2025-07-09
Add info about calling `remove()` on nodes that have no parent node

lionel-rowe:
### Description

Add info about calling `remove()` on nodes that have no parent node.

Also tightened up wording:
* `Element.remove` doesn't always "remove the element from the DOM" (e.g. if the parent is already detached from the DOM)
* `CharacterData.remove` doesn't "remove the text contained in the node", rather it removes the text node it from its parent, without altering its internal text data.

### Motivation

Currently this info is missing.

### Additional details

https://dom.spec.whatwg.org/#dom-childnode-remove
1. If [this](https://webidl.spec.whatwg.org/#this)’s [parent](https://dom.spec.whatwg.org/#concept-tree-parent) is null, then return.
2. [Remove](https://dom.spec.whatwg.org/#concept-node-remove) [this](https://webidl.spec.whatwg.org/#this).

N/A

### DIFF
--- a/files/en-us/web/api/characterdata/remove/index.md
+++ b/files/en-us/web/api/characterdata/remove/index.md
@@ -8,7 +8,8 @@ browser-compat: api.CharacterData.remove
 
 {{APIRef("DOM")}}
 
-The **`remove()`** method of the {{domxref("CharacterData")}} removes the text contained in the node.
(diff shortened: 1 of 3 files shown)


Josh-Cena (approved):
Nice addition. Thanks for propagating it to all other pages!

PR #40275, opened 2025-07-09
Fix the "Exceptions" section of the VideoEncoder.encode() page

okyanusoz:
### Description

The "Exceptions" section of the page for the new Web Codecs' API's VideoEncoder's encode() method (https://developer.mozilla.org/en-US/docs/Web/API/VideoEncoder/encode) is incorrect, specifically the description for DataError. The [official spec](https://w3c.github.io/webcodecs/#dom-videoencoder-encode) specifies that this is thrown when the "active orientation" does not match. The original description is likely a leftover from the VideoDecoder.decode() method.

### Motivation

The description for the DataError exception on that page is incorrect. This could confuse readers who are getting started with this new API.

### Additional details

[Official spec page](https://w3c.github.io/webcodecs/#dom-videoencoder-encode)

[The page in question](https://developer.mozilla.org/en-US/docs/Web/API/VideoEncoder/encode)

N/A (as far as I know)


files/en-us/web/api/videoencoder/encode/index.md
```diff
@@ -52,7 +52,7 @@ None ({{jsxref("undefined")}}).
 - `InvalidStateError` {{domxref("DOMException")}}
   - : Thrown if the {{domxref("VideoEncoder.state","state")}} is not `"configured"`.
 - `DataError` {{domxref("DOMException")}}
-  - : Thrown if the `chunk` cannot be decoded due to relying on other frames for decoding.
+  - : Thrown if the given `frame` object's rotation and flip do not match the first {{domxref("VideoFrame","VideoFrame")}}'s rotation and flip passed to the {{domxref("VideoEncoder")}} object (the "active orientation").
```
Josh-Cena:
```suggestion
  - : Thrown if the given `frame` object's rotation and flip do not match the rotation and flip of the first {{domxref("VideoFrame")}} passed to `encode()` (the "active orientation").
```

Josh-Cena (approved):
Great catch. Thanks!

PR #40284, opened 2025-07-09
fix(ff-relnote-experimental): Reorder the updates in reverse chronological order

dipikabh:
### Description

This PR reorders the entries in each technology section in the descending order by the **_Nightly_** version they were introduced in. This is already the order in the [HTML DOM API](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#html_dom_api) section.

I've also removed some H4s because the parent H3s were sufficient and the prose contains other details:
- "Primary payment handling" under [Payment Request API](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#payment_request_api)
- "ScreenOrientation.lock()" under [Screen Orientation API](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#screen_orientation_api)
- "WebVR API (Disabled)" under [WebVR API](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#webvr_api)

Updated the heading "[Reporting API support for CSP Violations](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#reporting_api_support_for_csp_violations)" to "Reporting API" for consistency with other headings in the API section.

### Motivation

To allow developers to view the latest changes quickly without having to scroll through older updates.

### DIFF
--- a/files/en-us/mozilla/firefox/experimental_features/index.md
+++ b/files/en-us/mozilla/firefox/experimental_features/index.md
@@ -21,20 +21,6 @@ Check the [Firefox Configuration Editor](https://support.mozilla.org/en-US/kb/ab
 
 ## HTML
 
-### Layout for input type="search"
-
-Layout for `input type="search"` has been updated. This causes a search field to have a clear icon once someone starts typing in it, to match other browser implementations. (See [Firefox bug 558594](https://bugzil.la/558594) for more details.)
-
-| Release channel   | Version added | Enabled by default? |
-| ----------------- | ------------- | ------------------- |
-| Nightly           | 81            | No                  |
-| Developer Edition | 81            | No                  |
-| Beta              | 81            | No                  |
-| Release           | 81            | No                  |
-
-- `layout.forms.input-type-search.enabled`
-  - : Set to `true` to enable.
-
 ### Toggle password display
 
 HTML password input elements ([`<input type="password">`](/en-US/docs/Web/HTML/Reference/Elements/input/password)) include an "eye" icon that can be toggled to display or obscure the password text ([Firefox bug 502258](https://bugzil.la/502258)).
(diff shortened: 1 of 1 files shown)


hamishwillee (commented):
Who the audience for this page and how will they will consume the content?

Sorting the page by most recent change first is good if people come here to find out the most recent changes. I'm not sure that is true - they come to the release note for that now. If you come to this page, isn't it just as likely that you're going to be interested in a specific technology, so you'd prefer an alphabetic view?

Assuming this is the best fit for audience, the cost of this is churn. If some older things gets worked on we may have to move the whole section even if the only change is a number in the columns. Is it worth it?

If there are multiple sections (H4) in a parent, do we move the whole H3 even if only one part has changed?

I am not sure about removing the H4s. I completely agree that they are unnecessary, but they may serve a purpose. For example, say I add a new feature for Payments "Fast Find Payments". Now I have to make up the heading your removed "Primary payment handling">
In Web API docs, we add an H3 heading under H2 Examples even if it isn't needed, because in theory it might be.

Upshot, not opposed (my first reaction was very positive), but want to better understand the audience and thinking since this will create more work for unclear value.

pepelsbey (approved):
Looks good! I like the idea. Thank you :)

dipikabh:
Thanks @hamishwillee for your thoughtful feedback and poking at this ordering from all perspectives, especially the ROI vis-a-vis the maintenance requirement should an older feature receive an update. I agree with your concerns. 

The [WebGPU API](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#graphics_canvas_webgl_and_webgpu) is a perfect example of the issue you've raised about the maintenance burden - as it recently changed from version 113 to 141 - which would've required moving the entire parent section to the top of the API list.

While the idea of reverse chronological ordering is good and would surface the newest incoming changes first, the ongoing maintenance cost would be high. 

I also looked at the usage data for this page and it is pretty low compared to the main release notes pages. While that was an argument to update this page because the impact would be minimal to the few existing users, it is also an argument for not spending any more effort on it and not creating a maintenance burden for ourselves for the future. And so after considering this a bit more, I've decided to close this PR.

Thank you @pepelsbey as well for taking a look at this PR. 🙏

PR #40287, opened 2025-07-10
Fix example in `String#replaceAll` to show advantage of `RegExp.escape` vs string arg

lionel-rowe:
### Description

Fix example in `String#replaceAll` to show potential advantage of `RegExp.escape` over passing a string argument directly.

### Motivation

Current wording "it is better to just use `replaceAll()` and pass the string" isn't necessarily correct, e.g. you may want to only match at word boundaries (I've used non-Unicode-aware `\b` boundaries in the example to avoid overcomplicating).


Josh-Cena (approved):
Sure, thank you!

PR #40300, opened 2025-07-10
Docs: Use && instead of | in text-shadow syntax example

### LINKED ISSUE #39659: Grammar inconsistency in box-shadow syntax example: | vs. &&
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow

Syntax   CSS

I've noticed a recurring issue with the syntax notation in example comments across MDN Web Docs pages. This often leads to confusion when trying to interpret the required components of CSS properties based on formal grammar.

For example, on the box-shadow page:
URL: https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow

Under the "Syntax" section, an example comment provides:

**CSS**

```css
/* A color and two length values */
/* <color> | <length> | <length> */
box-shadow: red 60px -16px;
```

**The Problem:**

In this specific example, the box-shadow property must contain a color and two length values (at minimum, for offset-x and offset-y). However, the comment uses the pipe symbol |.

According to standard BNF (Backus-Naur Form) grammar and even within the specific CSS Value Definition Syntax, the | symbol denotes an "OR" relationship, meaning one choice out of several options. This directly contradicts the requirement for these components to be combined (AND), not chosen exclusively (OR).

**Suggestion for Improvement:**

To accurately represent that these components are all required as part of the combined value, the syntax comment should use && (double ampersand), which in CSS Value Definition Syntax signifies that 'all components must be present, and their order is arbitrary.'

Therefore, I suggest the syntax comment be rewritten as follows:

CSS

```
/* <color> && <length> && <length> */
```

This correction would greatly improve clarity and align the example comments more closely with formal CSS grammar principles, reducing potential confusion for developers.

In all MDN Web Docs where Syntax CSS available

_No response_

kannanwisen:
### Description

This PR corrects the syntax comment in a text-shadow example, changing the incorrect | (OR) separator to && (AND) for better accuracy and clarity.

### Motivation

The previous comment used /* offset-x | offset-y */, which was misleading as it implied only one of the components could be used. Using && correctly signifies that these components are used together to form the text-shadow value. This change aligns the example with formal CSS value syntax and reduces potential confusion for developers reading the documentation.

Fixes #39659

### DIFF
--- a/files/en-us/web/css/text-shadow/index.md
+++ b/files/en-us/web/css/text-shadow/index.md
@@ -58,19 +58,19 @@ p {
 ## Syntax
 
 ```css
-/* offset-x | offset-y | blur-radius | color */
+/* offset-x && offset-y && blur-radius && color */
 text-shadow: 1px 1px 2px black;
 
-/* color | offset-x | offset-y | blur-radius */
+/* color && offset-x && offset-y && blur-radius */
 text-shadow: #fc0 1px 0 10px;
 
-/* offset-x | offset-y | color */
+/* offset-x && offset-y && color */
 text-shadow: 5px 5px #558abb;
 
-/* color | offset-x | offset-y */
+/* color && offset-x && offset-y */
 text-shadow: white 2px 5px;
 
-/* offset-x | offset-y
+/* offset-x && offset-y */
 /* Use defaults for color and blur-radius */
 text-shadow: 5px 10px;
 


chrisdavidmills (changes requested):
@kannanwisen hi there, and thanks for your contribution to MDN!

I am not sure the page author meant `|` to signify logical OR (which is actually `||` in JS terms). I think they were just using it as some kind of generic separator.

Saying that, I can see how it might be possible to make this incorrect assumption. If you've interpreted it in this way, others might do also.

However, I'm not sure if adding logical AND syntax is the answer — this might equally confuse people and make them think they have to use it in the syntax in some way. I think a better solution would be to remove the separator altogether, and have each value component separated by a space, just like they are in the actual syntax.

What do you think?

Josh-Cena:
CSS syntax uses `|` to mean "or". Using it as a separator only causes confusion. `&&` does not mean "and"; it means "must specify both parts".

kannanwisen:
Hi @chrisdavidmills, thanks for the feedback. This is a great discussion.

My initial thought was similar, that a space should be sufficient based on general BNF where a space implies concatenation.

However, after looking at the official CSS Value Definition Syntax, there's a subtle but important difference in the CSS grammar:

- A **space** as a combinator means all components must appear **in that exact order**.
- A **double ampersand (&&)** means all components must be present, but **can appear in any order**.

Since the <color> value in box-shadow and text-shadow can validly appear before or after the <length> values (e.g., red 10px 5px and 10px 5px red are both valid), the && combinator is the most technically accurate way to represent that flexibility.

Therefore, I believe sticking with && in the comment is the best way to avoid ambiguity for developers. Thanks for pushing the discussion on this!

What do you think?

dipikabh:
Please refer to https://github.com/mdn/content/pull/40301 for feedback on similar PRs

files/en-us/web/css/text-shadow/index.md
```diff
@@ -58,19 +58,19 @@ p {
 ## Syntax
 
 ```css
-/* offset-x | offset-y | blur-radius | color */
+/* offset-x && offset-y && blur-radius && color */
```
dipikabh, line 61:
```suggestion
/* Two distances, a blur radius, and a color */
```

files/en-us/web/css/text-shadow/index.md
```diff
@@ -58,19 +58,19 @@ p {
...
 text-shadow: 1px 1px 2px black;
 
-/* color | offset-x | offset-y | blur-radius */
+/* color && offset-x && offset-y && blur-radius */
```
dipikabh, line 64:
```suggestion
/* A color, two distances, and a blur radius */
```

files/en-us/web/css/text-shadow/index.md
```diff
@@ -58,19 +58,19 @@ p {
...
 text-shadow: #fc0 1px 0 10px;
 
-/* offset-x | offset-y | color */
+/* offset-x && offset-y && color */
```
dipikabh, line 67:
```suggestion
/* Two distances and a color */
```

files/en-us/web/css/text-shadow/index.md
```diff
@@ -58,19 +58,19 @@ p {
...
 text-shadow: 5px 5px #558abb;
 
-/* color | offset-x | offset-y */
+/* color && offset-x && offset-y */
```
dipikabh, line 70:
```suggestion
/* A color and two distances */
```

files/en-us/web/css/text-shadow/index.md
```diff
@@ -58,19 +58,19 @@ p {
...
 text-shadow: white 2px 5px;
 
-/* offset-x | offset-y
+/* offset-x && offset-y */
```
dipikabh, line 73:
```suggestion
/* Two distances */
```

dipikabh (changes requested):

Josh-Cena:
Closing until https://github.com/mdn/content/issues/40500 reaches a conclusion. Afterwards we will give detailed guidelines about how the comments should be formatted and every doc in CSS should be updated in one batch instead of doing it piecemeal like this.

PR #40298, opened 2025-07-10
Docs: Use && instead of | in box-shadow syntax example

kannanwisen:
### Description

This PR corrects the syntax comment in a box-shadow example, changing the incorrect | (OR) separator to && (AND) for better accuracy and clarity.

### Motivation

The previous comment used /* <color> | <length> | <length> */, which was misleading as it implied only one of the components could be used. Using && correctly signifies that these components are used together to form the shadow value. This change aligns the example with formal CSS value syntax and reduces potential confusion for developers reading the documentation.

### Related issues and pull requests

Fixes #39659

Related PRs:
- https://github.com/mdn/content/pull/40299
- https://github.com/mdn/content/pull/40300
- https://github.com/mdn/content/pull/40301

PR #40301, opened 2025-07-10
Docs: Use && instead of | in padding syntax example

### LINKED ISSUE #39659: Grammar inconsistency in box-shadow syntax example: | vs. &&

kannanwisen:
### Description

This PR corrects the syntax comment in a padding example, changing the incorrect | (OR) separator to && (AND) for better accuracy and clarity.

### Motivation

The previous comment used /* top and bottom | left and right */, which was misleading as it implied only one of the components could be used. Using && correctly signifies that these components are used together to form the padding value. This change aligns the example with formal CSS value syntax and reduces potential confusion for developers reading the documentation.

Fixes #39659

### DIFF
--- a/files/en-us/web/css/padding/index.md
+++ b/files/en-us/web/css/padding/index.md
@@ -74,13 +74,13 @@ This property is a shorthand for the following CSS properties:
 /* Apply to all four sides */
 padding: 1em;
 
-/* top and bottom | left and right */
+/* top and bottom && left and right */
 padding: 5% 10%;
 
-/* top | left and right | bottom */
+/* top && left and right && bottom */
 padding: 1em 2em 2em;
 
-/* top | right | bottom | left */
+/* top && right && bottom && left */
 padding: 5px 1em 0 2em;
 
 /* Global values */


chrisdavidmills (changes requested):
See my thoughts here: https://github.com/mdn/content/pull/40300#pullrequestreview-3005361188

Also, I can see how my suggestion would not work so well in this instance. I'd also like to get @estelle's thoughts on this issue, as she is the main driving force behind the CSS docs.

estelle:
> See my thoughts here: [#40300 (review)](https://github.com/mdn/content/pull/40300#pullrequestreview-3005361188)
> 
> Also, I can see how my suggestion would not work so well in this instance. I'd also like to get @estelle's thoughts on this issue, as she is the main driving force behind the CSS docs.

The comments are generally more like sentences. So I think the following would be easier for users to parse:

```css
/* top and bottom, left and right */
/* top, left and right, bottom */
/* top, right, bottom, left */
```

chrisdavidmills:
> > See my thoughts here: [#40300 (review)](https://github.com/mdn/content/pull/40300#pullrequestreview-3005361188)
> > Also, I can see how my suggestion would not work so well in this instance. I'd also like to get @estelle's thoughts on this issue, as she is the main driving force behind the CSS docs.
> 
> The comments are generally more like sentences. So I think the following would be easier for users to parse:
> 
> ```css
> /* top and bottom, left and right */
> /* top, left and right, bottom */
> /* top, right, bottom, left */
> ```

+1 here — I like this.

Using human language will avoid the need to use any potentially confusing "logic" syntax.

kannanwisen:
Ah, that makes perfect sense. Great. Thank you, @estelle 

You're right, my suggestion of && is too rigid for the padding, margin, border property since the order of values is significant.

I agree that a descriptive, sentence-like comment such as /* top and bottom, left and right */ is much clearer for developers and is a better fit for MDN's style. 

I can make that change.

dipikabh:
I just left a similar feedback in one of the other PRs: https://github.com/mdn/content/pull/40298#issuecomment-3070009767

kannanwisen:
Hi @dipikabh and @estelle,

Thank you for the feedback and for explaining the different perspectives on this.

My goal was to make the formal syntax comment more accurate, as my main concern was ensuring clarity for new learners and job seekers.

I understand this is a decision about the project's documentation conventions, so I will leave it to you and the other maintainers to make the final call on the best approach. I am happy to make whatever changes you decide on.

Thanks again for your time and guidance.

files/en-us/web/css/padding/index.md
```diff
@@ -74,13 +74,13 @@ This property is a shorthand for the following CSS properties:
 /* Apply to all four sides */
 padding: 1em;
 
-/* top and bottom | left and right */
+/* top and bottom && left and right */
```
estelle, line 77:
```suggestion
/* top and bottom, left and right */
```

files/en-us/web/css/padding/index.md
```diff
@@ -74,13 +74,13 @@ This property is a shorthand for the following CSS properties:
...
 padding: 5% 10%;
 
-/* top | left and right | bottom */
+/* top && left and right && bottom */
```
estelle, line 80:
```suggestion
/* top, left and right, bottom */
```

files/en-us/web/css/padding/index.md
```diff
@@ -74,13 +74,13 @@ This property is a shorthand for the following CSS properties:
...
 padding: 1em 2em 2em;
 
-/* top | right | bottom | left */
+/* top && right && bottom && left */
```
estelle, line 83:
```suggestion
/* top, right, bottom, left */
```

estelle (commented):
made the suggestions as suggested edits.

Josh-Cena:
Closing until https://github.com/mdn/content/issues/40500 reaches a conclusion. Afterwards we will give detailed guidelines about how the comments should be formatted and every doc in CSS should be updated in one batch instead of doing it piecemeal like this.

PR #40305, opened 2025-07-10
Restructure display property values for better clarity

kannanwisen:
### Description

This PR reorganizes the 'Values' section for the CSS display property to better align with the modern multi-keyword syntax. The new structure groups values more logically by their function (e.g., two-keyword, internal, list-item) rather than the current mix of precomposed and multi-keyword values.

### Motivation

The current syntax section on the display page can be confusing. It mixes legacy single-keyword values with the modern syntax in a way that obscures the underlying structure of the property.

This new organization is more systematic and aligns directly with the concepts in the CSS Display Module specification. By grouping values into clear categories, it provides a much clearer mental model for understanding how the different display values relate to each other. This will help both new and experienced developers grasp the property's full capabilities more easily.

### DIFF
--- a/files/en-us/web/css/display/index.md
+++ b/files/en-us/web/css/display/index.md
@@ -77,34 +77,66 @@ code {
 ## Syntax
 
 ```css
-/* precomposed values */
-display: block;
-display: inline;
+/* Two Keyword values */
+display: block flow;
+display: block flow-root;
+display: block table;
+display: block flex;
+display: block grid;
+display: block ruby;
+
+display: inline flow;
+display: inline flow-root;
+display: inline table;
+display: inline flex;
+display: inline grid;
+display: inline ruby;
+
+
+display: run-in flow;
+display: run-in flow-root;
+display: run-in table;
+display: run-in flex;
+display: run-in grid;
+display: run-in ruby;
+
+
+/* List Item values */
+
+display: block flow list-item;
+display: block flow-root list-item;
+
+display: inline flow list-item;
+display: inline flow-root list-item;
+
+
+/* Precomposed Inline-level values */
+
 display: inline-block;
-display: flex;
+display: inline-table;
 display: inline-flex;
-display: grid;
 display: inline-grid;
-display: flow-root;
+
+
+/* Internal values */
+
+display: table-row-group;
+display: table-header-group;
+display: table-footer-group;
+display: table-row;
+display: table-cell;
+display: table-column-group;
+display: table-column;
+display: table-caption;
+display: ruby-base;
+display: ruby-text;
+display: ruby-base-container;
+display: ruby-text-container;
+   
 
 /* Box suppression */
 display: none;
 display: contents;
-
-/* multi-keyword syntax */
-display: block flex;
-display: block flow;
-display: block flow-root;
-display: block grid;
-display: inline flex;
-display: inline flow;
-display: inline flow-root;
-display: inline grid;
-
-/* other values */
-display: table;
-display: table-row; /* all table elements have an equivalent CSS display value */
-display: list-item;
 
 /* Global values */
 display: inherit;


Josh-Cena:
I'm not sure what you are trying to do here. Single-keyword values are not deprecated or legacy—and 99.9% of `display` values you see will be single-keyword. It's also our convention to show single-keyword values before double or more complicated values. Furthermore syntax sections are also not meant to be exhaustive—they are only examples. You are making it inappropriately long.

kannanwisen:
Kindly check the specification and the mdn syntax.

kannanwisen:
The syntax is inappropriate and incorrect. As you closed the pull request, I dont want comment more on this.

kannanwisen:
@Josh-Cena Thank you for the feedback. 

Hi @chrisdavidmills,

I've been thinking more about this, and I wanted to clarify the perspective behind my proposed changes.

My proposal is based on the idea that the modern display property has a single, underlying system: `display: <outer-type> <inner-type>`;. From this viewpoint, every single keyword (block, flex, etc.) is a convenient shorthand for a specific two-keyword combination.

While I understand the current separation is helpful for teaching common legacy values first, it can create confusion by suggesting "precomposed" and "multi-keyword" are two different systems, when they're really just shorthands and their full values.

For example, display: table is also a precomposed value (block table), but it's currently listed under the "other values" section. Similarly, another important precomposed value, inline-table, is missing from the main list entirely. These inconsistencies are what can make the current structure confusing.

However, further down the page under the <display-legacy> heading, inline-table is defined, and the text notes that these single-keyword values are primarily for "backward-compatibility". This can create a confusing narrative for someone trying to understand the modern, systematic nature of display.

My main concern is that this could be misleading for new developers and job seekers trying to build a clear mental model. I believe a more systematic structure that reflects the modern specification would be a great benefit for learners.

Would you be open to discussing a potential restructure from this perspective?

Thanks for your time and consideration.

Josh-Cena:
You can think of `display` as a shorthand for two hypothetical properties: `display-outside` and `display-inside`. You can either specify one of them (and leave the other as default), use a special keyword to specify both ("pre-composed"), or specify both separately. Therefore the approach to document this is not any different from, say, `flex`, `white-space`, or `margin`. All of these properties follow the approach of "shorter syntax first, followed by longer syntax".

kannanwisen:
@Josh-Cena 
Thanks, I understand the convention of documenting shorthands first, as with margin or the flex shorthand property.

However, I believe there's a key difference here. margin is a shorthand for four separate properties, whereas display is a single property. The keyword flex is just an alias for a more descriptive value, block flex, within that same single property.

Because it's a single property with a systematic outer/inner model, I feel that explaining this model clearly would be a great benefit for learners, even if the common single-keyword values are still presented first.

Does that distinction make sense? My main goal is to help reduce confusion for developers trying to understand the modern, underlying structure of display.

Josh-Cena:
The code block you edited just shows quick and representative examples; it's not intended  to convey any "mental model", and certainly should not exhaustively show every valid value. The prose that follows it describes the keywords exactly in the outside/inside/precomposed order as you envision.

kannanwisen:
@Josh-Cena I understand. While I see it from a different perspective, I respect the decision and the established conventions. Thanks again for the detailed feedback!

chrisdavidmills:
@kannanwisen I agree with @Josh-Cena here. A few more thoughts (if mentioned above, apologies!)

- The code block at the top of the Syntax section is supposed to give some examples of the different types of syntax that can be used, not give an exhaustive list. That's the job of the values sections that follow.
- The only display values in the display-legacy data type in the display spec are inline-block | inline-table | inline-flex | inline-grid. I'm pretty sure it doesn't say that single keyword values like block and inline are legacy...?
- In the wild, you will overwhelmingly see single keyword values being used. I think it would confuse learners more to not see those mentioned here.

One thing - you have a lot of ideas and a passion for making positive change here. It might perhaps be good for you to open a discussion item at https://github.com/orgs/mdn/discussions to share some of your ideas, so we can discuss them before you open any more PRs. This way, we can discuss why things have been done the way they have been, and where some of your ideas might fit in, leading to more productive interactions going forward. I want to see you waste as little time as possible.

PR #40307, opened 2025-07-10
List Translator and Language Detector APIs restricted to secure contexts

michaelwasserman:
### Description

List Translator and Language Detector APIs restricted to secure contexts

### Motivation

Match the Spec:
https://webmachinelearning.github.io/translation-api/#translator-api
https://webmachinelearning.github.io/translation-api/#language-detector-api

### Additional details

Perhaps the API docs should also be updated?
https://developer.mozilla.org/en-US/docs/Web/API/Translator_and_Language_Detector_APIs

Reported via Chromium issue:
https://issues.chromium.org/u/1/issues/428561254


michaelwasserman:
FYI @chrisdavidmills and @tomayac

hamishwillee:
@michaelwasserman Thankyou for this. A bit more had to be done, which was adding the secure context headers. I've added to overview page, properties, methods. 

1. It wasn't clear to me if the `destroy()` methods need secure context. If they don't, you could drop the last commit in which they were added (separately)
2. Is there anything missing? (Can you sanity check?)
3. Some of the methods already had secure context. That makes me worry that perhaps the specification changed since this was implemented, and perhaps there are versions in which secure context only applies to some things and not others. Can you confirm that this is not the case? [if it is, we need to add information to the browser compatibility data as well]

tomayac (commented):
LGTM as a non-owner

michaelwasserman:
Thanks for the additional fixes! Everything lgtm, responses:

> 1. It wasn't clear to me if the `destroy()` methods need secure context. If they don't, you could drop the last commit in which they were added (separately)

AFAIK, the entire interface and everything therein is SecureContext in spec and impl.

> 2. Is there anything missing? (Can you sanity check?)

I did a pass over the live pages and changes. This looks complete, thank you!

> 3. Some of the methods already had secure context. That makes me worry that perhaps the specification changed since this was implemented, and perhaps there are versions in which secure context only applies to some things and not others. Can you confirm that this is not the case? [if it is, we need to add information to the browser compatibility data as well]

This seems to be a documentation discrepancy, the interfaces themselves have been SecureContext from initial spec work, and before Chrome origin trials and launches.

hamishwillee (approved):
@michaelwasserman @tomayac Thanks both of you for confirmations and for reporting. This is good to go!

PR #40313, opened 2025-07-11
Clarify and expand example for `HTMLAnchorElement#href`

lionel-rowe:
### Description

Clarify and expand example for `HTMLAnchorElement#href`.

### Motivation

Current example doesn't illustrate how the getter and setter actually work, as it just refers to a non-existent `document.getElementById("myAnchor")`.


lionel-rowe:
Alternative, expanded example including `getAttribute('href')` for comparison:

```js
const anchor = document.createElement("a");
// Initially empty if the element doesn't have an href attribute set
console.log(anchor.href); // ""
console.log(anchor.getAttribute("href")); // null

// Can set to relative URL...
anchor.href = "../../..";
// ...but getter always returns absolute URL
console.log(anchor.href); // "https://developer.mozilla.org/en-US/docs/"
console.log(anchor.getAttribute("href")); // "../../.."

// Setting empty string is interpreted as a relative URL, resolving to the current page
anchor.href = "";
console.log(anchor.href); // "https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/href"
console.log(anchor.getAttribute("href")); // ""

anchor.removeAttribute("href");
// Getter is empty again due to removing attribute
console.log(anchor.href); // ""
console.log(anchor.getAttribute("href")); // null
```

However, that probably just overcomplicates things, as the page isn't about `getAttribute`.

wbamberg (commented):
Sorry to be slow to review this. I think this is a good addition but I think we could present it better: in particular I think examples should illustrate concepts that are explained in the rest of the document, and in this PR the example you've added comes with a lot of explanation that could usefully be put further up the page.

What do you think about moving that explanation after the first paragraph pf the page, something like:

```
- If the `href` attribute is not set, the getter returns an empty string.
- If the `href` attribute is a relative URL, the getter returns an absolute URL relative to the URL of the document.
- If the `href` attribute is set to an empty string, the getter returns the URL of the document.
```

?

Also I think it might be better for the example to use HTML and JS, instead of dynamically setting the href value. Just fewer moving parts that way, making it easier to understand, e.g.

```html
<a id="no_href">no href</a>
<a id="empty_href" href="">empty href</a>
<a id="relative_href" href="relative/url"></a>
```

```js
const noHref = document.querySelector("#no_href");
const emptyHref = document.querySelector("#empty_href");
const relativeHref = document.querySelector("#relative_href");

console.log(noHref.href);
console.log(emptyHref.href);
console.log(relativeHref.href);
```

...you could even make this a live sample, [logging the values](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Live_samples#displaying_a_single_entry_log).

Josh-Cena:
I think it's better to keep the example this way, because it is directly runnable in the console. Also the live sample / playground has some pretty absurd URL that I don't want to expose.

wbamberg (approved):

PR #40319, opened 2025-07-11
Docs: Clarify that flow-root is an inner display type

kannanwisen:
### Description

This PR clarifies the documentation for the display: flow-root value to provide a more complete and accurate picture of its usage, especially in the context of the modern two-keyword display syntax.

### Motivation

The current definition states that flow-root "generates a block box". While this is true for the flow-root keyword itself (which is a shorthand for block flow-root), this can be misleading for developers learning the full capabilities of display.

The flow-root value is technically an <display-inside> value. As such, it can also be paired with an inline outer type (e.g., display: inline flow-root;), which generates an inline-level box, not a block box.

This clarification is important for providing a clearer mental model and preventing confusion. It helps developers understand that flow-root's primary job is to create a new block formatting context, a behavior that can be applied to both block-level and inline-level elements.

### DIFF
--- a/files/en-us/web/css/display/index.md
+++ b/files/en-us/web/css/display/index.md
@@ -147,7 +147,7 @@ The keyword values can be grouped into six value categories.
 ### Inside
 
 - {{CSSxRef("&lt;display-inside&gt;")}}
-  - : These keywords specify the element's inner display type, which defines the type of formatting context that its contents are laid out in (assuming it is a non-replaced element):
+  - : These keywords specify the element's inner display type, which defines the type of formatting context that its contents are laid out in (assuming it is a non-replaced element). When one of these keywords is used by itself as a single value, the element's outer display type defaults to `block` (with the exception of `ruby`, which defaults to `inline`).
     - `flow`
       - : The element lays out its contents using flow layout (block-and-inline layout).
 
@@ -156,7 +156,7 @@ The keyword values can be grouped into six value categories.
         Depending on the value of other properties (such as {{CSSxRef("position")}}, {{CSSxRef("float")}}, or {{CSSxRef("overflow")}}) and whether it is itself participating in a block or inline formatting context, it either establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context) (BFC) for its contents or integrates its contents into its parent formatting context.
 
     - `flow-root`
-      - : Depending on its display property, the element generates either a block box (with block flow-root) or an inline-level box (with inline flow-root). In both cases, it generates a block container that establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context), defining where the formatting root lies.
+      - : The element generates a block box that establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context), defining where the formatting root lies.
 
     - `table`
       - : These elements behave like HTML {{HTMLElement("table")}} elements. It defines a block-level box.
--- a/files/en-us/web/css/display/index.md
+++ b/files/en-us/web/css/display/index.md
@@ -156,7 +156,9 @@ The keyword values can be grouped into six value categories.
         Depending on the value of other properties (such as {{CSSxRef("position")}}, {{CSSxRef("float")}}, or {{CSSxRef("overflow")}}) and whether it is itself participating in a block or inline formatting context, it either establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context) (BFC) for its contents or integrates its contents into its parent formatting context.
 
     - `flow-root`
-      - : The element generates a block box that establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context), defining where the formatting root lies.
+      - : Depending on its display property, the element generates either a block box (with block flow-root) or an inline-level box (with inline flow-root). In both cases, it generates a block container that establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context), defining where the formatting root lies.
+
+
     - `table`
       - : These elements behave like HTML {{HTMLElement("table")}} elements. It defines a block-level box.
     - `flex`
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/display/index.md
```diff
@@ -156,7 +156,8 @@ The keyword values can be grouped into six value categories.
         Depending on the value of other properties (such as {{CSSxRef("position")}}, {{CSSxRef("float")}}, or {{CSSxRef("overflow")}}) and whether it is itself participating in a block or inline formatting context, it either establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context) (BFC) for its contents or integrates its contents into its parent formatting context.
 
     - `flow-root`
-      - : The element generates a block box that establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context), defining where the formatting root lies.
+      - : Depending on its display property, the element generates either a block box (with block flow-root) or an inline-level box (with inline flow-root). In both cases, it generates a block container that establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context), defining where the formatting root lies.
```
chrisdavidmills: Maybe something like this:

```suggestion
      - : Generates a block container that establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context), defining where the formatting root lies. When combined with an outer display type of `block` (that is, `block flow-root`), the element generates a block box; when combined with an outer display type of inline (that is, `inline flow-root`), the element generates an inline box. When the single keyword value `flow-root` is specified, the outer display type defaults to `block`.
```

chrisdavidmills (dismissed):
Hi @kannanwisen!

I think it's a good idea to try to clarify this; I've suggested an update that makes things a bit clearer.

kannanwisen:
Hi @chrisdavidmills,

Thank you for the suggested wording for flow-root. It got me thinking about the documentation's overall consistency, and I noticed something interesting.

The current MDN definition for display: flex or grid explains the outer display type first ("The element behaves like a block-level element...") before it mentions the inner flexbox model.

My original proposed text for flow-root ("Depending on its display property, the element generates either a block box or an inline-level box...") was attempting to follow this same "outer-type first" pattern for the sake of consistency.

Your new suggestion for flow-root ("Generates a block container...") is excellent at explaining the core purpose, but it prioritizes the inner behavior first, which is a different approach than the one used for flex or grid.

This raises a great question about the best documentation strategy: For these layout modes, is it better to consistently explain the outer display type first (like the flex doc does), or the primary purpose/inner type first (as in your suggestion for flow-root)?

chrisdavidmills:
@kannanwisen I think the overall flow of the doc is fine — it makes sense to start by explaining the outer display type, and then the inner display type. Inside each inner display type explanation, I think it makes sense to first define what the inner type does, and then talk about how it can be combined with different outer display types, if the difference in effect is significant.

Thinking about it — if the effect is basically the same in each case ("When combined with an outer display type of `block`, the element generates a block box; when combined with an outer display type of inline, the element generates an inline box.") then it is probably not worth repeating in each case. I worry that my suggestion might look basically the same for each inner type.

We need to remember that this is a reference page, not a guide, so too much surrounding explanation is not appropriate here.

kannanwisen:
Hi @chrisdavidmills,

Thanks again for the detailed discussion and for explaining the project's perspective.

I'd like to state final position clearly for flow-root explanation. My main concern has always been ensuring the highest level of technical accuracy for new developers and Job Seekers, as I believe small imprecisions in foundational topics can lead to confusion.

Because I feel strongly that some of the explanatory text deviates from the specification's precise definitions, I cannot personally agree that the current flow is fine as-is.

That being said, I completely respect that this is a project with its own conventions and editorial standards. I will defer to the judgment of the maintainers for the final decision.

Thank you for considering my feedback.

chrisdavidmills:
> Hi @chrisdavidmills,
> 
> Thanks again for the detailed discussion and for explaining the project's perspective.
> 
> I'd like to state final position clearly for flow-root explanation. My main concern has always been ensuring the highest level of technical accuracy for new developers and Job Seekers, as I believe small imprecisions in foundational topics can lead to confusion.
> 
> Because I feel strongly that some of the explanatory text deviates from the specification's precise definitions, I cannot personally agree that the current flow is fine as-is.
> 
> That being said, I completely respect that this is a project with its own conventions and editorial standards. I will defer to the judgment of the maintainers for the final decision.
> 
> Thank you for considering my feedback.

I'm not saying I don't want to make any changes here, just voicing my concerns about repetition and the amount of content included.

Going back to the original text, and your original suggestion, how about we do something like this:

> The element generates either a block box (when combined with an outer display type of `block`) or an inline box (when combined with an outer display type of `inline`) that establishes a new [block formatting context](/en-US/docs/Web/CSS/CSS_display/Block_formatting_context), defining where the formatting root lies. When the single keyword value `flow-root` is specified, the outer display type defaults to `block`.

?

kannanwisen:
Hi @chrisdavidmills,

Thank you for the continued discussion. To be very clear, let me break down my thoughts.

### Part 1: The flow-root Definition

Regarding the proposed text for flow-root, I believe it could be made more precise for learners by introducing the term "block container" first, as that is the key role the flow-root inner type creates. This is especially important since technically not all block-level boxes are block containers, and vice-versa. Therefore, personally I strict with my orginal statement.
	
### Part 2: Regarding the sentence placement, "When the single keyword value flow-root is specified, the outer display type defaults to block."

To clarify, I did not originally suggest including this statement. If you are interested in adding it, I see two possible outcomes:

**Condition 1 (Your Suggested Approach. Figure 1 (attached)):** If it is included under the flow-root section, then for consistency, a similar statement should be repeated for every corresponding display-inside value, as shown in the "Figure 1" (attached) I created. In my opinion, this goes against the DRY (Don't Repeat Yourself) principle.

**Condition 2 (My Suggested Approach. Figure 2 (attached)):** A better alternative would be to include a single, general explanation under the main <display-inside> heading, as shown in my "Figure 2" (attached).


image

The final decision is, of course, up to the maintainers. My only concern is determining what is best for new learners and job seekers.

chrisdavidmills:
@kannanwisen your new suggestion here (Figure 2) is great. Please go ahead with the necessary update.

Josh-Cena:
@chrisdavidmills Could you take another look?

chrisdavidmills (approved):
Looks good; let's get this merged.

PR #40121, opened 2025-06-28
Avoid inline style (CSS&SVG)

Josh-Cena:
Continuation of https://github.com/mdn/content/pull/39965. The use of presentation attributes results in much terser code as well, in addition to the already listed benefits.

Issue #36276, opened 2024-10-09
dataTransfer is null


rahulbansal16:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/dataTransfer

### What specific section or headline is this issue about?

Code example given below

### What information was incorrect, unhelpful, or incomplete?

In the code example, it is mentioned that if I try to paste some text inside the content editable div, the result div just below the content editable div will access the data transfer object, get that content, and display it.

 However, running the example doesn't show any output in the result. Instead, there is an error saying "unable to read getData of null," basically indicating that the data transfer object is null.

### What did you expect to see?

By the example, whatever I have pasted inside the content editable div should be displayed in the result div.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/api/inputevent/datatransfer`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/dataTransfer
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/inputevent/datatransfer/index.md
* Last commit: https://github.com/mdn/content/commit/980b5a01c4527ef69fee3b865c68ee3ffb09d612
* Document last modified: 2024-07-05T01:52:57.000Z

</details>

wbamberg:
This works for me in Firefox and Safari, but not in Chrome. I don't know why though.

himanshugarg:
Screenshot from Chrome for reference:
![image](https://github.com/user-attachments/assets/edca8e98-de0f-4c7b-a17d-02b9dd3e0bd7)

Josh-Cena:
This is likely a Chrome bug, if the other browsers behave differently. I've submitted a Chrome bug: https://issues.chromium.org/issues/401593412 and I'll close this unless I'm informed that it's intentional.

PR #40339, opened 2025-07-13
Remove remaining usage of inline style

Josh-Cena:
Continuation of https://github.com/mdn/content/pull/40121. After this PR, we have almost no use of `style` and `on*` HTML attributes.

In a lot of cases, we are adding ad-hoc styles to each element in a list. In this case I'm using JavaScript instead, which better parallels how this kind of code likely works IRL (it will come from some external data).

### DIFF
--- a/files/en-us/web/css/filter-function/hue-rotate/index.md
+++ b/files/en-us/web/css/filter-function/hue-rotate/index.md
@@ -158,8 +158,8 @@ These values produce the same results:
 
 ```css
 filter: hue-rotate(90deg); /* 90deg rotation */
-filter: url(#hue-rotate); /* with embedded SVG */
-filter: url(folder/fileName.svg#hue-rotate); /* external svg filter definition */
+filter: url("#hue-rotate)"; /* with embedded SVG */
+filter: url("folder/fileName.svg#hue-rotate"); /* external svg filter definition */
 ```
 
 This example shows three images: the image with a `hue-rotate()` filter function applied, the image with an equivalent `url()` filter applied, and the original images for comparison:
@@ -177,13 +177,13 @@ This example shows three images: the image with a `hue-rotate()` filter function
     <tr>
       <td>
         <img
-          style="filter: hue-rotate(90deg)"
+          class="css-filter"
           src="https://mdn.github.io/shared-assets/images/examples/progress-pride-flag.jpg"
           alt="Pride flag with rotated colors" />
       </td>
       <td>
         <img
-          style="filter: url(#hue-rotate)"
+          class="svg-filter"
           src="https://mdn.github.io/shared-assets/images/examples/progress-pride-flag.jpg"
           alt="Pride flag with rotated colors" />
       </td>
@@ -195,6 +195,15 @@ This example shows three images: the image with a `hue-rotate()` filter function
     </tr>
   </tbody>
 </table>
+```
+
+```css hidden live-sample___svg_filter
+.css-filter {
+  filter: hue-rotate(90deg);
+}
+.svg-filter {
+  filter: url("#hue-rotate");
+}
 ```
 
 {{EmbedLiveSample('svg_filter','100%','280')}}
--- a/files/en-us/web/css/filter-function/hue-rotate/index.md
+++ b/files/en-us/web/css/filter-function/hue-rotate/index.md
@@ -204,6 +204,9 @@ This example shows three images: the image with a `hue-rotate()` filter function
 .svg-filter {
   filter: url("#hue-rotate");
 }
+svg:not(:root) {
+  display: none;
+}
 ```
 
 {{EmbedLiveSample('svg_filter','100%','280')}}
--- a/files/en-us/web/css/flex-grow/index.md
+++ b/files/en-us/web/css/flex-grow/index.md
@@ -113,12 +113,12 @@ In this example, the sum of six flex-grow factors is equal to eight, meaning eac
   <code>flex-grow: 2</code> set.
 </p>
 <div id="content">
-  <div class="small" style="background-color:red;">A</div>
-  <div class="small" style="background-color:lightblue;">B</div>
-  <div class="small" style="background-color:yellow;">C</div>
-  <div class="double" style="background-color:brown;">D</div>
-  <div class="double" style="background-color:lightgreen;">E</div>
-  <div class="small" style="background-color:brown;">F</div>
+  <div class="box1">A</div>
+  <div class="box2">B</div>
+  <div class="box3">C</div>
+  <div class="box4">D</div>
+  <div class="box5">E</div>
+  <div class="box6">F</div>
 </div>
 ```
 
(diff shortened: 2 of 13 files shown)


files/en-us/web/css/filter-function/hue-rotate/index.md
```diff
@@ -197,6 +197,15 @@ This example shows three images: the image with a `hue-rotate()` filter function
 </table>
 ```
 
+```css hidden live-sample___svg_filter
+.css-filter {
+  filter: hue-rotate(90deg);
+}
+.svg-filter {
+  filter: url("#hue-rotate");
+}
```
bsmth, line 206: You can add this to stop the first SVG from pushing the table down the page. This is [also on prod](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/hue-rotate#with_url_and_the_svg_hue-rotate_filter):

```suggestion
}
svg:not(:root) {
  display: none;
}
```

files/en-us/web/css/flex-grow/index.md
```diff
@@ -133,14 +133,37 @@ div > div {
   border: 3px solid rgb(0 0 0 / 20%);
 }
```
bsmth, line 135: Any objection to making this a little fancier? I find something like this easier to read:

```suggestion
body {
  font-family: sans-serif;
}
code {
  background-color: lightgrey;
  border-radius: 5%;
}
```

files/en-us/web/css/if/index.md
```diff
@@ -401,7 +401,7 @@ Next, we target the `<h2>` element's {{cssxref("::before")}} pseudo-element, set
 ```css-nolint live-sample___basic
 h2::before {
   content: if(
-    style(--show-apple:true): "🍎 ";
+    style(--show-apple: true): "🍎 ";
```
bsmth, line 404: L356 has `--show-apple:true` FYI

bsmth (approved):
Only minor suggestions (mostly aesthetic), otherwise looking good, ty!

files/en-us/web/css/flex-grow/index.md
Josh-Cena: Um, I prefer the vanilla version. It would require much more code to make it look actually nice (for example you would need some padding for the code)

files/en-us/web/css/if/index.md
Josh-Cena: Sure, we don't have a formatter for style attributes in general though. For example whether we should do `display:block` (terseness) or `display: block;` (consistent with actual CSS declarations). So I'm only formatting real CSS here.

files/en-us/web/css/flex-grow/index.md
bsmth: OK as you like 👍🏻

files/en-us/web/css/if/index.md
bsmth: Alright, not super important anyway

Josh-Cena:
Thanks for the suggestions!

files/en-us/web/css/filter-function/hue-rotate/index.md
myfonj: Applying `display: none` to any element in an SVG `filter`'s DOM path breaks the effect in Firefox. Consequently, https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/hue-rotate#frame_svg_filter now displays:

Table with three times pride flag photo, first labelled 'hue-rotate' and tinted into magenta, second labelled 'url()' and third labelled 'Original image'. Second and third look identical, despite second's alt text reads 'Pride flag with rotated colors'.

in all Gecko-based browsers. In Goanna and older Firefox ESRs (102, for example), this construct would make the image disappear entirely, had those samples but worked in such older versions at all.

I've flagged this incident on the long-standing https://bugzilla.mozilla.org/show_bug.cgi?id=376027 as an argument for increasing its priority. (Given that there are permalinks involved, I suppose this could be corrected at the maintainers' discretion, assuming Firefox compatibility still holds any priority here.)

Josh-Cena: @myfonj Thanks for the info! I think browser bugs do not motivate content updates, but if you have a nice way to fix this without reintroducing inline style (assuming this problem wasn't there before) I'm all ears.

myfonj: > way to fix this

Well the "idiomatic" way is usually to do the same trick as with making any other element "visually hidden" so removed from rendering flow and display but preserved for other reasons; I think that currently the "hottest" way is making it absolutely positioned and clip, or something like that. But pragmatically, adding `width="0" height="0"` to the SVG would do the same trick without unnecessarily polluting the CSS and I think super clear communication of its purpose. Technically "not an inline style", even. (Semi-joking.) I guess any article showing this technique how to use SVG effects in CSS would use some "safe" hiding technique, or even mention not to do the tempting `display: none` and explain why. Random pick: https://frontendmasters.com/blog/grainy-gradients/ . Ana uses that

```HTML
<svg width='0' height='0' aria-hidden='true'>
```
and to remove it from flow also:
```CSS
svg[height='0'][aria-hidden='true'] { position: fixed }
```

(The `aria-hidden` might be a good remark here, but for different discussion.)

> browser bugs do not motivate content updates

Interesting. Out of curiosity, would you say the same if the rendering engine in question was, say, Blink, affecting more than half of web users worldwide?

This problem in the browser engine (Gecko) was reported 18 years ago, by the way.  While super annoying and periodically reappearing, real-world usage seems to be adjusted to that in some way or another.

myfonj: > without reintroducing inline style

(OT for this specific topic) I don't follow updates here closely, but if I may give my 2 cents: if your current crusade is really to banish **all** inline styles, I don't see it as universally beneficial. In case of changing **SINGLE** CSS property set through an inline style attribute only to create class (or use other selector) with that single CSS property in a separate style block … This, in my perspective, only introduces unnecessary complexity, puts more demand on the reader, who has to extend the context and follow more "hops". So I hope you are not blindly changing minimal stuff like 

```HTML
<p style="magic: abraca">Demo of spell.
<p style="magic: dabra">Demo of different spell.
```
into
```HTML
<p>Demo of spell.
<p>Demo of different spell.
```
```CSS
p:first-of-type {
 magic: abraca;
}
p:nth-of-type(2) {
 magic: dabra;
}
```
because such change would be clearly detrimental for teaching purposes: some linter paradigms are totally orthogonal to making a demonstration of some topic as clear as possible. So I hope you are applying these changes with reason and care.

Josh-Cena: Thanks. Your suggestion about zero-size SVG makes sense to me and I would gladly take it.

> Out of curiosity, would you say the same if the rendering engine in question was, say, Blink, affecting more than half of web users worldwide?

I would be bold and say "yes", but of course we would need to assess it case-by-case. For example, https://github.com/mdn/content/issues/36276 is a Chrome bug which resulted in me filing a Chromium bug instead of working around it in content.

> because such change would be clearly detrimental for teaching purposes: some linter paradigms are totally orthogonal to making a demonstration of some topic as clear as possible.

People copy our examples as-is into their projects, so we need to make sure we show industry best practices, not just what's convenient for teaching. As I mentioned in the PR description:

- Inline style is not lintable: people use CSS linters in their projects and they should not be able to work around them using inline styles.
- Inline style may be rejected by CSP settings: we are promoting high security by default.
- Inline style is not composable: they quickly lead to large amounts of duplication. `style="display: none"` is longer than `class="hidden"` and results in larger markup sizes. Realistically, most people use some form of CSS methodology (say, Tailwind, styled-components) and it's very convenient for them to add these ad-hoc utility classes anyway.

So, we decided that it's not only beneficial for us, but beneficial for our readers in the long run as well. We are not dogmatic though: for example, for SVG elements (not the root, but its children), we are fine with inline styles because most SVG editors produce them by default so they are sort of idiomatic.

PR #40342, opened 2025-07-13
Remove incorrect usage of "declarative" in Popover_API index.md

mfp22:
### Description

Describing the HTML attribute usage as "declarative" is redundant at best, and incorrect if using a definition of "declarative" that isn't just "whatever makes it into HTML."

In the case of popover attributes, the button is commanding the popover to appear. That's imperative by the English definition of "imperative": "Expressive of a command." https://www.merriam-webster.com/dictionary/imperative

The dialog `open` attribute is an example of what a declarative version of this would look like, with the dialog declaring its own behavior instead. 

### Motivation

Clarity

### DIFF
--- a/files/en-us/web/api/popover_api/index.md
+++ b/files/en-us/web/api/popover_api/index.md
@@ -10,7 +10,7 @@ browser-compat:
 
 {{DefaultAPISidebar("Popover API")}}
 
-The **Popover API** provides developers with a standard, consistent, flexible mechanism for displaying popover content on top of other page content. Popover content can be controlled either declaratively using HTML attributes, or via JavaScript.
+The **Popover API** provides developers with a standard, consistent, flexible mechanism for displaying popover content on top of other page content. Popover content can be controlled either using HTML attributes, or via JavaScript.
 
 ## Concepts and usage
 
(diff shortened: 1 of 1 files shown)


hamishwillee (approved):
Thanks, and welcome to MDN.

The text is not _wrong_. Using HTML attributes to define behaviour is declarative programming - it describes the intent without spelling out the control flow. The use of declarative and imperative to differentiate between HTML-alike and using programming languages normal in software programming.

That said, completely agree it is redundant. I much prefer your wording. 

I'm going to merge this, but note that others might not for similar changes in future. Because there are so few reviewers, we try to focus on making sure that we concentrate on things that are wrong or ambiguous - that's why you'll see grammar changes being rejected unless they make a concrete difference.

mfp22:
Thank you for the feedback. I wasn't entirely sure what the response would be.

Originally, imperative and declarative were used to describe paradigms for isolated, static programs. Once programs became more interactive and time entered the equation, people continued to think of imperative and declarative as only describing how static things were described instead of adjusting based on the meaning of the word itself. FRP in 1997 introduced the idea that behavior over time could be declarative, but the industry is still in the process of absorbing that mindset. But it's happening naturally everywhere reactivity is introduced, including JSX and signals. You'll see people describing those as declarative, in opposition to imperative patterns. 

If you consider the behavior of the popover over time, _that_ is what the command is contributing to. The button is part of the control flow of the popover's behavior _over time_.

I know people will disagree, but I prefer this definition for many reasons, but most of all because it follows the literal English definition of "imperative" instead of industry jargon with a lot of baggage and contradictions.

Thank you again for being nice while disagreeing. Not everybody is :)

PR #40344, opened 2025-07-14
Remove a codepen link about testing support

Josh-Cena:
Part of https://github.com/mdn/content/issues/16120. This codepen only demonstrates support, which is already shown in BCD. It's also written in CoffeeScript.


bsmth (approved):
Makes sense, tnx 👍🏻

PR #40348, opened 2025-07-14
Docs: fix ReadableByteStreamController: enqueue()

OpportunityLiu:
### Description

Change the summary of `enqueue` method.

### Motivation

Clarify how the chunk is handled by `enqueue` method.

### Additional details

See: https://streams.spec.whatwg.org/#readable-byte-stream-controller-enqueue, step 7.


wbamberg (approved):
👍 thank you, and sorry to be so slow with this one.

PR #40358, opened 2025-07-14
chore: move sidebar into fm

bsmth:
### Description

Moves the sidebar into front matter

### Motivation

We're using front matter instead of the macro going forward

### Additional details

I think the `captured-surface-control` page landed while the other HTTP section cleanup PR was in-flight: https://github.com/mdn/content/pull/39982


Josh-Cena:
Just fixed the same issue in https://github.com/mdn/content/pull/40359 ^^

Josh-Cena (approved):
Leaving an approval so you can merge whichever

bsmth:
Thanks a lot, let's do the other one 👍🏻

PR #40361, opened 2025-07-14
Add example to force import re-evaluation

jmcmenamy:
### Description

Added another import example to show how to bypass the module cache and force re-evaluation without using a unique 'cache-busting' query parameter

### Motivation

The existing example:

```import(`/my-module.js?t=${Date.now()}`);```

felt a bit hacky to me, and it took a while to find a (subjectively) 'cleaner' alternative online. So I thought it would be helpful to add an alternative example.

### Additional details

None

None

### DIFF
--- a/files/en-us/web/javascript/reference/operators/import/index.md
+++ b/files/en-us/web/javascript/reference/operators/import/index.md
@@ -200,6 +200,23 @@ Promise.all(
 ).then((modules) => modules.forEach((module) => module.load()));
 ```
 
+### Importing a module with forced re-evaluation
+
+Another way to re-import and re-evaluate a module without restarting the entire JavaScript environment is to use a [`Blob`](/en-US/docs/Web/API/Blob) with [`URL.createObjectUrl`](/en-US/docs/Web/API/URL/createObjectURL_static):
+
+```js
+fetch("/my-module.js")
+  .then((r) => r.text())
+  .then((code) => {
+    const url = URL.createObjectURL(
+      new Blob([code], { type: "text/javascript" }),
+    );
+    return import(url).finally(() => {
+      URL.revokeObjectURL(url);
+    });
+  });
+```
+
 ## Specifications
 
 {{Specifications}}

### CODE REVIEW

**Comment by Josh-Cena** [comment]
Thanks. In what way is this an improvement from the existing example? Although you managed to free the underlying code blob, the compiled module is still stuck in the module cache. It also relies on web APIs which may not be available in every environment, while being much longer.

**Comment by jmcmenamy** [comment]
> In what way is this an improvement from the existing example?

Hi @Josh-Cena , good question. Here's why I think this example offers some improvements over the existing one:

- The query param trick relies on browser module caching behavior that is common but not guaranteed by a specification (as far as I could tell). So different environments may ignore or strip the query parameters, especially if custom URL resolution is used (e.g., service workers, import maps, extensions, etc.), causing the existing example to not work. By contrast, using a Blob and object URL forces re-evaluation regardless of URL resolution policies. While it’s more verbose, it is a more reliable method for ensuring re-imports in all environments that support the standard Web APIs.
- `Date.now()` will return the same value if called within the same millisecond. So there’s a rare but real chance that ```import(`/my-module.js?t=${Date.now()}`);``` might not force a re-import. In those cases, either more code would be needed to ensure uniqueness, or an approach like the `Blob` example avoids the issue entirely. 

> the compiled module is still stuck in the module cache

Yes, without a way to interact with the module cache, both examples unfortunately suffer from this issue

> It also relies on web APIs which may not be available in every environment

Good point, this example does rely on more web APIs. However, the [`import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) function that both examples use became widely available in 2020, while [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob), [`createObjectUrl()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static), [`revokeObjectUrl()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static), and [`then()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then) became widely available in 2015 (same time as [`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now)), [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) in 2017, and [`finally()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally) in 2018. So in practice, if the `import()` function is supported, it seems likely that the other APIs will be supported as well.

So while the existing example is nice and concise, I think it would be good to include another example that could be more reliable

**Comment by Josh-Cena** [comment]
> The query param trick relies on browser module caching behavior that is common but not guaranteed by a specification (as far as I could tell). 

It is. The string passed to `import()` is an opaque specifier that browsers are not allowed to parse and re-interpret. The spec requires engines to maintain a mapping of `{ <specifier> => <compiled module> }`, and two different import paths are always distinct specifiers.

I'm also not talking about these web APIs' availability on the web; I'm talking about non-browser environments, such as Node.js, or some embedded system such as React Native, Moddable XS, etc.

**Comment by jmcmenamy** [comment]
> It is. The string passed to import() is an opaque specifier that browsers are not allowed to parse and re-interpret.

Ah, thanks for the clarification. So currently I can't come up with an example (e.g. using import maps or service workers) that reliably breaks the existing example. However, if import maps are extended in the future to allow for wildcards or other matching, then the import map could potentially be used to strip query parameters away. This [issue](https://github.com/WICG/import-maps/issues/7) proposes that, but hasn't yet been discussed in any issues [here](https://github.com/whatwg/html/issues) as far as I can tell since import maps became part of the HTML standard. So the new example

- would be future proofed against import maps being used to strip query parameters away if wildcard matching or similar becomes supported
- still avoids the the rare but real possibility of `Date.now()` returning the same value, which would break the existing example
- introduces readers to other useful Web APIs they might not have encountered yet and shows an example using them, which helps them learn the docs

> I'm also not talking about these web APIs' availability on the web; I'm talking about non-browser environments, such as Node.js, or some embedded system such as React Native, Moddable XS, etc.

The existing example relies on the environment making `Date.now()` available, which the new example doesn't. The APIs each example relies on are incomparable, so I think it would be helpful to include both examples.

**Comment by Josh-Cena** [comment]
`Date.now()` is a core JS API that every standard-compliant JS engine has to implement; the web APIs are only required for browsers and very likely to not be available outside of the web.

Realistically, the only use case I found for force re-importing is to implement hot module reloading, which can only happen in a server-side environment, aka Node. What's an actual use case of force re-import in the browser?

**Comment by jmcmenamy** [comment]
> `Date.now()` is a core JS API that every standard-compliant JS engine has to implement; the web APIs are only required for browsers and very likely to not be available outside of the web.

Fair point. Fwiw, Deno and Node are included in the compatibility information and began supporting the baseline of all these APIs no later than 2022, so the example should work in a typical server side environment. But I see your point that other non browser environments may not implement these APIs.

> What's an actual use case of force re-import in the browser?

Here's a [convenient example](https://web.mit.edu/6.102/www/sp25/inclass/ic19-music-language/). The html on the browser has a button that plays some music. It plays music by running some bundled javascript code that uses the [MIDI API](https://developer.mozilla.org/en-US/docs/Web/API/Web_MIDI_API). The MIDI API currently isn't widely available to Workers ([open Chromium issue](https://issues.chromium.org/issues/379805727)), only the main thread, so a worker can't be used to run the bundled js and play the music. The browser dynamically imports the bundled js in order to run the code and play the music as a side effect. We need to re-import and re-evaluate it so the music is played each time the button is clicked. We could export a function in the bundled js that the main js code can import and call to play the music, but the bundled js will be changing as we change the music notes that are played, so we'll still need to force re-import each time we want to play the music. We don't want to have to reload the webpage every time we change the notes to play.

Stated generally, one use case of force re-import in the browser is for repeatedly running external code that uses APIs not accessible to Web Workers (or doesn't need to be run in a separate thread) and that will be updated while the browser is open (basically lightweight hot module reloading in the browser), without adding a dependency (e.g. Webpack).

**Comment by Josh-Cena** [comment]
Thanks. I really tried to give your PR enough charitable consideration, but I still cannot convince myself of its utility. This function fixes the edge case of importing two things in the same millisecond, at the cost of:

- Does not admit any import options such as import attributes, lazy import, etc.
- May have subtly different `fetch` configurations from `import`, such as `destination`, `Accept` headers, etc.
- Does not use the module resolution algorithm (which is fixable with `import.meta.resolve`, so this one is minor)
- Is subject to CORS and CSP restrictions which may not allow executing JS from blob: URLs (disclaimer: I didn't test this)
- Uses browser-specific APIs that don't work in Node, Deno, or Bun

Given all these gotchas, while it may work for the ordinary case, I really don't think it's fool-proof enough to be recommended as a general pattern for MDN, and honestly I think the edge case of `Date` colliding is completely ignorable (because remember you also need the module to behave differently within that same millisecond in order to import two differentiable modules).

Sorry if that's disappointing for you. I do buy your use case and I think there are certain modules on the web where fresh import may be useful, but I think you should just go with the `Date.now()` trick. And as an aside, I think a module that _requires_ you to do a fresh import is a bad design—the module should be stateless and expose functions that are polymorphic. Not in your control though.

Here was the suggestion I was going to give to move this PR forward, before I gave up on the idea:

````md
If you need to re-import and re-evaluate a module without restarting the entire JavaScript environment, in addition to the ``import(`my-module.js?t=${Date.now()}`)`` trick, you can consider the following function. Every time, it generates a new importable URL with [`URL.createObjectURL`](/en-US/docs/Web/API/URL/createObjectURL_static) to a [`Blob`](/en-US/docs/Web/API/Blob) of JavaScript code:

```js
async function importFresh(modulePath) {
  const resolvedPath = import.meta.resolve(modulePath);
  const code = await fetch(resolvedPath).then((r) => r.text());
  const url = URL.createObjectURL(
    new Blob([code], { type: "text/javascript" }),
  );
  try {
    return await import(url);
  } finally {
    URL.revokeObjectURL(url);
  }
}
```

Note that this relies on some browser APIs that may not be available in your JavaScript environment, in which case the `Date.now()` trick is still necessary.
````

jmcmenamy:
I understand, your points make sense for why it shouldn't be a general pattern for MDN. This discussion was quite informative, thank you very much for your time and consideration!

PR #40362, opened 2025-07-14
Uint8Array.prototype.toHex: correct output

phoddie:
### Description

The output displayed for the second part of the example is incorrect. This is easily verified in a web browser.

### Motivation

Developers may be confused by incorrect output in the example.


Josh-Cena (approved):
Good catch. Thanks!

Issue #40285, opened 2025-07-09
Need to explicitly specify default step base with min/value absent


jack5github:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/date

### What specific section or headline is this issue about?

#step

### What information was incorrect, unhelpful, or incomplete?

Article stops short of specifying what the 'step' property is relative to when both 'min' and 'value' are not set, only saying "an appropriate default value if neither of those is provided".

### What did you expect to see?

A concrete definition of what date is used as the basis for the input in this context. The same should be the case for all input that behave in this way.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/html/reference/elements/input/date`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/date
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/reference/elements/input/date/index.md
* Last commit: https://github.com/mdn/content/commit/c1564acf160ef4b320fb7b89ab65211b9c50cf1b
* Document last modified: 2025-06-07T15:17:44.000Z

</details>

Josh-Cena:
Thanks for the report. All types have the default step base of `0`, except `type="week"` which is `1970-W01` (`−259,200,000`)---which is already documented. We need to change "an appropriate default value" to the actual value for each input type.

PR #40365, opened 2025-07-15
docs(html): Clarify default step base is 0 (1970-01-01) for date inputs

Shrinivassab:
### Description
Updates the `input's` documentation to explicitly state that the default `step`. (when neither `min` nor `value` is provided )

#40285

### DIFF
--- a/files/en-us/web/html/reference/elements/input/date/index.md
+++ b/files/en-us/web/html/reference/elements/input/date/index.md
@@ -82,7 +82,7 @@ If both the `max` and `min` attributes are set, this value must be a date string
 
 ### step
 
-The `step` attribute is a number that specifies the granularity that the value must adhere to, or the special value `any`, which is described below. Only values which are equal to the basis for stepping ([`min`](#min) if specified, [`value`](/en-US/docs/Web/HTML/Reference/Elements/input#value) otherwise, or **`0`** (the Unix epoch, `1970-01-01`) if neither is provided) are valid.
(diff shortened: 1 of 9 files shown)


files/en-us/web/html/reference/elements/input/date/index.md
```diff
@@ -82,7 +82,7 @@ If both the `max` and `min` attributes are set, this value must be a date string
 
 ### step
 
-The `step` attribute is a number that specifies the granularity that the value must adhere to, or the special value `any`, which is described below. Only values which are equal to the basis for stepping ([`min`](#min) if specified, [`value`](/en-US/docs/Web/HTML/Reference/Elements/input#value) otherwise, and an appropriate default value if neither of those is provided) are valid.
+The `step` attribute is a number that specifies the granularity that the value must adhere to, or the special value `any`, which is described below. Only values which are equal to the basis for stepping ([`min`](#min) if specified, [`value`](/en-US/docs/Web/HTML/Reference/Elements/input#value) otherwise, or **`0`** (the Unix epoch, `1970-01-01`) if neither is provided) are valid.
```
Josh-Cena:
```suggestion
The `step` attribute is a number that specifies the granularity that the value must adhere to, or the special value `any`, which is described below. Only values which are a whole number of steps from the step base are valid. The step base is [`min`](#min) if specified, [`value`](/en-US/docs/Web/HTML/Reference/Elements/input#value) otherwise, or `0` (the Unix epoch, `1970-01-01`) if neither is provided.
```

Josh-Cena (changes requested):
Could you also check all other input types and propagate this same wording?

Shrinivassab:
Sure. will check and update with similar wording.

estelle (approved):
Thanks! 🎉

Josh-Cena (approved):
Nice, thank you! I just reordered the paragraphs a bit to make it less repetitive.

PR #40369, opened 2025-07-15
Glossary idempotent

mb21:
### Description
Clarify PATCH idempotency in body text

### Motivation

Previously the paragraph mentioned PUT, DELETE and POST but curiously left out PATCH, which on a cursory reading could be interpreted as PATCH being idempotent (yes, the cursory reader was me).

### DIFF
--- a/files/en-us/glossary/idempotent/index.md
+++ b/files/en-us/glossary/idempotent/index.md
@@ -7,13 +7,17 @@ sidebar: glossarysidebar
 
 An HTTP method is **idempotent** if the intended effect on the server of making a single request is the same as the effect of making several identical requests.
 
-This does not necessarily mean that the request does not have _any_ unique side effects: for example, the server may log every request with the time it was received. Idempotency only applies to effects intended by the client: for example, a POST request intends to send data to the server, or a DELETE request intends to delete a resource on the server.
+The HTTP specification defines several HTTP methods and their semantics, which includes whether they are idempotent or not. All {{glossary("Safe/HTTP", "safe")}} methods are idempotent, as well as {{HTTPMethod("PUT")}} and {{HTTPMethod("DELETE")}}. The {{HTTPMethod("POST")}} and {{HTTPMethod("PATCH")}} methods are not guaranteed to be idempotent.
 
-All {{glossary("Safe/HTTP", "safe")}} methods are idempotent, as well as {{HTTPMethod("PUT")}} and {{HTTPMethod("DELETE")}}. The {{HTTPMethod("POST")}} and {{HTTPMethod("PATCH")}} methods are not idempotent.
+Idempotent methods are distinguished because the client can retry a request if its not sure whether the request reached the server. If both requests happen to reach the server, as long as the method is idempotent, no harm is done.
 
-To be idempotent, only the state of the server is considered. The response returned by each request may differ: for example, the first call of a {{HTTPMethod("DELETE")}} will likely return a {{HTTPStatus("200")}}, while successive ones will likely return a {{HTTPStatus("404")}}. Another implication of {{HTTPMethod("DELETE")}} being idempotent is that developers should not implement RESTful APIs with a _delete last entry_ functionality using the `DELETE` method.
+Idempotency, as defined by the HTTP specification, only considers the _intented_ effect of the client on the server: for example, a POST request intends to send data to the server, or a DELETE request intends to delete a resource on the server. A request with an idempotent method does not necessarily mean that the request does not have _any_ unique side effects: for example, the server may log every request with the time it was received.
 
-Note that the idempotence of a method is not guaranteed by the server and some applications may incorrectly break the idempotence constraint.
+Also, the response returned by each request may differ: for example, the first call of a {{HTTPMethod("DELETE")}} will likely return a {{HTTPStatus("200")}}, while successive ones will likely return a {{HTTPStatus("404")}}. Another implication of {{HTTPMethod("DELETE")}} being idempotent is that developers should not implement RESTful APIs with a _delete last entry_ functionality using the `DELETE` method.
+
+While servers are very much encouraged to adhere to the semantics laid out by the HTTP specification, the spec does not mandate it. There is nothing preventing a server in the wild from exposing a non-idempotent endpoint under an idempotent HTTP method – although clients will be surprised.
+
+A few examples:
 
 `GET /pageX HTTP/1.1` is idempotent, because it is a safe (read-only) method. Successive calls may return different data to the client, if the data on the server was updated in the meantime.
 
--- a/files/en-us/glossary/idempotent/index.md
+++ b/files/en-us/glossary/idempotent/index.md
@@ -17,7 +17,7 @@ Also, the response returned by each request may differ: for example, the first c
 
 While servers are very much encouraged to adhere to the semantics laid out by the HTTP specification, the spec does not mandate it. Nothing is preventing a server in the wild from exposing a non-idempotent endpoint under an idempotent HTTP method, although clients may well be surprised.
 
-A few examples:
+## Examples
 
 `GET /pageX HTTP/1.1` is idempotent, because it is a safe (read-only) method. Successive calls may return different data to the client, if the data on the server was updated in the meantime.
 
--- a/files/en-us/glossary/idempotent/index.md
+++ b/files/en-us/glossary/idempotent/index.md
@@ -9,7 +9,7 @@ An HTTP method is **idempotent** if the intended effect on the server of making 
 
 This does not necessarily mean that the request does not have _any_ unique side effects: for example, the server may log every request with the time it was received. Idempotency only applies to effects intended by the client: for example, a POST request intends to send data to the server, or a DELETE request intends to delete a resource on the server.
 
-All {{glossary("Safe/HTTP", "safe")}} methods are idempotent, as well as {{HTTPMethod("PUT")}} and {{HTTPMethod("DELETE")}}. The {{HTTPMethod("POST")}} method is not idempotent.
+All {{glossary("Safe/HTTP", "safe")}} methods are idempotent, as well as {{HTTPMethod("PUT")}} and {{HTTPMethod("DELETE")}}. The {{HTTPMethod("POST")}} and {{HTTPMethod("PATCH")}} methods are not idempotent.
 
 To be idempotent, only the state of the server is considered. The response returned by each request may differ: for example, the first call of a {{HTTPMethod("DELETE")}} will likely return a {{HTTPStatus("200")}}, while successive ones will likely return a {{HTTPStatus("404")}}. Another implication of {{HTTPMethod("DELETE")}} being idempotent is that developers should not implement RESTful APIs with a _delete last entry_ functionality using the `DELETE` method.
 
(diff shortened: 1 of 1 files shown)


chrisdavidmills (changes requested):
Hi @mb21, and thanks for the contribution!

I'm not an expert on this, but from what I've read on the subject, your addition seems like an oversimplification. `PATCH` can be idempotent or non-idempotent, depending on how your request is written to update the data you are modifying.

This thread contains some good examples: https://softwareengineering.stackexchange.com/questions/260818/why-patch-method-is-not-idempotent

Perhaps it'd be a good idea to include `PATCH` in a separate paragraph, along the lines of "{{HTTPMethod("PATCH")}} is a special case...",

mb21:
From your link:

> A PATCH request can be idempotent, but it isn't required to be. That is the reason it is characterized as non-idempotent.

That's exactly right. And it's exactly the same for `POST`. Perhaps the whole article should be clarified then? I can take a stab at it.

chrisdavidmills:
> From your link:
> 
> > A PATCH request can be idempotent, but it isn't required to be. That is the reason it is characterized as non-idempotent.
> 
> That's exactly right. And it's exactly the same for `POST`. Perhaps the whole article should be clarified then? I can take a stab at it.

Yeah, I think it might be a good idea. I read the "That is the reason it is characterized as non-idempotent" line and thought, "well, that sounds a bit hand-wavy and vague".

mb21:
Updated! Let me know whether its clearer now. btw. actually the linked spec is quite well-written: https://httpwg.org/specs/rfc9110.html#idempotent.methods

files/en-us/glossary/idempotent/index.md
```diff
@@ -7,13 +7,17 @@ sidebar: glossarysidebar
...
 
-This does not necessarily mean that the request does not have _any_ unique side effects: for example, the server may log every request with the time it was received. Idempotency only applies to effects intended by the client: for example, a POST request intends to send data to the server, or a DELETE request intends to delete a resource on the server.
+The HTTP specification defines several HTTP methods and their semantics, which includes whether they are idempotent or not. All {{glossary("Safe/HTTP", "safe")}} methods are idempotent, as well as {{HTTPMethod("PUT")}} and {{HTTPMethod("DELETE")}}. The {{HTTPMethod("POST")}} and {{HTTPMethod("PATCH")}} methods are not guaranteed to be idempotent.
 
-All {{glossary("Safe/HTTP", "safe")}} methods are idempotent, as well as {{HTTPMethod("PUT")}} and {{HTTPMethod("DELETE")}}. The {{HTTPMethod("POST")}} method is not idempotent.
+Idempotent methods are distinguished because the client can retry a request if its not sure whether the request reached the server. If both requests happen to reach the server, as long as the method is idempotent, no harm is done.
```
chrisdavidmills:
```suggestion
A client can safely retry a request that uses an idempotent method, for example, in cases where there is doubt as to whether the request reached the server. If multiple identical requests happen to reach the server, as long as the method is idempotent, no harm is done.
```

files/en-us/glossary/idempotent/index.md
```diff
@@ -7,13 +7,17 @@ sidebar: glossarysidebar
...
 
-To be idempotent, only the state of the server is considered. The response returned by each request may differ: for example, the first call of a {{HTTPMethod("DELETE")}} will likely return a {{HTTPStatus("200")}}, while successive ones will likely return a {{HTTPStatus("404")}}. Another implication of {{HTTPMethod("DELETE")}} being idempotent is that developers should not implement RESTful APIs with a _delete last entry_ functionality using the `DELETE` method.
+Idempotency, as defined by the HTTP specification, only considers the _intented_ effect of the client on the server: for example, a POST request intends to send data to the server, or a DELETE request intends to delete a resource on the server. A request with an idempotent method does not necessarily mean that the request does not have _any_ unique side effects: for example, the server may log every request with the time it was received.
```
chrisdavidmills:
```suggestion
In the context of HTTP, idempotency only considers the _intended_ effect of the client on the server. For example, a POST request intends to send data to the server, whereas a DELETE request intends to delete a resource on the server. A request with an idempotent method does not necessarily mean that the request has _no_ unique side effects: For example, the server may log the time each request is received.
```

You mentioned something to the effect of "defined by HTTP specification" twice, so I reworded one to get rid of repetition.

files/en-us/glossary/idempotent/index.md
```diff
@@ -7,13 +7,17 @@ sidebar: glossarysidebar
...
 
-Note that the idempotence of a method is not guaranteed by the server and some applications may incorrectly break the idempotence constraint.
+Also, the response returned by each request may differ: for example, the first call of a {{HTTPMethod("DELETE")}} will likely return a {{HTTPStatus("200")}}, while successive ones will likely return a {{HTTPStatus("404")}}. Another implication of {{HTTPMethod("DELETE")}} being idempotent is that developers should not implement RESTful APIs with a _delete last entry_ functionality using the `DELETE` method.
```
chrisdavidmills:
```suggestion
In addition, the response returned by each request may differ. For example, an initial {{HTTPMethod("DELETE")}} request will likely return a {{HTTPStatus("200")}} status, while successive identical `DELETE` requests will probably return a {{HTTPStatus("404")}}. Another implication of {{HTTPMethod("DELETE")}} being idempotent is that developers should not implement RESTful APIs with a _delete last entry_ function that uses the `DELETE` method.
```

You might also want to add another sentence to explain why the last sentence is true.

files/en-us/glossary/idempotent/index.md
```diff
@@ -7,13 +7,17 @@ sidebar: glossarysidebar
...
+
+While servers are very much encouraged to adhere to the semantics laid out by the HTTP specification, the spec does not mandate it. There is nothing preventing a server in the wild from exposing a non-idempotent endpoint under an idempotent HTTP method – although clients will be surprised.
```
chrisdavidmills:
```suggestion
While servers are very much encouraged to adhere to the semantics laid out by the HTTP specification, the spec does not mandate it. Nothing is preventing a server in the wild from exposing a non-idempotent endpoint under an idempotent HTTP method, although clients may well be surprised.
```

files/en-us/glossary/idempotent/index.md
```diff
@@ -7,13 +7,17 @@ sidebar: glossarysidebar
...
+
+A few examples:
```
chrisdavidmills:
```suggestion
## Examples
```

chrisdavidmills (changes requested):
This is great, thanks @mb21! I've included a few wording suggestions, but I think this is pretty close now.

files/en-us/glossary/idempotent/index.md
mb21: We can also delete that last sentence. It was already there before, and while true, it's a bit of a weird example, if you ask me.

files/en-us/glossary/idempotent/index.md
mb21: I wouldn't mind the repetition, as it's kinda crucial to understand that we're only talking about the HTTP spec here.

I was actually looking into this for this chapter of a guide I'm writing: https://mastrojs.github.io/guide/http-forms-and-rest-apis/#client-side-fetching-a-rest-api and worded it there:

> While the HTTP specification only talks about the effect on the server that the client _intended_, in practice it falls to the server to make sure the routes it exposes actually adhere to these semantics.

Not sure that's sufficiently clear? Because in practice, whether a request actually is idempotent  or not (like whether actually it has the same effect whether it comes in multiple times or once) is entirely up to the server. But the HTTP spec didn't want to mandate that to servers, so it only talks about the effects the client _intended_. So they kinda say "btw FYI server: if a client sends you a `GET`, and since we're assuming the client adheres to the spec, then obviously the client intended an idempotent operation."

chrisdavidmills: I don't see the repetition as a deal breaker.

I like the sentence you've provided here, but it is a bit different in form to the original you provided. Maybe we could combine the two, maybe something like:

```suggestion
The HTTP specification only defines idempotency in terms of the _intented_ effect of the client on the server. For example, a `POST` request intends to send data to the server, whereas a `DELETE` request intends to delete a resource on the server. In practice, it falls to the server to make sure the routes it exposes adhere to these semantics. Also, a request with an idempotent method does not necessarily mean that the request has _no_ unique side effects: For example, the server may log the time each request is received.
```

WDYT?

files/en-us/glossary/idempotent/index.md
chrisdavidmills: I am happy for you to delete it.

files/en-us/glossary/idempotent/index.md
mb21: Ah, I committed that suggestion, only to realize it's kinda the same point as the last paragraph?

> While servers are very much encouraged to adhere to the semantics laid out by the HTTP specification, the spec does not mandate it. Nothing is preventing a server in the wild from exposing a non-idempotent endpoint under an idempotent HTTP method, although clients may well be surprised.

chrisdavidmills: I did think about that, but I don't think it is making exactly the same point.

The first instance is saying that it is down to the server to adhere to those semantics.

The second instance is a follow-on — it says that servers are encouraged to adhere to the semantics, but nothing is forcing them to.

Tell you what, I'll edit it and move things around a bit, to make this link clearer, then I'll ask you what you think of the result.

chrisdavidmills: @mb21 OK, done. WDYT? I'm not sure how long it will take to propagate to the live preview, but you can view the changes immediately in the "Files changed" tab.

files/en-us/glossary/idempotent/index.md
```diff
@@ -7,13 +7,21 @@ sidebar: glossarysidebar
...
+The HTTP specification only defines idempotency in terms of the _intented_ effect of the client on the server. For example, a `POST` request intends to send data to the server, whereas a `DELETE` request intends to delete a resource on the server. In practice, it falls to the server to make sure the routes it exposes adhere to these semantics.
 
-Note that the idempotence of a method is not guaranteed by the server and some applications may incorrectly break the idempotence constraint.
+> [!NOTE]
+> While servers are very much encouraged to adhere to the semantics laid out by the HTTP specification, the spec does not mandate it. Nothing is preventing a server in the wild from exposing a non-idempotent endpoint under an idempotent HTTP method, although clients may be surprised.
+
+Also, bear in mind:
+
+- A request with an idempotent method does not necessarily mean that the request has _no_ unique side effects: For example, the server may log the time each request is received.
```
mb21:
```suggestion
- A request with an idempotent method does not necessarily mean that the request has _no_ side effects on the server, only that the client intended none: For example, the server may log the time each request is received.
```

files/en-us/glossary/idempotent/index.md
mb21: nice! I just added one minor clarification, then I think it's good to go.

chrisdavidmills: Perfect, that looks good!

chrisdavidmills (approved):
OK, we are ready to publish this one. Thanks so much for your work on this, @mb21 — this reads much better now.

PR #40370, opened 2025-07-15
feat: move Quick_Links into front matter

bsmth:
### Description

For the "Related" and "IMSC" sections:

* Add yaml sidebars
* Use front matter sidebars
* Remove `Quick_links`

### Motivation

We're setting sidebars in front matter.

### DIFF
--- a/files/sidebars/related.yaml
+++ b/files/sidebars/related.yaml
@@ -1,0 +1,15 @@
+# Do not add comments to this file. They will be lost.
+
+sidebar:
+  - type: section
+    link: /Related
+  - type: section
+    link: /Related/IMSC
+  - /Related/IMSC/Basics
+  - /Related/IMSC/Using_the_imscJS_polyfill
+  - /Related/IMSC/Styling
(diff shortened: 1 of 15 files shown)


Josh-Cena (approved):
:+1:

files/sidebars/related.yaml
Josh-Cena, line 1: IMO: if we think this "related" thing is a good idea (which I'm skeptical) and we plan to add more stuff here, then we should have a new sidebar for each technology, just like the API area, because it will necessarily be ever-growing. Then the "Related" page itself would not have a sidebar, just like the status quo.

bsmth: >if we think this "related" thing is a good idea

We don't :) -> we are likely going to move the IMSC docs elsewhere (into Media?). We need to discuss again, but that would make this section redundant.

Josh-Cena: That sounds better :)

bsmth:
Thanks, merging shortly 👍🏻

PR #40374, opened 2025-07-15
Improve wording of then() return value

### LINKED ISSUE #40352: Grammatical ambiguity in .then() return value description
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then

Return value

"Returns a new [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) immediately. This new promise is always pending when returned, regardless of the current promise's status.

One of the onFulfilled and onRejected handlers will be executed to handle the current promise's fulfillment or rejection. The call always happens asynchronously, even when the current promise is already settled. The behavior of the returned promise (call it p) depends on the handler's execution result, following a specific set of rules. If the handler function:

returns a value: p gets fulfilled with the returned value as its value.
doesn't return anything: p gets fulfilled with undefined as its value.
throws an error: p gets rejected with the thrown error as its value.
returns an already fulfilled promise: p gets fulfilled with that promise's value as its value.
returns an already rejected promise: p gets rejected with that promise's value as its value.
returns another pending promise: p is pending and becomes fulfilled/rejected with that promise's value as its value immediately after that promise becomes fulfilled/rejected."

In the second paragraph the phrase "the returned promise (call it p)" can be grammatically ambigious. As its a noun phrase, it can either mean 'p' is returned by .then() pointing back to the only returned promise in the previous context as antecedent or it can imply the handlers have returned a promise 'p' where its acts as noun to introduce the promise returned by the handler. And both are grammatically valid, which can create ambiguity.
In the later case where 'p' can grammatically mean the promise returned by handler, the sentence will be read as "The behavior of the returned promise (call it p) by handler depends on the handler's execution result, following a specific set of rules". It can be a circular sentence but its still grammatically valid , as behaviour of promise by handler can be a term and the handler execution result i.e a promise now can be another term.

Also the terms used can be inconsistent, as the promise returned by .then() is said to be "new promise" repeatedly, and the promise is on which .then() is attached is termed as "current promise". 
The introduction of term "returned promise" without it being specified can create confusion and ambiguity.

It would be better to specify directly to whom the returned promise 'p' belongs to, instead of needing to imply what it belongs to.

(https://github.com/mdn/content/commit/9a771cdf71361a2d7528fa729e1512461a97db28)
In above link it can be seen from and onwards line 54, it was directly specified that behaviour of promise returned by .then() is depend upon the handler execution result. Its easier to understand and direct.

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/javascript/reference/global_objects/promise/then`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/global_objects/promise/then/index.md
* Last commit: https://github.com/mdn/content/commit/544b843570cb08d1474cfc5ec03ffb9f4edc0166
* Document last modified: 2025-07-10T09:07:55.000Z

</details>

Josh-Cena:
Fix https://github.com/mdn/content/issues/40352.

@OP-TechTrekker How does this look to you?

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/promise/then/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/promise/then/index.md
@@ -49,7 +49,7 @@ then(onFulfilled, onRejected)
 
 Returns a new {{jsxref("Promise")}} immediately. This returned promise is always pending when returned, regardless of the current promise's status.
 
-One of the `onFulfilled` and `onRejected` handlers will be executed to handle the current promise's fulfillment or rejection. The call always happens asynchronously, even when the current promise is already settled. The behavior of the promise returned by `then()` (call it `p`) depends on the handler's execution result, following a specific set of rules. If the handler function:
+One of the `onFulfilled` and `onRejected` handlers will be executed to handle the current promise's fulfillment or rejection. The call always happens asynchronously, even when the current promise is already settled. The behavior of the promise returned by `then()` (referred to as `p` in the following list) depends on the handler's execution result, following a specific set of rules. If the handler function:
 
 - returns a value: `p` gets fulfilled with the returned value as its value.
 - doesn't return anything: `p` gets fulfilled with `undefined` as its value.
--- a/files/en-us/web/javascript/reference/global_objects/promise/then/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/promise/then/index.md
@@ -47,9 +47,9 @@ then(onFulfilled, onRejected)
 
 ### Return value
 
-Returns a new {{jsxref("Promise")}} immediately. This new promise is always pending when returned, regardless of the current promise's status.
+Returns a new {{jsxref("Promise")}} immediately. This returned promise is always pending when returned, regardless of the current promise's status.
 
-One of the `onFulfilled` and `onRejected` handlers will be executed to handle the current promise's fulfillment or rejection. The call always happens asynchronously, even when the current promise is already settled. The behavior of the returned promise (call it `p`) depends on the handler's execution result, following a specific set of rules. If the handler function:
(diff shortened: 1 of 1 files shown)


files/en-us/web/javascript/reference/global_objects/promise/then/index.md
```diff
@@ -47,9 +47,9 @@ then(onFulfilled, onRejected)
 
 ### Return value
 
-Returns a new {{jsxref("Promise")}} immediately. This new promise is always pending when returned, regardless of the current promise's status.
+Returns a new {{jsxref("Promise")}} immediately. This returned promise is always pending when returned, regardless of the current promise's status.
 
-One of the `onFulfilled` and `onRejected` handlers will be executed to handle the current promise's fulfillment or rejection. The call always happens asynchronously, even when the current promise is already settled. The behavior of the returned promise (call it `p`) depends on the handler's execution result, following a specific set of rules. If the handler function:
+One of the `onFulfilled` and `onRejected` handlers will be executed to handle the current promise's fulfillment or rejection. The call always happens asynchronously, even when the current promise is already settled. The behavior of the promise returned by `then()` (call it `p`) depends on the handler's execution result, following a specific set of rules. If the handler function:
```
chrisdavidmills:
```suggestion
One of the `onFulfilled` and `onRejected` handlers will be executed to handle the current promise's fulfillment or rejection. The call always happens asynchronously, even when the current promise is already settled. The behavior of the promise returned by `then()` (referred to as `p` in the following list) depends on the handler's execution result, following a specific set of rules. If the handler function:
```

chrisdavidmills: Maybe? The existing wording seemed a bit strange.

chrisdavidmills (approved):
FWIW, the wording makes sense to me. Just a suggestion for improving the wording.

OP-TechTrekker:
I am thankful that my suggestion helped to improve something.
It looks great now. Thank You ! again.

Josh-Cena:
Thanks both!

PR #22032, opened 2022-11-03
TE Expand warning for using header in HTTP/2

zargold:
I recently experienced a cryptic protocol error as a result of an upstream HTTP/2.0 interceptor (envoy mesh network/sidecar) reading a header for `Transfer-Encoding`. Due to the documentation here only warning about `chunked` I specified a different value than chunked in case there was a HTTP2 error. Turns out ALL usage of Transfer-Encoding is forbidden very explicitly in the HTTP/2.0 RFC now linked.

### Description

The documentation for the Transfer-Encoding header on MDN incorrectly advised that only `"chunked"` did not work with HTTP/2. In fact all of those headers listed should be indicated as "HTTP/1.1"-only. 

### Motivation

This misled me into hours of extra debugging as I did not use "chunked" value for Transfer-Encoding, but instead used `gzip` thinking that as long as I use a value other than chunked I'd be safe in case some upstream service used HTTP/2. In fact, I was unsafe in using any value for TransferEncoding and got a cryptic Protocol error which ended up resulting from passing any Transfer-Encoding header.

### Additional details

In the RFC I found for HTTP/2 Protocol:

It talks about how the Transfer-Encoding (and others) header MUST be removed before sending something with the HTTP/2 protocol:

> This means that an intermediary transforming an HTTP/1.x message to
   HTTP/2 will need to remove any header fields nominated by the
   Connection header field, along with the Connection header field
   itself.  Such intermediaries SHOULD also remove other connection-
   specific header fields, such as Keep-Alive, Proxy-Connection,
   Transfer-Encoding, and Upgrade, even if they are not nominated by the
   Connection header field.

It then explicitly talks about the Transfer-Encoding header as allowed only if used with the value of `trailers`. 
>    The only exception to this is the TE header field, which MAY be
   present in an HTTP/2 request; when it is, it MUST NOT contain any
   value other than "trailers".

Indeed, failure to do so with a service using HTTP/2 will result in a protocol error at least in my lived experience.

PR #40376, opened 2025-07-15
Transfer-Encoding Header: Remove mention of `trailers` value

mhils:
### Description

The page for `Transfer-Encoding` currently incorrectly mentions a permissible `trailers` value for HTTP/2. But `trailers` is only valid for the `TE` header. `Transfer-Encoding: trailers` is not a thing. There already is a (correct) note on https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/TE.

The PR that introduced this (https://github.com/mdn/content/pull/22032) correctly cites the RFC for `TE`, but then mixes up TE and Transfer-Encoding.

### Additional details

From https://datatracker.ietf.org/doc/html/rfc9113#section-8.2.2 (emphasis mine):

> An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields. This includes the Connection header field and those listed as having connection-specific semantics in [Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)] (that is, Proxy-Connection, Keep-Alive, **Transfer-Encoding**, and Upgrade).

> The only exception to this is the **TE** header field, which MAY be present in an HTTP/2 request; when it is, it MUST NOT contain any value other than "trailers".

Relates to https://github.com/mdn/content/pull/22032

### DIFF
--- a/files/en-us/web/http/reference/headers/transfer-encoding/index.md
+++ b/files/en-us/web/http/reference/headers/transfer-encoding/index.md
@@ -22,7 +22,7 @@ The header is optional in responses to a {{HTTPMethod("HEAD")}} request as these
 When present it indicates the value that would have applied to the corresponding response to a {{HTTPMethod("GET")}} message, if that `GET` request did not include a preferred `Transfer-Encoding`.
 
 > [!WARNING]
-> HTTP/2 disallows all uses of the `Transfer-Encoding` header other than the HTTP/2 specific value `"trailers"`.
+> HTTP/2 disallows all uses of the `Transfer-Encoding` header.
 > HTTP/2 and later provide more efficient mechanisms for data streaming than chunked transfer.
 > Usage of the header in HTTP/2 may likely result in a specific `protocol error`.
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/reference/headers/transfer-encoding/index.md
```diff
@@ -105,5 +105,4 @@ Developer Network\r\n
 - {{HTTPHeader("Accept-Encoding")}}
 - {{HTTPHeader("Content-Encoding")}}
 - {{HTTPHeader("Content-Length")}}
-- Header fields that regulate the use of trailers: {{HTTPHeader("TE")}} (requests) and {{HTTPHeader("Trailer")}} (responses).
```
hamishwillee, line 108: Not relevant if trailers is not allowed.

hamishwillee (approved):
Thank you very much - good catch.

And welcome to MDN!

PR #40384, opened 2025-07-15
Synchronize with BCD v6.0.32

OnkarRuikar:
[BCD release v6.0.32](https://github.com/mdn/browser-compat-data/releases/tag/v6.0.32)

/cc cc/ @bsmth @wbamberg


Josh-Cena (approved):
On second thought, including browser-compat for related website sets is a bad idea, because it's not actually the same spec as any web feature. LGTM otherwise.

PR #40390, opened 2025-07-16
Editorial review: Add docs for the HighlightRegistry.highlightsFromPoint() method

chrisdavidmills:
### Description

Chrome 140 adds support for the [`HighlightRegisty.highlightsFromPoint()`](https://drafts.csswg.org/css-highlight-api-1/#dom-highlightregistry-highlightsfrompoint) method; see https://chromestatus.com/feature/4552801607483392.

This PR adds docs for the new method.

The Spec and browser compat tables are not currently showing, but will soon. The necessary data is added in https://github.com/mdn/browser-compat-data/pull/27310.


files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,198 @@
...
+### Parameters
+
+- `x`
+  - : The x coordinate of the point within the viewport to return custom highlight information for.
+- `y`
+  - : The y coordinate of the point within the viewport to return custom highlight information for.
+- `options` {{optional_inline}}
+  - : An object containing options, which can include:
+    - `shadowRoots`
+      - : An array of {{domxref("ShadowRoot")}} objects. Any shadow roots included in this array will have custom highlights applied inside them that exist at the specified point included in the return value. By default, shadow root custom highlights are not included in the return value.
```
ffiori: The second sentence wasn't so easy for me to follow, would you consider something like this instead?

An array of {{domxref("ShadowRoot")}} objects. If provided, custom highlights that exist at the specified point inside these shadow roots will be included in the result. By default, highlights inside shadow roots are not returned unless you explicitly specify them with this option.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Return value
+
+An array of objects representing the custom highlights applied at the specified point. Each object contains the following properties:
```
ffiori: Should we mention `HighlightHitResult` type here so developers know what to expect exactly? Maybe something like:

Returns an array of [HighlightHitResult](https://drafts.csswg.org/css-highlight-api-1/#dictdef-highlighthitresult) objects representing the custom highlights applied at the specified point. Each HighlightHitResult object contains the following properties:

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,198 @@
...
+
+- `highlight`
+  - : A {{domxref("Highlight")}} object representing the applied custom highlight.
+- `ranges`
+  - : An array of {{domxref("AbstractRange")}} objects representing the ranges the applied custom highlight is applied to.
```
ffiori: Should we avoid using "apply" twice here? Probably something like

```suggestion
  - : An array of {{domxref("AbstractRange")}} objects representing the ranges to which the custom highlight is applied.
```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,198 @@
...
+
+:root::highlight(highlight3) {
+  background-color: rgb(0 0 255 / 0.75);
+  color: white;
+}
+```
+
+#### JavaScript
+
+In our script, we start by grabbing references to the text node inside our `<p>` element, and our `<section>` element. We then create an array called `highlights` than contains our custom highlight names, and a variable called `curHighlight`, initially set to `0`, which will be used to specify which custom highlight to apply later on.
```
ffiori: Small typo on "than"

```suggestion
In our script, we start by grabbing references to the text node inside our `<p>` element, and our `<section>` element. We then create an array called `highlights` that contains our custom highlight names, and a variable called `curHighlight`, initially set to `0`, which will be used to specify which custom highlight to apply later on.
```

ffiori: Consider renaming `curHighlight` to `currentHighlight` for better clarity

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,198 @@
...
+For each highlight, we grab the first range in the [`ranges`](#ranges) array (there is only ever one range in each highlight, in this case), then get a reference to the node that contains the highlight via the {{domxref("AbstractRange.startContainer")}} property. We then get the exact highlighted string by accessing this node's [`textContent`](/en-US/docs/Web/API/Node/textContent) and running {{jsxref("String.substring()")}} on it, with the offsets specified as the range's {{domxref("AbstractRange.startOffset")}} and {{domxref("AbstractRange.endOffset")}}.
+
+The final step is to add the highlighted text inside the `<section>` element's `innerHTML`, inside an `<article>` element.
+
+```js live-sample___highlights-from-point-example
+document.addEventListener("dblclick", (event) => {
+  const mouseX = event.clientX;
+  const mouseY = event.clientY;
+  const highlights = CSS.highlights.highlightsFromPoint(mouseX, mouseY);
+  if (highlights.length !== 0) {
```
ffiori: is this `if` statement needed? Should the bottom fragments of text be cleaned if the user double clicks outside of any highlights?

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,198 @@
...
+    section.innerHTML = "";
+    for (highlight of highlights) {
+      const range = highlight.ranges[0];
+      const highlightedContainer = range.startContainer;
+      const textSelection = highlightedContainer.textContent.substring(
```
ffiori: The demo doesn't work if the user selects all the text (for example with triple click), then h, then double clicks anywhere in the text. It should output all the text in the boxes below, but instead it displays empty boxes like this:

image

I think you could use Range.toString() here, that problem would be solved and it would be simpler code.

ffiori (commented):
Thanks for working on this article! I left some comments after a first review but it's looking great.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Yeah, that was horrible. I found it quite tricky to construct this sentence, and your suggestion is certainly better.

I played with it a bit more and came up with this:

> An array of {{domxref("ShadowRoot")}} objects. Custom highlights that exist at the specified point inside shadow roots in the array will be included in the return value. By default, highlights inside shadow roots are not returned.

WDYT?

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: We don't usually do that, TBH. We usually either just say "an object" if it is small and can be fitted on the same page, or we create a separate dictionary page with the same name as the dictionary and link to it. We also don't tend to link to the spec unless absolutely necessary — the idea being that anything web devs need to know should be included on MDN.

But I can see how it is worth mentioning the name for the case where someone is reading about this method on MDN, and still wants to go look it up in the spec to find some extra information (browser engineers, for example!)

I've added the first instance, but not linked it. I'll see what the editorial review thinks, when it comes to that stage.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Good call; I've used your wording.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: fixed, ta!

chrisdavidmills: ok, done.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Yeah, I think you are right. I've made this change.

I also added an if check to the first function that checks whether the selected range start and end container are both the paragraph text node, and only then applies a highlight. I noticed that you can add highlights outside the paragraph, including cross-container highlights. That really messed the demo up ;-)

chrisdavidmills: I also updated the name of the initial `highlights` array to `customHighlights`, to avoid confusion with the same-named value in the second function.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Ah ha, good catch. I overengineered this one; I'd not even thought of using `toString()` ;-)

It does, of course, work much better, and gets rid of that bug. Updated!

chrisdavidmills:
Thanks for the review, @ffiori! I've responded to all your comments. Please let me know if you think this needs anything else, or give me a "LGTM" if you think it is ready to go to the editorial review stage.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
ffiori: Oh I see, ok that makes sense, thanks for the explanation, mentioning the name sounds good to me 👍

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
ffiori: That looks even better, thanks! Do you think it's clear that highlights in the light DOM are still returned when there are shadow roots provided? I'm thinking maybe some web dev can get confused thinking that option is for returning exclusively those highlights present in the shadow roots that you pass. I'd probably add an "also" in the second sentence there, or maybe even something like "alongside those in the light DOM". Dropping both suggestions below, let me know your thoughts.

> An array of {{domxref("ShadowRoot")}} objects. Custom highlights that exist at the specified point inside shadow roots in the array will also be included in the return value. By default, highlights inside shadow roots are not returned.

> An array of {{domxref("ShadowRoot")}} objects. Custom highlights that exist at the specified point inside shadow roots in the array will be included in the return value, alongside those present in the light DOM. By default, highlights inside shadow roots are not returned.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Examples
+
+### Output custom highlights applied at the mouse pointer position
+
+In this example, we provide a paragraph of text that you can apply custom highlights to. You can then double-click a point in the paragraph to output the text that has custom highlights applied at that point.
+
+#### HTML
+
+The markup contains a {{htmlelement("p")}} element containing the text to apply custom highlights to, and a {{htmlelement("section")}} element that we will output the highlighted text fragments to.
```
ffiori: Small suggestion to avoid using "contain" twice in the same sentence:

```suggestion
The markup includes a {{htmlelement("p")}} element containing the text to apply custom highlights to, and a {{htmlelement("section")}} element that we will output the highlighted text fragments to.
```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,193 @@
...
+
+Next, we check that the `range` object's [`startContainer`](/en-US/docs/Web/API/Range/startContainer) and [`endContainer`](/en-US/docs/Web/API/Range/endContainer) are both the paragraph text node, to make sure we don't allow any cross-container highlights. If so, we select the custom highlight we want to apply to the `range` using `customHighlights[currentHighlight]`, then increment `currentHighlight`; if it reaches `3`, we set it back to `0`. This has the effect of cycling through the available highlights in order as they are set.
+
+Finally for the `keydown` event handler, we create a new highlight object using the {{domxref("Highlight.Highlight", "Highlight()")}} constructor, passing it the `range` we created earlier. We then apply the chosen custom highlight style to the `highlight` using the {{domxref("HighlightRegistry.set()")}} method.
```
ffiori: `highlight` is mentioned in the second sentence but it's not specified where it came from. Do you think it makes sense to add something like this?

```suggestion
Finally for the `keydown` event handler, we create a new highlight object `highlight` using the {{domxref("Highlight.Highlight", "Highlight()")}} constructor, passing it the `range` we created earlier. We then apply the chosen custom highlight style to `highlight` using the {{domxref("HighlightRegistry.set()")}} method.
```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,193 @@
...
+```js live-sample___highlights-from-point-example
+pElem.addEventListener("dblclick", (event) => {
+  const mouseX = event.clientX;
+  const mouseY = event.clientY;
+  const highlights = CSS.highlights.highlightsFromPoint(mouseX, mouseY);
+
+  section.innerHTML = "";
+  for (highlight of highlights) {
+    const range = highlight.ranges[0];
+    const highlightedContainer = range.startContainer;
```
ffiori: I think `highlightedContainer` is not used anymore, right?

```suggestion
```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,193 @@
...
+      const highlight = new Highlight(range);
+      CSS.highlights.set(highlightStyle, highlight);
+    }
+  }
+});
+```
+
+Next, we define a [`dblclick`](/en-US/docs/Web/API/Element/dblclick_event) event handler function to handle outputting the highlighted text at the mouse cursor position when the event fires. We first grab the current mouse coordinates from the event object ({{domxref("MouseEvent")}}). We then pass those coordinates into a `highlightsFromPoint()` call to return the custom highlights applied at that point. We then clear the contents of the `<section>` element, and loop through each highlight in the `highlights` array.
+
+For each `highlight`, we grab the first range in the [`ranges`](#ranges) array (there is only ever one range in each highlight, in this case), then get a reference to the node that contains the highlight via the {{domxref("AbstractRange.startContainer")}} property. We then get the exact highlighted string using {{domxref("Range.toString()")}}, and add the highlighted text inside the `<section>` element's `innerHTML`, inside an `<article>` element.
```
ffiori: I think `highlightedContainer` is not used anymore after modifying the code, right? I think this paragraph could be updated too now, maybe something like below? Feel free to modify it.

```suggestion
For each `highlight`, we grab the first range in the [`ranges`](#ranges) array (there is only ever one range in each highlight, in this case) and get the exact highlighted string using {{domxref("Range.toString()")}}. This text is then added inside the `<section>` element's `innerHTML`, inside an `<article>` element.
```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
ffiori: Awesome, great changes. Thanks!

ffiori (commented):
Left a couple comments after reviewing the new version. And thanks for addressing the previous ones, it's looking really good!

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Yeah, that's a good consideration. I've updated it to your second suggestion.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: updated

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: I've done something similar to fix this; I just put "create a new `highlight` object" rather than "create a new highlight object `highlight`"

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Yes, really good point. I've removed it, and also simplified the preceding explanation to suit.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Done

chrisdavidmills:
That's the last few bits of cleanup dealt with; thanks @ffiori for taking the time to provide me with a great tech review.

I think this is now ready to go to the editorial review stage.

ffiori:
> That's the last few bits of cleanup dealt with; thanks @ffiori for taking the time to provide me with a great tech review.
> 
> I think this is now ready to go to the editorial review stage.

Agreed, LGTM! Thanks for addressing all the feedback and putting all this together :)

(I can't hit approve because it seems I don't have the permissions to do so in this repo but consider it approved by me)

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+
+#### Result
+
+Try selecting some sections of text, pressing <kbd>h</kbd> after each one to apply a highlight. Then try double-clicking on a highlighted section of text to see that section outputted at the bottom of the UI. If multiple highlights overlap the section, you'll see multiple text fragments outputted.
+
+{{EmbedLiveSample("highlights-from-point-example", "100%", "600")}}
+
+## Specifications
+
+{{Specifications}}
```
estelle, line 195: saying no spec, but i see https://drafts.csswg.org/css-highlight-api/#dom-highlightregistry-highlightsfrompoint

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
+- {{domxref("css_custom_highlight_api", "The CSS Custom Highlight API", "", "nocode")}}
```
estelle, line 203: - add the CSS module
- remove the css tricks (we link to it a gazillion other places, and it doesn't mention this property).

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+
+```js-nolint
+highlightsFromPoint(x, y)
+highlightsFromPoint(x, y, options)
+```
+
+### Parameters
+
+- `x`
+  - : The x coordinate of the point within the viewport to return custom highlight information for.
```
estelle:
```suggestion
  - : The x-coordinate of the point within the viewport from which to return custom highlight information.
```
to not end in "for". Same for Y.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+- `y`
+  - : The y coordinate of the point within the viewport to return custom highlight information for.
```
estelle:
```suggestion
  - : The y-coordinate of the point within the viewport from which to return custom highlight information.
```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+- `options` {{optional_inline}}
+  - : An object containing options, which can include:
+    - `shadowRoots`
+      - : An array of {{domxref("ShadowRoot")}} objects. Custom highlights that exist at the specified point inside shadow roots in the array will also be included in the return value, in addition to those present in the light DOM. By default, highlights inside shadow roots are not returned.
+
+### Return value
+
+An array of `HighlightHitResult` objects representing the custom highlights applied at the specified point. Each object contains the following properties:
```
estelle: We should have either:

```
### Return value

An array; a sequence of [highlightHitResult] objects each representing a custom highlight applied at the specified point, or an empty array if not custom highlights are applied at the specified point

````
and then add a highlightHitResult reference page.

alternatively, 

```
The **`highlightsFromPoint()`** method of the {{domxref("HighlightRegistry")}} interface returns an array of objects representing the custom highlights applied at a specific point within the viewport.

The method returns a sequence of `HighlightHitResult` objects. Each `HighlightHitResult` object represents a custom highlights applied at the point in the viewport specified by the x and y parameters. Each object contains a {{domxref("Highlight")}} object representing the applied custom highlight and a `ranges` array of {{domxref("AbstractRange")}} objects representing the ranges to which the custom highlight is applied.

If no custom highlights are applied at the specified point, or the specified point is outside the viewport, the method returns an empty array.

(then all the other stuff)

### Return value

An array; an empty array or a sequence of `HighlightHitResult` objects.

```

File: files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md [hunk 16]
```diff
@@ -0,0 +1,192 @@
...
+```html live-sample___highlights-from-point-example
+<h1>highlightsFromPoint() demo</h1>
+<p class="highlight-text">
+  Deadlights jack lad schooner scallywag dance the hempen jig carouser broadside
+  cable strike colors. Bring a spring upon her cable holystone blow the man down
+  spanker Shiver me timbers to go on account lookout wherry doubloon chase.
+  Belay yo-ho-ho keelhaul squiffy black spot yardarm spyglass sheet transom
+  heave to.
+</p>
+<h2>Highlighted text at point</h2>
```
estelle, line 64: this part isn't working.
i suggest adding an HR above this, or adding a thin border top.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+In this example, we provide a paragraph of text that you can apply custom highlights to. You can then double-click a point in the paragraph to output the text that has custom highlights applied at that point.
+
+#### HTML
+
+The markup includes a {{htmlelement("p")}} element containing the text to apply custom highlights to, and a {{htmlelement("section")}} element that we will output the highlighted text fragments to.
+
+```html live-sample___highlights-from-point-example
+<h1>highlightsFromPoint() demo</h1>
+<p class="highlight-text">
+  Deadlights jack lad schooner scallywag dance the hempen jig carouser broadside
```
estelle: optional, but i think it's a good idea: 

change the ipsum to an explanation of what is happening.

"When you select a subsection of  text, and then press the letter "h", the text you selected will be highlighted in yellow, red, or blue, in that order. ....."

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+- `ranges`
+  - : An array of {{domxref("AbstractRange")}} objects representing the ranges to which the custom highlight is applied.
+
+If no custom highlights are applied at the specified point, an empty array is returned. This includes instances where the specified point is outside the viewport, that is, the coordinate values are negative or greater than the viewport dimensions.
+
+## Examples
+
+### Output custom highlights applied at the mouse pointer position
+
+In this example, we provide a paragraph of text that you can apply custom highlights to. You can then double-click a point in the paragraph to output the text that has custom highlights applied at that point.
```
estelle:
```suggestion
In this example, you can apply custom highlights to up to three sections of content within a paragraph of text. These custom highlights can be overlapping. 
When the user clicks in the viewport, we use the `highlightsfrompoint()` method of the [highlightregistry]() interface to return the content of any custom highlights that are located where the user double-clicked
```

we should explain that we are creating the custom hightlights, and stying them, to demonstrate a use case. It wasn't clear why were doing all these other things to create an example where the method is relevant

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+
+#### HTML
+
+The markup includes a {{htmlelement("p")}} element containing the text to apply custom highlights to, and a {{htmlelement("section")}} element that we will output the highlighted text fragments to.
```
estelle:
```suggestion
The markup includes a {{htmlelement("p")}} element containing text to which you can apply custom highlights, and a {{htmlelement("section")}} element in which we will output the highlighted text fragments.
```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+  Belay yo-ho-ho keelhaul squiffy black spot yardarm spyglass sheet transom
+  heave to.
+</p>
+<h2>Highlighted text at point</h2>
+<section></section>
+```
+
+#### CSS
+
+We have hidden most of the CSS for brevity; the most relevant CSS defines styling for three custom highlights named `highlight1`, `highlight2`, and `highlight3`.
```
estelle: explain the `::hightlight()` pseudo element, and link the term.

maybe mention yellow, ,red, blue.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+  display: flex;
+  gap: 10px;
+}
+
+.highlight-text,
+article {
+  padding: 10px;
+  background-color: #eee;
+  border: 2px solid #ddd;
+  border-radius: 10px;
```
estelle:
```suggestion
  border-radius: 5px;
```
my preference. i like when the border is 50% of the padding

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+  color: white;
+}
+```
+
+#### JavaScript
+
+In our script, we start by grabbing references to the `<p>` element, its contained text node, and our `<section>` element. We then create an array called `customHighlights` that contains our custom highlight names, and a variable called `currentHighlight`, initially set to `0`, which will be used to specify which custom highlight to apply later on.
+
+```js live-sample___highlights-from-point-example
+const pElem = document.querySelector(".highlight-text");
```
estelle:
```suggestion
const pElem = document.querySelector(".highlightable-text");
```
pedantic

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+
+:root::highlight(highlight3) {
+  background-color: rgb(0 0 255 / 0.75);
+  color: white;
+}
+```
+
+#### JavaScript
+
+In our script, we start by grabbing references to the `<p>` element, its contained text node, and our `<section>` element. We then create an array called `customHighlights` that contains our custom highlight names, and a variable called `currentHighlight`, initially set to `0`, which will be used to specify which custom highlight to apply later on.
```
estelle:
```suggestion
We start by getting references to the `<p>` element, its contained text node, and the `<section>` element. We also define a `currentHighlight` variable which will be used to specify which custom highlight to apply later on.
```

estelle: I removed "We then create a `customHighlights` array with our custom highlight names, which is used in our CSS within our `::highlight()` pseudo-elements." 

in the codepen, you'll note that i was able to remove the array which I think simplifies the JS. If they're advanced enough to write the rest of this JS, they should be able to understand the iterator for the custom highlight names.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+
+```js live-sample___highlights-from-point-example
+const pElem = document.querySelector(".highlight-text");
+const textNode = pElem.firstChild;
+const section = document.querySelector("section");
+const customHighlights = ["highlight1", "highlight2", "highlight3"];
```
estelle:
```suggestion
```
my js suggestion below makes this not necessary, and can reduce the complexity.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+let currentHighlight = 0;
+```
+
+Next, we define a keyboard control — <kbd>h</kbd> — that when pressed applies a custom highlight to any text selected inside the paragraph text. This consists of a [`keydown`](/en-US/docs/Web/API/Element/keydown_event) event handler function that only runs its contents if the pressed key was <kbd>h</kbd>. If so, we grab the selected text using {{domxref("Window.getSelection()")}} and convert it to a {{domxref("Range")}} using {{domxref("Selection.getRangeAt()")}}.
```
estelle:
```suggestion
Next, we define a [`keydown`](/en-US/docs/Web/API/Element/keydown_event) event handler that applies a custom highlight to any selected text if the keyboard letter <kbd>h</kbd> keyboard is pressed. When <kbd>h</kbd> is pressed, the get the selected text using the {{domxref("Window.getSelection()")}} method and convert the text to a {{domxref("Range")}} using the {{domxref("Selection.getRangeAt()")}} method.
```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+
+In this example, we provide a paragraph of text that you can apply custom highlights to. You can then double-click a point in the paragraph to output the text that has custom highlights applied at that point.
+
+#### HTML
+
+The markup includes a {{htmlelement("p")}} element containing the text to apply custom highlights to, and a {{htmlelement("section")}} element that we will output the highlighted text fragments to.
+
+```html live-sample___highlights-from-point-example
+<h1>highlightsFromPoint() demo</h1>
+<p class="highlight-text">
```
estelle:
```suggestion
<p class="highlightable-text">
```
this isn't the highlight text, but it is what can be highlighted (pendantic)
we can't make this `contenteditable` b/c the "h" is usurped.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+window.addEventListener("keydown", (event) => {
+  if (event.key === "h") {
+    const selection = window.getSelection();
+    const range = selection.getRangeAt(0);
+    if (range.startContainer === textNode && range.endContainer === textNode) {
+      const highlightStyle = customHighlights[currentHighlight];
+      currentHighlight++;
+      if (currentHighlight === 3) {
+        currentHighlight = 0;
+      }
```
estelle: I did a codepen that shrinks the code a bit. See https://codepen.io/estelle/pen/jEbqrxx. I highlighted the edits i made so you know what to look for.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+#### JavaScript
+
+In our script, we start by grabbing references to the `<p>` element, its contained text node, and our `<section>` element. We then create an array called `customHighlights` that contains our custom highlight names, and a variable called `currentHighlight`, initially set to `0`, which will be used to specify which custom highlight to apply later on.
+
+```js live-sample___highlights-from-point-example
+const pElem = document.querySelector(".highlight-text");
+const textNode = pElem.firstChild;
+const section = document.querySelector("section");
```
estelle: move this to event handler (doesn't need to be global)

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+const customHighlights = ["highlight1", "highlight2", "highlight3"];
+let currentHighlight = 0;
+```
+
+Next, we define a keyboard control — <kbd>h</kbd> — that when pressed applies a custom highlight to any text selected inside the paragraph text. This consists of a [`keydown`](/en-US/docs/Web/API/Element/keydown_event) event handler function that only runs its contents if the pressed key was <kbd>h</kbd>. If so, we grab the selected text using {{domxref("Window.getSelection()")}} and convert it to a {{domxref("Range")}} using {{domxref("Selection.getRangeAt()")}}.
+
+Next, we check that the `range` object's [`startContainer`](/en-US/docs/Web/API/Range/startContainer) and [`endContainer`](/en-US/docs/Web/API/Range/endContainer) are both the paragraph text node, to make sure we don't allow any cross-container highlights. If so, we select the custom highlight we want to apply to the `range` using `customHighlights[currentHighlight]`, then increment `currentHighlight`; if it reaches `3`, we set it back to `0`. This has the effect of cycling through the available highlights in order as they are set.
```
estelle:
```suggestion
We check that the `range` object's [`startContainer`](/en-US/docs/Web/API/Range/startContainer) and [`endContainer`](/en-US/docs/Web/API/Range/endContainer) are both within the text node that is the paragraph's first child to ensure we don't try highlighting across multiple containers. If so, we defined a name for the custom highlight we want to apply to the `selectedRange` that is `highlight` appended by the modulo of the pre-incremented highlight couNTer (`highlightCount`).
```
range is an object and the variable name, which can be confusing, so let's use `selectedRange` instead.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,192 @@
...
+    const range = highlight.ranges[0];
+    const textSelection = range.toString();
+    section.innerHTML += `<article>${textSelection}</article>`;
+  }
+});
+```
+
+#### Result
+
+Try selecting some sections of text, pressing <kbd>h</kbd> after each one to apply a highlight. Then try double-clicking on a highlighted section of text to see that section outputted at the bottom of the UI. If multiple highlights overlap the section, you'll see multiple text fragments outputted.
```
estelle: we could put a chunk of this in the example.

estelle:
```suggestion
Selecting sections of text, then press <kbd>h</kbd> after each selection to apply a highlight. Once you have highlighted text, try double-clicking on a highlighted section of text, or overlapped highlighted areas, to see the text of the highlighted section(s) outputted at the bottom of the UI. When multiple highlights overlap, clicking in the overlapping area will lead to multiple text fragments being displayed.
```

estelle (commented):
The example is a good one, but there is so much that needs to happen to demonstrate this feature. I think it might be helpful to explain that the example is really two parts; the first step is creating the custom highlights, and the second is the point of the page, basically state that "this part isn't just so we can do the relevant part" (and maybe add a link there to the the create custom highlight examples, and to the page we just added yesterday).

There are a few minor edits too.


All the other edits are me trying to shrink the example. I did a CodePen, to hopefully reduce the complexity and shrink the explanation.  See https://codepen.io/estelle/pen/jEbqrxx. I highlighted the edits i made so you know what to look for.Please take a look at the codepen suggested shrinkage before reading all the code change suggestions. Then hopefully the feedback below won't be overwhelming. 

I did comment where I made suggested changes in that codepen. There are several variable name suggestions in the codepen.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Weird that it's not showing up on the preview URL. It works locally, and the `spec_url` is definitely in the data; see https://github.com/mdn/browser-compat-data/blob/main/api/HighlightRegistry.json#L243.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Done

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Well, I'm not going to add a separate reference page for the object, that seems pointless. 

I also don't understand your other suggestion. Some of the test suggestions are good (which I've used), but it is largely repetitive, and why would you put information about the return value at the top of the page, and not in the return value section?

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: I don't understand your comment — what's not working about it?

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Yeah, this is a good idea. It also saves me from having to add extra explanatory text to the demo.

chrisdavidmills: Done

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: OK, updated.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: updated

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Good call; done.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Yeah, this looks a bit more subtle. Updated.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: OK, I have taken on board several of your suggestions here, and it has made the code and explanation shorter. All good!

I didn't do everything you suggested, as in some cases, slightly longer code is better for understanding what is going on.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: done

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Text updated, array removed.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: removed

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: I've made most of the updates here.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: class updated

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: I don't know about this. Having it global doesn't do any harm, plus it means that only running this line once, rather than every time the handler function runs.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: I've applied quite a bit of this.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Yeah, I ended up doing that. With the explanation being in the example, I don't think we need it here as well.

chrisdavidmills: I transferred it to the example text, using some of these suggestions.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
estelle: sorry, should have removed my comment. I couldn't get it to work at first, then updated my chrome ;) oops. sorry about that.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,195 @@
...
+
+:root::highlight(highlight3) {
+  background-color: rgb(0 0 255 / 0.75);
+  color: white;
+}
+```
+
+#### JavaScript
+
+We start by grabbing references to the `<p>` element, its contained text node, and our `<section>` element. We then create a variable called `highlightCount`, initially set to `1`, which will be used to specify which custom highlight to apply later on.
```
estelle:
```suggestion
##### Create custom highlights

To create custom highlights, we start by grabbing references to the `<p>` element, its contained text node, and our `<section>` element. We then create a variable called `highlightCount`, initially set to `1`, which will be used to specify which custom highlight to apply later on.
```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,195 @@
...
+}
+
+:root::highlight(highlight3) {
+  background-color: rgb(0 0 255 / 0.75);
+  color: white;
+}
+```
+
+#### JavaScript
+
```
estelle, line 124:
```suggestion

To demonstrate the `highlightsFromPoint()` method, we need to have some highlights that can be returned. 

```

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
```diff
@@ -0,0 +1,195 @@
...
+        highlightCount = 1;
+      }
+      const highlight = new Highlight(selectedRange);
+      CSS.highlights.set(highlightName, highlight);
+    }
+  }
+});
+```
+
+Next, we define a [`dblclick`](/en-US/docs/Web/API/Element/dblclick_event) event handler function to handle outputting the highlighted text at the mouse cursor position when the event fires. We pass the current mouse coordinates into a `highlightsFromPoint()` call, clear the contents of the `<section>` element, then loop through each highlight in the `highlights` array.
```
estelle:
```suggestion
##### Returning custom highlights from a point

Now that we have the ability to create custom highlights, we can use the highlightsFromPoint() method to return custom highlights from a defined point.

We define a [`dblclick`](/en-US/docs/Web/API/Element/dblclick_event) event handler function to handle outputting the highlighted text at the mouse cursor position when the event fires. We pass the current mouse coordinates into a `highlightsFromPoint()` call, clear the contents of the `<section>` element, then loop through each highlight in the `highlights` array.
```

estelle (commented):
Looks good. I think it would be helpful to divide the example explanation into two parts, so I suggested a way of doing that... your text may vary.

Otherwise, LGTM.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Ah, ok. No worries ;-)

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: Yeah, this is quite a nice suggestion — see my next commit. I have also moved the `section` global definition into the second section, so it is right by the point where it gets used.

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: updated

files/en-us/web/api/highlightregistry/highlightsfrompoint/index.md
chrisdavidmills: updated

estelle (approved):
Yay!

PR #40391, opened 2025-07-16
Aggressively remove HTTP-only links

Josh-Cena:
A site without HTTPS is a common sign that it is obsolete. For many of these, the project hasn't been maintained for years, and there have been reports about the SSL cert being invalid/expired. This is a great chance to clean up these low-quality links. I think it's fine to set a very high standard for external links—HTTPS is a must regardless of the content of the target page.


files/en-us/web/accessibility/guides/seizure_disorders/index.md
```diff
@@ -615,9 +615,3 @@ Along with the Harding tool, is generally recognized to be one of the two "gold
 - [Understanding Success Criteria 1.4.3: Contrast (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html)
 - [Web Accessibility Initiative (WAI)](https://www.w3.org/WAI/)
 - [Web Content Accessibility Guidelines (WCAG) 2.2](https://w3c.github.io/wcag/guidelines/22/#dfn-relative-luminance) definition of relative luminance
-
-## Contributors
```
bsmth, line 619: I wonder if we can capture some of the acknowledgements somehow.

bsmth (approved):
Everything looks good, going to share with the team if there's any ideas on the the original contributor acknowledgements.

files/en-us/web/accessibility/guides/seizure_disorders/index.md
Josh-Cena: We usually don't have contributor acknowledgements anyway. These two pages are written in a far more academic way than the rest of the site, which I assume is the reason why we have it, but we should seek to reduce the amount of external references in general. If you look at https://jc-verse.github.io/mdn-graph/external-links, and group by page URL and sort by count, you see that this page has the most external links out of all non-relnote pages, and there isn't even a reason for it to.

bsmth: Right, this is a good move, I mean acknowledgements somewhere else outside of the reference/guide pages - we need to do something on this topic anyway with contributors.txt / wikihistory / credits

Josh-Cena:
lmk if there's any out of band team feedback, and you can merge if you think it's ready.

bsmth:
> lmk if there's any out of band team feedback, and you can merge if you think it's ready.

I don't think we need to block on that seeing as we have the diff, I've put it on our next internal meeting agenda

PR #40393, opened 2025-07-16
Update inconsistent var name in example

Erzender:
### Description

Updating a variable name from "someTargetElement" to "someElement" in a given code example

### Motivation

The previous name did not seem to match the one that is used in the very same example


dipikabh (changes requested):
Thanks! Appreciate you sending a PR.

Would you mind also updating the variable in line 26:
`console.log(target.innerHTML); // abc def`

dipikabh (approved):
Thanks @Erzender for your contribution!

Merging this now.

PR #40397, opened 2025-07-17
Mention shared memory in using web workers

Josh-Cena:
The current claim is incorrect: some data can be shared instead of copied. I added a very brief mention of `SharedArrayBuffer`, but did not go into details because that would be https://github.com/mdn/mdn/issues/341.

I also removed an unilluminating example about "simulating copy" because it's a very bad way to demonstrate this anyway, when you already have `structuredClone`.

### DIFF
--- a/files/en-us/web/api/web_workers_api/using_web_workers/index.md
+++ b/files/en-us/web/api/web_workers_api/using_web_workers/index.md
@@ -243,51 +243,9 @@ The exception to this is if the worker script's origin is a globally unique iden
 
 ## Transferring data to and from workers: further details
 
-Data passed between the main page and workers is **copied**, not shared. Objects are serialized as they're handed to the worker, and subsequently, de-serialized on the other end. The page and worker **do not share the same instance**, so the end result is that **a duplicate** is created on each end. Most browsers implement this feature as [structured cloning](/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/web_workers_api/using_web_workers/index.md
```diff
@@ -243,71 +243,27 @@ The exception to this is if the worker script's origin is a globally unique iden
...
-  this.type = type;
-  this.age = age;
-}
-const example5 = new Animal("Cat", 3);
-alert(example5.constructor); // Animal
-alert(emulateMessage(example5).constructor); // Object
-```
-
-A value that is cloned and not shared is called _message_. As you will probably know by now, _messages_ can be sent to and from the main thread by using `postMessage()`, and the `message` event's {{domxref("MessageEvent.data", "data")}} attribute contains data passed back from the worker.
+As will probably know by now, data is exchanged between the two threads via messages using `postMessage()`, and the `message` event's {{domxref("MessageEvent.data", "data")}} attribute contains data passed back from the worker.
```
sideshowbarker:
```suggestion
As you probably know by now, data is exchanged between the two threads via messages using `postMessage()`, and the `message` event's {{domxref("MessageEvent.data", "data")}} attribute contains data passed back from the worker.
```

sideshowbarker (approved):

PR #40400, opened 2025-07-17
docs: add a warning for `Customizable select elements`

paoloricciuti:
### Description

This PR adds a warning to the `Customizable select elements` article to warn about the fact that those features are not widely available.

### Motivation

The way this article is presented and described make it seems like this is the modern, generally available way to style selects ( https://github.com/sveltejs/svelte/issues/15617#issuecomment-3083080567 ). This couldn't be further from being true as this is only available in Chrome. Furthermore since the changes to the parser (again only available in Chrome) most Javascript frameworks that handle hydration will either prevent you from writing such code (like Svelte and Solid) or even worse just straight up fail hydration. It will also probably be a while before frameworks can adapt since they basically need to branch their hydration logic based on the availability of this feature.

I'm open to rephrase this if it's too scary but i think we should have some sort of warning here.

https://github.com/sveltejs/svelte/issues/15963
https://github.com/sveltejs/svelte/issues/15617


files/en-us/learn_web_development/extensions/forms/customizable_select/index.md
```diff
@@ -10,6 +10,9 @@ sidebar: learnsidebar
 
 This article explains how to use dedicated, modern HTML and CSS features together to create fully-customized {{htmlelement("select")}} elements. This includes having full control over styling the select button, drop-down picker, arrow icon, current selection checkmark, and each individual {{htmlelement("option")}} element.
```
chrisdavidmills:
```suggestion
This article explains how to create fully-customized {{htmlelement("select")}} elements using experimental browser features. This includes having full control over styling the select button, drop-down picker, arrow icon, current selection checkmark, and each individual {{htmlelement("option")}} element.
```

chrisdavidmills: I'd probably start by toning down the opening paragraph and making it clearer that this is currently experimental.

files/en-us/learn_web_development/extensions/forms/customizable_select/index.md
```diff
@@ -10,6 +10,9 @@ sidebar: learnsidebar
...
 
+> [!WARNING]
+> The CSS and HTML features demonstrated in this article have limited browser support and may not work reliably in production. Most JavaScript frameworks either block these features or cause hydration failures when Server-Side Rendering (SSR) is enabled.
```
chrisdavidmills: For the first sentence, maybe something like:

```suggestion
> The CSS and HTML features demonstrated in this article currently have limited browser support; check the browser compatibility tables on the individual feature reference pages for more details. Some JavaScript frameworks block these features; in others, they cause hydration failures when Server-Side Rendering (SSR) is enabled.
```

chrisdavidmills: For the second sentence — is this just because of the issue with the `<selectedcontent>` rendering (which led to us [adding the warning here](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/selectedcontent#description)), or is the problem more widespread than we thought?

I'll also get some other folks to look at this PR, who have more framework experience than me.

chrisdavidmills (commented):
Hi @paoloricciuti, and thanks for your consideration here. I agree that something needs to be added. I've provided some initial thoughts, and I'll get some other folks to look at this too.

files/en-us/learn_web_development/extensions/forms/customizable_select/index.md
paoloricciuti: I think `modern` here was an attempt at that...but yeah this feels better already

files/en-us/learn_web_development/extensions/forms/customizable_select/index.md
paoloricciuti: It's way more widespread (it's basically a breaking change for the web unfortunately). Even without using any frameworks if you did something like this before

```html
<template id="tmp">
  <select>
    <!-- this is wrong but it was possible nonetheless -->
    <option><b>Empty</b></option>
    <option>A</option>
    <option>B</option>
    <option>C</option>
  </select>
</template>
```
and then in JS
```js
function create_select(placeholder){
  const template = document.getElementById("tmp");
  const select = template.content.cloneNode(true);
  if(placeholder){    
    const option = select.querySelector("option");
	// this is the wrong way to do it obviously but it was possible nonetheless
    option.firstChild.data = placeholder;
  }
  document.body.append(select);
}

create_select("Cool");
```

Above code will work in Firefox and Safari and fail in Chrome.

Frameworks are not doing something like this obviously but let's say that we permit the usage of rich html selects. If the user does this

```svelte
<script>
  let count = $state(0);
</script>

<select>
  <option><strong>{count}</strong></option>
</select>
```
during hydration we need to get the `strong` tag with Javascript and set `innerText` to `count`...except that the `strong` tag will not be there in Firefox and Safari and the first element will actually be a text node. 😱

From my experiments every framework that allows you to write rich html select fails during hydration in Firefox and Safari if you do so.

chrisdavidmills: This is a bit of a tricky situation. I'm not getting much back from other interested parties, so I think for now I'm going to commit my suggestions and then merge it. 

We can always come back to it and make further updates if the behavior changes or if others have input.

chrisdavidmills (approved):
OK, let's get this in.

files/en-us/learn_web_development/extensions/forms/customizable_select/index.md
paoloricciuti: Looks sensible...thanks!

PR #40405, opened 2025-07-17
Internalize data attributes examples

Josh-Cena:
Almost finishing https://github.com/mdn/content/issues/16120.

I'm having a really hard time coming up with convincing examples, because the existing ones surely don't convince me. @pepelsbey pointed me to the new Fred architecture which is helpful. I'm also pointing to another internal example for reference.

### DIFF
--- a/files/en-us/web/html/how_to/use_data_attributes/index.md
+++ b/files/en-us/web/html/how_to/use_data_attributes/index.md
@@ -166,7 +166,7 @@ There are multiple benefits of this:
 
 ### Associating arbitrary data with DOM elements
 
-Most web apps have JavaScript data as the source-of-truth for their UI state, so you only add HTML attributes where needed. But occasionally, it may be more convenient to have everything be present in the markup, and only use JavaScript to handle events, sync state, etc.
+Many web apps have JavaScript data as the source-of-truth for their UI state. In these cases, you only add HTML attributes necessary for rendering. Data attributes are useful in the cases where everything is present in the markup, and JavaScript is only needed for handling events, syncing state, etc.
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/html/how_to/use_data_attributes/index.md
```diff
@@ -92,15 +92,87 @@ article[data-columns="4"] {
...
+
+There are multiple benefits of this:
+
+- It eliminates a lot of invalid states, such as applying `callout--note` without also adding `callout`, or applying multiple variants simultaneously.
+- A separate `data-variant` attribute allows static analysis for valid values via linting or type checking.
+- Toggling the variant is more intuitive: you can use `div.dataset.variant = "warning";` instead of manipulating the [`classList`](/en-US/docs/Web/API/Element/classList) which requires multiple steps.
+
+### Associating arbitrary data with DOM elements
+
+Most web apps have JavaScript data as the source-of-truth for their UI state, so you only add HTML attributes where needed. But occasionally, it may be more convenient to have everything be present in the markup, and only use JavaScript to handle events, sync state, etc.
```
estelle:
```suggestion
Many web apps have JavaScript data as the source-of-truth for their UI state. In these cases, you only add HTML attributes necessary for rendering. Data attributes are useful in the cases where everything is present in the markup, and JavaScript is only needed for handling events, syncing state, etc.
```
less opinionated and dissuasive. "Most"? not sure of the definitions, so let's make it less debatable

estelle (commented):
generally LGTM, but made one suggestion

estelle (approved):

PR #40409, opened 2025-07-17
chore: move EmbedGH live sample into the page

bsmth:
### Description

Moves an EmbedGH live sample into the page.

### Motivation

It's easier to maintain this way.


hamishwillee (approved):
Since I'm here, yay. 
We could do other things such as format this similar to the Web API examples etc. But that's in no way a requirement.

bsmth:
Nice, thanks!

PR #40411, opened 2025-07-17
Editorial review: Document recent FedCM additions, Chrome 122-136

chrisdavidmills:
### Description

Chrome has added several [FedCM](https://developer.mozilla.org/en-US/docs/Web/API/FedCM_API) features that we are missing from MDN. This PR adds docs for them.

The features are:

- Support for multiple IdPs in a single `get()` call, as specified inside the `providers` array, added in Chrome 136: https://chromestatus.com/feature/5067784766095360
- The above ChromeStatus entry also includes the `IdentityCredential.configURL` (if you look at the [associated spec changes](https://github.com/w3c-fedid/FedCM/pull/686/files#diff-40cc3a1ba233cc3ca7b6d5873260da9676f6ae20bb897b62f7871c80d0bda4e9R445))
- `IdentityCredential.disconnect()`, added in Chrome 122: https://chromestatus.com/feature/5202286040580096. 
- `identity.mode` in `get()` calls, added in Chrome 132: https://chromestatus.com/feature/4689551782313984. 
- `identity.fields` and `identity.params` in `get()` calls, added in Chrome 132. I'm not sure what ChromeStatus entry covers these, but [developer.chrome.com](https://privacysandbox.google.com/cookies/fedcm/implement/relying-party#providers_property) says they were added in Chrome 132.
- `account_label` / `label_hints` in the IdP config file / accounts list endpoint. This was added in Chrome 132. Note that this won't have browser compat data - we don't really have a way of recording data for features in endpoints related to an API.
- `supports_use_other_account` in the IdP config file.
- `domain_hints` in the accounts list endpoint and `domainHint` in `get()` calls.

One point I'd especially like the tech review to comment on — the specified [options object for `disconnect()`](https://w3c-fedid.github.io/FedCM/#dictdef-identitycredentialdisconnectoptions) does not match up with the description in the [Chrome docs](https://privacysandbox.google.com/cookies/fedcm/implement/relying-party#disconnect_the_idp_from_the_rp). I think the latter makes more sense, but I'd like to know what is going on here.

### DIFF
--- a/files/en-us/web/api/fedcm_api/idp_integration/index.md
+++ b/files/en-us/web/api/fedcm_api/idp_integration/index.md
@@ -48,6 +48,7 @@ The config file (hosted at `https://accounts.idp.example/config.json` in our exa
 {
   "accounts_endpoint": "/accounts.php",
   "client_metadata_endpoint": "/client_metadata.php",
+  "disconnect_endpoint": "/disconnect.php",
   "id_assertion_endpoint": "/assertion.php",
   "login_url": "/login",
   "branding": {
@@ -69,6 +70,8 @@ The properties are as follows:
   - : The URL for the accounts list endpoint, which returns a list of accounts that the user is currently signed in to on the IdP. The browser uses these to create a list of sign-in choices to show to the user in the browser-provided FedCM UI.
 - `client_metadata_endpoint` {{optional_inline}}
   - : The URL for the client metadata endpoint, which provides URLs pointing to the RP's metadata and terms of service pages, to be used in the FedCM UI.
+- `disconnect_endpoint` {{optional_inline}}
+  - : The URL for the disconnect endpoint, which is used by the RP to disconnect from the IdP, via the {{domxref("IdentityCredential.disconnect_static", "IdentityCredential.disconnect()")}} method.
 - `id_assertion_endpoint`
   - : The URL for the ID assertion endpoint, which when sent valid user credentials should respond with a validation token that the RP can use to validate the authentication.
 - `login_url`
@@ -83,6 +86,7 @@ The following table summarizes the different requests made by the FedCM API:
 | `well-known`/`config.json` | `GET`  | No                          | No                                |
 | `accounts_endpoint`        | `GET`  | Yes                         | No                                |
 | `client_metadata_endpoint` | `GET`  | No                          | Yes                               |
+| `disconnect_endpoint`      | `POST` | Yes                         | Yes                               |
 | `id_assertion_endpoint`    | `POST` | Yes                         | Yes                               |
 
 > [!NOTE]
@@ -174,6 +178,46 @@ The response to a successful request includes URLs pointing to the RP's metadata
   "terms_of_service_url": "https://rp.example/terms_of_service.html"
 }
 ```
+
+### The disconnect endpoint
+
+By invoking {{domxref("IdentityCredential.disconnect_static", "IdentityCredential.disconnect()")}}, the browser sends a cross-origin {{httpmethod("POST")}} request with cookies and a {{httpheader("Content-Type")}} of `application/x-www-form-urlencoded` to the disconnect endpoint with the following information:
+
+- `account_hint`
+  - : A string specifying an account hint that the IdP uses the identify the account to disconnect.
+- `client_id`
+  - : A string specifying the RP's client identifier.
+
+For example:
+
+```http
+POST /disconnect HTTP/1.1
+Host: idp.example
+Origin: rp.example
+Content-Type: application/x-www-form-urlencoded
+Cookie: 0x123
+Sec-Fetch-Dest: webidentity
+
+account_hint=account456&client_id=rp123
+```
+
+Upon receiving the request, the IdP server should:
+
+1. Respond to the request with [CORS (Cross-Origin Resource Sharing)](/en-US/docs/Web/HTTP/Guides/CORS).
+2. Verify that the request contains a {{httpheader("Sec-Fetch-Dest")}} HTTP header with a directive of `webidentity`.
+3. Match the {{httpheader("Origin")}} header against the RP origin determined by the `client_id`. Reject the promise if they don't match.
+4. Find the account that matches the `account_hint`.
+5. Disconnect the user account from the list of RP's connected accounts.
+6. Respond with the identified user's `account_id` in JSON format:
+
+   ```json
+   {
+     "account_id": "account456"
+   }
+   ```
+
+> [!NOTE]
+> If the IdP wishes to disconnect all accounts associated with the RP, it can pass a string that does not match any `account_id`, for example `"account_id": "*"`.
 
 ### The ID assertion endpoint
 
@@ -289,17 +333,17 @@ The IdP should update its login status when a user signs into or out of the IdP.
 
 When an [RP attempts federated sign-in](/en-US/docs/Web/API/FedCM_API/RP_sign-in), the login status is checked:
 
-- If the login status is `"logged-in"`, a request is made to the IdP's [accounts list endpoint](#the_accounts_list_endpoint) and available accounts for sign-in are displayed to the user in the browser-provided FedCM dialog.
-- If the login status is `"logged-out"`, the promise returned by the FedCM `get()` request rejects without making a request to the accounts list endpoint. In such a case it is up to the developer to handle the flow, for example by prompting the user to go and sign in to a suitable IdP.
-- If the login status is `"unknown"`, a request is made to the IdP's accounts list endpoint and the login status is updated depending on the response:
+- If an IdP's login status is `"logged-in"`, a request is made to the [accounts list endpoint](#the_accounts_list_endpoint) and available accounts for sign-in are displayed to the user in the browser-provided FedCM dialog.
+- If all IdPs' login statuses are `"logged-out"`, the promise returned by the FedCM `get()` request rejects without making a request to the accounts list endpoint. In such a case, it is up to the developer to handle the flow, for example by prompting the user to go and sign in to a suitable IdP.
+- If an IdP's login status is `"unknown"`, a request is made to the accounts list endpoint and the login status is updated depending on the response:
   - If the endpoint returns a list of available accounts for sign-in, update the status to `"logged-in"` and display the sign-in options to the user in the browser-provided FedCM dialog.
-  - If the endpoint returns no accounts, update the status to `"logged-out"`; the promise returned by the FedCM `get()` request then rejects.
+  - If the endpoint returns no accounts, update the status to `"logged-out"`; the promise returned by the FedCM `get()` request will reject in no other `logged-in` IdPs are available.
 
 ### What if the browser and the IdP login status become out of sync?
 
-Despite the Login Status API informing the browser of the IdP's login status, it is possible for the browser and IdP to become out of sync. For example, the IdP sessions might expire, meaning that all user accounts end up signed out but the login status is still set to `"logged-in"` (the application was not able to set the login status to `"logged-out"`). In such a case, when federated sign-in is attempted, a request will be made to the IdP's accounts list endpoint but no available accounts will be returned because the session is no longer available.
+Despite the Login Status API informing the browser of IdP login status, it is possible for the browser and an IdP to become out of sync. For example, the IdP sessions might expire, meaning that all user accounts end up signed out but the login status is still set to `"logged-in"` (the application was not able to set the login status to `"logged-out"`). In such a case, when federated sign-in is attempted, a request will be made to the IdP's accounts list endpoint but no available accounts will be returned because the session is no longer available.
 
-When this occurs, the browser can dynamically let a user sign into the IdP by opening the IdP's sign-in page in a dialog (the sign-in URL is found in the IdP's [config file](#provide_a_config_file_and_endpoints) `login_url` ). The exact nature of this flow is up to the browser; for example, [Chrome handles it like this](https://privacysandbox.google.com/blog/fedcm-chrome-120-updates#what_if_the_user_session_expires_let_the_user_sign_in_through_a_dynamic_login_flow).
+When this occurs, the browser can dynamically let a user sign into an IdP by opening the IdP's sign-in page in a dialog (the sign-in URL is found in the IdP's [config file](#provide_a_config_file_and_endpoints) `login_url`). The exact nature of this flow is up to the browser; for example, [Chrome handles it like this](https://privacysandbox.google.com/blog/fedcm-chrome-120-updates#what_if_the_user_session_expires_let_the_user_sign_in_through_a_dynamic_login_flow).
 
 Once the user is signed in to the IdP, the IdP should:
 
--- a/files/en-us/web/api/fedcm_api/idp_integration/index.md
+++ b/files/en-us/web/api/fedcm_api/idp_integration/index.md
@@ -353,7 +353,7 @@ When an [RP attempts federated sign-in](/en-US/docs/Web/API/FedCM_API/RP_sign-in
 - If all IdPs' login statuses are `"logged-out"`, the promise returned by the FedCM `get()` request rejects without making a request to the accounts list endpoint. In such a case, it is up to the developer to handle the flow, for example by prompting the user to go and sign in to a suitable IdP.
 - If an IdP's login status is `"unknown"`, a request is made to the accounts list endpoint and the login status is updated depending on the response:
   - If the endpoint returns a list of available accounts for sign-in, update the status to `"logged-in"` and display the sign-in options to the user in the browser-provided FedCM dialog.
-  - If the endpoint returns no accounts, update the status to `"logged-out"`; the promise returned by the FedCM `get()` request will reject in no other `logged-in` IdPs are available.
+  - If the endpoint returns no accounts, update the status to `"logged-out"`; the promise returned by the FedCM `get()` request will reject if no other `logged-in` IdPs are available.
 
 ### What if the browser and the IdP login status become out of sync?
 
--- a/files/en-us/web/api/fedcm_api/idp_integration/index.md
+++ b/files/en-us/web/api/fedcm_api/idp_integration/index.md
@@ -47,6 +47,8 @@ The config file (hosted at `https://accounts.idp.example/config.json` in our exa
 ```json
 {
   "accounts_endpoint": "/accounts.php",
+  "account_label": "developer",
+  "supports_use_other_account": true,
   "client_metadata_endpoint": "/client_metadata.php",
   "disconnect_endpoint": "/disconnect.php",
   "id_assertion_endpoint": "/assertion.php",
@@ -68,6 +70,12 @@ The properties are as follows:
 
 - `accounts_endpoint`
   - : The URL for the accounts list endpoint, which returns a list of accounts that the user is currently signed in to on the IdP. The browser uses these to create a list of sign-in choices to show to the user in the browser-provided FedCM UI.
+- `account_label` {{optional_inline}}
+  - : A string that, if included, specifies an identifier for a subset of accounts that should be returned when this IdP is used for federated authentication. When a `get()` request is made, only accounts matching this string in their `label_hints` parameters will be returned from the [accounts endpoint](#the_accounts_list_endpoint).
+- `supports_use_other_account` {{optional_inline}}
+  - : A boolean that defaults to `false`; if set to `true`, it means that users can sign in with an account different from the one they're currently logged in with (if the IdP supports multiple accounts). This only applies to `get()` calls that specify [active mode](/en-US/docs/Web/API/IdentityCredentialRequestOptions#active).
+    > [!NOTE]
+    > In the browser sign-in UI, this will likely manifest as some kind of "Choose other account" button.
 - `client_metadata_endpoint` {{optional_inline}}
   - : The URL for the client metadata endpoint, which provides URLs pointing to the RP's metadata and terms of service pages, to be used in the FedCM UI.
 - `disconnect_endpoint` {{optional_inline}}
@@ -121,6 +129,8 @@ The response to a successful request returns a list of all the IdP accounts that
       "email": "elaina_maduro@idp.example",
       "picture": "https://idp.example/profile/123",
       "approved_clients": ["123", "456", "789"],
+      "domain_hints": ["rp1.example.com", "rp3.example.com"],
+      "label_hints": ["developer", "admin"],
       "login_hints": ["elaina_maduro", "elaina_maduro@idp.example"]
     },
     {
@@ -130,6 +140,8 @@ The response to a successful request returns a list of all the IdP accounts that
       "email": "Elly@idp.example",
       "picture": "https://idp.example/profile/456",
       "approved_clients": ["abc", "def", "ghi"],
+      "domain_hints": ["rp1.example.com", "rp2.example.com"],
+      "label_hints": ["developer", "test"],
       "login_hints": ["elly", "elly@idp.example"]
     }
   ]
@@ -150,6 +162,10 @@ This includes the following information:
   - : The URL of the user's avatar image.
 - `approved_clients` {{optional_inline}}
   - : An array of RP clients that the user has registered with.
+- `domain_hints` {{optional_inline}}
+  - : An array of domains the account is associated with. The RP can make a `get()` call that includes a [`domainHint`](/en-US/docs/Web/API/IdentityCredentialRequestOptions#domainhint) property to filter the returned accounts by domain.
+- `label_hints` {{optional_inline}}
+  - : An array of strings specifying labels that define account types that the account is identified with. If the config file specifies an [`account_label`](#account_label), only accounts that contain that label in their `label_hints` will be returned from the accounts list endpoint.
 - `login_hints` {{optional_inline}}
   - : An array of strings representing the account. These strings are used to filter the list of account options that the browser offers for the user to sign-in. This occurs when the `loginHint` property is provided within [`identity.providers`](/en-US/docs/Web/API/IdentityCredentialRequestOptions#providers) in a related `get()` call. Any account with a string in its `login_hints` array that matches the provided `loginHint` is included.
 
--- a/files/en-us/web/api/fedcm_api/idp_integration/index.md
+++ b/files/en-us/web/api/fedcm_api/idp_integration/index.md
@@ -6,7 +6,7 @@ page-type: guide
 
 {{DefaultAPISidebar("FedCM API")}}
 
-This article details all the steps an identity provider (IdP) needs to take to integrate with the Federated Credential Management (FedCM) API.
+This article details all the steps an {{glossary("Identity provider", "identity provider")}} (IdP) needs to take to integrate with the Federated Credential Management (FedCM) API.
 
 ## IdP integration steps
 
@@ -18,9 +18,9 @@ To integrate with FedCM, an IdP needs to do the following:
 
 ## Provide a well-known file
 
-There is a potential privacy issue whereby an [IdP is able to discern whether a user visited an RP without explicit consent](https://github.com/w3c-fedid/FedCM/issues/230). This has tracking implications, so an IdP is required to provide a well-known file to verify its identity and mitigate this issue.
+There is a potential privacy issue whereby an [IdP is able to discern whether a user visited a relying party (RP) without explicit consent](https://github.com/w3c-fedid/FedCM/issues/230). This has tracking implications, so an IdP is required to provide a well-known file to verify its identity and mitigate this issue.
 
-The well-known file is requested via an uncredentialed [`GET`](/en-US/docs/Web/HTTP/Reference/Methods/GET) request, which doesn't follow redirects. This effectively prevents the IdP from learning who made the request and which RP is attempting to connect.
+The well-known file is requested via an uncredentialed [`GET`](/en-US/docs/Web/HTTP/Reference/Methods/GET) request, which doesn't follow redirects. This effectively prevents the IdP from learning who made the request and which {{glossary("Relying party", "RP")}} is attempting to connect.
 
 The well-known file must be served from the [eTLD+1](https://web.dev/articles/same-site-same-origin#site) of the IdP at `/.well-known/web-identity`. For example, if the IdP endpoints are served under `https://accounts.idp.example/`, they must serve a well-known file at `https://idp.example/.well-known/web-identity`. The well-known file's content should have the following JSON structure:
 
@@ -85,7 +85,7 @@ The properties are as follows:
 - `login_url`
   - : The login page URL for the user to sign into the IdP.
 - `branding` {{optional_inline}}
-  - : Contains branding information that will be used in the browser-supplied FedCM UI to customize its appearance as desired by the IdP.
+  - : Contains branding information that will be used in the browser-supplied FedCM UI to customize its appearance as desired by the IdP. The provided icon size must be greater than or equal to `25` (`25px`) in passive mode and greater than or equal to `40` (`40px`) in active mode (see [Active versus passive mode](/en-US/docs/Web/API/FedCM_API/RP_sign-in#active_versus_passive_mode) for more details).
 
 The following table summarizes the different requests made by the FedCM API:
 
--- a/files/en-us/web/api/fedcm_api/index.md
+++ b/files/en-us/web/api/fedcm_api/index.md
@@ -38,7 +38,8 @@ There are two parts to using the FedCM API, which are covered in the linked guid
 
 ## Permissions Policy integration and `<iframe>` support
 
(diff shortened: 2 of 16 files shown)


hamishwillee:
@chrisdavidmills Can you ping me or/and Will when you're ready for the editorial review.

files/en-us/web/api/fedcm_api/idp_integration/index.md
```diff
@@ -289,17 +349,17 @@ The IdP should update its login status when a user signs into or out of the IdP.
...
+- If an IdP's login status is `"logged-in"`, a request is made to the [accounts list endpoint](#the_accounts_list_endpoint) and available accounts for sign-in are displayed to the user in the browser-provided FedCM dialog.
+- If all IdPs' login statuses are `"logged-out"`, the promise returned by the FedCM `get()` request rejects without making a request to the accounts list endpoint. In such a case, it is up to the developer to handle the flow, for example by prompting the user to go and sign in to a suitable IdP.
+- If an IdP's login status is `"unknown"`, a request is made to the accounts list endpoint and the login status is updated depending on the response:
   - If the endpoint returns a list of available accounts for sign-in, update the status to `"logged-in"` and display the sign-in options to the user in the browser-provided FedCM dialog.
-  - If the endpoint returns no accounts, update the status to `"logged-out"`; the promise returned by the FedCM `get()` request then rejects.
+  - If the endpoint returns no accounts, update the status to `"logged-out"`; the promise returned by the FedCM `get()` request will reject in no other `logged-in` IdPs are available.
```
ewewraw:
```suggestion
  - If the endpoint returns no accounts, update the status to `"logged-out"`; the promise returned by the FedCM `get()` request will reject if no other `logged-in` IdPs are available.
```

files/en-us/web/api/identitycredentialrequestoptions/index.md
```diff
@@ -40,19 +40,29 @@ It is used to request an {{domxref("IdentityCredential")}} provided by a {{gloss
...
     - `loginHint` {{optional_inline}}
       - : A string providing a hint about the account option(s) the browser should provide for the user to sign in with. This is useful in cases where the user has already signed in and the site asks them to reauthenticate. Otherwise, the reauthentication process can be confusing when a user has multiple accounts and can't remember which one they used to sign in previously. The value for the `loginHint` property can be taken from the user's previous sign-in, and is matched against the `login_hints` values provided by the IdP in the array of user information returned from the IdP's [accounts list endpoint](/en-US/docs/Web/API/FedCM_API/IDP_integration#the_accounts_list_endpoint).
     - `nonce` {{optional_inline}}
       - : A random string that can be included to ensure the response is issued specifically for this request and prevent {{glossary("replay attack", "replay attacks")}}.
-
-    > [!NOTE]
-    > Currently FedCM only allows the API to be invoked with a single IdP, i.e., the `providers` array has to have a length of 1. Multiple IdPs must be supported via different `get()` calls.
```
ewewraw, line 55: FedCM still only supports a single `get()` (subsequent get requests will fail), but it can support multiple IdPs in the `providers` array. Maybe it's worthy to call it out?

chrisdavidmills: Possibly, but this isn't really the right place to do so. A better place would be the [`get()` page "Exceptions" section](https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/get#exceptions). But then, that raises more questions:

1. What exactly is the nature of the failure? Is it simply that a `get()` call cannot be make when a previous `get()` call promise remains unresolved? Or is it that you can only be signed in using one federated identity provider per domain, and all other galls will fail until the existing sign-in is signed out again? Is it both?
2. I'm assuming the exception type would be an `InvalidStateError`, but I could be wrong.
3. Is this the case for all `get()` call types, or just for FedCM/identity types?

chrisdavidmills: I think I'm going to ask for an editorial review of this PR now, but feel free to chime in more on this issue. Thanks for the review, @ewewraw!

chrisdavidmills:
This one is ready for an editorial review, @hamishwillee and @wbamberg. Whoever fancies it.

hamishwillee:
> account_label / label_hints in the IdP config file / accounts list endpoint. This was added in Chrome 132. Note that this won't have browser compat data - we don't really have a way of recording data for features in endpoints related to an API.

Suggest asking BCD for how to do this. You might need to invent a way.

files/en-us/web/api/identitycredential/disconnect_static/index.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+### Parameters
+
+- `options`
+  - : An options object, which can contain the following properties:
+    - `accountHint`
+      - : A string specifying an account hint that the IdP uses the identify the account to disconnect. The hint can be an arbitrary string as long as the `disconnect_endpoint` can identify the account — for example an email address or user ID. This will not necessarily match the account ID provided by the `accounts_endpoint`.
+    - `clientId`
+      - : A string specifying the RP's client identifier, as specified during sign-in in the `providers` [`clientId`](/en-US/docs/Web/API/IdentityCredentialRequestOptions#clientid) property.
```
hamishwillee: The double -in makes this hard to read.
```suggestion
      - : A string specifying the RP's client identifier, as specified in the `providers` [`clientId`](/en-US/docs/Web/API/IdentityCredentialRequestOptions#clientid) property during sign-in.
```

files/en-us/web/api/identitycredential/disconnect_static/index.md
```diff
@@ -0,0 +1,78 @@
...
+    - `configURL`
+      - : A string specifying the config file URL of the IdP, as specified during sign-in in the `providers` [`configURL`](/en-US/docs/Web/API/IdentityCredentialRequestOptions#configurl) property.
```
hamishwillee:
```suggestion
      - : A string specifying the config file URL of the IdP, as specified in the `providers` [`configURL`](/en-US/docs/Web/API/IdentityCredentialRequestOptions#configurl) property during sign-in.
```

files/en-us/web/api/identitycredential/disconnect_static/index.md
```diff
@@ -0,0 +1,78 @@
...
+  - : Thrown if:
+    - The IdP's `configURL` is invalid or missing the `disconnect_enpoint`.
+    - The document's origin does not match the `configURL`.
+- `NetworkError` {{domxref("DOMException")}}
+  - : Thrown if:
+    - The browser is unable to connect to the IdP.
+    - The request is disallowed by a [`connect-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/connect-src) {{httpheader("Content-Security-Policy")}}.
+    - Another `disconnect()` call was previously made that has not yet resolved.
+    - The FedCM API has been disabled globally.
+    - The IdP's `configURL` is not trustworthy
```
hamishwillee: What is a trustworthy URL? I think we could at last note that this means a secure context. I have run into this before but I can't remember if it means more than that.

files/en-us/web/api/identitycredential/disconnect_static/index.md
```diff
@@ -0,0 +1,78 @@
...
+- `NotAllowedError` {{domxref("DOMException")}}
+  - : Thrown if the embedding `<iframe>` does not have a {{httpheader("Permissions-Policy/identity-credentials-get", "identity-credentials-get")}} [Permissions-Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) set to allow the use of `disconnect()` or if the FedCM API is disabled globally by a policy set on the top-level document.
+
+## Examples
+
```
hamishwillee, line 55: The templates now expect a level 3 heading in Examples to describe what the specific example does.

files/en-us/web/api/identitycredential/disconnect_static/index.md
```diff
@@ -0,0 +1,78 @@
...
+```js-nolint
+IdentityCredential.disconnect(options)
+```
+
+### Parameters
+
+- `options`
+  - : An options object, which can contain the following properties:
+    - `accountHint`
+      - : A string specifying an account hint that the IdP uses the identify the account to disconnect. The hint can be an arbitrary string as long as the `disconnect_endpoint` can identify the account — for example an email address or user ID. This will not necessarily match the account ID provided by the `accounts_endpoint`.
```
hamishwillee: Link `accounts_endpoint` (not mentioned or linked elsewhere on this page)

hamishwillee: The disconnect_endpoint should be linked /Web/API/FedCM_API/IDP_integration#disconnect_endpoint - you do so below but this is first mention.

files/en-us/web/api/fedcm_api/idp_integration/index.md
```diff
@@ -67,8 +70,16 @@ The properties are as follows:
 
 - `accounts_endpoint`
   - : The URL for the accounts list endpoint, which returns a list of accounts that the user is currently signed in to on the IdP. The browser uses these to create a list of sign-in choices to show to the user in the browser-provided FedCM UI.
```
hamishwillee, line 72: On first instance in all these docs please link IdP to https://developer.mozilla.org/en-US/docs/Glossary/Identity_provider

You might have done it in these existing ones but it is omitted from the new docs such as disconnect_static.

files/en-us/web/api/identityprovider/close_static/index.md
```diff
@@ -14,7 +14,7 @@ The **`close()`** static method of the {{domxref("IdentityProvider")}} interface
 
 ## Usage notes
```
hamishwillee: FYI We're slowly purging the use of `## Usage notes` from API ref. This kind of stuff either appears as part of the description OR in a section before examples called `## Description`. I would have said the heading wasn't needed here. Ditto elsewhere. Note, not a requirement for this to be accepted but desirable.

files/en-us/web/api/fedcm_api/index.md
```diff
@@ -38,7 +38,11 @@ There are two parts to using the FedCM API, which are covered in the linked guid
 
 ## Permissions Policy integration and `<iframe>` support
 
-The {{httpheader("Permissions-Policy/identity-credentials-get", "identity-credentials-get")}} [Permissions-Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) can be used to control permission to use FedCM, more specifically usage of the {{domxref("CredentialsContainer.get", "get()")}} method.
+The {{httpheader("Permissions-Policy/identity-credentials-get", "identity-credentials-get")}} [Permissions-Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) can be used to control permission to use FedCM, more specifically related usage of the following methods:
```
hamishwillee: ", more specifically related usage of the" is not quite right. Maybe

```suggestion
The {{httpheader("Permissions-Policy/identity-credentials-get", "identity-credentials-get")}} [Permissions-Policy](/en-US/docs/Web/HTTP/Guides/Permissions_Policy) can be used to control permission to use FedCM.
More specifically, it permits usage of the following methods:
```

files/en-us/web/api/identitycredential/index.md
```diff
@@ -19,11 +19,18 @@ A successful {{domxref("CredentialsContainer.get", "navigator.credentials.get()"
 
 _Inherits properties from its ancestor, {{domxref("Credential")}}._
 
+- {{domxref("IdentityCredential.configURL")}} {{ReadOnlyInline}} {{experimental_inline}}
+  - : A string specifying the config file URL of the IdP used for sign-in.
```
hamishwillee: Even at this level maybe

```suggestion
  - : A string specifying the [config file](/en-US/docs/Web/API/FedCM_API/IDP_integration#provide_a_config_file_and_endpoints) URL of the IdP used for sign-in.
```

files/en-us/web/api/identitycredentialrequestoptions/index.md
```diff
@@ -40,19 +40,29 @@ It is used to request an {{domxref("IdentityCredential")}} provided by a {{gloss
 
     The default value is `"signin"`.
 
+- `mode` {{optional_inline}}
+  - : A string specifying the UI mode to use for the sign-in flow, which browsers can use to specify different behaviors if wished. Possible values are:
```
hamishwillee: - " which browsers can use to specify different behaviors if wished." - This kind of implies that browsers set the mode. What would normally be referred to as a "hint".
- I'm not sure from the spec that browsers have any choice - if they do this would be a hint, and I think it is a commanded behaviour with specific expectation on the browser UIk

files/en-us/web/api/identitycredentialrequestoptions/index.md
```diff
@@ -40,19 +40,29 @@ It is used to request an {{domxref("IdentityCredential")}} provided by a {{gloss
...
+    - `active`
+      - : The sign-in flow must be initiated via a user action such as clicking a button. If `mode` is set to `active`, `providers` can only have a length of `1`, otherwise the `get()` promise will reject.
+    - `passive`
+      - : The sign-in flow can be initiated without direct user interaction. This is the default value.
+
 - `providers`
-  - : An array containing a single object specifying details of an IdP to be used to sign in. This object can contain the following properties:
+  - : An array of objects specifying details of the IdPs the user should be presented with as options for signing in. These objects can contain the following properties:
```
hamishwillee:
```suggestion
  - : An array of objects specifying details of the IdPs that the user should be presented with as options for signing in. These objects can contain the following properties:
```

files/en-us/web/api/identitycredentialrequestoptions/index.md
```diff
@@ -40,19 +40,29 @@ It is used to request an {{domxref("IdentityCredential")}} provided by a {{gloss
...
     - `configURL`
       - : A string specifying the URL of the IdP's config file. See [Provide a config file](/en-US/docs/Web/API/FedCM_API/IDP_integration#provide_a_config_file_and_endpoints) for more information.
     - `clientId`
       - : A string specifying the RP's client identifier. This information is issued by the IdP to the RP in a separate process that is specific to the IdP.
+    - `domainHint` {{optional_inline}}
+      - : A string representing a hint corresponding to a domain the RP is interested in. If provided, the user agent will only show accounts that match the domain hint value in their [`domain_hints`](/en-US/docs/Web/API/FedCM_API/IDP_integration#domain_hints) array. If `"any"` is specified, the RP will show any account that is associated with at least one domain hint.
```
hamishwillee: The first sentence is very hard to parse. The problem is most likely the awkward "representing" syntax (yes, I know it is much used). Also, is it really a hint? It's a hint if the browser can choose to use it or not.

```suggestion
      - : A string hinting at the domain of accounts that the RP is interested in. If provided, the user agent will only show accounts that match the domain hint value in their [`domain_hints`](/en-US/docs/Web/API/FedCM_API/IDP_integration#domain_hints) array. If `"any"` is specified, the RP will show any account that is associated with at least one domain hint.
```

files/en-us/web/api/identitycredentialrequestoptions/index.md
hamishwillee: FWIW I agree with you here Chris (unless I misunderstand the point). The existing note is incorrect because you can have multiple providers. More information needed on how `get()` fails, but likely and exception on get, and not something to mention in this particular interface.

files/en-us/web/api/identitycredentialrequestoptions/index.md
```diff
@@ -40,19 +40,29 @@ It is used to request an {{domxref("IdentityCredential")}} provided by a {{gloss
 
     The default value is `"signin"`.
 
+- `mode` {{optional_inline}}
```
hamishwillee, line 43: There is a line in the introduction

> It is used to request an {{domxref("IdentityCredential")}} provided by a {{glossary("identity provider", "federated identity provider")}} that supports the [Federated Credential Management (FedCM) API](/en-US/docs/Web/API/FedCM_API).

This is ultimately true, but it is sufficient/comprehensive? i.e. we request a list of providers be offered to a user and get back a single credential from the selected IdP.

files/en-us/web/api/identitycredential/configurl/index.md
```diff
@@ -0,0 +1,57 @@
...
+The **`configURL`** read-only property of the {{domxref("IdentityCredential")}} interface returns a string specifying the config file URL of the IdP used for sign-in.
+
+See [Provide a config file](/en-US/docs/Web/API/FedCM_API/IDP_integration#provide_a_config_file_and_endpoints) for more information.
+
+## Value
+
+A string.
+
+## Examples
+
```
hamishwillee, line 22: As I said somewhere else, it is now usual to have a level three heading describing the precise example shown, even if only one example is provided.

files/en-us/web/api/identitycredential/configurl/index.md
```diff
@@ -0,0 +1,57 @@
...
+Relying parties (RPs) can call `navigator.credentials.get()` with the `identity` option to make a request for users to sign in to the RP via an identity provider (IdP), using identity federation. A typical request would look like this:
```
hamishwillee: Maybe, just because from now one we're going to get typical requests with multiple providers.
```suggestion
Relying parties (RPs) can call `navigator.credentials.get()` with the `identity` option to make a request for users to sign in to the RP via an identity provider (IdP), using identity federation. A request indicating a single provider would look like this:
```

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
```diff
@@ -10,16 +10,16 @@ This article describes the process by which a relying party (RP) can use the [Fe
 
 ## Calling the get() method
 
-RPs can call {{domxref("CredentialsContainer.get", "navigator.credentials.get()")}} with an `identity` option to request that a user signs in to the RP with an existing IdP account that they are already signed in to on the browser. The IdP identifies the RP by its `clientId`, which was issued by the IdP to the RP in a separate process that is specific to the IdP. The IdP identifies the specific user using credentials (cookies) provided to the browser on login.
+RPs can call {{domxref("CredentialsContainer.get", "navigator.credentials.get()")}} with an `identity` option to request that a user is given the option to sign in to the RP with an a choice of existing IdP accounts that they are already signed in to on the browser. The IdPs identify the RP by its `clientId`, which was issued by each IdP to the RP in a separate IdP-specific process. The chosen IdP identifies the specific user using credentials (cookies) provided to the browser on login.
```
hamishwillee: Typo "an a"

```suggestion
RPs can call {{domxref("CredentialsContainer.get", "navigator.credentials.get()")}} with an `identity` option to request that a user be given the option to sign in to the RP with a choice of existing IdP accounts (that they are already signed in to on the browser). The IdPs identify the RP by its `clientId`, which was issued by each IdP to the RP in a separate IdP-specific process. The chosen IdP identifies the specific user using credentials (cookies) provided to the browser on login.
```

hamishwillee: > The chosen IdP identifies the specific user using credentials (cookies) provided to the browser on login.

Edited. This feels not quite right. I think it is, but perhaps link down to "FedCM sign-in flow"

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
```diff
@@ -135,7 +135,7 @@ Auto-reauthentication can occur if `mediation` is set to `optional` or `silent`.
 With these `mediation` options, auto-reauthentication will occur under the following conditions:
 
 - FedCM is available to use. For example, the user has not disabled FedCM either globally or in the RP's settings.
-- The user has only used one account to sign into the RP website on this browser via FedCM.
+- The user has only used one account to sign into the RP website on this browser via FedCM. If returning accounts exist for multiple IdPs, the user won't be automatically re-authenticated.
```
hamishwillee: What is a returning account? I think you're saying that the RP has account on multiple IdPs - i.e. reinforcing the previous sentence.

```suggestion
- The user has only used one account to sign into the RP website on this browser via FedCM. If accounts exist for multiple IdPs, the user won't be automatically re-authenticated.
```

files/en-us/web/api/identitycredentialrequestoptions/index.md
```diff
@@ -40,19 +40,29 @@ It is used to request an {{domxref("IdentityCredential")}} provided by a {{gloss
...
+  - : A string specifying the UI mode to use for the sign-in flow, which browsers can use to specify different behaviors if wished. Possible values are:
+    - `active`
+      - : The sign-in flow must be initiated via a user action such as clicking a button. If `mode` is set to `active`, `providers` can only have a length of `1`, otherwise the `get()` promise will reject.
+    - `passive`
+      - : The sign-in flow can be initiated without direct user interaction. This is the default value.
```
hamishwillee, line 48: We need to say more about both of these - the only other mention is in https://pr40411.review.mdn.allizom.net/en-US/docs/Web/API/FedCM_API/IDP_integration#supports_use_other_account ... and I am not certain how it works.
My thinking is that probably a whole section on the differences in either that doc - but mostly likely in https://pr40411.review.mdn.allizom.net/en-US/docs/Web/API/FedCM_API/RP_sign-in

At the moment this says that active is kicked off by a button and that only one provider is allowed. That implies this is for the case when you know only one provider can be used. But what is the overarching situation that would lead to this, and to needing this particular flow vs passive. There is also talk somewhere of using a model dialog and also of the passive case allowing login without user interaction - presumably auto-reauthentication.

Upshot, this is not enough for anyone to understand the difference between the flow options - what the user will see, what the site should do, what the integration looks like.

chrisdavidmills:
> > account_label / label_hints in the IdP config file / accounts list endpoint. This was added in Chrome 132. Note that this won't have browser compat data - we don't really have a way of recording data for features in endpoints related to an API.
> 
> Suggest asking BCD for how to do this. You might need to invent a way.

I don't see it as that useful. Most of the features you can add to the IdP endpoint have some kind of corresponding feature in the actual RP developer-visible API surface, for example:

- `supports_use_other_account` in the IdP config file has `mode: "passive"` in the `get()` call `identity` object.
- `domain_hints` in the accounts list endpoint has `domainHint` in the `get()` call `identity` object.

This is somewhat of an exception.

files/en-us/web/api/identitycredential/disconnect_static/index.md
chrisdavidmills: It's not just that: see https://w3c-fedid.github.io/FedCM/#browser-api-identity-credential-disconnect, and see https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin.

I wasn't sure what to put, without opening a confusing can of worms.

files/en-us/web/api/identitycredential/disconnect_static/index.md
chrisdavidmills: Fair point. I've added a descriptive H3 for all examples in the `IdentityCredential` and `IdentityProvider` page trees.

files/en-us/web/api/identitycredential/disconnect_static/index.md
chrisdavidmills: I've linked both here.

files/en-us/web/api/fedcm_api/idp_integration/index.md
chrisdavidmills: I have gone through and made sure that all related pages have a decent first link to the identity provider and relying party glossary entries. I have also updated the relying party glossary entry to make the relationship between IdP and RP clearer on that page.

files/en-us/web/api/identityprovider/close_static/index.md
chrisdavidmills: Good call. I've removed the heading from the `close()` page, and moved the section above "Examples" and renamed it "Description" on the `getUserInfo()` page, as it is a bit more substantial.

files/en-us/web/api/identitycredentialrequestoptions/index.md
chrisdavidmills: I think you're right. I've updated it to make it sound less like a hint and more like a definite.

files/en-us/web/api/identitycredentialrequestoptions/index.md
chrisdavidmills: I've rewritten the intro like so:

> The **`IdentityCredentialRequestOptions`** dictionary represents the object passed to {{domxref("CredentialsContainer.get()")}} as the value of the `identity` option.
>
> When an `identity` option is provided in a `get()` call made on a {{glossary("Relying party", "relying party")}} (RP) website, the user is offered a list of {{glossary("identity provider", "federated identity providers")}} (IdPs) as sign-in options. Once the user signs in successfully using one of these options, the promise returned by the `get()` call returns an {{domxref("IdentityCredential")}} object.

WDYT?

files/en-us/web/api/identitycredential/configurl/index.md
chrisdavidmills: updated (see earlier comment).

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
chrisdavidmills: I've updated the final sentence to 

> The chosen IdP identifies the specific user using credentials (cookies) provided to the browser during the [sign-in flow](#fedcm_sign-in_flow).

Does that work for you?

files/en-us/web/api/identitycredentialrequestoptions/index.md
chrisdavidmills: OK, I've provided a section about the two modes in the RP sign-in article, and linked to it from appropriate places. Let me know what you think.

hamishwillee:
> I don't see it as that useful. Most of the features you can add to the IdP endpoint have some kind of corresponding feature in the actual RP developer-visible API surface, for example:
> 
> * `supports_use_other_account` in the IdP config file has `mode: "passive"` in the `get()` call `identity` object.
> * `domain_hints` in the accounts list endpoint has `domainHint` in the `get()` call `identity` object.
> 
> This is somewhat of an exception.

As long as people can work out the support story then fine.

files/en-us/web/api/identitycredential/disconnect_static/index.md
hamishwillee: OK, so I recall this now. It is basically wriggle room added to secure context - things that are strictly speaking not secure because they don't use https but in practice likely to be - such as file access, localhost, and other things that the browser decides it will trust.

We have two other mentions of this in the docs so it is worth looking at. Proposal, we change this to `trustworthy or [potentially trustworthy](/en-US/docs/Web/Security/Secure_Contexts#potentially_trustworthy)`", and create a section for that link. It doesn't have to be a huge can of worms - it really is just "in some cases sites that are not strictly speaking secure are allowed to be trusted blah blah blah"

files/en-us/web/api/fedcm_api/index.md
```diff
@@ -9,13 +9,13 @@ browser-compat: api.IdentityCredential
 
 {{SeeCompatTable}}{{DefaultAPISidebar("FedCM API")}}
 
-The **Federated Credential Management API** (or _FedCM API_) provides a standard mechanism for identity providers (IdPs) to make identity federation services available on the web in a privacy-preserving way, without the need for [third-party cookies](/en-US/docs/Web/Privacy/Guides/Third-party_cookies) and redirects. This includes a JavaScript API that enables the use of federated authentication for activities such as signing in or signing up on a website.
+The **Federated Credential Management API** (or _FedCM API_) provides a standard mechanism for {{glossary("Identity provider", "identity providers")}} (IdPs) to make identity federation services available on the web in a privacy-preserving way, without the need for [third-party cookies](/en-US/docs/Web/Privacy/Guides/Third-party_cookies) and redirects. This includes a JavaScript API that enables the use of federated authentication for activities such as signing in or signing up on a website.
```
hamishwillee, line 12: FWIW There is nothing wrong with this, but it did make me wonder what "federated" means. I.e. this credential management service is differentiated by the fact that it is "federated" but I don't understand why this term is used. Why don't we call it Cheryl Identity Providers?

In the next section we do explain it is the delegation of identity validation to another site. Perhaps the origin of the name is not relevant. I just like to know.

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
```diff
@@ -103,6 +103,32 @@ The flow is as follows:
...
+          configURL: "https://accounts.idp.example/config.json",
+          clientId: "********",
+        },
+      ],
+    },
+  });
+}
+```
+
+The default value for `mode` is `passive`. If `mode` is not set, or is set explicitly to `passive`, the browser can initiate the sign-in flow via a `get()` call without direct user interaction. In this mode, browsers typically present the user with a sign-in dialog window containing all the different sign-in options specified in the `providers` object, and they can choose whichever one suits them best and then enter the appropriate credentials.
```
hamishwillee: I like this new section. 

Just FMI, do we need to say why/when/how the browser would want to initiate sign in without direct user interaction? 
I kind of think yes, because I don't know. I'm guessing that this might be because you want the site to auto-log the user in when they navigate to the sign in page?

files/en-us/web/api/identitycredentialrequestoptions/index.md
hamishwillee: Looks excellent.

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
```diff
@@ -6,17 +6,17 @@ page-type: guide
...
 
-## Calling the get() method
+## Calling the `get()` method
 
-RPs can call {{domxref("CredentialsContainer.get", "navigator.credentials.get()")}} with an `identity` option to request that a user is given the option to sign in to the RP with an a choice of existing IdP accounts that they are already signed in to on the browser. The IdPs identify the RP by its `clientId`, which was issued by each IdP to the RP in a separate IdP-specific process. The chosen IdP identifies the specific user using credentials (cookies) provided to the browser on login.
+RPs can call {{domxref("CredentialsContainer.get", "navigator.credentials.get()")}} with an `identity` option to request that a user be given the option to sign in to the RP with a choice of existing IdP accounts (that they are already signed in to on the browser). The IdPs identify the RP by its `clientId`, which was issued by each IdP to the RP in a separate IdP-specific process. The chosen IdP identifies the specific user using credentials (cookies) provided to the browser during the [sign-in flow](#fedcm_sign-in_flow).
```
hamishwillee: The approach is is good but I can't cope with "specific user using credentials"
```suggestion
RPs can call {{domxref("CredentialsContainer.get", "navigator.credentials.get()")}} with an `identity` option to request that a user be given the option to sign in to the RP with a choice of existing IdP accounts (that they are already signed in to on the browser). The IdPs identify the RP by its `clientId`, which was issued by each IdP to the RP in a separate IdP-specific process. The chosen IdP identifies the specific user who is attempting to sign-in with the credentials (cookies) provided to the browser during the [sign-in flow](#fedcm_sign-in_flow).
```

hamishwillee (commented):
Thanks for the comprehensive changes in response. Looks great. A few nits. I'd quite like to resolve https://github.com/mdn/content/pull/40411#discussion_r2220540282 before merging.

files/en-us/web/api/fedcm_api/index.md
chrisdavidmills: I don't really get this comment. Federated identity services are a common, well-known type of identity management service whereby a third-party service is used to handle the auth.

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
chrisdavidmills: Good call. I have added some text to explain this. Your example is basically exactly what I'd think of for an example of when to use this.

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
chrisdavidmills: Just trying to keep you on your toes, y'know? ;-)

chrisdavidmills: Updated.

files/en-us/web/api/identitycredential/disconnect_static/index.md
chrisdavidmills: OK, this isn't quite as bad as I first thought. I have added a section and linked to it, as suggested.

hamishwillee: What you have done looks great.

hamishwillee (approved):
Thanks very much @chrisdavidmills - looks excellent. Approving.

chrisdavidmills:
> Thanks very much @chrisdavidmills - looks excellent. Approving.

Super, thanks for the awesome review, @hamishwillee!

PR #40418, opened 2025-07-18
Make CSS attr types definition list items

hamishwillee:
This makes the options of https://developer.mozilla.org/en-US/docs/Web/CSS/attr#attr-type into definition list items. 
This allows them to be linked, including from BCD, and is IMO always good practice

Related to discussion in https://github.com/mdn/browser-compat-data/pull/26624#pullrequestreview-3031312598


dipikabh (approved):
LGTM, thanks a lot!

PR #40420, opened 2025-07-18
TrustedTypes: Add boilerplate for the xxxHtmlUnsafe() methods

hamishwillee:
This updates the following methods to explain how they are used with TrustedTypes:

- [x] [`Document.parseHTMLUnsafe()` ](https://developer.mozilla.org/en-US/docs/Web/API/Document/parseHTMLUnsafe_static)
- [x] [`Element.setHTMLUnsafe()` ](https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTMLUnsafe)
- [x] [`ShadowRoot.setHTMLUnsafe()` ](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/setHTMLUnsafe)

The advice from https://github.com/WICG/sanitizer-api/issues/303 was that the recommendation is to the safe methods where possible, once it is defined. The setUnsafeHTML should be treated the same way as innerHTML w.r.t. trusted types. I believe this is because the point of TT is to provide easily auditable policy files that can then be enforced.

There was no specific comment on whether you should use a noop policy to avoid double sanitization: but the inference was that if that is allowed by your website policies then you can. However it isn't particularly safe. 

I have updated all three to have the new header and a security considerations section that says "use the safe versions, if supported. If you need to use this you can, but use trusted types "first" - up to you if you then use the sanitizer.

The two `setHTMLUnsafe()` methods have examples that show basic usage that includes TT in line with the guidance. I have explicitly omitted the use of TT in the live examples to keep them simple, but added a note disclaimer.
There was no example for the parseHTMLUnsafe() and I have not added on.

### DIFF
--- a/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
+++ b/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
@@ -15,7 +15,7 @@ browser-compat: api.Document.parseHTMLUnsafe_static
 > For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
 > This means you can be sure that the input has been passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
 
-The **`parseHTMLUnsafe()`** static method of the {{domxref("Document")}} object is used to parse an HTML input, optionally filtering unwanted HTML elements and attributes, in order to create a new {{domxref("Document")}} instance.
+The **`parseHTMLUnsafe()`** static method of the {{domxref("Document")}} object is used to parse HTML input, optionally filtering unwanted HTML elements and attributes, in order to create a new {{domxref("Document")}} instance.
 
 Unlike with {{domxref("Document.parseHTML_static", "Document.parseHTML()")}}, XSS-unsafe HTML entities are not guaranteed to be removed.
 
--- a/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
+++ b/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
@@ -12,8 +12,8 @@ browser-compat: api.Document.parseHTMLUnsafe_static
 > This method parses its input as HTML, writing the result into the DOM.
 > APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
 >
-> For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
-> This means you can be sure that the input has been passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
+> You can mitigate this risk by always passing `TrustedHTML` objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](#security_considerations) for more information.
 
 The **`parseHTMLUnsafe()`** static method of the {{domxref("Document")}} object is used to parse HTML input, optionally filtering unwanted HTML elements and attributes, in order to create a new {{domxref("Document")}} instance.
 
@@ -56,18 +56,25 @@ A {{domxref("Document")}}.
 The **`parseHTMLUnsafe()`** static method can be used to create a new {{domxref("Document")}} instance, optionally filter out unwanted elements and attributes.
 The resulting `Document` will have a [content type](/en-US/docs/Web/API/Document/contentType) of "text/html", a [character set](/en-US/docs/Web/API/Document/characterSet) of UTF-8, and a URL of "about:blank".
 
-The suffix "Unsafe" in the method name indicates that, while the method does allow the input string to be filtered of unwanted HTML entities, it does not enforce the sanitization or removal of potentially unsafe XSS-relevant input.
-If no sanitizer configuration is specified in the `options.sanitizer` parameter, `parseHTMLUnsafe()` is used without any sanitization.
-Note that {{htmlelement("script")}} elements are not evaluated during parsing.
-
 The input HTML may include [declarative shadow roots](/en-US/docs/Web/HTML/Reference/Elements/template#declarative_shadow_dom).
 If the string of HTML defines more than one [declarative shadow root](/en-US/docs/Web/HTML/Reference/Elements/template#declarative_shadow_dom) in a particular shadow host then only the first {{domxref("ShadowRoot")}} is created — subsequent declarations are parsed as {{htmlelement("template")}} elements within that shadow root.
 
-`parseHTMLUnsafe()` should be instead of {{domxref("Document.parseHTML_static", "Document.parseHTML()")}} when parsing potentially unsafe strings of HTML that for whatever reason need to contain XSS-unsafe elements or attributes.
-If the HTML to be parsed doesn't need to contain unsafe HTML entities, then you should use `Document.parseHTML()`.
+{{domxref("Document/parseHTML_static", "Document.parseHTML()")}} should almost always be used instead of this method — on browsers where it is supported — as it always removes XSS-unsafe HTML entities.
+`parseHTMLUnsafe()` may be useful in the rare cases where you _need_ to allow for at least some XSS-unsafe elements or attributes in the input.
+In this case the unsafe elements that aren't needed can still be filtered, which may reduce the overall risk.
 
-Note that since this method does not necessarily sanitize input strings of XSS-unsafe entities, input strings should also be validated using the [Trusted Types API](/en-US/docs/Web/API/Trusted_Types_API).
-If the method is used with both a trusted types and a sanitizer, the HTML input will be passed through the trusted type transformation function before it is sanitized.
+### Security considerations
+
+The suffix "Unsafe" in the method name indicates that while the method allows the input string to be filtered of unwanted and XSS-unsafe HTML entities, unlike {{domxref("Document/parseHTML_static", "Document.parseHTML()")}} it does not enforce this sanitization.
+In fact, by default no sanitizer is used!
+The method is therefore a possible vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, where potentially unsafe strings provided by a user are injected into the DOM without first being sanitized.
+
+You should mitigate this risk by always passing {{domxref("TrustedHTML")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup (such as {{htmlelement("script")}} elements and event handler attributes), before it is injected.
+
+Note that this may lead to "double-sanitization" of the input — once in the transformation function, and again in the `parseHTMLUnsafe()` method.
+The recommendation is to use trusted types to sanitize according to your website policies, because trusted types allow auditing of potential injection sinks.
+If you wish to sanitize the input again by passing a sanitizer to this method, that's up to you!
 
 ## Specifications
 
--- a/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
+++ b/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
@@ -35,6 +35,7 @@ Document.parseHTMLUnsafe(input, options)
   - : An options object with the following optional parameters:
     - `sanitizer` {{optional_inline}}
       - : A {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} object which defines what elements of the input will be allowed or removed.
+        This can also be a string with the value `"default"`, which applies a `Sanitizer` with the default (XSS-safe) configuration.
         Note that generally a `"Sanitizer` is expected than the to be more efficient than a `SanitizerConfig` if the configuration is to reused.
         If not specified, no sanitizer is used.
 
--- a/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
+++ b/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
@@ -9,7 +9,7 @@ browser-compat: api.Document.parseHTMLUnsafe_static
 {{APIRef("DOM")}}
 
 > [!WARNING]
-> This API parses its input as HTML, writing the result into the DOM.
+> This method parses its input as HTML, writing the result into the DOM.
 > APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
 >
 > For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
--- a/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
+++ b/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
@@ -36,8 +36,9 @@ Document.parseHTMLUnsafe(input, options)
     - `sanitizer` {{optional_inline}}
       - : A {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} object which defines what elements of the input will be allowed or removed.
         This can also be a string with the value `"default"`, which applies a `Sanitizer` with the default (XSS-safe) configuration.
-        Note that generally a `"Sanitizer` is expected than the to be more efficient than a `SanitizerConfig` if the configuration is to reused.
         If not specified, no sanitizer is used.
+        
+        Note that generally a `Sanitizer` is expected than the to be more efficient than a `SanitizerConfig` if the configuration is to reused.
 
 ### Return value
 
@@ -74,7 +75,7 @@ You should mitigate this risk by always passing {{domxref("TrustedHTML")}} objec
 This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup (such as {{htmlelement("script")}} elements and event handler attributes), before it is injected.
 
 Using `TrustedHTML` makes it possible to audit and check that sanitization code is effective in just a few places, rather than scattered across all your injection sinks.
-It should be unnecessary to additionally pass a sanitizer to the method when using `TrustedHTML`.
+You should not need to pass a sanitizer to the method when using `TrustedHTML`.
 
 If for any reason you can't use `TrustedHTML` (or even better, `setHTML()`) then the next safest option is to use `setHTMLUnsafe()` with the XSS-safe default {{domxref("Sanitizer")}}.
 
--- a/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
+++ b/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
@@ -8,6 +8,13 @@ browser-compat: api.Document.parseHTMLUnsafe_static
 
 {{APIRef("DOM")}}
 
+> [!WARNING]
+> This API parses its input as HTML, writing the result into the DOM.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
+>
+> For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+> This means you can be sure that the input has been passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
+
 The **`parseHTMLUnsafe()`** static method of the {{domxref("Document")}} object is used to parse an HTML input, optionally filtering unwanted HTML elements and attributes, in order to create a new {{domxref("Document")}} instance.
 
 Unlike with {{domxref("Document.parseHTML_static", "Document.parseHTML()")}}, XSS-unsafe HTML entities are not guaranteed to be removed.
@@ -22,7 +29,7 @@ Document.parseHTMLUnsafe(input, options)
 ### Parameters
 
 - `input`
-  - : A string or {{domxref("TrustedHTML")}} instance defining HTML to be parsed.
+  - : A {{domxref("TrustedHTML")}} or string instance defining HTML to be parsed.
 - `options` {{optional_inline}}
   - : An options object with the following optional parameters:
     - `sanitizer` {{optional_inline}}
--- a/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
+++ b/files/en-us/web/api/document/parsehtmlunsafe_static/index.md
@@ -15,9 +15,10 @@ browser-compat: api.Document.parseHTMLUnsafe_static
 > You can mitigate this risk by always passing `TrustedHTML` objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
 > See [Security considerations](#security_considerations) for more information.
 
-The **`parseHTMLUnsafe()`** static method of the {{domxref("Document")}} object is used to parse HTML input, optionally filtering unwanted HTML elements and attributes, in order to create a new {{domxref("Document")}} instance.
+> [!NOTE]
+> {{domxref("Document/parseHTML_static", "Document.parseHTML()")}} should almost always be used instead of this method — on browsers where it is supported — as it always removes XSS-unsafe HTML entities.
 
-Unlike with {{domxref("Document.parseHTML_static", "Document.parseHTML()")}}, XSS-unsafe HTML entities are not guaranteed to be removed.
+The **`parseHTMLUnsafe()`** static method of the {{domxref("Document")}} object is used to parse HTML input, optionally filtering unwanted HTML elements and attributes, in order to create a new {{domxref("Document")}} instance.
 
 ## Syntax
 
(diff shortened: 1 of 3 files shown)


files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -8,6 +8,13 @@ browser-compat: api.ShadowRoot.setHTMLUnsafe
 
 {{APIRef("Shadow DOM")}}
 
+> [!WARNING]
+> This API parses its input as HTML, writing the result into the DOM.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
+>
+> For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+> This means you can be sure that the input has been passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
```
hamishwillee: I was thinking maybe for (all) these cases we modify the boilerplate like this.

```suggestion
> [!WARNING]
> This API parses its input as HTML, writing the result into the DOM.
> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
> It is better to use XSS-safe methods where possible, such as {{domxref("ShadowRoot.setHTML()")}}.
>
> You should ensure that this method is called with as restrictive a sanitizer as possible.
> If [Trusted Types are enforced](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types),  you will need pass {{domxref("TrustedHTML")}} objects into this method.
> The policy for creating these types should be a pass-through, as the inputs do not need to be sanitized twice.
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -22,12 +29,12 @@ setHTMLUnsafe(input, options)
 ### Parameters
 
 - `input`
-  - : A string or {{domxref("TrustedHTML")}} instance defining HTML to be parsed.
+  - : A {{domxref("TrustedHTML")}} or string instance defining HTML to be parsed.
```
hamishwillee, line 33: My intent is that for all injection sinks we will put the trusted type as the first option.

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -227,6 +235,31 @@ With this approach you can create safe HTML, but you aren't forced to.
 
 {{EmbedLiveSample("setHTMLUnsafe() live example","100","350px")}}
 
+<!--
```
hamishwillee: This is a (hidden) example of what we might show for the htmlunsafe examples.

The point being that you need to do this to avoid the methods throwing exceptions. 
What we need to show here is a very restrictive sanitiser that still allows something that is XSS unsafe. 
- Obvious thing is a script element, but is there anything else that might be better for an example? It is "the most unsafe" thing, so seems unsubtle to allow it.
  - Will modify to use default sanitizer then add an allowed "bad" item, so that we're demonstrating "disallow what you can".

Will also note that if you're using the default sanitizer there is no reason not to use the safe method setHTML.

Note, this is a "call for help". Is it reasonable? Should we be recommending something else? Who can we ask?

files/en-us/web/api/document/parsehtmlunsafe_static/index.md
```diff
@@ -8,6 +8,13 @@ browser-compat: api.Document.parseHTMLUnsafe_static
 
 {{APIRef("DOM")}}
 
+> [!WARNING]
+> This API parses its input as HTML, writing the result into the DOM.
```
hamishwillee:
```suggestion
> This method parses its input as HTML, writing the result into the DOM.
```

files/en-us/web/api/document/parsehtmlunsafe_static/index.md
```diff
@@ -8,6 +8,13 @@ browser-compat: api.Document.parseHTMLUnsafe_static
...
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
+>
+> For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+> This means you can be sure that the input has been passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
```
hamishwillee:
```suggestion
> For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
> You can reduce the risk by passing {{domxref("TrustedHTML")}} objects into this method instead of string, and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
> This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
```

lukewarlow:
So one thing to keep in mind is that right now no browser ships the sanitizer and even once they do on older browser releases all these functions will very much be XSS sinks. So for that reason I think MDN should be consistent with it's guidance, if the approach is to recommend using trusted types then it should be done here.

Regarding the case of once you have the sanitizer API and you use it, it does get interesting.

While the sanitizer API is better designed (avoids mXSS issues) it isn't *enforcing* anything in particular (of course on the safe variants it will). Trusted types allows you to enforce that it goes through an approved policy, which you control the substance of.

This is a tricky case where you don't necessarily want to provide a no-op policy because your sanitizer config might be wrong or too lax for your liking. But equally you probably don't want to run it through two sanitizers.

I think the general advice still applies though. They should be passing a TrustedHTML argument. It's very context dependent as to what the policy should look like though.

I'll raise an issue on the sanitizer API repo to discuss guidance on this point.

hamishwillee:
@lukewarlow You make some good points. I've been imagining the release in browsers of sanitizer as imminent, but of course, there is no guarantee that this is true. 

> I'll raise an issue on the sanitizer API repo to discuss guidance on this point.

Thank you.

hamishwillee:
@wbamberg @lukewarlow This is now ready for review. I have integrated the unsatisfying advice from the spec authors as best I can, as covered in the description above https://github.com/mdn/content/pull/40420#issue-3241656902

Note, if you can look by Friday that would be great. I'm gone for 3 ish weeks after Tuesday, so that would allow some iteration.  Fine of course if you can't.

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -22,12 +29,12 @@ setHTMLUnsafe(input, options)
...
 
 - `input`
-  - : A string or {{domxref("TrustedHTML")}} instance defining HTML to be parsed.
+  - : A {{domxref("TrustedHTML")}} instance or string defining HTML to be parsed.
 - `options` {{optional_inline}}
   - : An options object with the following optional parameters:
     - `sanitizer` {{optional_inline}}
       - : A {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} object which defines what elements of the input will be allowed or removed.
-        Note that generally a `"Sanitizer` is expected than the to be more efficient than a `SanitizerConfig` if the configuration is to reused.
+        Note that generally a `"Sanitizer` is expected to be more efficient than a `SanitizerConfig` if the configuration is to reused.
```
wbamberg:
```suggestion
        Note that generally a `Sanitizer` is expected to be more efficient than a `SanitizerConfig` if the configuration is to reused.
```

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -22,12 +29,12 @@ setHTMLUnsafe(input, options)
 ### Parameters
 
 - `input`
-  - : A string or {{domxref("TrustedHTML")}} instance defining HTML to be parsed.
+  - : A {{domxref("TrustedHTML")}} instance or string defining HTML to be parsed.
 - `options` {{optional_inline}}
   - : An options object with the following optional parameters:
     - `sanitizer` {{optional_inline}}
       - : A {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} object which defines what elements of the input will be allowed or removed.
```
wbamberg: Under "Exceptions" it's implied that it can be a string with the value "default".

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -46,46 +53,87 @@ None (`undefined`).
...
 
-Like `setHTML()`, `setHTMLUnsafe()` may be used instead of {{domxref("Element.innerHTML")}} in order to parse strings of HTML that may contain declarative shadow roots.
-`setHTMLUnsafe()` should be instead of {{domxref("Element.setHTML()")}} when parsing potentially unsafe strings of HTML that for whatever reason need to contain XSS-unsafe elements or attributes.
-If strings to be injected don't need to contain unsafe HTML entities, then you should always use {{domxref("Element.setHTML()")}}.
+{{domxref("Element.setHTML()")}} should almost always be used instead of this method — on browsers where it is supported — as it always removes XSS-unsafe HTML entities.
+`setHTMLUnsafe()` may be useful in the rare cases where you _need_ to allow for at least some XSS-unsafe elements or attributes in the input.
+In this case the unsafe elements that aren't needed can still be filtered, which may reduce the overall risk.
```
wbamberg: I wonder if it's worth an example to illustrate. For example, setHTML() will remove all event handler attributes, while omitting the sanitizer argument or using `innerHTML` will allow all event handler attributes. Calling `setHTMLUnsafe()` with an appropriate sanitizer configuration enables you to allow only some event handler attributes: while this is still unsafe, it is safer than allowing all of them:

```js
const target = document.querySelector("#target");

const config = {
    elements: ["img"],
    attributes: ["src", "onclick"]
}

const input = "<img src=x onclick=alert('onclick') onerror=alert('onerror')>";

target.setHTML(input, {sanitizer: config});
// will remove both event handler attributes

target.setHTMLUnsafe(input);
target.innerHTML = input;
// will remove neither event handler attribute

target.setHTMLUnsafe(input, {sanitizer: config});
// will remove only `onerror`
```

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -46,46 +53,87 @@ None (`undefined`).
...
+
+You should mitigate this risk by always passing {{domxref("TrustedHTML")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup (such as {{htmlelement("script")}} elements and event handler attributes), before it is injected.
+
+Note that this may lead to "double-sanitization" of the input — once in the transformation function, and again in the `setHTMLUnsafe()` method.
+The recommendation is to use trusted types to sanitize according to your website policies, because trusted types allow auditing of potential injection sinks.
+If you wish to sanitize the input again by passing a sanitizer to this method, that's up to you!
```
wbamberg: I can't see why you should (in terms of "security guidance") ever use this method with a sanitizer unless for some reason you can't enforce trusted types. Because it seems preferable to use and enforce trusted types, because you know you're not messing up at the call sites. And if you are using and enforcing trusted types, then it's surely redundant to sanitize the trusted type.

And if you're not using it with a sanitizer I don't really see how it is different from `innerHTML`.

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -46,46 +53,87 @@ None (`undefined`).
...
-// setHTML() with no sanitizer
-target.setHTMLUnsafe(unsanitizedString);
+// setHTMLUnsafe() with no sanitizer
+target.setHTMLUnsafe(trustedHTML);
 
 // Define custom Sanitizer and use in setHTMLUnsafe()
 // This allows only elements: div, p, button, script
 const sanitizer1 = new Sanitizer({
   elements: ["div", "p", "button", "script"],
 });
```
wbamberg, line 185: Except why would you do this, because if your TT policy put it through DOMPurify, any script elements would have been removed, wouldn't they? So why use a sanitizer config that allows them?

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -46,46 +53,87 @@ None (`undefined`).
...
+// Create a TrustedHTML instance using the policy
+const trustedHTML = policy.createHTML(untrustedString);
+```
+
+#### Using setHTMLUnsafe() with and without a sanitizer
+
+Now that we have `trustedHTML`, the code below shows how you can use it with `setHTMLUnsafe()`, both with and without a sanitizer.
+
+> [!NOTE]
+> The input may already sanitized by the trusted type policy at this point.
```
wbamberg:
```suggestion
> The input may already have been sanitized by the trusted type policy at this point.
```

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -96,6 +144,10 @@ This example provides a "live" demonstration of the method when called with diff
 The code defines buttons that you can click to inject a string of HTML that is not sanitized, and that uses and a custom sanitizer, respectively.
```
wbamberg:
```suggestion
The code defines buttons that you can click to inject a string of HTML. One button injects the HTML without sanitizing it at all, and the second uses a custom sanitizer than allows `<script>` elements but not other unsafe items.
```

wbamberg (commented):
Comments only on `element.setHTMLUnsafe()` but I expect the others are going to be similar.

files/en-us/web/api/element/sethtmlunsafe/index.md
hamishwillee: @wbamberg I expect APIs to "make sense" and to be able to explain how they should be used, but the cases in which this "might" make sense are very narrow. So you're right - there is no reason to do this, but it is how the sanitizer part might be used.

This is why I tried to get clarity from the spec authors/Anne. I'm not entirely sure what else I can do. This isn't supposed to be creative writing.

files/en-us/web/api/element/sethtmlunsafe/index.md
hamishwillee: That's true. Though you probably wouldn't do it.

```suggestion
      - : A {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} object which defines what elements of the input will be allowed or removed.
         This can also be a string with the value `"default"`, which applies a `Sanitizer` with the default (XSS-safe) configuration.
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -48,25 +55,65 @@ None (`undefined`).
...
+const trustedHTML = policy.createHTML(untrustedString);
+```
+
+#### Using the method
 
-First we will create the {{domxref("ShadowRoot")}} we want to target.
+Now that we have `trustedHTML`, the code below shows how you can use it with `setHTMLUnsafe()`, both with and without a sanitizer.
+
+> [!NOTE]
+> The input may already sanitized by the trusted type policy at this point.
```
hamishwillee:
```suggestion
> The input may already have been sanitized by the trusted type policy at this point.
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -110,6 +154,10 @@ This example provides a "live" demonstration of the method when called with diff
 The code defines buttons that you can click to sanitize and inject a string of HTML using a default and a custom sanitizer, respectively.
```
hamishwillee:
```suggestion
The code defines buttons that you can click to inject a string of HTML. One button injects the HTML without sanitizing it at all, and the second uses a custom sanitizer than allows `<script>` elements but not other unsafe items.
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -22,12 +29,12 @@ setHTMLUnsafe(input, options)
 ### Parameters
 
 - `input`
-  - : A string or {{domxref("TrustedHTML")}} instance defining HTML to be parsed.
+  - : A {{domxref("TrustedHTML")}} or string instance defining HTML to be parsed.
 - `options` {{optional_inline}}
   - : An options object with the following optional parameters:
     - `sanitizer` {{optional_inline}}
       - : A {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} object which defines what elements of the input will be allowed or removed.
```
hamishwillee:
```suggestion
      - : A {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} object which defines what elements of the input will be allowed or removed.
        This can also be a string with the value `"default"`, which applies a `Sanitizer` with the default (XSS-safe) configuration.
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -22,12 +29,12 @@ setHTMLUnsafe(input, options)
...
-        Note that generally a `"Sanitizer` is expected than the to be more efficient than a `SanitizerConfig` if the configuration is to reused.
+        Note that generally a `"Sanitizer` is expected to be more efficient than a `SanitizerConfig` if the configuration is to reused.
```
hamishwillee:
```suggestion
        Note that generally a `Sanitizer` is expected to be more efficient than a `SanitizerConfig` if the configuration is to reused.
```

files/en-us/web/api/element/sethtmlunsafe/index.md
hamishwillee: The theory from the sanitizer explainer is that for the HTML setting cases the browser can do a better job than something like DOMPurify because it knows the context - i.e. it will additionally strip out things that don't make sense in the context (such as table rows outside of a table).
Further in all contexts (not just the HTML methods) it knows the subtleties of the browser implementation so it can catch things that an external library wouldn't be able to "see". This is true for both safe and unsafe methods.

It differs from innerHTML in two ways. 
1. innerHTML is an older method that does not allow injection of ShadowRoots. So if you need to use shadowRoots you have to use this or the safe method for injection.
2. innerHTML has some (ineffective) protections against script execution - specifically, it doesn't allow script elements to execute during import. With this method you can inject whatever you like.

So it is actually more dangerous that innerHTML if used without a sanitiser. 

EDIT Anyway, see link below for a story that positions when/why you might use the sanitizer.

files/en-us/web/api/element/sethtmlunsafe/index.md
hamishwillee: OK. Tried to integrate this.

files/en-us/web/api/element/sethtmlunsafe/index.md
hamishwillee: @wbamberg OK, I believe I have "fixed this" at least structurally (bit of a rush job, so needs another read over).

The story I am telling is that:
- Always use the setHTML variant if you can
- If you use the setHTMLUnsafe() variant you must use trusted types and you shouldn't need to supply a sanitizer
- You might use setHTMLUnsafe() in cases where you can't use TT or setHTML because it allows declarative shadow roots, and using it with a default sanitizer is much safer than using innerHTML with no sanitizer.

The examples show strictly that - the TT example passes no sanitizer. There is a section on "When should I use it" that captures these bits ^^^.

It works because I don't need to explain that the sanitizer bit here is mostly pointless or that the cases are unlikely, just that it would be useful if you had to.

files/en-us/web/api/element/sethtmlunsafe/index.md
hamishwillee: See https://github.com/mdn/content/pull/40420#discussion_r2281294967

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -46,56 +56,153 @@ None (`undefined`).
...
+The method is therefore a possible vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, where potentially unsafe strings provided by a user are injected into the DOM without first being sanitized.
+
+You should mitigate this risk by always passing {{domxref("TrustedHTML")}} objects instead of strings, and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup (such as {{htmlelement("script")}} elements and event handler attributes), before it is injected.
+
+Using `TrustedHTML` makes it possible to audit and check that sanitization code is effective in just a few places, rather than scattered across all your injection sinks.
+It should be unnecessary to additionally pass a sanitizer to the method when using `TrustedHTML`.
```
wbamberg: less passive?

```suggestion
Using `TrustedHTML` makes it possible to audit and check that sanitization code is effective in just a few places, rather than scattered across all your injection sinks.
You should not have to pass a sanitizer to the method when using `TrustedHTML`.
```

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -46,56 +56,153 @@ None (`undefined`).
...
-Note that since this method does not necessarily sanitize input strings of XSS-unsafe entities, input strings should also be validated using the [Trusted Types API](/en-US/docs/Web/API/Trusted_Types_API).
-If the method is used with both a trusted types and a sanitizer, the input string will be passed through the trusted transformation function before it is sanitized.
+  You can't use `setHTML()` in this case, because it strips all unsafe entities.
+  You could use `setHTMLUnsafe()` without a sanitizer or `innerHTML`, but that would allow all unsafe entities.
+
+  A better option here is to call `setHTMLUnsafe()` with a sanitizer that allows just those dangerous elements and attributes we actually need.
+  While this is still unsafe, it is safer than allowing all of them.
+
+For the last point, consider a situation where your code relies on being able to use unsafe `onclick` handlers.
+The following code shows the effect of different the different methods and sanitizers for this case.
```
wbamberg:
```suggestion
The following code shows the effect of the different methods and sanitizers for this case.
```

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -46,56 +56,153 @@ None (`undefined`).
...
+target.setHTMLUnsafe(input);
+target.innerHTML = input;
+
+// Safe - removes all XSS-unsafe entities.
+const configSafe = new Sanitizer();
+target.setHTMLUnsafe(input, { sanitizer: configSafe });
+
+// Removes all XSS-unsafe entities except `onerror`
+const configLessSafe = new Sanitizer();
+config.allowAttribute("onerror");
```
wbamberg:
```suggestion
// Removes all XSS-unsafe entities except `onclick`
const configLessSafe = new Sanitizer();
config.allowAttribute("onclick");
```

files/en-us/web/api/element/sethtmlunsafe/index.md
```diff
@@ -46,56 +56,153 @@ None (`undefined`).
...
   sanitizer: { removeElements: ["script"] },
 });
 ```
 
 ### `setHTMLUnsafe()` live example
 
 This example provides a "live" demonstration of the method when called with different sanitizers.
-The code defines buttons that you can click to inject a string of HTML that is not sanitized, and that uses and a custom sanitizer, respectively.
+The code defines buttons that you can click to inject a string of HTML.
+One button injects the HTML without sanitizing it at all, and the second uses a custom sanitizer than allows `<script>` elements but not other unsafe items.
```
wbamberg:
```suggestion
One button injects the HTML without sanitizing it at all, and the second uses a custom sanitizer that allows `<script>` elements but not other unsafe items.
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -48,25 +58,102 @@ None (`undefined`).
...
+`setHTMLUnsafe()` doesn't perform any sanitization by default.
+If no sanitizer is passed as a parameter, all HTML entities in the input will be injected.
+
+### Security considerations
+
+The suffix "Unsafe" in the method name indicates that it does not enforce removal of all XSS-unsafe HTML entities (unlike {{domxref("ShadowRoot.setHTML()")}}).
+While it can do so if used with an appropriate sanitizer, it doesn't have to use an effective sanitizer, or any sanitizer at all!
+The method is therefore a possible vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, where potentially unsafe strings provided by a user are injected into the DOM without first being sanitized.
+
+You should mitigate this risk by always passing {{domxref("TrustedHTML")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
```
wbamberg:
```suggestion
You should mitigate this risk by always passing {{domxref("TrustedHTML")}} objects instead of strings, and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -48,25 +58,102 @@ None (`undefined`).
...
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup (such as {{htmlelement("script")}} elements and event handler attributes), before it is injected.
+
+Using `TrustedHTML` makes it possible to audit and check that sanitization code is effective in just a few places, rather than scattered across all your injection sinks.
+It should be unnecessary to additionally pass a sanitizer to the method when using `TrustedHTML`.
```
wbamberg:
```suggestion
You should not have to pass a sanitizer to the method when using `TrustedHTML`.
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -48,25 +58,102 @@ None (`undefined`).
...
+- You have an edge case where you have to allow HTML input that includes a known set of unsafe HTML entities.
+
+  You can't use `setHTML()` in this case, because it strips all unsafe entities.
+  You could use `setHTMLUnsafe()` without a sanitizer or `innerHTML`, but that would allow all unsafe entities.
+
+  A better option here is to call `setHTMLUnsafe()` with a sanitizer that allows just those dangerous elements and attributes we actually need.
+  While this is still unsafe, it is safer than allowing all of them.
+
+For the last point, consider a situation where your code relies on being able to use unsafe `onclick` handlers.
+The following code shows the effect of different the different methods and sanitizers on this case.
```
wbamberg:
```suggestion
The following code shows the effect of the different methods and sanitizers on this case.
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -48,25 +58,102 @@ None (`undefined`).
...
 
-Note that since this method does not necessarily sanitize input strings of XSS-unsafe entities, input strings should also be validated using the [Trusted Types API](/en-US/docs/Web/API/Trusted_Types_API).
-If the method is used with both a trusted types and a sanitizer, the input string will be passed through the trusted transformation function before it is sanitized.
+// Safe - removes all XSS-unsafe entities.
+const configSafe = new Sanitizer();
+shadow.setHTMLUnsafe(input, { sanitizer: configSafe });
+
+// Removes all XSS-unsafe entities except `onerror`
+const configLessSafe = new Sanitizer();
+config.allowAttribute("onerror");
```
wbamberg:
```suggestion
// Removes all XSS-unsafe entities except `onclick`
const configLessSafe = new Sanitizer();
config.allowAttribute("onclick");
```

files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
```diff
@@ -77,38 +164,52 @@ This could be created programmatically using {{domxref("Element.attachShadow()")
...
 });
 ```
 
 ### `setHTMLUnsafe()` live example
 
 This example provides a "live" demonstration of the method when called with different sanitizers.
-The code defines buttons that you can click to sanitize and inject a string of HTML using a default and a custom sanitizer, respectively.
-The original string and sanitized HTML are logged so you can inspect the results in each case.
+The code defines buttons that you can click to inject a string of HTML.
+One button injects the HTML without sanitizing it at all, and the second uses a custom sanitizer than allows `<script>` elements but not other unsafe items.
```
wbamberg:
```suggestion
One button injects the HTML without sanitizing it at all, and the second uses a custom sanitizer that allows `<script>` elements but not other unsafe items.
```

files/en-us/web/api/document/parsehtmlunsafe_static/index.md
```diff
@@ -49,18 +57,25 @@ A {{domxref("Document")}}.
...
+The method is therefore a possible vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, where potentially unsafe strings provided by a user are injected into the DOM without first being sanitized.
+
+You should mitigate this risk by always passing {{domxref("TrustedHTML")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup (such as {{htmlelement("script")}} elements and event handler attributes), before it is injected.
+
+Using `TrustedHTML` makes it possible to audit and check that sanitization code is effective in just a few places, rather than scattered across all your injection sinks.
+It should be unnecessary to additionally pass a sanitizer to the method when using `TrustedHTML`.
```
wbamberg:
```suggestion
You should not need to pass a sanitizer to the method when using `TrustedHTML`.
```

files/en-us/web/api/document/parsehtmlunsafe_static/index.md
```diff
@@ -22,7 +30,7 @@ Document.parseHTMLUnsafe(input, options)
 ### Parameters
 
 - `input`
-  - : A string or {{domxref("TrustedHTML")}} instance defining HTML to be parsed.
+  - : A {{domxref("TrustedHTML")}} or string instance defining HTML to be parsed.
 - `options` {{optional_inline}}
   - : An options object with the following optional parameters:
     - `sanitizer` {{optional_inline}}
```
wbamberg, line 36: This also needs to list `"default"`, and has the rogue opening quote before `Sanitizer`.

wbamberg (approved):
Thanks for all your work on this and sorry it's annoying. I just made little comments so have approved in case you are able to get this in before you go away. I don't know if you also wanted to wait for Luke's review.

I still don't really think it makes sense but I don't think this is your fault!

files/en-us/web/api/document/parsehtmlunsafe_static/index.md
```diff
@@ -22,11 +30,12 @@ Document.parseHTMLUnsafe(input, options)
...
       - : A {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} object which defines what elements of the input will be allowed or removed.
+        This can also be a string with the value `"default"`, which applies a `Sanitizer` with the default (XSS-safe) configuration.
         Note that generally a `"Sanitizer` is expected than the to be more efficient than a `SanitizerConfig` if the configuration is to reused.
         If not specified, no sanitizer is used.
```
hamishwillee:
```suggestion
        This can also be a string with the value `"default"`, which applies a `Sanitizer` with the default (XSS-safe) configuration.
        If not specified, no sanitizer is used.
        
        Note that generally a `Sanitizer` is expected than the to be more efficient than a `SanitizerConfig` if the configuration is to reused.
```

hamishwillee:
Thanks very much for the review @wbamberg - I've accepted all the suggestions. I was a bit sloppy - just trying to get something "acceptable".

I'm not unhappy with the docs. I think this tells it how it is in the nicest way possible. In Australia we call this "polishing a turd".

PR #40423, opened 2025-07-18
Element.innerHTML - add info trusted types

hamishwillee:
This updates the following properties to explain how they are used with TrustedTypes:

- [x] [`Element.innerHTML` ](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML)
- [x] [`Element.insertAdjacentHTML()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)

This is in progress:
- Add boilerplate
- Add TT information, including the new exception and tinyfill
- Restructured - this isn't to current templates. Need to move the guid-ish stuff either down as examples or up into description

This does not mention the sanitizer methods as alternatives since the `setHTML()` is not implemented anywhere, and the sanitizer in `setHTMLUnsafe()` is not yet implemented. We're in separate discussion on those and when they are in a release we can revisit.

### DIFF
--- a/files/en-us/web/api/element/innerhtml/index.md
+++ b/files/en-us/web/api/element/innerhtml/index.md
@@ -23,11 +23,8 @@ To insert the HTML into the document rather than replace the contents of an elem
 
 Getting the property returns a string containing the HTML serialization of the element's descendants.
 
-Setting the value of `innerHTML` removes all of the element's descendants and replaces them with nodes constructed by parsing the HTML given in the assigned {{domxref("TrustedHTML")}} or string.
+Setting the property accepts either a {{domxref("TrustedHTML")}} object or a string. It parses this value as HTML and replaces all the element's descendants with the result.
 When set to the `null` value, that `null` value is converted to the empty string (`""`), so `elt.innerHTML = null` is equivalent to `elt.innerHTML = ""`.
-Note that {{htmlelement("script")}} elements in the assigned value are injected but not executed.
-
-Shadow roots are dropped from the serialized and injected HTML.
 
 ### Exceptions
 
@@ -102,7 +99,7 @@ if (typeof trustedTypes === "undefined")
   trustedTypes = { createPolicy: (n, rules) => rules };
 ```
 
-Next create a {{domxref("TrustedTypePolicy")}} that defines a {{domxref("TrustedTypePolicy/createHTML", "createHTML()")}} for transforming an input string into {{domxref("TrustedHTML")}} instances.
+Next we create a {{domxref("TrustedTypePolicy")}} that defines a {{domxref("TrustedTypePolicy/createHTML", "createHTML()")}} for transforming an input string into {{domxref("TrustedHTML")}} instances.
 Commonly implementations of `createHTML()` use a library such as [DOMPurify](https://github.com/cure53/DOMPurify) to sanitize the input as shown below:
 
 ```js
@@ -111,7 +108,7 @@ const policy = trustedTypes.createPolicy("my-policy", {
 });
 ```
 
-Then use this `policy` object to create a `TrustedHTML` object from the potentially unsafe input string, and assign the result to the element:
+Then we use this `policy` object to create a `TrustedHTML` object from the potentially unsafe input string, and assign the result to the element:
 
 ```js
 // The potentially malicious string
@@ -138,9 +135,6 @@ For example, the code below clears the entire contents of a document by setting 
 // Create a TrustedHTML instance using the policy
 document.body.textContent = policy.createHTML(policy.createHTML(""));
 ```
-
-Note that in this case we're setting the element to an empty string, which we know is safe.
-If we weren't [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) we could instead directly assign the empty string:
 
 ```js
 document.body.textContent = "";
--- a/files/en-us/web/api/element/innerhtml/index.md
+++ b/files/en-us/web/api/element/innerhtml/index.md
@@ -92,8 +92,8 @@ console.log(contents); // "\n  <p>My name is Joe</p>\n"
 
 ### Replacing the contents of an element
 
-Setting the value of `innerHTML` lets you replace the existing contents of an element with a new DOM tree parsed from an input.
-This completely removes the original content, including any event handlers associated with the removed elements.
+In this example we'll replace an element's DOM by assigning HTML to the element's `innerHTML` property.
+To mitigate the risk of XSS, we'll first create a `TrustedHTML` object from the string containing the HTML, and then assign that object to `innerHTML`.
 
 Trusted types are not yet supported on all browsers, so first we define the [trusted types tinyfill](/en-US/docs/Web/API/Trusted_Types_API#trusted_types_tinyfill).
 This acts as a transparent replacement for the trusted types JavaScript API:
--- a/files/en-us/web/api/element/innerhtml/index.md
+++ b/files/en-us/web/api/element/innerhtml/index.md
@@ -271,3 +271,4 @@ You can see output into the log by moving the mouse in and out of the box, click
 - {{domxref("ShadowRoot.getHTML()")}}
 - {{domxref("Element.setHTMLUnsafe()")}}
 - {{domxref("ShadowRoot.setHTMLUnsafe()")}}
+- [Trusted Types API](/en-US/docs/Web/API/Trusted_Types_API)
--- a/files/en-us/web/api/element/innerhtml/index.md
+++ b/files/en-us/web/api/element/innerhtml/index.md
@@ -8,36 +8,69 @@ browser-compat: api.Element.innerHTML
 
 {{APIRef("DOM")}}
 
-The **`innerHTML`** property of the {{domxref("Element")}} interface gets or sets the HTML or XML markup contained within the element.
+> [!WARNING]
+> This property parses its input as HTML, writing the result into the DOM.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
+>
+> You can reduce the risk by assigning {{domxref("TrustedHTML")}} objects instead of strings, and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+> This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
 
-More precisely, `innerHTML` gets a serialization of the nested child DOM elements within the element, or sets HTML or XML that should be parsed to replace the DOM tree within the element.
+The **`innerHTML`** property of the {{domxref("Element")}} interface gets or sets the HTML or XML markup contained within the element, omitting any {{glossary("shadow tree", "shadow roots")}} in both cases.
 
 To insert the HTML into the document rather than replace the contents of an element, use the method {{domxref("Element.insertAdjacentHTML", "insertAdjacentHTML()")}}.
 
-The serialization of the DOM tree read from the property does not include {{glossary("shadow tree", "shadow roots")}} — if you want to get a HTML string that includes shadow roots, you must instead use the {{domxref("Element.getHTML()")}} or {{domxref("ShadowRoot.getHTML()")}} methods.
-Similarly, when setting element content using `innerHTML`, the HTML string is parsed into DOM elements that do not contain shadow roots.
-
-So for example [`<template>`](/en-US/docs/Web/HTML/Reference/Elements/template) is parsed into as {{domxref("HTMLTemplateElement")}}, whether or not the [`shadowrootmode`](/en-US/docs/Web/HTML/Reference/Elements/template#shadowrootmode) attribute is specified
-In order to set an element's contents from an HTML string that includes declarative shadow roots, you must use either {{domxref("Element.setHTMLUnsafe()")}} or {{domxref("ShadowRoot.setHTMLUnsafe()")}}.
-
-Note that some browsers serialize the `<` and `>` characters as `&lt;` and `&gt;` when they appear in attribute values (see [Browser compatibility](#browser_compatibility)).
-This is to prevent a potential security vulnerability ([mutation XSS](https://www.securitum.com/mutation-xss-via-mathml-mutation-dompurify-2-0-17-bypass.html)) in which an attacker can craft input that bypasses a [sanitization function](/en-US/docs/Web/Security/Attacks/XSS#sanitization), enabling a cross-site scripting (XSS) attack.
-
 ## Value
 
-A string containing the HTML serialization of the element's descendants.
-Setting the value of `innerHTML` removes all of the element's descendants and replaces them with nodes constructed by parsing the HTML given in the string _htmlString_.
+Getting the property returns a string containing the HTML serialization of the element's descendants.
 
+Setting the value of `innerHTML` removes all of the element's descendants and replaces them with nodes constructed by parsing the HTML given in the assigned {{domxref("TrustedHTML")}} or string.
 When set to the `null` value, that `null` value is converted to the empty string (`""`), so `elt.innerHTML = null` is equivalent to `elt.innerHTML = ""`.
+Note that {{htmlelement("script")}} elements in the assigned value are injected but not executed.
+
+Shadow roots are dropped from the serialized and injected HTML.
 
 ### Exceptions
 
 - `SyntaxError` {{domxref("DOMException")}}
   - : Thrown if an attempt was made to set the value of `innerHTML` using a string which is not properly-formed HTML.
+- `TypeError`
+  - : Thrown if the property is set to a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
 - `NoModificationAllowedError` {{domxref("DOMException")}}
   - : Thrown if an attempt was made to insert the HTML into a node whose parent is a {{domxref("Document")}}.
 
-## Usage notes
+## Description
+
+`innerHTML` gets a serialization of the nested child DOM elements within the element, or sets HTML or XML that should be parsed to replace the DOM tree within the element.
+
+The serialization of the DOM tree read from the property does not include {{glossary("shadow tree", "shadow roots")}} — if you want to get a HTML string that includes shadow roots, you must instead use the {{domxref("Element.getHTML()")}} or {{domxref("ShadowRoot.getHTML()")}} methods.
+
+Similarly, when setting element content using `innerHTML`, the HTML string is parsed into DOM elements that do not contain shadow roots.
+So for example [`<template>`](/en-US/docs/Web/HTML/Reference/Elements/template) is parsed into as {{domxref("HTMLTemplateElement")}}, whether or not the [`shadowrootmode`](/en-US/docs/Web/HTML/Reference/Elements/template#shadowrootmode) attribute is specified.
+In order to set an element's contents from an HTML string that includes declarative shadow roots, you must instead use {{domxref("Element.setHTMLUnsafe()")}} or {{domxref("ShadowRoot.setHTMLUnsafe()")}}.
+
+Note that some browsers serialize the `<` and `>` characters as `&lt;` and `&gt;` when they appear in attribute values (see [Browser compatibility](#browser_compatibility)).
+This is to prevent a potential security vulnerability ([mutation XSS](https://research.securitum.com/dompurify-bypass-using-mxss/)) in which an attacker can craft input that bypasses a [sanitization function](/en-US/docs/Web/Security/Attacks/XSS#sanitization), enabling a cross-site scripting (XSS) attack.
+
+### Security considerations
+
+The `innerHTML` property is probably the most common vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, where potentially unsafe strings provided by a user are injected into the DOM without first being sanitized.
+While the property does prevent {{HTMLElement("script")}} elements from executing when they are injected, it is susceptible to many other ways that attackers can craft HTML to run malicious JavaScript.
+For example, the following example would execute the code in the `error` event handler, because the {{htmlelement("img")}} `src` value is not a valid image URL:
+
+```js
+const name = "<img src='x' onerror='alert(1)'>";
+el.innerHTML = name; // shows the alert
+```
+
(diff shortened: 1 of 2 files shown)


files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -79,15 +153,6 @@ document.documentElement.innerHTML = `<pre>${document.documentElement.innerHTML.
 )}</pre>`;
 ```
 
-#### Operational details
```
hamishwillee, line 82: This is interesting I guess. But it doesn't actually matter. If it did/does it should be in the description part.

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -115,47 +180,11 @@ list.innerHTML += `<li><a href="#">Item ${list.children.length + 1}</a></li>`;
...
-- {{domxref("Node.textContent")}} when inserting plain text, as this inserts it as raw text rather than parsing it as HTML.
-
-> [!WARNING]
-> If your project is one that will undergo any form of security review, using `innerHTML` most likely will result in your code being rejected.
-> For example, [if you use `innerHTML`](https://wiki.mozilla.org/Add-ons/Reviewers/Guide/Reviewing#Step_2:_Automatic_validation) in a [browser extension](/en-US/docs/Mozilla/Add-ons/WebExtensions) and submit
-> the extension to [addons.mozilla.org](https://addons.mozilla.org/), it may be rejected in the review process.
-> Please see [Safely inserting external content into a page](/en-US/docs/Mozilla/Add-ons/WebExtensions/Safely_inserting_external_content_into_a_page) for alternative methods.
-
-## Examples
+## Adding a logging mechanism using innerHTML
```
hamishwillee:
```suggestion
### Adding a logging mechanism using innerHTML
```

hamishwillee:
FYI @lukewarlow @wbamberg - this is the trusted types update for `Element.innerHTML`. Luke,  I've taken on board your comments about sanitizer not yet being implemented. I still mention the `setHTMLUnsafe()` because it  allows set with shadow roots, but not as alternatives with sanitization. It should however be possible to graft that information on here quite easily.

All the examples use trustedHTML. This also shows the first live example use of the tinyfill in the docs. This is duplication with the main docs, but it feels very practical to me given the current state of trustedtypes.

files/en-us/web/api/element/insertadjacenthtml/index.md
```diff
@@ -105,15 +119,35 @@ code {
 
 #### JavaScript
 
+```js hidden
```
hamishwillee: Decided to hide the tinyfill in this case. Do you think it should be here. Or perhaps a note "if trusted types not supported you should add a tinyfill".

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -8,36 +8,69 @@ browser-compat: api.Element.innerHTML
...
-Note that some browsers serialize the `<` and `>` characters as `&lt;` and `&gt;` when they appear in attribute values (see [Browser compatibility](#browser_compatibility)).
-This is to prevent a potential security vulnerability ([mutation XSS](https://research.securitum.com/dompurify-bypass-using-mxss/)) in which an attacker can craft input that bypasses a [sanitization function](/en-US/docs/Web/Security/Attacks/XSS#sanitization), enabling a cross-site scripting (XSS) attack.
-
 ## Value
 
-A string containing the HTML serialization of the element's descendants.
-Setting the value of `innerHTML` removes all of the element's descendants and replaces them with nodes constructed by parsing the HTML given in the string _htmlString_.
+Getting the property returns a string containing the HTML serialization of the element's descendants.
 
+Setting the value of `innerHTML` removes all of the element's descendants and replaces them with nodes constructed by parsing the HTML given in the assigned {{domxref("TrustedHTML")}} or string.
```
wbamberg: I think it might help some of the other discussion to be clearer about the types that this property can accept. As it stands, especially the "Setting the value" para starts with what it does, not what types it accepts, which turns out to be important of course for when we want to talk about trusted types.

Maybe something like:

```suggestion
Getting the property returns a string containing the HTML serialization of the element's descendants.

Setting the property accepts either a {{domxref("TrustedHTML")}} object or a string. It parses this value as HTML and replaces all the element's descendants with the result.
```

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -8,36 +8,69 @@ browser-compat: api.Element.innerHTML
...
 When set to the `null` value, that `null` value is converted to the empty string (`""`), so `elt.innerHTML = null` is equivalent to `elt.innerHTML = ""`.
+Note that {{htmlelement("script")}} elements in the assigned value are injected but not executed.
```
wbamberg: Maybe better in the security section, where it is mentioned? Anyway, seems a poor fit for "Value".

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -8,36 +8,69 @@ browser-compat: api.Element.innerHTML
...
+
+Shadow roots are dropped from the serialized and injected HTML.
```
wbamberg: Also doesn't seem a good fit for "Value", and is already covered in "Description".

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -8,36 +8,69 @@ browser-compat: api.Element.innerHTML
...
-## Usage notes
+## Description
+
+`innerHTML` gets a serialization of the nested child DOM elements within the element, or sets HTML or XML that should be parsed to replace the DOM tree within the element.
+
+The serialization of the DOM tree read from the property does not include {{glossary("shadow tree", "shadow roots")}} — if you want to get a HTML string that includes shadow roots, you must instead use the {{domxref("Element.getHTML()")}} or {{domxref("ShadowRoot.getHTML()")}} methods.
+
+Similarly, when setting element content using `innerHTML`, the HTML string is parsed into DOM elements that do not contain shadow roots.
+So for example [`<template>`](/en-US/docs/Web/HTML/Reference/Elements/template) is parsed into as {{domxref("HTMLTemplateElement")}}, whether or not the [`shadowrootmode`](/en-US/docs/Web/HTML/Reference/Elements/template#shadowrootmode) attribute is specified.
+In order to set an element's contents from an HTML string that includes declarative shadow roots, you must instead use {{domxref("Element.setHTMLUnsafe()")}} or {{domxref("ShadowRoot.setHTMLUnsafe()")}}.
```
wbamberg: Because this is about a very distinct aspect of innerHTML, I would have these two paragraphs under an H3 like "Shadow DOM considerations" or something, and move it after the para about "Note that some browsers...".

You might also consider moving "Note that some browsers..." into "Security considerations", maybe.

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -8,36 +8,69 @@ browser-compat: api.Element.innerHTML
...
+For example, the following example would execute the code in the `error` event handler, because the {{htmlelement("img")}} `src` value is not a valid image URL:
+
+```js
+const name = "<img src='x' onerror='alert(1)'>";
+el.innerHTML = name; // shows the alert
+```
+
+You can mitigate these issues by always assigning {{domxref("TrustedHTML")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup before it is injected.
+This is demonstrated in the following sections.
```
wbamberg: This is unclear. Which of the following sections demonstrate this? As a reader I want a direct link to a concrete example of this.

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -8,36 +8,69 @@ browser-compat: api.Element.innerHTML
...
+
+> [!NOTE]
+> {{domxref("Node.textContent")}} should be used when you know that the user provided content should be plain text.
+> This prevents it being parsed as HTML.
+
+## Examples
 
 ### Reading the HTML contents of an element
```
wbamberg, line 73: In the line below, "comprised of" -> "comprising"?

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -51,126 +84,145 @@ const contents = myElement.innerHTML;
 This lets you look at the HTML markup of the element's content nodes.
 
 > [!NOTE]
-> The returned HTML or XML fragment is generated based on the current contents of the element, so the markup and formatting of the returned fragment is likely not to match the original page markup.
+> The returned HTML or XML fragment is generated based on the current contents of the element, so the markup and formatting of the returned fragment may not match the original page markup.
+> It will also omit any shadow roots.
```
wbamberg: We already say this in the description, I don't think we need to say it again in the example. Examples are illustrative of things we have already described. If we want to talk about shadow DOM in this section it should be via an example (i.e. a shadow DOM that's not being serialized).

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -51,126 +84,145 @@ const contents = myElement.innerHTML;
...
 
 ### Replacing the contents of an element
 
-Setting the value of `innerHTML` lets you easily replace the existing contents of an element with new content.
+Setting the value of `innerHTML` lets you replace the existing contents of an element with a new DOM tree parsed from an input.
+This completely removes the original content, including any event handlers associated with the removed elements.
```
wbamberg: Again, stuff like this is descriptive, not illustrative. It should be (and is) covered in Description, not Examples. We might replace this preamble with something like:

> In this example we'll replace an element's DOM by assigning HTML to the element's `innerHTML` property. To mitigate the risk of XSS, we'll first create a TrustedHTML object from the string containing the HTML, and then assign that object to `innerHTML`.

files/en-us/web/api/element/innerhtml/index.md
wbamberg: Actually I wonder if any of this note belongs in the example. It would be better for the example to show some actual DOM that's being retrieved, like:

```html
<div id="example">
  <p>My name is Joe</p>
</div>
```

```js
const myElement = document.querySelector("#example");
const contents = myElement.innerHTML; // "\n  <p>My name is Joe</p>\n"
```

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -51,126 +84,145 @@ const contents = myElement.innerHTML;
...
 
-> [!WARNING]
-> This is a [security risk](#security_considerations) if the string to be inserted might contain potentially malicious content.
-> When inserting user-supplied data you should always consider using a sanitizer library, in order to sanitize the content before it is inserted.
-
-For example, you can erase the entire contents of a document by clearing the contents of the document's {{domxref("Document.body", "body")}} attribute:
+Trusted types are not yet supported on all browsers, so first we define the [trusted types tinyfill](/en-US/docs/Web/API/Trusted_Types_API#trusted_types_tinyfill).
+This provides an implementation of `trustedTypes.createPolicy()` which just returns the `policyOptions` object it was passed.
+The object defines sanitization functions for data, and these functions are expected to return strings.
+Effectively it is a transparent replacement for the trusted types JavaScript API:
```
wbamberg: I agree it's worth including the tinyfill but might omit the description - we do link to the tinyfill description in case people don't want to take this on trust (and IMO if someone is to really understand how the tinyfill works they need to read that description anyway).

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -51,126 +84,145 @@ const contents = myElement.innerHTML;
...
-document.body.textContent = "";
+if (typeof trustedTypes === "undefined")
+  trustedTypes = { createPolicy: (n, rules) => rules };
 ```
 
-This example fetches the document's current HTML markup and replaces the `"<"` characters with the {{glossary("character reference")}} `"&lt;"`, thereby essentially converting the HTML into raw text.
-This is then wrapped in a {{HTMLElement("pre")}} element.
-Then the value of `innerHTML` is changed to this new string.
-As a result, the document contents are replaced with a display of the page's entire source code.
+Next create a {{domxref("TrustedTypePolicy")}} that defines a {{domxref("TrustedTypePolicy/createHTML", "createHTML()")}} for transforming an input string into {{domxref("TrustedHTML")}} instances.
```
wbamberg:
```suggestion
Next we create a {{domxref("TrustedTypePolicy")}} that defines a {{domxref("TrustedTypePolicy/createHTML", "createHTML()")}} for transforming an input string into {{domxref("TrustedHTML")}} instances.
```

Very pedantic, but this started "first we" so perhaps "next we" afterwards? Just "next" implicitly changes it from first-person to second.

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -51,126 +84,145 @@ const contents = myElement.innerHTML;
...
-  /</g,
-  "&lt;",
-)}</pre>`;
+const policy = trustedTypes.createPolicy("my-policy", {
+  createHTML: (input) => DOMPurify.sanitize(input),
+});
 ```
 
-#### Operational details
+Then use this `policy` object to create a `TrustedHTML` object from the potentially unsafe input string, and assign the result to the element:
```
wbamberg:
```suggestion
Then we use this `policy` object to create a `TrustedHTML` object from the potentially unsafe input string, and assign the result to the element:
```

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -51,126 +84,145 @@ const contents = myElement.innerHTML;
...
 
-```html
-<ul id="list">
-  <li><a href="#">Item 1</a></li>
-  <li><a href="#">Item 2</a></li>
-  <li><a href="#">Item 3</a></li>
-</ul>
+```js
+// Create a TrustedHTML instance using the policy
+document.body.textContent = policy.createHTML(policy.createHTML(""));
```
wbamberg: Is this supposed to be assigning to textContent?

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -51,126 +84,145 @@ const contents = myElement.innerHTML;
...
 ```
 
-#### JavaScript
+Note that in this case we're setting the element to an empty string, which we know is safe.
+If we weren't [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) we could instead directly assign the empty string:
```
wbamberg: I'm not sure we should encourage this. AIUI, if we don't enforce the use of trusted types, we lose a lot of the protection they offer (because you don't know any more whether somewhere in your codebase you might be unsafe).

So I would omit this.

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -51,126 +84,145 @@ const contents = myElement.innerHTML;
...
 ```js
-const list = document.getElementById("list");
-
-list.innerHTML += `<li><a href="#">Item ${list.children.length + 1}</a></li>`;
+document.body.textContent = "";
 ```
 
-Please note that using `innerHTML` to append HTML elements (e.g., `el.innerHTML += "<a href='…'>link</a>"`) will result in the removal of any previously set event listeners.
-That is, after you append any HTML element that way you won't be able to listen to the previously set event listeners.
+### Logging messages using innerHTML
```
wbamberg: Is this example worth keeping? Is this a pattern we want to encourage, versus explicitly creating the DOM elements to put in the log?

and I also wonder, is it a good idea to recommend people to define a no-op TT policy? To what extent does that open up a hole in the whole thing?

wbamberg (commented):
Just comments on `innerHTML` for now.

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -8,36 +8,69 @@ browser-compat: api.Element.innerHTML
...
-Setting the value of `innerHTML` removes all of the element's descendants and replaces them with nodes constructed by parsing the HTML given in the string _htmlString_.
+Getting the property returns a string containing the HTML serialization of the element's descendants.
 
+Setting the value of `innerHTML` removes all of the element's descendants and replaces them with nodes constructed by parsing the HTML given in the assigned {{domxref("TrustedHTML")}} or string.
 When set to the `null` value, that `null` value is converted to the empty string (`""`), so `elt.innerHTML = null` is equivalent to `elt.innerHTML = ""`.
+Note that {{htmlelement("script")}} elements in the assigned value are injected but not executed.
+
+Shadow roots are dropped from the serialized and injected HTML.
 
 ### Exceptions
```
hamishwillee:
```suggestion

### Exceptions
```

files/en-us/web/api/element/innerhtml/index.md
hamishwillee: Just removing from here. As you say, mentioned in the security section.

files/en-us/web/api/element/innerhtml/index.md
```diff
@@ -51,126 +84,145 @@ const contents = myElement.innerHTML;
...
-  <li><a href="#">Item 3</a></li>
-</ul>
+```js
+// Create a TrustedHTML instance using the policy
+document.body.textContent = policy.createHTML(policy.createHTML(""));
 ```
 
-#### JavaScript
+Note that in this case we're setting the element to an empty string, which we know is safe.
+If we weren't [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) we could instead directly assign the empty string:
```
hamishwillee:
```suggestion
```

files/en-us/web/api/element/innerhtml/index.md
hamishwillee: Thanks. FYI I've done the first, but not the "you might consider" - it's not something you need to consider in order to be secure.

files/en-us/web/api/element/innerhtml/index.md
hamishwillee: Just removing this. They'll get the examples when I'm good and ready :-)

files/en-us/web/api/element/innerhtml/index.md
hamishwillee: Avoided it altogether.

files/en-us/web/api/element/innerhtml/index.md
hamishwillee: Yes to all. Stole your example.

files/en-us/web/api/element/innerhtml/index.md
hamishwillee: Doh. Actually though, you'd be safer to use textContent for this - and less hassle.. I'm going to delete the section as pointless

files/en-us/web/api/element/innerhtml/index.md
hamishwillee: I've decided to delete for the reasons given - it doesn't teach anything new, except possibly wrong things.

> and I also wonder, is it a good idea to recommend people to define a no-op TT policy? To what extent does that open up a hole in the whole thing?

Right now you should mostly be able to avoid recommending this for HTML but not for script trusted types. In future there are likely to be cases where your policy may need to be no-op - such as when using the unsafe HTML parsing methods. Right now and in this doc you make a good point.  We shouldn't.

files/en-us/web/api/element/innerhtml/index.md
hamishwillee:
```suggestion
In this example we'll replace an element's DOM by assigning HTML to the element's `innerHTML` property.
To mitigate the risk of XSS, we'll first create a `TrustedHTML` object from the string containing the HTML, and then assign that object to `innerHTML`.
```

hamishwillee:
Thanks @wbamberg ! Feedback for the first document integrated. Much appreciated.

files/en-us/web/api/element/insertadjacenthtml/index.md
wbamberg: I think you should do the same thing here that you've done for the other one.

files/en-us/web/api/element/insertadjacenthtml/index.md
```diff
@@ -139,5 +173,5 @@ reset.addEventListener("click", () => {
 - {{domxref("Element.insertAdjacentElement()")}}
 - {{domxref("Element.insertAdjacentText()")}}
 - {{domxref("XMLSerializer")}}: Serialize a DOM tree into an XML string
-- [hacks.mozilla.org guest post](https://hacks.mozilla.org/2011/11/insertadjacenthtml-enables-faster-html-snippet-injection/) by Henri Sivonen including benchmark showing
-  that insertAdjacentHTML can be way faster in some cases.
+- [Trusted Types API](/en-US/docs/Web/API/Trusted_Types_API)
+- [hacks.mozilla.org guest post](https://hacks.mozilla.org/2011/11/insertadjacenthtml-enables-faster-html-snippet-injection/) by Henri Sivonen including benchmark showing that `insertAdjacentHTML()` can be way faster in some cases.
```
wbamberg: It might be time to retire this "look at this new thing in Firefox" blog post.

files/en-us/web/api/element/insertadjacenthtml/index.md
```diff
@@ -105,15 +119,35 @@ code {
...
 #### JavaScript
 
+```js hidden
+if (typeof trustedTypes === "undefined")
+  trustedTypes = { createPolicy: (n, rules) => rules };
+```
+
+While not required for this example, below we follow the recommendation of defining a policy to create {{domxref("TrustedHTML")}} objects from the input (we should also enforce the policy `safe-content-policy` using CSP).
+In this case we know the input is safe so this policy passes it through without modification.
+The commented code shows how you might instead use the "DOMPurify" library to sanitize content that wasn't trusted.
```
wbamberg: Similarly to the other example, I wonder about whether we should give people guidance like "we know the input is safe so this policy passes it through without modification".

I suppose part of the issue here is that you want to make it a live sample, but you can't do that and use DOMPurify, so it's hard to make it a "real" example.

One option would be, keep the same code but change the guidance from "we don't need a policy that does anything here" to "in order for this example to work without a third-party dependency, we've implemented a no-op policy here, but you shouldn't do this in real life".

Another option would be to stop making this a live sample, and use DOMPurify, but that would be sad because we want to show the effect of `position`.

Yet a third option would be to implement our own basic sanitizer, but that might be tricky to get right.

I'd love to know what @aaronshim thinks our guidance here ought to be. Is it good practice to have a no-op policy for content that we "know" is safe?

wbamberg (commented):
Thanks for the updates Hamish, they look great to me. I just had a couple of comments on the other page.

files/en-us/web/api/element/insertadjacenthtml/index.md
hamishwillee: I've done a modification of  "in order for this example to work without a third-party dependency, we've implemented a no-op policy here, but you shouldn't do this in real life".
Its the safest option and the one that most closely matches what we want users to do.
I definitely do not want to remove the live example - it is very helpful for visualising the insert behaviour :-)

hamishwillee: > I'd love to know what @aaronshim thinks our guidance here ought to be. Is it good practice to have a no-op policy for content that we "know" is safe?

Me too. 

I tend to think that the advice should generally be no - if you want to use these unsafe APIs then you're better always sanitizing. In most cases you'd instead replace using TT and use  `setHTML()` (once supported).
Of course the API itself doesn't care, and if the website has a policy that allows noops then developers will be allowed to use said noop policy.

The more interesting noop question is what you do with `setHTMLUnsafe()` (once supported) - the case where you know you have unsafe strings, but you perhaps need to allow them anyway. If that is a valid case, my contention is that you'd have a noop trusted type to avoid double sanitization.

That's a problem for https://github.com/mdn/content/pull/40420#issuecomment-3092282214

wbamberg: I'm happy with your choice here and let's merge it. We can always follow up if Aaron recommends something else.

wbamberg (approved):
Thank you Hamish!

hamishwillee:
No, thanks for the review and all the great advice @wbamberg . I feel better that we're finally getting these updated.

PR #40424, opened 2025-07-18
[CSS] Docs: Clarify the dual role of the font-size property

kannanwisen:
### Description

This PR reframes the introduction to the font-size page to more clearly and explicitly state the two primary roles the property serves in CSS.

### Motivation

The current introduction is correct, but it presents the property's role as a reference for other units (like em) as a secondary effect. For learners, it can be clearer to understand that font-size has two equally important, distinct jobs. This change helps build a more accurate mental model from the start.

### DIFF
--- a/files/en-us/web/css/font-size/index.md
+++ b/files/en-us/web/css/font-size/index.md
@@ -6,7 +6,9 @@ browser-compat: css.properties.font-size
 sidebar: cssref
 ---
 
-The **`font-size`** [CSS](/en-US/docs/Web/CSS) property sets the size of the font. Changing the font size also updates the sizes of the font size-relative {{cssxref("&lt;length&gt;")}} units, such as `em`, `ex`, and so forth.
+The **`font-size`** [CSS](/en-US/docs/Web/CSS) property has two primary functions:
+1. **It sets the size of the font.** Its most direct job is to specify the size of the glyphs for the text within an element. 
+2. **It acts as a reference for other properties.** It is the foundational unit used to compute font-relative {{cssxref("&lt;length&gt;")}} units like em, rem, ch etc. These units are then often used to set the value of other properties, such as width, margin, and padding.
 
 {{InteractiveExample("CSS Demo: font-size")}}
 
--- a/files/en-us/web/css/font-size/index.md
+++ b/files/en-us/web/css/font-size/index.md
@@ -7,7 +7,8 @@ sidebar: cssref
 ---
 
 The **`font-size`** [CSS](/en-US/docs/Web/CSS) property has two primary functions:
-1. **It sets the size of the font.** Its most direct job is to specify the size of the glyphs for the text within an element. 
+
+1. **It sets the size of the font.** Its most direct job is to specify the size of the glyphs for the text within an element.
 2. **It acts as a reference for other properties.** It is the foundational unit used to compute font-relative {{cssxref("&lt;length&gt;")}} units like em, rem, ch etc. These units are then often used to set the value of other properties, such as width, margin, and padding.
 
 {{InteractiveExample("CSS Demo: font-size")}}


Josh-Cena (commented):
The updated description sounds like `font-size` has two unrelated jobs, which it doesn't. It has a single job: it sets the font size. `em` and `ex` by definition are relative to the font size, so naturally they update w.r.t. to `font-size`, but that's the behavior of the units and not `font-size`.

kannanwisen:
How is height: 1ex calculated?

Josh-Cena:
em:

> Equal to the computed value of the [font-size](https://drafts.csswg.org/css-fonts-4/#propdef-font-size) property of the element on which it is used.

ex:

> Equal to the used x-height of the [first available font](https://www.w3.org/TR/css3-fonts/#first-available-font) [[CSS3-FONTS]](https://drafts.csswg.org/css-values/#biblio-css3-fonts).

Both are downstream definitions of the font size. The other units are, indeed, "a secondary effect".

kannanwisen:
Kindly don’t copy from Csswg document and explain this.

You read the csswg document applied to option.

Font size applies to element and text.

kannanwisen:
And your mdn docs says in applies to

kannanwisen:
A simple code



body style="font-size:100px"
span style="background-color:blue;"   /span
/body



can you please vary the height of the span according to x character height?

Kindly remember that, it has to work in responsive design too.

kannanwisen:
@Josh-Cena 

> The updated description sounds like `font-size` has two unrelated jobs, which it doesn't. It has a single job: it sets the font size. `em` and `ex` by definition are relative to the font size, so naturally they update w.r.t. to `font-size`, but that's the behavior of the units and not `font-size`.


Could you kindly justify your explanation regarding the three lines of code above? Please note that the <span> element does not contain any text — and this is intentional, as I do not want any text, since font-size is not solely intended for text.

Josh-Cena:
font-size has nothing to do with text in particular. An element can have an associated font size that would be used by any text in it, without actually having any text. It's just a configuration.

kannanwisen:
You can't copy from csswg and suggest the way you need,

image

image


Kindly provide with my code

Josh-Cena:
Your code has a span element that inherits the `font-size` setting from the body. It has no text in it, sure; but should there be any, they would be rendered using that size. You can think of the span as containing text of length zero. That doesn't change the conclusion that `em` is a downstream effect of `font-size` changing.

kannanwisen:
Your code has a span element that inherits the font-size setting from the body.

I am sorry to say this your are conpletely incorrect.

Kindly read my question.

can you please vary the height of the span according to x character height?

Josh-Cena:
There's nothing to argue here. Your change has been rejected because the wording before is clear enough and does not introduce any misunderstanding about the behavior of this property. We only accept fixes to incorrect or misleading content, because whether something is "clear" or "builds an accurate mental model" is very subject and YMMV. It's no longer up for debate.

kannanwisen:
Sorry to say rhis You are completely incorrect. I can increase the span height without the body fiont-size

Josh-Cena:
Sorry, what does that even mean? Sure, if you specify `height: 10ex`, then you can change the size of `ex` by either increasing font-size, or using a font with a larger x; how does that pertain to this section at all?

kannanwisen:
Hopefully I think my thought proceess is not wrong,

It is a open source project. The contributor has to prove his statement rather than copy from the CSSWG specification. And arguing that his interpretation of CSSWG statment is correct

kannanwisen:
> Sorry, what does that even mean? Sure, if you specify `height: 10ex`, then you can change the size of `ex` by either increasing font-size, or using a font with a larger x; how does that pertain to this section at all?

Sorry that is not correct. I will NOT change the 3 lines code.

Unfortunately you are are agreeing the my statement font-size is not only for text. My 3 lines code does not have any text.

Josh-Cena:
I repeat: `font-size` has a single job, which is to set the font size of the element. There are other features that are defined relative to the font size, which are therefore also affected, but those are secondary effects and not something independent of "sets the size of the font". You are free to have your own mental model, but this is the mode of explanation we chose for MDN, and we are not open to bikeshedding about this detail.

Josh-Cena:
> The **`font-size`** [CSS](/en-US/docs/Web/CSS) property sets the size of the font. Changing the font size also updates the sizes of the font size-relative {{cssxref("&lt;length&gt;")}} units, such as `em`, `ex`, and so forth.

Where in this paragraph do we mention "text" at all? Font size can exist without text, because it's just a configuration.

kannanwisen:
I repeat: font-size has a single job, which is to set the font size of the element. 
Your document says

image

It is not only for element.

It is for element and text.

kannanwisen:
Sorry to say your suggestion is incorrect for 50%.

Josh-Cena:
You can't set CSS directly on text. CSS has to be set on elements and inherited to text.

kannanwisen:
> I repeat: `font-size` has a single job, which is to set the font size of the element.

Your statement, what it does with the element.

kannanwisen:
> You can't set CSS directly on text. CSS has to be set on elements and inherited to text.

I knew from 1998.

Josh-Cena:
The content before is correct, so we don't accept rewrite PRs; that's the general rule. Please don't send rewrite PRs unless there's something obviously misleading or incorrect. This wastes the maintainers' time. And our decisions are final: we didn't accept your change.

kannanwisen:
@Josh-Cena 

Let not discuss for nothing,

My code is 3 lines.

no text is involved.

increase the height of the span without font-size accorindly to my requirement.

kannanwisen:
If you are able to prove it, we will stop this discussion here.

kannanwisen:
Kindly show me a complete code.

Josh-Cena:
OK let me phrase it more nicely if you are hung up on this: an element can have an associated font. A piece of text can have an associated font. The font has properties that can be customized via CSS properties. The CSS property `font-size` customizes the size of the font, by being applied to the element or text that the font is associated with. There are other things that reference the size of the font, such as `em` and `ex`. But `font-size` and "size of font" are two separate things, and the former happens to control the latter. The chain of cause-effect goes: `font-size` -> size of font -> value of `em`.

kannanwisen:
@Josh-Cena Kindly understand I am testing the 3 AI LLM for code generation. I am not here just to create a Pull request. I asked the LLM where you learned this concept. If they say MDN, then I raise the pull request.

Even they all agreed.

kannanwisen:
> OK let me phrase it more nicely if you are hung up on this: an element can have an associated font. A piece of text can have an associated font. The font has properties that can be customized via CSS properties. The CSS property `font-size` customizes the size of the font, by being applied to the element or text that the font is associated with. There are other things that reference the size of the font, such as `em` and `ex`. But `font-size` and "size of font" are two separate things, and the former happens to control the latter. The chain of cause-effect goes: `font-size` -> size of font -> value of `em`.

Sorry, no explanation. Just show your code.

Josh-Cena:
What does any of this have to do with code? Your change before and after have the exact same implications about behavior. You said:

> The current introduction is correct, but it presents the property's role as a reference for other units (like em) as a secondary effect. For learners, it can be clearer to understand that font-size has two equally important, distinct jobs. This change helps build a more accurate mental model from the start.

I'm telling you that this "more accurate mental model" is not what we chose.

kannanwisen:
Only the code the proof.

I am saying font-size is not for text alone.

Josh-Cena:
No one is saying that though? An element can also have a font size.

kannanwisen:
It is open source. It is not individual perception or conclusion.

If you conclude wrongly and closed the pull request, the whole world suffers.

Kindly provide your concept in code. No theorically.

Josh-Cena:
I have clearly explained the case: "font size" is a concept shared by both elements and text, and they have the exact same effect. I am a maintainer, and together with other maintainers, we own the final right of interpretation over what we deem as correct and accurate.

Locking because this discussion is not going anywhere. Your contributions are still very much appreciated, but this particular case is closed.

PR #40427, opened 2025-07-18
Add a "Guides" section to the API landing page template

Josh-Cena:
Many guide pages are not reachable from their parent page, which is a minus for discoverability and for consistent IA. This adds a "Guides" section to enforce consistent structure; afterwards I can send another PR to apply this structure to API pages.

### DIFF
--- a/files/en-us/mdn/writing_guidelines/page_structures/page_types/api_landing_page_template/index.md
+++ b/files/en-us/mdn/writing_guidelines/page_structures/page_types/api_landing_page_template/index.md
@@ -109,7 +109,7 @@ For a practical usage guide with code examples, you should include a "Usage…" 
 
 ## Guides
 
-Include a list of guide pages under this landing page. Each DT should link to the guide page.
+Include a list of guide pages under this landing page. Each DT should link to the guide page. This section is optional; if there's only a single "Using" guide, along with a few other conceptual guides, you may find it more convenient to link to it at the end of the "Concepts and usage" section instead. This section may be more helpful if there are so many guides that prose becomes hard to scan.
 
 - Using the ... API
   - : Intro paragraph of this guide page
--- a/files/en-us/mdn/writing_guidelines/page_structures/page_types/api_landing_page_template/index.md
+++ b/files/en-us/mdn/writing_guidelines/page_structures/page_types/api_landing_page_template/index.md
@@ -107,7 +107,14 @@ Don't go into a lot of detail in this section, and don't include code examples.
 If there are a lot of concepts to explain around this API, you should explain them in a separate "Fundamentals" or "Concepts" article (e.g., [Fundamentals of WebXR](/en-US/docs/Web/API/WebXR_Device_API/Fundamentals)).
 For a practical usage guide with code examples, you should include a "Usage…" article in your API docs (e.g., [Using the WebVR API](/en-US/docs/Web/API/WebVR_API/Using_the_WebVR_API)).
 
-To help improve content discoverability and {{Glossary("SEO")}}, keep the following tips in mind:
+## Guides
+
+Include a list of guide pages under this landing page. Each DT should link to the guide page.
+
+- Using the ... API
+  - : Intro paragraph of this guide page
+- Guide 2
+  - : Intro paragraph of this guide page
 
 ## Interfaces
 
--- a/files/en-us/mdn/writing_guidelines/page_structures/page_types/api_landing_page_template/index.md
+++ b/files/en-us/mdn/writing_guidelines/page_structures/page_types/api_landing_page_template/index.md
@@ -109,7 +109,7 @@ For a practical usage guide with code examples, you should include a "Usage…" 
 
 ## Guides
 
-Include a list of guide pages under this landing page. Each DT should link to the guide page. This section is optional; if there's only a single "Using" guide, along with a few other conceptual guides, you may find it more convenient to link to it at the end of the "Concepts and usage" section instead. This section may be more helpful if there are so many guides that prose becomes hard to scan.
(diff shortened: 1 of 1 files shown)


files/en-us/mdn/writing_guidelines/page_structures/page_types/api_landing_page_template/index.md
```diff
@@ -107,7 +107,14 @@ Don't go into a lot of detail in this section, and don't include code examples.
 If there are a lot of concepts to explain around this API, you should explain them in a separate "Fundamentals" or "Concepts" article (e.g., [Fundamentals of WebXR](/en-US/docs/Web/API/WebXR_Device_API/Fundamentals)).
 For a practical usage guide with code examples, you should include a "Usage…" article in your API docs (e.g., [Using the WebVR API](/en-US/docs/Web/API/WebVR_API/Using_the_WebVR_API)).
 
-To help improve content discoverability and {{Glossary("SEO")}}, keep the following tips in mind:
+## Guides
```
hamishwillee, line 110: FWIW I like the idea but I don't think it makes much "difference". After all, if you're on this page you can see the guides on the sidebar.

What would actually make a difference for users is to make the sidebar work properly and include these guides in the sidebar when you aren't on a guide page.

Josh-Cena: My checker requires every parent page to link to all its children. The main goal is to detect missing members on interface pages and other landing pages, but I think it provides value for this to be a general rule. The sidebar doesn't provide short descriptions about the page's content.

hamishwillee: Yes, I agree it is good - in particular there is an opportunity to talk about the content. My whine is because the guide isn't in the sidebar at all if you're not in a guide.

I have no objection to this at all. But since I wasn't assigned this is just a fly by :-)

Josh-Cena: If you look at https://developer.mozilla.org/en-US/docs/Web/API/URL for example, the sidebar does indeed contain guides. This was added last month: https://github.com/mdn/rari/pull/224

hamishwillee: That's awesome. Been complaining about it for years. Ignore me then.

hamishwillee (approved):
I'm approving. I think this is a good idea though ideally a guide should be well enough named that it would be hard to add useful extra information in just a paragraph.

Would be good to get another view - @chrisdavidmills ?

chrisdavidmills:
> I'm approving. I think this is a good idea though ideally a guide should be well enough named that it would be hard to add useful extra information in just a paragraph.
> 
> Would be good to get another view - @chrisdavidmills ?

I don't agree with this change. I don't think structuring it like this is necessarily a bad idea, however, this isn't the way we've traditionally done it. The guides are clearly linked at the top of the sidebar for those who want to find them right away, and we tend to put a line at the bottom of the "Concepts and usage" section linking to the guide(s), for more information on usage.

We'd have to update every page listed at https://developer.mozilla.org/en-US/docs/Web/API#specifications, for very little gain.

I do love that we are now linking guides from the interface page sidebar <3

Josh-Cena:
> We'd have to update every page listed at https://developer.mozilla.org/en-US/docs/Web/API#specifications, for very little gain.

I already have automation to catch all such cases, so it won't be a pain. The gains OTOH:

1. Every other page type has this requirement; it would be odd for API landing pages to _not_ have this. Interface pages also list all their members despite the exact same information present in the sidebar.
2. An extra paragraph of description for at-a-glance understanding of the guide landscape, which is especially useful for sequential guides such as WebXR, Web Audio, or Canvas.
3. It makes the IA more apparent because currently the guides feel more like peripheral "stuff" that happen to be nested under the landing page, but there's no strong logical connection between them, unlike interface & its members. A "guides" section makes it clear that these guides _belong_ to the API.

> we tend to put a line at the bottom of the "Concepts and usage" section linking to the guide(s), for more information on usage.

Sure, we have some cases of doing this, but in other cases the guides are completely unlinked from the landing pages, which is the main case I'm trying to fix here. If there's already a link, then this section isn't necessary.

chrisdavidmills:
> I already have automation to catch all such cases, so it won't be a pain. The gains OTOH:

OK, if we can fix these up automatically without too much hassle, I am less opposed to it.

> Sure, we have some cases of doing this, but in other cases the guides are completely unlinked from the landing pages, which is the main case I'm trying to fix here. If there's already a link, then this section isn't necessary.

Right. I can see how this section would be more useful in cases where there are several sequential guides. In the (numerous) cases where there is only one guide, I think it is a bit pointless. Can we include it, but mark it as optional, saying when to use it and when not to, and saying what the alternative is in the "not" cases?

Josh-Cena:
Of course, happy to do that.

files/en-us/mdn/writing_guidelines/page_structures/page_types/api_landing_page_template/index.md
```diff
@@ -107,7 +107,14 @@ Don't go into a lot of detail in this section, and don't include code examples.
...
+
+Include a list of guide pages under this landing page. Each DT should link to the guide page. This section is optional; if there's only a single "Using" guide, along with a few other conceptual guides, you may find it more convenient to link to it at the end of the "Concepts and usage" section instead. This section may be more helpful if there are so many guides that prose becomes hard to scan.
```
Josh-Cena:
```suggestion
Include a list of guide pages under this landing page. Each DT should link to the guide page. This section is optional; if there's only a single "Using" guide, along with a few other conceptual guides, you may find it more convenient to link to them as a paragraph at the end of the "Concepts and usage" section instead. This section may be more helpful if there are so many guides that prose becomes hard to scan.
```

Josh-Cena:
@chrisdavidmills How does this look?

chrisdavidmills:
> @chrisdavidmills How does this look?

Yup, that sounds fine. Merging!

PR #40434, opened 2025-07-18
Improve HTML formatting

alistairmcmillan:
Removing line breaks from two LABEL tags.

### Description

Simply removing line breaks, and placing two LABEL tags on their own lines.

### Motivation

Improves readability a little.


Josh-Cena (approved):
Sure, thank you

PR #40440, opened 2025-07-19
Remove broken links, fix content issues

Josh-Cena:
Remove sites that are no longer available or SSL cert has expired. We can entirely remove mentions of these products if their websites aren't available anyway.

For the NASA site, I've kept it for now for historical reference. Maybe the site has been decommissioned.

Also fixed a few linting errors.


bsmth (approved):
Great, LGTM! 👍🏻

PR #40444, opened 2025-07-19
Remove PaymentRequestEvent.instrumentKey

Josh-Cena:
This feature is already removed from BCD in https://github.com/mdn/browser-compat-data/pull/26253


bsmth (approved):
LGTM

PR #40445, opened 2025-07-19
Remove `@media aural`

Josh-Cena:
This feature has never been implemented. And it's irrelevant for any browser because it's for speech-only user agents, and screen readers are supposed to use `@media screen`.


bsmth (approved):
Thank you 👍🏻

Issue #14429, opened 2022-03-29
Issue with "Navigator.platform": Cross-browser API marked deprecated and Chromium-only alternative suggested


hsivonen:
MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform

#### What information was incorrect, unhelpful, or incomplete?

The page claims that `navigator.platform` is deprecated even though it's not marked deprecated [in the spec](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-platform-dev).

The page suggests using a Chromium-only alternative even though the page notes that it's neither supported by all major browsers nor accepted by a standards group.

#### Specific section or headline?

The very top of the document.

#### What did you expect to see?

Expected the page not to claim deprecation and expected not to suggest using a Chromium-only API instead.

#### Did you test this? If so, how?

Not applicable.

<details>
<summary>MDN Content page report details</summary>

* Folder: `en-us/web/api/navigator/platform`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/navigator/platform/index.md
* Last commit: https://github.com/mdn/content/commit/e752a8fe731fbdb9061e39a6e2b797a7ccb7e2b5
* Document last modified: 2022-01-19T17:57:58.000Z

</details>

eemeli:
Went down the rabbit hole a bit chasing the origin of this, just out of interest. The origin of the change is in browser-compat-data, where the marking-as-deprecated appears to originate in https://github.com/mdn/browser-compat-data/issues/6567#issuecomment-683547363 by @sideshowbarker, for which the subsequent change was confirmed by @foolip in https://github.com/mdn/browser-compat-data/pull/6612#discussion_r479986556.

sideshowbarker:
Thanks much @eemeli for investigating this and pinging me, and thanks @hsivonen for raising this.

I have a few questions and comments about this that’ll take me some time to write up, but for now I’ve opened https://github.com/mdn/content/pull/14452 with changes that I hope might have a chance of resolving it to everybody’s satisfaction. So please take a look at that PR and comment there — either if you have suggestions for specific refinements to it, or if you think it’s not actually at all the right solution to the issue.

sideshowbarker:
@hsivonen, in case you don’t find #14452 to be a satisfactory resolution to this issue, then some questions:

[...]

PR #40446, opened 2025-07-19
Remove deprecation status for navigator.platform

Pomax:
### Description

The WhatWG spec for this property (https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-platform-dev) does not have it marked as deprecated, so this deprecation was either erroneously added because it was never officially deprecated, or if at some point in time it was then the removal was missed when the official decision was made to un-deprecate it.

### Motivation

MDN is a trustworthy source, factually incorrect claims should get fixed.

https://github.com/mdn/content/issues/14429

### DIFF
--- a/files/en-us/web/api/navigator/platform/index.md
+++ b/files/en-us/web/api/navigator/platform/index.md
@@ -3,12 +3,8 @@ title: "Navigator: platform property"
 short-title: platform
 slug: Web/API/Navigator/platform
 page-type: web-api-instance-property
-status:
-  - deprecated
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Deprecation status is synced from browser compat data, so you need to remove it there and it will automatically disappear here. It was marked as deprecated in https://github.com/mdn/browser-compat-data/pull/17322, but I don't know if the reasoning given there fits our deprecation data schema, so you will probably have luck if you start a discussion over there.

PR #40450, opened 2025-07-20
Clarify container queries always use content box (fix #40306)

### LINKED ISSUE #40306: @container size queries aren't clear about which box size is used
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/@container

Size container descriptors

I think it's the same for all of the descriptors, but just to pick one as an example:

> The [block-size](https://developer.mozilla.org/en-US/docs/Web/CSS/block-size) of the container expressed as a [<length>](https://developer.mozilla.org/en-US/docs/Web/CSS/length) value.

Which `block-size` is that? From the text above, I'd assume that it's the one CSS usually uses, as specified by [`box-sizing`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing). However, https://drafts.csswg.org/css-conditional-5/#block-size says:

> The [block-size](https://drafts.csswg.org/css-conditional-5/#descdef-container-block-size) [container feature](https://drafts.csswg.org/css-conditional-5/#container-feature) queries the [size](https://drafts.csswg.org/css-sizing-3/#size) of the [query container](https://drafts.csswg.org/css-conditional-5/#query-container)’s [content box](https://drafts.csswg.org/css-box-4/#content-box) in the query container’s [block axis](https://drafts.csswg.org/css-writing-modes-4/#block-axis).

A clear description of which of the container's boxes is used for the size queries.

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/css/@container`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/@container
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/@container/index.md
* Last commit: https://github.com/mdn/content/commit/63cbf204323f117a2a80c7aa6273e50253ab9d07
* Document last modified: 2025-07-01T15:43:47.000Z

</details>

Shrinivassab:
### Description

Clarified that all container size queries measure the content box, ignoring `box-sizing`.

Fixes #40306


estelle (commented):
instead of making any of these changes, change line 130 to something like

```suggestion
The `<container-condition>` can include one or more boolean size queries, each within a set of parentheses. A size query includes a size descriptor, a value, and — depending on the descriptor — a comparison operator. The queries always measures the [content box](/en-US/docs/Web/CSS/box-edge#content-box) as the comparison. The syntax for including multiple conditions is the same as for [`@media`](/en-US/docs/Web/CSS/@media) size feature queries.
```

only use bolding when really necessary. When all cases are the same, favor adding a single note rather than repeating text.

Shrinivassab:
Sure. I revered the changes and added your suggestion.

estelle (approved):
Thanks! 🎉

PR #39141, opened 2025-04-15
Merge Transforming_XML_with_XSLT subarticles into itself

Josh-Cena:
These two pages are not discoverable via sidebar, and they are quite absurdly structured. This PR merges them into the main doc (since they are so thin anyway) and changed the DL to normal prose.

PR #38057, opened 2025-02-09
Move XSLTProcessor guide pages to DOM API

Josh-Cena:
Fix https://github.com/mdn/content/issues/35324.

In general the DOM guides are pretty terrible, but I don't want to split hairs over them. I also just consolidated the three pages into one.

PR #40453, opened 2025-07-20
Merge Transforming with XSLT guide in DOM API with the one in XSLT

### LINKED ISSUE #40223: Duplicated guide content for transforming with XSLT
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Transforming_with_XSLT

_No response_

It seems like there are two guide pages for transforming with XSLT.

- https://developer.mozilla.org/en-US/docs/Web/XML/XSLT/Guides/Transforming_XML_with_XSLT
- https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Transforming_with_XSLT

Probably they should be combined into the one in the Guides namespace. We can then put all the external links there. Other articles like [XSLTProcessor](https://developer.mozilla.org/en-US/docs/Web/API/XSLTProcessor) should then link just to that guide in the See Also section, so all the external links are in one place.

_No response_

_No response_

Josh-Cena:
Fix https://github.com/mdn/content/issues/40223.

Apparently I keep merging XSLT guides: https://github.com/mdn/content/pull/39141, https://github.com/mdn/content/pull/38057. After this one we have a single guide as opposed to 6 that talk about the same thing...

I'm ignoring all the intro in the DOM API guide. This information is all present in the XSLT guide article, and in a more elaborate fashion.

I'm moving the DOM API guide into the XSLT guide, because in fact these examples don't invoke JS at all, so it shouldn't belong in the API docs. It was only there because it was originally in `XSLTProcessor` and I was only looking to improve IA at that time.


dipikabh:
I'm taking a look at this now

dipikabh (approved):
Good idea to merge the two guides, thanks for taking this on @Josh-Cena!

I want to keep the focus of this review on combining the two guides. That part looks great, so leaving a +1.

At the same time, I'll make a note here of possible future improvements:
- As a follow up, we could look at cleaning up the guide. For example, the "What is XSLT?" section comes way later, after too much introductory text. Also, some of the text in the intro (starting from "Because a browser provides..." to the end of the section) could be moved to a separate section, something like "XSLT in browsers". But again, "What is XSLT?" should be moved up on the page.
- The intro text in the guide under DOM, the one we're deleting, is actually a much nicer writeup for intro. If and when we do the cleanup, we could borrow some of that text (we'll need to remember to check the history).

Josh-Cena:
Note that this page is supposed to be a usage guide and we do have a separate introductory landing page, so I agree this one could be shortened.

PR #40458, opened 2025-07-20
Clarify `<template>` permitted content

### LINKED ISSUE #38984: The `<template>` tag says it permitted content has "No restrictions", but that is not the case
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template

Technical summary > Permitted content

In the Technical Summary > Permitted content section, it says that there are "No restrictions" on a `<template>`'s permitted content. That, however, isn't true. Since `<template>`s are `DocumentFragments`, they obey document fragment parsing rules, which I presume means having no `<head>` nor `<body>` (and probably other restrictions, as well).

This is surprising if you expect to be able to put *any* arbitrary text/HTML content inside of `<template>`s, as I was recently trying to do, and was surprised when my `<head>` tag was stripped from it.

I believe the doc page hints at it, but I believe that the documentation could be a little bit more explicit in that `<template>`s do *not* accept arbitrary text/HTML content, but *only* HTML fragment content.

Additional, the "Permitted Content" section should be changed to make the `DocumentFragment` restrictions more evident.

Minimal example that works in both FF and Chrome:

```html
<!DOCTYPE html>
<body>
  <p>hello</p>
  <template>
    <head>
      <title>this is the title</title>
    </head>
  </template>
</body>
```

produces:

```html
<html>
  <head></head>
  <body>
    <p>hello</p>

    <template>
      <title>this is the title</title>
    </template>
  </body>
</html>
```

, stripping the `<head>` that was initially in my template.

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/html/element/template`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/element/template/index.md
* Last commit: https://github.com/mdn/content/commit/afaa8a0345fc2603e20503f00c7ab79f4911b55b
* Document last modified: 2025-01-10T07:26:01.000Z

</details>

Josh-Cena:
Fix https://github.com/mdn/content/issues/38984

### DIFF
--- a/files/en-us/web/html/reference/elements/template/index.md
+++ b/files/en-us/web/html/reference/elements/template/index.md
@@ -45,6 +45,10 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
 
 ## Usage notes
 
+This element has no permitted content, because everything nested inside it in the HTML source does not actually become the children of the `<template>` element. The {{domxref("Node.childNodes")}} property of the `<template>` element is always empty, and you can only access said nested content via the special {{domxref("HTMLTemplateElement.content", "content")}} property. However, if you call {{domxref("Node.appendChild()")}} or similar methods on the `<template>` element, then you would be inserting children into the `<template>` element itself, which is a violation of its content model and does not actually update the {{domxref("DocumentFragment")}} returned by the `content` property.
+
+Due to the way the `<template>` element is parsed, all `<html>`, `<head>`, and `<body>` opening and closing tags inside the template are syntax errors and are ignored by the parser, so `<template><head><title>Test</title></head></template>` is the same as `<template><title>Test</title></template>`.
+
 There are two main ways to use the `<template>` element.
 
 ### Template document fragment
@@ -319,7 +323,7 @@ Since `firstClone` is a `DocumentFragment`, only its children are added to `cont
     </tr>
     <tr>
       <th scope="row">Permitted content</th>
-      <td>No restrictions</td>
+      <td>Nothing (see <a href="#usage_notes">Usage notes</a>)</td>
     </tr>
     <tr>
       <th scope="row">Tag omission</th>


files/en-us/web/html/reference/elements/template/index.md
```diff
@@ -45,6 +45,10 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
 
 ## Usage notes
 
+This element has no permitted content, because everything nested inside it in the HTML source does not actually become the children of the `<template>` element. The {{domxref("Node.childNodes")}} property of the `<template>` element is always empty, and you can only access said nested content via the special {{domxref("HTMLTemplateElement.content", "content")}} property. However, if you call {{domxref("Node.appendChild()")}} or similar methods on the `<template>` element, then you would be inserting children into the `<template>` element itself, which is a violation of its content model and does not actually update the {{domxref("DocumentFragment")}} returned by the `content` property.
```
estelle, line 48:
```suggestion
The `<template>` element has no permitted content; nothing nested within the `<template>` element within the HTML source becomes a child of the `<template>` element. The {{domxref("Node.childNodes")}} property of the `<template>` element is always empty.

If you call {{domxref("Node.appendChild()")}} or similar methods on the `<template>` element, then you would be inserting children into the `<template>` element itself, which is a violation of its content model and does not actually update the {{domxref("DocumentFragment")}} returned by the `content` property.

Content nested within the `<template>` element can only be accessed via the special {{domxref("HTMLTemplateElement.content", "content")}} property. 
```
moved the last one down to put it before the part where you talk about stuff that is nested.

files/en-us/web/html/reference/elements/template/index.md
```diff
@@ -45,6 +45,10 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
+
+Due to the way the `<template>` element is parsed, all `<html>`, `<head>`, and `<body>` opening and closing tags inside the template are syntax errors and are ignored by the parser, so `<template><head><title>Test</title></head></template>` is the same as `<template><title>Test</title></template>`.
```
estelle, line 50:
```suggestion
Due to the way the `<template>` element is parsed, all `<html>`, `<head>`, and `<body>` opening and closing tags inside the `<template>` are syntax errors and are ignored by the parser, so `<template><head><title>Test</title></head></template>` is the same as `<template><title>Test</title></template>`.
```

estelle: Above, we say nothing can be nested, and here we are nesting, so, adding an intro before the "due" explaining this dissonance would be helpful

estelle (commented):
grammer and clarification feedback

files/en-us/web/html/reference/elements/template/index.md
Josh-Cena: > Above, we say nothing can be nested, and here we are nesting

Not sure I follow? The message above (and everything else on this page) is that `template` is supposed to nest (HTML-wise), but the HTML source does not map to DOM the way you expect.

Josh-Cena: There are two distinct things I'm saying here:

1. HTML source ≠ DOM; DOM children should always be empty and the HTML source maps to `content`. This is a runtime DOM requirement because the source code never generates a template with children.
2. The HTML source cannot contain head or body. This is a source code syntax requirement, because later you can insert a `head` yourself anyway if you really want and that wouldn't violate anything.

files/en-us/web/html/reference/elements/template/index.md
Josh-Cena: I believe your intention here is for the "can only be accessed via `content`" paragraph to be a segue into the next paragraph, but as I explained below that this is actually a non-sequitur. This paragraph conveys a single point which is that HTML source goes to `content` and `children` is unrelated.

estelle (approved):

PR #40459, opened 2025-07-20
Clarify how close-path command works

### LINKED ISSUE #37983: Close Path description incorrect (over simplified)
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Paths

Close Path

This command draws a straight line from the current position back to the first point of the path.

This statement is incorrect. Each M within a path starts a new subpath. The path will close to the start of the current subpath not the start of the whole path.

https://www.w3.org/TR/SVG2/paths.html#PathDataClosePathCommand

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/svg/tutorial/paths`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Paths
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/svg/tutorial/paths/index.md
* Last commit: https://github.com/mdn/content/commit/01b8471b84e1d157cbddbb3ffaf560a86b082070
* Document last modified: 2024-11-14T23:45:45.000Z

</details>

Josh-Cena:
Fixes https://github.com/mdn/content/issues/37983

### DIFF
--- a/files/en-us/web/svg/tutorials/svg_from_scratch/paths/index.md
+++ b/files/en-us/web/svg/tutorials/svg_from_scratch/paths/index.md
@@ -122,7 +122,7 @@ button.addEventListener("click", () => {
 
 {{ EmbedLiveSample('rectangle-lines', 100, 130) }}
 
-We can shorten the above path declaration a little bit by using the "Close Path" command, called with `Z`. This command draws a straight line from the current position back to the first point of the path. It is often placed at the end of a path node, although not always. There is no difference between the uppercase and lowercase command.
+We can shorten the above path declaration a little bit by using the "Close Path" command, called with `Z`. This command draws a straight line from the current position back to the first unclosed point (first point after the last `Z` command if there is one, or the first point in the path otherwise), and closes off the path with a line joint. It is often placed at the end of a path node, although not always. There is no difference between the uppercase and lowercase command.
 
 ```plain
 Z


files/en-us/web/svg/tutorials/svg_from_scratch/paths/index.md
```diff
@@ -122,7 +122,7 @@ button.addEventListener("click", () => {
 
 {{ EmbedLiveSample('rectangle-lines', 100, 130) }}
 
-We can shorten the above path declaration a little bit by using the "Close Path" command, called with `Z`. This command draws a straight line from the current position back to the first point of the path. It is often placed at the end of a path node, although not always. There is no difference between the uppercase and lowercase command.
+We can shorten the above path declaration a little bit by using the "Close Path" command, called with `Z`. This command draws a straight line from the current position back to the first unclosed point (first point after the last `Z` command if there is one, or the first point in the path otherwise), and closes off the path with a line joint. It is often placed at the end of a path node, although not always. There is no difference between the uppercase and lowercase command.
```
bsmth, line 125: Nice, makes sense, thanks. One comment here, I think the `plain` block can be moved inline like

```suggestion
We can shorten the above path declaration a little bit by using the "Close Path" command, called with `Z`. This command draws a straight line from the current position back to the first unclosed point (first point after the last `Z` command if there is one, or the first point in the path otherwise), and closes off the path with a line joint. It is often placed at the end of a path node, although not always. There is no difference between the uppercase and lowercase command (`z` and `Z` are equivalent).
```

bsmth (approved):
LGTM with one hint for making the path using `Z` appear right after the explanation.

files/en-us/web/svg/tutorials/svg_from_scratch/paths/index.md
Josh-Cena: Every path command has a plain block like this, so this is only for consistency (which happens to be trivial)

PR #40461, opened 2025-07-21
Typos in listing of the safe methods in Santiizer

hamishwillee:
The methods that support sanitizer were listed as both unsafe - setHTMLUnsafe and setHTMLUnsafe rather than, rather setHTML safe and setHTMLUnsafe. A typo.


sideshowbarker (approved):

sideshowbarker (approved):

captainbrosset:
See #40471 for another place where the same fix is needed.

PR #40463, opened 2025-07-21
`bufferedamountlow` is also fired with threshold zero

### LINKED ISSUE #29087: RTCDataChannel bufferedamountlow event not sent if the threshold is 0?
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/bufferedamountlow_event

The first section

The intro paragraph states:

> bufferedamountlow events aren't sent if bufferedAmountLowThreshold is 0.

This is not mentioned in [the spec](https://www.w3.org/TR/webrtc), nor does it seem to hold through experimental observation.

It was added in a generic ["initial en-us content checkin"](https://github.com/mdn/content/commit/cbe151a06d6e5b4d1fbb46081bd16e69ef4c1630) commit (warning, very large diff) so I'm not sure of it's provenance.

Where did this come from?

This sentence probably shouldn't be there.

The spec - https://www.w3.org/TR/webrtc

A demo - https://codepen.io/achingbrain/pen/qBLjZow?editors=0012

In the console you should see something like:

```
"peer1 opened datachannel"
"peer1 buffered amount" 65536 "waiting for it to be zero"
"peer2 got dataChannel" "test"
"peer1 buffered amount low event fired with threshold set to" 0
"peer1 buffered amount now" 0
```

Works in Chrome & Firefox

_No response_

Josh-Cena:
Fix https://github.com/mdn/content/issues/29087. Since this seems true in both implementations and standard, and we can't tell why it was that way originally, it's better to stay handwavy and not call out `0` as a special case.

### DIFF
--- a/files/en-us/web/api/rtcdatachannel/bufferedamountlow_event/index.md
+++ b/files/en-us/web/api/rtcdatachannel/bufferedamountlow_event/index.md
@@ -8,7 +8,7 @@ browser-compat: api.RTCDataChannel.bufferedamountlow_event
 
 {{APIRef("WebRTC")}}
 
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Nice, thanks 👍🏻

PR #40468, opened 2025-07-21
Add note on CSP sandbox causing  to iframe sandbox section

### LINKED ISSUE #40093: Mention that Content-Security-Policy: sandbox can lead to Origin: null
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin

Where it discusses the cases when it can be null

It doesn't mention `Content-Security-Policy: sandbox` leads to `Origin: null`

When a page is served with the `Content-Security-Policy: sandbox` header without `allow-same-origin`, then the `Origin` header will be `null` for any outgoing requests that the page makes.

That should be mentioned on this page.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/sandbox#allow-same-origin

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/http/reference/headers/origin`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/reference/headers/origin/index.md

</details>

krikera:
## Description

Added a new bullet point to the list of cases where the `Origin` header may be `null`.  
It covers the scenario where a page is served with a `Content-Security-Policy: sandbox` header **without** the `allow-same-origin` directive.

Fixes #40093

### DIFF
--- a/files/en-us/web/http/reference/headers/origin/index.md
+++ b/files/en-us/web/http/reference/headers/origin/index.md
@@ -63,6 +63,7 @@ The `Origin` header value may be `null` in a number of cases, including (non-exh
 - Documents created programmatically using {{domxref("DOMImplementation.createDocument", "createDocument()")}}, generated from a `data:` URL, or that do not have a creator browsing context.
 - Redirects across origins.
 - {{HTMLElement("iframe", "iframes")}} with a sandbox attribute that doesn't contain the value `allow-same-origin`.
+- Documents served with the {{HTTPHeader("Content-Security-Policy")}} `sandbox` directive that doesn't include `allow-same-origin`.
 - Responses that are network errors.
 - {{HTTPHeader("Referrer-Policy")}} set to `no-referrer` for non-`cors` request modes (e.g., basic form posts).
 
--- a/files/en-us/web/http/reference/headers/origin/index.md
+++ b/files/en-us/web/http/reference/headers/origin/index.md
@@ -62,8 +62,8 @@ The `Origin` header value may be `null` in a number of cases, including (non-exh
 - Cross-origin images and media data, including that in {{HTMLElement("img")}}, {{HTMLElement("video")}} and {{HTMLElement("audio")}} elements.
 - Documents created programmatically using {{domxref("DOMImplementation.createDocument", "createDocument()")}}, generated from a `data:` URL, or that do not have a creator browsing context.
 - Redirects across origins.
-- {{HTMLElement("iframe", "iframes")}} with a sandbox attribute that doesn't contain the value `allow-same-origin`.
-- Documents served with the {{HTTPHeader("Content-Security-Policy")}} `sandbox` directive that doesn't include `allow-same-origin`.
+- Documents served with the {{HTTPHeader("Content-Security-Policy")}} `sandbox` directive that don't include `allow-same-origin`.
+- {{HTMLElement("iframe", "iframes")}} with a sandbox attribute that don't contain the value `allow-same-origin`.
 - Responses that are network errors.
 - {{HTTPHeader("Referrer-Policy")}} set to `no-referrer` for non-`cors` request modes (e.g., basic form posts).
 


files/en-us/web/http/reference/headers/origin/index.md
```diff
@@ -63,6 +63,7 @@ The `Origin` header value may be `null` in a number of cases, including (non-exh
 - Documents created programmatically using {{domxref("DOMImplementation.createDocument", "createDocument()")}}, generated from a `data:` URL, or that do not have a creator browsing context.
 - Redirects across origins.
 - {{HTMLElement("iframe", "iframes")}} with a sandbox attribute that doesn't contain the value `allow-same-origin`.
+- Documents served with the {{HTTPHeader("Content-Security-Policy")}} `sandbox` directive that doesn't include `allow-same-origin`.
```
hamishwillee: Don't, not doesn't. I've also put the new item first, because if the document doesn't have the directive, then it doesn't matter if the frame does (I know it doesn't matter here).

```suggestion
- Documents served with the {{HTTPHeader("Content-Security-Policy")}} `sandbox` directive that don't include `allow-same-origin`.
- {{HTMLElement("iframe", "iframes")}} with a sandbox attribute that don't contain the value `allow-same-origin`.
```

hamishwillee (approved):
Thanks very much!

files/en-us/web/http/reference/headers/origin/index.md
```diff
@@ -62,7 +62,8 @@ The `Origin` header value may be `null` in a number of cases, including (non-exh
 - Cross-origin images and media data, including that in {{HTMLElement("img")}}, {{HTMLElement("video")}} and {{HTMLElement("audio")}} elements.
 - Documents created programmatically using {{domxref("DOMImplementation.createDocument", "createDocument()")}}, generated from a `data:` URL, or that do not have a creator browsing context.
 - Redirects across origins.
-- {{HTMLElement("iframe", "iframes")}} with a sandbox attribute that doesn't contain the value `allow-same-origin`.
+- Documents served with the {{HTTPHeader("Content-Security-Policy")}} `sandbox` directive that don't include `allow-same-origin`.
+- {{HTMLElement("iframe", "iframes")}} with a sandbox attribute that don't contain the value `allow-same-origin`.
```
Josh-Cena, line 66: Shouldn't it be "that doesn't" since "that" refers to the singular "directive"?

hamishwillee: The subject is the iframes - my understanding is that the modifier "with a sandbox attribute" does not change that.

But in honestly, I didn't think about it at all - the other way "feels wrong".

Josh-Cena: The iframe doesn't "contain the value"; the attribute "contains the value". It should read "iframes (with a sandbox attribute (that doesn't contain the value))", not "iframes (with a sandbox attribute) (that don't contain the value)" (and in this case the latter should use "and which don't contain the value" anyway)

hamishwillee: In English, the iframe contains the attribute so it also contains the `allow-same-attribute`. This isn't ambiguous, but I guess our job is to be very pedantic.

```
- Documents served with the {{HTTPHeader("Content-Security-Policy")}} `sandbox` directive, where the directive value doesn't include `allow-same-origin`.
- {{HTMLElement("iframe", "iframes")}} with a sandbox attribute where the attribute doesn't contain the value `allow-same-origin`.
```

hamishwillee: https://github.com/mdn/content/pull/40493

BVRmart (commented):
IMG-20250722-WA0112

PR #40469, opened 2025-07-21
Correct `navigate` addeventlistener

tunetheweb:
Missing a `navigation` from the `addEventListener` example, without which it doesn't work. You can see the rest of the examples in the doc use the correct syntax.

### Motivation

It's wrong :-)

### Additional details

You can test it with this SQL in the console of any page in Chrome:

```js
// This will log
navigation.addEventListener("navigate", (event) => {console.log('navigation event:',event)});
// This won't log
addEventListener("navigate", (event) => {console.log('window event',event)});
// Neither will this
window.addEventListener("navigate", (event) => {console.log('explicit window event',event)})
// Nor this
document.addEventListener("navigate", (event) => {console.log('document event',event)})

// Push a history to trigger above
history.pushState({}, "", "/barrry");
```

FYI: @mattzeunert

### DIFF
--- a/files/en-us/web/api/navigation/navigate_event/index.md
+++ b/files/en-us/web/api/navigation/navigate_event/index.md
@@ -17,7 +17,7 @@ The **`navigate`** event of the {{domxref("Navigation")}} interface is fired whe
 Use the event name in methods like {{domxref("EventTarget.addEventListener", "addEventListener()")}}, or set an event handler property.
 
 ```js-nolint
-addEventListener("navigate", (event) => { })
+navigation.addEventListener("navigate", (event) => { })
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Hi, this is the general MDN syntax convention: we don't add the preceding object for any method invocation. See for example: https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/abort_event, https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event. The page template: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types/API_event_subpage_template

PR #40489, opened 2025-07-22
CSP sandbox allow-same-origin - clarify

### LINKED ISSUE #40094: The Content-Security-Policy: sandbox allow-same-origin section has some ambiguity
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/sandbox

allow-same-origin

It says

>[allow-same-origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/sandbox#allow-same-origin)
>If this token is not used, the resource is treated as being from a special origin that always fails the [same-origin policy](https://developer.mozilla.org/en-US/docs/Glossary/Same-origin_policy) (potentially preventing access to [data storage/cookies](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs).

3 suggestions:

# directionality ambiguity

That's ambiguous. It could refer to 3 things:

* requests made by this page to other pages fail the same-origin policy
* requests made to this page by other pages fail the same-origin policy
* both of the above

In reality, it's only the 1st bullet point that applies. The `Content-Security-Policy: sandbox` doesn't impact when this page is requested by other pages.

# requests

It mentions "data storage/cookies and some JavaScript APIs", but doesn't explicitly mention requests to other sites (fetch). I think it would be good to also mention that.

# mention null

It doesn't mention that the `Origin` header will be `null`. I think it should mention that, and link to the `Origin` header page. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin

see above

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/http/reference/headers/content-security-policy/sandbox`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/sandbox
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
* Last commit: https://github.com/mdn/content/commit/40699574190b48488c35d37c89d631e15907c94e
* Document last modified: 2025-05-26T02:16:43.000Z

</details>

hamishwillee:
If found the description of the CSP sandbox value `allow-same-origin` to be a little confusing. This rewords to match the pattern of the other values, and adds the note that the `Origin` header will be null if this is not set.

This fell out of reviewing https://github.com/mdn/content/issues/40093

Fixes #40094

### DIFF
--- a/files/en-us/glossary/origin/index.md
+++ b/files/en-us/glossary/origin/index.md
@@ -9,6 +9,20 @@ Web content's **origin** is defined by the _scheme_ (protocol), _hostname_ (doma
 
 Some operations are restricted to same-origin content, and this restriction can be lifted using {{Glossary("CORS")}}.
 
+## Opaque origin
+
+An opaque origin is a special type of "internal" origin that obscures the true origin of a resource (opaque origins are always serialized as `null`), and ensures that it never matches as same-origin with other resources — including those from other opaque origins.
+
+Opaque origins are applied in cases where the true origin of a resource is sensitive, cannot be safely used for security checks, or does not exist.
+A resource with an opaque origin will have its {{httpheader("Origin")}} HTTP header in requests set to [`null`](/en-US/docs/Web/HTTP/Reference/Headers/Origin#null).
+It will also fail same-origin checks with any other resource, and hence be restricted to only those operations available to cross-origin resources.
+
+Common cases where opaque origins are used include:
+
+- A document within an iframe that has the [sandbox](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/sandbox) attribute set, and does not include the `allow-same-origin` flag.
+- `file:` URLs are usually treated as opaque origins so that files on they file system cannot read each other.
+- Documents created programatically using APIs like https://developer.mozilla.org/en-US/docs/Web/API/ {{domxref("DOMImplementation.createDocument()")}}.
+
 ## Examples
 
 These are same origin because they have the same scheme (`http`) and hostname (`example.com`), and the different file path does not matter:
--- a/files/en-us/glossary/origin/index.md
+++ b/files/en-us/glossary/origin/index.md
@@ -11,7 +11,7 @@ Some operations are restricted to same-origin content, and this restriction can 
 
 ## Opaque origin
 
-An opaque origin is a special type of "internal" origin that obscures the true origin of a resource (opaque origins are always serialized as `null`), and ensures that it never matches as same-origin with other resources — including those from other opaque origins.
+An opaque origin is a special type of browser-internal value that obscures the true origin of a resource (opaque origins are always serialized as `null`). They are used by the browser to ensure resource isolation as they are never considered equal to any other origin — including other opaque origins.
 
 Opaque origins are applied in cases where the true origin of a resource is sensitive, cannot be safely used for security checks, or does not exist.
 A resource with an opaque origin will have its {{httpheader("Origin")}} HTTP header in requests set to [`null`](/en-US/docs/Web/HTTP/Reference/Headers/Origin#null).
--- a/files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
+++ b/files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
@@ -59,7 +59,6 @@ where `<value>` can optionally be one of the following values:
   - : Allows a sandboxed resource to retain its {{Glossary("origin")}}.
     A sandboxed resource is otherwise treated as being from an opaque origin, which ensures that it will always fail {{Glossary("same-origin policy")}} checks, and hence cannot access [`localstorage` and `document.cookie`](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs.
     The {{httpheader("Origin")}} of sandboxed resources without the `allow-same-origin` keyword is `null`.
-
 - `allow-scripts`
   - : Allows the page to run scripts (but not create pop-up windows). If this keyword is not used, this operation is not allowed.
 - `allow-storage-access-by-user-activation` {{experimental_inline}}
--- a/files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
+++ b/files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
@@ -56,9 +56,8 @@ where `<value>` can optionally be one of the following values:
 - `allow-presentation`
   - : Allows embedders to have control over whether an iframe can start a [presentation session](/en-US/docs/Web/API/PresentationRequest).
 - `allow-same-origin`
-  - : Allows a same-origin sandboxed resource to be treated as same-origin, reducing the restrictions normally applied to a sandbox.
-
-    If this token is not used, the resource is treated as being from a special origin that always fails the {{Glossary("same-origin policy")}} (potentially preventing access to [data storage/cookies](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs).
+  - : Allows a sandboxed resource to retain its {{Glossary("origin")}}.
+    A sandboxed resource is otherwise treated as being from an opaque origin, which ensures that it will always fail {{Glossary("same-origin policy")}} checks, and hence cannot access [`localstorage` and `document.cookie`](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs.
     The {{httpheader("Origin")}} of sandboxed resources without the `allow-same-origin` keyword is `null`.
 
 - `allow-scripts`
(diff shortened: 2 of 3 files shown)


Josh-Cena:
1. Perhaps you could use this chance to fix https://github.com/mdn/content/issues/40094?
2. #40093 has a grammatical error; it should be "that doesn't".

hamishwillee:
I already fixed the second in another issue. The first outlines a number of problems, the only one that is valid (Origin) is already fixed. I've added "Fixes" above for that. Nothing else to do.

files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
```diff
@@ -56,7 +56,11 @@ where `<value>` can optionally be one of the following values:
 - `allow-presentation`
   - : Allows embedders to have control over whether an iframe can start a [presentation session](/en-US/docs/Web/API/PresentationRequest).
 - `allow-same-origin`
-  - : If this token is not used, the resource is treated as being from a special origin that always fails the {{Glossary("same-origin policy")}} (potentially preventing access to [data storage/cookies](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs).
+  - : Allows a same-origin sandboxed resource to be treated as same-origin, reducing the restrictions normally applied to a sandbox.
+
+    If this token is not used, the resource is treated as being from a special origin that always fails the {{Glossary("same-origin policy")}} (potentially preventing access to [data storage/cookies](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs).
```
bsmth: Tough one to summarize, is this any better?

```suggestion
  - : Allows a sandboxed resource to maintain its {{Glossary("origin")}}.
    A sandboxed resource is otherwise treated as being from an origin that always fails {{Glossary("same-origin policy")}} checks, which prevents access to [`localstorage` and `document.cookie`](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs.
```

>causes the content to be treated as being from its real origin instead of forcing it into an [opaque origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque);

* https://html.spec.whatwg.org/multipage/browsers.html#sandboxed-origin-browsing-context-flag

files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
```diff
@@ -56,7 +56,11 @@ where `<value>` can optionally be one of the following values:
...
+    The {{httpheader("Origin")}} of such a request would be `null`.
```
bsmth:
```suggestion
    The {{httpheader("Origin")}} of sandboxed resources without the `allow-same-origin` keyword is `null`.
```

files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
hamishwillee: Better :-). I  added mention of opaque. I also think retain is perhaps better than maintain. How about this @bsmth ?

```suggestion
  - : Allows a sandboxed resource to retain its {{Glossary("origin")}}.
    A sandboxed resource is otherwise treated as being from an opaque origin, which ensures that it will always fail {{Glossary("same-origin policy")}} checks, and hence cannot access [`localstorage` and `document.cookie`](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs.
```

files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
```diff
@@ -56,7 +56,10 @@ where `<value>` can optionally be one of the following values:
 - `allow-presentation`
   - : Allows embedders to have control over whether an iframe can start a [presentation session](/en-US/docs/Web/API/PresentationRequest).
 - `allow-same-origin`
-  - : If this token is not used, the resource is treated as being from a special origin that always fails the {{Glossary("same-origin policy")}} (potentially preventing access to [data storage/cookies](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs).
+  - : Allows a sandboxed resource to retain its {{Glossary("origin")}}.
+    A sandboxed resource is otherwise treated as being from an opaque origin, which ensures that it will always fail {{Glossary("same-origin policy")}} checks, and hence cannot access [`localstorage` and `document.cookie`](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs.
+    The {{httpheader("Origin")}} of sandboxed resources without the `allow-same-origin` keyword is `null`.
+
```
hamishwillee:
```suggestion
```

hamishwillee:
@bsmth Thought about this some more and accepted my one-word tweak of your suggestion here https://github.com/mdn/content/pull/40489#discussion_r2230016794

I was a little uncomfortable initially because the name `allow-same-origin` implies that this only allows the origin to be retained for same-origin content - this change makes it clear that cross-origin pages also get to keep their origin. 
AIUI that is irrelevant though - w.r.t. the embedder the frame is still cross origin, and while it could in theory access its own database etc, it can't unless you allow scripts as well - and even if you do that it is still cross origin to the embedder and isolated from it..

files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
```diff
@@ -56,7 +56,9 @@ where `<value>` can optionally be one of the following values:
 - `allow-presentation`
   - : Allows embedders to have control over whether an iframe can start a [presentation session](/en-US/docs/Web/API/PresentationRequest).
 - `allow-same-origin`
-  - : If this token is not used, the resource is treated as being from a special origin that always fails the {{Glossary("same-origin policy")}} (potentially preventing access to [data storage/cookies](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs).
+  - : Allows a sandboxed resource to retain its {{Glossary("origin")}}.
+    A sandboxed resource is otherwise treated as being from an opaque origin, which ensures that it will always fail {{Glossary("same-origin policy")}} checks, and hence cannot access [`localstorage` and `document.cookie`](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs.
```
bsmth: I think we don't define "opaque origin" anywhere, maybe there's a follow-up opportunity on the `origin` glossary entry?

```suggestion
    A sandboxed resource is otherwise treated as being from an _opaque origin_, which ensures that it will always fail {{Glossary("same-origin policy")}} checks, and hence cannot access [`localstorage` and `document.cookie`](/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs.
```

When searching, I find:

```
./files/en-us/web/api/credentialscontainer/get/index.md:    - The calling origin is an [opaque origin](/en-US/docs/Web/HTTP/Reference/Headers/Origin#null).
```

Any thoughts on linking to https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin#null?

bsmth (approved):
Looks good thanks! One opportunity for linking "opaque origin", comment there to consider 👍🏻

files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
bsmth: This is better, thanks!

files/en-us/web/http/reference/headers/content-security-policy/sandbox/index.md
hamishwillee: I'll create a glossary entry and you can re-review.

files/en-us/glossary/origin/index.md
```diff
@@ -9,6 +9,20 @@ Web content's **origin** is defined by the _scheme_ (protocol), _hostname_ (doma
 
 Some operations are restricted to same-origin content, and this restriction can be lifted using {{Glossary("CORS")}}.
 
+## Opaque origin
```
hamishwillee, line 12: @bsmth OK, I ended up adding to the origin glossary. I think that makes sense. Can split out if you want.

bsmth: Super, thanks a lot!

files/en-us/glossary/origin/index.md
```diff
@@ -9,6 +9,20 @@ Web content's **origin** is defined by the _scheme_ (protocol), _hostname_ (doma
...
+
+An opaque origin is a special type of "internal" origin that obscures the true origin of a resource (opaque origins are always serialized as `null`), and ensures that it never matches as same-origin with other resources — including those from other opaque origins.
```
bsmth: What do you think of this?

```suggestion
An opaque origin is a special type of browser-internal value that obscures the true origin of a resource (opaque origins are always serialized as `null`). They are used by the browser to ensure resource isolation as they are never considered equal to any other origin — including other opaque origins.
```

bsmth (approved):
Thanks a lot, Hamish. One wording suggestion, but I think we're good to go when you've had a look 👀

files/en-us/glossary/origin/index.md
hamishwillee: Sounds like the spec, but I'm OK with that :-)

PR #40491, opened 2025-07-22
FF142 Selection.getComposedRanges() - Relnote/Expr

hamishwillee:
FF142 adds support for [`Selection.getComposedRanges()`](https://developer.mozilla.org/en-US/docs/Web/API/Selection/getComposedRanges) in https://bugzilla.mozilla.org/show_bug.cgi?id=1903870 (along with modifying [`Selection.setBaseAndExtent()`](https://developer.mozilla.org/en-US/docs/Web/API/Selection/setBaseAndExtent),[`Selection.collapse()`](https://developer.mozilla.org/en-US/docs/Web/API/Selection/collapse), [`Selection.extend()`](https://developer.mozilla.org/en-US/docs/Web/API/Selection/extend) to accept nodes inside a shadow root)

This adds the release note and removes the information from the experimental features.

Related docs work can be tracked in #40475

### DIFF
--- a/files/en-us/mozilla/firefox/experimental_features/index.md
+++ b/files/en-us/mozilla/firefox/experimental_features/index.md
@@ -563,28 +563,6 @@ It is disabled by default on all builds [Firefox bug 1750902](https://bugzil.la/
 - `dom.vr.enabled`
   - : Set to `true` to enable.
 
-### HTML DOM API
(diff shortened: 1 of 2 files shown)


dipikabh (approved):
Excellent, thank you!

PR #40505, opened 2025-07-22
Replace archive link with actual link to the blog post

gurjeet:
### Description

Ideally, archive links should be used only if the original URL/domain does not
work anymore, or has changed content that no longer matches the intent here.
That is not the case for this blog post, hence replacing the archive link with
actual URL feels like the right thing to do.

### Motivation

Doing the right thing on the internet :-)


Josh-Cena (approved):
Nice thanks. It has been a web archive link since day 1; probably the site was down at that time

PR #40506, opened 2025-07-23
Better explanation for why `innerHTML` is not a good substitute for `textContent`

lionel-rowe:
### Description

Better explanation for why `innerHTML` is not a good substitute for `textContent`.

### Motivation

Current version mentions performance before any other reason, which isn't a primary concern. For any use case where performance had a noticeable impact, correctness and vulnerability to XSS would likely be a much greater concern; and for any use case trivial enough that correctness and vulnerability weren't practically impacted (e.g. setting `button.innerHTML = "Loading..."`), performance impact would be negligible.

### DIFF
--- a/files/en-us/web/api/node/textcontent/index.md
+++ b/files/en-us/web/api/node/textcontent/index.md
@@ -58,10 +58,7 @@ important differences:
 
 ### Differences from innerHTML
 
-{{domxref("Element.innerHTML")}} returns HTML, as its name indicates. Sometimes people
-use `innerHTML` to retrieve or write text inside an element, but it is unsuited to
-this purpose, as it deals with raw HTML rather than plain text and can be succeptable
-to {{glossary("Cross-site_scripting", "XSS attacks")}}. Use `textContent` instead.
+{{domxref("Element.innerHTML")}} gets or sets HTML, as its name indicates. Sometimes people use `innerHTML` to get or set text inside an element, but it is unsuited to this purpose, as it deals with raw HTML rather than plain text and can be susceptible to {{glossary("Cross-site_scripting", "XSS attacks")}}. Use `textContent` instead for this purpose.
 
 ## Examples
 
--- a/files/en-us/web/api/node/textcontent/index.md
+++ b/files/en-us/web/api/node/textcontent/index.md
@@ -59,11 +59,9 @@ important differences:
 ### Differences from innerHTML
 
 {{domxref("Element.innerHTML")}} returns HTML, as its name indicates. Sometimes people
-use `innerHTML` to retrieve or write text inside an element, but
-`textContent` has better performance because its value is not parsed as
-HTML.
-
-Moreover, using `textContent` can prevent {{glossary("Cross-site_scripting", "XSS attacks")}}.
+use `innerHTML` to retrieve or write text inside an element, but it is unsuited to
+this purpose, as it deals with raw HTML rather than plain text and can be succeptable
+to {{glossary("Cross-site_scripting", "XSS attacks")}}. Use `textContent` instead.
 
 ## Examples
 
--- a/files/en-us/web/api/node/textcontent/index.md
+++ b/files/en-us/web/api/node/textcontent/index.md
@@ -58,7 +58,7 @@ important differences:
 
 ### Differences from innerHTML
 
-{{domxref("Element.innerHTML")}} gets or sets HTML, as its name indicates. Sometimes people use `innerHTML` to get or set text inside an element, but it is unsuited to this purpose, as it deals with raw HTML rather than plain text and can be susceptible to {{glossary("Cross-site_scripting", "XSS attacks")}}. Use `textContent` instead for this purpose.
+{{domxref("Element.innerHTML")}} gets or sets HTML, as its name indicates. We advise against using `innerHTML` to get or set text inside an element because it deals with raw HTML rather than plain text and can be susceptible to {{glossary("Cross-site_scripting", "XSS attacks")}}. Even if you are sure that the text never contains HTML syntax, it is still less semantic and slower because it needs to invoke the HTML parser.
 
 ## Examples
 


files/en-us/web/api/node/textcontent/index.md
```diff
@@ -59,11 +59,9 @@ important differences:
...
 
 {{domxref("Element.innerHTML")}} returns HTML, as its name indicates. Sometimes people
-use `innerHTML` to retrieve or write text inside an element, but
-`textContent` has better performance because its value is not parsed as
-HTML.
-
-Moreover, using `textContent` can prevent {{glossary("Cross-site_scripting", "XSS attacks")}}.
+use `innerHTML` to retrieve or write text inside an element, but it is unsuited to
+this purpose, as it deals with raw HTML rather than plain text and can be succeptable
+to {{glossary("Cross-site_scripting", "XSS attacks")}}. Use `textContent` instead.
```
Josh-Cena: 1. Remove the line wrapping
2. I would still suggest adding something about performance. Something like "Even if you are completely sure that there's no HTML injection risk, using `innerHTML` is slower because it invokes the HTML parser even when there's no HTML markup to insert".
3. Spelling: "susceptible"
4. "Use `textContent` instead." seems too unnuanced because we do have the case of HTML markup insertion (which can be safe). How about "Use `textContent` for inserting or reading plain text, and DOM methods such as {{domxref("document.createElement()")}} and {{domxref("Node.appendChild()")}} for manipulating DOM".

lionel-rowe: Fixed 1, 3, and 4. I don't agree that performance should be mentioned, as I can't think of a case where it would be a worthwhile consideration over other concerns.

For example, even if multiplication is slower than addition, you would never cite performance as a reason to use addition over multiplication. Correctness is the only concern worth mentioning in this case, even if they happen to give the same result in certain cases (`1*2*3 == 1+2+3`).

Josh-Cena:
People often do `innerHTML = "<b>foo</b>"`, or `innerHTML = "Hello"`, which has zero XSS risk. The only difference is it's slower. The truth is, people often try to confront our content, and they believe they have use cases that could overrule our suggestion, so we need to be bulletproof and preempt their use case and say why it's still bad.

files/en-us/web/api/node/textcontent/index.md
lionel-rowe: @Josh-Cena replying inline here as gh doesn't allow threaded replies to top-level comments

> People often do `innerHTML = "<b>foo</b>"`, or `innerHTML = "Hello"`, which has zero XSS risk. The only difference is it's slower. The truth is, people often try to confront our content, and they believe they have use cases that could overrule our suggestion, so we need to be bulletproof and preempt their use case and say why it's still bad.

The first of those is a legitimate use case for `innerHTML`, as the content is intended to be parsed and rendered as HTML, and setting `textContent` would not give the intended result. Performance characteristics aren't relevant here.

The second is semantically wrong (assuming `"Hello"` is just supposed to mean the literal text "Hello", rather than e.g. being retrieved from a database of trusted HTML fragments, some of which happen not to contain internal tags). It's the equivalent of using `1*2*3` to mean `1+2+3` just because they happened to give the same result with the given operands. `*` being faster or slower than `+` isn't relevant.

Josh-Cena: For the first one, my suggestion as outlined above is to use DOM methods. The reason is two-fold:(1) later you may decide to parameterize your markup and you need to refactor it anyway (2) even using multiple DOM methods in JS is still faster than invoking the HTML parser. In this example:

```js
const x = document.createElement("div")
x.innerHTML = "<b>foo</b>"
```

```js
const x = document.createElement("div")
const b = document.createElement("b")
b.textContent = "foo"
x.appendChild(b)
```

I benchmarked this in Chrome and the second one is 35% faster.

> Performance characteristics aren't relevant here.

Performance is always relevant. If you have two methods that do the same thing, always pick the faster one. MDN is opinionated and we need to give authoritative guidelines about preferences whenever there's a choice for our readers.

> The second is semantically wrong

People do do that, so we need to dissuade them. You can of course explicitly write that in content ("Don't use `innerHTML` where `textContent` would do the same thing because it's semantically wrong"), but I'm just collecting more reasons because people do argue against our suggestions and we need to really buff our arguments. Semantics are subjective and may come off as our own style guide, so we need to pick some objective reasons.

lionel-rowe: > Performance is always relevant. If you have two methods that do the same thing, always pick the faster one. MDN is opinionated and we need to give authoritative guidelines about preferences whenever there's a choice for our readers.

OK, but then the comparison isn't between `innerHTML` and `textContent` any more, it's between `innerHTML` and manually calling `createElement` etc. That also might not fit the use case, e.g. if it's trusted content from a CMS stored as serialized HTML — it's not possible to implement a full HTML parser in JS that's more performant than the browser's built-in one.

> Semantics are subjective

I don't think they are, at least in this case:
* Does the string represent HTML content? If yes, use `innerHTML` (or other relevant APIs like `setHTML`, `setHTMLUnsafe`, `insertAdjacentHTML`, `DOMParser`, etc. depending on specifics of use case)
* Otherwise, if the string represents plain text, use `textContent` (or rarely in very specific use cases, `innerText`)

That's not a style guide recommendation like "prefer tabs over spaces", that's a factual statement about how those setter methods interpret and handle their arguments.

Josh-Cena: If two things behave exactly the same, then which one you pick is always a stylistic decision. Here what you are proposing reads like a style guide for me, and I don't see everyone necessarily agreeing with it. For example, I could easily list a few reasons to use `innerHTML`:

- Consistency: for example, if you later want to add bold or code formatting, you can easily do so without changing other stuff. This is a really big thing in making style guides: use the most permissive thing in the first place so future changes are as small and localized as possible.
- HTML doesn't have to contain markup. Plain text is still valid HTML.

The only things you can ground in reality are: which one is shorter, and which one gives a better UX. Here we need to argue for the latter.

I agree about the CMS case, and that would remain one of the few valid use cases of `innerHTML` (similar to cases where you need to use `dangerouslySetInneerHTML` in React). I'm strictly talking about inserting predictable markup/DOM structure.

files/en-us/web/api/node/textcontent/index.md
```diff
@@ -58,12 +58,7 @@ important differences:
 
 ### Differences from innerHTML
 
-{{domxref("Element.innerHTML")}} returns HTML, as its name indicates. Sometimes people
-use `innerHTML` to retrieve or write text inside an element, but
-`textContent` has better performance because its value is not parsed as
-HTML.
-
-Moreover, using `textContent` can prevent {{glossary("Cross-site_scripting", "XSS attacks")}}.
+{{domxref("Element.innerHTML")}} gets or sets HTML, as its name indicates. Sometimes people use `innerHTML` to get or set text inside an element, but it is unsuited to this purpose, as it deals with raw HTML rather than plain text and can be susceptible to {{glossary("Cross-site_scripting", "XSS attacks")}}. Use `textContent` instead for this purpose.
```
Josh-Cena:
```suggestion
{{domxref("Element.innerHTML")}} gets or sets HTML, as its name indicates. We advise against using `innerHTML` to get or set text inside an element because it deals with raw HTML rather than plain text and can be susceptible to {{glossary("Cross-site_scripting", "XSS attacks")}}. Even if you are sure that the text never contains HTML syntax, it is still less semantic and slower because it needs to invoke the HTML parser.
```

Josh-Cena (commented):
I just realized we are on the `textContent` page, which means we don't need to talk about inserting DOM, just plain text. I still want us to talk about performance though.

Josh-Cena (approved):
Nice, thank you!

PR #40507, opened 2025-07-23
Add a browser compat section for Tk header

Josh-Cena:
This is the only case where we have a deprecated/experimental header but not browser compat section. A section would make the anchor in the banner valid, and also explain why there is no compat.


bsmth (approved):
Good, tnx 👍🏻

PR #40516, opened 2025-07-24
doc(api): IntersectionObserver rootMargin option value

### LINKED ISSUE #40515: IntersectionObserver rootMargin option can only be in pixels or percentage
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver#rootmargin

Instance property value

On https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver clearly mentioned that `rootMargin` should be `can be only expressed in pixels (px) or as a percentage (%).`
I got hit by exception in my production build while following this page https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver#rootmargin.

Image

Correct instance value as `can be only expressed in pixels (px) or as a percentage (%).`

Ref: https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver

Ref exception: Image

ShubhamOulkar:
closes #40515


Josh-Cena (changes requested):
This should be mentioned in https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#the_intersection_root_and_root_margin and https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver instead, because the property merely reflects the input value.

ShubhamOulkar:
> This should be mentioned in https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#the_intersection_root_and_root_margin and https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver instead, because the property merely reflects the input value.

@Josh-Cena thank you for pointing out, I changed instance property page. Actually, the page I visited was [this](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver#rootmargin). Issue is updated.

I will do necessary changes on following sections :

https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#the_intersection_root_and_root_margin

https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver#rootmargin

Josh-Cena (approved):
Nice, thank you! Apparently this is a spec bug: https://github.com/w3c/IntersectionObserver/issues/308

PR #40521, opened 2025-07-24
Clarify the meaning of absolute units length

matklad:
These units are named `pixels` and `centimeters`, but it is important to call out that those are "CSS" pixels and centimeters,  and  that they _don't_ match usual definitions of things. 

Call out explicitly that `1px` != pixel and that `1cm` != centimeter. 

I explicitly decided _not_ to mention angular measures, as my goal here is to give correct intuition (or rather, avoid incorrect one), rather than being fully pedantic (for which we have a spec).

### DIFF
--- a/files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
+++ b/files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
@@ -145,6 +145,10 @@ The following are all **absolute** length units — they are not relative to any
 
 Most of these units are more useful when used for print, rather than screen output. For example, we don't typically use `cm` (centimeters) on screen. The only value that you will commonly use is `px` (pixels).
 
+Note that `1px` doesn't necessarily equal one physical device pixel. On HD displays, it may span multiple physical pixels.
+Similarly, `1cm` in CSS often doesn't correspond to one tenth of [SI](https://en.wikipedia.org/wiki/International_System_of_Units) meter. On a large TV screen, it typically be longer than that.
+The lengths are perceptual: `16px` looks roughly the same on a phone, laptop, or TV screen.
+
 #### Relative length units
 
 Relative length units are relative to something else. For example:
--- a/files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
+++ b/files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
@@ -146,7 +146,7 @@ The following are all **absolute** length units — they are not relative to any
 Most of these units are more useful when used for print, rather than screen output. For example, we don't typically use `cm` (centimeters) on screen. The only value that you will commonly use is `px` (pixels).
 
 Note that `1px` doesn't necessarily equal one physical device pixel. On HD displays, it may span multiple physical pixels.
-Similarly, `1cm` in CSS often doesn't correspond to one tenth of [SI](https://en.wikipedia.org/wiki/International_System_of_Units) meter. On a large TV screen, it typically be longer than that.
+Similarly, `1cm` in CSS often doesn't correspond to one hundredth of [SI](https://en.wikipedia.org/wiki/International_System_of_Units) meter. On a large TV screen, it typically is longer than that.
 The lengths are perceptual: `16px` looks roughly the same on a phone, laptop, or TV screen at typical viewing distance.
 
 #### Relative length units
--- a/files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
+++ b/files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
@@ -147,7 +147,7 @@ Most of these units are more useful when used for print, rather than screen outp
 
 Note that `1px` doesn't necessarily equal one physical device pixel. On HD displays, it may span multiple physical pixels.
 Similarly, `1cm` in CSS often doesn't correspond to one tenth of [SI](https://en.wikipedia.org/wiki/International_System_of_Units) meter. On a large TV screen, it typically be longer than that.
-The lengths are perceptual: `16px` looks roughly the same on a phone, laptop, or TV screen.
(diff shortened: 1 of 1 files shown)


files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
```diff
@@ -145,6 +145,10 @@ The following are all **absolute** length units — they are not relative to any
 
 Most of these units are more useful when used for print, rather than screen output. For example, we don't typically use `cm` (centimeters) on screen. The only value that you will commonly use is `px` (pixels).
 
+Note that `1px` doesn't necessarily equal one physical device pixel. On HD displays, it may span multiple physical pixels.
+Similarly, `1cm` in CSS often doesn't correspond to one tenth of [SI](https://en.wikipedia.org/wiki/International_System_of_Units) meter. On a projector, it typically appears much larger.
+The lengths are perceptual: `16px` looks roughly the same on a phone, desktop, or projector.
```
Josh-Cena: I don't think this is true. cm and px are both physical, and correspond to real-world measurements. You are right about px not being physical pixels, but the part about centimeters is more confusing than helpful. A projector is like pinch-zooming a page which by definition cannot change how the page is laid out. It's also just physically not possible: how can the projector know how long 1cm would be on the screen? Does it need laser measurements of the screen distance? On any digital display (including TV) 1cm will be 1cm.

Josh-Cena: Edit: I realized it may not be true for TVs because they are "devices with unusual viewing distances" per the spec, so if you want to use "large-screen TVs" as an example I would be fine, but we would need to verify if that's indeed the case. At least 1cm is actually 1cm for both my phone and laptop.

matklad: >I don't think this is true. cm and px are both physical, and correspond to real-world measurements.

My understanding is that, in the context of screens, the physical measure they correspond to the visual angle. 

>For screen media (including high-resolution devices), low-resolution devices, and devices with unusual viewing distances, it is recommended instead that the anchor unit be the [pixel unit](https://drafts.csswg.org/css-values-3/#visual-angle-unit). For such devices it is recommended that the pixel unit refer to the whole number of [device pixels](https://drafts.csswg.org/css-values-3/#device-pixel) that best approximates the reference pixel.
>
>The reference pixel is the visual angle of one pixel on a device with a [device pixel](https://drafts.csswg.org/css-values-3/#device-pixel) density of 96dpi and a distance from the reader of an arm’s length. For a nominal arm’s length of 28 inches, the visual angle is therefore about 0.0213 degrees. For reading at arm’s length, 1px thus corresponds to about 0.26 mm (1/96 inch).

So, the measure is physical, but, contrary to the name, it's not linear (meters), but angular (degrees). 

Which is subtle and confusing, so I don't want to get to that level of details. But I think it's important to make sure that the reade leaves with understanding that if they put a ruler to a display, then 96px pixel line would not necessarily measure as 1 inch, despite the pixel begin defined as 1/96 of an inch. 

Agree that projector would be a bad example, substituted that for a large TV screen. 

As an illustration, here's how 1cm looks on my desktop and on my mobile phone:

telegram-cloud-photo-size-2-5192885756497097823-y

files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
```diff
@@ -145,6 +145,10 @@ The following are all **absolute** length units — they are not relative to any
 
 Most of these units are more useful when used for print, rather than screen output. For example, we don't typically use `cm` (centimeters) on screen. The only value that you will commonly use is `px` (pixels).
 
+Note that `1px` doesn't necessarily equal one physical device pixel. On HD displays, it may span multiple physical pixels.
+Similarly, `1cm` in CSS often doesn't correspond to one tenth of [SI](https://en.wikipedia.org/wiki/International_System_of_Units) meter. On a large TV screen, it typically be longer than that.
+The lengths are perceptual: `16px` looks roughly the same on a phone, laptop, or TV screen.
```
Josh-Cena:
```suggestion
The lengths are perceptual: `16px` looks roughly the same on a phone, laptop, or TV screen at typical viewing distance.
```

files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
```diff
@@ -145,6 +145,10 @@ The following are all **absolute** length units — they are not relative to any
 
 Most of these units are more useful when used for print, rather than screen output. For example, we don't typically use `cm` (centimeters) on screen. The only value that you will commonly use is `px` (pixels).
 
+Note that `1px` doesn't necessarily equal one physical device pixel. On HD displays, it may span multiple physical pixels.
+Similarly, `1cm` in CSS often doesn't correspond to one tenth of [SI](https://en.wikipedia.org/wiki/International_System_of_Units) meter. On a large TV screen, it typically be longer than that.
```
Josh-Cena: Mmm

```suggestion
Similarly, `1cm` in CSS often doesn't correspond to one hundredth of [SI](https://en.wikipedia.org/wiki/International_System_of_Units) meter. On a large TV screen, it typically be longer than that.
```

Josh-Cena (approved):
Just one last thing. Great addition; thank you!

files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
matklad: lol, thanks! Gotta get one more cup of tea 🤣

matklad: wait, one more typo!

matklad: ok, should be ~~fun~~ fine (def need that cup) now!

Josh-Cena:
🍵 🍵

PR #40524, opened 2025-07-24
Migrate all images with editable source to shared-assets

Josh-Cena:
This migrates all `.drawio` and `.html` assets over, so they can be managed in a central place, and also enables deduplication.

I'm also changing SVG to PNG. SVG generated from draw.io is very cool because it supports dark mode OOTB on MDN :)

See https://github.com/mdn/shared-assets/pull/57


bsmth (approved):
Great job, thank you 🙏🏻

PR #40529, opened 2025-07-25
Fix variable redeclaration in url.origin code example

jiangying000:
```
const url = new URL("blob:https://mozilla.org:443/");
console.log(url.origin); // 'https://mozilla.org'

const url = new URL("http://localhost:80/");
console.log(url.origin); // 'http://localhost'

const url = new URL("https://mozilla.org:8080/");
console.log(url.origin); // 'https://mozilla.org:8080'
```
Error:
`Uncaught SyntaxError: Identifier 'url' has already been declared`

### DIFF
--- a/files/en-us/web/api/url/origin/index.md
+++ b/files/en-us/web/api/url/origin/index.md
@@ -28,6 +28,7 @@ A string.
 const url = new URL("blob:https://mozilla.org:443/");
 console.log(url.origin); // 'https://mozilla.org'
 ```
+
 ```js
 const url = new URL("http://localhost:80/");
 console.log(url.origin); // 'http://localhost'
--- a/files/en-us/web/api/url/origin/index.md
+++ b/files/en-us/web/api/url/origin/index.md
@@ -24,6 +24,8 @@ A string.
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


dipikabh:
Thanks for the PR @jiangying000.

I see that the current presentation of the code might be confusing. The likely intention of the original code would have been to show these as standalone examples and to not run all the lines together. It might be better instead to separate these into three JS code blocks.

Would you be able to send an update? Thanks!

jiangying000:
> Thanks for the PR @jiangying000. 
> 
> I see that the current presentation of the code might be confusing. The likely intention of the original code would have been to show these as standalone examples and to not run all the lines together. It might be better instead to separate these into three JS code blocks. 
> 
> Would be able to send an update? Thanks!

Thanks for the feedback. Done.

files/en-us/web/api/url/origin/index.md
```diff
@@ -27,10 +27,14 @@ A string.
 ```js
```
dipikabh, line 29:
```suggestion
The following examples show how the `origin` property is computed for a `blob:` URL, an `http:` URL, and one using a non-default port:

```js
```

dipikabh (changes requested):

dipikabh (approved):
Thanks a lot, @jiangying000!

PR #40534, opened 2025-07-25
docs (content-length) : Add HTTP version comparison

### LINKED ISSUE #36681: Content-Length information incomplete
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length

Specification 

the page do not include the information about the handling in different HTTP versions.  HTTP v1 vs v2 

some hint about the header heandling in different HTTP verisons

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/http/headers/content-length`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/headers/content-length/index.md
* Last commit: https://github.com/mdn/content/commit/92b03e46cef6be37de60799363e3e33e3415b491
* Document last modified: 2024-10-30T19:43:09.000Z

</details>

Julien-Dochez:
### Description

Added a comparison table clarifying 'Content-Length' header behavior differences between HTTP/1.1 and HTTP/2.

### Motivation

I made this simple change because it helps resolve confusion in a simple way. Developers may use this information to avoid unecessary header usage.

### Additional details

https://httpwg.org/specs/rfc9113.html#rfc.section.8.1

Fixes #36681

### DIFF
--- a/files/en-us/web/http/reference/headers/content-length/index.md
+++ b/files/en-us/web/http/reference/headers/content-length/index.md
@@ -47,11 +47,11 @@ Content-Length: <length>
 
 The handling of `Content-Length` differs between HTTP versions:
 
-| Feature               | HTTP/1.1                         | HTTP/2                    |
-|-----------------------|----------------------------------|---------------------------|
-| Required              | Yes (for persistent connections) | Optional                  |
-| Primary purpose       | Frame body size                  | Backward compatibility    |
-| Alternatives          | `Transfer-Encoding`              | Inferred from DATA frames |
+| Feature         | HTTP/1.1                         | HTTP/2                    |
+| --------------- | -------------------------------- | ------------------------- |
+| Required        | Yes (for persistent connections) | Optional                  |
+| Primary purpose | Frame body size                  | Backward compatibility    |
+| Alternatives    | `Transfer-Encoding`              | Inferred from DATA frames |
 
 ## Specifications
 
--- a/files/en-us/web/http/reference/headers/content-length/index.md
+++ b/files/en-us/web/http/reference/headers/content-length/index.md
@@ -45,14 +45,15 @@ Content-Length: <length>
 
 ## HTTP version comparison
 
-The handling of `Content-Length` differs between HTTP versions:
+`Content-Length` is required in HTTP/1.0.
+`Content-Length` is limited in that the message size must be known up front, before sending the headers, which is a problem when content is dynamically generated or streamed.
 
-| Feature         | HTTP/1.1                         | HTTP/2                    |
-| --------------- | -------------------------------- | ------------------------- |
-| Required        | Yes (for persistent connections) | Optional                  |
-| Primary purpose | Frame body size                  | Backward compatibility    |
-| Alternatives    | `Transfer-Encoding`              | Inferred from DATA frames |
-
+In HTTP/1.1 either `Content-Length` or {{httpheader("Transfer-Encoding","Transfer-Encoding: chunked")}} can be used to indicate message size (`Content-Length` is ignored if `Transfer-Encoding: chunked` is set).
+At least one of these is required for persistent connections.
+Note that `Transfer-Encoding: chunked` allows the content to be sent out in parts as its size is calculated, which is recommended better when the message size is not known up front.
+    
+In HTTP/2 `Content-Length` is optional and redundant, because the content length may be inferred from DATA frames. It may still be included for backwards compatibility.
+Note that `Transfer-Encoding` should not be used in HTTP/2.
 ## Specifications
 
 {{Specifications}}
--- a/files/en-us/web/http/reference/headers/content-length/index.md
+++ b/files/en-us/web/http/reference/headers/content-length/index.md
@@ -9,6 +9,12 @@ sidebar: http
 
 The HTTP **`Content-Length`** header indicates the size, in bytes, of the message body sent to the recipient.
 
+`Content-Length` is limited in that the message size must be known up front, before sending the headers, which is a problem when content is dynamically generated or streamed.
+
+- In HTTP/1.0, it is required.
+- In HTTP/1.1, it could be replaced with {{httpheader("Transfer-Encoding", "Transfer-Encoding: chunked")}} for responses sent out in parts as its size is calculated.
+- In HTTP/2, `Content-Length` is redundant, because the content length may be inferred from DATA frames. It may still be included for backwards compatibility.
+
 <table class="properties">
   <tbody>
     <tr>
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/reference/headers/content-length/index.md
```diff
@@ -43,6 +43,16 @@ Content-Length: <length>
...
 
+## HTTP version comparison
+
+The handling of `Content-Length` differs between HTTP versions:
+
+| Feature         | HTTP/1.1                         | HTTP/2                    |
+| --------------- | -------------------------------- | ------------------------- |
+| Required        | Yes (for persistent connections) | Optional                  |
+| Primary purpose | Frame body size                  | Backward compatibility    |
+| Alternatives    | `Transfer-Encoding`              | Inferred from DATA frames |
```
Josh-Cena: Just want to point out that this is very low information density and very un-MDN because you could have been a single sentence in the intro:

```suggestion
`Content-Length` was required in HTTP/1.1 for persistent connections to indicate the frame body size. It became optional in HTTP/2 because the size can now be inferred from DATA frames, and this header is only kept for compatibility.
```

files/en-us/web/http/reference/headers/content-length/index.md
```diff
@@ -43,6 +43,16 @@ Content-Length: <length>
...
+
```
hamishwillee: I think having this information is very useful, but I'm concerned that this table is "too succinct". It took me quite a while to parse, understand and confirm it. 

As I understand it (without checking)
- Either `Transfer-Encoding: chunked` or `Content-Length` can be in HTTP/1.1 for persistent connections. 
- If `Transfer-Encoding: chunked` is used, `Content-Length` is ignored.
- `Transfer-Encoding` should not be used in HTTP/2
- `Content-Length` can be used but is redundant.
-  `Content-Length` is limited in that the message size must be known up front, before sending the headers, which is a problem when content is dynamically generated or streamed. `Transfer-Encoding: chunked` allows the content to be sent out in parts as its size is calculated. 

Assuming that is correct, then perhaps something like 

```suggestion
## HTTP version comparison

`Content-Length` is required in HTTP/1.0.
`Content-Length` is limited in that the message size must be known up front, before sending the headers, which is a problem when content is dynamically generated or streamed.

In HTTP/1.1 either `Content-Length` or {{httpheader("Transfer-Encoding","Transfer-Encoding: chunked")}} can be used to indicate message size (`Content-Length` is ignored if `Transfer-Encoding: chunked` is set).
At least one of these is required for persistent connections.
Note that `Transfer-Encoding: chunked` allows the content to be sent out in parts as its size is calculated, which is recommended better when the message size is not known up front.
    
In HTTP/2 `Content-Length` is optional and redundant, because the content length may be inferred from DATA frames. It may still be included for backwards compatibility.
Note that `Transfer-Encoding` should not be used in HTTP/2.
```

What do you think?

files/en-us/web/http/reference/headers/content-length/index.md
hamishwillee: @Josh-Cena I was writing my response at the same time as you - https://github.com/mdn/content/pull/40534#discussion_r2234208086 - took me a lot longer

Similar feeling except that I thought (something like) this was useful to add in the introduction too:

> `Content-Length` is limited in that the message size must be known up front, before sending the headers, which is a problem when content is dynamically generated or streamed. Transfer-Encoding: chunked allows the content to be sent out in parts as its size is calculated.

hamishwillee: So maybe we combine those two approaches in the intro?

Josh-Cena: Yes let's go with the wording you would prefer; I'm just paraphrasing here and you would know better

hamishwillee: Cool. We agree on the main point of "MDN style". I lean towards https://github.com/mdn/content/pull/40534#discussion_r2234208086

files/en-us/web/http/reference/headers/content-length/index.md
hamishwillee: Actually I know this to be more inline with MDN patterns, so I'll commit this and we can then argue on this as the baseline.

files/en-us/web/http/reference/headers/content-length/index.md
```diff
@@ -43,6 +43,17 @@ Content-Length: <length>
...
 
+## HTTP version comparison
+
+`Content-Length` is required in HTTP/1.0.
+`Content-Length` is limited in that the message size must be known up front, before sending the headers, which is a problem when content is dynamically generated or streamed.
+
+In HTTP/1.1 either `Content-Length` or {{httpheader("Transfer-Encoding","Transfer-Encoding: chunked")}} can be used to indicate message size (`Content-Length` is ignored if `Transfer-Encoding: chunked` is set).
+At least one of these is required for persistent connections.
+Note that `Transfer-Encoding: chunked` allows the content to be sent out in parts as its size is calculated, which is recommended better when the message size is not known up front.
+In HTTP/2 `Content-Length` is optional and redundant, because the content length may be inferred from DATA frames. It may still be included for backwards compatibility.
```
hamishwillee:
```suggestion
Note that `Transfer-Encoding: chunked` allows the content to be sent out in parts as its size is calculated, which is recommended better when the message size is not known up front.

In HTTP/2 `Content-Length` is optional and redundant, because the content length may be inferred from DATA frames. It may still be included for backwards compatibility.
```

files/en-us/web/http/reference/headers/content-length/index.md
```diff
@@ -43,6 +43,18 @@ Content-Length: <length>
 - `<length>`
   - : The length in octets.
 
+## HTTP version comparison
```
Josh-Cena: FWIW I don't like ad-hoc H2 headings, would prefer if you change it to "description" or just put this in the intro.

files/en-us/web/http/reference/headers/content-length/index.md
```diff
@@ -43,6 +43,18 @@ Content-Length: <length>
...
+
+`Content-Length` is required in HTTP/1.0.
+`Content-Length` is limited in that the message size must be known up front, before sending the headers, which is a problem when content is dynamically generated or streamed.
+
+In HTTP/1.1 either `Content-Length` or {{httpheader("Transfer-Encoding","Transfer-Encoding: chunked")}} can be used to indicate message size (`Content-Length` is ignored if `Transfer-Encoding: chunked` is set).
+At least one of these is required for persistent connections.
+Note that `Transfer-Encoding: chunked` allows the content to be sent out in parts as its size is calculated, which is recommended better when the message size is not known up front.
+
+In HTTP/2 `Content-Length` is optional and redundant, because the content length may be inferred from DATA frames. It may still be included for backwards compatibility.
+Note that `Transfer-Encoding` should not be used in HTTP/2.
```
Josh-Cena:
```suggestion
`Content-Length` is limited in that the message size must be known up front, before sending the headers, which is a problem when content is dynamically generated or streamed.

- In HTTP/1.0, it is required.
- In HTTP/1.1, it could be replaced with {{httpheader("Transfer-Encoding", "Transfer-Encoding: chunked")}} for responses sent out in parts as its size is calculated.
- In HTTP/2, `Content-Length` is redundant, because the content length may be inferred from DATA frames. It may still be included for backwards compatibility.
```

A shorter wording for you to consider

hamishwillee: Yeah, this is better thank you. I've added @bsmth in case he has a comment. But otherwise perhaps will merge tomorrow.

Josh-Cena:
Brian is ooo for a few weeks. Happy to give you a rubberstamp.

hamishwillee (approved):
Thanks @Josh-Cena and @Julien-Dochez .

PR #40540, opened 2025-07-26
Improve XSLT translate description

### LINKED ISSUE #29823: Improve XPath translate function examples
This whole document could be improved. If I was doing this I would make the examples into live examples: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Code_examples#traditional_live_samples

I would also completely rewrite the text to make it clear that this is not a string replacement or regexp - it's more like a substitution cypher. So if a letter in the nth position of originalLetters appears in string, then it is replaced by the letter at the nth position in replacementLetters

```
translate(string, originalLetters, replacementLetters)
```

_Originally posted by @hamishwillee in https://github.com/mdn/content/issues/22616#issuecomment-1694791787_

Josh-Cena:
Fix https://github.com/mdn/content/issues/29823

I did not add a live sample. None of the other XSLT examples have one and it would involve far too much instrumentation.

### DIFF
--- a/files/en-us/web/xml/xpath/reference/functions/translate/index.md
+++ b/files/en-us/web/xml/xpath/reference/functions/translate/index.md
@@ -34,13 +34,13 @@ For every character in `string`, if `abc` contains that character, it is replace
 - If `XYZ` contains more characters than `abc`, the extra characters are ignored.
 - If a character appears multiple times in `abc`, then the first occurrence determines the replacement character.
 
-Note that these cases should be avoided. `translate()` is not a regexp or string replacement function. It is a character-by-character substitution function.
+`translate()` is a character-by-character substitution function, not a regexp or string replacement function. The `abc` and `XYZ` strings represent _character ciphers_, not substrings. This means that if run into any of the cases above, you may be using the method incorrectly (except perhaps having a longer `abc` to remove certain characters).
+
+A substitution like this will not output `The quick red fox` as you might expect; instead the result is `The quick red fdx`.
 
 ```xml example-bad
 <xsl:value-of select="translate('The quick brown fox', 'brown', 'red')" />
 ```
-
-This does not output `The quick red fox` as you might expect, but rather `The quick red fdx`.
 
 ## Examples
 
--- a/files/en-us/web/xml/xpath/reference/functions/translate/index.md
+++ b/files/en-us/web/xml/xpath/reference/functions/translate/index.md
@@ -30,7 +30,7 @@ The translated string.
 
 For every character in `string`, if `abc` contains that character, it is replaced with the character at the same index in `XYZ`. If `abc` does not contain that character, it is kept as-is.
 
-- If `abc` is longer than `XYZ`, then every occurrence of characters in `abc` that do not have a corresponding character in `XYZ` will be removed.
+- If `abc` is longer than `XYZ`, then the extra characters at the end of `abc` are mapped to the empty string (i.e., they are removed from the source string).
 - If `XYZ` contains more characters than `abc`, the extra characters are ignored.
 - If a character appears multiple times in `abc`, then the first occurrence determines the replacement character.
 
--- a/files/en-us/web/xml/xpath/reference/functions/translate/index.md
+++ b/files/en-us/web/xml/xpath/reference/functions/translate/index.md
@@ -28,11 +28,27 @@ The translated string.
 
 ## Description
 
+For every character in `string`, if `abc` contains that character, it is replaced with the character at the same index in `XYZ`. If `abc` does not contain that character, it is kept as-is.
+
+- If `abc` is longer than `XYZ`, then every occurrence of characters in `abc` that do not have a corresponding character in `XYZ` will be removed.
+- If `XYZ` contains more characters than `abc`, the extra characters are ignored.
+- If a character appears multiple times in `abc`, then the first occurrence determines the replacement character.
+
+Note that these cases should be avoided. `translate()` is not a regexp or string replacement function. It is a character-by-character substitution function.
+
+```xml example-bad
+<xsl:value-of select="translate('The quick brown fox', 'brown', 'red')" />
+```
+
+This does not output `The quick red fox` as you might expect, but rather `The quick red fdx`.
+
+## Examples
+
+### Using `translate()` for case conversion
+
 XPath notes that the translate function is not a sufficient solution for case conversion in all languages. A future version of XPath may provide additional functions for case conversion.
 
 However, this is the closest we have at present to a function that can convert a string to uppercase or lowercase.
-
-Example
 
 ```xml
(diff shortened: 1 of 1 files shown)


files/en-us/web/xml/xpath/reference/functions/translate/index.md
```diff
@@ -28,38 +28,38 @@ The translated string.
 
 ## Description
 
-XPath notes that the translate function is not a sufficient solution for case conversion in all languages. A future version of XPath may provide additional functions for case conversion.
+For every character in `string`, if `abc` contains that character, it is replaced with the character at the same index in `XYZ`. If `abc` does not contain that character, it is kept as-is.
 
-However, this is the closest we have at present to a function that can convert a string to uppercase or lowercase.
+- If `abc` is longer than `XYZ`, then every occurrence of characters in `abc` that do not have a corresponding character in `XYZ` will be removed.
```
hamishwillee: I find this confusing. Do you mean?

```suggestion
- If `abc` is longer than `XYZ`, then `abc` is effectively truncated to match the length of `XYZ`.
```

files/en-us/web/xml/xpath/reference/functions/translate/index.md
```diff
@@ -28,38 +28,38 @@ The translated string.
...
+- If `XYZ` contains more characters than `abc`, the extra characters are ignored.
+- If a character appears multiple times in `abc`, then the first occurrence determines the replacement character.
 
-Example
+Note that these cases should be avoided. `translate()` is not a regexp or string replacement function. It is a character-by-character substitution function.
```
hamishwillee: The sentence splits two sections of docs and they should not be on the same line.

I would be tempted to move `Note that these cases should be avoided.` above the list.
So:
```
The string of characters to be replaced should not contain duplicates, and should have the same length as the replacement string (the following cases should be avoided:

- If `abc` is longer than `XYZ`, then `abc` is effectively truncated to match the length of `XYZ`.
- ...
```

The second sentence feels a bit like it doesn't sit with what precedes it. Maybe

```
Note that `translate()` is not a regexp or string replacement function. It is a character-by-character substitution function.
This does not output `The quick red fox` as you might expect, 

<Example bad>

Instead it will output
`The quick red fdx`.
````

^^^ in "pseudo code"

**Review by hamishwillee — COMMENTED** [review]
A few comments.

Do we know if the substitutions are one a single time or a recursive? See https://github.com/mdn/content/issues/29823#issuecomment-1782183371

File: files/en-us/web/xml/xpath/reference/functions/translate/index.md (hunk 1 above)
Comment (Josh-Cena) [reply files/en-us/web/xml/xpath/reference/functions/translate/index.md]: No what it's saying is that the extra characters in `abc` are removed from the input string, i.e., replaced with the empty string

File: files/en-us/web/xml/xpath/reference/functions/translate/index.md (hunk 2 above)
Comment (Josh-Cena) [reply files/en-us/web/xml/xpath/reference/functions/translate/index.md]: What I intended is this: if you use `translate` as a char-by-char replacement, then there's no reason you would have two ciphers of different lengths. If you run into any of these cases (perhaps except `abc` longer than `xyz` if you actually want to remove some characters), you are likely using it completely wrong, such as doing string or regex replacement, and here's an example of such an invalid use case.

**Comment by Josh-Cena** [comment]
> Do we know if the substitutions are one a single time or a recursive

My first opening sentence is intended to address this, by stating the algorithm very explicitly:

> For every character in `string`, if `abc` contains that character, it is replaced with the character at the same index in `XYZ`. If `abc` does not contain that character, it is kept as-is.

That is, for every character, you do the lookup in `abc` only once. Essentially you build a replacement dictionary like `zip(*abc, *xyz)`.

File: files/en-us/web/xml/xpath/reference/functions/translate/index.md (hunk 1 above)
Comment (Josh-Cena) [reply files/en-us/web/xml/xpath/reference/functions/translate/index.md]: I could say:


```suggestion
- If `abc` is longer than `XYZ`, then the extra characters of `abc` are mapped to the empty string (i.e., removed).
```

hamishwillee: Thanks.

```suggestion
- If `abc` is longer than `XYZ`, then the extra characters at the end of `abc` are mapped to the empty string (i.e., they are removed from the source string).
```

files/en-us/web/xml/xpath/reference/functions/translate/index.md
hamishwillee: OK, it says what you think it says. I'm going to propose the following for your consideration, because even so, I was able to misinterpret this:

````
For every character in `string`, if `abc` contains that character, it is replaced with the character at the same index in `XYZ`. If `abc` does not contain that character, it is kept as-is.

Generally the replaced (`abc`) and replacement (`XYZ`) strings should be the same length, and `abc` should not contain duplicates:

- If `abc` is longer than `XYZ`, then the extra characters at the end of `abc` are mapped to the empty string (i.e., they are removed from the source string).
- If `XYZ` contains more characters than `abc`, the extra characters are ignored.
- If a character appears multiple times in `abc`, then the first occurrence determines the replacement character.

If the two strings have different lengths it is likely that you're using the method incorrectly (except perhaps having a longer `abc` to remove additional characters).

> [!WARNING]
`translate()` is a character-by-character substitution function, not a regexp or string replacement function. 
> 
> A substitution like this will not output `The quick red fox` as you might expect:
>
> ```xml example-bad
>  <xsl:value-of select="translate('The quick brown fox', 'brown', 'red')" />
> ```
>
> Instead the result is `The quick red fdx`.
````

File: files/en-us/web/xml/xpath/reference/functions/translate/index.md [hunk 3]
```diff
@@ -28,38 +28,38 @@ The translated string.
...
+For every character in `string`, if `abc` contains that character, it is replaced with the character at the same index in `XYZ`. If `abc` does not contain that character, it is kept as-is.
 
-However, this is the closest we have at present to a function that can convert a string to uppercase or lowercase.
+- If `abc` is longer than `XYZ`, then the extra characters at the end of `abc` are mapped to the empty string (i.e., they are removed from the source string).
+- If `XYZ` contains more characters than `abc`, the extra characters are ignored.
+- If a character appears multiple times in `abc`, then the first occurrence determines the replacement character.
 
-Example
+`translate()` is a character-by-character substitution function, not a regexp or string replacement function. The `abc` and `XYZ` strings represent _character ciphers_, not substrings. This means that if run into any of the cases above, you may be using the method incorrectly (except perhaps having a longer `abc` to remove certain characters).
```
Josh-Cena:
```suggestion
`translate()` is a character-by-character substitution function, not a regexp or string replacement function. The `abc` and `XYZ` strings represent _character ciphers_, not substrings. This means that if you run into any of the cases above, you may be using the method incorrectly (except perhaps having a longer `abc` to remove certain characters).
```

hamishwillee (approved):
Thanks, much better.

PR #40544, opened 2025-07-27
Specify getFloatTimeDomainData() encoding range

### LINKED ISSUE #25571: AnalyserNode::getFloatTimeDomainData() doesn't specify the encoding of the waveform
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode/getFloatTimeDomainData

Top-level description

Missing encoding of the waveform. Because the parameter is a `Float32Array`, one can reasonably assume that the encoding is `f32` (that is, PCM data amplitude in [-1:1]). However it would be better to confirm this in writing instead of leaving developers to their guess. There's also no mention of planar/non-planar encoding, which is necessary to do anything with the data retrieved (unless the signal has a single channel).

Something akin to:

```
The time-domain data is encoded in `f32` format with the amplitude of samples in [-1:1].
```

_No response_

The 3 other instance methods of `AnalyzerNode` have the same issue. Especially for the `byte` variant where it's unclear if the data is signed [-128:127] or unsigned [0:255]. The example in https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode/getByteTimeDomainData hints that it may be the former.

### MDN metadata

<details>

* Folder: `en-us/web/api/analysernode/getfloattimedomaindata`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode/getFloatTimeDomainData
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/analysernode/getfloattimedomaindata/index.md
* Last commit: https://github.com/mdn/content/commit/bca8d1ab2bc4f5a1ef6b39c454b0229539178e98
* Document last modified: 2023-02-18T07:09:59.000Z

</details>

Josh-Cena:
Fix https://github.com/mdn/content/issues/25571. Wondering if you could advise @padenot—thanks!

### DIFF
--- a/files/en-us/web/api/analysernode/getfloattimedomaindata/index.md
+++ b/files/en-us/web/api/analysernode/getfloattimedomaindata/index.md
@@ -10,7 +10,7 @@ browser-compat: api.AnalyserNode.getFloatTimeDomainData
 
 The **`getFloatTimeDomainData()`** method of the {{ domxref("AnalyserNode") }} Interface copies the current waveform, or time-domain, data into a {{jsxref("Float32Array")}} array passed into it. Each array value is a _sample_, the magnitude of the signal at a particular time.
 
-The waveform is encoded as PCM data, which has a nominal range of -1.0 to 1.0, but values can exceed the range such as when down-mixing stereo to mono.
+The waveform is represented as PCM data, which has a nominal range of -1.0 to 1.0, but values can exceed the range such as when down-mixing stereo to mono.
 
 ## Syntax
 
--- a/files/en-us/web/api/analysernode/getfloattimedomaindata/index.md
+++ b/files/en-us/web/api/analysernode/getfloattimedomaindata/index.md
@@ -10,6 +10,8 @@ browser-compat: api.AnalyserNode.getFloatTimeDomainData
 
 The **`getFloatTimeDomainData()`** method of the {{ domxref("AnalyserNode") }} Interface copies the current waveform, or time-domain, data into a {{jsxref("Float32Array")}} array passed into it. Each array value is a _sample_, the magnitude of the signal at a particular time.
 
+The waveform is encoded as PCM data, which has a nominal range of -1.0 to 1.0, but values can exceed the range such as when down-mixing stereo to mono.
+
 ## Syntax
 
 ```js-nolint


files/en-us/web/api/analysernode/getfloattimedomaindata/index.md
```diff
@@ -10,6 +10,8 @@ browser-compat: api.AnalyserNode.getFloatTimeDomainData
 
 The **`getFloatTimeDomainData()`** method of the {{ domxref("AnalyserNode") }} Interface copies the current waveform, or time-domain, data into a {{jsxref("Float32Array")}} array passed into it. Each array value is a _sample_, the magnitude of the signal at a particular time.
 
+The waveform is encoded as PCM data, which has a nominal range of -1.0 to 1.0, but values can exceed the range such as when down-mixing stereo to mono.
```
padenot: It's a bit of a duplicate of the paragraph above, that says it's a copy of the waveform data. But I looked elsewhere on MDN, and I think we should say that waveform data are nominally in `[-1, 1]` at least once, e.g. in https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Basic_concepts_behind_Web_Audio_API in which we say it's float 32-bits.

`encoded` might be a bit strange because the data is copied unaltered (which is important).

Josh-Cena: Re: duplication, "A sample, the magnitude of the signal" sounds a bit less straightforward than "PCM data" to me as an amateur. The guide page you linked doesn't really talk about the data encoding at all, so I think this looks like the best place to put it. Would "represented" be a better word?

sideshowbarker (approved):
@Josh-Cena I don’t have any special insight to offer here, so I’m going ahead and marking it approved, but leaving it unmerged and up to you whether to wait for resolution of the review comment (or else leave that as an issue for a follow-up PR)

files/en-us/web/api/analysernode/getfloattimedomaindata/index.md
Josh-Cena:
```suggestion
The waveform is represented as PCM data, which has a nominal range of -1.0 to 1.0, but values can exceed the range such as when down-mixing stereo to mono.
```

Josh-Cena:
@padenot PTAL, thanks!

padenot:
Looks good, thanks!

Josh-Cena:
Thanks both :)

PR #40546, opened 2025-07-27
Refine CSS Value Processing document

sakupi01:
### Description

The change mainly aims for better accuracy to describe CSS value processing flow according to the specification.

### Motivation

The previous version sometimes used incorrect terminology and descriptions, which will ultimately lead differences between mdn and the specs.

e.g.

- CSS property value result will not come only from specificity - it will be the result of the entire cascade process
- "specified values" were mostly used incorrectly/confused with "declared values" - "specified values" result from the defaulting process after cascading, but the previous version didn't make this clear
- The document mainly described the stages after the [Defaulting process](https://www.w3.org/TR/css-cascade-4/#defaulting), even though the terminology and phases _before_ Defaulting
 (which are Filtering, Cascading, and related terms) are necessary to make _after_-Defaulting values understandable. Omitting the before-Defaulting process essentially leads to misworded descriptions.
- Initial value shouldn't be listed as a value processing stage according to the spec - it is the terminology used in the specified value resolution process 

The change also includes some typo/wording updates, which will make the description more accurate or simpler.

### Additional details

The changes ensure compliance with:
- [CSS Syntax Module Level 3](https://www.w3.org/TR/css-syntax-3/)
- [CSS Cascading and Inheritance Level 4](https://www.w3.org/TR/css-cascade-4/)


files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+Every CSS property's value comes from the declaration that wins the {{cssxref("Cascade")}} algorithm. When multiple declarations provide different property values for the same element, the declaration that ranks highest in the [Cascade Sorting Order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) gets applied.
 
-Each property value comes from a single property-value pair; is single value is applied from each property. Even if the value is a comma separated list of values, that list of values came from a single declaration.
+Each property value comes from a single property name-value pair; a single value is applied from each property. Even if the value is a comma separated list of values, that list of values came from a single declaration.
 
-To determine which specified value is applied, the user agent gathers and processes all the styles from different sources, such as inline styles, and internal and external stylesheets.
+To determine which [declared value](#declared-value) is applied, the user agent gathers and processes all the styles from different sources, such as inline styles, and internal and external stylesheets.
```
Josh-Cena:
```suggestion
To determine which [declared value](#declared_value) is applied, the user agent gathers and processes all the styles from different sources, such as inline styles, and internal and external stylesheets.
```

Make sure all your anchor links are valid

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
 
 ## Property values
 
-Every CSS property's value comes from the declaration with the greatest {{cssxref("specificity")}}. If two or more declarations with the same specificity provide different property values for the same element, the declaration value whose selector has the greatest algorithmic weight gets applied.
+Every CSS property's value comes from the declaration that wins the {{cssxref("Cascade")}} algorithm. When multiple declarations provide different property values for the same element, the declaration that ranks highest in the [Cascade Sorting Order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) gets applied.
```
Josh-Cena: `CSS/Cascade` is a redirect, plus `{{cssxref("Cascade")}}` would be rendered as code, so you should just use a markdown link instead. Also sentence case.

```suggestion
Every CSS property's value comes from the declaration that wins the [cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) algorithm. When multiple declarations provide different property values for the same element, the declaration that ranks highest in the [cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) gets applied.
```

Same below.

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+```css
+/* Result of defaulting - Specified Value */
+font-size: 2em;
+```
+
+For properties without [cascaded values](#cascaded-value), the [defaulting](#defaulting) process fills in the gaps:
+
+```css
+/* Example of inherited property */
+color: /* inherits from parent's computed value */
```
Josh-Cena: ?

Josh-Cena (commented):
Drive-by, just looking at Markdown correctness. Not interested in reviewing, although this looks like good work; thank you!

files/en-us/web/css/css_cascade/value_processing/index.md
estelle:
```suggestion
Every CSS property's value comes from the declaration that wins the {{cssxref("cascade")}} algorithm. When multiple declarations provide different property values for the same element, the declaration that ranks highest in the [cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) gets applied.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+Each property value comes from a single property name-value pair; a single value is applied from each property. Even if the value is a comma separated list of values, that list of values came from a single declaration.
 
-To determine which specified value is applied, the user agent gathers and processes all the styles from different sources, such as inline styles, and internal and external stylesheets.
+To determine which [declared value](#declared-value) is applied, the user agent gathers and processes all the styles from different sources, such as inline styles, and internal and external stylesheets.
+
+The {{cssxref("Cascade")}} determines which value should be applied when multiple conflicting styles target the same element. The [cascade algorithm](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) defines how user agents combine property values originating from different sources, scopes, and/or layers. When a selector matches an element, the property's [declared value](#declared-value) from the origin with the highest precedence gets applied, even if a selector from a lower precedence origin or layer has greater {{cssxref("specificity")}}.
```
estelle:
```suggestion
When multiple conflicting styles target the same element, the {{cssxref("cascade")}} determines which value should be applied. The [cascade algorithm](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) defines how user agents combine property values originating from different sources, scopes, and/or layers. When a selector matches an element, the property's [declared value](#declared-value) from the origin with the highest precedence gets applied, even if a selector from a lower precedence origin or layer has greater {{cssxref("specificity")}}.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
 
-The [cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) determines which value should be applied when multiple conflicting styles target the same element. The cascade algorithm defines how user agents combine property values originating from different sources, scopes, and/or layers. When a selector matches an element, the property's [specified value](#specified_value) from the origin with the highest precedence gets applied, even if a selector from a lower precedence origin or layer has greater {{cssxref("specificity")}}.
+After applying the [cascading](#cascading) rules and defaulting values step by step, the browser ensures the visual presentation matches the processed CSS.
+
+## Processing overview
+
+Before diving into the individual value stages, it's important to understand the three main phases that occur in value processing; [Filtering](#filtering), [Cascading](#cascading), and [Defaulting](#defaulting).
```
estelle:
```suggestion
Before diving into the individual value stages, it's important to understand the three main phases that occur in value processing; [filtering](#filtering), [cascading](#cascading), and [defaulting](#defaulting).
```
lowercase everywhere except at the start of sentences and headers

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+```css
+/* Valid declarations become declared values */
+p {
+  font-size: 1.25em; /* Valid */
+  colr: blue; /* Invalid property name - filtered out */
+  font-weight: bold; /* Valid */
+}
+```
+
+As a result of filtering, each element has zero or more [**declared values**](#declared-value) for each property. These declared values are the starting point for the next processing stages; [Cascading](#cascading).
```
estelle:
```suggestion
When filtering is complete, every element has zero or more [declared values](#declared-value) for every CSS property. These declared values are the starting point for the [cascading](#cascading) processing stage.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+
+### Cascading
+
+{{cssxref("Cascade")}} resolves conflicts when multiple declarations apply to the same property on the same element. {{cssxref("Cascade")}} sorts declarations using [the Cascade Sorting Order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) algorithm.
```
estelle:
```suggestion
The {{cssxref("cascade")}} phase resolves the conflicts that would otherwise occur when multiple declarations apply to the same property on the same element by sorting declarations based on [the cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) algorithm.
```

sakupi01:
@Josh-Cena 

Thanks for the quick review! Fixed in:

[correct anchor links](https://github.com/mdn/content/pull/40546/commits/7a730adc1e0167932a89767be91c56eb8fe2cd37)
[use non-code/redirect ref](https://github.com/mdn/content/pull/40546/commits/e16c142c859fcd3d166ef010b790943864f01670)
[update non-inherited/inherited property code description](https://github.com/mdn/content/pull/40546/commits/0a5567bc22a9311c3bc2e577dabf18ff3c343a80)

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
 
 ## Property values
 
-Every CSS property's value comes from the declaration with the greatest {{cssxref("specificity")}}. If two or more declarations with the same specificity provide different property values for the same element, the declaration value whose selector has the greatest algorithmic weight gets applied.
+Every CSS property's value comes from the declaration that wins the [cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) algorithm. When multiple declarations provide different property values for the same element, the declaration that ranks highest in the [cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) gets applied.
 
-Each property value comes from a single property-value pair; is single value is applied from each property. Even if the value is a comma separated list of values, that list of values came from a single declaration.
+Each property value comes from a single property name-value pair; a single value is applied from each property. Even if the value is a comma separated list of values, that list of values came from a single declaration.
```
sakupi01: property = name + value

see: https://www.w3.org/TR/css-syntax-3/#declaration

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
 
-To determine which specified value is applied, the user agent gathers and processes all the styles from different sources, such as inline styles, and internal and external stylesheets.
+To determine which [declared value](#declared_value) is applied, the user agent gathers and processes all the styles from different sources, such as inline styles, and internal and external stylesheets.
```
sakupi01, line 17: The term "specified value" was confusing in the entire context.
"specified value" comes after declaration value and cascaded value, but here it sounded like it comes before cascading.

https://www.w3.org/TR/css-cascade-4/#value-stages

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
 
 ## Property values
 
-Every CSS property's value comes from the declaration with the greatest {{cssxref("specificity")}}. If two or more declarations with the same specificity provide different property values for the same element, the declaration value whose selector has the greatest algorithmic weight gets applied.
+Every CSS property's value comes from the declaration that wins the [cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) algorithm. When multiple declarations provide different property values for the same element, the declaration that ranks highest in the [cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) gets applied.
```
sakupi01: It could be read like property value comes because of only from the resolution of the specificity algorithm, but it's actually not.

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
 
 Certain properties inherit values from their parent elements unless explicitly overridden. [Inheritance](/en-US/docs/Web/CSS/CSS_cascade/Inheritance) occurs when no style information exists for a specific property on an element. If the property is inherited, the value is set to the [computed value](#computed_value) of the parent element. If the property is not inherited, its value is set to the [initial value](#initial_value) for that element.
 
-The [cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) determines which value should be applied when multiple conflicting styles target the same element. The cascade algorithm defines how user agents combine property values originating from different sources, scopes, and/or layers. When a selector matches an element, the property's [specified value](#specified_value) from the origin with the highest precedence gets applied, even if a selector from a lower precedence origin or layer has greater {{cssxref("specificity")}}.
+After applying the [cascading](#cascading) rules and defaulting values step by step, the browser ensures the visual presentation matches the processed CSS.
+
+## Processing overview
```
sakupi01, line 25: In order not to use "specified value" incorrectly (or to use "declared value" and "cascaded value" correctly), and in order to follow [Value Processing steps](https://www.w3.org/TR/css-cascade-4/#value-stages) based on the spec, the pre-description of three processing steps, which are [Filtering](https://www.w3.org/TR/css-cascade-4/#filtering), [Cascading](https://www.w3.org/TR/css-cascade-4/#cascading),  and [Defaulting](https://www.w3.org/TR/css-cascade-4/#defaulting), is essentially needed.

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+All elements that are part of the document's flattened element tree have declared, cascaded, specified, computed, used, and actual values. For a specific property, these values may or may not be the same. For example, if your large code base includes the CSS `p { font-size: 1.25em; }` and your HTML includes `<p class="large">CSS is fun!</p>`, what size will the paragraph be? The {{cssxref("font-size")}} value moves through a few stages to go from the `em` specified value to the rendered `px` value.
+
+The value processing stages are:
 
-- [Initial value](#initial_value)
+- [Declared value](#declared_value)
+- [Cascaded value](#cascaded_value)
 - [Specified value](#specified_value)
 - [Computed value](#computed_value)
 - [Used value](#used_value)
```
sakupi01, line 92: The terminology "Initial value" is used IN the resolution phase of specified value, that is Defaulting phase - it's not the term to describe the part of value processing flow.

https://www.w3.org/TR/css-cascade-4/#value-stages
https://www.w3.org/TR/css-cascade-4/#defaulting

sakupi01:
@estelle 

I would love a review from you, if that's possible!

files/en-us/web/css/css_cascade/value_processing/index.md
estelle:
```suggestion
Every CSS property is set to its [cascaded value](#cascaded_value), which comes from the property declaration that ranks highest in the [cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) based on the [cascade algorithm](/en-US/docs/Web/CSS/CSS_cascade/Cascade).
```
If the term is cascaded value, let's use that. And leave the "win" to that area and move the "multiple declarations" to the next paragraph.

files/en-us/web/css/css_cascade/value_processing/index.md
estelle:
```suggestion
Even when there are multiple [declared values](#declared_value), with multiple declarations providing the same or different property values for the same element, each property value must still come from a single property name-value pair as only a single value is applied from each property. Even if the value is a comma-separated list of values, that list of values came from a single declaration.
```

files/en-us/web/css/css_cascade/value_processing/index.md
estelle: moved "declared" up, as there may be zero (or many) declared values applied to the element

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+Each property value comes from a single property name-value pair; a single value is applied from each property. Even if the value is a comma separated list of values, that list of values came from a single declaration.
 
-To determine which specified value is applied, the user agent gathers and processes all the styles from different sources, such as inline styles, and internal and external stylesheets.
+To determine which [declared value](#declared_value) is applied, the user agent gathers and processes all the styles from different sources, such as inline styles, and internal and external stylesheets.
+
+The [cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) determines which value should be applied when multiple conflicting styles target the same element. The [cascade algorithm](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) defines how user agents combine property values originating from different sources, scopes, and/or layers. When a selector matches an element, the property's [declared value](#declared_value) from the origin with the highest precedence gets applied, even if a selector from a lower precedence origin or layer has greater {{cssxref("specificity")}}.
```
estelle: let's link the terms "layer" and "origin"

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+## Processing overview
+
+Before diving into the individual value stages, it's important to understand the three main phases that occur in value processing; [Filtering](#filtering), [Cascading](#cascading), and [Defaulting](#defaulting).
+
+### Filtering
+
+**Filtering** is the process of identifying all declarations that apply to each element. During this phase, the browser:
+
+- Verifies that declarations belong to active stylesheets
+- Checks that property names are valid and recognized
```
estelle: can we merge #2 into #4? If the property name is not valid or recognized, it's invalid according to the browser (even if valid in supporting browsers. 

We may also want to link to the error processing (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_syntax/Error_handling) and link from there back to here.

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+```css
+/* Valid declarations become declared values */
+p {
+  font-size: 1.25em; /* Valid */
+  colr: blue; /* Invalid property name - filtered out */
+  font-weight: bold; /* Valid */
+}
+```
+
+As a result of filtering, each element has zero or more [**declared values**](#declared_value) for each property. These declared values are the starting point for the next processing stages; [Cascading](#cascading).
```
estelle:
```suggestion
As a result of filtering, each element has zero or more [**declared values**](#declared_value) for each property. These declared values are the starting point for the next processing stages; [cascading](#cascading).
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+- Checks that property names are valid and recognized
+- Validates that property values match the expected syntax
+- Eliminates declarations that are syntactically invalid
+
+```css
+/* Valid declarations become declared values */
+p {
+  font-size: 1.25em; /* Valid */
+  colr: blue; /* Invalid property name - filtered out */
+  font-weight: bold; /* Valid */
```
estelle:
```suggestion
p {
  font-size: 1.25em;
  colr: blue;
  font-weight: bold;
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,156 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+
+### Filtering
+
+**Filtering** is the process of identifying all declarations that apply to each element. During this phase, the browser:
+
+- Verifies that declarations belong to active stylesheets
+- Checks that property names are valid and recognized
+- Validates that property values match the expected syntax
+- Eliminates declarations that are syntactically invalid
+
```
estelle, line 41:
```suggestion

Only valid declarations become declared values. Declarations with invalid property names or invalid values get filtered out. In this example, only the {{cssxref("font-size")}} and {{cssxref("font-weight")}} declarations are processed. The [CSS parser filters out errors](/en-US/docs/Web/CSS/CSS_syntax/Error_handling#css_parser_errors), ignoring or "filtering" out the declaration with the invalid property name:
```

in all code examples on the page, remove as many comments as possible, opting to introduce the issue displayed in the code block as much as possible in prose rather than as code comments.

estelle (commented):
This PR is excellent. I only got through the first bit, so have more to review, but I am heading out in 20 minutes for the week so I wanted to commit what I have instead of leaving you waiting.

I realize from the number of comments in just the first portion of this it seems like i may be being critical. I am just trying to ensure that the level is such that people with less knowledge than us aren't confused. So please don't take the number of edits as a reflection of the work you did. It's excellent. I am just making suggestions that I hope will answer the readers questions before they ask them.

files/en-us/web/css/css_cascade/value_processing/index.md
sakupi01: > Every CSS property is set to its [cascaded value](#cascaded_value),

Every CSS property doesn't necessarily have its cascaded value - if there are multiple declarations set, there's a cascaded value. 
see: https://www.w3.org/TR/css-cascade-4/#stages-examples

So, I mended the phrase a bit, embracing your suggestion!

files/en-us/web/css/css_cascade/value_processing/index.md
sakupi01: fixed in: [links for layers and origins](https://github.com/mdn/content/pull/40546/commits/57d9f3511c4cfb4ff82cff0af87e6f04fc5d3e33)

files/en-us/web/css/css_cascade/value_processing/index.md
sakupi01: fixed in: [modify&split the first paragraph](https://github.com/mdn/content/pull/40546/commits/118b89e3c623f62af28c98e557f32c3b437fecf3)

files/en-us/web/css/css_cascade/value_processing/index.md
sakupi01: fixed in: [modify&split the first paragraph](https://github.com/mdn/content/pull/40546/commits/118b89e3c623f62af28c98e557f32c3b437fecf3)

files/en-us/web/css/css_cascade/value_processing/index.md
sakupi01: fixed in: [modify&split the first paragraph](https://github.com/mdn/content/pull/40546/commits/118b89e3c623f62af28c98e557f32c3b437fecf3)

files/en-us/web/css/css_cascade/value_processing/index.md
sakupi01: fixed in: [merge bullet items and add link to error handling](https://github.com/mdn/content/pull/40546/commits/6515be73380a31cce2cb74c4e8ff2056aa28122a)

files/en-us/web/css/css_cascade/value_processing/index.md
sakupi01: fixed in: [make all code comments removed and replaced with prose explanations](https://github.com/mdn/content/pull/40546/commits/10974c491fc927dedbae4ab871a2c400d3955fe6)

sakupi01:
Thank you so much for taking the time to review this, @estelle! 

I really appreciate your thoughtful feedback, especially knowing how busy you are.
I've addressed the points you raised and commented respectively. 
Looking forward to your thoughts on the rest when you get a chance. 

Enjoy your time off!

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
 
 ## Property values
 
-Every CSS property's value comes from the declaration with the greatest {{cssxref("specificity")}}. If two or more declarations with the same specificity provide different property values for the same element, the declaration value whose selector has the greatest algorithmic weight gets applied.
+Every CSS property that has declarations applied to it is set to its [cascaded value](#cascaded_value), which comes from the property declaration that ranks highest in the [cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) based on the [cascade algorithm](/en-US/docs/Web/CSS/CSS_cascade/Cascade).
```
estelle:
```suggestion
Every style applied to an element or pseudo-element is based on a single CSS property declaration. Each CSS property has only one value. The value that is applied is determined by the [cascaded values](#cascaded_value) of all the declarations of that property that apply to that element or pseudo-element, with the single value applied coming from the property declaration that ranks highest in the [cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) based on the [cascade algorithm](/en-US/docs/Web/CSS/CSS_cascade/Cascade).
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
 
-Each property value comes from a single property-value pair; is single value is applied from each property. Even if the value is a comma separated list of values, that list of values came from a single declaration.
+When there are multiple [declared values](#declared_value), with multiple declarations providing the same or different property values for the same element, each property value must still come from a single property name-value pair as only a single value is applied from each property. Even if the value is a comma-separated list of values, that list of values came from a single declaration.
```
estelle:
```suggestion
When there are multiple [declared values](#declared_value), with multiple declarations, providing the same or different property values for the same element, each property value must still come from a single property name-value pair as only a single value is applied from each property, even if the value is a comma-separated list of values.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
 
-The [cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) determines which value should be applied when multiple conflicting styles target the same element. The cascade algorithm defines how user agents combine property values originating from different sources, scopes, and/or layers. When a selector matches an element, the property's [specified value](#specified_value) from the origin with the highest precedence gets applied, even if a selector from a lower precedence origin or layer has greater {{cssxref("specificity")}}.
+Certain properties inherit values from their parent elements unless explicitly overridden. [inheritance](/en-US/docs/Web/CSS/CSS_cascade/Inheritance) occurs when no style information exists for a specific property on an element. If the property is inherited, the value is set to the [computed value](#computed_value) of the parent element. If the property is not inherited, its value is set to the [initial value](#initial_value) for that element.
```
estelle:
```suggestion
Certain properties inherit values from their parent elements unless explicitly overridden. [Inheritance](/en-US/docs/Web/CSS/CSS_cascade/Inheritance) may occur when no style information exists for a specific property on an element. If the property is inherited, the value is set to the [computed value](#computed_value) of the parent element. If the property is not inherited, its value is set to the [initial value](#initial_value) for that element.
```
 I added the "may"  because if the property is not inherited, its value is set to the initial value, so it doesn't occur in that case. The original was correct, so this edit is optional, but the "may" might help some, but I'll leave it up to you.

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+
+Before diving into the individual value stages, it's important to understand the three main phases that occur in value processing; [filtering](#filtering), [cascading](#cascading), and [defaulting](#defaulting).
+
+### Filtering
+
+**Filtering** is the process of identifying all declarations that apply to each element. A declaration applies to an element only if:
+
+- It belongs to a style sheet that currently applies to this document
+- Any [conditional rules](/en-US/docs/Web/CSS/CSS_conditional_rules) (like {{cssxref("@media")}} or {{cssxref("@supports")}}) that contain it are currently true, and it belongs to a style rule whose selector matches the element
+- It is syntactically valid: the property name is recognized by the browser and the value matches the expected syntax for that property
```
estelle:
```suggestion
- The declaration belongs to a style sheet that currently applies to this document
- Any [conditional rules](/en-US/docs/Web/CSS/CSS_conditional_rules) (like {{cssxref("@media")}} or {{cssxref("@supports")}}) that contain the declaration are currently true.
- The declaration belongs to a style rule whose selector matches the element
- The declaration is syntactically valid: the property name is recognized by the browser and the value matches the expected syntax for that property
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+  colr: blue;
+  font-weight: bold;
+}
+```
+
+When filtering is complete, every element has zero or more [declared values](#declared-value) for every CSS property. These declared values are the starting point for the [cascading](#cascading) processing stage.
+
+### Cascading
+
+[Cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) resolves conflicts when multiple declarations apply to the same property on the same element. [cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) sorts declarations using [the cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) algorithm.
```
estelle:
```suggestion
[Cascade](/en-US/docs/Web/CSS/CSS_cascade/Cascade) resolves conflicts when multiple declarations apply to the same property on the same element. Cascade sorts declarations using [the cascade sorting order](/en-US/docs/Web/CSS/CSS_cascade/Cascade#cascading_order) algorithm.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+
+For example, if you have the following declarations for `<p class="large">CSS is fun!</p>`, the second declaration wins because it has higher specificity. Both declarations have author origin, but the second selector has specificity of 0,1,1 while the first has 0,0,1:
```
estelle:
```suggestion
For example, both {{cssxref("font-size")}} declarations match `<p class="large">CSS is fun!</p>`, but the second declaration gets applied because it has higher {{cssxref("specificity")}}. Both declarations have author origin, but the second selector has specificity of `0-1-1` while the first has `0-0-1`:
```
we use the `x-y-z` syntax on the specificity page, and i think everywhere.

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+p {
+  font-size: 1em;
+}
+
+p.large {
+  font-size: 1.5em;
+}
+```
+
+After cascading, the browser determines the [**cascaded value**](#cascaded_value) for each property on each element. This value is the one that will be used in the next processing stage; [defaulting](#defaulting).
```
estelle:
```suggestion
After cascading, the browser determines the [**cascaded value**](#cascaded_value) for each property on each element. This is the value used in the next processing stage; [defaulting](#defaulting).
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+
+### Defaulting
+
+**Defaulting** ensures every property on every element has a value, even when no declarations apply. This involves:
```
estelle:
```suggestion
**Defaulting** ensures every property on every element has a value. This involves applying default property values when no  CSS declarations explicitly set that property value. The CSS parser sets a value by:
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+
+- Using **inherited values** for [inherited properties](/en-US/docs/Web/CSS/CSS_cascade/Inheritance#inherited_properties)
+- Using **initial values** for [non-inherited properties](/en-US/docs/Web/CSS/CSS_cascade/Inheritance#non-inherited_properties)
+- Applying **explicit defaulting keywords** ([`initial`](/en-US/docs/Web/CSS/initial), [`inherit`](/en-US/docs/Web/CSS/inherit), [`unset`](/en-US/docs/Web/CSS/unset), [`revert`](/en-US/docs/Web/CSS/revert), [`revert-layer`](/en-US/docs/Web/CSS/revert-layer))
```
estelle: Does the parser ever set anything other than initial or inherit? Should the first two read:

```suggestion
-  Setting `inherit` for [inherited properties](/en-US/docs/Web/CSS/CSS_cascade/Inheritance#inherited_properties)
- Setting `initial` for [non-inherited properties](/en-US/docs/Web/CSS/CSS_cascade/Inheritance#non-inherited_properties)
```

and the last be removed?

Aren't these already in the browsers UA stylesheet? So is the parser doing this explicitly, or is it just reading the UA styles?

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+All elements that are part of the document's flattened element tree have declared, cascaded, specified, computed, used, and actual values. For a specific property, these values may or may not be the same. For example, if your large code base includes the CSS `p { font-size: 1.25em; }` and your HTML includes `<p class="large">CSS is fun!</p>`, what size will the paragraph be? The {{cssxref("font-size")}} value moves through a few stages to go from the `em` specified value to the rendered `px` value.
 
-- [Initial value](#initial_value)
+The value processing stages are:
+
+- [Declared value](#declared_value)
+- [Cascaded value](#cascaded_value)
 - [Specified value](#specified_value)
 - [Computed value](#computed_value)
 - [Used value](#used_value)
```
estelle, line 92: should "actual value" be added here?

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+
+As a result of defaulting, every property is guaranteed to have a [specified value](#specified_value).
 
 ## Processing stages
 
-All elements that are part of the document's flattened element tree have declared, cascaded, specified, computed, used, and actual values. For a specific property, these values may or may not be the same. For example, if your large code base includes the CSS `p { font-size: 1.25em; }` and your HTML includes `<p>CSS is fun!</p>`, what size will the paragraph be? The {{cssxref("font-size")}} value moves through a few stages to go from the `em` specified value to the rendered `px` value.
+All elements that are part of the document's flattened element tree have declared, cascaded, specified, computed, used, and actual values. For a specific property, these values may or may not be the same. For example, if your large code base includes the CSS `p { font-size: 1.25em; }` and your HTML includes `<p class="large">CSS is fun!</p>`, what size will the paragraph be? The {{cssxref("font-size")}} value moves through a few stages to go from the `em` specified value to the rendered `px` value.
```
estelle:
```suggestion
All elements that are part of the document's flattened element tree have [declared](#declared_value), [cascaded](#cascaded_value), [specified](#specified_value), [computed](#computed_value), [used](#used_value), and [actual](actual_value) values. For a specific property, these values may or may not be the same. For example, if your large code base includes the CSS `p { font-size: 1.25em; }` and your HTML includes `<p class="large">CSS is fun!</p>`, what size will the paragraph be? The {{cssxref("font-size")}} value moves through a few stages to go from the `em` specified value to the rendered `px` value.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+A **declared value** is any syntactically valid value from a declaration that applies to an element. An element can have zero or more declared values for each property. These values come from style sheets (author, user, or user-agent) and are identified during the [filtering](#filtering) stage.
 
-- For [inherited properties](/en-US/docs/Web/CSS/CSS_cascade/Inheritance#inherited_properties), the initial value is used on the _root element only_, as long as no [specified value](#specified_value) is supplied.
-- For [non-inherited properties](/en-US/docs/Web/CSS/CSS_cascade/Inheritance#non-inherited_properties), the initial value is used on _all elements_, as long as no specified value is supplied.
+For our example `p { font-size: 1.25em; }` for `<p class="large">CSS is fun!</p>`, if multiple stylesheets declare font-size for the same paragraph, the user-agent stylesheet might set `font-size: 1em` for paragraphs, while the author stylesheet sets `font-size: 1.25em` for paragraphs and `font-size: 2em` for elements with class "large":
```
estelle:
```suggestion
Continuing our example, in which our stylesheet includes an occurrence of `p { font-size: 1.25em; }` and the document linking to that stylesheet includes `<p class="large">CSS is fun!</p>`, there may be other `font-size` declarations that could potentially apply to the same paragraph. The user-agent stylesheet might set `font-size: 1em` for all paragraphs, while another author declaration sets `font-size: 2em` for elements with class "large":
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
 
-You can explicitly set the initial value by using the {{cssxref("initial")}} keyword.
+```css
+p {
```
estelle, line 105:
```suggestion
/* User agent styles */
p {
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+  font-size: 1em;
+}
 
-> [!NOTE]
-> The initial value for can be found in the formal syntax section of each CSS property reference page. For example, the [initial value of `font-size` is `medium`](/en-US/docs/Web/CSS/font-size#formal_definition). The initial value should not be confused with the value specified by the browser's style sheet.
+p {
```
estelle, line 110:
```suggestion
/* author styles */
p {
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+  font-size: 1.25em;
+}
+
+.large {
+  font-size: 2em;
+}
+```
+
+Only declarations whose selectors match the element become declared values. In this example, if our `<p>` element has `class="large"`, then all three declarations would be declared values for that element.
```
estelle:
```suggestion
There may be many other `font-size` declarations in our stylesheets, but only declarations whose selectors match the element become declared values. In this example, as our `<p>` element has `class="large"`, all three declarations are declared values for this element.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+
+### Cascaded value
+
+The **cascaded value** is the declared value that wins the [cascade](#cascading). There is at most one cascaded value per property per element.
+
+From our declared values above, the cascaded value would be `font-size: 2em` from the author origin with specificity 0,1,1:
```
estelle:
```suggestion
From our declared values, author styles win over user agent styles. Within the same origin, higher specifity styles win over lower specificity styles. In this case, the cascaded value would be `font-size: 2em`, from the author origin with specificity `0-1-1`:
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+
+```css
+font-size: 2em;
+```
+
+If there are no declared values for a property, there is no cascaded value, which means the value will be determined by the [defaulting](#defaulting) process.
```
estelle:
```suggestion
If there are no declared values for a property, there is no cascaded value, which means the [specified value](#specified_value) for that property is determined by the [defaulting](#defaulting) process.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
 
 ### Specified value
 
-The **specified value** is the value initially assigned in the CSS file or by the `style` attribute. The specified value for a given property is determined according to the following rules:
+The **specified value** is the result of the [defaulting](#defaulting) process. It is guaranteed to exist for every property on every element. The specified value is determined as follows:
 
-1. If the document's style sheet explicitly specifies a value for the property, the given value will be used.
-2. If the document's style sheet doesn't specify a value but it is an inherited property, the value will be taken from the parent element.
-3. If none of the above apply, the element's [initial value](#initial_value) will be used.
+1. If there is a [cascaded value](#cascaded_value), it becomes the specified value.
```
estelle:
```suggestion
1. If there is a [cascaded value](#cascaded_value), the cascaded value is the specified value.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+2. If there is _no_ [cascaded value](#cascaded_value) and the property is inherited, the specified value is the [computed value](#computed_value) of the parent element.
```
estelle:
```suggestion
2. If there is _no_ cascaded value and the property is [inherited](/en-US/docs/Web/CSS/CSS_cascade/Inheritance), the specified value is the [computed value](#computed_value) of the parent element.
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
+3. If there is _no_ [cascaded value](#cascaded_value) and the property is _not_ inherited, the specified value is the property's [initial value](#initial_value).
```
estelle:
```suggestion
3. If there is _no_ cascaded value and the property is _not_ inherited, the specified value is the property's [initial value](#initial_value).
```

files/en-us/web/css/css_cascade/value_processing/index.md
```diff
@@ -10,50 +10,152 @@ For every element in a document tree, the browser assigns a value to every CSS p
...
 
-In the example, `p { font-size: 1.25em; }`, the specified value is `1.25em`, unless the codebase includes a `font-size` declaration with greater {{cssxref("specificity")}}.
+In our example, since we have a [cascaded value](#cascaded_value) of `2em`, this becomes the specified value:
+
+```css
+font-size: 2em;
+```
+
+For properties without [cascaded values](#cascaded_value), the [defaulting](#defaulting) process fills in the gaps. For example, if `color` is not specified, it inherits from the parent's computed value since it's an inherited property. If `margin` is not specified, it uses its initial value of `0` since it's a non-inherited property:
```
estelle:
```suggestion
For properties without cascaded values, the defaulting process determines the value. For example, if `color` is not specified, the `color` is inherited from the parent's computed value since it's an inherited property. If `margin` is not specified, the `initial` value of `0` is used as `margin` is not an [inherited property](/en-US/docs/Web/CSS/CSS_cascade/Inheritance#inherited_properties):
```

estelle (commented):
a few more suggestions and questions

files/en-us/web/css/css_cascade/value_processing/index.md
sakupi01: @estelle 
Thank you!

`initial` or `inherit` **keywords itself** are not going to be set by CSS parser if I am being accurate to the spec. It could be any initial value defined in spec(, which is not limited to `initial`) or inherited from the parent (, which is also not limited to `inherit`), if I read the spec correctly:
https://www.w3.org/TR/css-cascade-4/#defaulting

I also thought using the word "The CSS parser" here is misleading. The parsing phase corresponds to Filtering phase, which is 2 phases before Defaulting phase.

Also, explicit defaulting keywords also contributes to set specified values which is a result of Defaulting process here. (Each explicit defaulting keywords are also converted to the proper specified values.)
So, I thought it'd be better to mention about explicit defaulting keywords conversion, not entirely deleting.

What do you think about: [fixed defaulting description with accuracy](https://github.com/mdn/content/pull/40546/commits/864427f20421d4db519c7279f96a6b252becd00a)?

sakupi01:
Thank you very much for spending time, @estelle !

I made changes of: 
[grammar or english](https://github.com/mdn/content/pull/40546/commits/9702e1ee1afadfb039950b928fe2b561ed20ae06), 
[defaulting description with accuracy](https://github.com/mdn/content/pull/40546/commits/864427f20421d4db519c7279f96a6b252becd00a),
and [CSS value processing index with 'Actual value'](https://github.com/mdn/content/pull/40546/commits/af0de08998abf7286cbd3f883158e6e1eccb0352).

I hope they make sense!

estelle (approved):
Thank you!

estelle (approved):
Thank you! 
Yay. That was quite the first content PR. Welcome to the team!

PR #40547, opened 2025-07-27
New CSSOM view API landing page, move CSS guides under

Josh-Cena:
I've been long bothered by the fact that "CSSOM view" is not a real API group on MDN. It makes our IA extremely awkward because existing references either have get the sidebar of its owner interface (such as "DOM" for `Window` and `Document` extensions, or "UI events" for `MouseEvent` extensions) or be fakely labeled as "CSSOM". "Visual Viewport API" is also not a real spec anymore. So we basically hacked around the whole issue of lacking a proper API group.

Meanwhile we do have a "CSSOM view" landing page, but it's confusingly categorized as CSS, which means it cannot be used as the landing page of a web API group, because of the way our groupdata.json works. All the other CSS specs that define JS bindings, such as CSSOM, CSS Typed OM, etc. are under web API, so I think this should be no exception: the CSS docs should be left to the CSS language proper, not for JS bindings.

I'm adding the groupdata for the new interfaces only. There are too many interface extensions and I'm not interested in updating their sidebars in this PR, although a future one might.


estelle:
We should probably move this to a discussion.

I just recently (last week) added https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_custom_highlight_API to the docs. We also have a https://developer.mozilla.org/en-US/docs/Web/API/CSS_custom_highlight_API page. They serve different purposes.

Similarly, https://developer.mozilla.org/en-US/docs/Web/CSS/CSSOM_view could have a https://developer.mozilla.org/en-US/docs/Web/API/CSSOM_view page (we have https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model, so we do have it).

The page should not be moved / removed from the CSS module landing pages section because CSSOM is a CSS module. Every CSS spec should have a CSS module landing page. I've been working on that for .... checks calendar .... almost 2 years?

Josh-Cena:
If a CSS spec does not define any at-rules, selectors, properties, functions, or types, then it would have nothing to be associated with in the CSS docs. In the future when we have features nested under the module folders, these modules would weirdly have nothing under them because well they are not CSS. CSS highlight API does have both CSS and API, so it should rightfully be documented under both, but CSSOM view is purely a web API that just happens to be standardized by CSSWG. For example, the websec group also publishes API bindings such as the reporting API, and that is documented as a web API as opposed to an HTTP feature.

Thinking about it from another perspective: anything called "X object model" inherently is a web API, not "X" itself. For example DOM is not nested under XML or HTML, and neither is SVG DOM or HTML DOM. They are all inherently bindings in the API realm that enable working with the said language. Similarly anything branded as "CSSOM" is not "CSS", but a foreign interface for interacting with CSS.

Rumyra:
Update as said in the last editorial:

I think we should keep the module page (as it is a CSS module page type) but review the documentation of this under the API section and make sure it is complete there as well 👍

Josh-Cena:
Porting over my question from Discord: What's the definition of a "CSS module"? Why can something that defines no CSS language feature be considered a CSS module?

estelle:
We should include a module landing page in the CSS module landing page section for all CSSWG specifications. The CSS module pages include the IDLs introduced in the spec, so omitting the a spec because it doesn't include properties would mean those are the only CSS spec IDLs not included in the CSS module section.

All the CSSWG specs are modules, but modules are not limited to the CSSWG . Specs from https://drafts.fxtf.org like masking, filtering, motion, etc., are also modules.

Josh-Cena:
Fair enough. I have no intention to argue with that decision. I've restored the "CSSOM view" module landing page. However, I would like the guides to live under API, because they talk about the semantics of the JavaScript properties and not any CSS feature.

files/en-us/_redirects.txt
```diff
@@ -12141,6 +12141,8 @@
 /en-US/docs/Web/CSS/Aural	/en-US/docs/Web/CSS/@media
 /en-US/docs/Web/CSS/Block_formatting_context	/en-US/docs/Web/CSS/CSS_display/Block_formatting_context
 /en-US/docs/Web/CSS/CSS3_Columns	/en-US/docs/Web/CSS/CSS_multicol_layout/Using_multicol_layouts
+/en-US/docs/Web/CSS/CSSOM_view/Coordinate_systems	/en-US/docs/Web/API/CSSOM_view_API/Coordinate_systems
+/en-US/docs/Web/CSS/CSSOM_view/Viewport_concepts	/en-US/docs/Web/API/CSSOM_view_API/Viewport_concepts
```
estelle: OK to move  Coordinate_systems but not Viewport_concepts

estelle (changes requested):
The viewport concepts is about JS, HTML, SVG, and CSS, and is very relevant to several CSS concepts, including viewport related values, the CSS viewport module, and the CSS environment variables module. The viewport concepts guide needs to remain in the CSS section.

The coordinate system is related to SVG, JS, and CSS (think <basic-shapes> all the properties that accept a shape as a value. The example on the coordinate system page is more JS than CSS, so can be moved to the CSSOM View API section under APIs.,

estelle (approved):

PR #40548, opened 2025-07-27
Provide more guidelines about labeling input

### LINKED ISSUE #37665: Better documentation for labelling inputs
The discussion at https://github.com/orgs/mdn/discussions/763 proposed changing example code on MDN from "explicit labels" (in which input elements are associated with labels using `for` and `id` attributes) to "implicit labels" (in which the association is defined by nesting the input inside the label.

Discussion suggests that we don't want to do this, mainly because (according to https://css-tricks.com/html-inputs-and-labels-a-love-story/): "an implicit label is not handled correctly by all assistive technologies".

However we don't document this in the reference page for [`<label>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label): we do describe both forms but don't give any guidance about which to use. We ought to do this.

I'd love to know more concrete details about the implications of this choice for assistive technologies and wonder if @scottaohara or @ericwbailey might be able to help with this <3.

Josh-Cena:
Fix https://github.com/mdn/content/issues/37665, and its two linked issues, which as I work through this PR realized are not actually duplicates of https://github.com/mdn/content/issues/37665.

### DIFF
--- a/files/en-us/web/html/reference/elements/label/index.md
+++ b/files/en-us/web/html/reference/elements/label/index.md
@@ -12,12 +12,12 @@ The **`<label>`** [HTML](/en-US/docs/Web/HTML) element represents a caption for 
 
 ```html interactive-example
 <div class="preference">
-  <label for="cheese">Do you like cheese?</label>
+  <label for="cheese">I like cheese.</label>
   <input type="checkbox" name="cheese" id="cheese" />
 </div>
 
 <div class="preference">
-  <label for="peas">Do you like peas?</label>
+  <label for="peas">I like peas.</label>
   <input type="checkbox" name="peas" id="peas" />
 </div>
 ```
@@ -64,7 +64,7 @@ Alternatively, you can nest the `<input>` directly inside the `<label>`, in whic
 
 ```html
 <label>
-  Do you like peas?
+  I like peas.
   <input type="checkbox" name="peas" />
 </label>
 ```
--- a/files/en-us/web/html/reference/elements/label/index.md
+++ b/files/en-us/web/html/reference/elements/label/index.md
@@ -36,25 +36,15 @@ The **`<label>`** [HTML](/en-US/docs/Web/HTML) element represents a caption for 
 This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Global_attributes).
 
 - [`for`](/en-US/docs/Web/HTML/Reference/Attributes/for)
-  - : The value of the `for` attribute must be a single [`id`](/en-US/docs/Web/HTML/Reference/Global_attributes/id) for a [labelable](/en-US/docs/Web/HTML/Guides/Content_categories#labelable) form-related element in the same document as the `<label>` element. So, any given `label` element can be associated with only one form control.
-
-    > [!NOTE]
-    > To programmatically set the `for` attribute, use [`htmlFor`](/en-US/docs/Web/API/HTMLLabelElement/htmlFor).
-
-    The first element in the document with an `id` attribute matching the value of the `for` attribute is the _labeled control_ for this `label` element — if the element with that `id` is actually a [labelable element](https://html.spec.whatwg.org/multipage/forms.html#category-label). If it is _not_ a labelable element, then the `for` attribute has no effect. If there are other elements that also match the `id` value, later in the document, they are not considered.
-
-    Multiple `label` elements can be given the same value for their `for` attribute; doing so causes the associated form control (the form control that `for` value references) to have multiple labels.
-
-    > [!NOTE]
-    > A `<label>` element can have both a `for` attribute and a contained control element, as long as the `for` attribute points to the contained control element.
+  - : The value is the [`id`](/en-US/docs/Web/HTML/Reference/Global_attributes/id) of the [labelable](/en-US/docs/Web/HTML/Guides/Content_categories#labelable) form control in the same document, [associating the `<label>` with that form control](#associating_a_label_with_a_form_control). Note that its JavaScript reflection property is [`htmlFor`](/en-US/docs/Web/API/HTMLLabelElement/htmlFor).
 
 ## Usage notes
 
-### Styling with CSS
-
-There are no special styling considerations for `<label>` elements — structurally they are inline elements, and so can be styled in much the same way as a {{htmlelement("span")}} or {{htmlelement("a")}} element. You can apply styling to them in any way you want, as long as you don't cause the text to become difficult to read.
-
 ### Associating a label with a form control
+
+The first element in the document with an `id` attribute matching the value of the `for` attribute is the _labeled control_ for this `label` element — if the element with that `id` is actually a [labelable element](/en-US/docs/Web/HTML/Guides/Content_categories#labelable). If it is _not_ a labelable element, then the `for` attribute has no effect. If there are other elements that also match the `id` value, later in the document, they are not considered.
+
+Multiple `label` elements can be given the same value for their `for` attribute; doing so causes the associated form control (the form control that `for` value references) to have multiple labels.
 
 Associating a `<label>` with a form control, such as {{htmlelement("input")}} or {{htmlelement("textarea")}} offers some major advantages:
 
@@ -79,11 +69,14 @@ Alternatively, you can nest the `<input>` directly inside the `<label>`, in whic
 </label>
 ```
 
-These two methods are equivalent in modern browsers. However, there are some other considerations:
+> [!NOTE]
+> A `<label>` element can have both a `for` attribute and a contained control element, as long as the `for` attribute points to the contained control element.
 
-- Not all assistive technologies support implicit association.
-- Implicit association may be harder to style because the label box contains the form control instead of being separate elements.
-- When using component frameworks with explicit association, it's often hard to ensure that the control has a globally unique `id` (React introduced the [`useId()`](https://react.dev/reference/react/useId) hook for this purpose).
+These two methods are equivalent, but there are a few considerations:
+
+- While common browser and {{glossary("screen reader")}} combinations support implicit association, not all assistive technologies do.
+- Depending on your design, the type of association may impact stylability. Explicit association makes the label and form control separate boxes, enabling more customizable layout such as lining them up with grid or flex layout methods.
+- Explicit association requires the form control to have an `id`, which must be unique in the whole document. This is hard especially in a componentized application. Frameworks often provide their own solutions, such as React's [`useId()`](https://react.dev/reference/react/useId), but it still requires extra orchestration to get right.
 
 Generally, we recommend using explicit association with the `for` attribute, to ensure compatibility with external tools and assistive technologies. In fact, you can simultaneously nest _and_ provide `id`/`for` for maximum compatibility.
 
@@ -125,7 +118,7 @@ Don't place additional interactive elements such as {{HTMLElement("a", "anchors"
 ```
 
 > [!NOTE]
-> It is also a good practice to place any necessary context, such as the link to the terms and conditions, before the form control, so that the user can read it before they interact with the control.
+> It is a good practice to place any necessary context, such as the link to the terms and conditions, before the form control, so that the user can read it before they interact with the control.
 
 ### Headings
 
--- a/files/en-us/web/html/reference/elements/label/index.md
+++ b/files/en-us/web/html/reference/elements/label/index.md
@@ -44,7 +44,7 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
 
 The first element in the document with an `id` attribute matching the value of the `for` attribute is the _labeled control_ for this `label` element — if the element with that `id` is actually a [labelable element](/en-US/docs/Web/HTML/Guides/Content_categories#labelable). If it is _not_ a labelable element, then the `for` attribute has no effect. If there are other elements that also match the `id` value, later in the document, they are not considered.
 
-Multiple `label` elements can be given the same value for their `for` attribute; doing so causes the associated form control (the form control that `for` value references) to have multiple labels.
+Multiple `<label>` elements can be associated with the same form control by having multiple `<label>` elements with the same `for` attribute value, which gives the form control multiple labels.
 
 Associating a `<label>` with a form control, such as {{htmlelement("input")}} or {{htmlelement("textarea")}} offers some major advantages:
 
@@ -75,7 +75,7 @@ Alternatively, you can nest the `<input>` directly inside the `<label>`, in whic
 These two methods are equivalent, but there are a few considerations:
 
 - While common browser and {{glossary("screen reader")}} combinations support implicit association, not all assistive technologies do.
-- Depending on your design, the type of association may impact stylability. Explicit association makes the label and form control separate boxes, enabling more customizable layout such as lining them up with grid or flex layout methods.
+- Depending on your design, the type of association may impact stylability. Making the `<label>` and form control sibling elements instead of parent-child means they are separate, adjacent boxes, enabling more customizable layout such as lining them up with grid or flex layout methods.
 - Explicit association requires the form control to have an `id`, which must be unique in the whole document. This is hard especially in a componentized application. Frameworks often provide their own solutions, such as React's [`useId()`](https://react.dev/reference/react/useId), but it still requires extra orchestration to get right.
 
 Generally, we recommend using explicit association with the `for` attribute, to ensure compatibility with external tools and assistive technologies. In fact, you can simultaneously nest _and_ provide `id`/`for` for maximum compatibility.
@@ -94,7 +94,7 @@ Elements that can be associated with a `<label>` element include {{HTMLElement('
 
 ### Interactive content
 
-Don't place additional interactive elements such as {{HTMLElement("a", "anchors")}} or {{HTMLElement("button", "buttons")}} inside a `label` (the only interactive child should be the element being labeled, if you are using [implicit association](#associating_a_label_with_a_form_control)). Doing so makes it difficult for people to activate the form input associated with the `label`.
+Other than the [implicitly associated](#associating_a_label_with_a_form_control) form control, don't place additional interactive elements such as {{HTMLElement("a", "anchors")}} or {{HTMLElement("button", "buttons")}} inside a `<label>`. Doing so makes it difficult for people to activate the form input associated with the `label`.
 
 **Don't do this:**
 
--- a/files/en-us/web/html/reference/elements/label/index.md
+++ b/files/en-us/web/html/reference/elements/label/index.md
@@ -31,32 +31,6 @@ The **`<label>`** [HTML](/en-US/docs/Web/HTML) element represents a caption for 
 }
 ```
 
-Associating a `<label>` with a form control, such as {{htmlelement("input")}} or {{htmlelement("textarea")}} offers some major advantages:
-
-- The label text is not only visually associated with its corresponding text input; it is programmatically associated with it too. This means that, for example, a screen reader will read out the label when the user is focused on the form input, making it easier for an assistive technology user to understand what data should be entered.
-- When a user clicks or touches/taps a label, the browser passes the focus to its associated input (the resulting event is also raised for the input). That increased hit area for focusing the input provides an advantage to anyone trying to activate it — including those using a touch-screen device.
-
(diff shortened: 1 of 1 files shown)


files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -31,13 +31,45 @@ The **`<label>`** [HTML](/en-US/docs/Web/HTML) element represents a caption for
 }
 ```
 
+## Attributes
+
+This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Global_attributes).
+
+- [`for`](/en-US/docs/Web/HTML/Reference/Attributes/for)
+  - : The value of the `for` attribute must be a single [`id`](/en-US/docs/Web/HTML/Reference/Global_attributes/id) for a [labelable](/en-US/docs/Web/HTML/Guides/Content_categories#labelable) form-related element in the same document as the `<label>` element. So, any given `label` element can be associated with only one form control.
```
estelle:
```suggestion
  - : The value is the [`id`](/en-US/docs/Web/HTML/Reference/Global_attributes/id) of the [labelable](/en-US/docs/Web/HTML/Guides/Content_categories#labelable) form control in the same document, associating the `<label>` with that form control.
  
## Usage notes
```
suggested shrinkage


Alternatively, it can end with
`associating the two.` 
instead of
`associating the <label> with that form control.`
which is shorter, but may be less clear.

estelle: The rest of this section should be under a "description" or "usage notes" header, rather than under "attributes", which is a concise overview of non-global attributes.

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -31,13 +31,45 @@ The **`<label>`** [HTML](/en-US/docs/Web/HTML) element represents a caption for
...
+    > To programmatically set the `for` attribute, use [`htmlFor`](/en-US/docs/Web/API/HTMLLabelElement/htmlFor).
+
+    The first element in the document with an `id` attribute matching the value of the `for` attribute is the _labeled control_ for this `label` element — if the element with that `id` is actually a [labelable element](https://html.spec.whatwg.org/multipage/forms.html#category-label). If it is _not_ a labelable element, then the `for` attribute has no effect. If there are other elements that also match the `id` value, later in the document, they are not considered.
+
+    Multiple `label` elements can be given the same value for their `for` attribute; doing so causes the associated form control (the form control that `for` value references) to have multiple labels.
+
+    > [!NOTE]
+    > A `<label>` element can have both a `for` attribute and a contained control element, as long as the `for` attribute points to the contained control element.
+
+## Usage notes
```
estelle, line 41:
```suggestion
```
moved up

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -31,13 +31,45 @@ The **`<label>`** [HTML](/en-US/docs/Web/HTML) element represents a caption for
...
+
+    > [!NOTE]
+    > To programmatically set the `for` attribute, use [`htmlFor`](/en-US/docs/Web/API/HTMLLabelElement/htmlFor).
+
+    The first element in the document with an `id` attribute matching the value of the `for` attribute is the _labeled control_ for this `label` element — if the element with that `id` is actually a [labelable element](https://html.spec.whatwg.org/multipage/forms.html#category-label). If it is _not_ a labelable element, then the `for` attribute has no effect. If there are other elements that also match the `id` value, later in the document, they are not considered.
+
+    Multiple `label` elements can be given the same value for their `for` attribute; doing so causes the associated form control (the form control that `for` value references) to have multiple labels.
+
+    > [!NOTE]
+    > A `<label>` element can have both a `for` attribute and a contained control element, as long as the `for` attribute points to the contained control element.
```
estelle:
```suggestion

### The `for` attribute

Include the `for` attribute to associate the `<label>` with a single form control. The value of the `for` attribute is the unique [`id`](/en-US/docs/Web/HTML/Reference/Global_attributes/id) of the [labelable](/en-US/docs/Web/HTML/Guides/Content_categories#labelable) form-related element within the same document as the `<label>`. Every `<label>` element can be associated with zero or one form controls.

The first element in the document with an `id` attribute matching the value of the `for` attribute is the _labeled control_ for this `label` element — if the element with that `id` is actually a [labelable element](https://html.spec.whatwg.org/multipage/forms.html#category-label). If it is _not_ a labelable element, then the `for` attribute has no effect. If there are other elements that also match the `id` value, later in the document, they are not considered.

Multiple `label` elements can be given the same value for their `for` attribute; doing so causes the associated form control (the form control that `for` value references) to have multiple labels.

To programmatically set the `for` attribute, use [`htmlFor`](/en-US/docs/Web/API/HTMLLabelElement/htmlFor).

An implicit `<label>` element, a label that has a form control as a descendant, can have both the `for` attribute and the contained control element, as long as the `for` attribute value is the `id` of that nested form control.
```

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -31,13 +31,45 @@ The **`<label>`** [HTML](/en-US/docs/Web/HTML) element represents a caption for
...
+
+## Usage notes
+
+### Styling with CSS
+
+There are no special styling considerations for `<label>` elements — structurally they are inline elements, and so can be styled in much the same way as a {{htmlelement("span")}} or {{htmlelement("a")}} element. You can apply styling to them in any way you want, as long as you don't cause the text to become difficult to read.
```
estelle: I realize this was just moved, but it can be deleted. It's basically a section that says "nothing special to see here". We can add it in if we decide that every element should have a "styling with CSS" section. The last section, which is important, should be added to accessibility concerns

```suggestion
```

files/en-us/web/html/reference/elements/label/index.md
estelle: Actually, this should all be moved to the "associating with form control" section

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -47,6 +79,14 @@ Alternatively, you can nest the `<input>` directly inside the `<label>`, in whic
 </label>
 ```
 
+These two methods are equivalent in modern browsers. However, there are some other considerations:
```
estelle:
```suggestion
These two methods are equivalent, but there are a few considerations:
```

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -47,6 +79,14 @@ Alternatively, you can nest the `<input>` directly inside the `<label>`, in whic
...
+
+- Not all assistive technologies support implicit association.
```
estelle:
```suggestion
- While common browser and {{glossary("screen reader")}} combinations support implicit labels, not all assistive technologies support implicit association in all user agents.
```
this is kind of a one off case of non-support that should be fixed rather than developers catering to a specific instance.

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -47,6 +79,14 @@ Alternatively, you can nest the `<input>` directly inside the `<label>`, in whic
...
+- Implicit association may be harder to style because the label box contains the form control instead of being separate elements.
```
estelle: could also be easier to style. maybe something like this:

```suggestion
- Depending on your design, the type of association may impact the stylability; with an implicit association, the control is nested ensuring the visual association, while explicit associations enable the label and form controls to be adjacent, enabling lining up the elements with grid or flex layout methods.
```
There is little inheritance with form controls, so that's not an issue. I assume that the design issue is that flex and grid can't be used. There used to be an issue in IE6 or 7, which I blogged about over 15 years ago, where the browser divided the implicit label into two elements, causing duplicated backgrounds and causing borders to close and reopen, but that is old news.

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -47,6 +79,14 @@ Alternatively, you can nest the `<input>` directly inside the `<label>`, in whic
...
+- When using component frameworks with explicit association, it's often hard to ensure that the control has a globally unique `id` (React introduced the [`useId()`](https://react.dev/reference/react/useId) hook for this purpose).
```
estelle: hesitant to include this as IDs should always be unique and adding react's fix is like saying they're a little less shitty than other options.  Maybe something like:
```suggestion
- Implicit labels do not require a form control to have an `id`. Every `id` in a document should be unique. However, some component frameworks may not ensure every form control has a unique `id`. While this is bad in general, it can make the resulting output inaccessible.
```

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -47,6 +79,14 @@ Alternatively, you can nest the `<input>` directly inside the `<label>`, in whic
...
+
+Generally, we recommend using explicit association with the `for` attribute, to ensure compatibility with external tools and assistive technologies. In fact, you can simultaneously nest _and_ provide `id`/`for` for maximum compatibility.
```
estelle, line 81:
```suggestion
It is considered a best practice to explicitly associate every `<label>` with a form control using the `for` attribute even if that form control is nested within the `<label>.
```

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -96,15 +115,18 @@ Don't place interactive elements such as {{HTMLElement("a", "anchors")}} or {{HT
...
   <input id="tac" type="checkbox" name="terms-and-conditions" />
   I agree to the Terms and Conditions
 </label>
-<p>
-  <a href="terms-and-conditions.html">Read our Terms and Conditions</a>
-</p>
 ```
 
+> [!NOTE]
+> It is also a good practice to place any necessary context, such as the link to the terms and conditions, before the form control, so that the user can read it before they interact with the control.
```
estelle:
```suggestion
> It is a good practice to place any necessary context, such as the link to the terms and conditions, before the form control, so that the user can read it before they interact with the control.
```

estelle (commented):
lots of feedback, including on stuff you just moved (sorry)

files/en-us/web/html/reference/elements/label/index.md
Josh-Cena: IMO "in modern browsers" is important. (1) Legacy browsers are generally irrelevant, but we are talking about a11y anyway, and supporting more browsers is part of a11y; (2) We talk about non-browser consumers of the HTML, so mentioning "browsers" is also important, otherwise saying "they are functionally equivalent, except in certain cases they are not functionally equivalent" is weird.

files/en-us/web/html/reference/elements/label/index.md
Josh-Cena: I was thinking about, for example, the label and the input being visually separate, in which case having a nested box means you have a box straddling between them that pushes everything else away. Sort of a weird headache I run into often.

files/en-us/web/html/reference/elements/label/index.md
Josh-Cena: It's straight out invalid HTML and should be treated as a syntax error, which is worse than "inaccessible", so I was phrasing it as a first principle that IDs must be unique. The point is how we mitigate such issues for framework users (which represent 99% of developers), and I'm providing React as an example, which probably represents 50%+ of our readers.

files/en-us/web/html/reference/elements/label/index.md
Josh-Cena: I don't think it's universally considered a best practice due to the third point about unique IDs; it's just our own recommendation

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -47,6 +69,17 @@ Alternatively, you can nest the `<input>` directly inside the `<label>`, in whic
 </label>
 ```
 
+> [!NOTE]
+> A `<label>` element can have both a `for` attribute and a contained control element, as long as the `for` attribute points to the contained control element.
+
+These two methods are equivalent, but there are a few considerations:
+
+- While common browser and {{glossary("screen reader")}} combinations support implicit association, not all assistive technologies do.
+- Depending on your design, the type of association may impact stylability. Explicit association makes the label and form control separate boxes, enabling more customizable layout such as lining them up with grid or flex layout methods.
```
estelle:
```suggestion
- Depending on your design, the type of association may impact stylability. Sibling `<label>` and form control pairs that are explicitly associated means the label and form control are separate, adjacent boxes, enabling more customizable layout such as lining them up with grid or flex layout methods.
```
it's being siblings that makes it easier, not the association.

files/en-us/web/html/reference/elements/label/index.md
estelle: i don't think "we" should "recommend" stuff, especially if we don't officially have a meeting to decide we're making a recommendation.

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -96,15 +108,18 @@ Don't place interactive elements such as {{HTMLElement("a", "anchors")}} or {{HT
...
   <input id="tac" type="checkbox" name="terms-and-conditions" />
   I agree to the Terms and Conditions
 </label>
-<p>
-  <a href="terms-and-conditions.html">Read our Terms and Conditions</a>
-</p>
 ```
 
+> [!NOTE]
+> It is a good practice to place any necessary context, such as the link to the terms and conditions, before the form control, so that the user can read it before they interact with the control.
```
estelle, line 121: not the norm for radio and checkboxes. https://www.w3.org/TR/WCAG20-TECHS/G162.html#G162-description, and we just did the opposite 5 lines before

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -57,32 +90,11 @@ The form control that a label is labeling is called the _labeled control_ of the
...
-## Styling with CSS
-
-There are no special styling considerations for `<label>` elements — structurally they are inline elements, and so can be styled in much the same way as a {{htmlelement("span")}} or {{htmlelement("a")}} element. You can apply styling to them in any way you want, as long as you don't cause the text to become difficult to read.
-
 ## Accessibility
 
 ### Interactive content
 
-Don't place interactive elements such as {{HTMLElement("a", "anchors")}} or {{HTMLElement("button", "buttons")}} inside a `label`. Doing so makes it difficult for people to activate the form input associated with the `label`.
+Don't place additional interactive elements such as {{HTMLElement("a", "anchors")}} or {{HTMLElement("button", "buttons")}} inside a `label` (the only interactive child should be the element being labeled, if you are using [implicit association](#associating_a_label_with_a_form_control)). Doing so makes it difficult for people to activate the form input associated with the `label`.
```
estelle:
```suggestion
Other than the [implicitly associated](#associating_a_label_with_a_form_control) form control, don't place additional interactive elements such as {{HTMLElement("a", "anchors")}} or {{HTMLElement("button", "buttons")}} inside a `<label>`. Doing so makes it difficult for people to activate the form input associated with the `label`.
```
fewer words

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -31,13 +31,35 @@ The **`<label>`** [HTML](/en-US/docs/Web/HTML) element represents a caption for
...
+- [`for`](/en-US/docs/Web/HTML/Reference/Attributes/for)
+  - : The value is the [`id`](/en-US/docs/Web/HTML/Reference/Global_attributes/id) of the [labelable](/en-US/docs/Web/HTML/Guides/Content_categories#labelable) form control in the same document, [associating the `<label>` with that form control](#associating_a_label_with_a_form_control). Note that its JavaScript reflection property is [`htmlFor`](/en-US/docs/Web/API/HTMLLabelElement/htmlFor).
+
+## Usage notes
+
+### Associating a label with a form control
+
+The first element in the document with an `id` attribute matching the value of the `for` attribute is the _labeled control_ for this `label` element — if the element with that `id` is actually a [labelable element](/en-US/docs/Web/HTML/Guides/Content_categories#labelable). If it is _not_ a labelable element, then the `for` attribute has no effect. If there are other elements that also match the `id` value, later in the document, they are not considered.
+
+Multiple `label` elements can be given the same value for their `for` attribute; doing so causes the associated form control (the form control that `for` value references) to have multiple labels.
```
estelle:
```suggestion
Multiple `<label>` elements can be associated with the same form control by having multiple `<label>` elements with the same `for` attribute value, which gives the form control multiple labels.
```
While `<label>` is a tag, not an element name, i think we always list html names as tags to differentiate elements from attributes and other features.

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -31,13 +31,35 @@ The **`<label>`** [HTML](/en-US/docs/Web/HTML) element represents a caption for
...
 
 - The label text is not only visually associated with its corresponding text input; it is programmatically associated with it too. This means that, for example, a screen reader will read out the label when the user is focused on the form input, making it easier for an assistive technology user to understand what data should be entered.
 - When a user clicks or touches/taps a label, the browser passes the focus to its associated input (the resulting event is also raised for the input). That increased hit area for focusing the input provides an advantage to anyone trying to activate it — including those using a touch-screen device.
 
+There are two ways to associate a `<label>` with a form control, commonly referred to as _explicit_ and _implicit_ association.
+
 To explicitly associate a `<label>` element with an `<input>` element, you first need to add the `id` attribute to the `<input>` element. Next, you add the `for` attribute to the `<label>` element, where the value of `for` is the same as the `id` in the `<input>` element.
 
+```html
+<label for="peas">Do you like peas?</label>
```
estelle: This is a yes / no, so would be the `<legend>` of a `<fieldset>` with a "yes" for the label, or, alternatively, two different radio buttons or checkboxes (yes and no) with the label after the checkbox..

files/en-us/web/html/reference/elements/label/index.md
```diff
@@ -96,15 +108,18 @@ Don't place interactive elements such as {{HTMLElement("a", "anchors")}} or {{HT
...
-  <a href="terms-and-conditions.html">Read our Terms and Conditions</a>
-</p>
 ```
 
+> [!NOTE]
+> It is a good practice to place any necessary context, such as the link to the terms and conditions, before the form control, so that the user can read it before they interact with the control.
+
 ### Headings
 
 Placing [heading elements](/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements) within a `<label>` interferes with many kinds of assistive technology, because headings are commonly used as [a navigation aid](/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements#navigation). If the label's text needs to be adjusted visually, use CSS classes applied to the `<label>` element instead.
```
estelle, line 125: the next line talkes about legend and fieldset, but the example-good doesn't include them. this needs work too. not sure if you want to add to this PR.

estelle (commented):
more feedback

files/en-us/web/html/reference/elements/label/index.md
Josh-Cena: We do recommend things all the time though. For example we say "don't use `eval()`", or "prefer `rgb()` to `rgba()`", or "use `Temporal` instead of `Date`". In this case the reasoning is well-founded (explicit association has more support). Remember that not all ATs are full HTML implementations and they are not required to be.

files/en-us/web/html/reference/elements/label/index.md
Josh-Cena: I could get away with radio buttons, but I would argue that a checkbox is the most traditional UX for boolean inputs. Radios are for enumerated values.

files/en-us/web/html/reference/elements/label/index.md
Josh-Cena:
```suggestion
- Depending on your design, the type of association may impact stylability. Making the `<label>` and form control sibling elements instead of parent-child means they are separate, adjacent boxes, enabling more customizable layout such as lining them up with grid or flex layout methods.
```

files/en-us/web/html/reference/elements/label/index.md
Josh-Cena: The associated label can be thought of as "parallel" to the control and their order is irrelevant to screen reader because they are announced together. What I'm saying here is that you should have the ToC link above the label-control pair instead of below them, so screen reader users can reach them in the intended order

files/en-us/web/html/reference/elements/label/index.md
estelle: "I like peas" could get a checkmark, or "check if you like peas", but "do you like peas?" doesn't work.

Josh-Cena: Alright I could go with that :)

```suggestion
<label for="peas">I like peas.</label>
```

estelle (approved):

PR #40550, opened 2025-07-27
[CSS] Clarify that float has no effect when display: contents is applied

kannanwisen:
### Description

This PR adds a clarification note explaining that the float property also has no effect on elements with display: contents, to help readers better understand the limitations mentioned in the formal definition.

### DIFF
--- a/files/en-us/web/css/float/index.md
+++ b/files/en-us/web/css/float/index.md
@@ -120,6 +120,8 @@ The `float` property is specified as a single keyword, chosen from the list of v
 
 {{cssinfo}}
 
+**Note:** The `float` property has no effect when applied to elements with `display: none` or `display: contents`
+
 ## Formal syntax
 
 {{csssyntax}}


files/en-us/web/css/float/index.md
```diff
@@ -120,6 +120,8 @@ The `float` property is specified as a single keyword, chosen from the list of v
 
 {{cssinfo}}
 
+**Note:** The `float` property has no effect when applied to elements with `display: none` or `display: contents`
```
Josh-Cena, line 123: This section does not admit extra text. If you want to add freeform prose put it somewhere else

kannanwisen: Is it possible to change in

Applies to	all elements, but has no effect when the display property is none or contents..

Josh-Cena: I don't think so. "Applies to" is a different concept from "has an effect". Generally, whether something has an effect is a very ad-hoc note and you should probably put it in the intro, as a note, or immediately after where we talk about `display`.

kannanwisen:
That's a fair point regarding the conceptual distinction between the "Applies to" definition and a property's practical effects.

However, my proposal is based directly on the precedent already set in the current text for this very property. The existing "Applies to" line reads:

    all elements, but has no effect when the display property is none.

Since the specification already uses this "has no effect" phrasing in the "Applies to" line to describe the interaction with display: none, my suggestion is to simply make this statement more complete. Both display: none and display: contents create a situation where the element generates no principal box for the float property to act upon.

Given this existing structure, adding contents here seems to be the most consistent and logical location for this clarification.

Josh-Cena:
The CSSInfo is directly extracted from the CSS spec. We have very little say in its content. But it seems that the CSS 2 spec says "all, but see Relationships between 'display', 'position', and 'float'", so we could take your suggestion. For this we would modify mdn/data instead, and you would want to put in the whole section at https://www.w3.org/TR/2011/REC-CSS2-20110607/visuren.html#dis-pos-flo.

kannanwisen:
> The CSSInfo is directly extracted from the CSS spec. 

The CSS 2.2 Spec contains the below table.

image

And MDN Table has different content.

Kindly check.

Josh-Cena:
That's exactly what I'm saying though

kannanwisen:
> That's exactly what I'm saying though

The CSS 2.2 Spec says for 'Applies to':

Applies to: all, but see [9.7]

According to your conversation, this 'Applies to' value should be fetched from the CSS 2.2 Spec and displayed in the MDN Docs.

However, the MDN document displays:
Applies to: all elements, but has no effect if the value of display is none.".

kannanwisen:
> The CSSInfo is directly extracted from the CSS spec.

Please verify whether the information is being extracted from the correct specification document.

Josh-Cena:
I'm saying we manually extract it from the CSS specs, so in this case I agree with you that CSSInfo could be updated, and to do that you would be updating mdn/data instead. You would see how it works once you open that repo.

I'm also not sure if there's a spec that says exactly what you are proposing, so there would be no "specification document" to extract it from anyway; you are just manually fixing it yourself because you know it's correct.

kannanwisen:
Thank you for the pointer to the mdn/data repo. This is very helpful.

I have located the appliesto key ("allElementsNoEffectIfDisplayNone") in css/properties.schema.json and can see that it is validated by the schema.

However, I am having trouble locating the file where the text definition for this key is stored so that I can add a new one for contents. Could you please point me to the correct file to edit?

Thank you again for your guidance.

Josh-Cena:
You can do a project-wide search for this key.

kannanwisen:
Ok, Thanks,

Josh-Cena:
Closing and moving to mdn/data instead...

PR #40556, opened 2025-07-28
Fix bug in example code

Calculamatrise:
### Description

An error was found in the example code in the first positional argument for Promise#then ([1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then)). Instead of providing a function, a function call was provided which would run before the promise is resolved.

### Motivation

I believe this would help developers who are new to this feature and are trying to learn and test using the example code, as was I. An error would be thrown after attempting to use the example code provided here.

### References

1. [Promise#then](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then)


Josh-Cena (approved):
Great catch, thank you

PR #40559, opened 2025-07-28
Clarify setDate() behavior during timezone changes (fixes #40531)

### LINKED ISSUE #40531: Clarify effects of timezone changes in Date.setDate()
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate

Description

I think the docs could be clearer about what happens to a `Date` object when you shift it over a net timezone difference (e.g. the beginning or end of DST). True, it already states: 

> The setDate() method of [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) instances changes the day of the month for this date according to local time

But this is not a terribly clear description. In particular, it is not obvious what "according to local time" means. To someone struggling with unexpected effects - in this case, the difference between new and old timestamps being too short or too long -, this phrasing provides very little insight or help on what to do different.

In the Description, we might add something like this:

If the update moves the target date over a net timezone shift (e.g. over the beginning or end of Daylight Savings Time), `setDate` preserves local time: the hours value will be the same before and after the update. By extension, this means that the difference between the new and the old timestamp may vary. If you need to shift a date by a fixed amount of time, use {{{jsxref("Date.prototype.setUTCDate()")}} instead.

_No response_

I am willing to make this change in all affected methods and submit it as a pull request. But since this affects multiple pages, I would like some sort of confirmation that such a change would be welcomed before investing the work.

### MDN metadata

<details>

* Folder: `en-us/web/javascript/reference/global_objects/date/setutcdate`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setUTCDate
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/global_objects/date/setutcdate/index.md
* Last commit: https://github.com/mdn/content/commit/544b843570cb08d1474cfc5ec03ffb9f4edc0166
* Document last modified: 2025-07-10T09:07:55.000Z

</details>

Shrinivassab:
### Description

This PR clarifies the behavior of `Date.prototype.setDate()` during Daylight Saving Time (DST) transitions by adding explicit documentation about timezone shift effects. The changes explain how spring-forward and fall-back transitions affect elapsed time calculations and recommend using `setUTCDate()` for fixed time adjustments.

Fixes #40531


Josh-Cena:
@BadIdeaException what are the other pages you think we should add this note to?

BadIdeaException:
Any method that can potentially exhibit this problem, I would say. That means basically all the setters except:

* `setTime` because it only changes the timestamp
* `setFullYear` because setting the year _presumably_ can never lead to timezone changes? Or does this include cases where e.g. a country that used to have DST opted out of it at some point? Not an expert on how closely the `Date` API follows that sort of thing...

Josh-Cena:
Yeah a country changing TZ would fall into that. I will update the other pages.

Josh-Cena (approved):
The wording is really great. Thank you!

PR #40567, opened 2025-07-29
Update Japanese translations in jssidebar.yaml

mfuji09:
### Description

Update Japanese translations in jssidebar.yaml


dipikabh (approved):
Thank you!

PR #40568, opened 2025-07-29
Synchronize with BCD v6.0.36

OnkarRuikar:
[BCD release v6.0.36](https://github.com/mdn/browser-compat-data/releases/tag/v6.0.36)

/cc cc/ @bsmth @wbamberg


Josh-Cena:
I know this is autogenerated, but I'm going to close it because https://github.com/mdn/browser-compat-data/pull/27386 landed after this release, so all changes will be reverted by next release anyway, which doesn't seem productive to me.

OnkarRuikar:
On live, the 22 pages are not displaying the specs and correct BCD status. Can we request an early BCD release?

Josh-Cena:
I don't think it's a huge deal for them to be broken for a few days

PR #40572, opened 2025-07-30
Fix getAsFileSystemHandle() example

### LINKED ISSUE #36496: Broken example code for DataTransferItem.getAsFileSystemHandle()
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/getAsFileSystemHandle

Examples

The "Return value" section states the method must be "accessed in the dragstart or drop event handlers". However, the example code uses `await`, meaning that if more than one item is dropped, the second item's getAsFileSystemHandle will not be invoked "during" the event handler and will not return a file handle.

The example code should be updated so it accesses the items correctly during the event handler

_No response_

_No response_

Josh-Cena:
Fix https://github.com/mdn/content/issues/36496.

This makes this example consistent with https://wicg.github.io/file-system-access/#dom-datatransferitem-getasfilesystemhandle

### DIFF
--- a/files/en-us/web/api/datatransferitem/getasfilesystemhandle/index.md
+++ b/files/en-us/web/api/datatransferitem/getasfilesystemhandle/index.md
@@ -39,6 +39,9 @@ None.
 This example uses the `getAsFileSystemHandle()` method to return
 {{domxref('FileSystemHandle', 'file handles', '', 'nocode')}} for dropped items.
 
+> [!NOTE]
+> Because `getAsFileSystemHandle()` can only retrieve the entry handle the same tick as the `drop` event handler, there must be no `await` before it. This is why we synchronously invoke `getAsFileSystemHandle()` for all items first, and then wait for their results concurrently.
+
 ```js
 elem.addEventListener("dragover", (e) => {
   // Prevent navigation.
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/datatransferitem/getasfilesystemhandle/index.md
```diff
@@ -39,6 +39,9 @@ None.
 This example uses the `getAsFileSystemHandle()` method to return
 {{domxref('FileSystemHandle', 'file handles', '', 'nocode')}} for dropped items.
 
+> [!NOTE]
+> Because `getAsFileSystemHandle()` can only retrieve the entry handle the same tick as the `drop` event handler, there must be no `await` before it. This is why we synchronously invoke `getAsFileSystemHandle()` for all items first, and then wait for their results concurrently.
```
sideshowbarker:
```suggestion
> Because `getAsFileSystemHandle()` can only retrieve the entry handle in the same tick as the `drop` event handler, there must be no `await` before it. This is why we synchronously invoke `getAsFileSystemHandle()` for all items first, and then wait for their results concurrently.
```

sideshowbarker (approved):

PR #25283, opened 2023-03-13
Service workers cannot do dynamic import

hamishwillee:
The HTML spec explicitly prohibits use of `import()` in service workers and worklets here: https://whatpr.org/html/6395/webappapis.html#hostimportmoduledynamically(referencingscriptormodule,-specifier,-promisecapability)

This adds a note in a few key places - the `import()` statement being the most important, but also the javascript module guide (as a note) and the shared worker docs. 

This came out of discussion in https://github.com/mdn/browser-compat-data/pull/19054

It is associated with work for FF111 in https://github.com/mdn/content/issues/24402

FYI @Josh-Cena

PR #40574, opened 2025-07-30
Remove URL hash from import()

SadCar5295:
### Description

Remove unnecessary URL hash from the `import()`-link

### Motivation

The linked section is unnecessary for the reader, who expects to read about the `import()` function, not specifically what the browser support is.

### Additional details

Link before:
[`import()`](/en-US/docs/Web/JavaScript/Reference/Operators/import#browser_compatibility)
Link after:
[`import()`](/en-US/docs/Web/JavaScript/Reference/Operators/import)

### DIFF
--- a/files/en-us/web/api/service_worker_api/index.md
+++ b/files/en-us/web/api/service_worker_api/index.md
@@ -15,7 +15,7 @@ A service worker is an event-driven [worker](/en-US/docs/Web/API/Worker) registe
 
 Service workers run in a worker context: they therefore have no DOM access and run on a different thread to the main JavaScript that powers your app. They are non-blocking and designed to be fully asynchronous. As a consequence, APIs such as synchronous [XHR](/en-US/docs/Web/API/XMLHttpRequest) and [Web Storage](/en-US/docs/Web/API/Web_Storage_API) can't be used inside a service worker.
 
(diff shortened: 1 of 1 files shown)


Josh-Cena:
@hamishwillee I assume you added this anchor because at that time the compat table said "Available in workers - no". Do we have support in service workers now? I'm not sure if worker support can imply service worker support; if not then we should add another compat row, unless that's a fact given in the service worker spec or some vendor-agreed behavior.

hamishwillee:
@Josh-Cena dynamic `import()` is not supported in service workers or worklets (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#description) which I added in #25283

> Dynamic module import is not permitted in all execution contexts. For example, import() can be used in the main thread, a shared worker, or a dedicated worker, but will throw if called within a [service worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) or a [worklet](https://developer.mozilla.org/en-US/docs/Web/API/Worklet).

This is in the HTML spec here: https://whatpr.org/html/6395/webappapis.html#hostimportmoduledynamically(referencingscriptormodule,-specifier,-promisecapability)

I think this fix is correct and I should not have added the hash (ever).

I would love it if the BCD showed that it wasn't supported in shared workers/worklets, but at least at the time they didn't have a mechanism to show this case of "supported in the workers except this kind of worker"

hamishwillee (approved):
Thanks very much, and welcome to MDN.

Issue #38481, opened 2025-03-06
Runtime Error in The Code Example of The "Compiling from Rust to WebAssembly" Article


shawnrivers:
### MDN URL

https://developer.mozilla.org/en-US/docs/WebAssembly/Guides/Rust_to_Wasm

### What specific section or headline is this issue about?

Using the npm package on the web

### What information was incorrect, unhelpful, or incomplete?

When running the `npm run serve`, I don't see the alert dialog showing `"Hello, WebAssembly with npm!"`, instead I see the following runtime error in my browser console:

```
hello_wasm.js:86 Uncaught TypeError: Cannot read properties of undefined (reading '__wbindgen_malloc')
    at Module.greet (hello_wasm.js:86:47)
    at eval (index.js:5:41)
    at ./index.js (index.js:41:1)
    at __webpack_require__ (index.js:217:32)
    at index.js:1283:37
    at index.js:1285:12
```

### What did you expect to see?

I should see the alert dialog showing `"Hello, WebAssembly with npm!"` in the browser.

### Do you have anything more you want to share?

It seems like the WASM module needs to be initialized at first. Only then you can start calling any of its function.

After I modified the code of `site/index.js`, I was able to see the alert dialog and the runtime error was gone:

#### Before
```js
import * as wasm from "hello-wasm";

wasm.greet("WebAssembly with npm");
```

#### After
```js
import init, { greet } from 'hello-wasm';

init().then(() => {
  greet('WebAssembly with npm');
});
```

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/webassembly/guides/rust_to_wasm`
* MDN URL: https://developer.mozilla.org/en-US/docs/WebAssembly/Guides/Rust_to_Wasm
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/webassembly/guides/rust_to_wasm/index.md
* Last commit: https://github.com/mdn/content/commit/df9d06402163f77fc3e2d327ab63f9dd4af15b38
* Document last modified: 2025-01-31T14:58:40.000Z

</details>

Josh-Cena:
I cannot reproduce this error, and your hello-wasm should not contain an init function. It should look like this:

```js
import * as wasm from "./hello_wasm_bg.wasm";
export * from "./hello_wasm_bg.js";
import { __wbg_set_wasm } from "./hello_wasm_bg.js";
__wbg_set_wasm(wasm);
wasm.__wbindgen_start();
```

[...]

PR #40583, opened 2025-07-31
rust_to_wasm : fix site/index.js codeblock

pourquoi:
### Description

In the final index.js (site/index.js) the wasm module is not initialized leading to an undefined error when calling greet.


Josh-Cena:
Please see https://github.com/mdn/content/issues/38481

PR #40584, opened 2025-07-31
Correct typo in brightness example

mikoMK:
One of the examples uses `brightens()` instead of `brightness()`


estelle (approved):
Thank you!

PR #40588, opened 2025-08-01
Add scrollIntoView() container option

aguingand:
### Description

Documentation of the new `Element.scrollintoView()` `container` option.
 
### Motivation

The `container` option has been specified & scheduled to ship in Chrome 140.

### Additional details

https://drafts.csswg.org/cssom-view/#dom-scrollintoviewoptions-container
https://cr-status.appspot.com/feature/5100036528275456?gate=5140738792488960

### Related issues
https://github.com/mdn/browser-compat-data/pull/27463

### DIFF
--- a/files/en-us/web/api/element/scrollintoview/index.md
+++ b/files/en-us/web/api/element/scrollintoview/index.md
@@ -37,36 +37,36 @@ scrollIntoView(options)
   - : An object with the following properties:
     - `behavior` {{optional_inline}}
       - : Determines whether scrolling is instant or animates smoothly. Its value can be one of the following:
-        - `"smooth"`: scrolling should animate smoothly
-        - `"instant"`: scrolling should happen instantly in a single jump
-        - `"auto"`: scroll behavior is determined by the computed value of {{cssxref("scroll-behavior")}}
+        - `smooth`: scrolling should animate smoothly
+        - `instant`: scrolling should happen instantly in a single jump
+        - `auto`: scroll behavior is determined by the computed value of {{cssxref("scroll-behavior")}}
 
-        The default is `"auto"`.
+        The default is `auto`.
 
     - `block` {{optional_inline}}
       - : Defines the vertical alignment of the element within the scrollable ancestor container. Its value can be one of the following:
-        - `"start"`: Aligns the element's top edge with the top of the scrollable container, making the element appear at the start of the visible area vertically.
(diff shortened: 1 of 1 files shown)


aguingand:
I don't know if there should be more details to the option description

sideshowbarker (commented):
Howdy. The documentation you added for the `container` option is great, but can you say why you added quotes around all the option values? And why you didn’t also add any quotes around the option _names_?

Are you maybe following the example of a how some similar content on another existing page is formatted?

I’d personally prefer that we not put quotes around any of them. And I’m not sure if the MDN style guide provides any guidance on this. But at least I don’t think we have any linter that actually enforces that style.

But maybe this style (putting quotes around the option values but not the option names) is how most MDN pages are styled.

If so, I guess maybe the logic for that would be, the option values are strings but the option names aren’t? Regardless, from just that, I don’t personally think it follows that it should be necessary or useful to quote the option values — especially not in a case like this page, where all the options are limited to an enumerated set of values, and can’t be any arbitrary string.

aguingand:
I don't think there should be quotes either, @Josh-Cena I will remove them (but keep the reordering you made)

Josh-Cena:
We indeed don't have written guidelines, but I'm saying "yes" in https://github.com/orgs/mdn/discussions/779. You are welcome to give opinions for "no". `smooth` is an identifier and corresponds to `{ behavior: smooth }`; `"smooth"` is a string and corresponds to `{ behavior: "smooth" }`. Otherwise, `false`, `1`, or `NodeFilter.NONE` would be ambiguous.

aguingand:
I guess it should be a separate PR that updates all string values, for now other pages seems to be unquoted (e.g. https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo, https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#open...)

Josh-Cena:
Alright, okay to keep it in this form

sideshowbarker (approved):

sideshowbarker (approved):

sideshowbarker:
@aguingand, 👍 and congrats on landing your first docs change here — welcome aboard 🎉

aguingand:
@sideshowbarker Thank you ! Long live MDN !

aguingand:
@sideshowbarker I think BCD PR should be merged too (https://github.com/mdn/browser-compat-data/pull/27463)

PR #40591, opened 2025-08-01
Update external XSLT links and move them to the guide

mathewhodson:
The "Hands-on XSL" article is no longer hosted at developer.ibm.com


Josh-Cena (approved):
Thanks! Great call

PR #40597, opened 2025-08-02
PasswordCredential cannot be passed an object from fetch()

wbamberg:
The bit in https://developer.mozilla.org/en-US/docs/Web/API/PasswordCredential/PasswordCredential:

> In supporting browsers, an instance of this class may be passed the credential from the init object for global [fetch()](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch).

...as far as we can tell makes no sense.


hamishwillee (approved):
Thanks very much.

PR #40600, opened 2025-08-02
Update README.md - MAJOR CLARIFICATION ADDED

gitboy2022:
### Description

Add a line to clarify to minimize errors when running the site locally.
(Running this site locally inside a directory with spaces makes the site not run properly.)

### Motivation

I made these changes because I faced that type of problem before.

### Additional details

none

```
resolves issue #40595
```


Josh-Cena:
This should be reported as a bug to https://github.com/mdn/rari instead.

PR #40228, opened 2025-07-07
Expand CSS style guide to cover more conventions and recommendations

Josh-Cena:
Following my Discord discussion, I am renovating the CSS style guide to add more normative requirements, so I can go ahead and implement linting. I am expecting debate over some of these (mostly arbitrary) decisions, but where appropriate I have added some reasoning. Please do argue if you believe otherwise.

I am looping in everyone who may be interested in CSS code style. You can cc others. (cc @OnkarRuikar here)

PR #40603, opened 2025-08-03
Lint: identifiers should use kebab-case

Josh-Cena:
After https://github.com/mdn/content/pull/40228, we are a lot more explicit on some CSS stylistic decisions. This PR makes sure that all CSS identifiers use `kebab-case`, while SVG value keywords use camelCase (mostly `currentColor`).


dipikabh (approved):
Thanks a lot!

PR #40604, opened 2025-08-03
Fix hyperlink

thermosflasche:
### Description

Fixed wrong CSS' `%` (percentage) unit hyperlink. I think the reason is self-explanatory.

### Motivation

It helps with docs navigation.


Josh-Cena (approved):
Great catch, thanks

PR #40605, opened 2025-08-03
Fix typo `Scx`->`scx` for short name of `Script_Extensions` prop

lionel-rowe:
### Description

Fix typo `Scx`->`scx` for short name of `Script_Extensions` prop.

### Motivation

`Scx` capitalization is invalid, must be `scx`:

```js
/\p{scx=Thaana}/u // ok
/\p{Scx=Thaana}/u // Uncaught SyntaxError: Invalid regular expression: /\p{Scx=Thaana}/u: Invalid property name
```


Josh-Cena (approved):
Great catch. Thanks

PR #40607, opened 2025-08-03
Feat/property example

yashrajbharti:
Added animating gradient example which is now possible through CSS `@property`

### DIFF
--- a/files/en-us/web/css/@property/index.md
+++ b/files/en-us/web/css/@property/index.md
@@ -128,7 +128,7 @@ For item two, CSS global keywords are set for both custom properties which are v
 
 For item three, the `--item-size` value gets set to `1000px`. While `1000px` is a {{cssxref("length")}} value, the `@property` declaration requires the value be a `<percentage>`, so the declaration is not valid and is ignored, meaning the inheritable `20%` set on the parent is used. The `xyz` value is also invalid. As `registerProperty()` set `--item-color` to not be inherited, the default initial value of `aqua` is used and not the parent's `orange` value.
 
-### Using `@property` to animate linear gradient
+### Animating a custom property value
 
 In this example we define a custom property `--progress` that animates the position of a progress bar within a linear gradient.
 
--- a/files/en-us/web/css/@property/index.md
+++ b/files/en-us/web/css/@property/index.md
@@ -130,7 +130,7 @@ For item three, the `--item-size` value gets set to `1000px`. While `1000px` is 
 
 ### Animating a custom property value
 
-In this example, we define a custom property called `--progress` using `@property`: this accepts [`<percentage>`](/en-US/docs/Web/CSS/percentage) values and has an initial value of `25%`. We use `--progress` to define the position value of the color stops in a {{cssxref("<linear-gradient>")}}, specifying where a green color stops, and black starts. We then animate the value of `--progress` to `100%` over 2.5 seconds, giving the effect of animating a progress bar.
+In this example, we define a custom property called `--progress` using `@property`: this accepts [`<percentage>`](/en-US/docs/Web/CSS/percentage) values and has an initial value of `25%`. We use `--progress` to define the position value of the color stops in a {{cssxref("linear-gradient()")}}, specifying where a green color stops, and black starts. We then animate the value of `--progress` to `100%` over 2.5 seconds, giving the effect of animating a progress bar.
 
 ```html
 <div class="bar"></div>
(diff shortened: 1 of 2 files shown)


files/en-us/web/css/@property/index.md
```diff
@@ -128,6 +128,43 @@ For item two, CSS global keywords are set for both custom properties which are v
 
 For item three, the `--item-size` value gets set to `1000px`. While `1000px` is a {{cssxref("length")}} value, the `@property` declaration requires the value be a `<percentage>`, so the declaration is not valid and is ignored, meaning the inheritable `20%` set on the parent is used. The `xyz` value is also invalid. As `registerProperty()` set `--item-color` to not be inherited, the default initial value of `aqua` is used and not the parent's `orange` value.
 
+### Using `@property` to animate linear gradient
```
chrisdavidmills: Maybe something like:

```suggestion
### Animating a custom property value
```

?

It isn't accurate to say that `@property` is being used to animate the linear gradient. It just defines the property that we then animate.

files/en-us/web/css/@property/index.md
```diff
@@ -128,6 +128,43 @@ For item two, CSS global keywords are set for both custom properties which are v
...
+
+In this example we define a custom property `--progress` that animates the position of a progress bar within a linear gradient.
```
chrisdavidmills:
```suggestion
In this example, we define a custom property called `--progress` using `@property`: this accepts `<percentage>` values and has an initial value of `25%`. We use `--progress` to define the position value of the color stops in a {{cssxref("<linear-gradient>")}}, specifying where a green color stops, and black starts. We then animate the value of `--progress` to `100%` over 2.5 seconds, giving the effect of animating a progress bar.
```

chrisdavidmills: Or something similar...I just thought a bit more explanation was needed.

files/en-us/web/css/@property/index.md
```diff
@@ -128,6 +128,43 @@ For item two, CSS global keywords are set for both custom properties which are v
...
+}
+
+.bar {
+  display: inline-block;
+  --progress: 25%;
+  width: 100%;
+  height: 5px;
+  background: linear-gradient(
+    to right,
+    #00d230 0 var(--progress),
```
chrisdavidmills:
```suggestion
    #00d230 var(--progress),
```

I don't think the solitary `0` is needed, from testing?

files/en-us/web/css/@property/index.md
```diff
@@ -128,6 +128,43 @@ For item two, CSS global keywords are set for both custom properties which are v
...
+}
+
+@keyframes progressAnimation {
+  to {
+    --progress: 100%;
+  }
+}
+```
+
+{{ EmbedLiveSample('Using `@property` to animate linear gradient', '100%', '250px') }}
```
chrisdavidmills: Change the heading here, too, if you agree with my heading change suggestion above.

Also, you don't need to set the container to `250px` high. `60px` would do, which is the minimum allowed.

files/en-us/web/css/@property/index.md
```diff
@@ -128,6 +128,43 @@ For item two, CSS global keywords are set for both custom properties which are v
...
+  display: inline-block;
+  --progress: 25%;
+  width: 100%;
+  height: 5px;
+  background: linear-gradient(
+    to right,
+    #00d230 0 var(--progress),
+    #000000 var(--progress)
+  );
+  animation: progressAnimation 2.5s ease infinite;
```
chrisdavidmills, line 156: `linear` easing might be better/slightly more typical for a progress bar? Hrm, maybe not; I guess not all progress is linear. I'll leave this up to you.

chrisdavidmills (changes requested):
Hi there @yashrajbharti, and thank you for your example! I've left you a few comments to consider.

files/en-us/web/css/@property/index.md
yashrajbharti: True, it works without it!

files/en-us/web/css/@property/index.md
yashrajbharti: To me these look more "realistic" as often the loading slows at the end.
Reminds me of the loading meme gif

chrisdavidmills: Yes, totally fair! ;-)

chrisdavidmills (approved):
@yashrajbharti This is looking great, thank you for the fabulous addition to MDN!

Let's get this merged.

PR #40608, opened 2025-08-03
Lint: spaces around CSS comments, from/to keyframe selectors

Josh-Cena:
After https://github.com/mdn/content/pull/40228, we are a lot more explicit on some CSS stylistic decisions. This PR fixes two basic things: comments should start and end with a space, and the use of `from` and `to` selectors should follow our style guide.


dipikabh (approved):
Thanks!

PR #40495, opened 2025-07-22
Ff142 RTCInboundRtpStreamStats fixes

hamishwillee:
This fixes up the [`RTCInboundRtpStreamStats`](https://developer.mozilla.org/en-US/docs/Web/API/RTCInboundRtpStreamStats) to match the BCD as per FF142, which adds support for some more properties.

RTC is hard IMO, and the specs are not easy. A lot of this is nothing more than the spec. A little of it is. For example, I did learn some interesting things about the jitter buffer and associated stats in https://webrtchacks.com/how-webrtcs-neteq-jitter-buffer-provides-smooth-audio/

Contributes to https://github.com/mdn/mdn/issues/384

Related docs work can be tracked in #40478

Removals (not in BCD): `averageRtcpInterval`, `firCount`, `lastPacketReceivedTimestamp`,  `packetsDuplicated`,  `packetsFailedDecryption`,  `perDscpPacketsReceived`,  `pliCount`,  `receiverId`,  `sliCount`,

PR #40613, opened 2025-08-04
FF142 RTCInboundRtpStreamStats stats - a few more supported

hamishwillee:
FF142 adds support for some  [`RTCInboundRtpStreamStats`](https://developer.mozilla.org/en-US/docs/Web/API/RTCInboundRtpStreamStats). This adds the release note.

The docs for the properties listed in this are added in #40495

Related docs work can be tracked in #40478


dipikabh (approved):
Cool, thanks!

PR #40623, opened 2025-08-04
Lint: use named color if possible (5)

Josh-Cena:
After https://github.com/mdn/content/pull/40228, we are a lot more explicit on some CSS stylistic decisions. In particular, we now have systematic rules surrounding the preferred color notations. As a starter, I'm porting all colors that have named equivalents to named colors, per our recommendation of using "common named colors". Afterwards, I'm going to convert the remaining colors to the preferred notation, such as preferring `rgb()` and preferring number parameters.

### DIFF
--- a/files/en-us/web/css/color_value/color-mix/index.md
+++ b/files/en-us/web/css/color_value/color-mix/index.md
@@ -24,9 +24,9 @@ color-mix(in hsl, hsl(200 50 80), coral 80%)
 color-mix(in lch longer hue, hsl(200deg 50% 80%), coral)
 
 /* color-mix(in <rectangular-color-space>, <color>, <color>) */
-color-mix(in srgb, plum, #f00)
+color-mix(in srgb, plum, #123456)
 /* color-mix(in <rectangular-color-space>, <color> <percentage>, <color> <percentage> */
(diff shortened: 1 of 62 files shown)


files/en-us/web/css/color_value/color-mix/index.md
```diff
@@ -24,9 +24,9 @@ color-mix(in hsl, hsl(200 50 80), coral 80%)
 color-mix(in lch longer hue, hsl(200deg 50% 80%), coral)
 
 /* color-mix(in <rectangular-color-space>, <color>, <color>) */
-color-mix(in srgb, plum, #f00)
+color-mix(in srgb, plum, #123456)
```
dipikabh, line 27: why not use a known hex?

dipikabh (approved):
LGTM. Thanks for the updates!!
Left one question out of curiosity

files/en-us/web/css/color_value/color-mix/index.md
Josh-Cena: By "a known hex" I assume you mean why not just `#f00`? Technically it won't be a problem for me, but I was doing a global search and replace for `#f00`, so I need to avoid these false positives

PR #40628, opened 2025-08-04
fix: replace "fragment" with "query" in URI query ref

tales-aparecida:
### Description

This PR fixes a typo in the argument name used in URI/reference/query.
The current line seems to be copypasta from the "fragment" reference page.

### Motivation

This caught my eyes while reading the page, and I've interpreted as a simple typo that could cause some confusion.

### Additional details

I couldn't find any open issue about this.


Josh-Cena (approved):
Good catch, thanks!

PR #40629, opened 2025-08-05
RTCRtpSender.get/setParameters - encodings.codec property

hamishwillee:
FF142 adds support for the dictionary property `RTCRtpEncodingParameters.codec` in https://bugzilla.mozilla.org/show_bug.cgi?id=1894137

This is used when setting encodings in [`RTCRtpSender.setParameters(parameter.encodings)`](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/setParameters#encodings) and [`RTCPeerConnection.addTransceiver(sendencodings)`](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addTransceiver#sendencodings), and returned in [`RTCRtpSender/getParameters()`](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/getParameters#encodings) (`encodings.codec`).

This updates the set/getParameters on the sender. The transceiver just references the sender, so doesn't need an update.

Related docs work can be tracked in #40477

### DIFF
--- a/files/en-us/web/api/rtcrtpsender/getparameters/index.md
+++ b/files/en-us/web/api/rtcrtpsender/getparameters/index.md
@@ -32,6 +32,29 @@ An object indicating the current configuration of the sender. <!-- RTCRtpSendPar
     - `active`
       - : `true` (the default) if the encoding is being sent, `false` if it is not being sent or used.
 
+    - `codec` {{optional_inline}}
+      - : Selects the [media codec](/en-US/docs/Web/Media/Guides/Formats/WebRTC_codecs) that is used for this encoding's RTP stream.
+        If not set, the user agent may select any codec negotiated for sending.
+        <!-- RTCRtpCodec -->
+        - `channels` {{optional_inline}}
+          - : A positive integer indicating the number of channels supported by the codec.
+            For example, for audio codecs a value of 1 specifies monaural sound, while 2 indicates stereo.
+
+        - `clockRate`
+          - : A positive integer specifying the codec's clock rate in Hertz (Hz).
+            The clock rate is the rate at which the codec's RTP timestamp advances.
+            Most codecs have specific values or ranges of values they permit.
+            The IANA maintains a [list of codecs and their parameters](https://www.iana.org/assignments/rtp-parameters/rtp-parameters.xhtml#rtp-parameters-1), including their clock rates.
+
+        - `mimeType`
+          - : A string indicating the codec's MIME media type and subtype, specified as a string of the form `"type/subtype"`.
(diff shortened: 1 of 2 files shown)


files/en-us/web/api/rtcrtpsender/getparameters/index.md
```diff
@@ -32,6 +32,29 @@ An object indicating the current configuration of the sender. <!-- RTCRtpSendPar
     - `active`
       - : `true` (the default) if the encoding is being sent, `false` if it is not being sent or used.
 
+    - `codec` {{optional_inline}}
```
hamishwillee, line 35: FYI This is the same type of object as used in the following `codecs` array except that has a payload type as well. Decided not to update that one to say "this is the same as codec + payload type".

files/en-us/web/api/rtcrtpsender/getparameters/index.md
```diff
@@ -96,8 +119,15 @@ An object indicating the current configuration of the sender. <!-- RTCRtpSendPar
   - : An array of zero or more RTP header extensions, each identifying an extension supported by the sender or receiver. Header extensions are described in {{RFC(3550, "", "5.3.1")}}.
     This parameter cannot be changed once initially set.
 - `rtcp`
-  - : An {{domxref("RTCRtcpParameters")}} object providing the configuration parameters used for {{Glossary("RTCP")}} on the sender or receiver.
-    This parameter cannot be changed once initially set.
+  - : An object providing the configuration parameters used for {{Glossary("RTCP")}} on the sender.
```
hamishwillee, line 122: Decided to fix up missing object while here.

files/en-us/web/api/rtcrtpsender/setparameters/index.md
```diff
@@ -28,7 +28,30 @@ setParameters(parameters)
       - : An array of objects, each specifying the parameters for a single codec that could be used to encode the track's media.
         The properties of the objects include:
         - `active`
-          - : Setting this value `true` (the default) causes this encoding to be sent, while `false` stops it from being sent and used (but does not cause the SSRC to be removed).
+          - : Setting this value `true` (the default) causes this encoding to be sent, while `false` stops it from being sent and used (but does not cause the SSRC to be removed). RTCRtpCodec
```
wbamberg:
```suggestion
          - : Setting this value `true` (the default) causes this encoding to be sent, while `false` stops it from being sent and used (but does not cause the SSRC to be removed).
```

wbamberg (approved):
👍

PR #40630, opened 2025-08-05
fix the code comment with incorrect location

nano71:
### Description

The code comments were placed in the wrong position.


Josh-Cena (approved):
Sure, I think this makes sense. Thanks

PR #40633, opened 2025-08-05
Element.outerHTML and TrustedTypes

hamishwillee:
This updates the  [`Element.outerHTML` ](https://developer.mozilla.org/en-US/docs/Web/API/Element/outerHTML) to show how it is used with TrustedTypes. It is based on the work done in #40423 for `innerHTML`.

- Added boilerplate
- Restructured to match current template, such as copying up things that were notes previously.
- Copied in security considerations and shadow dom considerations from the innerHTML and updated appropriately. 
- Add TT information, including the new exception and tinyfill.
- Mirrored the innerHTML example behavior

### DIFF
--- a/files/en-us/web/api/document/write/index.md
+++ b/files/en-us/web/api/document/write/index.md
@@ -24,11 +24,11 @@ browser-compat: api.Document.write
 > > For all these reasons, use of this method is strongly discouraged.
 
 > [!WARNING]
-> This API parses its input as HTML, writing the result into the DOM.
+> This method parses its input as HTML, writing the result into the DOM.
 > APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
 >
-> For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
-> This means you can be sure that the input has been passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
+> You can mitigate this risk by always assigning `TrustedHTML` objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](#security_considerations) for more information.
 
 The **`write()`** method of the {{domxref("Document")}} interface writes text in one or more {{domxref("TrustedHTML")}} or string parameters to a document stream opened by {{domxref("document.open()")}}.
 
@@ -60,10 +60,6 @@ None ({{jsxref("undefined")}}).
 
 `document.write()` parses the markup text in the objects passed as parameters into the open document's object model (DOM), in the order that the parameters are specified.
 
-The passed objects may be {{domxref("TrustedHTML")}} instances or strings.
-It is much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
-The guarantees that the input has been passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
-
 Because `document.write()` writes to the document **stream**, calling `document.write()` on a closed (loaded) document (without first calling {{domxref("document.open()")}}) automatically calls {{domxref("document.open()")}}, which will clear the document.
 
 The exception is that if the `document.write()` call is embedded within an inline HTML `<script>` tag, then it will not automatically call `document.open()`:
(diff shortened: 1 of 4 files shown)


files/en-us/web/api/element/outerhtml/index.md
```diff
@@ -8,108 +8,160 @@ browser-compat: api.Element.outerHTML
...
+You can get and log the markup for the {{htmlelement("div")}} as shown:
+
+```js
+const myElement = document.querySelector("#example");
+const contents = myElement.outerHTML;
+console.log(contents);
+// '<div id="example">\n  <p>Content</p>\n  <p>Further Elaborated</p>\n</div>'
 ```
 
-## Notes
```
hamishwillee, line 83: FYI Notes moved up into description.

files/en-us/web/api/element/outerhtml/index.md
```diff
@@ -8,108 +8,160 @@ browser-compat: api.Element.outerHTML
...
+p.outerHTML = "<div>This div replaced a paragraph.</div>";
+console.log(p.nodeName); // still "P";
 ```
 
-### Replacing a node by setting the outerHTML property
+### Escaped attribute values
 
-#### HTML
+The returned value will escape some values in HTML attributes.
+Here we see that the `&` character is escaped:
```
hamishwillee, line 64: This bit comes from the notes, which said HTML attribute values are escaped, and demonstrates with the &.
However that can't be entirely correct, because I know that this bit below on "Some browsers also serialize" is very recent. So what I have done is make this "will escape some values" - its a bit weasely, but it means that users won't be surprised that not all are escaped all the time.

wbamberg (approved):
I think this is great. I do wonder (and sorry, this is surely my fault) if we have too much repetition between the boilerplate warning and "security considerations". Should we consider keeping only the first paragraph of the warning:

> Warning: This property parses its input as HTML, writing the result into the DOM. APIs like this are known as [injection sinks](https://pr40633.review.mdn.allizom.net/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](https://pr40633.review.mdn.allizom.net/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.

...and then adding another paragraph like:

> See [Security considerations]() to learn how to mitigate this issue by always passing `TrustedHTML` objects into this API, instead of strings.

?

hamishwillee:
> I think this is great. I do wonder (and sorry, this is surely my fault)

@wbamberg It is your fault, because that original header is such a succinct summary of the problem and solution. It's hard to replace something you like - even though I did feel the same way on this.

Anyway, I've committed a slightly modified version for your review (and put the current version below for comparison). If you like this, don't merge - I'll roll it out to the other already merged things that use this header.

image

wbamberg:
The warning in https://pr40633.review.mdn.allizom.net/en-US/docs/Web/API/Element/outerHTML looks the same as your screenshot, what am I missing?

hamishwillee:
> The warning in https://pr40633.review.mdn.allizom.net/en-US/docs/Web/API/Element/outerHTML looks the same as your screenshot, what am I missing?

Annoyingly, nothing. The commit is now pushed.  The only change is the new header. The content in security section is still OK. I think this is slightly better.

image

wbamberg:
> Anyway, I've committed a slightly modified version for your review (and put the current version below for comparison). If you like this, don't merge - I'll roll it out to the other already merged things that use this header.

Yes, I do like this version, thank you! Per your request here, not merging :).

files/en-us/web/api/document/write/index.md
```diff
@@ -24,11 +24,11 @@ browser-compat: api.Document.write
...
+> This method parses its input as HTML, writing the result into the DOM.
 > APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
 >
-> For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
-> This means you can be sure that the input has been passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
+> You can mitigate this risk by always assigning `TrustedHTML` objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
```
hamishwillee:
```suggestion
> You can mitigate this risk by always passing `TrustedHTML` objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
```

files/en-us/web/api/document/writeln/index.md
```diff
@@ -11,18 +11,14 @@ browser-compat: api.Document.writeln
...
+> This method parses its input as HTML, writing the result into the DOM.
 > APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, if the input originally came from an attacker.
 >
-> For this reason it's much safer to pass only {{domxref("TrustedHTML")}} objects into this method, and to [enforce](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) this using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
-> This means you can be sure that the input has been passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input to remove potentially dangerous markup, such as {{htmlelement("script")}} elements and event handler attributes.
+> You can mitigate this risk by always assigning `TrustedHTML` objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
```
hamishwillee:
```suggestion
> You can mitigate this risk by always passing `TrustedHTML` objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
```

hamishwillee:
Thanks @wbamberg - I've rolled this out to the other docs we already saw. Not forcing a re-review - changes fairly minimal and constrained.

PR #40641, opened 2025-08-05
caret-color: update 

estelle:
caret-color is now supported everywhere.
added an explanation of what carets are and when they are used.


files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,35 @@ caret-color: unset;
 ### Values
 
 - `auto`
-  - : The user agent selects an appropriate color for the caret. This is generally {{cssxref("&lt;color&gt;","currentColor","#currentcolor_keyword")}}, but the user agent may choose a different color to ensure good visibility and contrast with the surrounding content, taking into account the value of `currentColor`, the background, shadows, and other factors.
-
-    > [!NOTE]
-    > While user agents may use `currentColor` (which is usually animatable) for the `auto` value, `auto` is not interpolated in transitions and animations.
+  - : Generally resolves to [`currentcolor`](/en-US/docs/Web/CSS/color_value#currentcolor_keyword), the {{cssxref("color")}} of the text that will but modified.
```
Josh-Cena: Drive-by. Same in other places; we require saying `currentColor` instead of `currentcolor` from now on

```suggestion
  - : Generally resolves to [`currentColor`](/en-US/docs/Web/CSS/color_value#currentcolor_keyword), the {{cssxref("color")}} of the text that will but modified.
```

files/en-us/web/css/caret-color/index.md
```diff
@@ -6,7 +6,7 @@ browser-compat: css.properties.caret-color
 sidebar: cssref
 ---
 
-The **`caret-color`** [CSS](/en-US/docs/Web/CSS) property sets the color of the **insertion caret**, the visible marker where the next character typed will be inserted. This is sometimes referred to as the **text input cursor**. The caret appears in elements such as {{HTMLElement("input")}} or those with the [`contenteditable`](/en-US/docs/Web/HTML/Reference/Global_attributes/contenteditable) attribute. The caret is typically a thin vertical line that flashes to help make it more noticeable. By default, it is black, but its color can be altered with this property.
+The **`caret-color`** [CSS](/en-US/docs/Web/CSS) property sets the color of the **insertion caret**, sometimes referred to as the **text input cursor**, which is the visible marker appearing at the insertion point where the next character typed will be added or where the next character deleted will be removed.
```
chrisdavidmills:
```suggestion
The **`caret-color`** [CSS](/en-US/docs/Web/CSS) property sets the color of the **insertion caret**, sometimes referred to as the **text input cursor**. This is the visible marker appearing at the insertion point where the next character typed will be added or where the next character deleted will be removed.
```

files/en-us/web/css/caret-color/index.md
chrisdavidmills: Agree with @Josh-Cena; you also have a typo to fix here: but -> be

```suggestion
  - : Generally resolves to [`currentColor`](/en-US/docs/Web/CSS/color_value#currentcolor_keyword), the {{cssxref("color")}} of the text that will be modified.
```

files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,35 @@ caret-color: unset;
...
-    > [!NOTE]
-    > While user agents may use `currentColor` (which is usually animatable) for the `auto` value, `auto` is not interpolated in transitions and animations.
+  - : Generally resolves to [`currentcolor`](/en-US/docs/Web/CSS/color_value#currentcolor_keyword), the {{cssxref("color")}} of the text that will but modified.
 
 - {{cssxref("&lt;color&gt;")}}
   - : The color of the caret.
 
+## Description
+
+An insertion caret is a visible indicator of the spot within editable text or within an element that accepts text input where content will be inserted (or deleted) by the user. The caret is typically a thin vertical line indicating where the addition or deletion of characters will occur. It generally blinks, or flashes on and off, to help make itself more noticeable. The caret only appears when the potentially editable element has focus. By default, this caret is the color of the text. The `caret-color` property is used to set the color of this caret to something other than the `currentcolor`, or, alternatively, to reset a colored caret back to its default.
```
chrisdavidmills:
```suggestion
An insertion caret is a visible indicator of the spot within editable text — or within an element that accepts text input — where content will be inserted (or deleted) by the user. The caret is typically a thin vertical line indicating where the addition or deletion of characters will occur. It generally blinks (flashes on and off), so it is more noticeable. The caret only appears when the editable element has focus. By default, this caret is the color of the text. The `caret-color` property can be used to set the color of this caret to something other than the `currentColor`, or to reset a colored caret back to its default.
```

files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,35 @@ caret-color: unset;
...
+
+Setting `auto` sets the insertion caret to `currentcolor`, which is the {{cssxref("color")}} of the text that is being added or deleted. While user agents may choose a different color to ensure good visibility and contrast with the surrounding content, taking into account the {{cssxref("color")}}, the {{cssxref("background-color")}}, shadows, and other factors, in practice, all browsers use the current color when `caret-color` is set or defaults to `auto`. You can set any valid `<color>` as the value.
```
chrisdavidmills:
```suggestion
The `auto` value sets the insertion caret to `currentColor`, which is the {{cssxref("color")}} of the text that is being added or deleted. User agents may choose a different color to ensure good visibility and contrast with the surrounding content, taking into account the {{cssxref("color")}}, {{cssxref("background-color")}}, shadows, and other factors. In practice, however, all browsers use the current color by default, and when `caret-color` is set to `auto`. You can set any valid `<color>` as the value.
```

files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,35 @@ caret-color: unset;
...
+
+### Understanding insertion carets
+
+The insertion caret, and therefore this property, only applies to text or elements that can accept text input. The caret appears in focused user interface elements where users can update content, such as {{HTMLElement("input")}} elements that accept freeform text, the {{HTMLElement("textarea")}} element, and elements with the [`contenteditable`](/en-US/docs/Web/HTML/Reference/Global_attributes/contenteditable) attribute set.
+
+The caret may appear in `<input>` elements of type `password`, `text`, `search`, `tel`, and `email`. No caret exists with date, `color`, `hidden`, `radio`, or `checkbox` input types. Some browsers display a caret with the `number` input type. If an element that never has text content is altered, such as by setting [`appearance: none`](/en-US/docs/Web/CSS/appearance) and adding the `contenteditable` attribute, it is possible in some browsers to make a caret appear, but don't do that!
```
chrisdavidmills:
```suggestion
The caret may appear in `<input>` elements of type `password`, `text`, `search`, `tel`, and `email`. No caret exists with date, `color`, `hidden`, `radio`, or `checkbox` input types. Some browsers display a caret with the `number` input type. It is possible in some browsers to make a caret appear in elements that never have text content — for example, by setting [`appearance: none`](/en-US/docs/Web/CSS/appearance) and adding the `contenteditable` attribute. However, this is not recommended.
```

files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,35 @@ caret-color: unset;
...
+
+The caret may be displayed if the element is either an editable element or a descendent of an editable element, as long as the editability is not disabled, such as by setting the descendent element's `contentEditable` attribute to `false`. If an element is not editable or selectable, such as if {{cssxref("user-select")}} is set to `none`, the caret should not appear.
```
chrisdavidmills:
```suggestion
A caret may be displayed in an editable element or its descendants, provided the editability is not disabled, for example by setting a descendant element's `contentEditable` attribute to `false`. If an element is not editable or selectable, for example if {{cssxref("user-select")}} is set to `none`, the caret should not appear.
```

files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,35 @@ caret-color: unset;
...
+
+### Caret versus cursor
+
+The insertion caret is only one type of caret, and the only type of caret effected by the `caret-color` property.
```
chrisdavidmills:
```suggestion
There are multiple types of caret. The insertion caret is the only type affected by the `caret-color` property.
```

files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,35 @@ caret-color: unset;
...
+
+Many browsers have a **navigation caret**, which acts similarly to an insertion caret but can be moved around in non-editable text. The mouse cursor image shown when hovering over text where the {{cssxref("cursor")}} property is `auto`, or when hovering over an element where the `cursor` property is `text` or `vertical-text`, may resemble a caret. But it's not. It's a cursor.
```
chrisdavidmills:
```suggestion
Many browsers have a **navigation caret**, which acts similarly to an insertion caret but can be moved around in non-editable text.

The mouse cursor image shown for certain {{cssxref("cursor")}} property values (for example, `auto` or `text`) may resemble a caret, but it's not. It's a cursor.
```

chrisdavidmills: The stuff about mouse cursors and exactly when they are shown seemed difficult to read and totally overkill for this page. You might want to make sure it is accurately represented on the cursor reference page, but for this page, I think something like the above suggestion is fine.

files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,35 @@ caret-color: unset;
...
+
+### Animating from `auto`
+
+Generally, when the `caret-color` is set to or defaults to `auto`, user agents use `currentcolor` which is animatable. However, `auto` is not by default an animatable value. Therefor, if animating or transitioning the `caret-color` from `auto` to any color value, no interpolation happens. Instead, after any delay expires, the color switches from or to the `currentcolor` color at the midpoint of the {{cssxref("animation-duration")}} or {{cssxref("transition-duration")}}.
```
chrisdavidmills:
```suggestion
Generally, when the `caret-color` is set to or defaults to `auto`, user agents use `currentColor`, which is animatable. However, `auto` is not an animatable value by default: when animating or transitioning the `caret-color` from `auto` to any color value, no interpolation happens. Instead, the color switches from or to the `currentColor` color at the midpoint of the {{cssxref("animation-duration")}} or {{cssxref("transition-duration")}}.
```

chrisdavidmills: Also, this is standard [discrete](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties#discrete) behavior anyway, isn't it? Maybe just say animations are discrete when animating to or from `auto`, and link to there for an explanation?

chrisdavidmills (changes requested):
@estelle Nice work; just a few suggestions for you to look over.

files/en-us/web/css/caret-color/index.md
estelle: oh. we had a discussion 5 years ago or so and went in the other direction. I remember, because i was the voice arguing for camel case, and lost. Right now we have `currentcolor` 101 times and `currentColor` 68 times. 

the spec authors I think were the ones who managed to convince me that non-camel case was the right way in the end, but even the spec uses camelcase. I wonder what changed?

files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,37 @@ caret-color: unset;
...
+
+The caret may appear in `<input>` elements of type `password`, `text`, `search`, `tel`, and `email`. No caret exists with date, `color`, `hidden`, `radio`, or `checkbox` input types. Some browsers display a caret with the `number` input type. It is possible in some browsers to make a caret appear in elements that never have text content — for example, by setting [`appearance: none`](/en-US/docs/Web/CSS/appearance) and adding the `contenteditable` attribute. However, this is not recommended.
+
+A caret may be displayed in an editable element or its descendants, provided the editability is not disabled, for example by setting a descendant element's `contentEditable` attribute to `false`. If an element is not editable or selectable, for example if {{cssxref("user-select")}} is set to `none`, the caret should not appear.
+
+### Caret versus cursor
+
+There are multiple types of caret. The insertion caret is the only type affected by the `caret-color` property.
```
estelle:
```suggestion
There are multiple types of carets. The insertion caret is the only type affected by the `caret-color` property.
```

files/en-us/web/css/caret-color/index.md
```diff
@@ -64,14 +62,37 @@ caret-color: unset;
...
+
+There are multiple types of carets. The insertion caret is the only type affected by the `caret-color` property.
+
+Many browsers have a **navigation caret**, which acts similarly to an insertion caret but can be moved around in non-editable text.
+
+The mouse cursor image shown for certain {{cssxref("cursor")}} property values (for example, `auto` or `text`) may resemble a caret, but it's not. It's a cursor.
+
+### Animating from `auto`
+
+Generally, when the `caret-color` is set to or defaults to `auto`, user agents use `currentColor`, which is animatable. However, `auto` is not an animatable value by default: when animating or transitioning the `caret-color` from `auto` to any color value, no interpolation happens. Instead, the color switches from or to the `currentColor` color at the midpoint of the {{cssxref("animation-duration")}} or {{cssxref("transition-duration")}}.
```
estelle:
```suggestion
Generally, when the `caret-color` is set to or defaults to `auto`, user agents use `currentColor`, which is animatable. However, `auto` is not an animatable value by default: when animating or transitioning the `caret-color` from `auto` to any color value, no interpolation happens. The animation is [discrete](/en-US/docs/Web/CSS/CSS_animated_properties#discrete); the color switches from or to the `currentColor` color at the midpoint of the {{cssxref("animation-duration")}} or {{cssxref("transition-duration")}}.
```

files/en-us/web/css/caret-color/index.md
chrisdavidmills: Hrm; I don't think it's that big a deal really.

chrisdavidmills (approved):
@estelle I'm happy with this! Approving.

files/en-us/web/css/caret-color/index.md
Josh-Cena: It changed in the new CSS style guide: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide/CSS#casing, and I already fixed all content in https://github.com/mdn/content/pull/40603. If there was a discussion before then it predates me and the outcome wasn't coded anywhere. FWIW I don't feel very strongly about `currentcolor`, but all lowercase looks very weird for all other SVG keywords, so I decided to use camelCase for this as well for consistency.

PR #40643, opened 2025-08-05
docs(fetchevent): improve example clarity

jmdarling:
### Description
Updates the "example" for [FetchEvent](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent#examples) to improve clarity and better match likely real-world implementations.

### Motivation
When leveraging the [FetchEvent](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent) documentation to help educate my co-workers, the example provided lead to confusion about how to leverage the API. The improvements made should more clearly explain what is being done in the example and why each line matters.

### Additional details

N/A

N/A

### DIFF
--- a/files/en-us/web/api/fetchevent/index.md
+++ b/files/en-us/web/api/fetchevent/index.md
@@ -63,14 +63,22 @@ self.addEventListener("fetch", (event) => {
       const cachedResponse = await cache.match(event.request);
 
       if (cachedResponse) {
-        // If we found a match in the cache, return it, but also
-        // update the entry in the cache in the background.
+        // Asynchronously update the entry in the cache,
+        // implementing the "stale-while-revalidate" pattern.
         event.waitUntil(cache.add(event.request));
+
+        // Immediately return the cached response.
         return cachedResponse;
       }
 
       // If we didn't find a match in the cache, use the network.
-      return fetch(event.request);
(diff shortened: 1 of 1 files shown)


jakearchibald:
The text above the example:

> This fetch event uses the browser default for non-GET requests. For GET requests it tries to return a match in the cache, and falls back to the network. If it finds a match in the cache, it asynchronously updates the cache for next time.

Whereas this PR adds an additional behaviour: If there's no match in the cache, it's added to the cache.

Indiscriminately adding items to the cache is one of the old AppCache gotchas, so I don't recommend deliberately recreating this pattern. You might do it for very particular URL patterns, like a URL pattern for avatars, but I think that's overcomplicating this example.

Maybe what's missing here is a statement/example that shows that the expected cache entries are set up in an install event?

sideshowbarker:
> Indiscriminately adding items to the cache is one of the old AppCache gotchas, so I don't recommend deliberately recreating this pattern. You might do it for very particular URL patterns, like a URL pattern for avatars, but I think that's overcomplicating this example.

Given that assessment, it seems like the existing example is serving the purpose it was written for — so I’m going ahead and closing this PR.

> Maybe what's missing here is a statement/example that shows that the expected cache entries are set up in an install event?

It seems like that could be handled in a different PR

PR #40654, opened 2025-08-06
Add `for...of` loop with `const` as modern best practice to "Creating closures in loops: A common mistake"

miskr-instructure:
### Description

As the PR title suggests, using `for (const <varname> of <iterable>) { ... }` should be a highly recommended syntax for developers when creating closures in loops (in general, it should probably be the most recommended way to loop on iterables, but that's not the scope of this PR).

### Motivation

 This syntax is the most concise (and arguably intuitive) out of all the other options, it should be highly recommended:
 
 * the iteration variable is defined with `const`: immutability is a good practice for a loop variable by default
 * it is the behavior a naive user may intuitively expect from the `let` and `var` variant of the loop, which don't work that way, but `const` does

### Additional details

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

### DIFF
--- a/files/en-us/web/javascript/guide/closures/index.md
+++ b/files/en-us/web/javascript/guide/closures/index.md
@@ -438,28 +438,6 @@ setupHelp();
 
 This works as expected. Rather than the callbacks all sharing a single lexical environment, the `makeHelpCallback` function creates _a new lexical environment_ for each callback, in which `help` refers to the corresponding string from the `helpText` array.
 
-If the index is not needed during iteration, using the **for...of** statement with the `const` modifier is the most modern and concise solution:
-
-```js
-function setupHelp() {
-  const helpText = [
-    { id: "email", help: "Your email address" },
-    { id: "name", help: "Your full name" },
-    { id: "age", help: "Your age (you must be over 16)" },
-  ];
-
-  for (const item of helpText) {
-    document.getElementById(item.id).onfocus = () => {
-      document.getElementById("help").textContent = item.help;
-    };
-  }
-}
-
-setupHelp();
-```
-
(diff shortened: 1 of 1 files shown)


files/en-us/web/javascript/guide/closures/index.md
```diff
@@ -438,6 +438,28 @@ setupHelp();
 
 This works as expected. Rather than the callbacks all sharing a single lexical environment, the `makeHelpCallback` function creates _a new lexical environment_ for each callback, in which `help` refers to the corresponding string from the `helpText` array.
 
+If the index is not needed during iteration, using the **for...of** statement with the `const` modifier is the most modern and concise solution:
```
miskr-instructure: I was considering putting this as the first solution (at line 400), but in context of this article, I guess the first solution is better at demonstrating how closures work, while this new solution is providing an overall "better code" but less demonstrative solution so I opted to put it secpmd.

Josh-Cena (approved):
Thanks :) I decided to merge it with the point about using `forEach`. We put the closure one first because that's the ES3 solution, and then we talk about ES5/ES6. WDYT?

miskr-instructure:
Looks good to me either way. Just thought it'd be nice to include the most modern JS solution _somewhere_, so that's done this way too 👍

miskr-instructure:
Also, I was actually mistaken myself initially, `for...of` works the same wrt. closures no matter whether `let` or `const` is used (like your correction says). I thought it only worked that way with `const`, so, TIL.

Josh-Cena:
Thanks! Great addition btw. Yeah `let` and `const` work the same in every regard except reassignability, so sometimes extra machinery is invented to avoid the closure problem. For example you may find https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for#lexical_declarations_in_the_initialization_block fun to read.

Issue #31841, opened 2024-01-21
Fetch Basic Sample Code 404


Fooeybar:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

### What specific section or headline is this issue about?

Fetch Basic Sample Code

### What information was incorrect, unhelpful, or incomplete?

The first sample given for fetch:
`async function logMovies() {
  const response = await fetch("http://example.com/movies.json");
  const movies = await response.json();
  console.log(movies);
}
`
results in a json parse error, since the sample url returns a 404.

Recommend a sample URL which will return 200 status and sample json, such as 'https://jsonplaceholder.typicode.com/users':
`async function logMovies() {
    const response = await fetch("https://jsonplaceholder.typicode.com/users");
    const users = await response.json();
    console.log(users);
  }
`

### What did you expect to see?

A successfully parsed JSON object printed to the console. This is the most basic example provided, from the source reference itself, and it should result in an error free response when copied & pasted.

### Do you have any supporting links, references, or citations?

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/api/fetch_api/using_fetch`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/fetch_api/using_fetch/index.md
* Last commit: https://github.com/mdn/content/commit/aae16b81e18d13dd006d418983558578563e9746
* Document last modified: 2023-08-18T00:00:48.000Z

</details>

bsmth:
Thanks for flagging, alternatively, we could do:

```js
// Define the function
async function logContributors() {
    const r = await fetch("https://developer.mozilla.org/en-US/docs/Glossary/TOFU/contributors.txt");
    const textData = await r.text();
    console.log(textData);
}
```

If we don't want to point too much traffic at `jsonplaceholder.typicode.com`. Note that the rest of the examples on the page point to `example.com` and so a lot of them will also fail, but it's possible this could be pointed out with a note that they're for illustration only.

wbamberg:
This is an interesting suggestion.

[...]

PR #40659, opened 2025-08-07
change the url

Shrinivassab:
### Description
Updates the Fetch API documentation to: --- Replace the non-working example URL (`example.org/products.json`) with a functional endpoint (`jsonplaceholder.typicode.com/users`)

Fix - #31841


wbamberg:
I'm closing this because I don't think we have consensus on what (if anything) we want to do here.

PR #40666, opened 2025-08-07
"Headers: forEach() method": `See also`: add link to Array.prototype.forEach

vitaly-zdanevich:


wbamberg:
Thanks for your PR! What's the reasoning for adding this link? I understand that reason for the link to Map's `forEach()`, as this API has similar behavior (i.e. it's a pair iterator (https://webidl.spec.whatwg.org/#idl-iterable). But not sure what value adding Array's version gives us here.

vitaly-zdanevich:
As a reader, I was surprised that such a basic link did not exist, coming from a search engine to this page, but looking for `Array.prototype.forEach`

wbamberg:
But what was the reason you were expecting to see such a link? This function is like Map's forEach, not Array's forEach. The `Headers` interface has what the spec calls a [pair iterator](https://webidl.spec.whatwg.org/#idl-iterable), about which it says:

> The JavaScript forEach method that is generated for a [value iterator](https://webidl.spec.whatwg.org/#dfn-value-iterator) invokes its callback like Array.prototype.forEach does, and **the forEach method for a [pair iterator](https://webidl.spec.whatwg.org/#dfn-pair-iterator) invokes its callback like Map.prototype.forEach does.**

So I don't see the point in linking to Array's forEach as well as Map's.

wbamberg:
Oh, or did you mean, you were searching for Array's forEach and ended up on the Headers.forEach page? I don't think that's a good justification for adding this see also. See also should list items that are semantically related to the item documented in the page, not unrelated items that the user might have been looking for.

wbamberg:
I'm closing this PR because AFAICT this is not a thing we want to do, but if there is a good reason for it, I'd be happy to hear it.

PR #40677, opened 2025-08-08
Fix HTMLOptionElement Option() constructor example

### LINKED ISSUE #40665: example results are not achieved with JS code
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option

https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option#append_options_with_different_parameters

The key value parameter is incorrect, as it is the same (`s.options.length`) for each new Option:
```
options.forEach((element, key) => {
  if (element === "zero") {
    s[key] = new Option(element, s.options.length, false, false);
  }
  if (element === "one") {
    s[key] = new Option(element, s.options.length, true, false); // Will add the "selected" attribute
  }
  if (element === "two") {
    s[key] = new Option(element, s.options.length, false, true); // Just will be selected in "view"
  }
});
```
Is give this:
Image

per the example output, it needs to produce:
```
/* Result
<select id="s">
  <option value="0">zero</option>
  <option value="1" selected="">one</option>
  <option value="2">two</option> // User will see this as 'selected'
</select>
*/
```

_No response_

I believe `s.options.length` should be `key`.

### MDN metadata

<details>

* Folder: `en-us/web/api/htmloptionelement/option`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/htmloptionelement/option/index.md
* Last commit: https://github.com/mdn/content/commit/e9b6cd1b7fa8612257b72b2a85a96dd7d45c0200
* Document last modified: 2025-04-10T14:56:07.000Z

</details>

adi-ray:
### Description
Corrected the option value parameter and separated HTML/JS code blocks.

### Motivation
The existing example code was using `s.options.length` instead of key which caused all the options to have same value. Additionally, the HTML was embedded in JS comments making it less readable.

Fixes #40665

### DIFF
--- a/files/en-us/web/api/htmloptionelement/option/index.md
+++ b/files/en-us/web/api/htmloptionelement/option/index.md
@@ -65,15 +65,11 @@ options.forEach((element, key) => {
 
 ### Append options with different parameters
 
-```js
-/* assuming we have the following HTML
-<select id="s">
-    <option>First</option>
-    <option>Second</option>
-    <option>Third</option>
-</select>
-*/
+```html
+<select id="s"></select>
+```
 
+```js
 const s = document.getElementById("s");
 const options = ["zero", "one", "two"];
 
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
Nice, thank you

PR #40688, opened 2025-08-10
docs(css): add performance considerations for :has() pseudo-class

### LINKED ISSUE #40280: Add CSS `:has()` selector performance tips
### Proposal

I recently learned from Firefox developers that certain forms of CSS `:has()` selectors can have vastly different performance characteristics when I reported a [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1896380).

They shared the following tips:

> anchoring :has() to an element (or to all elements) slow all DOM mutations inside that subtree (more the more complex the selector inside :has() is). So stuff like body:has() or :root:has() or *:has() are just a recipe for bad performance

> 1. :has(...) with an under-constrained compound may trigger many subtree-traversals: e.g. With :has(.foo), every DOM mutation will cause subtree traversals to see if that element does have .foo in the subtree.
> 2. :has(...) with an under-constrained inner selector compound may trigger many ancestor-traversals: e.g. With .ancestor:has(.foo > *), any DOM mutation will cause ancestor traversals to see if there's a :has() anchor to be invalidated. i.e. We need to check if the mutated DOM element has a .foo parent, and if we do, need to check for .ancestor to restyle.
> 3. Any occurrence of above can affect invalidation performance of otherwise well-constrained :has() selectors.

I would like to add this somewhere on MDN so these performance implications can become more widely known, but I wasn't sure where to put it.

Potential pages that occurred to me include:

- [:has()](https://developer.mozilla.org/en-US/docs/Web/CSS/:has)
- [CSS performance optimization](https://developer.mozilla.org/en-US/docs/Learn/Performance/CSS)

I wasn't sure if the `:has()` page itself was okay or not, as it might be required to match a very specific format like other CSS pages.

### Browser support

_No response_

### Tasks

- [ ]

### Dependencies

- [ ]

### Additional information

_No response_

Yes, I can help write content once the contect place for it to live is decided.

Shrinivassab:
### Description

Adds performance considerations section to `:has()` documentation based on feedback from Firefox engineers.

Fixes #40280


Josh-Cena:
Don't duplicate explanations. Write it in one place and link to it from the other place; I think here we should write it on the `:has()` page and keep the learning page's heading but reduce its content to a single paragraph that links to the reference.

Shrinivassab:
@Josh-Cena Thanks for the review... Even I had the same thought, not sure about the MDN approach ---  Now, I consolidated all performance details in the `:has()` reference page...

Josh-Cena:
I'll defer to @chrisdavidmills for the actual review

files/en-us/web/css/_colon_has/index.md
```diff
@@ -217,6 +217,66 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
 
+## Performance considerations
+
+Certain uses of the `:has()` pseudo-class can significantly impact page performance, particularly during dynamic updates (DOM mutations). Browser engines must re-evaluate `:has()` selectors when the DOM changes, and complex or poorly constrained selectors can lead to expensive computations.
+
+### Avoid broad anchoring
+
+Anchoring `:has()` to very general selectors like `body`, `:root`, or `*` can degrade performance. Because `:has()` depends on the descendants of its anchor, any DOM change within the entire subtree of a broadly selected element requires the browser to re-check the `:has()` condition. The more complex the selector inside `:has()`, the greater the impact.
+
+```css example-bad
```
chrisdavidmills, line 252: These examples still feel quite specific...

files/en-us/web/css/_colon_has/index.md
```diff
@@ -217,6 +217,66 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
+}
+:root:has(main > article[data-priority="high"]) {
+  /* styles */
+}
+*:has(> img[data-loaded="false"]) {
+  /* styles */
+}
+```
+
+Instead, anchor `:has()` to the most specific element possible that still logically contains the elements you are checking for.
```
chrisdavidmills: ...below here, add an `example-good` block showing what the selectors should be to follow best practice. A before and after, if you like.

files/en-us/web/css/_colon_has/index.md
```diff
@@ -217,6 +217,66 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
+
+### Minimize subtree traversals
```
chrisdavidmills, line 280: Reading this subsection and the previous one, they sound like the same issue but written in slightly different words. Is this the case? If they do differ, how? It is not obvious to me at the moment.

Did you use AI to help you generate some of this text? Generating different text to describe the same problem and presenting it as separate problems is a common issue with this approach, so check the output carefully if you do so.

files/en-us/web/css/_colon_has/index.md
```diff
@@ -217,6 +217,66 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
+Certain selector patterns involving `:has()` can force the browser to traverse up the ancestor chain for every DOM mutation, looking for potential `:has()` anchors that might need updating. This happens when the structure implies a need to check ancestors of the mutated element.
+
+```css example-bad
+/* Might trigger ancestor traversal */
+.ancestor:has(.foo > *) {
+  /* styles */
+}
+```
+
+In this example, any DOM change requires checking if the changed element is the `*` (any element) that is a direct child of `.foo`, and if its parent (or further ancestors) is `.ancestor`.
```
chrisdavidmills: Again, might be good to show an `example-good` block to illustrate how to fix the problem - before and after.

chrisdavidmills (changes requested):
Hi @Shrinivassab, and thank you for your continued work on MDN. I've provided a review, with a few suggestions and questions.

files/en-us/web/css/_colon_has/index.md
Josh-Cena: In a selector like `A:has(B)`, the first section says "A can't be too broad (avoid `body`, `:root`, `*`, etc.)", this says "B can't be too broad (use `>`, etc.)", so it makes sense to me.

chrisdavidmills: OK, that makes sense. But I think my point still stands that the difference could be made clearer.

files/en-us/web/css/_colon_has/index.md
Shrinivassab: Does this example works, Please share your thought?
/* Constrain the inner selector to avoid ancestor traversals */
.ancestor:has(.foo > .specific-child) {
  /* styles */
}

files/en-us/web/css/_colon_has/index.md
Shrinivassab: Hi @chrisdavidmills and @Josh-Cena, thank you for the feedback... In order to distinguish between these two subsection(Avoid vs Minimize), I will make the below changes in my next commit.

1. "Avoid broad anchoring," I will add an `example-good` as you suggested with specific selectors (e.g., `.container:has(.sidebar-expanded)`) to contrast with broad selectors (`body`, `:root`, `*`), making the guidance more actionable.
2. To clarify the distinction, I will revise both sections with introductory sentences: "Avoid broad anchoring" focuses on the anchor selector (`A` in `A:has(B)`), while "Minimize subtree traversals" addresses the inner selector (`B`). At the end the note will summarizes these differences.
3. I will make sure the revised content is concise and distinct.

--- Pls share your thought... Thanks in advance

files/en-us/web/css/_colon_has/index.md
Shrinivassab: share your thought about these example: 
/* Prefer specific anchors to limit scope */
.container:has(.sidebar-expanded) {
  /* styles */
}
.content-wrapper:has(> article[data-priority="high"]) {
  /* styles */
}
.gallery:has(> img[data-loaded="false"]) {
  /* styles */
}

files/en-us/web/css/_colon_has/index.md
Shrinivassab: Hi @chrisdavidmills, thanks for your suggestion. Sure I will add a `example-good` block, targeting more "specific anchor selectors like (e.g., `.container`, `.content-wrapper`, `.gallery`) to contrast with the problematic broad selectors like (`body`, `:root`, `*`)". So that these examples will provide a practical guidance for developers to optimize `:has()` usage while keeping the scope manageable. Please share your thought about the example also, If anything can be improved?

files/en-us/web/css/_colon_has/index.md
chrisdavidmills: So, you've still not added the `example-good` block, and the examples in the `example-bad` block still feel quite specific.

files/en-us/web/css/_colon_has/index.md
```diff
@@ -217,6 +217,66 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
+
+When the selector _inside_ `:has()` is not tightly constrained (for example, it doesn't use child `>` or sibling `+` / `~` combinators), the browser might need to traverse the entire subtree of the anchor element on every DOM mutation to check if the condition still holds.
+
+```css example-bad
+/* May trigger full subtree traversal */
+.ancestor:has(.foo) {
+  /* styles */
+}
+```
+
```
chrisdavidmills, line 292: This section is definitely better now, and the difference between it and the previous section is better explained.

I would suggest putting a sentence in between the two code blocks to introduce the `example-good` block and state briefly why these examples are better. I know you've kind of done that in the comments, but it would still look better to have a brief explanatory sentence here.

files/en-us/web/css/_colon_has/index.md
chrisdavidmills: Yup, makes sense. Like the previous section, include an explainer sentence between the two code blocks.

chrisdavidmills (changes requested):
Thanks for your further work on this, @Shrinivassab. There is a bit more to do, but this is definitely going in the right direction now.

files/en-us/web/css/_colon_has/index.md
```diff
@@ -241,6 +241,90 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
+Certain selector patterns involving `:has()` can force the browser to traverse up the ancestor chain for every DOM mutation, looking for potential `:has()` anchors that might need updating. This happens when the structure implies a need to check ancestors of the mutated element.
+
+```css example-bad
+/* Might trigger ancestor traversal */
+.ancestor:has(.foo > *) {
+  /* styles */
+}
+```
+
+In this example, any DOM change requires checking if the changed element is the `*` (any element) that is a direct child of `.foo`, and if its parent (or further ancestors) is `.ancestor`. Constraining the inner selector with specific classes or direct child combinators reduces ancestor traversals, improving performance.
```
Shrinivassab: I tried to explain each section as much as possible- Please let me know if this part is redundant? "Constraining the inner..."

Shrinivassab (commented):
Sorry for the delay... I have updated the wording, please let me know if there is any grammatical or redundant information. As there is subtle difference, I might over explained it, happy to correct it based on MDN standard.

files/en-us/web/css/_colon_has/index.md
```diff
@@ -241,6 +241,90 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
+  /* styles */
+}
+.gallery:has(> img[data-loaded="false"]) {
+  /* styles */
+}
+```
+
+### Minimize subtree traversals
+
+This section focuses on the inner selector (the `B` in `A:has(B)`), which should use combinators like `>` or `+` to limit traversal. When the selector inside `:has()` is not tightly constrained (for example, it doesn't use child `>` or sibling `+` / `~` combinators), the browser might need to traverse the entire subtree of the anchor element on every DOM mutation to check if the condition still holds.
```
chrisdavidmills:
```suggestion
This section focuses on the inner selector (the `B` in `A:has(B)`), which should use combinators like `>` or `+` to limit traversal. When the selector inside `:has()` is not tightly constrained, the browser might need to traverse the entire subtree of the anchor element on every DOM mutation to check if the condition still holds.

In this example, any change within `.ancestor` requires checking all descendants for `.foo`:
```

chrisdavidmills: Little bit too much repetition here.

chrisdavidmills: I've also added a bit from the paragraph below, which I think can be removed after its text has been better placed.

files/en-us/web/css/_colon_has/index.md
```diff
@@ -241,6 +241,90 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
+
+```css example-bad
+/* May trigger full subtree traversal */
+.ancestor:has(.foo) {
+  /* styles */
+}
+```
+
+Using child or sibling combinators limits the scope of the inner selector, reducing the performance cost of DOM mutations.
```
chrisdavidmills:
```suggestion
Using child or sibling combinators limits the scope of the inner selector, reducing the performance cost of DOM mutations. In this example, the browser only needs to check direct children or a specific sibling's descendants:
```

files/en-us/web/css/_colon_has/index.md
```diff
@@ -241,6 +241,90 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
+/* More constrained - limits traversal */
+.ancestor:has(> .foo) {
+  /* direct child */
+}
+.ancestor:has(+ .sibling .foo) {
+  /* descendant of adjacent sibling */
+}
+```
+
+In the first example, any change within `.ancestor` requires checking all descendants for `.foo`. In the good examples, the browser only needs to check direct children or a specific sibling's descendants.
```
chrisdavidmills: remove

files/en-us/web/css/_colon_has/index.md
```diff
@@ -241,6 +241,90 @@ The analogous construct in CSS would be `.abc:has(+ .xyz)`: it selects the eleme
...
+
+### Be mindful of ancestor traversals
+
+Certain selector patterns involving `:has()` can force the browser to traverse up the ancestor chain for every DOM mutation, looking for potential `:has()` anchors that might need updating. This happens when the structure implies a need to check ancestors of the mutated element.
```
chrisdavidmills:
```suggestion
Certain selector patterns involving `:has()` can force the browser to traverse up the ancestor chain for every DOM mutation, looking for potential `:has()` anchors that might need updating. This happens when the structure implies a need to check ancestors of the mutated element.

In this example, any DOM change requires checking if the changed element is the `*` (any element) that is a direct child of `.foo`, and if its parent (or further ancestors) is `.ancestor`.
```

files/en-us/web/css/_colon_has/index.md
chrisdavidmills:
```suggestion
Constraining the inner selector with specific classes or direct child combinators reduces ancestor traversals, improving performance.
```

chrisdavidmills: Maybe add a sentence to explain specifically how this relates to the next example: "In the next example..."

chrisdavidmills (changes requested):
@Shrinivassab nice work; this is getting really close now! My suggestions below are now fairly minor restructuring.

files/en-us/web/css/_colon_has/index.md
Shrinivassab: Thanks a lot for the suggestion and helping to reduce the redundancy.

files/en-us/web/css/_colon_has/index.md
Shrinivassab: Removed the section...

files/en-us/web/css/_colon_has/index.md
Shrinivassab: In the next example, using a specific class like `.specific-child` avoids ancestor traversals by limiting the browser’s check to a well-defined element   --- Please let me know your suggestion about this text. If it looks good i can add it before the good example section.

chrisdavidmills: Hrm, this would make the paragraph rather repetitive. How about just modifying the existing sentence, something like this:

> Constraining the inner selector with specific classes or direct child combinators (see how `.specific-child` is used in the next code snippet, for example) reduces ancestor traversals by limiting the scope of the DOM section the browser needs to check to a well-defined element, improving performance.

?

Shrinivassab: Thanks a lot for your suggestion, that improvise a lot the current wording. Your suggestion spot on -  I made a small few word changes, please share your thought. "Constraining the inner selector with specific classes or direct child combinators `(e.g: .specific-child in the next snippet)` reduces expensive ancestor traversals by limiting the browser's check to a well defined element, improving performance."

I will use your wording as it is, if mine is not spot on...

chrisdavidmills: Sounds good; make the update, and I'll see if it needs any small tweaks after that.

chrisdavidmills (approved):
OK, I've read through this again, made a couple of small tweaks, and now I think it is ready to merge. Well done to everyone involved for the hard work on this, and especially @Shrinivassab!

PR #40696, opened 2025-08-11
Fix grammar in captureEvents() note

neeko-cat:
### Description

Fix grammar in the captureEvents() note by replacing “This is an method” with “This method was…”.

### Motivation

Readability and correctness of the documentation.

### DIFF
--- a/files/en-us/web/api/window/captureevents/index.md
+++ b/files/en-us/web/api/window/captureevents/index.md
@@ -13,7 +13,7 @@ browser-compat: api.Window.captureEvents
 The **`Window.captureEvents()`** method does nothing.
 
 > [!NOTE]
-> This is an method long removed from the specification. It is kept in browsers to prevent code breakage but does nothing.
+> This method was long removed from the specification. It is kept in browsers to prevent code breakage but does nothing.
 
 ## Syntax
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/window/captureevents/index.md
```diff
@@ -13,7 +13,7 @@ browser-compat: api.Window.captureEvents
 The **`Window.captureEvents()`** method does nothing.
 
 > [!NOTE]
-> This is an method long removed from the specification. It is kept in browsers to prevent code breakage but does nothing.
+> This method was long removed from the specification. It is kept in browsers to prevent code breakage but does nothing.
```
wbamberg: Thank you for finding the grammar error! This is still a bit weird, because (1) technically the method hasn't been removed, and (2) we already said it does nothing, so this is repetitive.

Perhaps we could remove this note and update the intro paragraph?

> The **`Window.captureEvents()`** method does nothing. Its original behavior has been removed from the specification, but the method itself has been retained so as not to break code that calls it.

neeko-cat: yeah makes sense!

wbamberg (approved):
👍 thank you!

PR #40714, opened 2025-08-13
fix cssxref typo

zhangenming:
It might be necessary to have a universal method to find all similar errors


Josh-Cena (approved):
Good catch, thanks

aalonso777777 (commented):
Okay thanks aalonso777777@gmail.gmail.com

PR #40728, opened 2025-08-13
Add "See also" links to Temporal polyfills

arshaw:
### Description

Adds the `@js-temporal/polyfill` and `temporal-polyfill` polyfills to the Temporal "See also" section.

### Motivation

It might be useful to link to the two established Temporal polyfills for people wishing to use the API now.

### Additional details

I've attempted to follow MDN-content conventions for linking to polyfills.

I am the author of the "FullCalendar" Temporal polyfill. Also, I've been working with the Temporal champions over the past three years to help refine the spec.

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/temporal/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/temporal/index.md
@@ -377,5 +377,5 @@ The `Temporal` objects will refuse to construct an instance representing a date/
 - {{jsxref("Intl.DateTimeFormat")}}
 - {{jsxref("Intl.RelativeTimeFormat")}}
 - {{jsxref("Intl.DurationFormat")}}
-- [Temporal Polyfill by proposal champions](https://github.com/js-temporal/temporal-polyfill#readme) (also available as an [npm package](https://www.npmjs.com/package/@js-temporal/polyfill))
-- [Temporal Polyfill by FullCalendar](https://github.com/fullcalendar/temporal-polyfill#readme) (also available as an [npm package](https://www.npmjs.com/package/temporal-polyfill))
+- [Temporal polyfill by proposal champions](https://www.npmjs.com/package/@js-temporal/polyfill)
+- [Temporal polyfill by FullCalendar](https://www.npmjs.com/package/temporal-polyfill)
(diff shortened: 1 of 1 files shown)


files/en-us/web/javascript/reference/global_objects/temporal/index.md
```diff
@@ -377,3 +377,5 @@ The `Temporal` objects will refuse to construct an instance representing a date/
 - {{jsxref("Intl.DateTimeFormat")}}
 - {{jsxref("Intl.RelativeTimeFormat")}}
 - {{jsxref("Intl.DurationFormat")}}
+- [Temporal Polyfill by proposal champions](https://github.com/js-temporal/temporal-polyfill#readme) (also available as an [npm package](https://www.npmjs.com/package/@js-temporal/polyfill))
+- [Temporal Polyfill by FullCalendar](https://github.com/fullcalendar/temporal-polyfill#readme) (also available as an [npm package](https://www.npmjs.com/package/temporal-polyfill))
```
Josh-Cena:
```suggestion
- [Temporal polyfill by FullCalendar](https://www.npmjs.com/package/temporal-polyfill)
```

Thanks. I personally use your polyfill :) I suggest removing the other one because it's intended for experimentation and isn't really great for production (or so I heard)

arshaw: Thanks for looking at this @Josh-Cena. Hopefully my polyfill is serving you well!

What you're describing about the other polyfill sounds like the distinction between `@js-temporal/polyfill` vs the "non-production" polyfill described [here](https://github.com/tc39/proposal-temporal/tree/main/?tab=readme-ov-file#polyfills), which is basically an unoptimized reference implementation that powers the playground on the [proposal-temporal docs](https://tc39.es/proposal-temporal/docs/).

I would suggest still keeping `@js-temporal/polyfill` in the MDN list. It's more than twice the size, but it is slightly more compliant than FullCalendar's. You can see the FullCalendar plugin's deviations [here](https://github.com/fullcalendar/temporal-polyfill/blob/main/packages/temporal-polyfill/scripts/test262-config/expected-failures.txt). Some people might appreciate the strictness, and don't care about file size, like those using Node-only environments.

Josh-Cena: Thanks. Let's go with this then:

```suggestion
- [Temporal polyfill by proposal champions](https://www.npmjs.com/package/@js-temporal/polyfill)
- [Temporal polyfill by FullCalendar](https://www.npmjs.com/package/temporal-polyfill)
```

arshaw: Sounds good to me @Josh-Cena. Committed

Josh-Cena (approved):
Nice, thank you!

PR #40729, opened 2025-08-13
Fix typo

### LINKED ISSUE #40726: TYPO: "The" should be "This"
### MDN URL

https://developer.mozilla.org/en-US/docs/Glossary/Browsing_context

Browsing context 4th paragraph

By default a document opened from a browser context group is opened in the same group whether or not it is cross-origin or same-origin. The Cross-Origin-Opener-Policy can be used to control whether the document is instead opened in its own new browsing context group and cross-origin isolated from other contexts (in particular cross-origin contexts). **The** can mitigate the risk of cross-origin attacks and the side-channel attacks referred to as XS-Leaks.

By default a document opened from a browser context group is opened in the same group whether or not it is cross-origin or same-origin. The Cross-Origin-Opener-Policy can be used to control whether the document is instead opened in its own new browsing context group and cross-origin isolated from other contexts (in particular cross-origin contexts). **This** can mitigate the risk of cross-origin attacks and the side-channel attacks referred to as XS-Leaks.

_No response_

_No response_

dipikabh:
### Description

Fixes https://github.com/mdn/content/issues/40726

...plus a few other edits.

### DIFF
--- a/files/en-us/glossary/browsing_context/index.md
+++ b/files/en-us/glossary/browsing_context/index.md
@@ -10,14 +10,14 @@ In modern browsers, it usually is a _tab_, but it can be a _window_, a _popup_, 
 
 Each browsing context has an origin (that of the active document) and an ordered history of previously displayed documents.
 Communication and resource sharing between browsing contexts is constrained, in particular between cross-origin contexts.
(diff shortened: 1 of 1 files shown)


hamishwillee (approved):
Thank you. If I wrote that, it was a very bad day :-(

PR #40730, opened 2025-08-14
Fix grammar in list of abstract roles to avoid

redwoodsteve:
Added a space in the list.
List before:
..., ..., ..., ...,and ...
List now:
..., ..., ..., ..., and ...
This will probably help readers understand the list a bit better and help keep the flow of the page when reading.


Josh-Cena (approved):
Nice catch. Thanks

PR #40734, opened 2025-08-14
Bug-1542358 omnibox.onInputEntered is a user action trigger

rebloor:
### Description

This change:

- adds the information that clicking an extension suggestion in the address bar (omnibox) is now a user action, including a release note.
- adds details of all the user action enabled APIs.
- tidies up the content on the user action page.

### Motivation

In addition to the change requested as part of a dev-doc-needed, the user actions page seemed hard to read and didn't provide a comprehensive guide to the related API's.

[Bug 1542358](https://bugzilla.mozilla.org/show_bug.cgi?id=1542358) Allow omnibox.onInputEntered as a user action trigger e.g. for clipboard & browserAction.openPopup()

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/user_actions/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/user_actions/index.md
@@ -18,7 +18,7 @@ The APIs enabled by a user action are:
 - The {{WebExtAPIRef("action.openPopup")}}, {{WebExtAPIRef("browserAction.openPopup")}}, and {{WebExtAPIRef("pageAction.openPopup")}} APIs that open an extension's browser or page action popup. Users do the same by clicking the browser or page action.
 - The {{WebExtAPIRef("sidebarAction.open")}}, {{WebExtAPIRef("sidebarAction.close")}}, and {{WebExtAPIRef("sidebarAction.toggle")}} APIs open and close an extension's sidebar. The user does this from some part of the browser's built-in user interface, such as the **View** > **Sidebar** menu.
 - The {{WebExtAPIRef("downloads.open")}} API opens a downloaded file. The user does this from some part of the browser's built-in user interface, such as the **Tools** > **Downloads** menu.
-- The {{WebExtAPIRef("management.setenabled")}} API. The user can turn off a theme extension from the extension's Add-on Manager page.
+- The {{WebExtAPIRef("management.setEnabled")}} API. The user can turn off a theme extension from the extension's Add-on Manager page.
 - The {{WebExtAPIRef("permissions.request")}} API. The user can grant permissions from the extension's Add-on Manager permissions and data tab.
 
 For example:
--- a/files/en-us/mozilla/add-ons/webextensions/user_actions/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/user_actions/index.md
@@ -5,15 +5,7 @@ page-type: guide
 sidebar: addonsidebar
 ---
 
-Some WebExtension APIs perform functions that usually occur as a result of a user action. These APIs are:
-
-- The {{WebExtAPIRef("action.openPopup")}}, {{WebExtAPIRef("browserAction.openPopup")}}, and {{WebExtAPIRef("pageAction.openPopup")}} APIs that open an extension's browser or page action popup. Users do the same by clicking the browser or page action.
-- The {{WebExtAPIRef("sidebarAction.open")}}, {{WebExtAPIRef("sidebarAction.close")}}, and {{WebExtAPIRef("sidebarAction.toggle")}} APIs open and close an extension's sidebar. The user does this from some part of the browser's built-in user interface, such as the **View** > **Sidebar** menu.
-- The {{WebExtAPIRef("downloads.open")}} API opens a downloaded file. The user does this from some part of the browser's built-in user interface, such as the **Tools** > **Downloads** menu.
-- The {{WebExtAPIRef("management.setenabled")}} API. The user can turn off a theme extension from the extension's Add-on Manager page.
-- The {{WebExtAPIRef("permissions.request")}} API. The user can grant permissions from the extension's Add-on Manager permissions and data tab.
-
-Following the principle of "no surprises", these APIs can only be called from inside the handler for a user action. User actions include:
+Some WebExtension APIs perform functions that usually occur as a result of a user action (also referred to as user gestures and user activation). Following the principle of "no surprises", these APIs can only be called from inside the handler for a user action. These user actions are:
 
 - Clicking the extension's browser action or page action.
 - Selecting a context menu item defined by the extension.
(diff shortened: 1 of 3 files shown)


files/en-us/mozilla/add-ons/webextensions/user_actions/index.md
```diff
@@ -5,17 +5,21 @@ page-type: guide
...
-- if an extension adds a sidebar, it is typically opened by the user via some part of the browser's built-in user interface, such as the View/Sidebar menu. But there's also a {{WebExtAPIRef("sidebarAction.open")}} API enabling an extension to open their sidebar programmatically.
+- The {{WebExtAPIRef("action.openPopup")}}, {{WebExtAPIRef("browserAction.openPopup")}}), and {{WebExtAPIRef("pageAction.openPopup")}}) APIs that open an extension's browser or page action popup. Users do the same by clicking the browser or page action.
+- The {{WebExtAPIRef("sidebarAction.open")}}, {{WebExtAPIRef("sidebarAction.close")}}, and {{WebExtAPIRef("sidebarAction.toggle")}} APIs open and close an extensions sidebar. The user does this from some part of the browser's built-in user interface, such as the **View** > **Sidebar** menu.
+- The {{WebExtAPIRef("downloads.open")}} API opens a downloaded file. The user does this from some part of the browser's built-in user interface, such as the **Tools** > **Downloads** menu.
+- The {{WebExtAPIRef("management.setenabled")}} API. The user can disable a theme extension from the extension's Add-on Manager page.
```
Rob--W:
```suggestion
- The {{WebExtAPIRef("management.setEnabled")}} API. The user can disable a theme extension from the extension's Add-on Manager page.
```

files/en-us/mozilla/add-ons/webextensions/user_actions/index.md
```diff
@@ -5,17 +5,21 @@ page-type: guide
...
 
-To follow the principle of "no surprises", APIs like this can only be called from inside the handler for a user action. User actions include the following:
+Following the principle of "no surprises", these APIs can only be called from inside the handler for a user action. User actions include:
 
 - Clicking the extension's browser action or page action.
 - Selecting a context menu item defined by the extension.
 - Activating a keyboard shortcut defined by the extension (this only treated as a user action from Firefox 63 onwards).
-- Clicking a button in a page bundled with the extension.
+- Clicking a button on a page bundled with the extension.
+- Clicking an extension suggestion in the address bar (omnibox). (This action also grants the `"activeTab"` permission.)
```
Rob--W: activeTab is not unique to this API; it is granted with all of the listed examples.

Also added compatibility information for consistency with the other information listed above.

```suggestion
- Clicking an extension suggestion in the address bar (omnibox). (this only treated as a user action from Firefox 143 onwards).
```

Rob--W (approved):
Approved with the following feedback addressed.

files/en-us/mozilla/add-ons/webextensions/user_actions/index.md
rebloor:
```suggestion
- Clicking an extension suggestion in the address bar (omnibox) (this is treated as a user action from Firefox 143).

In addition to enabling the APIs, these actions also grant the `"activeTab"` permission.
```

rebloor: @Rob--W I added the comment about activeTab as this isn't mentioned on the page. I'm not sure how the reader would know this without us stating it. So is my change OK?

Rob--W: I agree that it would make sense to reference activeTab. Could you use phrasing consistent with what is mentioned at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#activetab_permission and also link to this "user action" article from that page?

rebloor: @Rob--W done. Although the redundancy between the two articles is not ideal. Perhaps a separate exercise we could consider moving all of the activeTab details into the user action page?

Rob--W: In an attempt to address the duplication issue that you raised, I took a closer look at what we had and noticed inconsistencies that is caused by a reframe of the "user action" article.

Originally, the "user action" article's purpose was to document the special requirement that some API methods had in common, coincidentally with examples from the "action" group of APIs (browser action, sidebar action).

Now you tried to generalize it, but the introduction has expanded to exhaustively list all APIs, which distracts from the list of actual user actions.

To fix this, I would start with the current introduction that some APIs require user action, then list the qualifying actions, and then list the APIs themselves. When used in this way, the article becomes a more obvious canonical reference for the concept, and it would be much easier to drop the listing from the activeTab section.

(Side note: activeTab is a relatively large article, yet difficult to find when searching for "mdn activeTab". I would be in favor of a dedicated activeTab article that can be linked from various places)

For completeness, "user action" is presented as if it is an established name for the concept. While understandable, it is not a universally used name. Other related terms in use are:

-In Chrome, "user gesture", e.g. mentioned in their activeTab docs at https://developer.chrome.com/docs/extensions/develop/concepts/activeTab
- In Firefox, "user interaction" or "user input" internally.
- On the web platform, "user activation" - https://developer.mozilla.org/en-US/docs/Web/API/UserActivation

files/en-us/mozilla/add-ons/webextensions/user_actions/index.md
```diff
@@ -5,17 +5,21 @@ page-type: guide
 sidebar: addonsidebar
 ---
 
-Some WebExtension APIs perform functions that are generally performed as a result of a user action. For example:
-
-- a browser action with a popup will display the popup when the user clicks it, but there's also a {{WebExtAPIRef("browserAction.openPopup")}} API enabling an extension to open the popup programmatically.
-- if an extension adds a sidebar, it is typically opened by the user via some part of the browser's built-in user interface, such as the View/Sidebar menu. But there's also a {{WebExtAPIRef("sidebarAction.open")}} API enabling an extension to open their sidebar programmatically.
-
-To follow the principle of "no surprises", APIs like this can only be called from inside the handler for a user action. User actions include the following:
+Some WebExtension APIs perform functions that usually occur as a result of a user action (also referred to as user gestures and user activation). Following the principle of "no surprises", these APIs can only be called from inside the handler for a user action. These user actions are:
```
Rob--W: The way this is phrased is odd. An API is gated on user gestures/interaction, not because the APIs are typically user-generated, but because they are powerful or annoying if triggered without explicit user interaction. Here is an attempt to move some words around, but I am sure that it can be better.

```suggestion
Some WebExtension APIs perform functions that usually occur as a result of a user action. Following the principle of "no surprises", these APIs can only be called from inside the handler for a user action (also referred to as user gestures). These user actions are:
```

I also dropped the reference to user activation. What counts as user activation on the web is not counting as a user action in extension APIs. There are discussions ongoing about whether it should, but as of now it doesn't. (This fact is already mentioned in the current article at "User interactions in normal web pages aren't treated as user actions").

PR #40735, opened 2025-08-14
Add background color change example to <button> element documentation

tejsapkal:

### DIFF
--- a/files/en-us/web/html/reference/elements/button/index.md
+++ b/files/en-us/web/html/reference/elements/button/index.md
@@ -225,9 +225,9 @@ Whether clicking on a `<button>` or {{HTMLElement("input")}} button types causes
 ## Examples
 
 ```html
-<button name="button">Press me</button>
-```
 
+<button onclick="this.style.backgroundColor='lightgreen'">Click me</button>
+```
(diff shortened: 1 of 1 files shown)


chrisdavidmills (commented):
Hi @tejsapkal, and welcome to MDN.

I'm afraid I don't agree with this change.

- First of all, you are using an inline event handler attribute, which is considered bad practice in this day and age (in terms of code separation, unobtrusive JS, etc.)
- Second, you've changed "Press me" to "Click me", which is less good as it makes assumptions about how the user is going to activate the button (what about smartphone or keyboard users?)
- Third, pressing a button to change its own background color is not very real-world, on its own. You might do that, plus activate some kind of functionality.
- Fourth, we always like to include a minimal functionality example first on the page, before then showing something more complex afterwards. This example is fine as it is.

I'd be inclined to just close this one, but if you do want to add a slightly more interesting example, I would suggest:

- Leaving the existing example as-is, but adding an H3 heading of "Basic `<button>` example"
- Adding a more interesting example below it, under its own H3, that shows using a button to activate some kind of simple functionality, like a "Machine on / Machine off" label toggle. Use `addEventListener()` rather than an inline event handler.

PR #40747, opened 2025-08-15
FF142 Relnote/Expr features: Integrity-Policy can be enforced on styles behind pref

hamishwillee:
FF142 adds support for specifying a style as a blocked destination (behind a preference) in https://bugzilla.mozilla.org/show_bug.cgi?id=1974247. This updates the release note and experimental features to note that this is an allowed value for the `Integrity-Policy` and `Integrity-Policy-Report-Only` headers.

Note that I did these as separate entries in relnote and experimental features. I could have done them together perhaps, but they are under different preferences, and may well release separately.

Related docs work can be tracked in #40667


pepelsbey (approved):
Looks good! Thank you :)

PR #40756, opened 2025-08-15
Update example to check for status on fetch response

michaelmberger:
### Description

Add a check for the fetch response status to make sure the response is ok 

### Motivation

Change in code improves the example to use best practice and improves understandability

### Additional details
Discussion
https://discord.com/channels/1009925603572600863/1405917321767292979/1405939031958618153

### DIFF
--- a/files/en-us/web/api/response/text/index.md
+++ b/files/en-us/web/api/response/text/index.md
@@ -64,7 +64,7 @@ function getData(pageId) {
         throw new Error(`HTTP error, status = ${response.status}`);
       }
       return response.text();
-    )}
+    })
     .then((text) => {
       myArticle.innerText = text;
     });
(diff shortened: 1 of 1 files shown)


sideshowbarker:
@wbamberg requesting your review since the Discord discussion seems to indicate you’re familiar with the context

wbamberg (commented):
Thank you for the PR! If we're updating this to match [the version in dom-examples](https://github.com/mdn/dom-examples/blob/main/fetch/fetch-text/index.html) that the text references, we ought also to add a `catch` block to handle the error that gets thrown.

michaelmberger:
@wbamberg Good point, sorry for the delay. It's been added.

wbamberg (approved):
👍 thank you for the fix, @michaelmberger !

PR #40758, opened 2025-08-16
Fix two typos

doctoon:
### Description

Fixes two typos in the extension CSP page.

### Motivation

So the reader doesn't get confused :D

### Additional details

None.

None.


Josh-Cena (approved):
Thank you

PR #40771, opened 2025-08-18
Rearrange, make `content` more obvious

zhangenming:


chrisdavidmills (approved):
@zhangenming I agree with this. We tend to order lists alphabetically anyway, so this fits well with our guidelines.

Thanks for your MDN contribution!

Issue #40780, opened 2025-08-18
Small contradiction in "Attacks/XSS" about contexts


Seb35:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/XSS

### What specific section or headline is this issue about?

Document contexts

### What information was incorrect, unhelpful, or incomplete?

There is a logical contradiction between the second and the fourth code snippet in the paragraph [Document contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/XSS#document_contexts):
* if the second code snippet (`<div {{ my_input }}></div>
`) has an XSS with the payload `onmouseover="alert('XSS')"` (the text says it is XSS-vulnerable)
* then the fourth code snippet `<div class="{{ my_class }}">...</div>` has an XSS with the payload `some_id" onmouseover="alert('XSS!')` (but the text says it is **not** XSS-vulnerable) (NB: the place of the double-quotes, assumed not escaped in the second code snippet).

After checking the [escaping rules of Django](https://docs.djangoproject.com/en/5.2/ref/templates/language/#automatic-html-escaping) [I don’t know well Django], it turns out that single-quotes and double-quotes are escaped, so the fourth code snipped is indeed XSS-safe (as said), but the payload described for the second code snippet does not trigger an XSS (as it is said).

I just installed Django 4.2 and confirmed that the payload for the second code snippet does not trigger an XSS, and the payload the fourth code snipped does not trigger an XSS.

### What did you expect to see?

The second code snippet is indeed XSS-vulnerable but for a more complex payload: ```onmouseover=alert(`XSS!`)``` (or even ```onmouseover=alert`XSS!` ```) or if we want to hide the complexity of the backtick ```onmouseover=alert(1)```

It should be precised that the double-quote is escaped as `&quot;`, so an attacker cannot use it (and this is a clue for the reader [like me] wondering why the fourth code snippet is not XSS-vulnerable).

### Do you have any supporting links, references, or citations?

* [Escaping rules of Django](https://docs.djangoproject.com/en/5.2/ref/templates/language/#automatic-html-escaping)

### Do you have anything more you want to share?

Many thanks for the whole MDN ressource!

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/security/attacks/xss`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/XSS
[...]

PR #40786, opened 2025-08-18
docs(xss): clarify attribute-context examples and fix contradiction

badwriter123:
### Description (#40780)

Hello,
For the file --> files/en-us/web/security/attacks/xss/index.md , I fixed the contradiction in the “Document contexts” section.  I have updated the example to use a working payload (onmouseover=alert(1) or backticks). I also corrected the unquoted class example to demonstrate real attribute injection and specified that quoting the placeholder prevents it. No other sections were changed.

### Motivation
I wanted to remove the contradiction and make the examples accurate and unambiguous.

### DIFF
--- a/files/en-us/web/security/attacks/xss/index.md
+++ b/files/en-us/web/security/attacks/xss/index.md
@@ -203,7 +203,7 @@ However, suppose the template is like this:
 <div \{{ my_input }}></div>
 ```
 
-In this context the browser will treat the `my_input` variable as an HTML attribute. If `my_input` is `onmouseover="alert('XSS')"`, the output encoding provided by Django won't prevent the attack.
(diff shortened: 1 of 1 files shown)


files/en-us/web/security/attacks/xss/index.md
```diff
@@ -203,7 +203,7 @@ However, suppose the template is like this:
 <div \{{ my_input }}></div>
 ```
 
-In this context the browser will treat the `my_input` variable as an HTML attribute. If `my_input` is `onmouseover="alert('XSS')"`, the output encoding provided by Django won't prevent the attack.
+In this context the browser will treat the `my_input` variable as an HTML attribute. Because Django encodes quotes (`"` → `&quot;`, `'` → `&#x27;`), the payload `onmouseover="alert('XSS')"` will not execute. However, an unquoted payload like `onmouseover=alert(1)` (or using backticks, `onmouseover=alert(`XSS`)`) will still execute, because attribute values need not be quoted and backticks are not escaped by default.
```
hamishwillee: Use double backticks to properly render the inner backticks.
```suggestion
In this context the browser will treat the `my_input` variable as an HTML attribute. Because Django encodes quotes (`"` → `&quot;`, `'` → `&#x27;`), the payload `onmouseover="alert('XSS')"` will not execute.
However, an unquoted payload like `onmouseover=alert(1)` (or using backticks, ``onmouseover=alert(`XSS`)``) will still execute, because attribute values need not be quoted and backticks are not escaped by default.
```

hamishwillee (approved):
Thank you. I think this is better.

PR #40788, opened 2025-08-19
`Highlight#priority` explain that higher=greater priority

lionel-rowe:
### Description

* Add explanation that higher=greater priority in `Highlight#priority` article
* Also slightly re-ordered the intro to front-load the definition of what the property actually does (rather than front-loading the problem it solves, burying the definition several paragraphs deep)

### Motivation

Current article doesn't mention whether e.g. priority=1 has more or less preference than priority=2.

### Additional details

https://drafts.csswg.org/css-highlight-api/#priorities

> A higher [priority](https://drafts.csswg.org/css-highlight-api/#priority) results in being above in the stacking order.

### DIFF
--- a/files/en-us/web/api/highlight/priority/index.md
+++ b/files/en-us/web/api/highlight/priority/index.md
@@ -8,7 +8,7 @@ browser-compat: api.Highlight.priority
 
 {{APIRef("CSS Custom Highlight API")}}
 
-The `priority` property of the {{domxref("Highlight")}} interface is a {{jsxref("Number")}} used to determine which highlight's styles should be used to resolve style conflicts in overlapping parts. Highlights with a higher `priority` number have preference over those with a lower `priority`.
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/highlight/priority/index.md
```diff
@@ -8,15 +8,15 @@ browser-compat: api.Highlight.priority
 
 {{APIRef("CSS Custom Highlight API")}}
 
+The `priority` property of the {{domxref("Highlight")}} interface is a {{jsxref("Number")}} used to determine which highlight's styles should be used to resolve style conflicts in overlapping parts. Highlights with a higher `priority` number have preference over those with a lower `priority`.
```
wbamberg: (I know you didn't introduce this, but we don't treat basic types as JS objects.)

```suggestion
The `priority` property of the {{domxref("Highlight")}} interface is a number used to determine which highlight's styles should be used to resolve style conflicts in overlapping parts. Highlights with a higher `priority` number have preference over those with a lower `priority`.
```

wbamberg (approved):
👍 thank you, this is much better.

PR #40789, opened 2025-08-19
Some more `WebSocketStream` fixes

tomayac:
Some more fixes as a follow-up from https://github.com/mdn/content/pull/40782.  

### Description

- Cited some text from the design doc
- Clarified handshake

### Motivation

Fix some incorrect statements.

### Additional details

https://docs.google.com/document/d/1La1ehXw76HP6n1uUeks-WJGFgAnpX2tCjKts7QFJ57Y/edit?pli=1&tab=t.0


files/en-us/web/api/websocketstream/websocketstream/index.md
```diff
@@ -29,7 +29,8 @@ new WebSocketStream(url, options)
...
       - : A single string or an array of strings representing the sub-protocol(s) that the client would like to use, for example `"amqp"` or `"mqtt"`. Subprotocols may be selected from the [IANA WebSocket Subprotocol Name Registry](https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name) or may be custom names jointly understood by the client and the server. A single server can implement multiple WebSocket sub-protocols, and handle different types of interactions depending on the specified value. If it is omitted, an empty array is used by default. If `protocols` is included, the connection will only be established if the server reports that it has selected one of these sub-protocols.
     - `signal` {{optional_inline}}
-      - : An {{domxref("AbortSignal")}} belonging to an {{domxref("AbortController")}} that you want to use to close the WebSocket connection.
+      - : An {{domxref("AbortSignal")}} belonging to an {{domxref("AbortController")}}, which can be used to abort the *handshake*. This is specifically intended to make it easy to implement connection timeouts. As such, it does nothing after the connection is established.
```
tomayac:
```suggestion
      - : An {{domxref("AbortSignal")}} belonging to an {{domxref("AbortController")}}, which can be used to abort the connection *before the handshake*. This is specifically intended to make it easy to implement connection timeouts. As such, it does nothing after the connection is established.
```

files/en-us/web/api/websocketstream/websocketstream/index.md
```diff
@@ -29,7 +29,8 @@ new WebSocketStream(url, options)
...
       - : A single string or an array of strings representing the sub-protocol(s) that the client would like to use, for example `"amqp"` or `"mqtt"`. Subprotocols may be selected from the [IANA WebSocket Subprotocol Name Registry](https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name) or may be custom names jointly understood by the client and the server. A single server can implement multiple WebSocket sub-protocols, and handle different types of interactions depending on the specified value. If it is omitted, an empty array is used by default. If `protocols` is included, the connection will only be established if the server reports that it has selected one of these sub-protocols.
     - `signal` {{optional_inline}}
-      - : An {{domxref("AbortSignal")}} belonging to an {{domxref("AbortController")}} that you want to use to close the WebSocket connection.
+      - : An {{domxref("AbortSignal")}} belonging to an {{domxref("AbortController")}}, which can be used to abort the connection *before the handshake*. This is specifically intended to make it easy to implement connection timeouts. As such, it does nothing after the connection is established.
```
wbamberg: Does your "abort the connection before the handshake" mean, before the handshake has started, or before it has finished (i.e. before the client gets `opened`)? My suggestion below was made on the basis that it's the latter, but maybe this is wrong?

Some other changes here:
- don't say "belonging to an AbortController" since AFAICS this is not true if it's a timeout
- link to the docs on handshake and connect it to opened
- specifically->primarily ("primarily" implies that there might be other uses, if there aren't, we should just delete "specifically" because it adds nothing)
- don't say things are easy

```suggestion
      - : An {{domxref("AbortSignal")}}, which can be used to abort the connection before the [handshake](/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#the_websocket_handshake) has completed (that is, before the {{domxref("WebSocketStream.opened", "opened")}} promise resolves). This is primarily intended to help implement connection timeouts. As such, it does nothing after the connection is established.
```

files/en-us/web/api/websocketstream/websocketstream/index.md
```diff
@@ -47,26 +48,28 @@ const wss = new WebSocketStream("wss://example.com/wss");
 A more advanced example could also include an options object containing custom protocols and/or an {{domxref("AbortSignal")}}:
```
wbamberg:
```suggestion
A more advanced example could also include an options object containing custom protocols and/or an {{domxref("AbortSignal")}}. The following example will time out if the connection is not established within 5 seconds:
```

files/en-us/web/api/websocketstream/websocketstream/index.md
```diff
@@ -47,26 +48,28 @@ const wss = new WebSocketStream("wss://example.com/wss");
...
-const controller = new AbortController();
 const queueWSS = new WebSocketStream("wss://example.com/queue", {
   protocols: ["amqp", "mqtt"],
-  signal: controller.signal,
+  signal: AbortSignal.timeout(5000),
 });
 ```
 
-At a later time, {{domxref("AbortController.abort()")}} can be called when required to close the connection:
+and the connect will timeout after 5 seconds.
```
wbamberg:
```suggestion
```

files/en-us/web/api/websocketstream/websocketstream/index.md
```diff
@@ -47,26 +48,28 @@ const wss = new WebSocketStream("wss://example.com/wss");
...
 
-```js
-controller.abort();
-```
+If you're connecting to localhost, it's likely to succeed or fail almost instantly, so this will have no effect.
 
-Alternatively, you can use the {{domxref("WebSocketStream.close()")}} method to close a connection, however this is mainly needed if you wish to specify a custom code and/or reason for the server to report.
+You can use the {{domxref("WebSocketStream.close()")}} method to close a connection.
+
+`WritableStream.close()` and `WritableStreamDefaultWriter.close()` from the `writable` of the fulfilled `opened` `Promise` also closes the connection.
```
wbamberg: I'd lean towards omitting this, as it looks a bit random after this update.

```suggestion
```

I suppose the argument for keeping it is to forestall confusion about using signal to close the connection. If we wanted to do that, we ought to frame it properly: "Note that you can't use `signal` to close a connection once it has been established. To close an established connection...". And we ought to put that explanation in the main body of the page (perhaps in the `signal` `<dd>`) and not in the examples (examples should illustrate, not explain new things).

files/en-us/web/api/websocketstream/websocketstream/index.md
```diff
@@ -47,26 +48,28 @@ const wss = new WebSocketStream("wss://example.com/wss");
...
 
 See [Using WebSocketStream to write a client](/en-US/docs/Web/API/WebSockets_API/Using_WebSocketStream) for a complete example with full explanation.
```
wbamberg, line 63: Hmm it looks like https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Using_WebSocketStream#closing_the_connection also needs updating.

wbamberg (changes requested):
Thanks for the PR, @tomayac !

files/en-us/web/api/websocketstream/websocketstream/index.md
tomayac: I think it makes sense to keep it, and to add something along the lines of what you wrote. Can you help author that?

files/en-us/web/api/websocketstream/websocketstream/index.md
tomayac: Opened https://github.com/mdn/content/pull/40807 for this.

files/en-us/web/api/websocketstream/websocketstream/index.md
wbamberg: It's going to be much easier to do that as a follow up then, so I'll merge this as it is.

wbamberg (approved):
Thank you for the fixes, @tomayac !

PR #40792, opened 2025-08-19
Clarify example in Callback function

alexcastrodev:
### Description

This PR improves the **Callback function glossary page** by clarifying the examples that illustrate synchronous vs asynchronous callbacks.  
- Updated the synchronous example to show why `console.log(value)` prints `2` (because `value = 2` has already run).  
- Simplified explanatory text for beginners while keeping it technically accurate.  

### Motivation

The previous example was incomplete and sometimes confusing for readers.  
I’ve received feedback that when sharing the page, the example “didn’t work” for people who had not yet read other parts of the documentation, because the context was missing.  
These changes make the examples self-contained and explicit, so they work for first-time readers as well as those familiar with other MDN docs.

### DIFF
--- a/files/en-us/glossary/callback_function/index.md
+++ b/files/en-us/glossary/callback_function/index.md
@@ -25,7 +25,7 @@ console.log(value); // 1 or 2?
 
 If `doSomething` calls the callback synchronously, then the last statement would log `2` because `value = 2` is synchronously executed; otherwise, if the callback is asynchronous, the last statement would log `1` because `value = 2` is only executed after the `console.log` statement.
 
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
Sure, thanks

files/en-us/glossary/callback_function/index.md
```diff
@@ -14,37 +14,31 @@ There are two ways in which the callback may be called: _synchronous_ and _async
 Understanding whether the callback is synchronously or asynchronously called is particularly important when analyzing side effects. Consider the following example:
 
 ```js
-function doSomething(callback) {
-  callback();
-}
-
```
alexcastrodev: @Josh-Cena sorry, but, why has it been removed?

Josh-Cena: Because it was moved after?

alexcastrodev: The main problem was that if a beginner ran this code, it wouldn’t work because the doSomething function isn’t included in the first example.

Josh-Cena: The only section on MDN written for "beginners" is the learning area. Everything else assumes the reader has perfect sense of what runs and isn't supposed to run. These example implementations of `doSomething()` below are enough courtesy for our readers.

alexcastrodev: Makes sense. 
Thank you :)

PR #40795, opened 2025-08-19
Remove redundant indefinite articles from var/param/etc names

### LINKED ISSUE #40794: Weird param names like `aCommandName`, `aShowDefaultUI`, `aValueArgument`
### MDN URL

E.g. https://developer.mozilla.org/docs/Web/API/Document/execCommand

Parameters

Weird param names like `aCommandName`, `aShowDefaultUI`, `aValueArgument`, instead of the more conventional `commandName`, `showDefaultUi`, `valueArgument`. Not clear what `a` refers to, as it doesn't even make sense as an indefinite article for `aShowDefaultUI`...?

Less `a`s

[Guidelines for writing JavaScript code examples § Variable names](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide/JavaScript#variable_names):

> Avoid using articles and possessives. For example, use `car` instead of `myCar` or `aCar`.

Grepping `\ba[A-Z]\w{2,30}(?![+/=])\b` (case-sensitive) yields 211 results in 49 files, most of which also appear to be occurrences of this problem.

lionel-rowe:
### Description

Remove redundant indefinite articles `a` from var/param/etc names (e.g. `aCar` -> `car`).

Mostly this is just grepping `\ba([A-Z])(\w{2,30}(?![+/=])\b)` and replacing with `\l$1$2`, with a few context-sensitive exceptions.

Excluded cases where the `a` had a clear purpose (e.g. `aElem` = an anchor element), as well as files under `mozilla/firefox/releases` (release notes) and `webgl*,webxr*` (I don't know if webgl has some convention around this or if the `a`s there had some significance).

### Motivation

Per linked issue

### Additional details

Bikeshedding — does MDN have a general style preference between `showDefaultUI` vs `showDefaultUi` style for camel-cased vars? The [writing guide](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide/JavaScript#variable_names) currently doesn't seem to mention it.

Fixes https://github.com/mdn/content/issues/40794

### DIFF
--- a/files/en-us/learn_web_development/extensions/server-side/express_nodejs/introduction/index.md
+++ b/files/en-us/learn_web_development/extensions/server-side/express_nodejs/introduction/index.md
@@ -366,19 +366,19 @@ const express = require("express");
 const app = express();
 
 // An example middleware function
-function aMiddlewareFunction(req, res, next) {
+function middlewareFunction(req, res, next) {
   // Perform some operations
   next(); // Call next() so Express will call the next middleware function in the chain.
 }
 
 // Function added with use() for all routes and verbs
-app.use(aMiddlewareFunction);
+app.use(middlewareFunction);
 
 // Function added with use() for a specific route
-app.use("/some-route", aMiddlewareFunction);
+app.use("/some-route", middlewareFunction);
 
 // A middleware function added for a specific HTTP verb and route
-app.get("/", aMiddlewareFunction);
+app.get("/", middlewareFunction);
 
 app.listen(3000);
 ```
--- a/files/en-us/learn_web_development/extensions/server-side/express_nodejs/mongoose/index.md
+++ b/files/en-us/learn_web_development/extensions/server-side/express_nodejs/mongoose/index.md
@@ -127,7 +127,7 @@ The code in the asynchronous function then executes until either another `await`
 
 You can see how this works in the example below.
 `myFunction()` is an asynchronous function that is called within a [`try...catch`](/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) block.
-When `myFunction()` is run, code execution is paused at `methodThatReturnsPromise()` until the promise resolves, at which point the code continues to `aFunctionThatReturnsPromise()` and waits again.
+When `myFunction()` is run, code execution is paused at `methodThatReturnsPromise()` until the promise resolves, at which point the code continues to `functionThatReturnsPromise()` and waits again.
 The code in the `catch` block runs if an error is thrown in the asynchronous function, and this will happen if the promise returned by either of the methods is rejected.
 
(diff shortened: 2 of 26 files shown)


Josh-Cena:
> Bikeshedding — does MDN have a general style preference between showDefaultUI vs showDefaultUi style for camel-cased vars?

The web platform overwhelmingly uses the former and I prefer that as well; capitalization is part of the correct spelling and must be respected (for example "npm" must be that way, invariant).


Also FYI I believe that `aParamName` is actually Hungarian naming where `a` means "argument" not an article. Definitely should change though

lionel-rowe:
> The web platform overwhelmingly uses the former and I prefer that as well; capitalization is part of the correct spelling and must be respected (for example "npm" must be that way, invariant).

I'll stick with the web platform-like version then. But I think requiring things like `npm` and `URL` _as single words_ to retain their casing is taking it too far, as it conflicts with the useful convention of `Class, instance, CONST`. And spelling `fetchNpmDependencies` as `fetchnpmDependencies` would just make it less readable. Also, the web platform has examples like `Response#url` (not `Response#URL`).

Roughly, I think the web platform "rule" for camel case is that the first word is always fully lowercased, then subsequent words have their leading char capitalized, retaining whatever trailing case they already had. And the second rule is no-one talks about `XMLHttpRequest`.

> Also FYI I believe that aParamName is actually Hungarian naming where a means "argument" not an article. Definitely should change though

I've heard various more-or-less useful versions of Hungarian naming, but never one that differentiates args from other variables. Also that might explain the motivating case, but lots of the other cases aren't even args (maybe the _other_ ones are indefinite articles?)

Josh-Cena:
Re: `Response.url`. My personal rule is this: `URL` can be all-lowercased as `url`; `npm` can be all-uppercased as `NPM`, but never mixed-case. And yeah no one talks about XHR :-) (I would have gone with `XMLHTTPRequest`). I think we actually had this discussion before when you were updating the base64 glossary entry, where I let `Url` fly.

Re: `a` prefix, yes a lot of these are still articles, and I wrote a few of these, well knowing that our style guide says otherwise. The idea is that `Promise.resolve(aThenable)` is actually supposed to be read like generic prose "Promise resolve a thenable" where `aThenable` is a stand-in for any variable name, not this particular thenable, but now you changed it, I think it reads just as fine, so I think we can move forward like this.

rebloor (commented):
No issues with the change in the one with extensions file.

Josh-Cena:
I will give it the approval&merge this evening unless someone gets ahead of me

Josh-Cena (approved):
Sure, thank you!

PR #40808, opened 2025-08-20
meta: Close comment in PR template

tomayac:
<!-- 🙌 Thanks for contributing to MDN Web Docs. 🙌 →

This led to empty comments if you didn't notice the problem.

### Description

Everything became a comment since it was never closed.

### Motivation

Avoid confusion.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,8 +1,10 @@
 <!-- 🙌 Thanks for contributing to MDN Web Docs. 🙌 -->
 
+<!--
 Add details below to help us review your pull request (PR).
 Explain your changes and link to a related issue or pull request.
-Your PR may be delayed or closed if you don't provide enough information. -->
+Your PR may be delayed or closed if you don't provide enough information.
+-->
 
 ### Description
 
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,4 +1,4 @@
-<!-- 🙌 Thanks for contributing to MDN Web Docs. 🙌
+<!-- 🙌 Thanks for contributing to MDN Web Docs. 🙌 -->
 
 Add details below to help us review your pull request (PR).
 Explain your changes and link to a related issue or pull request.


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,4 +1,4 @@
-<!-- 🙌 Thanks for contributing to MDN Web Docs. 🙌
+<!-- 🙌 Thanks for contributing to MDN Web Docs. 🙌 -->
 
 Add details below to help us review your pull request (PR).
 Explain your changes and link to a related issue or pull request.
```
bsmth, line 5: It's closed on L5 though, you can check here where it's un-modified: https://redirect.github.com/mdn/content/pull/40760

Would it be clearer if each line had open/close comment tags?

tomayac: > Would it be clearer if each line had open/close comment tags?

100%. And as is, it results in an empty overall GitHub comment.

Josh-Cena: Sorry I don't understand. What empty comment? If you post this template as-is you still see the headings because they are not inside any MD comment.

Also strong -1 to changing anything; adding comment markers to every line makes the explanation even harder to read.

tomayac: This is what my brain did:

"Ah, comment on line 1, and placeholder explanation text on the following lines, let's get rid of it."

Screenshot 2025-08-20 at 13 38 33

After that, you end up with the unclosed comment.

Josh-Cena: How about we add line breaks around the comment markers so it's more salient that they are multiline?

```md
<!--
Blah blah

Blah blah
-->

tomayac: I have updated the PR to do that in https://github.com/mdn/content/pull/40808/commits/9846c42a496c42ee737c2c1e89d22f310cb0f13c.

bsmth (approved):
LGTM, thanks! @Josh-Cena any other comments?

Josh-Cena:
I'm good with this

PR #40811, opened 2025-08-20
fix: typo in learning outcomes

FredTheJanitor:
### Description

There is a typo, so I corrected it.

### Motivation

I am making this change because typos can be confusing, and it will help readers by them being able to read the correct word.


chrisdavidmills (approved):
Good catch. Thanks for the fix, @FredTheJanitor!

PR #40809, opened 2025-08-20
relnote(Fx142): & inside at-scope blocks implicit specificity

bsmth:
### Description

Using `&` inside `@scope` changes according to spec updates

**Explainer:** https://css.oddbird.net/scope/parent-selector/

### Bugs

- [Change `&` behaviour inside `@scope` block as per spec resolution](https://bugzilla.mozilla.org/show_bug.cgi?id=1975531)

### Additional details

- [ ] Parent issue https://github.com/mdn/content/issues/40481

PR #40814, opened 2025-08-20
docs(Fx142): css ampersand in at-scope content and `:scope`

bsmth:
### Description

Using `&` inside `@scope` changes according to spec updates (and relation to `:scope`).

This makes `& &` redundant as far as I can see. The explainer has good context and I hope I've conveyed that in these changes.

**Explainer:** https://css.oddbird.net/scope/parent-selector/

### Bugs

- [Change `&` behaviour inside `@scope` block as per spec resolution](https://bugzilla.mozilla.org/show_bug.cgi?id=1975531)

### Additional details

- [ ] Parent issue https://github.com/mdn/content/issues/40481
- [x] relnote https://github.com/mdn/content/pull/40809

### DIFF
--- a/files/en-us/web/css/@scope/index.md
+++ b/files/en-us/web/css/@scope/index.md
@@ -199,8 +199,7 @@ Because [`:where()`](/en-US/docs/Web/CSS/:where) has zero specificity, the only 
 }
 ```
 
-By contrast, using `:scope` explicitly selects the scope root with class specificity.
-Like other pseudo-classes, `:scope` has a specificity of `0-1-0`.
+By contrast, using `:scope` explicitly selects the scope root and adds class-level specificity (`0-1-0`), since `:scope` is a pseudo-class.
 In the following example, `:scope img` has a specificity of `0-1-1`:
 
 ```css
--- a/files/en-us/web/css/@scope/index.md
+++ b/files/en-us/web/css/@scope/index.md
@@ -115,9 +115,9 @@ Or you could include your `@scope` block inline inside a `<style>` element, whic
 > [!NOTE]
 > It is important to understand that, while `@scope` allows you to isolate the application of selectors to specific DOM subtrees, it does not completely isolate the applied styles to within those subtrees. This is most noticeable with inheritance — properties that are inherited by children (for example {{cssxref("color")}} or {{cssxref("font-family")}}) will still be inherited, beyond any set scope limit.
 
-### The `:scope` pseudo-class
+### `:scope` pseudo-class within `@scope` blocks
 
-In the context of a `@scope` block, the {{cssxref(":scope")}} pseudo-class represents the scope root — it provides an easy way to apply styles to the scope root itself, from inside the scope:
+In the context of an `@scope` block, the {{cssxref(":scope")}} pseudo-class provides a convenient way to directly apply styles to the scope root, like so:
 
 ```css
 @scope (.feature) {
@@ -129,25 +129,9 @@ In the context of a `@scope` block, the {{cssxref(":scope")}} pseudo-class repre
 }
 ```
 
-In the following code snippet, the three rules select the same element, but differ in specificity (see [Specificity in @scope](#specificity_in_scope) for details):
+Here's some considerations for `:scope` within `@scope` blocks:
 
-```css
-@scope (.feature) {
-  img {
-    /* … */
-  }
-
-  & img {
-    /* … */
-  }
-
-  :scope img {
-    /* … */
-  }
-}
-```
-
-### Notes on scoped selector usage
+- `:scope` adds class-level specificity (see [Specificity in @scope](#specificity_in_scope) for details).
 
 - A scope limit can use `:scope` to specify a specific relationship requirement between the scope limit and root. For example:
 
@@ -182,8 +166,11 @@ In the following code snippet, the three rules select the same element, but diff
 
 ### Specificity in `@scope`
 
-Inside an `@scope` rule, both bare selectors and `&` are treated as if `:where(:scope)` were prepended, meaning they select the scope root with zero specificity.
-Because [`:where()`](/en-US/docs/Web/CSS/:where) has zero specificity, the only specificity comes from `img` (`0-0-1`).
+Inside an `@scope` rule, both bare selectors and `&` behave as if `:where(:scope)` were prepended to the selector.
+Because [`:where()`](/en-US/docs/Web/CSS/:where) has zero specificity, bare selectors and `&` add zero weight and only the specificity of the rest of the selector counts.
+An `& img` selector is the equivalent to writing `:where(:scope) img`.
+
+In both cases in the following example, the only specificity comes from `img` (`0-0-1`):
 
 ```css
 @scope (.article-body) {
--- a/files/en-us/web/css/@scope/index.md
+++ b/files/en-us/web/css/@scope/index.md
@@ -193,9 +193,9 @@ Including a ruleset inside a `@scope` block does not affect the specificity of i
 }
 ```
 
-When using the `&` selector inside a `@scope` block, `&` represents the implicit context (calculated as `:where(:scope)`) which has zero specificity.
-In the following example, `& img` is essentially equivalent to `:where(:scope) img`.
-Because `:where()` always has zero specificity, the only specificity comes from `img` (i.e., 0-0-1):
+The `&` nesting selector, when prepended to a selector inside an `@scope` block, represents the implicit scope context (`:where(:scope)`), which has zero specificity.
+In the following example, `& img` is equivalent to writing `:where(:scope) img`.
+Because [`:where()`](/en-US/docs/Web/CSS/:where) has zero specificity, the only specificity comes from `img` (`0-0-1`).
 
 ```css
 @scope (figure, #primary) {
--- a/files/en-us/web/css/@scope/index.md
+++ b/files/en-us/web/css/@scope/index.md
@@ -206,7 +206,7 @@ Because `:where()` always has zero specificity, the only specificity comes from 
 }
 ```
 
-However, if you decide to explicitly prepend the `:scope` pseudo-class to your scoped selectors, you'll need to factor it in when calculating their specificity. `:scope`, like all regular pseudo-classes, has a specificity of 0-1-0. For example:
+If you explicitly prepend the `:scope` pseudo-class to the selectors inside the `@scope` block, you need to factor in its specificity. Like other pseudo-classes, `:scope` has a specificity of `0-1-0`. In the following example, `:scope img` has a specificity of `0-1-1`:
 
 ```css
 @scope (.article-body) {
--- a/files/en-us/web/css/@scope/index.md
+++ b/files/en-us/web/css/@scope/index.md
@@ -137,11 +137,11 @@ In the following code snippet, the three rules select the same element, but diff
     /* … */
   }
 
-  :scope img {
+  & img {
     /* … */
   }
 
-  & img {
+  :scope img {
     /* … */
   }
 }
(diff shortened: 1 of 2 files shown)


dipikabh:
I can take up the review for this

files/en-us/web/css/@scope/index.md
```diff
@@ -129,9 +129,8 @@ In the context of a `@scope` block, the {{cssxref(":scope")}} pseudo-class repre
 }
 ```
 
-In fact, `:scope` is implicitly prepended to all scoped style rules. If you want, you can explicitly prepend `:scope` or prepend the [nesting](/en-US/docs/Web/CSS/CSS_nesting) selector (`&`) to get the same effect if you find these representations easier to understand.
-
-The three rules in the following block are all equivalent in what they select:
+If you want, you can explicitly prepend `:scope` or prepend the [nesting](/en-US/docs/Web/CSS/CSS_nesting) selector (`&`) to get the same effect if you find these representations easier to understand.
+The three rules in the following block are equivalent in what they select, but note that they differ in specificity, see [Specificity in @scope](#specificity_in_scope) for details:
```
dipikabh: imo, this should stay as a separate, new para
```suggestion
In the following code snippet, the three rules select the same element, but differ in specificity (see [Specificity in @scope](#specificity_in_scope) for details):
```

files/en-us/web/css/@scope/index.md
```diff
@@ -129,9 +129,8 @@ In the context of a `@scope` block, the {{cssxref(":scope")}} pseudo-class repre
 }
 ```
 
-In fact, `:scope` is implicitly prepended to all scoped style rules. If you want, you can explicitly prepend `:scope` or prepend the [nesting](/en-US/docs/Web/CSS/CSS_nesting) selector (`&`) to get the same effect if you find these representations easier to understand.
-
-The three rules in the following block are all equivalent in what they select:
+If you want, you can explicitly prepend `:scope` or prepend the [nesting](/en-US/docs/Web/CSS/CSS_nesting) selector (`&`) to get the same effect if you find these representations easier to understand.
```
dipikabh: - The previous code block already shows `:scope` "explicitly". Should we say what would happen if we omit it and how that would have a different (lower) specificity. Maybe we can introduce here the fact that "implicit scoping" behaves like `:where(:scope)`.
- Also, let's drop the mention of `&` from this instance - because if we do, we'd need to mention that they have different specificity
- In fact, maybe we can move the three rule example that comes after this to the "Specificity in @scope" section

files/en-us/web/css/@scope/index.md
```diff
@@ -195,42 +194,43 @@ Including a ruleset inside a `@scope` block does not affect the specificity of i
 }
 ```
```
dipikabh: we can add more clarity to L186 (and also include the explanation to prose in addition to it being a code comment, maybe even drop "as expected" from the comment):

"Including a rule inside an `@scope` block does not affect the specificity of its selector. The scope root and scope limit only restrict _where_ the selector can match. For example, the specificity of `img` in the following code is `0-0-1`, as is expected, and this is not affected by the scope root `.article-body`."

files/en-us/web/css/@scope/index.md
```diff
@@ -195,42 +194,43 @@ Including a ruleset inside a `@scope` block does not affect the specificity of i
...
 
-However, if you decide to explicitly prepend the `:scope` pseudo-class to your scoped selectors, you'll need to factor it in when calculating their specificity. `:scope`, like all regular pseudo-classes, has a specificity of 0-1-0. For example:
+When using the `&` selector inside a `@scope` block, `&` represents the implicit context (calculated as `:where(:scope)`) which has zero specificity.
+In the following example, `& img` is essentially equivalent to `:where(:scope) img`.
+Because `:where()` always has zero specificity, the only specificity comes from `img` (i.e., 0-0-1):
```
dipikabh: some tiny tweaks - instead of saying "when", we can phrase it to state what its use means

```suggestion
The `&` nesting selector, when prepended to a selector inside an `@scope` block, represents the implicit scope context (`:where(:scope)`), which has zero specificity.
In the following example, `& img` is equivalent to writing `:where(:scope) img`.
Because [`:where()`](/en-US/docs/Web/CSS/:where) has zero specificity, the only specificity comes from `img` (`0-0-1`).
```

files/en-us/web/css/@scope/index.md
```diff
@@ -195,42 +194,43 @@ Including a ruleset inside a `@scope` block does not affect the specificity of i
...
-  :scope img {
+@scope (figure, #primary) {
+  & img {
     /* … */
   }
 }
 ```
 
-When using the `&` selector inside a `@scope` block, `&` represents the scope root selector; it is internally calculated as that selector wrapped inside an {{cssxref(":is", ":is()")}} pseudo-class function. So for example, in:
+However, if you decide to explicitly prepend the `:scope` pseudo-class to your scoped selectors, you'll need to factor it in when calculating their specificity. `:scope`, like all regular pseudo-classes, has a specificity of 0-1-0. For example:
```
dipikabh:
```suggestion
If you explicitly prepend the `:scope` pseudo-class to the selectors inside the `@scope` block, you need to factor in its specificity. Like other pseudo-classes, `:scope` has a specificity of `0-1-0`. In the following example, `:scope img` has a specificity of `0-1-1`:
```

files/en-us/web/css/@scope/index.md
```diff
@@ -195,42 +194,43 @@ Including a ruleset inside a `@scope` block does not affect the specificity of i
...
     /* … */
   }
 
-  /* Doesn't work */
-  :scope :scope {
+  /* Same selection, but adds specificity 0-1-0 */
+  :scope img {
     /* … */
   }
 }
```
dipikabh: - This section feels like a repeat of what we've already mentioned in the previous section. Probably the only additional part is what they select. Maybe merge and reconcile?
- "they select the scope root." => should this be "they select the descendants inside the scope root"or "they select the elements within the scope root"? 
- "both bare selectors and `&`" => "both bare selectors and `&` with a selector"?

dipikabh (approved):
Great, thanks for capturing the changes in the specificity behavior.
I've added a few rephrasing suggestions.

files/en-us/web/css/@scope/index.md
bsmth: Maybe this is better? Let's remove this sentence completely:

```suggestion
```

I'll apply this commit, then let's take a look again

files/en-us/web/css/@scope/index.md
bsmth: Agreed, I've made some changes in 2a4fe65f2286274ac0a88364d4c998d2d793eda8 which should tidy this a little more.

bsmth:
Thanks for the review! @dipikabh do you want to have another look?

files/en-us/web/css/@scope/index.md
```diff
@@ -129,21 +129,19 @@ In the context of a `@scope` block, the {{cssxref(":scope")}} pseudo-class repre
...
 ```
```
dipikabh, line 131: Some other cleanup suggestions:

- L118:

```diff
-- ### The `:scope` pseudo-class
++ ### `:scope` pseudo-class within `@scope` block
```

- L120:

Not sure what we mean by "from inside the scope". Is this closer:
```diff
-- In the context of a `@scope` block, the {{cssxref(":scope")}} pseudo-class represents the scope root — it provides an easy way to apply styles to the scope root itself, from inside the scope:
++ In the context of an `@scope` block, the {{cssxref(":scope")}} pseudo-class represents the scope root — it provides an easy way to directly apply styles to the scope root, like so:

File: files/en-us/web/css/@scope/index.md [hunk 8]
```diff
@@ -129,21 +129,19 @@ In the context of a `@scope` block, the {{cssxref(":scope")}} pseudo-class repre
...
 
-In fact, `:scope` is implicitly prepended to all scoped style rules. If you want, you can explicitly prepend `:scope` or prepend the [nesting](/en-US/docs/Web/CSS/CSS_nesting) selector (`&`) to get the same effect if you find these representations easier to understand.
-
-The three rules in the following block are all equivalent in what they select:
+In the following code snippet, the three rules select the same element, but differ in specificity (see [Specificity in @scope](#specificity_in_scope) for details):
```
dipikabh: - This and the accompanying code feels out of place in the "`:scope` pseudo-class within `@scope` block" section. I think it might fit better in "Specificity in @scope" section.
- As I see it, we could remove the heading "### Notes on scoped selector usage" (L150) and move all those bullet points in the "`:scope` pseudo-class within `@scope` block" section as they all pertain to `:scope`.

files/en-us/web/css/@scope/index.md
```diff
@@ -184,53 +182,31 @@ The three rules in the following block are all equivalent in what they select:
 
 ### Specificity in `@scope`
 
-Including a ruleset inside a `@scope` block does not affect the specificity of its selector, regardless of the selectors used inside the scope root and limit. For example:
+Inside an `@scope` rule, both bare selectors and `&` are treated as if `:where(:scope)` were prepended, meaning they select the scope root with zero specificity.
+Because [`:where()`](/en-US/docs/Web/CSS/:where) has zero specificity, the only specificity comes from `img` (`0-0-1`).
```
dipikabh: something like this...to chain the first sentence with `:where()`'s zero specificity:
```suggestion
Inside an `@scope` rule, both bare selectors and leading `&` behave as if `:where(:scope)` were prepended to the selector. Because [`:where()`](/en-US/docs/Web/CSS/:where) has zero specificity, bare selectors and `&` add zero weight and only the specificity of the rest of the selector counts.

In both cases in the following example, the only specificity comes from `img` (`0-0-1`).
```

- Might be important to say "leading `&`" or "`&` at the start of a selector" - to differentiate it from the bare `&`.
WDYT?
IIUC `&` by itself will target the scope root. (in contrast, `& img` will target elements within the scope)

- We could also add to second para: "`& img` is equivalent to writing `:where(:scope) img`."

files/en-us/web/css/@scope/index.md
```diff
@@ -184,53 +182,31 @@ The three rules in the following block are all equivalent in what they select:
...
 ```
 
-`& img` is equivalent to `:is(figure, #primary) img`. Since `:is()` takes the specificity of its most specific argument (`#primary`, in this case), the specificity of the scoped `& img` selector is therefore 1-0-0 + 0-0-1 = 1-0-1.
-
-### The difference between `:scope` and `&` inside `@scope`
-
-`:scope` represents the matched scope root, whereas `&` represents the selector used to match the scope root. Because of this, it is possible to chain `&` multiple times. However, you can only use `:scope` once — you can't match a scope root inside a scope root.
+By contrast, using `:scope` explicitly selects the scope root with class specificity.
+Like other pseudo-classes, `:scope` has a specificity of `0-1-0`.
+In the following example, `:scope img` has a specificity of `0-1-1`:
```
dipikabh: could probably tie up the first two sentences
```suggestion
By contrast, using `:scope` explicitly selects the scope root and adds class-level specificity (`0-1-0`), since `:scope` is a pseudo-class.
In the following example, `:scope img` has a specificity of `0-1-1`:
```

dipikabh (approved):
Much better after the cleanup, thanks!

files/en-us/web/css/@scope/index.md
bsmth: Thanks! Can you check again 4580ac85defd9b543d54d69e7867c3951efd2bcd

files/en-us/web/css/@scope/index.md
bsmth: Done in 4580ac85defd9b543d54d69e7867c3951efd2bcd

files/en-us/web/css/@scope/index.md
bsmth: changed in 4580ac85defd9b543d54d69e7867c3951efd2bcd

bsmth:
Thanks, Dipika! Do you want to have another look?

files/en-us/web/css/@scope/index.md
dipikabh: thanks!

dipikabh:
Looks really good, thanks a lot @bsmth!
Let's land these updates 🎉

bsmth:
> Looks really good, thanks a lot @bsmth! Let's land these updates 🎉

Great, thanks a lot!

PR #40821, opened 2025-08-20
More fixes for socket stream close documentation

wbamberg:
Follow-up to https://github.com/mdn/content/pull/40782 and https://github.com/mdn/content/pull/40789.

Main point being that you _can't_ use the `signal` option to close an already-established connection, you can only use it to cancel a connection attempt before it has been established.

cc @tomayac .

This PR:
- updates the guide page at https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Using_WebSocketStream to give correct info about usage of `signal` and how to close a connection. 
- adds some subheadings to the examples in https://developer.mozilla.org/en-US/docs/Web/API/WebSocketStream/WebSocketStream#examples, to make the structure clearer, and provides a bit of context for the stuff about closing a connection.

### DIFF
--- a/files/en-us/web/api/websockets_api/using_websocketstream/index.md
+++ b/files/en-us/web/api/websockets_api/using_websocketstream/index.md
@@ -28,13 +28,13 @@ To create a WebSocket client, you first need to create a new `WebSocketStream` i
 const wss = new WebSocketStream("wss://example.com/wss");
 ```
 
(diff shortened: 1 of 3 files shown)


tomayac (commented):
Non-owner LGTM, thank you!

sideshowbarker (approved):

PR #40828, opened 2025-08-21
correction of a code-example property, use `prefix` instead of `namespaceURI`

bhamblok:
### Description

Correction of a code-example property. The property `prefix` should have been used instead of `namespaceURI`. 

### Motivation

I think it was a plain copy/paste mistake


Josh-Cena (approved):
Good catch, thanks

PR #40829, opened 2025-08-21
Update Learn changelog with entry about summer 2025 changes

chrisdavidmills:
### Description

This PR updates the [Learn web development changelog](https://developer.mozilla.org/en-US/docs/Learn_web_development/Changelog) with an entry describing the test your skills quality/consistency updates made in summer 2025.

### DIFF
--- a/files/en-us/learn_web_development/changelog/index.md
+++ b/files/en-us/learn_web_development/changelog/index.md
@@ -5,6 +5,14 @@ page-type: landing-page
 sidebar: learnsidebar
 ---
 
+## August 2025
+
+In the summer of 2025, we completed work on significant changes to the structure and positioning of our "Test your skills" articles and other reader exercises:
+
+- The "Test your skills" article content has been updated so that the language and approach in each one is consistent. For example, all of the old interactive editors have been replaced with live sample exercises that make use of the [MDN Playground](/en-US/play), and most of the repetitive boilerplate has been consolidated into a single [Test your skills help section](/en-US/docs/Learn_web_development#test_your_skills).
+- The position of the "Test your skills" articles in the navigation has been updated. Instead of being available from "Test your skills" submenus and linked to from the end of associated tutorials, the articles now appear in the sidebar and in the previous/next links. They are now mandatory articles in the flow, rather than optional resources.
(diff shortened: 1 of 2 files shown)


Josh-Cena:
@chrisdavidmills I don't know if you prefer other ways to be reached but there's no tracking issue/discussion for this so this may be the best place. I don't know if your goal is to fix "Test your skills" pages only or all "active learning" sections, but in particular I'd like to see two custom editors removed and replaced with MDN playground: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Including_vector_graphics_in_HTML, https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/General_embedding_technologies. Perhaps you may want to restructure the other "active learning" sections too.

chrisdavidmills:
> @chrisdavidmills I don't know if you prefer other ways to be reached but there's no tracking issue/discussion for this so this may be the best place. I don't know if your goal is to fix "Test your skills" pages only or all "active learning" sections, but in particular I'd like to see two custom editors removed and replaced with MDN playground: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Including_vector_graphics_in_HTML, https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/General_embedding_technologies. Perhaps you may want to restructure the other "active learning" sections too.

Slack is always good ;-)

To be honest, I'd forgotten about the ones that are in core modules, but not in the main article flow. I'll do those ones now.

Also, the extensions modules weren't as high a priority, so there might be instances there too. 

I've shared the tracking sheet with you for this work, so you can add other work items to it.

Josh-Cena:
My checker is reporting all inline `style` attributes in code examples and it looks like we either only have these two instances of custom editors or the other modules don't use inline `style`

chrisdavidmills:
> My checker is reporting all inline `style` attributes in code examples and it looks like we either only have these two instances of custom editors or the other modules don't use inline `style`

Yup, I can only see those two as well. Thanks for checking.

bsmth (approved):
LGTM, thanks 👍🏻

PR #40120, opened 2025-06-28
CSS tree counting functions - `sibling-index()` and `sibling-count()`

ramiy:
### Description

CSS value functions lists the available functions supported by different browser. Tree counting functions never added to the list because they were never supported, until now.

### Motivation

Chrome 138 added support for `sibling-index()` and `sibling-count()` functions.

Ref:

* https://developer.chrome.com/release-notes/138#css_sibling-index_and_sibling-count
* https://developer.chrome.com/blog/chrome-138-beta#css_sibling-index_and_sibling-count

### Additional details

Spec: https://drafts.csswg.org/css-values-5/#tree-counting
BCD: https://github.com/mdn/browser-compat-data/pull/27019

PR #40830, opened 2025-08-21
Update example in CSS `sibling-count()` function

### LINKED ISSUE #40755: Bad Sibling-Count Example
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/sibling-count

Dynamic column count

The example as a whole does not mirrow what `sibling-count()` is about.

The sibling-count() returns the number of children 
of a container **on the child it's used on.**

In the example that would mean using it on an `<li>` 
element would return `4` since there are `4` children / siblings.

The example is written as if the `sibling-count` 
had to be used on the parent element. In the example
it return `3` as the parent ( `<ul>` ) has two
sibling `<script>` tags.

Image

_No response_

_No response_

ramiy:
### Description

Fix bad example in CSS `sibling-count()` function.

Fixes #40755

Related #40120


Josh-Cena (changes requested):
You would need to fix the descriptions too, which still refer to this function as a child-counting algorithm.

ramiy:
I've reverted the description to the original description I use in my initial PR (#40120), based on the spec description.

@Josh-Cena tell me if you approve this change.

Josh-Cena (approved):
Thank you for the fix! The spec uses language like "direct children of the parent", which is slightly less straightforward, and your modification still suggests "direct children of the element" which is incorrect, so I'm opting for the more straightforward (but maybe less precise) wording "siblings including itself".

PR #40835, opened 2025-08-21
Editorial review: Document RTCEncodedAudioFrame.getMetadata() audioLevel return property

chrisdavidmills:
### Description

Chrome 139 updates the [`RTCEncodedAudioFrame.getMetadata()`](https://developer.mozilla.org/en-US/docs/Web/API/RTCEncodedAudioFrame/getMetadata) method so that the [`audioLevel`](https://w3c.github.io/webrtc-encoded-transform/#dom-rtcencodedaudioframemetadata-audiolevel) property is available in its return object. See https://chromestatus.com/feature/5206106602995712.

This PR adds content for this new return property.

### DIFF
--- a/files/en-us/web/api/rtcencodedaudioframe/getmetadata/index.md
+++ b/files/en-us/web/api/rtcencodedaudioframe/getmetadata/index.md
@@ -27,7 +27,7 @@ None.
 An object with the following properties:
 
 - `audioLevel`
-  - : A number representing the audio level of this frame. The value is between 0 and 1 inclusive (linear), where 1 represents 0 dBov, 0 represents silence, and 0.5 represents approximately 6 dBSPL change in the sound pressure level from 0 dBov. The value is converted from the -127 to 0 range specified in [RFC6464](https://www.rfc-editor.org/rfc/rfc6464) via the equation `10^(-rfc_level/20)`. If the RFC6464 header extension is not present in the received packets of the frame, `audioLevel` will be `undefined`.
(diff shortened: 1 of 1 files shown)


guidou (commented):
lgtm

files/en-us/web/api/rtcencodedaudioframe/getmetadata/index.md
```diff
@@ -26,6 +26,8 @@ None.
 
 An object with the following properties:
 
+- `audioLevel`
+  - : A number representing the audio level of this frame. The value is between 0 and 1 inclusive (linear), where 1 represents 0 dBov, 0 represents silence, and 0.5 represents approximately 6 dBSPL change in the sound pressure level from 0 dBov. The value is converted from the -127 to 0 range specified in [RFC6464](https://www.rfc-editor.org/rfc/rfc6464) via the equation `10^(-rfc_level/20)`. If the RFC6464 header extension is not present in the received packets of the frame, `audioLevel` will be `undefined`.
```
wbamberg: I had this comment on the stats PR that exposes audioLevel: https://github.com/mdn/content/pull/40495#discussion_r2268252656 and think it applies here too.

Is this the same thing as that one? They have slightly different information associated, should it be the same? Should we find a single place to document this, rather than repeating it?

chrisdavidmills: It is essentially the same piece of information, except that in this case, it is extracted from the metadata about an individual frame, rather than from statistics related to a stream.

I'd certainly be happy to make the language more consistent with what Hamish wrote; I'm not sure if there is a sensible common place we could document this and link to from both.

Would you rather I linked to Hamish's definition?

wbamberg: I think:

1. _ideally_, the WebRTC docs might have a single guide page that precisely defines concepts like audio level, that reference pages like this and @hamishwillee 's could link to. This page doesn't exist AFAIK and it's surely out of scope to ask you to add it in this PR.
2. _minimally_ I think we should consider if we want to explain what things like dBov and dBSPL are, which was my original comment on Hamish's PR.
3. _optionally_ we should make this language more consistent with the language in the other page.

I don't think linking from one reference page to another on an ad hoc basis is particularly useful. Linking to a single canonical definition ought to be part of a systematic organization of this documentation, which gets us into (1) territory.

chrisdavidmills: That sounds reasonable. I've updated the description to improve consistency and add the same explanatory links as Hamish added. I've kept the extra bit of text I included.

wbamberg (approved):
👍 thank you!

PR #40836, opened 2025-08-21
Fix typo in index.md

chphmh:
### Description

Fixes a typo ("justify-items and justify-self **align** items on the inline axis" instead of "... **aline** items ...")

### Motivation

Apparently according to Merriam-Webster dictionary, "aline" can also be used as a less common alternative spelling for "align". But the page uses "align" over a hundred times and "aline" is used once, so I think it's a typo.

### Additional details

["align" on Merriam-Webster](https://www.merriam-webster.com/dictionary/align)


chrisdavidmills (approved):
Hi @chphmh, and thanks for the fix!

PR #40760, opened 2025-08-16
Adds `form` html attribute.

younisdev:
### Description

Adds `form` HTML attribute, providing all the supported elements with links for `form` attribute.
### Motivation

These changes ensure that `form` attribute have its own standalone page and help the reader know about this attribute in a centralized place.
### Additional details

This PR might need `label`, `meter`, and `progress` BCD updates. Plus, these elements `meter` , `object` , `output` might have limited support or no support at all after some research.
### Related issues and pull requests

Fixes #28375

PR #40841, opened 2025-08-22
Enhance HTMLFormControlsCollection description

estelle:
Added details about the elements included in the HTMLFormControlsCollection.

>The elements IDL attribute must return an [HTMLFormControlsCollection](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#htmlformcontrolscollection) rooted at the [form](https://html.spec.whatwg.org/multipage/forms.html#the-form-element) element's [root](https://dom.spec.whatwg.org/#concept-tree-root), whose filter matches [listed elements](https://html.spec.whatwg.org/multipage/forms.html#category-listed) whose [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner) is the [form](https://html.spec.whatwg.org/multipage/forms.html#the-form-element) element, with the exception of [input](https://html.spec.whatwg.org/multipage/input.html#the-input-element) elements whose [type](https://html.spec.whatwg.org/multipage/input.html#attr-input-type) attribute is in the [Image Button](https://html.spec.whatwg.org/multipage/input.html#image-button-state-(type=image)) state, which must, for historical reasons, be excluded from this particular collection.

related to https://github.com/mdn/content/pull/40760


files/en-us/web/api/htmlformcontrolscollection/index.md
```diff
@@ -9,6 +9,8 @@ browser-compat: api.HTMLFormControlsCollection
 
 The **`HTMLFormControlsCollection`** interface represents a _collection_ of HTML _form control elements_, returned by the {{domxref("HTMLFormElement")}} interface's {{domxref("HTMLFormElement.elements", "elements")}} property.
 
+The collection of form control elements includes {{htmlelement("button")}}, {{htmlelement("fieldset")}}, {{htmlelement("input")}}, {{htmlelement("object")}}, {{htmlelement("output")}}, {{htmlelement("select")}}, and {{htmlelement("textarea")}} elements whose form owner is the {{htmlelement("form")}} element.
```
Josh-Cena, line 12: Personally I think this should be documented in `elements` (which we already do). `HTMLFormControlsCollection` is just a collection, and what's put inside this collection is up to each property that returns it, of which there happens to be only one, but this still feels out of place.

estelle: there are form controls that are not sent to the server that are included, so the list is an odd one, which is why i included it. It should be included on this page b/c of that, but perhaps in a different way. Do you have a suggestion? 

For example, 
1. fieldset is not sent to the server, but is in the collection.
2.  If a control is nested in a form, but has a `form` attribute that dissociates it from its ancestor form, it is not included. 
3. If a fieldset is not in the form, but has a `form` attribute that associates it, it is included, but not sent. It's descendant form controls are only included and sent if they too have the `form` attribute.

estelle: @Josh-Cena I am not saying my solution is the right one, but i do think we should inform the reader that the result may not be obvious. Suggestions?

Josh-Cena:
```suggestion
The collection returned by {{domxref("HTMLFormElement.elements")}} includes {{htmlelement("button")}}, {{htmlelement("fieldset")}}, {{htmlelement("input")}}, {{htmlelement("object")}}, {{htmlelement("output")}}, {{htmlelement("select")}}, and {{htmlelement("textarea")}} elements whose form owner is the {{htmlelement("form")}} element. This collection is unrelated to what would be sent when the form is submitted.
```

If the intention is to document the dissociation between what's included and what's submitted, then sure we could keep this paragraph, but make that purpose more salient. Also: (1) we should add this caveat to `HTMLFormElement.elements` as well, because that's the actually relevant documentation page. (2) the `HTMLFormElement.elements` page talks about `<input type="image">` not being included, and form-associated custom elements being included, which should be mentioned here as well. Alternatively we could just replace this list with a link saying "See `HTMLFormElement.elements` for the list of elements included", which I prefer because it makes it clear that this collection itself doesn't control what can be inside it.

estelle:
replaced with https://github.com/mdn/content/pull/41096

PR #40842, opened 2025-08-22
Update HTMLLabelElement 'form' property

estelle:
Clarified the definition of the 'form' property to include all form-associated elements which is different from "all form controls", so should be enumerated.


Josh-Cena (approved):
Thanks. I made it more explicit that "labelable" and "form-associated" refer to existing concepts; also "listed" has nothing to do with this API because "listed" is only relevant for enumeration APIs.

PR #40846, opened 2025-08-22
Remove cool-looking box challenge

chrisdavidmills:
### Description

This PR removes the [Cool-looking box challenge](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Cool-looking_box), as it is not really very useful.


bsmth (approved):
Great, tnx 👍🏻

PR #40851, opened 2025-08-23
avoid destructuring in `fib` example of `Iterator.prototype.take`

imcotton:
### Description

Avoid destructuring in `fib` example of `Iterator.prototype.take`.

### Motivation

This destructuring usage adds an extra layer of syntax noise that is unrelated to this very topic.

### Additional details

(n/a)

(n/a)

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/iterator/take/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/iterator/take/index.md
@@ -37,11 +37,14 @@ The following example creates an iterator that yields terms in the Fibonacci seq
 
 ```js
 function* fibonacci() {
+  let last = 0;
   let current = 1;
   let next = 1;
   while (true) {
     yield current;
-    [current, next] = [next, current + next];
+    last = current;
+    current = next;
+    next += last;
   }
 }
 


files/en-us/web/javascript/reference/global_objects/iterator/take/index.md
```diff
@@ -41,7 +41,8 @@ function* fibonacci() {
   let next = 1;
   while (true) {
     yield current;
-    [current, next] = [next, current + next];
+    current = next;
+    next = current + next;
```
zloirock: 🤦‍♂️

imcotton: :sweat_smile:

Josh-Cena (commented):
Hello! As you realized, the destructuring is needed for the two variables to be assigned simultaneously and not worry about ordering. Removing it makes the code a lot more verbose. Our code examples don't treat destructing as an anti pattern, and we believe that if a reader wants to copy&paste something into their project without thinking too much about it, we'd rather recommend destructuring by default.

imcotton:
For documentation purposes, adding a little verbosity would increase readability and clarity for new users.

Destructing has its place in coding activities, however:

> for the two variables to be assigned simultaneously and not worry about ordering

This is an anti-pattern, as you saw when it tripped me up. It also creates new arrays during destructing.

Could you reopen the PR for further discussion please?

Josh-Cena:
Conciseness is a big factor in readability as well: the more lines you have, the more time people need to spend on understanding what's going on. We won't be dumbing down the syntax for "new learners" because our reference docs are written for experienced coders, not "new learners", and the former would enjoy more modern, shorter, and more semantic code. Again, this is the exact pattern we'd like to recommend to our readers to be copy-pasted into their projects. My mind is made up on this particular matter and I'm not open to bikeshedding, especially when it relates to stylistic decisions only.

imcotton:
Quote from guides on: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide

- you should make sure that code examples are usable, follow generally accepted best practices, and **do not** do anything that will cause an application to be insecure, **inefficient**, bloated, or inaccessible.
- Code examples should be short and should ideally **only** show the feature you are immediately interested in.
- Write your code to be as understandable as possible, **even if it is not the most efficient way to write it**.

I want to emphasize:

> for the two variables to be assigned simultaneously and not worry about ordering (by creating a one time used array that only contains two elements and throw it away in a `while` loop)

It is an anti-pattern.

Josh-Cena:
You claim that it's an anti-pattern but I don't believe so and as I've repeatedly said, as a maintainer, I decide what we recommend and what we don't. I can be convinced about certain decisions but for this one I don't find the arguments to be persuasive. This pattern is synonymous with the swapping trick which we explicitly demonstrate on the destructuring page; you would find similar constructs in any modern language and the performance cost is negligible compared to invoking the iterator protocol.

imcotton:
Are you the sole maintainer on MDN responsible for making this call? If not, would it be possible for someone else to step in, given that you're both the original author of the code example and the one claiming on making the final decision? The overlap could present a potential conflict of interest.

Josh-Cena:
There's no "conflict of interest"; I don't have any vested interest to keep the content as-is. I review 100% of PRs that touch the JS docs and arbitrate what's best for the JS docs, and it's more often than not that I get convinced that someone else's opinion is better. Thank you a lot for your attention to this matter, but this constitutes my final review: the original is better and is the way MDN shall recommend. That means you and others should write code like that as opposed to using four separate lines.

Josh-Cena:
Thinking about it from another perspective, this function is not supposed to be read at all. It's not the focus of the example, and the longer we make it, the more we shift the focus away from the real part that readers should read. "New users" should treat this as a black box (say, imported from a utility library) and not worry about what's inside at all. If there's a one-liner to write Fibonacci's without tanking performance (that is, naïve recursion) I would have gone for it setting all concerns about readability aside, because the more obscure we make it the less readers are inclined to try to understand it. But that's of course taking it to extremes and I think the current one strikes a good balance between readability and terseness.

imcotton:
I strongly disagree.

Destructing is fine, destructing in an infinite while loop and creating a new temporary array each time just to cheat on one variable is not.

However, you set the tone for PR management. I won't spend any more time on this repo from now on.

Josh-Cena:
OK...? You are always welcome to contribute but it's ultimately your choice. There's no expectation or obligation anyway.

But just in case you are hung up on the performance point. Here's Chrome:

image

Here's Firefox:

image

And here's Safari:

image

The margin is within 5%, and it's even much faster on Chrome for some unknown reason. There are worse ways you could tank this function's performance, and I really don't think we should avoid constructing objects or arrays just for performance; on every `next()` call, a "useless" object gets constructed by the generator function anyway.

PR #40852, opened 2025-08-23
Fix formatting of valid duration string microsyntax

mat72coke:
### Description

This PR fixes formatting errors in the **Valid duration string** microsyntaxes on the `<time>` element reference page https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time .

### Motivation

The existing microsyntaxes contained formatting mistakes (extra spaces, inconsistent code markup), which could confuse readers. Correcting these helps ensure developers use valid ISO 8601 durations when working with the `datetime` attribute.

### Additional details

The closing tag for the second `<code>` element in the cell was missing, which caused display issues.

None.


Josh-Cena (approved):
Thanks, great catch

PR #40856, opened 2025-08-24
Remove sentence saying sidebars cannot be opened programmatically

l10nelw:
Untrue/out-of-date since `sidebarAction.open()`, `sidebarAction.close()` and `sidebarAction.toggle()` exist.


Rob--W (approved):
Thanks for the fix!

PR #40867, opened 2025-08-25
clarify default timezone in DateTimeFormat

felix9:
Experimentally, it's easy to see that `DateTimeFormat` will use the local timezone by default, but it's not clear if that's guaranteed or an accident.

It's guaranteed in the specification. This adds a sentence to the docs to indicate that.


Josh-Cena (approved):
Thanks! I just added a mention of the default value to other options too.

PR #40886, opened 2025-08-26
Clarify frame-ancestors behavior in nested frames

harunurhan:
### Description

Added clarification on the behavior of the frame-ancestors directive in nested frames.

### Motivation

While CSP 2 spec is clear about the frame-ancestor behavior when it comes to nested frames, MDN docs don't have any mention of it. This could lead to CSP errors in production if someone sets `frame-ancestor` directive solely relying on MDN.

### Additional details

https://www.w3.org/TR/CSP2/#frame-ancestors-and-frame-options

### DIFF
--- a/files/en-us/web/http/reference/headers/content-security-policy/frame-ancestors/index.md
+++ b/files/en-us/web/http/reference/headers/content-security-policy/frame-ancestors/index.md
@@ -16,7 +16,7 @@ Setting this directive to `'none'` is similar to {{HTTPHeader("X-Frame-Options",
 > This differs from **`frame-src`**, which allows you to specify where iframes in a page may be loaded from.
 
 > [!NOTE]
-> **`frame-ancestors`** directive [checks each ancestor](https://www.w3.org/TR/CSP2/#frame-ancestors-and-frame-options). If any ancestor doesn't match, the load is cancelled. Therefore all ancestors should be allowed by the **`frame-ancestors`** directive of leaf frame when using nested frames.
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/reference/headers/content-security-policy/frame-ancestors/index.md
```diff
@@ -15,6 +15,9 @@ Setting this directive to `'none'` is similar to {{HTTPHeader("X-Frame-Options",
 > **`frame-ancestors`** allows you to specify what parent source may embed a page.
 > This differs from **`frame-src`**, which allows you to specify where iframes in a page may be loaded from.
 
+> [!NOTE]
+> **`frame-ancestors`** directive [checks each ancestor](https://www.w3.org/TR/CSP2/#frame-ancestors-and-frame-options). If any ancestor doesn't match, the load is cancelled. Therefore all ancestors should be allowed by the **`frame-ancestors`** directive of leaf frame when using nested frames.
```
bsmth: Small wording update:

```suggestion
> The **`frame-ancestors`** directive [checks each ancestor](https://www.w3.org/TR/CSP2/#frame-ancestors-and-frame-options). If any ancestor doesn't match, the load is cancelled. Therefore all ancestors should be allowed by the **`frame-ancestors`** directive of leaf frames when using nested frames.
```

bsmth (approved):
Thanks a lot for the clarification! Welcome aboard!

PR #40890, opened 2025-08-26
Update font-display example to use correct format

HowieHz:
### Description

Update font-display example to use correct format.

### Motivation

format("eot") ×
format("embedded-opentype") √

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src#font_formats


estelle (approved):
Thank you. Also, congratulations on your first merged en-US MDN content PR. Welcome to the team!

PR #40891, opened 2025-08-26
replace ui pseudo-classes GH sample embeds

chrisdavidmills:
### Description

This PR is part of https://github.com/mdn/mdn/issues/605, which sets out the aim of removing embedded learning-area repo examples into regular content. See https://github.com/mdn/mdn/issues/605#issuecomment-3219673324 specifically for a list of GH embed macros we want to remove.

This PR removes GH embed macros from the [UI pseudo-classes article](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/UI_pseudo-classes), replacing them with regular live sample embeds.


bsmth (approved):
Nice, LGTM!

Issue #37057, opened 2024-12-03
HTTP status code pages imply the optional reason phrase is a required part of the status code


angleKH:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

### What specific section or headline is this issue about?

HTTP Status codes

### What information was incorrect, unhelpful, or incomplete?

The status codes are presented in the documentation and their respective pages are presented as if both the status code (e.g. 500) and reason phrase (e.g. Internal Server Error) are both part of a single one required token.

However from what I can gather, according to the official standards, the reason phrase is *optional* in HTTP/1.1 and *not supported* in HTTP/2.

From [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-overview-of-status-codes):
>  The reason phrases listed here are only recommendations -- they can be replaced by local equivalents or left out altogether without affecting the protocol.

### What did you expect to see?

It should be made more clear that the status code and reason phrase are distinct, and noted that the reason phrase is optional.

Example: Instead of linking to a status code page like [`400 Bad Request`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) as it is currently it can be linked to like [`400`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) (Bad Request).

Each individual page status page currently has a section titled "Status" where it presents the status code along with the reason phrase, this will also need to be adjusted and seperated into a "Status code" and "Recommended reason phraae" section.

### Do you have any supporting links, references, or citations?

https://stackoverflow.com/a/38655533

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/http/status`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/status/index.md
* Last commit: https://github.com/mdn/content/commit/bd4d7bc4176d9f67297e3940ae7163a258f07ef5
* Document last modified: 2024-10-18T02:40:47.000Z

</details>

bsmth:
Thank you for reporting, I think that's a good idea, in general. I would lean towards including the reason part in links, though:
* [400 (Bad Request)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400)

as opposed to:
 
* [400](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) (Bad Request)

[...]

PR #40892, opened 2025-08-26
chore(http): mention the reason-phrase in status line is optional

bsmth:
### Description

I'm updating the section that describes the status line to show that the reason phrase is optional.
I haven't looked at updating other sections for renaming, I'm not sure how much work this would be just yet, it may not be worth the churn.

### Motivation

Part of https://github.com/mdn/content/issues/37057


pepelsbey (approved):
Looks good! Thank you :)

PR #40898, opened 2025-08-26
Update index.md: minor wording fix on combining logical and nullish coalescing operators

thetumper:
Changed "both-and" to "either-or", to correctly word the disallowed combinations.


Josh-Cena (approved):
Oh... thanks. I have no idea why I wrote that.

PR #40906, opened 2025-08-27
Add auto sizes fallback img doc

### LINKED ISSUE #40902: Add more information about the <img sizes=auto> fallback strategy
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img

The auto value within the sizes attribute description

See this thread: https://mas.to/@patrickbrosset/115099135056881997

Roderick was asking if it was possible to specify fallback values within the `sizes` attribute, in case the `auto` keyword was not supported.
Simon responded that, yes, you can do that, per spec.

It seems like this information should be on MDN, and devs shouldn't have to hunt for it in the spec.

See https://html.spec.whatwg.org/multipage/images.html#images:~:text=For%20better%20backwards%2Dcompatibility%20with%20legacy%20user%20agents%20that%20don%27t%20support%20the%20auto%20keyword%2C%20fallback%20sizes%20can%20be%20specified%20if%20desired

> For better backwards-compatibility with legacy user agents that don't support the [auto](https://html.spec.whatwg.org/multipage/images.html#valdef-sizes-auto) keyword, fallback sizes can be specified if desired.

The spec provides this example:

```html
<img loading="lazy" width="200" height="200"
     sizes="auto, (max-width: 30em) 100vw, (max-width: 50em) 50vw, calc(33vw - 100px)"
     srcset="swing-200.jpg 200w, swing-400.jpg 400w, swing-800.jpg 800w, swing-1600.jpg 1600w"
     src="swing-400.jpg" alt="Kettlebell Swing">
```

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/html/reference/elements/img`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/reference/elements/img/index.md
* Last commit: https://github.com/mdn/content/commit/0754cd805a8e010d2e3a2a065f634a3bcf358252
* Document last modified: 2025-07-09T21:49:04.000Z

</details>

AyaanB24:
### Summary
Added clarification for the auto keyword and fallback values in the sizes attribute of the  element. This ensures developers know how to maintain backward compatibility for browsers that do not support auto.

### Problem

The existing MDN documentation did not mention:
How to specify fallback values when using auto.
The importance of ordering fallback values for older browsers.

### What Changed
Updated the sizes attribute description.
Added note about backward compatibility.
Included an example from the HTML specification.

### Before

image

### After
image


files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -210,6 +210,24 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
     The `sizes` attribute also accepts the following keyword values:
```
chrisdavidmills: I wouldn't introduce the `auto` keyword value down here. Really, this should be mentioned in the first sentence of the `sizes` description, maybe something like this:

> One or more values separated by commas, which can equal source sizes or the `auto` keyword.

You should then include the description of what a source size consists of, much the same as what is already there. Maybe start it in a new paragraph below the first sentence.

Then, include the description you've added about `auto` (including the bit about backwards compatibility and the example snippet) at the same indentation level as the source size description. Maybe start the first sentence like so:

> The `auto` keyword can replace the whole list...

We generally try not to start sentences with code font.

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -210,6 +210,24 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
     - `auto`
       - : `auto` can replace the whole list of sizes or the first entry in the list. It is only valid when combined with `loading="lazy"`, and resolves to the [concrete size](/en-US/docs/Web/CSS/image) of the image. Since the intrinsic size of the image is not yet known, `width` and `height` attributes (or CSS equivalents) should also be specified to [prevent the browser assuming a default width of 300px](https://html.spec.whatwg.org/multipage/images.html#sizes-attributes:attr-dim-width).
```
```suggestion
      - : `auto` can replace the whole list of sizes or the first entry in the list. It is only valid when combined with `loading="lazy"`, and resolves to the [concrete size](/en-US/docs/Web/CSS/image) of the image. Since the intrinsic size of the image is not yet known, `width` and `height` attributes (or CSS equivalents) should also be specified to prevent the browser from assuming the default image width of 300px.
```

chrisdavidmills: This spec link isn't very helpful, and we generally try not to link to the spec from prose unless absolutely necessary.

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -210,6 +210,24 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
+      - : For better backward compatibility with browsers that do not support the `auto` keyword, you can include fallback sizes before `auto` in the `sizes` attribute.
```
```suggestion
      - : For better backward compatibility with browsers that do not support the `auto` keyword, you can include fallback sizes after `auto` in the `sizes` attribute.
```

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -210,6 +210,24 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
+
+      **Example:**
+
+      ```html
+      <img
+        loading="lazy"
+        width="200"
+        height="200"
+        sizes="(max-width: 30em) 100vw, (max-width: 50em) 50vw, calc(33vw - 100px), auto"
```
chrisdavidmills: `auto` should be first in the list, no? This is what I'd expect, and what is shown in the spec example.

chrisdavidmills (changes requested):
@AyaanB24 thanks for another great contribution! I've got a few comments for you to consider.

files/en-us/web/html/reference/elements/img/index.md
Josh-Cena: Wrong DL syntax.

```suggestion

      For better backward compatibility with browsers that do not support the `auto` keyword, you can include fallback sizes after `auto` in the `sizes` attribute.
```

files/en-us/web/html/reference/elements/img/index.md
Josh-Cena: The spec link has always been there, and the last time I documented `auto`, I kept it because I find the "default width" very surprising and I would appreciate it to be backed up with an authoritative source (since this is not an official documentation of the default width, but just a mention in passing).

files/en-us/web/html/reference/elements/img/index.md
AyaanB24: Thanks for pointing that out!  You’re right — according to the spec example, auto should come first in the list. I’ve updated the PR accordingly.

files/en-us/web/html/reference/elements/img/index.md
AyaanB24: Thanks for the feedback! I’ve removed the spec link and kept the explanation clear in prose instead. Please let me know if you’d like any further tweaks.

files/en-us/web/html/reference/elements/img/index.md
AyaanB24: You're right, the spec requires fallback sizes to be listed after auto, not before. I’ll update the documentation accordingly.

files/en-us/web/html/reference/elements/img/index.md
AyaanB24: You’re right, according to the spec example auto should come first in the list. I`ve update the example to reflect that.

chrisdavidmills (approved):
Awesome, this is looking good. Thanks again @AyaanB24!

PR #40907, opened 2025-08-27
Make description less ambiguous

HammadAwan424:
### Description
The example code block [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_objects#objects_and_properties:~:text=There%20is%20no,result%3B%0A%7D) lists all the inherited properties, regardless whether they are enumerable or not. But the provided description ambiguously says that it lists "non-enumerable" properties. And the function name was made more descriptive in this commit.

### Motivation

I was confused at this thing for a moment, I don't want others to be in the same situation. Simple.


files/en-us/web/javascript/guide/working_with_objects/index.md
```diff
@@ -306,10 +306,10 @@ function showProps(obj, objName) {
 }
 ```
 
-There is no native way to list inherited non-enumerable properties. However, this can be achieved with the following function:
+There is no native way to list inherited properties. However, this can be achieved with the following function:
```
Josh-Cena:
```suggestion
There is no native way to list all inherited properties including non-enumerable ones. However, this can be achieved with the following function:
```

We have to mention it because there does exist a way to list inherited enumerable properties (`for...in`)

files/en-us/web/javascript/guide/working_with_objects/index.md
```diff
@@ -306,10 +306,10 @@ function showProps(obj, objName) {
...
 
 ```js
-function listAllProperties(myObj) {
+function listAllInheritedProperties(myObj) {
```
Josh-Cena: I would prefer the original name:

```suggestion
function listAllProperties(myObj) {
```

Because "inherited properties" excludes "own properties".

Josh-Cena (changes requested):

files/en-us/web/javascript/guide/working_with_objects/index.md
hammaddaftab: So the original description meant something completely different than what I supposed? Hmm, I am accepting your changes, at least it makes it more clear.

Josh-Cena (approved):
Good catch, thanks

PR #40911, opened 2025-08-27
Fix content issues

Josh-Cena:
- Fix "typos" (BrE spellings)
- Fix external links
- Follow up from https://github.com/mdn/content/pull/40891: use separate code blocks for styles/scripts.


bsmth (approved):
Great, thanks 👍🏻

PR #40926, opened 2025-08-28
fix(fred): js/css wasn't loading

LeoMcA:
### Description

Updates fred to the latest version and sets FRED_WRITER_MODE=true as it's now a runtime environment variable.

### Motivation

js/css weren't loading with the previous version of fred.

### Additional details

https://github.com/mdn/fred/blob/main/CHANGELOG.md#120-2025-08-28

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "lint:typos-words-only": "cspell --no-progress --gitignore --words-only --unique --config .vscode/cspell.json \"**/*.md\"",
     "lint:yml": "prettier -c \"**/*.yml\"",
     "start": "yarn -s info:rari && yarn up-to-date-check && env-cmd --silent cross-env CONTENT_ROOT=files REACT_APP_DISABLE_AUTH=true BUILD_OUT_ROOT=build rari-server",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,15 +38,15 @@
     "lint:typos-words-only": "cspell --no-progress --gitignore --words-only --unique --config .vscode/cspell.json \"**/*.md\"",
     "lint:yml": "prettier -c \"**/*.yml\"",
     "start": "yarn -s info:rari && yarn up-to-date-check && env-cmd --silent cross-env CONTENT_ROOT=files REACT_APP_DISABLE_AUTH=true BUILD_OUT_ROOT=build rari-server",
-    "start:fred": "yarn -s info:fred && yarn up-to-date-check && env-cmd --silent cross-env CONTENT_ROOT=files BUILD_OUT_ROOT=build NODE_ENV=production fred-server",
+    "start:fred": "yarn -s info:fred && yarn up-to-date-check && env-cmd --silent cross-env CONTENT_ROOT=files NODE_ENV=production FRED_WRITER_MODE=true fred-server",
```
caugner, line 41: Nit: As mentioned on Slack, I would find it preferable if we didn't have to set `FRED_WRITER_MODE` here.

caugner (approved):

PR #40188, opened 2025-07-03
Add WebSocket Connection best practices note to handle BFCache support

gilbertococchi:
Hi, I would like to suggest adding a note to help developers using the WebSocket API correctly to avoid negative UX side effects on sites.

### Motivation

 After having analyzed most of the WebSocket API usage on the Web via HTTP Archive I've found out that 99% of the WebSocket API usage is causing BFCache ineligibility on some modern browsers such as Chrome/Chromium based ones.

More info about BFCache and why Connections should be handled this way here: https://web.dev/articles/bfcache#close-open-connections

### Description

In a nutshell I wasn't able to find developers following these best practices when using WebSocket connections:

1) close WebSocket connections via `pagehide` to make the page eligible for BFCache.

2) restart WebSocket connection via `pageshow` (`event.persisted=true` case) when the page is restored from BFCache in order to restart the WebSocket connection to resume functionality.

I would appreciate MDN folks advice and suggestions about how we can all help developers better using this widely used API (approximately 600k Sites are using WebSocket so the impact on UX is quite big).

### Additional details

https://web.dev/articles/bfcache#close-open-connections

Related to Issue [#700](https://github.com/mdn/mdn/issues/700)

I am not a technical writer, I would appreciate some support to make it clear if anyone is also supportive to include this note in the WebSocket API MDN article.

Thanks
Gilberto

PR #40931, opened 2025-08-28
New WebSockets client guide

### LINKED ISSUE #29029: No information on how to handle errors
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications

The example doesn't handle errors

From reading the page, looking at `onopen` or `onmessage` handlers, there is no error handling. What if it failed to open, or there was an error ?

Some description on what happens if there was an error.

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/api/websockets_api/writing_websocket_client_applications`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
* Last commit: https://github.com/mdn/content/commit/acfe8c9f1f4145f77653a2bc64a9744b001358dc
* Document last modified: 2023-07-07T07:19:19.000Z

</details>

wbamberg:
This came out of https://github.com/mdn/content/pull/40188, which wanted to talk about how to make WebSocket clients bfcache-compatible, and https://github.com/orgs/mdn/discussions/817, in which we find that the example that's supposed to underpin the current [Writing WebSocket client applications](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications) page doesn't work.

In https://github.com/mdn/dom-examples/pull/326 I added a new example that does work and is bfcache-compatible.

This PR updates the guide page to use that example as a backbone. I hope I haven't jettisoned anything too important from the old page.

I wonder if we ought to say more about URL schemes: this example uses `ws` which is maybe shouldn't, but it can't use `wss` and still work with localhost AFAICT. But it could I think use a relative URL, and maybe should?

I also don't talk about the `protocols` option, as we don't use it in the example and it would feel like quite a big digression. Maybe that would be better as a separate guide?

On the differences between different browsers: this is mostly just what I have observed by testing. But this: https://docs.google.com/document/d/1JtDCN9A_1UBlDuwkjn1HWxdhQ1H2un9K4kyPLgBqJUc/edit?tab=t.0#heading=h.58d6ijfz2say is also a useful doc.

I wasn't sure about the "Security considerations" bit: I've kept it as-is but AFAIK mixed content just isn't allowed any more, and I couldn't find any more detail on "Most browsers now only allow secure WebSocket connections" (though I would love to know more about this, because if so it would be good to document it in BCD and describe it better here...).

I wouldn't be particularly averse to adding a note in https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API about bfcache-compatibility as well, if we think it merits more visibility than this.

If this merges I should file a follow-up to update https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_a_WebSocket_server_in_JavaScript_Deno with the new example, too.

cc @gilbertococchi and maybe @tunetheweb ?

Fixes #29029


files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
+In this guide we'll walk through the implementation of a WebSocket-based ping application. In this application, the client sends a "ping" message to the server every second, and the server responds with a "pong" message. The client listens for "pong" messages and logs them, keeping track of how many message exchanges there have been.
 
-> [!NOTE]
-> The example snippets in this article are taken from our WebSocket chat client/server sample.
-> [See the code](https://github.com/mdn/samples-server/tree/master/s/websocket-chat).
+Although this is a pretty minimal application, it covers the fundamental points involved in writing a WebSocket client.
 
-## Creating a WebSocket object
+You can find the complete example at [https://github.com/mdn/dom-examples/tree/main/websockets](https://github.com/mdn/dom-examples/tree/main/websockets). The server side is written in [Deno](https://deno.com/), so you'll have to install that first if you want to try running the example locally.
```
chrisdavidmills:
```suggestion
You can find the complete example at [https://github.com/mdn/dom-examples/tree/main/websockets](https://github.com/mdn/dom-examples/tree/main/websockets). The server side is written in [Deno](https://deno.com/), so you'll have to install that first if you want to run the example locally.
```

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
 
-In order to communicate using the WebSocket protocol, you need to create a {{domxref("WebSocket")}} object; this will automatically attempt to open the connection to the server.
+## Creating a `WebSocket` object
 
-The WebSocket constructor accepts one required and one optional parameter:
+In order to communicate using the WebSocket protocol, you need to create a {{domxref("WebSocket")}} object. As soon as you create this object, it will start trying to connect to the server.
```
chrisdavidmills:
```suggestion
To communicate using the WebSocket protocol, you need to create a {{domxref("WebSocket")}} object. As soon as you create this object, it will start trying to connect to the specified server.
```

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
 
-- `url`
-  - : The URL to which to connect; this should be the URL to which the WebSocket server will respond.
-    This should use the URL scheme `wss://`, although some software may allow you to use the insecure `ws://` for a local connection.
-    Relative URL values and `https://` and `http://` schemes are also allowed in [most recent browser versions](/en-US/docs/Web/API/WebSocket/WebSocket#browser_compatibility).
-- `protocols` {{optional_inline}}
-  - : Either a single protocol string or an array of protocol strings.
-    These strings are used to indicate sub-protocols, so that a single server can implement multiple WebSocket sub-protocols (for example, you might want one server to be able to handle different types of interactions depending on the specified `protocol`).
-    If you don't specify a protocol string, an empty string is assumed.
+The `WebSocket` constructor takes one mandatory argument, which is the URL of the WebSocket server to connect to. In this case, since we're running the server locally, we're using the address of localhost.
```
chrisdavidmills:
```suggestion
The `WebSocket` constructor takes one mandatory argument — the URL of the WebSocket server to connect to. In this case, since we're running the server locally, we're using the localhost address.
```

chrisdavidmills: Also, maybe add a note below this to mention the `ws` protocol, say what it is, say that we're using `ws` because we're connecting to localhost, and that in a real application you'd need to use `wss` to connect to a secure server?

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
+
+The constructor takes another optional argument [`protocols`](/en-US/docs/Web/API/WebSocket/WebSocket#protocols), which allows a single server to implement multiple sub-protocols. We're not using this feature in our example.
```
chrisdavidmills, line 29: I agree with your note that this probably should be discussed in a separate guide. It certainly doesn't belong in a basic usage guide, although from reading this and following the link, it feels like it'd be nicer if we provided some more info about it somewhere on MDN.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
-
-The browser may also output to its console a more descriptive error message as well as a closing code as defined in [RFC 6455, Section 7.4](https://datatracker.ietf.org/doc/html/rfc6455#section-7.4) through the {{domxref("CloseEvent")}}.
+## Listening for the `open` event
 
-### Examples
+Creating a `WebSocket` instance starts the process of establishing a connection to the server. Once the connection is established, the {{domxref("WebSocket.open_event", "open")}} event is fired, and after this point the socket is able to transmit data.
 
-This simple example creates a new WebSocket, connecting to the server at `wss://www.example.com/socketserver`.
-A custom protocol of "protocolOne" is named in the request for the socket in this example, though this can be omitted.
+In the example code below, when the `open` event is fired, we start sending one "ping" message every second, using the {{domxref("Window.setInterval()")}} API:
```
chrisdavidmills:
```suggestion
In the example code below, when the `open` event is fired, we start sending one "ping" message to the server every second, using the {{domxref("Window.setInterval()")}} API:
```

chrisdavidmills: I'm not sure if you feel like this is overkill, but to me it felt good to remind the reader that we are on the client and sending messages to the server.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
-  exampleSocket.send("Here's some text that the server is urgently awaiting!");
+const message = {
+  iteration: counter,
+  content: "ping",
 };
+websocket.send(JSON.stringify(message));
 ```
 
-### Using JSON to transmit objects
+The `send()` method is asynchronous, and the {{domxref("WebSocket.bufferedAmount")}} property represents the number of bytes that have not yet been transmitted. Note that the WebSockets protocol uses {{glossary("UTF-8")}} to encode text, so `bufferedAmount` is calculated based on the UTF-8 encoding of any buffered text data.
```
chrisdavidmills: I found this interesting — looking at the `send()` ref page, the method itself is not conventionally asynchronous (i.e. it is not promise or callback-based). It looks like the method itself is synchronous, but the underlying process of sending the messages is async (they are added to a queue and sent when it is their turn).

In light of this:

1. Is it accurate to say that `send()` is async?
2. Do you want to dig into `bufferedAmount` a bit more. Is it specifically the amount of bytes in the queue that are as yet unsent?

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
 ```
 
-## Receiving messages from the server
+The server can also send binary data, which is exposed to clients as a {{domxref("Blob")}} or an {{jsxref("ArrayBuffer")}}, based on the value of the {{domxref("WebSocket.binaryType")}} property.
 
-WebSockets is an event-driven API; when messages are received, a `message`
-event is sent to the `WebSocket` object. To handle it, add an event listener
-for the `message` event, or use the {{domxref("WebSocket/message_event", "onmessage")}} event handler.
-To begin listening for incoming data, you can do something like this:
+As we saw for sending messages, the server could also send JSON-serialized objects, which the client can then parse:
```
chrisdavidmills:
```suggestion
As we saw for sending messages, the server can also send JSON strings, which the client can then parse into an object:
```

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
+  }, 1000);
+});
 ```
 
-You can send data as a string, {{ domxref("Blob") }}, or {{jsxref("ArrayBuffer")}}.
+In our example we send text, but you can also send binary data as a {{domxref("Blob")}}, {{jsxref("ArrayBuffer")}}, {{jsxref("TypedArray")}}, or {{jsxref("DataView")}}.
 
-As establishing a connection is asynchronous and prone to failure there is no guarantee that calling the `send()` method immediately after creating a WebSocket object will be successful.
-We can at least be sure that attempting to send data only takes place once a connection is established by defining an {{domxref("WebSocket/open_event", "onopen")}} event handler to do the work:
+A common approach is to use {{glossary("JSON")}} to send and receive serialized JavaScript objects. For example, instead of just sending the text message "ping", our client could send a serialized object including the number of messages exchanged so far:
```
chrisdavidmills:
```suggestion
A common approach is to use {{glossary("JSON")}} to send serialized JavaScript objects as text. For example, instead of just sending the text message "ping", our client could send a serialized object including the number of messages exchanged so far:
```

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
 
-### Receiving and interpreting JSON objects
-
-Let's consider the chat client application first alluded to in [Using JSON to transmit objects](#using_json_to_transmit_objects). There are assorted types of data packets the client might receive, such as:
+## Handling disconnect
 
-- Login handshake
-- Message text
-- User list updates
+When the connection is closed, because either client or server closed it or because there was an error, then the {{domxref("WebSocket.close_event", "close")}} event will be fired.
```
chrisdavidmills:
```suggestion
When the connection is closed because either the client or the server closed it or an error occurred, the {{domxref("WebSocket.close_event", "close")}} event will be fired.
```

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
+  clearInterval(pingInterval);
+});
 ```
 
-Here we use {{jsxref("JSON.parse()")}} to convert the JSON object back into the original object, then examine and act upon its contents.
+## Working with the bfcache
+
+The back/forward cache, or {{glossary("bfcache")}}, enables much faster back and forward navigation between pages that the user has recently visited. It does this by storing a complete snapshot of the page, including the JavaScript heap.
+
+The browser pauses and then resumes JavaScript execution when a page is added to or loaded from the bfcache. This means that, depending on what the page is doing, it's not always safe for the browser to use the bfcache for the page. If the browser determines that it is not safe, then the page will not be added to the bfcache, and then the user will not get the performance benefit that it can bring.
```
chrisdavidmills:
```suggestion
The browser pauses and then resumes JavaScript execution when a page is added to or loaded from the bfcache. This means that, depending on what the page is doing, it's not always safe for the browser to use the bfcache for the page. If the browser determines that it is not safe, the page will not be added to the bfcache, and the user will not get the performance benefit that it can bring.
```

chrisdavidmills: Also, maybe link to https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Monitoring_bfcache_blocking_reasons for more information on why pages might not be added to the bfcache.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
+  if (websocket) {
+    log("CLOSING");
+    websocket.close();
+    websocket = null;
+    window.clearInterval(pingInterval);
+  }
+});
+```
+
+Conversely, by listening for the {{domxref("Window.pageshow_event", "pageshow")}} event, you can seamlessly start the connection again when the page is loaded, or when it is retrieved from the bfcache. So in our example we will add all the code to initialize our WebSocket and set up its event listeners inside the `pageshow` event handler:
```
chrisdavidmills:
```suggestion
Conversely, by listening for the {{domxref("Window.pageshow_event", "pageshow")}} event, you can seamlessly start the connection again when the page is loaded or retrieved from the bfcache. In our example, we add all the code to initialize our WebSocket and set up its event listeners inside the `pageshow` event handler:
```

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
+
+  websocket.addEventListener("error", (e) => {
+    log(`ERROR: ${e.data}`);
+  });
+});
 ```
 
-It may be helpful to examine the socket's {{domxref("WebSocket.bufferedAmount", "bufferedAmount")}} attribute before attempting to close the connection to determine if any data has yet to be transmitted on the network.
-If this value isn't 0, there's pending data still, so you may wish to wait before closing the connection.
+If you run our example, try navigating to a different page, then back to the example. In Chrome, you should see that the example starts the connection again, and keeps its original context: so, for example, it remembers the count of exchanged messages.
```
chrisdavidmills, line 185: In Firefox and Safari, it just seems to start the app again from its initial state, which is a shame. Worth mentioning?

chrisdavidmills (changes requested):
Nice work @wbamberg; this is a massive improvement over what we had before, in terms of both demo and guide. I've left you a few comments, but nothing major really.

I would still like to see a tech review from the other folks on the thread.

wbamberg:
Thanks for the review, Chris! I'm out until Tuesday but will look when I get back.

gilbertococchi:
Thanks everyone for your support for the WebSocket guide, sorry for late replying but I had been busy with lots of conflicting priorities these weeks.

I will review this first thing next week and will review the changes.

Thank you so much again for helping developers awareness about this opportunity for their sites when using WebSocket!

Josh-Cena:
Looks like this may fix https://github.com/mdn/content/issues/29029, though I'm not sure if completely.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: This is not what I see in Safari, and I would like to know why we see different things here.

But it is what I see in Firefox, and that's expected: because the page isn't added to the bfcache, then it is reloaded, so the log starts out empty and the counter starts again at zero.

In Safari, after a forward/back navigation, I see this:

Screenshot 2025-09-02 at 8 57 24 AM

Here, I clicked "forward" after exchange 4, and then "back". I think what happens here is that:

- the page was added to/loaded from the bfcache, so the HTML/JS is not reloaded, and we keep the contents of the log and the counter value
- but because Safari closed the network request as soon as we navigated forward, we get a WebSocket error notification


At least this is my understanding of the situation. It is perhaps worth spelling this out in the guide, maybe not to the screenshot level of detail though.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: To me it is async because the caller doesn't have to wait for the data to be sent before the function returns and the caller can continue:

```js
s.send(data);
// the data isn't actually sent when we get here
```

I appreciate it is unusual in that there is no way the caller is told that the operation has completed (i.e. with a promise or a callback or an event). But I would still call it async. Still I don't think getting hung up on the terminology is very useful so I'm happy to remove the word.

```suggestion
The `send()` method does not wait for the data to be transmitted before returning to the caller. It just adds the data to its internal buffer and begins the process of transmission. The {{domxref("WebSocket.bufferedAmount")}} property represents the number of bytes that have not yet been transmitted. Note that the WebSockets protocol uses {{glossary("UTF-8")}} to encode text, so `bufferedAmount` is calculated based on the UTF-8 encoding of any buffered text data.
```

> Do you want to dig into bufferedAmount a bit more. Is it specifically the amount of bytes in the queue that are as yet unsent?

Isn't that just what "The {{domxref("WebSocket.bufferedAmount")}} property represents the number of bytes that have not yet been transmitted" says?

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg:
```suggestion
When the connection is closed, because either the client or the server closed it or because an error occurred, the {{domxref("WebSocket.close_event", "close")}} event will be fired.
```

I think we need the extra "because" or it is ambiguous:

- the client closed it
- the server closed it or an error occurred

rather than:

- the client closed it
- the server closed it
- an error occurred

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: I wonder why you are deleting my "then"s? I like my thens. "If....then" I think is clearer structurally than an implicit then.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: Again I think I like my extra words around ", or when it is". I think it makes the structure clearer. I suppose this is subjective?

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: I think it's a shame that this is presented as a page on how to use a Chrome-only feature. It would be better to have a page under https://developer.mozilla.org/en-US/docs/Web/Performance, that talks about the bfcache as it works across browsers, and the various reasons pages might not be bfcache-friendly in various browsers, and links to `notRestoredReasons` as a tool you can use in that context.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
tunetheweb: FWIW I agree it's async. A promise or callback isn't necessary for a call to be considered async (although it's nice when it does!)

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
tunetheweb: I would say "restored from the bfcache" rather than "loaded from the bfcache".

The bfcache isn't a "cache" really. More a preservation and restore of a page's state.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
+
+The back/forward cache, or {{glossary("bfcache")}}, enables much faster back and forward navigation between pages that the user has recently visited. It does this by storing a complete snapshot of the page, including the JavaScript heap.
+
+The browser pauses and then resumes JavaScript execution when a page is added to or loaded from the bfcache. This means that, depending on what the page is doing, it's not always safe for the browser to use the bfcache for the page. If the browser determines that it is not safe, then the page will not be added to the bfcache, and then the user will not get the performance benefit that it can bring.
+
+Different browsers use different criteria for adding a page to the bfcache. If a page has an active WebSocket connection, then:
 
-### Text data format
+- Firefox and Chrome will not add the page to the bfcache.
+- Safari will add the page to the bfcache, but will cancel any active network requests. This is likely to generate an error in your WebSocket connection, leading to the connection being closed.
```
tunetheweb: I'd probably be less explicit here as this ius subject to change.

For example Chrome is looking into whether just using a WebSocket really should disqualify a page from the bfcache. Or if we can still pout it in the bfcache but evict it only if a message is received on that WebSocket connection.

So maybe just leave as "Different browsers use different criteria for adding a page to the bfcache. For maximum support of the bfcache it is recommended to close any WebSocket clients on the {{domxref("Window.pagehide_event", "pagehide")}} event..."

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
tunetheweb: `pageshow` is also fired on page load. So might wanna mention that (to avoid people opening a connection twice on pageload).

tunetheweb (commented):
Mostly LGTM but have made some comments.

Note that I'm far from a WebSockets expert, but I do know a fair bit about bfcache.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
tunetheweb: > I think it's a shame that this is presented as a page on how to use a Chrome-only feature. It would be better to have a page under https://developer.mozilla.org/en-US/docs/Web/Performance, that talks about the bfcache as it works across browsers, and the various reasons pages might not be bfcache-friendly in various browsers, and links to notRestoredReasons as a tool you can use in that context.

That's fair. The concepts are generally the same. It's just that only Chrome currently exposes the reasons in an easy manner. I think testing each in each browser is gonna be quite painful without that as could run into false positives (e.g. you think it didn't use bfcache for X reasons,  but it was actually for Y reason).

Either way I'm in two minds whether you should link it. I mean it is the only definitive list (and is part of the HTML spec btw!), but also is only queryable from Chrome as you say.

tunetheweb:
> I wouldn't be particularly averse to adding a note in https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API about bfcache-compatibility as well, if we think it merits more visibility than this.

I think this would still be good as a short note (perhaps linking to the relevant section in this guide?). This updated guide is good, but it feels a little out of the way.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
gilbertococchi: Hi @wbamberg, thanks for raising this.

I would love to hear from someone from WebKit if possible if they can share what Safari is doing with WebSocket.

I also noticed that Safari is behaving somehow differently from Chrome and Firefox.

It looks like pages using WebSocket without closing the connection via pagehide are still eligible for BFCache on Safari somehow.
From my tests it looked like the approach of closing the connection via pagehide and restarting the connection via pageshow was also working in Safari.

Can you please share the error you are getting?

gilbertococchi: Correction: Safari is definitively throwing a readyState 3 error (CLOSED), although I see the connection restarting on Safari too on my test page (https://output.jsbin.com/sepojum).

Did you see the WebSocket connection not working on a BFCache restored page on Safari while testing?

wbamberg: > It looks like pages using WebSocket without closing the connection via pagehide are still eligible for BFCache on Safari somehow.

This is what I would have expected, given the info in https://docs.google.com/document/d/1JtDCN9A_1UBlDuwkjn1HWxdhQ1H2un9K4kyPLgBqJUc/edit?tab=t.0#heading=h.yqabnzb880lx :

> Safari caches such pages but cancels active network requests.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: Yes, sorry, I didn't mean to sound so negative about it! It is a good article and obviously a useful API. And we should link to it. I just think it would be great to have a browser-agnostic article about bfcache compatibility on MDN, that's not tied to specific APIs. But I'm not planning to write that any time soon.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
chrisdavidmills: OK, your explanation works for me. I'd keep the word "async", but also add the additional bit of explanation you've provided in your above suggestion, just to clarify.

> Isn't that just what "The {{domxref("WebSocket.bufferedAmount")}} property represents the number of bytes that have not yet been transmitted" says?

Yeah, I guess so ;-) OK, nothing more to do there.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
chrisdavidmills: OK, WFM

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
chrisdavidmills: > Again I think I like my extra words around ", or when it is". I think it makes the structure clearer. I suppose this is subjective?

Yes, I think so. They seem a little wordy to me, but it is not wrong. Up to you whether you want to commit these suggestions.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
chrisdavidmills: > This is not what I see in Safari, and I would like to know why we see different things here.

Ah, hold on. I've tried it again in Safari. When I navigate to a new page by typing the URL into the address bar and pressing Enter (or following a link), then press "Back", I get the behavior I describe (I am using gamespot.com as a test).

When I then press "Forward" to go back to the new page, then press "Back", I get the behavior your screenshot is showing.

I've tried it with a few other example pages, and get the same effects,

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: Added in https://github.com/mdn/content/pull/40931/commits/9e4076fdddb7af3d145df010490defb352ecafbc. I don't think this note is great, and I don't fully understand how protocols work with WebSockets - this is a thing that ought to be explained in our docs, but I think it is out of scope for my change here.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: -> https://github.com/mdn/content/pull/40931/commits/768ae6e117d501adaec07a140690fc7ca7fedaa1

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: -> https://github.com/mdn/content/pull/40931/commits/ef5e49c14e077ebfc37950ec7aeb4078cbd6bdca and https://github.com/mdn/content/pull/40931/commits/86ebca5c5d6e506d43b0dd6b39fbe914ab605dd6

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: > pageshow is also fired on page load. So might wanna mention that (to avoid people opening a connection twice on pageload).

I do mention this:

> you can seamlessly start the connection again when the page **is loaded or** retrieved from the bfcache.

good enough?

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: I chose slightly different wording, hope this is OK? -> https://github.com/mdn/content/pull/40931/commits/1b0a04944397070b1bf1db4922b58d79d0c8e03d

This also gets us out of puzzling over Safari's behavior here.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: Yes, I see that as well. In https://github.com/mdn/content/pull/40931/commits/1b0a04944397070b1bf1db4922b58d79d0c8e03d I've been less specific about what the browsers do, so I hope that means we don't have to worry about this.

wbamberg:
> > I wouldn't be particularly averse to adding a note in https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API about bfcache-compatibility as well, if we think it merits more visibility than this.
> 
> I think this would still be good as a short note (perhaps linking to the relevant section in this guide?). This updated guide is good, but it feels a little out of the way.

-> https://github.com/mdn/content/pull/40931/commits/21592f9756a9e38bafa3e11201e391136ceba34d (I deleted a previous note that I didn't think was particularly useful)

wbamberg:
Thanks for the reviews, people! I think this is ready for another look.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
tunetheweb: FWIW I think the note is good!

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
tunetheweb: Mmm.... it's the "start the connection **again**" which is not quite working for me. If you're using this as the main way of initialising the web socket on initial page load, then it't not again. Also "when the page is loaded or retrieved from the bfcache" could easily be misread to mean "when the page is (loaded or retrieved) from the bfcache" rather than "when the page (is loaded) or (retrieved from the bfcache)"

WDYT about something more explicit like this?:

> Conversely, by listening for the {{domxref("Window.pageshow_event", "pageshow")}} event, you can seamlessly start the connection again when the page is retrieved from the bfcache. Since the `pageshow` event also fires on page load, this can also be used to initiate the websocket connection on initial page load."

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
tunetheweb: LGTM!

tunetheweb:
Looks great now. One nit on the page_show on page load to consider, but other than that LGTM.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
wbamberg: ah, makes sense, I missed the "again". Does https://github.com/mdn/content/pull/40931/commits/f15529db65e4abd362907808fe703e08f22ced13 work for you?

tunetheweb: LGTM! Thanks.

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
 ```
 
-Here we use {{jsxref("JSON.parse()")}} to convert the JSON object back into the original object, then examine and act upon its contents.
+## Working with the bfcache
 
-### Text data format
+The back/forward cache, or {{glossary("bfcache")}}, enables much faster back and forward navigation between pages that the user has recently visited. It does this by storing a complete snapshot of the page, including the JavaScript heap.
 
-Text received over a WebSocket connection is in UTF-8 format.
+The browser pauses and then resumes JavaScript execution when a page is added to or restored from the bfcache. This means that, depending on what the page is doing, it's not always safe for the browser to use the bfcache for the page. If the browser determines that it is not safe, then the page will not be added to the bfcache, and then the user will not get the performance benefit that it can bring.
```
chrisdavidmills:
```suggestion
The browser pauses and then resumes JavaScript execution when a page is added to or restored from the bfcache. This means that, depending on what the page is doing, it's not always safe for the browser to use the bfcache for the page. If the browser determines that it is not safe, the page will not be added to the bfcache, and the user will not get the performance benefit that it can bring.
```

chrisdavidmills (approved):
@wbamberg One minor language nitpick, which you can take or leave. Apart from that, LGTM!

files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
```diff
@@ -4,187 +4,189 @@ slug: Web/API/WebSockets_API/Writing_WebSocket_client_applications
...
 ```
 
-It may be helpful to examine the socket's {{domxref("WebSocket.bufferedAmount", "bufferedAmount")}} attribute before attempting to close the connection to determine if any data has yet to be transmitted on the network.
-If this value isn't 0, there's pending data still, so you may wish to wait before closing the connection.
+Conversely, by listening for the {{domxref("Window.pageshow_event", "pageshow")}} event, you can seamlessly start the connection again when the page is restored from the bfcache. Since the `pageshow` event also fires on page load, it can also be used to start the WebSocket connection when the page is first loaded:
+
+```js
+let websocket = null;
+
+window.addEventListener("pageshow", () => {
```
fergald, line 156: Sorry, I'm very late to this but shouldn't this be checking for `persisted`, otherwise it might fire on the initial `pageshow`.

gilbertococchi: Great catch, you are right.

I think that this demo is using the `pageshow` event handler to both initialize the WebSocket connection at initial load as well as restoring the same logic for BFCache restore.

I believe most developers will not use the pageshow event handler this way.

+1 to change this to initialize WebSocket normally on pageload and run the same logic on pageshow event.persisted=true to make it work again for BFCache.

bsmth: Thanks both for the feedback. I've created an issue to track this here so that this doesn't get lost in the chatter:

- [ ] https://github.com/mdn/content/issues/41357

Feel free to add more context or details in the issue 👍🏻

wbamberg: > I think that this demo is using the pageshow event handler to both initialize the WebSocket connection at initial load as well as restoring the same logic for BFCache restore.
>
> I believe most developers will not use the pageshow event handler this way.

For my education, why not?

tunetheweb: Yeah I'm curious too. IIRC we specially did think about this and added it this was intentionally. As [shown by the comment above](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications#working_with_the_bfcache):

> Since the pageshow event also fires on page load, it can also be used to start the WebSocket connection when the page is first loaded:

wbamberg: Just for visibility, I've linked these comments from the new issue: https://github.com/mdn/content/issues/41357#issuecomment-3362451522 and I guess we should talk about it in there :).

PR #40947, opened 2025-08-30
docs: clarify FOUT note in preload example

cupok:
The previous paragraph about preloading images ended with a note about fonts and FOUT,
which could be misleading in context. Updated the wording to clarify that FOUT reduction
applies when using preload with fonts, while keeping the image preload example intact.


estelle (approved):
Thanks! 🎉

PR #40948, opened 2025-08-30
typo fix in example code for sizes attribute

ksilix:
### Description

Fixed the example code based on below explanation (If the viewport is 600px wide or less, the slot size is 400px; otherwise, it is 800px.)

### DIFF
--- a/files/en-us/web/html/reference/elements/picture/index.md
+++ b/files/en-us/web/html/reference/elements/picture/index.md
@@ -118,7 +118,7 @@ For example:
 <picture>
   <source
     srcset="small.jpg 480w, medium.jpg 800w, large.jpg 1200w"
-    sizes="(max-width: 600px) 400px, 600px"
+    sizes="(max-width: 600px) 400px, 800px"
     type="image/jpeg" />
   <img src="fallback.jpg" alt="Example image" />
 </picture>


files/en-us/web/html/reference/elements/picture/index.md
```diff
@@ -118,7 +118,7 @@ For example:
 <picture>
   <source
     srcset="small.jpg 480w, medium.jpg 800w, large.jpg 1200w"
-    sizes="(max-width: 600px) 400px, 600px"
+    sizes="(max-width: 600px) 400px, 800px"
```
estelle, line 121: Thanks for catching and ediing the typo.

I think we should change line 129 instead. We wouldn't want an 800px wide img in a viewport that is less than 600px wide.

ksilix: I think `sizes="(max-width: 600px) 400px, 800px"` tells if width <= 600px choose 400px and otherwise width > 600px choose 800px no?
Since we already have a value for 800w in srcset, changing 600px to 800px seems more easy to me

Josh-Cena (approved):
Yeah this makes total sense to me. Thanks! Using `800px` is better because there won't be two occurrences of `600px` which would be confusing.

PR #40952, opened 2025-08-31
Use RFC macro consistently

mohd-akram:


Josh-Cena:
Thanks for the PR. We are not interested in manually keeping the macros' styles consistent. In the mid-/long-term we may have a formatter/linter to autofix capitalization, spacing, and empty slot vs. empty string, but so far no one finds it worth the churn.

mohd-akram:
Hmm, I understand, but these were the only two instances of empty slot. I found them because they broke my custom parser.

Josh-Cena:
I'm trying to get Rari to expose its macro parser and AST format because they are working on an editor extension. This would be useful for myself too.

PR #40959, opened 2025-09-01
Editorial review: Document __Http- and __Host-Http- cookie prefixes

chrisdavidmills:
### Description

Chrome 140 adds support for the `__Http-` and `__Host-Http-` cookie prefixes: see https://chromestatus.com/feature/5170139586363392, and also see https://github.com/httpwg/http-extensions/blob/main/draft-ietf-httpbis-layered-cookies.md#the-__http--prefix for further info.

This PR documents the new prefixes.

Can you give me a technical review, @yoavweiss?


files/en-us/web/api/document/cookie/index.md
```diff
@@ -54,11 +54,10 @@ You can also assign to this property a string of the form `"key=value"`, specify
 
 - The cookie value string can use {{jsxref("Global_Objects/encodeURIComponent", "encodeURIComponent()")}} to ensure that the string does not contain any commas, semicolons, or whitespace (which are disallowed in cookie values).
 - Some user agent implementations support the following cookie prefixes:
-  - `__Secure-` Signals to the browser that it should only include the cookie in requests transmitted over a secure channel.
-  - `__Host-` Signals to the browser that in addition to the restriction to only use the cookie from a secure origin, the scope of the cookie is limited to a path attribute passed down by the server.
-    If the server omits the path attribute the "directory" of the request URI is used.
-    It also signals that the domain attribute must not be present, which prevents the cookie from being sent to other domains.
-    For Chrome the path attribute must always be the origin.
+  - **`__Secure-`**: Cookies with names starting with `__Secure-` (dash is part of the prefix) must be set with the `Secure` attribute by a secure page (HTTPS).
```
yoavweiss: The "dash is part of the prefix" part is true for all prefixes. Should it be mentioned further up?

files/en-us/web/api/document/cookie/index.md
```diff
@@ -54,11 +54,10 @@ You can also assign to this property a string of the form `"key=value"`, specify
...
+  - **`__Host-`**: Cookies with names starting with `__Host-` must be set with the `Secure` attribute by a secure page (HTTPS) and in addition must be sent only to the host subdomain or domain that set them, and not to any other host. They must not have a `Domain` attribute specified, and the `Path` attribute must be set to `/` (for Chrome, the path attribute must always be the origin.).
```
yoavweiss: @johannhof @DCtheTall Can you look at the "for chrome" part? Is this correct?

yoavweiss: Maybe:
```
 - **`__Host-`**: Cookies with names starting with `__Host-` must be set with the `Secure` attribute by a secure page (HTTPS). In addition they must not have a `Domain` attribute specified, and the `Path` attribute must be set to `/`. That guarantees that such cookies are only sent to the host that set them, and not to any other host on the domain. It also guarantees that they are set host-wide and cannot be overridden on any path on that host.
```

files/en-us/web/api/document/cookie/index.md
```diff
@@ -54,11 +54,10 @@ You can also assign to this property a string of the form `"key=value"`, specify
...
+  - **`__Http-`**: Cookies with names starting with `__Http-` must be set with the `Secure` flag by a secure page (HTTPS) and in addition must have the `HttpOnly` attribute set to prove that they were set via the `Set-Cookie` header (they can't be set or modified via JavaScript features such as `Document.cookie` or the [Cookie Store API](/en-US/docs/Web/API/Cookie_Store_API)).
+  - **`__Host-Http-`**: Cookies with names starting with `__Host-Http-` must be set with the `Secure` flag by a secure page (HTTPS) and in addition must have the `HttpOnly` attribute set to prove that they were set via the `Set-Cookie` header and must be sent only to the host subdomain or domain that set them, and not to any other host. They must not have a `Domain` attribute specified, and the `Path` attribute must be set to `/`. This combination yields a cookie that is as close as a cookie can be to treating the origin as a security boundary while at the same time ensuring developers and server operators know that its scope is limited to HTTP requests.
```
yoavweiss: Maybe this can refer to the definition of `__Host-` instead of reiterating it here? If not, I think it may be clearer to turn this into a list of conditions.

files/en-us/web/api/document/cookie/index.md
chrisdavidmills: I think this line contains the first mention of an actual prefix. The earlier text talks about attributes. Or maybe I'm wrong? Where would you put it?

yoavweiss: Maybe instead of "Some user agent implementations support the following cookie prefixes" we could have something like "The cookie name can have a semantically meaningful prefix that impose restrictions on the cookie's attributes in supporting user-agents. Such prefixes all start with `__` and end in `-`. The following prefixes are defined:"

files/en-us/web/api/document/cookie/index.md
chrisdavidmills: I've removed the "for chrome" part for now (it has probably been on that page for a long time), but I can add it back in if it is deemed worthwhile.

I have also used your updated wording in all places where these descriptions are used. Thanks!

files/en-us/web/api/document/cookie/index.md
chrisdavidmills: OK, so I've updated the description to:

> Cookies with names starting with `__Host-Http-` must be set with the `Secure` flag by a secure page (HTTPS) and must have the `HttpOnly` attribute set to prove that they were set via the `Set-Cookie` header. In addition, they also have the same restrictions as `__Host-`-prefixed cookies. This combination yields a cookie that is as close as can be to treating the origin as a security boundary while at the same time ensuring developers and server operators know that its scope is limited to HTTP requests.

files/en-us/web/api/document/cookie/index.md
chrisdavidmills: Ah yes, I like this. I've updated all such descriptions to use this kind of form, with a few tweaks, and then deleted the "dash is part of the prefix" bit from subsequent bullets.

files/en-us/web/api/document/cookie/index.md
```diff
@@ -53,12 +53,11 @@ You can also assign to this property a string of the form `"key=value"`, specify
...
+  - **`__Http-`**: Cookies with names starting with `__Http-` must be set with the `Secure` flag by a secure page (HTTPS) and in addition must have the `HttpOnly` attribute set to prove that they were set via the `Set-Cookie` header (they can't be set or modified via JavaScript features such as `Document.cookie` or the [Cookie Store API](/en-US/docs/Web/API/Cookie_Store_API)).
+  - **`__Host-Http-`**: Cookies with names starting with `__Host-Http-` must be set with the `Secure` flag by a secure page (HTTPS) and must have the `HttpOnly` attribute set to prove that they were set via the `Set-Cookie` header. In addition, they also have the same restrictions as `__Host-`-prefixed cookies. This combination yields a cookie that is as close as can be to treating the origin as a security boundary while at the same time ensuring developers and server operators know that its scope is limited to HTTP requests.
```
yoavweiss, line 60: Should the "origin as a security boundary" part be mentioned in `__Host-` as well?

yoavweiss (commented):
LGTM

files/en-us/web/api/document/cookie/index.md
chrisdavidmills: That makes sense; added.

chrisdavidmills: Thanks for the technical review! I'll move this on to the editorial review stage now, but feel free to chime in if you have any more concerns.

files/en-us/web/api/document/cookie/index.md
```diff
@@ -53,12 +53,11 @@ You can also assign to this property a string of the form `"key=value"`, specify
...
 
 - The cookie value string can use {{jsxref("Global_Objects/encodeURIComponent", "encodeURIComponent()")}} to ensure that the string does not contain any commas, semicolons, or whitespace (which are disallowed in cookie values).
-- Some user agent implementations support the following cookie prefixes:
-  - `__Secure-` Signals to the browser that it should only include the cookie in requests transmitted over a secure channel.
-  - `__Host-` Signals to the browser that in addition to the restriction to only use the cookie from a secure origin, the scope of the cookie is limited to a path attribute passed down by the server.
-    If the server omits the path attribute the "directory" of the request URI is used.
-    It also signals that the domain attribute must not be present, which prevents the cookie from being sent to other domains.
-    For Chrome the path attribute must always be the origin.
+- The cookie name can have a semantically meaningful prefix that imposes restrictions on the cookie's attributes in supporting user-agents. All cookie prefixes start with a double-underscore (`__`) and end in a dash (`-`). The following prefixes are defined:
```
bsmth: Maybe we can drop "semantically meaningful" for brevity?

```suggestion
- The cookie name can have a prefix that imposes specific restrictions on the cookie's attributes in supporting user-agents. All cookie prefixes start with a double-underscore (`__`) and end in a dash (`-`). The following prefixes are defined:
```

files/en-us/web/http/reference/headers/set-cookie/index.md
```diff
@@ -60,21 +60,24 @@ Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnl
...
 
-    > [!NOTE]
-    > Some `<cookie-name>` have a specific semantic:
-    >
-    > **`__Secure-` prefix**: Cookies with names starting with `__Secure-` (dash is part of the prefix) must be set with the `secure` flag from a secure page (HTTPS).
-    >
-    > **`__Host-` prefix**: Cookies with names starting with `__Host-` are sent only to the host subdomain or domain that set them, and not to any other host.
-    > They must be set with the `secure` flag, must be from a secure page (HTTPS), must not have a domain specified, and the path must be `/`.
+  > [!NOTE]
+  > Some cookie names have prefixes added that confer specific semantics. All cookie prefixes start with a double-underscore (`__`) and end in a dash (`-`). The following prefixes are defined:
```
bsmth:
```suggestion
  > Some cookie names contain prefixes that impose specific restrictions on the cookie's attributes in supporting user-agents. All cookie prefixes start with a double-underscore (`__`) and end in a dash (`-`). The following prefixes are defined:
```

files/en-us/web/http/reference/headers/set-cookie/index.md
```diff
@@ -60,21 +60,24 @@ Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnl
...
+  >
+  > **`__Http-`**: Cookies with names starting with `__Http-` must be set with the `Secure` flag by a secure page (HTTPS) and in addition must have the `HttpOnly` attribute set to prove that they were set via the `Set-Cookie` header (they can't be set or modified via JavaScript features such as `Document.cookie` or the [Cookie Store API](/en-US/docs/Web/API/Cookie_Store_API)).
+  >
+  > **`__Host-Http-`**: Cookies with names starting with `__Host-Http-` must be set with the `Secure` flag by a secure page (HTTPS) and must have the `HttpOnly` attribute set to prove that they were set via the `Set-Cookie` header. In addition, they also have the same restrictions as `__Host-`-prefixed cookies. This combination yields a cookie that is as close as can be to treating the origin as a security boundary while at the same time ensuring developers and server operators know that its scope is limited to HTTP requests.
```
bsmth: Do you want to remove this note from the DLs and have the information in the "Cookie prefixes" section only?

files/en-us/web/api/document/cookie/index.md
chrisdavidmills: updated

files/en-us/web/http/reference/headers/set-cookie/index.md
chrisdavidmills: updated

files/en-us/web/http/reference/headers/set-cookie/index.md
```diff
@@ -60,21 +60,24 @@ Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnl
...
     The percent-encoding does help to satisfy the requirements of the characters allowed for `<cookie-value>`.
 
-    > [!NOTE]
-    > Some `<cookie-name>` have a specific semantic:
-    >
-    > **`__Secure-` prefix**: Cookies with names starting with `__Secure-` (dash is part of the prefix) must be set with the `secure` flag from a secure page (HTTPS).
-    >
-    > **`__Host-` prefix**: Cookies with names starting with `__Host-` are sent only to the host subdomain or domain that set them, and not to any other host.
-    > They must be set with the `secure` flag, must be from a secure page (HTTPS), must not have a domain specified, and the path must be `/`.
+  > [!NOTE]
```
bsmth: Depends on whether we decide to keep this note block as part of the `<dd>` but the indentation of this needs to be one level higher:

```suggestion
    > [!NOTE]
```

files/en-us/web/http/reference/headers/set-cookie/index.md
chrisdavidmills: Good idea. I've put this into its own separate "Cookie prefixes" section.

files/en-us/web/http/reference/headers/set-cookie/index.md
chrisdavidmills: OK, I think my previous update has fixed this one, too.

files/en-us/web/http/reference/headers/set-cookie/index.md
```diff
@@ -205,12 +210,11 @@ Set-Cookie: sessionId=e8bb43229de9; Domain=foo.example.com
 
 ### Cookie prefixes
```
bsmth, line 211: I actually missed that this section is under "Examples", so there's some small duplication, but it's also okay in this state, IMO.

bsmth (approved):
Thanks Chris! Maybe you want to de-dupe some of the information in the Set-Cookie page for cookie prefixes, but I think it's good either way! Reiterating before the example is also fine 👍🏻

PR #40960, opened 2025-09-01
Add dashed outline to checkbox checked state

estelle:
Added a dashed outline to the checked state of the question box so that color is not the only change (currently red/green change is not visible to 10% of the population)

### DIFF
--- a/files/en-us/web/css/_colon_state/index.md
+++ b/files/en-us/web/css/_colon_state/index.md
@@ -88,6 +88,7 @@ question-box::part(checkbox) {
 
 question-box::part(checkbox):state(checked) {
   color: green;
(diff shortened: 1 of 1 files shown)


chrisdavidmills (changes requested):
LGTM, but you should also make the same change in the same example code found at https://developer.mozilla.org/en-US/docs/Web/API/CustomStateSet#matching_a_custom_state_in_a_shadow_part_of_a_custom_element, which is referenced from the page you have edited.

estelle:
see https://github.com/mdn/content/pull/40995

chrisdavidmills (approved):
Perfect, thanks. This is now GTG.

PR #40964, opened 2025-09-01
Update CSS multi-column layout module landing page

estelle:
Some properties listed are in fragmentation, so moved those down to related (and also over to fragmentation, but that is a separate PR)

added to and organized related concepts


chrisdavidmills (approved):
LGTM

PR #40905, opened 2025-08-27
Replace GH sample embeds in the remaining Learn area articles

chrisdavidmills:
### Description

This PR is part of https://github.com/mdn/mdn/issues/605, which sets out the aim of removing embedded learning-area repo examples into regular content. See https://github.com/mdn/mdn/issues/605#issuecomment-3219673324 specifically for a list of GH embed macros we want to remove.

This PR removes GH embed macros from the remaining Learn area articles, replacing them with regular live sample embeds. 

<s>Note that one GH embed I haven't replaced is the Three.js example in the "Drawing graphics" article. It includes a copy of the Three library, and I wasn't sure how best to include this. We should explore that in a separate PR at some point.</s><br>^ done

PR #40965, opened 2025-09-01
Minor update to WebGL example

chrisdavidmills:
### Description

After my previous work on the "Drawing graphics" Learn area article, I merged the PR (#40905), but forgot to push another small fix that makes the `texture.repeat.set()` line in the Three.js example consistent across the live sample code and tutorial walkthrough code.

This PR provides that fix.


bsmth (approved):
Thanks for following up 👍🏻

PR #40972, opened 2025-09-01
Ensure tooltip stays open when hovered over

### LINKED ISSUE #24244: Tooltip example fails WCAG SC 1.4.13 Content on Hover or Focus
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role

Description and example

The SC requires the tooltip to stay open, if the mouse is moved over it (hovered).

This is not mentioned in the the description’s text, and the example does not take this into account.

Some addition that the tooltip must stay open if hovered, and either a warning along with the example, or a solution for this.

https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#hoverable

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/accessibility/aria/roles/tooltip_role`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/accessibility/aria/roles/tooltip_role/index.md
* Last commit: https://github.com/mdn/content/commit/7d2071f78f38f8be38888a32c599aebba348941b
* Document last modified: 2022-09-13T12:22:39.000Z

</details>

Josh-Cena:
Fix https://github.com/mdn/content/issues/24244

### DIFF
--- a/files/en-us/web/accessibility/aria/reference/roles/tooltip_role/index.md
+++ b/files/en-us/web/accessibility/aria/reference/roles/tooltip_role/index.md
@@ -76,26 +76,34 @@ The tooltip should appear on focus or when the element is hovered on, without ad
 The tooltip can be instantiated with CSS. Change the class name with JavaScript to a class that hides the tooltip if the user hits the <kbd>Escape</kbd> key.
 
 ```css
-[role="tooltip"],
-.hide-tooltip.hide-tooltip.hide-tooltip + [role="tooltip"] {
+[role="tooltip"] {
   visibility: hidden;
   position: absolute;
   top: 2rem;
   left: 2rem;
   background: black;
   color: white;
+  padding: 0.5rem;
+  border-radius: 0.25rem;
+  /* Give some time before hiding so mouse can exit the input
+  and enter the tooltip */
+  transition: visibility 0.5s;
 }
 [aria-describedby]:hover,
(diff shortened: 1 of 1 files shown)


Shrinivassab (commented):
Thanks for the PR. 
QQ - Will this bring time based fragility? as the current solution is relies on `transition delay`.
For a more robust solution can we go with wrapping the trigger and tooltip in a shared container? something like this -> `.tooltip-container` and using `.tooltip-container:hover [role="tooltip"]` to maintain visibility? This structural approach ensures hover continuity without relying on timing, and better supports users with magnification, large pointers etc.

Josh-Cena:
We can't wrap them in a shared container. Imagine you have a second control below it. We need to position the input in the normal flow but make the tooltip absolutely positioned. Also imagine you wrap both the tooltip and the input in a box; there's a part of this box that's visually inside neither, so if your cursor stays in that region then we should hide the tooltip.

Shrinivassab:
Thanks a lot for your response and your response about the container is valid, there is a possibility it can create a layout issue. I thought if the mouse moves slower than 0.5 between input to the tooltip, we might get flicker. Thanks for the PR.

estelle (approved):

PR #40976, opened 2025-09-02
remove superflous middle

kingma-sbw:
### Description

Remove the superfluous word "middle" in the phrase "middle value". 

### Motivation

the word middle in the phrase "middle value" is confusing. The phrase "middle value" could be confused with median or average both which are not correct. Function just returns a value.

### Additional details

- [Wikipedia on Median](https://en.wikipedia.org/wiki/Median)
- [W3 calles it central ](https://www.w3.org/TR/css-values-4/#funcdef-clamp)which imho is also confusing. The resulting value must not (at all) be at the *center* of the range.

### DIFF
--- a/files/en-us/web/css/clamp/index.md
+++ b/files/en-us/web/css/clamp/index.md
@@ -6,7 +6,7 @@ browser-compat: css.types.clamp
 sidebar: cssref
 ---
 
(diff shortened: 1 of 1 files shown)


estelle (approved):
Thanks! 🎉

PR #40978, opened 2025-09-02
change flex-basis to flex-grow

kingma-sbw:
the topic of the page is `flex-grow` not `flex-basis`. The text was apparently copy-pasted

change flex-basis to flex-grow

### Description

The topic of the article is `flex-grow` not `flex-basis`. The text appears to have been copy pasted from there.

### Motivation

Confusing mention of flex-basis.


estelle (approved):
Thanks! 🎉

PR #40982, opened 2025-09-02
Update styling bio challenge and delete fundamental comprehension

chrisdavidmills:
### Description

Specifically, this PR:

- Updates the [Styling a biography page](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Styling_a_bio_page) challenge to provide a bit more of a challenge and bring the structure into line with what I have been doing in the other updated challenges.
- Removes the [Fundamental CSS comprehension](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Fundamental_CSS_comprehension) challenge. It repeats some of the previous challenge and is no longer fit for purpose — it doesn't test the right things any more, and is oddly placed in the new sequence. I will add more challenges to test the functionality that will be missing because of this removal, and have them placed better.


bsmth (approved):
Great, LGTM, thanks 👍🏻

PR #40983, opened 2025-09-02
content(Fx142): & inside at-scope blocks has no implicit specificity

### LINKED ISSUE #40981: Wrong information for the selector &
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/@scope

Specificity in @scope

The first line of this headline says

`Inside an @scope rule, both bare selectors and & behave as if :where(:scope) were prepended to the selector.`

But from the articles below

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Specificity#how_scope_blocks_affect_specificity
https://developer.chrome.com/docs/css-ui/at-scope#scope_and_specificity

They mention `& img is equivalent to :is(figure, #primary) img.`

I expected & img to be equivalent to :is(figure, #primary) img on this page https://developer.mozilla.org/en-US/docs/Web/CSS/@scope#specificity_in_scope

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Specificity#how_scope_blocks_affect_specificity
https://developer.chrome.com/docs/css-ui/at-scope

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/css/@scope`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/@scope
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/@scope/index.md
* Last commit: https://github.com/mdn/content/commit/4708192749fe1ab205f018cd78f1422aa7b25050
* Document last modified: 2025-08-28T13:44:16.000Z

</details>

bsmth:
### Description

We have another case where we're describing specificity, `@scope` and `&`. Removing the duplication and pointing to browser compat

### Motivation

We have conflicting info, specs are updated and it looks like Fx142 is the only browser to make default behavior spec-compliant.

## Related issues and pull requests

- [x] https://github.com/mdn/content/issues/40481
- [x] https://github.com/mdn/browser-compat-data/pull/27723

Fixes #40981


files/en-us/web/css/@scope/index.md
bsmth, line 1: Q: Is this appropriate here:

```yaml
browser-compat:
  - css.at-rules.scope
  - css.selectors.nesting.at-scope
```

files/en-us/web/css/@scope/index.md
```diff
@@ -166,11 +166,15 @@ Here's some considerations for `:scope` within `@scope` blocks:
...
-Because [`:where()`](/en-US/docs/Web/CSS/:where) has zero specificity, bare selectors and `&` add zero weight and only the specificity of the rest of the selector counts.
-An `& img` selector is the equivalent to writing `:where(:scope) img`.
+Inside an `@scope` rule, both bare selectors and [&](/en-US/docs/Web/CSS/Nesting_selector) behave as if `:where(:scope)` were prepended to the selector.
+Because {{cssxref(":where", ":where()")}} has zero specificity, bare selectors and `&` add zero weight and only the specificity of the rest of the selector counts.
+An `& img` selector is the equivalent to writing `:where(:scope) img` (0-0-1).
 
-In both cases in the following example, the only specificity comes from `img` (`0-0-1`):
+> [!WARNING]
+> The specificity of `&` inside `@scope` blocks is handled differently according to the browser engine and release version.
+> Check [Browser compatibility](#browser_compatibility) for details.
```
bsmth, line 177: Either on this page (comment above) or:

```suggestion
> Check [`&` browser compatibility](/en-US/docs/Web/CSS/Nesting_selector#browser_compatibility) for details.
```

files/en-us/web/css/@scope/index.md
```diff
@@ -166,11 +166,15 @@ Here's some considerations for `:scope` within `@scope` blocks:
 
 ### Specificity in `@scope`
 
-Inside an `@scope` rule, both bare selectors and `&` behave as if `:where(:scope)` were prepended to the selector.
-Because [`:where()`](/en-US/docs/Web/CSS/:where) has zero specificity, bare selectors and `&` add zero weight and only the specificity of the rest of the selector counts.
-An `& img` selector is the equivalent to writing `:where(:scope) img`.
+Inside an `@scope` rule, both bare selectors and [&](/en-US/docs/Web/CSS/Nesting_selector) behave as if `:where(:scope)` were prepended to the selector.
```
estelle:
```suggestion
Inside an `@scope` rule, both bare selectors and the [`&`](/en-US/docs/Web/CSS/Nesting_selector) nesting selector behave as if `:where(:scope)` were prepended to the selector.
```

estelle: definitely the back ticks. the rest of the edit is optional

files/en-us/web/css/@scope/index.md
```diff
@@ -166,11 +166,15 @@ Here's some considerations for `:scope` within `@scope` blocks:
...
+Because {{cssxref(":where", ":where()")}} has zero specificity, bare selectors and `&` add zero weight and only the specificity of the rest of the selector counts.
```
estelle:
```suggestion
Because {{cssxref(":where", ":where()")}} has zero [specificity](/en-US/docs/Web/CSS/CSS_cascade/Specificity), bare selectors and `&` add zero weight. The specificity weight is determined by the rest of the selector.
```

files/en-us/web/css/@scope/index.md
```diff
@@ -166,11 +166,15 @@ Here's some considerations for `:scope` within `@scope` blocks:
...
+An `& img` selector is the equivalent to writing `:where(:scope) img` (0-0-1).
```
estelle:
```suggestion
For example, the specificity of the `& img` selector is equivalent to the specificity of `:where(:scope) img` (0-0-1).
```

files/en-us/web/css/@scope/index.md
estelle: I think we should add the css.selectors.nesting.at-scope data to css..at-rules.scope, but, yes, definitely add especially if the top `&` part doesn't show, but would be better to have one table instead of two.

files/en-us/web/css/@scope/index.md
estelle: if you include the table here, we can link to that table instead of to a different page.

files/en-us/web/css/@scope/index.md
```diff
@@ -166,11 +166,15 @@ Here's some considerations for `:scope` within `@scope` blocks:
...
 
-In both cases in the following example, the only specificity comes from `img` (`0-0-1`):
+> [!WARNING]
+> The specificity of `&` inside `@scope` blocks is handled differently according to the browser engine and release version.
+> Check [Browser compatibility](#browser_compatibility) for details.
+
+In both cases in the following example, the only specificity comes from `img`:
```
estelle:
```suggestion
In both cases in the following code block, the only specificity comes from `img`:
```

optional, but as there are examples below, i try to avoid using 'example' for stuff not under that heading (and line 194)

files/en-us/web/css/@scope/index.md
```diff
@@ -186,8 +190,8 @@ In both cases in the following example, the only specificity comes from `img` (`
 }
 ```
 
-By contrast, using `:scope` explicitly selects the scope root and adds class-level specificity (`0-1-0`), since `:scope` is a pseudo-class.
-In the following example, `:scope img` has a specificity of `0-1-1`:
+By contrast, using `:scope` explicitly selects the scope root and adds class-level specificity (0-1-0), since `:scope` is a pseudo-class.
+In the following example, `:scope img` has a specificity of 0-1-1:
```
estelle: link pseudo-class (it's not linked before this in the page. you could link the previous one, but this is a better place in terms of content).
optionally change "example" to "code block"

files/en-us/web/css/css_cascade/specificity/index.md
```diff
@@ -245,17 +245,9 @@ footer a {
...
-```css
-@scope (.article-body) {
-  /* img has a specificity of 0-0-1, as expected */
-  img {
-  }
-}
-```
-
-However, if you decide to explicitly prepend the `:scope` pseudo-class to your scoped selectors, you'll need to factor it in when calculating their specificity. `:scope`, like all regular pseudo-classes, has a specificity of 0-1-0. For example:
+Including a ruleset inside a {{cssxref("@scope")}} block does not affect the specificity of its selector, regardless of the selectors used inside the scope root and limit.
```
estelle:
```suggestion
Including a ruleset inside a {{cssxref("@scope")}} block does not affect the specificity of its selector, regardless of the selectors used inside the [scope root and limit](/en-US/docs/Web/CSS/@scope#syntax).
```
i don't think most readers of this page will know either term, so let's link directly to that.

estelle (commented):
looks good. Lots of suggestions, but they're all minor (suggesting links mostly)

estelle:
@bsmth are you still reviewing this, or should i re-review?

bsmth:
> @bsmth are you still reviewing this, or should i re-review?

Thanks for checking, I'll push some of the suggestions and re-request review shortly 👍🏻

files/en-us/web/css/@scope/index.md
bsmth: Done in c9ad1f5625f05b01d505c4c3319fab266361408c

files/en-us/web/css/@scope/index.md
bsmth: Left alone here c9ad1f5625f05b01d505c4c3319fab266361408c

files/en-us/web/css/@scope/index.md
bsmth: Added the extra entry in c9ad1f5625f05b01d505c4c3319fab266361408c

>css.selectors.nesting.at-scope data to css.at-rules.scope

Do you mean we should move it in BCD or duplicate the entry to be in both places there? It's tricky to decide where it should live, my first thoughts were that it's specific to `&`.

estelle: duplicate it, as i think it's a bug in both

estelle (approved):
Thanks! 🎉

PR #40993, opened 2025-09-02
relnote(FF143): Multi-pass grid track sizing enabled by default

dipikabh:
### Description

- **Relnote update**: The multi-pass algorithm has been enabled by default in FF143, which is aligned with the algorithm defined in the spec (https://drafts.csswg.org/css-grid-2/#algo-grid-sizing). Grid tracks that use percentage row sizes will now render correctly.
- **Content updates**: Also made some copy edits to `grid-template-rows` and `grid-template-columns` pages to clarify that the calculations are handled "by the browser".
- **Firefox release notes** (for comparison and reference): [1957244#c8](https://bugzilla.mozilla.org/show_bug.cgi?id=1957244#c8)

### Additional details

The algorithm is documented on the `grid-template-rows` page for [percentage](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows#percentage) values.

Doc issue: https://github.com/mdn/content/issues/40775
Bugzilla bug: [1957244](https://bugzilla.mozilla.org/show_bug.cgi?id=1957244)

### DIFF
--- a/files/en-us/mozilla/firefox/releases/143/index.md
+++ b/files/en-us/mozilla/firefox/releases/143/index.md
@@ -28,6 +28,8 @@ Firefox 143 is the current [Beta version of Firefox](https://www.firefox.com/en-
 
 - The {{cssxref("::details-content")}} pseudo-element is now enabled by default. It lets you style the content of the {{htmlElement("details")}} element.
   ([Firefox bug 1941406](https://bugzil.la/1941406)).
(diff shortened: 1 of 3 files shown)


pepelsbey (approved):
Looks good! Thank you :)

dipikabh:
Thanks for the quick review!

Issue #40500, opened 2025-07-22
box-shadow comments in Syntax section less helpful than what they used to be


sherlysheena:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow

### What specific section or headline is this issue about?

Syntax

### What information was incorrect, unhelpful, or incomplete?

```css
/* Keyword values */
box-shadow: none;

/* A color and two length values */
/* <color> | <length> | <length> */
box-shadow: red 60px -16px;

/* Three length values and a color */
/* <length> | <length> | <length> | <color> */
box-shadow: 10px 5px 5px black;

/* Four length values and a color */
/* <length> | <length> | <length> | <length> | <color> */
box-shadow: 2px 2px 2px 1px rgb(0 0 0 / 20%);

/* inset, length values, and a color */
/* <inset> | <length> | <length> | <color> */
box-shadow: inset 5em 1em gold;

/* Any number of shadows, separated by commas */
box-shadow:
  3px 3px red inset,
  -1em 0 0.4em olive;
```

### What did you expect to see?

```css
/* Keyword values */
box-shadow: none;

/* offset-x | offset-y | color */
box-shadow: 60px -16px teal;

/* offset-x | offset-y | blur-radius | color */
box-shadow: 10px 5px 5px black;

/* offset-x | offset-y | blur-radius | spread-radius | color */
box-shadow: 2px 2px 2px 1px rgba(0, 0, 0, 0.2);

/* inset | offset-x | offset-y | color */
box-shadow: inset 5em 1em gold;

/* Any number of shadows, separated by commas */
box-shadow:
  3px 3px red,
  -1em 0 0.4em olive;
```

### Do you have any supporting links, references, or citations?

I am trying to learn HTML & CSS and I noticed that the syntax examples for `box-shadow` were updated around 2023 in this [PR](https://github.com/mdn/content/pull/29013) to use generic descriptors like `<length>` and `<color>`, instead of the earlier more explicit syntax annotations such as:

```
/* offset-x | offset-y | blur-radius | spread-radius | color */
```

What used to be:

```
/* offset-x | offset-y | blur-radius | color */
box-shadow: 10px 5px 5px black;

/* offset-x | offset-y | blur-radius | spread-radius | color */
box-shadow: 2px 2px 2px 1px rgba(0, 0, 0, 0.2);
```

Became:

```
/* Three length values and a color */
/* <length> | <length> | <length> | <color> */
[...]

PR #41005, opened 2025-09-03
 `abs()` function: update parameter and return value

estelle:
syntax copies anchor()

original made it seem like the parameter had to be a number. made it clear what the options are.

Samsung is the only non-supporting browser.


files/en-us/web/css/abs/index.md
```diff
@@ -17,17 +17,23 @@ width: abs(20% - 100px);
...
+The `abs()` function's syntax is as follows:
 
-- `x`
-  - : A calculation which resolves to a number.
+```plain
+abs( <calc-sum>)
+```
+
+The parameter is:
+
```
Josh-Cena: Is this the new convention? Because this would go against all other references that use "Parameters", which all generally only include a DL. Our page template: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types/CSS_function_page_template#parameters also does not include this part. This should probably go into the syntax box above:

```plain
/* abs( <calc-sum> ) */
abs(20% - 100px)
```

(In other news we need to remove `width:` and `property:` from the syntax)

I'm not against the convention you are proposing here, but we should document it in the page template here, which may also help resolve https://github.com/mdn/content/issues/40500, alternatively we should consider https://github.com/mdn/content/issues/40500 a prerequisite, because if we settled on the approach I proposed above, then we can simply remove all this introduction.

dipikabh: I also don't think we should add syntax to the "Parameters" section. The first and most obvious reason is that that would make this section inconsistent with our template for CSS functions. Another is that this syntax is too simplistic compared to what we show in [Formal syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/abs#formal_syntax), where each value expands into more further underlying expressions - this could become a problem if we try to apply this approach to other function pages.
(side note: not sure why "Formal syntax" is not showing up in the [preview](https://pr41005.review.mdn.allizom.net/en-US/docs/Web/CSS/abs), it's there on prod)
A third reason is that we don't follow this convention on property pages either (in the "Values" section), so adding it here would introduce more inconsistency.

[anchor()](https://developer.mozilla.org/en-US/docs/Web/CSS/anchor) is probably not the best example to follow for this. [asin()](https://developer.mozilla.org/en-US/docs/Web/CSS/asin) and [clamp()](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp) are better aligned with the template (minus the "Syntax" block).

As for the "Syntax" block, although the [template](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types/CSS_function_page_template#syntax) says:
> Only include the function itself, not a complete declaration in which it occurs....Don't end the syntax lines with semicolons: this should emphasize that we are not showing complete valid CSS code here, just the syntax usage.

Currently this is pretty inconsistent across our function pages.
I don't quite agree with this part of the CSS function template (I believe there was extensive discussion around it before settling on that convention). Since our "Syntax" blocks are really more like "example usage", I think this block should include the complete usage **_with_** the property name and `;`.

files/en-us/web/css/abs/index.md
```diff
@@ -17,17 +17,23 @@ width: abs(20% - 100px);
 
 ### Parameters
 
-The `abs(x)` function accepts only one value as its parameter.
+The `abs()` function's syntax is as follows:
```
dipikabh: we should keep the fact that it takes only one parameter

files/en-us/web/css/abs/index.md
```diff
@@ -17,17 +17,23 @@ width: abs(20% - 100px);
...
-- `x`
-  - : A calculation which resolves to a number.
+```plain
+abs( <calc-sum>)
+```
+
+The parameter is:
+
+- `<calc-sum>`
+  - : An expression, or calculation, which resolves to a {{cssxref("number")}}, {{cssxref("length")}}, {{cssxref("percentage")}} or a {{cssxref("calc-keyword").
```
dipikabh: as per the Formal syntax, length should be dimension?

```suggestion
- `<calc-sum>`
  - : An expression or calculation that resolves to a {{cssxref("number")}}, a {{cssxref("dimension")}}, a {{cssxref("percentage")}} or a {{cssxref("calc-keyword")}}.
```

files/en-us/web/css/abs/index.md
```diff
@@ -17,17 +17,23 @@ width: abs(20% - 100px);
...
 
 ### Return value
 
-The absolute value of `x`.
+The absolute value of `<calc-sum>`.
 
-- if `x`'s numeric value is positive or `0⁺`, return `x`.
-- Otherwise, returns `-1 * x`.
+- if `<calc-sum>`'s numeric value is positive or `0⁺`, return `<calc-sum>`.
+- Otherwise, returns `-1 * <calc-sum>`.
```
dipikabh:
```suggestion
- If `<calc-sum>`'s numeric value is positive or `0⁺`, the function returns `<calc-sum>`.
- Otherwise, it returns `-1 * <calc-sum>`.
```

dipikabh (changes requested):

files/en-us/web/css/abs/index.md
Josh-Cena: A function is not a property value. Writing `property: function(x);` in a syntax box assumes more context than readers can assume. Using this syntax, we cannot possibly infer that `calc(1px * abs(x))` is valid syntax, because in fact, we do have CSS functions that only work at a property-value-level and cannot be nested; currently the only case is `toggle()` which is unimplemented but specified, but this remains a real concern. In JS and Web API, our method syntax is minimal for this purpose: we don't say `const y = await someObject.doSomething(x)`, we say `doSomething(x)`, because `someObject` may be inferred as an identifier, where in fact it can be substituted by any JS value; the method may not be necessary awaited; and it may not be necessarily assigned to an identifier. we only document the syntax that readers can _absolutely expect_, not peripheral things that _may be_ composed with it.

estelle:
thank you for the conversation and review.

files/en-us/web/css/abs/index.md
```diff
@@ -17,17 +17,17 @@ width: abs(20% - 100px);
 
 ### Parameters
 
-The `abs(x)` function accepts only one value as its parameter.
+The `abs( <calc-sum> )` function accepts one value as its parameter:
```
Josh-Cena: @dipikabh can also weigh in here, but I believe the template, and the general convention on MDN about parameter format, requires this paragraph to be completely gone, or somehow incorporated into the syntax box instead.

dipikabh: We do have these one liners in the CSS property pages, when we want to spell out how the values work in combination. Strangely, we have that text outside the "Values" section (like in [overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#values)). The thing is, we haven't defined this explicitly in the template (we should, I can open a PR), but it's okay to keep that text sometimes.

In this particular case, it's probably quite obvious that the function accepts only one parameter. We can omit the sentence altogether. But if we keep it, we should use the generic function notation `abs()` or even omit mentioning it.

```suggestion
The `abs()` function accepts one parameter.
```
OR
```suggestion
The function accepts one parameter.
```

files/en-us/web/css/abs/index.md
dipikabh: I guess this is all more to do with my long-standing issue with the section heading "Syntax" 🙂. The actual syntax is presented in the "Formal syntax" section. What we really show in the "Syntax" blocks (at least in CSS) is how to use the feature. Extending the way we present the "Syntax" block for CSS properties to CSS functions, my instinct is to also show the functions in context because that's how they will be used.

We do add sections to the "Syntax" block using comments to show various types of usage, we could do the same for functions. The "Syntax" block is where we can illustrate how a function can be used:
- If a function can be nested, we show a nested example in the "Syntax" block.
- If a function cannot be nested, we show only the property-value-level example.

In JS and Web APIs, the syntax in the "Syntax" block is really a syntax and not an example usage, without any conflicting Formal syntax section - so that format makes sense there!

Let's revisit the guidelines outside this PR.

For the scope of this PR, I believe @estelle can update the Syntax block with your previous suggestion, that is:

```plain
/* abs( <calc-sum> ) */
abs(20% - 100px)
```
...sans property and semicolon...because this aligns with what we have in the template at the moment.

dipikabh (approved):
Thanks, @estelle. Only two outstanding comments need fixing - the line in "Parameters" and the "Syntax" block.

PR #41006, opened 2025-09-03
blur() function: parameters and syntax

estelle:
working on making the functions all have the right parameters and return values listed.


files/en-us/web/css/filter-function/blur/index.md
```diff
@@ -35,23 +35,25 @@ filter: blur(1.5rem);
...
 
-### Examples of blur values
+```plain
+blur( <length> )
+```
 
-```css
-blur()         /* No effect */
-blur(0)        /* No effect */
+The parameter is:
```
Josh-Cena: See https://github.com/mdn/content/pull/41005#discussion_r2320275728

estelle (commented):
this comment is a non-comment. it says i have pending review changes, but i don't see any. so committing this so i can delete whatever is happending.

files/en-us/web/css/filter-function/blur/index.md
```diff
@@ -53,6 +42,13 @@ blur(8px)      /* Blur with 8px radius */
 blur(1.17rem)  /* Blur with 1.17rem radius */
 ```
 
+### Parameters
+
+The `blur( <length> )` function's parameter is:
```
dipikabh: sentence can be removed

files/en-us/web/css/filter-function/blur/index.md
```diff
@@ -53,6 +42,13 @@ blur(8px)      /* Blur with 8px radius */
...
+
+- {{cssxref("length")}} {{Optional_Inline}}
+  - : The parameter is the radius of the blur, specified as a {{cssxref("&lt;length&gt;")}}. It defines the value of the standard deviation to the Gaussian function, i.e., how many pixels on the screen blend into each other; thus, a larger value will create more blur. A value of `0` leaves the input unchanged. The initial value for {{Glossary("interpolation")}} is `0`. Percentage values are invalid. The default value is `0px`.
```
dipikabh:
```suggestion
  - : Specifies the radius of the blur. It defines the value of the standard deviation to the Gaussian function, that is, how many pixels on the screen blend into each other. Therefore, a larger value will create more blur. A value of `0` leaves the input unchanged. The initial value for {{Glossary("interpolation")}} is `0`. Percentage values are invalid. The default value is `0`.
```

files/en-us/web/css/filter-function/blur/index.md
```diff
@@ -53,6 +42,13 @@ blur(8px)      /* Blur with 8px radius */
...
+
 ## SVG filter
```
dipikabh, line 50: We don't have a "Description" section on this page. But we could enhance this section heading to make it relate to `blur()`. WDYT about:
```suggestion
## SVG filter alternative to `blur()`
```
OR
```suggestion
## Using an SVG filter instead of `blur()`
```

Also, I don't particularly like the fact that we have code snippets unrelated to `blur()` appear before the `blur()` examples. Could we perhaps move this section under "Examples" (as an H3 - "### Using an SVG filter instead of `blur()`")?

dipikabh: And the existing example in the "Examples" section could be placed under the heading "### Comparing the effects of `blur()` and SVG filter"

files/en-us/web/css/filter-function/blur/index.md
```diff
@@ -53,6 +42,13 @@ blur(8px)      /* Blur with 8px radius */
...
```
dipikabh, line 51: In the existing [example](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/blur#examples) code, L108-110:

```plain
      <th>Live example</th>
      <th>SVG Equivalent</th>
      <th>Original image</th>
```

could we change the order to:

```plain
      <th>Original image</th>
      <th>Blurred image</th> <-- instead of "Live example"
      <th>SVG filter equivalent</th> <-- instead of "SVG Equivalent"
```

dipikabh (changes requested):
Thanks @estelle. Added a few suggestions for your consideration.

files/en-us/web/css/filter-function/blur/index.md
estelle:
```suggestion
```

files/en-us/web/css/filter-function/blur/index.md
estelle: The goal of the PR was to fix the incorrect syntax / parameter thing. I don't have the bandwidth to address reordering right now

dipikabh (approved):
Thanks a lot, Estelle!

PR #41013, opened 2025-09-04
ci(pr-review-companion): use PR number from event payload

caugner:
### Description

Updates the `pr-review-companion` workflow to use the PR number from the payload of the event that triggered the workflow run, instead of extracting it from the artifact.

### Motivation

Improves security.


LeoMcA (approved):
Nice, looks like a sensible change

PR #41019, opened 2025-09-04
Tweak LCP render time wording

tunetheweb:
### Description

I noticed some inaccuracies in the LCP render time page:
- The top of the page states 0 will be returned for cross-origin requests but this is not the case now.
- Therefore you cannot use a non-zero render time as a measure whether the LCP is accurate.
- Reword the `startTime` section as I found that difficult to understand.

I changed this in https://github.com/mdn/content/pull/37743 but missed these.

### Motivation

Correct inaccuracies to avoid confusion

Relates to: https://github.com/mdn/content/pull/37743 
Relates to: https://github.com/w3c/largest-contentful-paint/pull/138


jasonwilliams:
is https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/startTime#largest-contentful-paint still accurate?

tunetheweb:
> is https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/startTime#largest-contentful-paint still accurate?

Yes it is. In most cases now the `renderTime` will no longer be zero, but old Chromes will be. Plus we don't know yet how Safari will handle this (or any other browser engines).

Ultimately the wording there is [still specified](https://w3c.github.io/largest-contentful-paint/#ref-for-dom-largestcontentfulpaint-rendertime) and also still matches the changes proposed here which still mention it might be zero.

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -97,13 +97,9 @@ For example, to allow `https://developer.mozilla.org` to see an accurate `render
 Timing-Allow-Origin: https://developer.mozilla.org
 ```
 
-Like in the code example, you can use {{domxref("PerformanceEntry.startTime", "startTime")}}, which returns the value of the entry's {{domxref("LargestContentfulPaint.renderTime", "renderTime")}} if it is not `0`, and otherwise the value of this entry's {{domxref("LargestContentfulPaint.loadTime", "loadTime")}}. However, it is recommended to set the {{HTTPHeader("Timing-Allow-Origin")}} header so that the metrics will be more accurate.
+It is recommended to set the {{HTTPHeader("Timing-Allow-Origin")}} header so that the metrics will be more accurate.
```
wbamberg: Is this sentence needed at all? It seems to me that the preceding 2 paras cover it:

> To expose more accurate cross-origin render-time information, the {{HTTPHeader("Timing-Allow-Origin")}} HTTP response header needs to be set.
>
> For example, to allow `https://developer.mozilla.org` to see an accurate `renderTime`, the cross-origin resource should send:

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -97,13 +97,9 @@ For example, to allow `https://developer.mozilla.org` to see an accurate `render
...
 
-If you use `startTime`, you can flag any inaccuracies by checking if `renderTime` was used:
-
-```js
-const isAccurateLCP = Boolean(entry.renderTime);
-```
+Like in the code example, it is recommended to use {{domxref("PerformanceEntry.startTime", "startTime")}}, which returns the value of the entry's {{domxref("LargestContentfulPaint.renderTime", "renderTime")}} if it is not `0`, and otherwise the value of this entry's {{domxref("LargestContentfulPaint.loadTime", "loadTime")}}.
```
wbamberg: I think this is a bit strange. We make 2 recommendations here and in the previous para:

- send `Timing-Allow-Origin`
- use `startTime`

AIUI "it is recommended" to use `startTime` to get around `renderTime` not always being available for cross-origin requests in the absence of `Timing-Allow-Origin`. But if you follow the first recommendation, `Timing-Allow-Origin` is not absent, so `renderTime` is available.

Could we say:

```suggestion
If you can't set the `Timing-Allow-Origin` header, use {{domxref("PerformanceEntry.startTime", "startTime")}} instead of `renderTime`. The `startTime` property returns the value of the entry's {{domxref("LargestContentfulPaint.renderTime", "renderTime")}} if it is not `0`, and otherwise returns the value of this entry's {{domxref("LargestContentfulPaint.loadTime", "loadTime")}}.
```

?

BTW I know this isn't yours but I don't like the passive in "it is recommended to use" - IMO we would be better to say "you should" or just "use".

wbamberg: Also, I'm not going to make you do this unless you really want to but there's no way this section should be in Examples. It should be part of the description of `renderTime`.

wbamberg (commented):
I don't understand why the "### Cross-origin image render time" is duplicated. Is there a reason to not just have it in the page for `renderTime`?

files/en-us/web/api/largestcontentfulpaint/index.md
tunetheweb: Yeah I think with the removal of the previous text this is no longer needed.

Previously it was "You can get around this somewhat by using `startTime` instead of `renderTime` but we recommend to use TAO". With the first section removed the second bit is no longer needed.

```suggestion
```

files/en-us/web/api/largestcontentfulpaint/index.md
tunetheweb: OK so I see how this can  be confusing. So submitted a new commit to hopefully make this clearer.

The advice is:
- Use TAO, which should now return a non-zero `renderTime` in most (**but not all**) browsers
- Use `startTime` regardless, which is the `renderTime`, but with an automatic fallback to `loadTime` for browsers that don't support this (old Firefox and old Chrome for now, but remains to be seen if Safari will implement a coarsened time).

> Also, I'm not going to make you do this unless you really want to but there's no way this section should be in Examples. It should be part of the description of renderTime.

100% agree - moved in both places.

tunetheweb:
> I don't understand why the "### Cross-origin image render time" is duplicated. Is there a reason to not just have it in the page for renderTime?

This is a somewhat fair comment. However, I'd prefer to have it duplicated for a couple of reasons:

1. We're talking about what to use for **LCP time**. So although this nuance is in regards to `renderTime`, it affects the over all value to be used for this API.
2. We recommend kinda ignoring `renderTime` anyway. So to bury this information in that page feels kinda like... well burying it.

Does that convince you at all that duplication, while not ideal, is important enough to have this on the top page?

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -30,6 +30,24 @@ Additional key paint moments are provided by the {{domxref("PerformancePaintTimi
 
 {{InheritanceDiagram}}
 
+### Cross-origin image render time
+
+For security reasons, the value of the {{domxref("LargestContentfulPaint.renderTime", "renderTime")}} property was originally `0` if the resource is a cross-origin request. Instead the {{domxref("LargestContentfulPaint.loadTime", "loadTime")}} property should be used as a fallback. The {{domxref("PerformanceEntry.startTime", "startTime")}}
```
wbamberg: There's a sentence fragment here. I didn't know if you also wanted to keep the preceding sentence on "Instead..." - I have removed it in my suggestion because it seems to me to be covered in the next section ("Use startTime") and because I think "Browsers may now" flows better without it. But your call.

```suggestion
For security reasons, the value of the {{domxref("LargestContentfulPaint.renderTime", "renderTime")}} property was originally `0` if the resource is a cross-origin request.
```

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -30,6 +30,24 @@ Additional key paint moments are provided by the {{domxref("PerformancePaintTimi
...
+
+Browsers [may now expose a slightly coarsened render time](https://github.com/w3c/paint-timing/issues/104) in these situations. Check for [browser support](#browser_compatibility).
```
wbamberg: This link goes to the interface BCD table, which doesn't show detailed compat for `renderTime`.

wbamberg (commented):
> > I don't understand why the "### Cross-origin image render time" is duplicated. Is there a reason to not just have it in the page for renderTime?
> 
> This is a somewhat fair comment. However, I'd prefer to have it duplicated for a couple of reasons:
> 
> 1. We're talking about what to use for **LCP time**. So although this nuance is in regards to `renderTime`, it affects the over all value to be used for this API.
> 
> 2. We recommend kinda ignoring `renderTime` anyway. So to bury this information in that page feels kinda like... well burying it.
> 
> 
> Does that convince you at all that duplication, while not ideal, is important enough to have this on the top page?

This is a good point. But it's quite a complex bit of exposition and I do think duplicating it is tough on readers. As a reader I have to try to work out if this is actually exactly the same or if there are subtle differences I need to worry about. And then perhaps some maintainer will change one and not the other, and now there is a difference, and readers have to decide whether the difference is intentional or not.

My suggestion is to add a note to the interface page somewhat like:

> To measure render time for cross-origin requests, set the {{httpheader("Timing-Allow-Origin")}} header.
> 
> To ensure more accurate results for cross-origin requests across all browsers, developers should use `startTime` instead of `renderTime` as the LCP value.
>
> See [Cross-origin image render time](/en-US/docs/Web/API/LargestContentfulPaint/renderTime#cross-origin_image_render_time) and [Use startTime over renderTime](/en-US/docs/Web/API/LargestContentfulPaint/renderTime#use_starttime_over_rendertime) for more details.

But I won't block approval on this, if you disagree.

files/en-us/web/api/largestcontentfulpaint/index.md
tunetheweb: OK I've removed this whole section from the LCP page and referenced the section in the renderTime page as suggested.

So I made this suggested change in that page instead.

files/en-us/web/api/largestcontentfulpaint/index.md
tunetheweb: Yes it does under `cross-origin` under `renderTime`:

image

I'm not convinced this is obvious to developer that this is what this represents but [was asked to change it from my original suggestion of `cross-origin-render-time`](https://github.com/mdn/browser-compat-data/pull/25735#discussion_r1935569072)

tunetheweb:
OK reworked it to have a lighter explanation on the LCP page and link to the heavier explanation on the `renderTime` page. PTAL!

files/en-us/web/api/largestcontentfulpaint/index.md
wbamberg: Oh yes you're right, sorry.

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -30,6 +30,12 @@ Additional key paint moments are provided by the {{domxref("PerformancePaintTimi
 
 {{InheritanceDiagram}}
 
+For accurate measurement cross-origin render time, set the {{httpheader("Timing-Allow-Origin")}} header.
+
+Developers should use `startTime` instead of `renderTime` as the LCP value, as the `renderTime` may not be set in some browsers.
+
+See [Cross-origin image render time](./renderTime#cross-origin_image_render_time) and [Use startTime over renderTime](./renderTime#use_starttime_over_rendertime) for more details.
```
wbamberg:
```suggestion
See [Cross-origin image render time](/en-US/docs/Web/API/LargestContentfulPaint/renderTime#cross-origin_image_render_time) and [Use startTime over renderTime](/en-US/docs/Web/API/LargestContentfulPaint/renderTime#use_starttime_over_rendertime) for more details.
```

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -30,6 +30,12 @@ Additional key paint moments are provided by the {{domxref("PerformancePaintTimi
 
 {{InheritanceDiagram}}
 
+For accurate measurement cross-origin render time, set the {{httpheader("Timing-Allow-Origin")}} header.
```
wbamberg:
```suggestion
To get an accurate measurement of render time for cross-origin resources, set the {{httpheader("Timing-Allow-Origin")}} header.
```

wbamberg (approved):
👍 thanks for your patience @tunetheweb !

PR #41023, opened 2025-09-04
add values, backgrounds & borders and sizing challenge

chrisdavidmills:
### Description

This PR specifically:

- Adds a new challenge that tests values, backgrounds & borders, and sizing.
- Removes the [Creating fancy letterheaded paper](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Fancy_letterheaded_paper) Challenge, which is rather old and not well-placed in the new structure.

One thing to note — the finished live sample in the new challenge is not rendering locally. I am not sure if I have done something stupid, or if this is a bug in Fred/Rari, but we need to keep an eye on what is going on here.
UPDATE: It works in the preview render, so it looks like it is a local rendering issue. I'll file a fred issue.


files/en-us/learn_web_development/core/styling_basics/size_decorate_content_panel/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+{{PreviousMenuNext("Learn_web_development/Core/Styling_basics/Test_your_skills/Backgrounds_and_borders", "Learn_web_development/Core/Styling_basics/Overflow", "Learn_web_development/Core/Styling_basics")}}
+
+In this challenge you are provided with a lightly-styled page structure that renders a content panel containing text and images, with a heading at the top and a button bar at the bottom. We want you to follow the instructions to size and decorate it, producing an interesting layout as a result. Along the way, we'll test your knowledge of CSS values and units, sizing, and background and borders.
+
+## Starting point
+
+We are going to get you to solve this challenge on your local development environment; ideally, you'll want to view the example in a full browser window to make sure you are going in the right direction.
+
+1. Create a new folder on your local hard drive called `size-decorate-content-panel`.
```
bsmth:
```suggestion
1. Create a new folder on your computer called `size-decorate-content-panel`.
```

files/en-us/learn_web_development/core/styling_basics/size_decorate_content_panel/index.md
```diff
@@ -0,0 +1,386 @@
...
+The finished project should look like this:
+
+{{EmbedLiveSample("content-pane-finish", "100%", 500)}}
+
+<details>
+<summary>Click here to show the solution</summary>
+
+The finished CSS looks like so:
+
+```html hidden live-sample___content-pane-finish
```
bsmth, line 202: Is it the same HTML as above? Can we remove this one and include the other one directly?

bsmth (approved):
Two comments, but a +1 from me, thanks!

files/en-us/learn_web_development/core/styling_basics/size_decorate_content_panel/index.md
chrisdavidmills: It is, but I wasn't sure if it was a good idea to include the full file including doctype, `<html>` element, etc., inside a live sample. I'll merge it for now as-is, and then we can always remove and simplify later if we think it is Ok.

Issue #30968, opened 2023-12-12
In the "Using custom elements" guide and related pages, add a disclaimer that WebKit does not plan to support Customized built-in elements


njanssoone:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements

### What specific section or headline is this issue about?

Types of custom element

### What information was incorrect, unhelpful, or incomplete?

The guide does not warn users that using Customized Built-in Elements currently does not work on WebKit-based browsers, and that it won't change for now because the WebKit team is refuting this part of the Web Components spec.

Also, on the ["CustomElementRegistry.prototype.define()" page](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define), the compatibility section does show the `define` method as fully supported by Safari (both desktop and mobile).

### What did you expect to see?

In the "Using custom elements" guide, part "Types of custom element": a disclaimer panel indicating that this type of custom elements is currently not supported by WebKit-based browsers.

In the "CustomElementRegistry.prototype.define()" page, part "Browser compatibility": a dedicated row for the `{ extend: 'tagName' }` option's compatibility, indicating WebKit's missing support.

### Do you have any supporting links, references, or citations?

https://bugs.webkit.org/show_bug.cgi?id=182671
https://github.com/WebKit/standards-positions/issues/97
https://github.com/WICG/webcomponents/issues/509

pepelsbey:
Hey! Nice catch :) I just stumbled upon the same part of the guide, but there’s already an issue for that. Thanks! This part of the guide would definitely benefit from clearing this out. It might be in the form of a note, linking to the WebKit’s standard position. Feel free to suggest a PR, or we can wait for the community to pick it up.

As for the compatibility table, this is something that’s based on [BCD repository](https://github.com/mdn/browser-compat-data/), and there’s a link on top of the table to file an issue with the data. Could you please file a feature request for documenting the `extend` support?

davidpcaldwell:
I filed my own duplicate of this that I'll link for reference. https://github.com/mdn/content/issues/33063

Here's the gist of what I wrote:

### What information was incorrect, unhelpful, or incomplete?

It seems like the page should mention somewhere that (as things stand) "customized built-in elements" will never be widely supported because [Safari has decided not to support them](https://bugs.webkit.org/show_bug.cgi?id=182671) on iOS or macOS.

### What did you expect to see?

[...]

PR #41036, opened 2025-09-05
Disambiguate between custom elements and customized built-in elements

kurowski:
### Description

Clarify that Safari does not intend to support custom**ized** built-in elements, leaving it clear that Safari _does_ support "custom" elements.

### Motivation

This note was unclear to me on first reading.

### Additional details

- [HTML Standard § 4.13 Custom elements](https://html.spec.whatwg.org/multipage/custom-elements.html) is consistent in its use of "custom" vs "customized" when referring to autonomous vs built-in elements, respectively.

- Relates to #30968

### DIFF
--- a/files/en-us/web/api/document/createelement/index.md
+++ b/files/en-us/web/api/document/createelement/index.md
@@ -83,7 +83,7 @@ function addElement() {
 ### Web component example
 
 > [!NOTE]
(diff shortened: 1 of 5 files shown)


Josh-Cena:
Thanks. If you search for "custom built-in", we have a lot of other occurrences. Could you update those as well?

Josh-Cena (approved):
I've fixed the rest. Thanks for pointing this out!

PR #41039, opened 2025-09-06
adding example-bad to code that shows a mistake

NookieGrey:
### Description

Just added example-bad to a block of code that shows an example of how not to write. The article below explains that this piece of code will not perform the expected behavior.

### Motivation

I added this parameter because I thought it would be more clear that this is not the code the developer expects.

When reading quickly, the user has the opportunity to copy the wrong piece of code and get the wrong result.

### Additional details

-

-


Josh-Cena (approved):
Sure, thanks

PR #41043, opened 2025-09-06
Fix AbortController being reused in multiple fetch requests in some o…

marcindz88:
### Description

The AbortController [examples](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/AbortController#examples) include code that reused the same AbortController instance across multiple fetch calls, which causes them to be instantly cancelled.

### Motivation

The documentation should be updated to help developers understand that an AbortController instance cannot be reused.

### DIFF
--- a/files/en-us/web/api/abortsignal/index.md
+++ b/files/en-us/web/api/abortsignal/index.md
@@ -56,7 +56,7 @@ The following snippet shows how we might use a signal to abort downloading a vid
 
 We first define a variable for our `AbortController`.
 
-Before each [fetch request](/en-US/docs/Web/API/Window/fetch) we create new instance of controller using the {{domxref("AbortController.AbortController","AbortController()")}} constructor, then grab a reference to its associated {{domxref("AbortSignal")}} object using the {{domxref("AbortController.signal")}} property.
+Before each [fetch request](/en-US/docs/Web/API/Window/fetch) we create a new controller using the {{domxref("AbortController.AbortController","AbortController()")}} constructor, then grab a reference to its associated {{domxref("AbortSignal")}} object using the {{domxref("AbortController.signal")}} property.
 
 > [!NOTE]
 > An `AbortSignal` can only be used once. After it is aborted, any fetch call using the same signal will be immediately rejected.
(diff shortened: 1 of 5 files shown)


wbamberg (changes requested):
Thank you for the fix, @marcindz88 ! The code looks good to me but the description needs tweaking a bit, especially the paragraph starting "We first create a controller...", because we don't, any more :).

marcindz88:
@wbamberg Do you think we could leave the example on the [AbortController page](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) and include only a link in the methods and constructor to not repeat this or is is a common practice here to include it on all pages for easier accessibility ?

wbamberg:
> @wbamberg Do you think we could leave the example on the [AbortController page](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) and include only a link in the methods and constructor to not repeat this

Yes, I think this is a good idea. I'm not really a fan of this style where we duplicate examples, it's such a maintenance burden.

files/en-us/web/api/abortsignal/index.md
```diff
@@ -54,10 +54,14 @@ Listen to this event using {{domxref("EventTarget.addEventListener", "addEventLi
...
 
-We first create an abort controller using the {{domxref("AbortController.AbortController","AbortController()")}} constructor, then grab a reference to its associated `AbortSignal` object using the {{domxref("AbortController.signal")}} property.
+We first define a variable for our `AbortController`.
 
-When the [fetch request](/en-US/docs/Web/API/Window/fetch) is initiated, we pass in the `AbortSignal` as an option inside the request's options object (the `{signal}` below). This associates the signal and controller with the fetch request, and allows us to abort it by calling {{domxref("AbortController.abort()")}}.
-Below you can see that the fetch operation is aborted in the second event listener, which triggered when the abort button (`abortBtn`) is clicked.
+Before each [fetch request](/en-US/docs/Web/API/Window/fetch) we create new instance of controller using the {{domxref("AbortController.AbortController","AbortController()")}} constructor, then grab a reference to its associated {{domxref("AbortSignal")}} object using the {{domxref("AbortController.signal")}} property.
```
wbamberg:
```suggestion
Before each [fetch request](/en-US/docs/Web/API/Window/fetch) we create a new controller using the {{domxref("AbortController.AbortController","AbortController()")}} constructor, then grab a reference to its associated {{domxref("AbortSignal")}} object using the {{domxref("AbortController.signal")}} property.
```

wbamberg (approved):
👍 thanks for the fix @marcindz88 !

PR #41044, opened 2025-09-06
Update index.md

Zearin:
Use prime symbol (`′`) instead of apostrophe (`'`).

### Description
N/A


pepelsbey:
Hey! Thank you for the suggestion. I know it kinda makes sense to use the prime symbol here, but in our [writing guidelines](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide#apostrophes_and_quotation_marks), we agreed not to use typographic symbols that might get in the way of writing code.

PR #41052, opened 2025-09-07
Safari also supports SVG WebExtension icons

stevennyman:
### Description

Safari also supports SVG WebExtension icons so the note should be updated to include it.

### Motivation

The page currently contradicts itself: the note about SVG WebExtension icon support says the feature is only available in Firefox, while the compatibility notes say support was added to Safari a couple of months later.

### Additional details

The compatibility notes state the feature is available in Safari: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/icons#browser_compatibility

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/manifest.json/icons/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/manifest.json/icons/index.md
@@ -70,7 +70,7 @@ You can use SVG, and the browser scales your icon appropriately. There are two c
    ```
 
 > [!NOTE]
-> Only Firefox and Safari are known to support SVG icons. Chromium has a bug about [unsupported SVG icons](https://crbug.com/29683).
+> Chromium-based browsers don't support this feature. See [Chromium bug 29683](https://crbug.com/29683).
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


files/en-us/mozilla/add-ons/webextensions/manifest.json/icons/index.md
```diff
@@ -70,7 +70,7 @@ You can use SVG, and the browser scales your icon appropriately. There are two c
    ```
 
 > [!NOTE]
-> Only Firefox is known to support SVG icons. Chromium has a bug about [unsupported SVG icons](https://crbug.com/29683).
+> Only Firefox and Safari are known to support SVG icons. Chromium has a bug about [unsupported SVG icons](https://crbug.com/29683).
```
rebloor: As it's only the Chromium-based browsers that don't have this feature, I suggest moving the note to the beginning of the section and simplifying:
```suggestion
> Chromium-based browsers don't support this feature. See [Chromium bug 29683](https://crbug.com/29683).
```

rebloor (approved):
@stevennyman sorry for the delay in getting back to this one

PR #41054, opened 2025-09-08
chore: align the output of the RangeError caused by invalid language tag

yin1999:
### Description

align the output of the RangeError caused by invalid language tag

### Motivation

The output in example and `InteractiveExample` is not the same, this pr align the output of them.


Josh-Cena (approved):
Thanks

PR #41056, opened 2025-09-08
fix: `::columns` -> `::column`

yamanoku:
### Description

This PR fixes a pseudo-element mismatch (`::columns` → `::column`) found in the documentation.


Josh-Cena (approved):
Good catch, thanks

PR #41058, opened 2025-09-08
Fix variable name

Scratchcat1:
### Description

`report` isn't defined, it should be `stat`

### Motivation

Fixes incorrect script

### Additional details

N/A


Josh-Cena (approved):
Good catch, thanks

PR #41059, opened 2025-09-08
Fixed demo: textContent is not a function

benhatsor:
### Description

Fixed the demo: textContent is not a function, and running the code throws a SyntaxError. Changed it to a property.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent


Josh-Cena (approved):
Good catch, thanks

PR #41061, opened 2025-09-08
Add House data UI challenge

chrisdavidmills:
### Description

This PR specifically adds a new challenge near the end of the Scripting module that tests the reader on JSON, network requests, DOM scripting, functions, loops, conditionals, arrays, and probably some other topics besides. I was originally intending it to appear earlier, but it ended up being a bit more complicated than I first envisaged.


files/en-us/learn_web_development/core/scripting/house_data_ui/index.md
```diff
@@ -0,0 +1,338 @@
...
+    });
+}
+
+function initializeForm() {
+  // Create options for all the different street names
+  const streetArray = [];
+  for (let house of houses) {
+    if (!streetArray.includes(house.street)) {
+      streetArray.push(house.street);
+      streetSelect.innerHTML += `<option>${house.street}</option>`;
```
bsmth: Should we avoid `innerHTML`? We're fetching the JSON and (in this scenario), we probably don't control the resource. So it could end up being something nasty:

```json
"street": "</option><script>alert('Hello World!');</script><option>thing"
```

Would it be better to do DOM methods like:

```html
thing.appendChild(document.createElement("option")).textContent = house.street;
```

Based on https://redirect.github.com/mdn/mdn/issues/567

chrisdavidmills: Good call. I've replaced the instances that need replacing and added a note telling the reader not to use `innerHTML`.

bsmth: Looks great, tnx

bsmth (approved):
Great, thanks for the updates. This is a challenging one 👍🏻

PR #41066, opened 2025-09-08
Changed "search term" to "username"

0yqc:
Small change for a small error I noticed:

The example for ### Making input required shows a form where one should choose a username, however the explanation below said search term.

### Description

The example for ### Making input required shows a form where one should choose a username, however the explanation below said search term.

### Motivation

Mainly new users could be confused as to what the text is referring to, so I edited it to be the correct form. Additionally, I have never contributed to an open-source project, so wanted to try it out with this basic edit.


estelle (approved):
Thanks! 🎉

PR #41067, opened 2025-09-08
Avoid usage of substr

Farzane2630:
### Description
The substr() method is removed (deprecated) in the latest JavaScript standard.

We should use substring() or slice() instead.

I replaced it with slice method.

### Motivation

I did this to avoid making confusions.


Josh-Cena (approved):
Thanks, it makes sense to me to replace usage of `substr` globally. I just did it for the other pages.

PR #41069, opened 2025-09-08
Mention that custom elements constructors can't be directly called

### LINKED ISSUE #41057: [Improvement] Add the required customElements.define() call directly after the very first snippet.
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements

Implementing a custom element

Please please, add

```
customElements.define("word-count-element", WordCount);  // Your custom element class won't work until this call
```

directly after the very first snippet of how to extend `HTMLElement`, since your code will never work without this. and it took me days to figure out why `super()` yields illegal constructor.

```js
class WordCount extends HTMLParagraphElement {
  constructor() {
    super();
  }
  // Element functionality written in here
}

customElements.define("word-count-element", WordCount);  // Your custom element class won't work until this call
```

HTML***Element classes are interfaces, you can not just simply extend them.

At least make a clear statement somewhere else, so users can google and find out why it won't work and stop getting non-sense errors.

During coding, I encounter these, and have no idea how to resolve them, every search result simply tells me to `extends` the class directly:

* HTML***Element (squiggles): you cannot extend an interface
* Uncaught TypeError: Illegal constructor

Josh-Cena:
Fix https://github.com/mdn/content/issues/41057.

At this point we haven't introduced registration yet, so code shouldn't use it, but it makes sense to point out that the most obvious way is not correct.

### DIFF
--- a/files/en-us/web/api/web_components/using_custom_elements/index.md
+++ b/files/en-us/web/api/web_components/using_custom_elements/index.md
@@ -49,6 +49,9 @@ class PopupInfo extends HTMLElement {
 
 In the class [constructor](/en-US/docs/Web/JavaScript/Reference/Classes/constructor), you can set up initial state and default values, register event listeners and perhaps create a shadow root. At this point, you should not inspect the element's attributes or children, or add new attributes or children. See [Requirements for custom element constructors and reactions](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-conformance) for the complete set of requirements.
 
+> [!NOTE]
+> This class is not meant to be directly constructed using `new`. If you do that, you will get an error like `TypeError: Failed to construct 'HTMLParagraphElement': Illegal constructor`. Instead, you must [register](#registering_a_custom_element) the custom element, and then create instances of it using standard DOM practices, such as writing the element in HTML markup, calling {{domxref("document.createElement()")}}, etc.
+
 ### Custom element lifecycle callbacks
 
 Once your custom element is registered, the browser will call certain methods of your class when code in the page interacts with your custom element in certain ways. By providing an implementation of these methods, which the specification calls _lifecycle callbacks_, you can run code in response to these events.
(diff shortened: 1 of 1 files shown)


wbamberg:
I'm not that keen on a documentation style where we add little hyper-specific notes to deal with misconceptions people get when they only read one section of the guide, and then guess what to do next. There are lots of ways people can get confused, and if we try to note them all we end up in a mess.

I would prefer a fix where we have a preamble explaining what we're going to cover, something like:

***

We're going to implement and use custom elements in three main steps:

- [Implementing a custom element](): creating a class that defines the behavior of the custom element.
- [Registering a custom element](): before a custom element can be used in a page, you need to register it, which maps its name to the constructor for the element's class
- [Using a custom element](): after registering the custom element, you can use in a page by including it in your HTML markup or dynamically creating it using APIs like `createElement()`.

***

I don't know if this would help much though.

Josh-Cena:
For me, I think this note is well deserved, because we do have certain HTML elements that are constructible, such as `Image` and `Option`, and in general it is a bit confusing to create a class but not call it. I think your suggestion can be implemented _in addition_ but not as a replacement of the current note. In addition this information is already sort of in the TOC.

imkzh:
Hi,

The current guide omits an important point: you cannot instantiate (via `new`) a custom element until it has been registered with `customElements.define()`. It also does not clearly explain why registration is necessary.

The documentation states:

> "To make a custom element available in a page, call the define() method of Window.customElements."

However, this phrasing implies that registration is only about making an element “available in a page.” In reality, many native elements (like `<canvas>`, ``, and `<video>`) can be created and used off-DOM. The issue is not about presence in the page — it's that the element constructor itself is not fully functional until the element is defined.

The current tone makes it sound like registration is optional — a “bonus” that enables custom tags in HTML. But in fact, it is mandatory. Without registration, you cannot even create an instance of your custom element using `new MyElement()`.

It should be explicitly stated that `customElements.define()` is required for the custom element to work at all. At this point, we could also briefly note that this is because custom element classes rely on the built-in mechanism that connects the class to the HTML element interface — a connection provided specifically by the `define()` method, which supplies the necessary implementation to complete the element's definition.

Josh-Cena:
@wbamberg I tried to incorporate this text into existing content, so it's written in a more affirmative fashion and doesn't look like a vestigial thing.

wbamberg (approved):
Thanks for the update, I'm much happier with this.

PR #41079, opened 2025-09-09
Remove onload and DOMContentLoaded handlers

Josh-Cena:
Listening to `DOMContentLoaded` or `load` for manipulating DOM is mostly an unnecessary practice. Usually, one can safely modify DOM simply by placing the script at the end of the file. It does avoid blocking render, but `async` does that job as well. However, the most common context for using this trick is for placing the script in `<head>`, which is a bad practice to begin with (since if you want to defer execution anyway you might as well defer parsing). It's definitely useless in our live sample / interactive example system, and most examples don't do this anyway, so for consistency I'm removing the rest.

(Tip: review this PR with hide whitespace.)

### DIFF
--- a/files/en-us/web/api/window/load_event/index.md
+++ b/files/en-us/web/api/window/load_event/index.md
@@ -19,6 +19,8 @@ This event is not cancelable and does not bubble.
 > [!NOTE]
 > The `load` event that is dispatched when the main document has loaded _is_ dispatched on the `window`, but has two mutated properties: `target` is `document`, and `path` is `undefined`. These two properties are mutated due to legacy conformance.
 
+Usually, to avoid running a script before the DOM it manipulates has been fully constructed, you can simply place the script at the end of the document body, immediately before the closing `</body>` tag, without wrapping it in an event listener. Use this event only to wait for external resources, such as images, to load.
+
 ## Syntax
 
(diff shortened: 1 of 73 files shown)


files/en-us/web/api/window/load_event/index.md
```diff
@@ -19,6 +19,8 @@ This event is not cancelable and does not bubble.
 > [!NOTE]
 > The `load` event that is dispatched when the main document has loaded _is_ dispatched on the `window`, but has two mutated properties: `target` is `document`, and `path` is `undefined`. These two properties are mutated due to legacy conformance.
 
+Usually, to avoid running a script before the DOM it manipulates has been fully constructed, you can simply place the script at the end of the document body, immediately before the closing `</body>` tag, without wrapping it in an event listener. Use this event only to wait for external resources, such as images, to load.
```
bsmth:
```suggestion
To avoid running a script before the DOM it manipulates has been fully constructed, you can place the script at the end of the document body, immediately before the closing `</body>` tag, without wrapping it in an event listener. You should usually only use the `load` event to wait for external resources, such as images or deferred scripts, to load.
```

bsmth (approved):
Great cleanup, thank you 👍🏻

PR #41081, opened 2025-09-09
Synchronize with BCD v7.1.4

OnkarRuikar:
[BCD release v7.1.4](https://github.com/mdn/browser-compat-data/releases/tag/v7.1.4)

/cc cc/ @bsmth @wbamberg

### DIFF
--- a/files/en-us/web/http/reference/headers/supports-loading-mode/index.md
+++ b/files/en-us/web/http/reference/headers/supports-loading-mode/index.md
@@ -45,7 +45,7 @@ The `Supports-Loading-Mode` header value is a list of one or more tokens, which 
 
 - `credentialed-prerender` {{experimental_inline}} {{non-standard_inline}}
   - : Indicates that a destination origin opts in to loading documents via cross-origin, same-site [prerendering](/en-US/docs/Web/API/Speculation_Rules_API#using_prerendering).
-- `fenced-frame`
+- `fenced-frame` {{experimental_inline}} {{non-standard_inline}}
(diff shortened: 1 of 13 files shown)


files/en-us/web/http/reference/headers/supports-loading-mode/index.md
```diff
@@ -42,7 +43,7 @@ Supports-Loading-Mode: <client-hint-headers>
 
 The `Supports-Loading-Mode` header value is a list of one or more tokens, which can include the following values:
 
-- `credentialed-prerender` {{experimental_inline}}
+- `credentialed-prerender` {{experimental_inline}} {{non-standard_inline}}
   - : Indicates that a destination origin opts in to loading documents via cross-origin, same-site [prerendering](/en-US/docs/Web/API/Speculation_Rules_API#using_prerendering).
 - `fenced-frame`
```
Josh-Cena: This is interesting. Is BCD wrong or is content wrong? BCD spells it as `fenced-frames`. I can find neither in the spec. Anyway I think it's very confusing for its status to be different from the header as a whole, and you may need to manually encode it as an exception.

OnkarRuikar: BCD has a typo: https://github.com/mdn/browser-compat-data/pull/27811
```suggestion
- `fenced-frame` {{experimental_inline}} {{non-standard_inline}}
```

Josh-Cena (approved):
Nice, thank you!

PR #41085, opened 2025-09-10
changed 'yield' to 'yieldFunc'

addsoupbase:
'yield' is a reserved word in strict mode and might be confusing

### Description

changed a function named 'yield' (a reserved word) to 'yieldFunc'


chrisdavidmills (approved):
Makes sense. Thanks, @addsoupbase!

PR #41086, opened 2025-09-10
40778 color input value

dletorey:
### Description

- Added pages for:
  - `alpha` attribute
  - `colorspace` attribute
- Added these attributes to:
  - the attributes input table
  - the attribute list
- On color input page:
  - updated the BCD data
  - updated the See Also list

### Motivation

- Working on [MDN issue #40778](https://github.com/mdn/content/issues/40778)

- [BCD PR](https://github.com/mdn/browser-compat-data/pull/27814)
- [FF Release PR](https://github.com/mdn/content/pull/41087)


Josh-Cena:
Hello, FYI we don't have an established convention yet, but it is an active discussion about what `HTML/Reference/Attribute` should contain: https://github.com/orgs/mdn/discussions/688. I think generally people agree that a page should only be created here in one of two cases: the attribute is shared between many elements, OR significant explanation/examples are required for its behavior (perhaps it is even an AND). I'm not sure either condition is met in this case (case in point: both of the pages you wrote are quite short). For these short pages, they don't add much beyond what's already present in the attributes DL on the element page, so creating another page adds close to zero value. This was why https://github.com/mdn/content/pull/40260 did not write these pages, and I believe there's really no reason to put in further effort beyond what's already in that PR.

files/en-us/web/html/reference/attributes/alpha/index.md
```diff
@@ -0,0 +1,50 @@
+---
+title: "HTML attribute: alpha"
+short-title: alpha
+slug: Web/HTML/Reference/Attributes/alpha
+page-type: html-attribute
+browser-compat: html.elements.input.alpha
+sidebar: htmlsidebar
+---
+
+The Boolean **`alpha`** attribute, valid only on the [`type="color"`](/en-US/docs/Web/HTML/Reference/Elements/input/color) {{HTMLElement("input")}} element, allows the end user to choose the opacity of the color being selected.
```
bsmth: How's this?

```suggestion
The Boolean **`alpha`** attribute, valid on [`<input type="color"/>`](/en-US/docs/Web/HTML/Reference/Elements/input/color) elements, allows the user to choose the opacity of the color being selected.
```

I used this format before here:

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/color-scheme#usage_notes

> A `<meta name="color-scheme">` element…

files/en-us/web/html/reference/attributes/alpha/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+<input type="color" id="no-alpha" name="no-alpha" value="#ff0000" />
+
+<label for="alpha">Color input using alpha:</label>
+
+<input
+  type="color"
+  id="alpha"
+  name="alpha"
+  alpha
```
bsmth: what should I be seeing here in the demo? Not sure what to look for

dletorey: This only works in Safari at present, when you click the input > select show colors > then you see an opacity field
Screenshot 2025-09-11 at 11 58 41

bsmth:
> Hello, FYI we don't have an established convention yet, but it is an active discussion about what `HTML/Reference/Attribute` should contain: https://github.com/orgs/mdn/discussions/688. I think generally people agree that a page should only be created here in one of two cases: the attribute is shared between many elements, OR significant explanation/examples are required for its behavior…

@dletorey I think this is a good point; details are worth mentioning here https://pr41086.review.mdn.allizom.net/en-US/docs/Web/HTML/Reference/Elements/input/color on `<input type="color"/>` but maybe the individual pages are not necessary in their current form.

bsmth (approved):
LGTM, thanks!

PR #41089, opened 2025-09-10
Change location type from string to number

nikneym:
### Description

Fixes the `location` argument type for `KeyboardEvent`.


wbamberg (approved):
👍 thank you!

PR #41091, opened 2025-09-10
The form attribute: link to the form ref page

estelle:
In the list of attributes on the form control pages, the form attribute itself links to the form attribute page, so might as well go there directly,.


wbamberg (approved):
👍

PR #41095, opened 2025-09-10
Document fetching a range of a blob

### LINKED ISSUE #24625: Fetch: blob: URL range requests
As of https://github.com/whatwg/fetch/pull/1520 `blob:` URLs now support `Range` headers. This impacts both `fetch()` and `XMLHttpRequest`.

cc @dlrobertson

Josh-Cena:
Fix https://github.com/mdn/content/issues/24625. Although this was added to the `fetch()`/`XMLHttpRequest` machinery, we don't document anything about the request specifics in `fetch()`. So instead, I'm just documenting it in the two relevant places.

### DIFF
--- a/files/en-us/web/http/reference/headers/range/index.md
+++ b/files/en-us/web/http/reference/headers/range/index.md
@@ -135,6 +135,22 @@ Content-Type: image/jpeg
 Accept-Ranges: bytes
 ```
 
+### Fetching a range from a blob URL
+
+The [`blob:`](/en-US/docs/Web/URI/Reference/Schemes/blob) URL also supports range requests by using [`fetch()`](/en-US/docs/Web/API/Window/fetch).
+
+```js
+const blob = new Blob(["Hello, world!"], { type: "text/plain" });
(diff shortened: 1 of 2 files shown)


files/en-us/web/http/reference/headers/range/index.md
```diff
@@ -135,6 +135,22 @@ Content-Type: image/jpeg
...
+### Fetching a range from a blob URL
+
+The [`blob:`](/en-US/docs/Web/URI/Reference/Schemes/blob) URL also supports range requests by using [`fetch()`](/en-US/docs/Web/API/Window/fetch).
+
+```js
+const blob = new Blob(["Hello, world!"], { type: "text/plain" });
+const url = URL.createObjectURL(blob);
+fetch(url, {
+  headers: {
+    Range: "bytes=7-11",
```
bsmth, line 147: Nice 👍🏻

bsmth (approved):
Interesting example, thanks 👍🏻

PR #41097, opened 2025-09-10
Bug-1986688 Cookies expirationDate in milliseconds

rebloor:
### Description

Addresses the dev-docs-needed requirement of [Bug 1986688](https://bugzilla.mozilla.org/show_bug.cgi?id=1986688) Cookies are returned with a fractional expirationDate for the change made in [Bug 1972757](https://bugzilla.mozilla.org/show_bug.cgi?id=1972757) Store cookie expiry value as msec.

Related BCD changes in TBC.

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
@@ -15,7 +15,7 @@ Values of this type are objects that can contain these properties:
 - `domain`
   - : A `string` representing the domain the cookie belongs to (e.g., "www.google.com" or "example.com").
 - `expirationDate` {{optional_inline}}
-  - : A `number` representing the expiration date of the cookie as the seconds since the UNIX epoch. From Firefox 142 includes millisecond in the fractional part. Not provided for session cookies.
+  - : A `number` representing the expiration date of the cookie as the seconds after the UNIX epoch. From Firefox 142 includes milliseconds in the fractional part. Not provided for session cookies.
 - `firstPartyDomain`
   - : A `string` representing the first-party domain associated with the cookie. This is an empty string if the cookie was set while first-party isolation was off. See [First-party isolation](/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies#first-party_isolation).
 - `hostOnly`
--- a/files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
@@ -15,7 +15,7 @@ Values of this type are objects that can contain these properties:
 - `domain`
   - : A `string` representing the domain the cookie belongs to (e.g., "www.google.com" or "example.com").
 - `expirationDate` {{optional_inline}}
-  - : A `number` representing the expiration date of the cookie as the seconds since the UNIX epoch. From Firefox 142 includes millisecond. Not provided for session cookies.
+  - : A `number` representing the expiration date of the cookie as the seconds since the UNIX epoch. From Firefox 142 includes millisecond in the fractional part. Not provided for session cookies.
 - `firstPartyDomain`
   - : A `string` representing the first-party domain associated with the cookie. This is an empty string if the cookie was set while first-party isolation was off. See [First-party isolation](/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies#first-party_isolation).
 - `hostOnly`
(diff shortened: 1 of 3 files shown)


files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
```diff
@@ -15,7 +15,7 @@ Values of this type are objects that can contain these properties:
 - `domain`
   - : A `string` representing the domain the cookie belongs to (e.g., "www.google.com" or "example.com").
 - `expirationDate` {{optional_inline}}
-  - : A `number` representing the expiration date of the cookie as the number of seconds since the UNIX epoch. Not provided for session cookies.
+  - : A `number` representing the expiration date of the cookie as the seconds since the UNIX epoch. From Firefox 142 includes millisecond in the fractional part. Not provided for session cookies.
```
vadcx: suggest to use plural millisecond**s**

files/en-us/mozilla/add-ons/webextensions/api/cookies/set/index.md
```diff
@@ -27,7 +27,7 @@ let setting = browser.cookies.set(
     - `domain` {{optional_inline}}
       - : A `string` representing the domain of the cookie. If omitted, the cookie becomes a host-only cookie.
     - `expirationDate` {{optional_inline}}
-      - : A `number` that represents the expiration date of the cookie as the number of seconds since the UNIX epoch. If omitted, the cookie becomes a session cookie.
+      - : A `number` that represents the expiration date of the cookie as the seconds since the UNIX epoch. From Firefox 142 can include millisecond in the fractional part. If omitted, the cookie becomes a session cookie.
```
vadcx: use **may** instead of "can" as it's the more common word thanks to [RFCs](https://www.rfc-editor.org/rfc/rfc2119)

rebloor: Thanks for the feedback, @vadcx. Best practice for technical writing is to avoid modal verbs, such as may, should, etc. While "may" is useful for indicating something is optional in a request for change, it is less appropriate given that we are describing an implemented feature and something that the developer is able to specify.

files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
rebloor: @vadcx thanks for spotting

files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
```diff
@@ -15,7 +15,7 @@ Values of this type are objects that can contain these properties:
 - `domain`
   - : A `string` representing the domain the cookie belongs to (e.g., "www.google.com" or "example.com").
 - `expirationDate` {{optional_inline}}
-  - : A `number` representing the expiration date of the cookie as the number of seconds since the UNIX epoch. Not provided for session cookies.
+  - : A `number` representing the expiration date of the cookie as the seconds after the UNIX epoch. From Firefox 142 includes milliseconds in the fractional part. Not provided for session cookies.
```
Rob--W: Omit the "Firefox 142" part; E.g. Chrome also did this already.

files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
```diff
@@ -15,7 +15,7 @@ Values of this type are objects that can contain these properties:
 - `domain`
   - : A `string` representing the domain the cookie belongs to (e.g., "www.google.com" or "example.com").
 - `expirationDate` {{optional_inline}}
-  - : A `number` representing the expiration date of the cookie as the number of seconds since the UNIX epoch. Not provided for session cookies.
+  - : A `number` representing the expiration date of the cookie as the seconds after the UNIX epoch. Includes milliseconds in the fractional part. Not provided for session cookies.
```
Rob--W, line 18:
```suggestion
  - : A `number` representing the expiration date of the cookie as the seconds after the UNIX epoch. May include milliseconds in the fractional part. Not provided for session cookies.
```

Maybe may to signal the optionalness? Commonly it would be an integer, e.g. the Expires flag in cookies APIs only have a resolution of seconds, without room for ms.

Rob--W (approved):
Approved with the comment acknowledged or addressed.

files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
rebloor: @Rob--W isn't the point that it now always includes milliseconds, even if the value of milliseconds is zero. The problem with using the modal verb is that it could be read as "may not include milliseconds even if they are in the source of value."

Rob--W: Previously it was unspecified what the resolution of the number is.

A fraction could be passed, and the implementation would truncate/round it to a second. Now it can include milliseconds, but unsurprisingly it can often be a whole number.

I'll leave it up to you on how to word this. If you think that the current phrasing is the best, feel free to stick with that.

rebloor: Thanks @Rob--W

PR #41098, opened 2025-09-11
Add missing "optional" label to "transports" in PublicKeyCredentialRequestOptions doc page

SagePtr:
### Description

Add missing "optional" label to "transports" parameter of "allowCredentials" array entry.

### Motivation

This parameter is optional, but was not marked with label, see below.

### Additional details

According to [Specification](https://w3c.github.io/webauthn/#dictionary-credential-descriptor)

> **transports, of type sequence<DOMString>**
This OPTIONAL member contains a hint as to how the client might communicate with the managing authenticator of the public key credential the caller is referring to. The values SHOULD be members of AuthenticatorTransport but client platforms MUST ignore unknown values.


wbamberg (approved):
👍 thank you!

PR #41106, opened 2025-09-11
Mention `new Set()` accepts `undefined`

fisker:
### Description

> 4. If iterable is either undefined or null, return set.

https://tc39.es/ecma262/multipage/keyed-collections.html#sec-set-iterable


Josh-Cena (approved):
Nice, thanks! Also congrats on your first PR here :)

PR #41107, opened 2025-09-12
chore: correct the `timeRange` function name

yin1999:
### Description

correct the `timeRange` function name

mdn/translated-content#28920


Josh-Cena (approved):
Thanks!

PR #41109, opened 2025-09-12
Remove inert "until supported"

estelle:
The attribute is baseline 2023 / fully supported, so the plugin and fully supported text is no longer needed.

There is  a bug in safari that allows in content search on inert content, which demoted safari support, but it is supported otherwise.


chrisdavidmills (approved):
Makes sense; thanks @estelle!

PR #41115, opened 2025-09-12
New attacks article: phishing

wbamberg:
Here's a new article on phishing attacks.

- I went back and forth a bit on how much to talk about the details of WebAuth. Minimally we could just say "a passkey is associated with an origin, and the browser will never allow the user to choose the passkey for the real site when the user is on the fake site". But it felt worth talking a bit more than this. 
- I'm not sure how much to talk about "downsides" of passkeys. I do feel like the tooling isn't so mature at the moment and things like passkey export are a bit unclear. But I don't know much about this (yet).
- I was also on the fence about including the bit on security indicators. In the end I have included it but not recommended it.


Josh-Cena:
Another defense I've seen websites do, is ask the user to enter a custom phrase that is displayed on the login page. That way, unless the phishing site knows the custom phrase (which is as bad and as hard as knowing the actual credentials), the user will be alerted that it's not the real site. This is perhaps significantly easier to implement than OTP and passkeys.

wbamberg:
> Another defense I've seen websites do, is ask the user to enter a custom phrase that is displayed on the login page. That way, unless the phishing site knows the custom phrase (which is as bad and as hard as knowing the actual credentials), the user will be alerted that it's not the real site. This is perhaps significantly easier to implement than OTP and passkeys.

I do remember some sites that used to do this a long time ago. Is it still current practice? Do you know any significant sites that still do this (e.g. banking sites)? I think it would still be vulnerable to the adversary in the middle attack, and I wonder how well defenses based on users paying attention work against attacks that are really all about users who are not paying attention...

Josh-Cena:
I have a banking site that does this, but everything else about the site is old enough so I'm not sure if it's in trend in general. Nevertheless I think it's a viable practice to recommend, especially for low-capability developers or needing to support extremely esoteric setup.

> vulnerable to the adversary in the middle attack

Isn't that solved by HTTPS?

> users paying attention work against attacks that are really all about users who are not paying attention

Yeah that's true. It has to be baked into the UI and the user has to be accustomed to seeing it. Anyway the site looks like this:

Screenshot 2025-09-15 at 15 30 38

files/en-us/web/security/attacks/phishing/index.md
```diff
@@ -0,0 +1,113 @@
...
+4. The user clicks the link and is asked to sign in. They enter their username and password, and the attacker now has their credentials.
+
+Phishing attacks sometimes use different techniques to contact their victims, such as text messages or voice calls, and sometimes don't involve websites at all, convincing users to tell them passwords or PIN codes verbally.
+
+Some phishing attacks are highly indiscriminate, sending out many messages to potential victims, in the hope that some will be tricked. In _spear-phishing_ attacks, on the other hand, attackers research specific victims, gathering personal information about them to make the lure more convincing. For example, the email may purport to come from someone the victim knows, and even include private information.
+
+Phishing attacks are not dependent on naive or inexperienced users: decades of experience has shown that even highly experienced and knowlegeable users can be vulnerable to phishing attacks, especially when they are busy, tired, or distracted.
```
Elchi3:
```suggestion
Phishing attacks are not dependent on naive or inexperienced users: decades of experience has shown that even highly experienced and knowledgeable users can be vulnerable to phishing attacks, especially when they are busy, tired, or distracted.
```

files/en-us/web/security/attacks/phishing/index.md
```diff
@@ -0,0 +1,113 @@
...
+sidebar: security
+---
+
+Phishing is a {{glossary("social engineering")}} attack in which a user is tricked into believing that they are interacting with a site with which they have an account, when in reality they are interacting with the attacker. The attacker convinces the user to enter their credentials on the fake site, and thereby steals the user's credentials.
+
+## Overview
+
+Phishing is a very old but still very common attack, which has gone through many variations, both to evade defenses and to take advantage of new vulnerabilities. In a basic form, though:
+
+1. The attacker registers a domain name that resembles the target site. For example, if the user's banking website is `my-bank.example.com`, the attacker could register `my-bank.examp1e.com`.
```
Elchi3, line 14: Optional: I think it might be worth calling out some tricks specifically. For example, say that unicode tricks (e.g. cyrillic ‘а’ vs Latin ‘a’) might get used and that the fake site might still use HTTPS and the original site's favicon to make the fake appear legitimate.

Elchi3 (approved):
This looks great to me, Will! 

Given many phishing attacks come via email I wondered what to do and I think this Cloudflare article has some good advice: https://www.cloudflare.com/learning/email-security/how-to-prevent-phishing/ From that we could maybe say that properly configuring SPF, DKIM, or DMARC email checks would be a thing that developers / domain owners should do? (among other things)

files/en-us/web/security/attacks/phishing/index.md
Josh-Cena: There's also [semantic attack](https://datatracker.ietf.org/doc/html/rfc3986#section-7.6) (which was briefly mentioned in https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Authority, although I just realized the Wikipedia link on that page is talking about a slightly different thing).

wbamberg:
> This looks great to me, Will!
> 
> Given many phishing attacks come via email I wondered what to do and I think this Cloudflare article has some good advice: https://www.cloudflare.com/learning/email-security/how-to-prevent-phishing/ From that we could maybe say that properly configuring SPF, DKIM, or DMARC email checks would be a thing that developers / domain owners should do? (among other things)

I have added a bit on this although I'm not sure how well I understand it!

files/en-us/web/security/attacks/phishing/index.md
wbamberg: We could add this, but I'm not sure it's that important in this guide to go into specific things attackers can do to make the fake domain resemble the target. None of this is actionable for developers AFAIK.

PR #41118, opened 2025-09-13
Docs: Clarify direct import option for ListView in Django examples

Jagan6991:
Description

I added a small clarification to the Django ListView example.
Along with using generic.ListView, I noted that developers can also directly import ListView from django.views.generic.

Motivation

This helps beginners who may find the longer generic.ListView import confusing. By showing the shorter direct import, the docs become a little clearer and more practical.

Additional details

No functional changes, only documentation improvement.

Related issues and pull requests

N/A


hamishwillee:
Thanks @Jagan6991 But I do not consider this documentation improvement necessary.

Yes it is true that you can do it this way and that it is equally valid Python. The reason it is done as it is is to highlight "even more" that they are generic views.

You might argue that this format as cleaner or that it is useful for people to know more about how Python imports work. While that is true this is not a Python lesson - we're trying to create a simple flow with as little diversion as possible so that people can follow the tutorial. I'd prefer to keep it straightforward at this point.

PR #41120, opened 2025-09-14
Make all JSON code syntactically valid

Josh-Cena:
This makes the code formattable.

### DIFF
--- a/files/en-us/web/api/performancelonganimationframetiming/tojson/index.md
+++ b/files/en-us/web/api/performancelonganimationframetiming/tojson/index.md
@@ -42,37 +42,37 @@ const observer = new PerformanceObserver((list) => {
 observer.observe({ type: "long-animation-frame", buffered: true });
 ```
 
-This would log a JSON object like so:
+This would log an object like so:
 
-```json
-{
-  "blockingDuration": 0,
-  "duration": 60,
-  "entryType": "long-animation-frame",
-  "firstUIEventTimestamp": 11801.099999999627,
-  "name": "long-animation-frame",
-  "renderStart": 11858.800000000745,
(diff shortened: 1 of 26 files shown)


files/en-us/web/api/performancelonganimationframetiming/tojson/index.md
```diff
@@ -42,37 +42,37 @@ const observer = new PerformanceObserver((list) => {
...
+      forcedStyleAndLayoutDuration: 0,
+      invoker: "DOMWindow.onclick",
+      invokerType: "event-listener",
+      name: "script",
+      pauseDuration: 0,
+      sourceURL: "https://web.dev/js/index-ffde4443.js",
+      sourceFunctionName: "myClickHandler",
+      sourceCharPosition: 17796,
+      startTime: 11803.199999999255,
+      window: {}, // Window object
```
hamishwillee: Not entirely sure about this. but assume it is OK to have an empty json object here.

hamishwillee (approved):
Makes sense to me - thank you. Approving. Merge at will.

files/en-us/web/api/performancelonganimationframetiming/tojson/index.md
Josh-Cena: Just realized that since this is not JSON anymore I can add comments here. Also I found out that `window` is in fact not included if you serialize `PerformanceScriptTiming` itself... which is actually a bit surprising for me.

hamishwillee: Thanks. This makes more sense now. I'm not sure what it would "mean" to serialize window.

PR #41121, opened 2025-09-14
Cross-reference `:target` and `::target-text` pages

lionel-rowe:
### Description

Cross-reference `:target` and `::target-text` pages under their respective _See also_ sections.

### Motivation

Closely conceptually related APIs that don't currently cross-reference each other. Also useful to have a handy reference to each other's syntax, given that `:target` is a single-colon pseudo-class whereas `::target-text` is a double-colon pseudo-element.


chrisdavidmills (approved):
This makes sense; thanks @lionel-rowe!

We don't usually put descriptions after links in "See also" sections, but I'm thinking perhaps it makes sense in this case, as the names are so similar. I'll merge this for now, and we can always modify it further in a future PR if others are uncomfortable with it.

PR #41126, opened 2025-09-15
FF143 GPU import external texture experimental relnote

hamishwillee:
FF143 adds support for [`GPUExternalTexture`](https://developer.mozilla.org/en-US/docs/Web/API/GPUExternalTexture) and [`GPUDevice.importExternalTexture()`](https://developer.mozilla.org/en-US/docs/Web/API/GPUDevice/importExternalTexture) behind a pref in 143.

This adds a release note to the experimental section.

Note that I have not added an update to the Experimental features page as this has already shipped in the now current beta 144. So I expect in a few days I would have to remove that, which would be pointless effort.

Related docs work can be tracked in #40774


dipikabh (approved):
Looks good, thank you!

PR #41127, opened 2025-09-15
Document trailing space handling in opaque URL paths

### LINKED ISSUE #23320: URL standard update: opaque paths cannot have trailing space
In https://github.com/whatwg/url/pull/728 we tweaked the `search` and `hash` setters to account for a minor roundtrip issue. Depending on the level of detail you go into this may impact documentation.

Josh-Cena:
Fix https://github.com/mdn/content/issues/23320. See also https://github.com/mdn/browser-compat-data/issues/27856

cc @annevk to see if this makes sense

### DIFF
--- a/files/en-us/web/api/url/pathname/index.md
+++ b/files/en-us/web/api/url/pathname/index.md
@@ -16,7 +16,7 @@ The `pathname` value for such URLs will therefore always have at least one `/` c
 For non-hierarchical schemes, the pathname is known as an _opaque path_ (meaning, the URL parser does not try to split it into a list of segments). In this case, an empty path results in the `pathname` property being the empty string. Trailing spaces in opaque paths are stripped during initial parsing if the `hash` and `search` are both empty; otherwise, they are percent-encoded as `%20` even when `hash` and `search` are later set to empty strings.
 
 > [!NOTE]
-> Percent-encoding trailing spaces in opaque paths is not widely implemented. Some browsers implement the old behavior of stripping trailing spaces from `pathname` whenever the `hash` and `search` properties are both empty strings. In these browsers, setting `hash` or `search` may change the `pathname` as well.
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/url/pathname/index.md
```diff
@@ -13,7 +13,10 @@ The **`pathname`** property of the {{domxref("URL")}} interface represents a loc
...
 The `pathname` value for such URLs will therefore always have at least one `/` character.
 
-For non-hierarchical schemes, if the URL has no path segments, the value of its `pathname` property will be the empty string.
+For non-hierarchical schemes, the pathname is known as an _opaque path_ (meaning, the URL parser does not try to split it into a list of segments). In this case, an empty path results in the `pathname` property being the empty string. Trailing spaces in opaque paths are stripped during initial parsing if the `hash` and `search` are both empty; otherwise, they are percent-encoded as `%20` even when `hash` and `search` are later set to empty strings.
+
+> [!NOTE]
+> Percent-encoding trailing spaces in opaque paths is not widely implemented. Some browsers implement the old behavior of stripping trailing spaces from `pathname` whenever the `hash` and `search` properties are both empty strings. In these browsers, setting `hash` or `search` may change the `pathname` as well.
```
annevk: Those setters with side effects have also only been in place for a relatively short period so older browsers do not even have that and instead end up with a trailing space in the path that would disappear after a serialize-then-parse.

Josh-Cena: Arguably this "even older behavior" is a bug, and has been gone for 2 years already, so I opted to not include it because it makes things even more complicated—it's fine to only document the status quo and not history, and I only included this note because of current divergence, not because of historical change. I've filed a BCD issue to get both behaviors' versions recorded, so this should be backed by real data soon. Of course if you have some wording suggestion in mind, I would be happy to make it more complete

annevk: Not encoding to %20 is also a bug at this point though. 😊

Josh-Cena:
I've updated with mentioning of the even-older behavior

sideshowbarker (approved):

PR #41128, opened 2025-09-15
HTMLScriptElement.text - trustedTypes for scripts

hamishwillee:
This updates the TrustedType documentation for `HTMLScriptElement.text` so that it includes boilerplate and security considerations.

Note that this links to `HTMLScriptElement.textContent` and `HTMLScriptElement.innerText` which are not yet defined. These were inherited from node/Element but are now in IDL - see https://w3c.github.io/trusted-types/dist/spec/#enforcement-in-scripts

Some comments by way of explanation/discussion inline.

### DIFF
--- a/files/en-us/web/api/htmlscriptelement/text/index.md
+++ b/files/en-us/web/api/htmlscriptelement/text/index.md
@@ -35,12 +35,12 @@ For other types it might represent an import map, speculation rules, or some oth
 
 Note that if the {{domxref('HTMLScriptElement/src','src')}} property is set the content of the `text` property is ignored.
 
-### text vs textContent vs innerText
+### `text` vs `textContent` vs `innerText`
 
 The `text` and {{domxref("HTMLScriptElement.textContent", "textContent")}} properties of `HTMLScriptElement` are equivalent: both can be set with a string or a `TrustedScript` type and both return a string representing the content of the script element.
 The main difference is that {{domxref("Node.textContent", "textContent")}} is also defined on {{domxref("Node")}} and can be used with other elements to set their content with a string.
 
-{{domxref("HTMLScriptElement.innerText", "innerText")}} will generally set and execute the text in the same as the other methods, but may return a slightly different value.
+{{domxref("HTMLScriptElement.innerText", "innerText")}} will generally set and execute the text in the same way as the other methods, but may return a slightly different value.
 The reason for this is that this property is designed for getting the rendered text of a string of HTML markup.
 When setting the value the text is treated as a text node, which normalizes the string as if it were visible text (collapsing spaces and converting `\n` to line breaks).
 This does not change the execution of the text, but it does alter the text that is stored and returned.
@@ -60,7 +60,7 @@ This ensures that the input is passed through a transformation function, which h
 
 The behavior of the transformation function will depend on the specific use case that requires a user provided script.
 If possible you should lock the allowed scripts to exactly the code that you trust to run.
-If that is not possible, you might whitelist or blacklist the use of certain functions within the provided string.
+If that is not possible, you might allow or block the use of certain functions within the provided string.
 
 ## Examples
 
@@ -115,7 +115,7 @@ const trustedScript = policy.createScript(untrustedScriptOne);
 el.text = trustedScript;
 ```
 
-### text and textContent properties are equivalent
+### Comparing `text` and `textContent`
 
 In this example we'll set the value of a script element by assigning a string of code to the element's `text` property and `textContent` properties, and read the result back to show that the results are equivalent.
 
--- a/files/en-us/web/api/htmlscriptelement/text/index.md
+++ b/files/en-us/web/api/htmlscriptelement/text/index.md
@@ -8,29 +8,127 @@ browser-compat: api.HTMLScriptElement.text
 
 {{APIRef("HTML DOM")}}
 
-The **`text`** property of the {{domxref("HTMLScriptElement")}} interface is a string that reflects the text content inside the {{HTMLElement("script")}} element. It acts the same way as the {{domxref("Node.textContent")}} property.
+> [!WARNING]
+> This property represents the text content of a script element, which may be executable depending on the script type.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
+>
+> You can mitigate this risk by always assigning {{domxref("TrustedScript")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](#security_considerations) for more information.
+
+The **`text`** property of the {{domxref("HTMLScriptElement")}} interface represents the inline text content of the script element.
+It acts the same way as the {{domxref("HTMLScriptElement.textContent","textContent")}} property.
 
 It reflects the `text` attribute of the {{HTMLElement("script")}} element.
 
 ## Value
 
-A string.
+Getting the property returns a string containing the element's text.
+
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/htmlscriptelement/text/index.md
```diff
@@ -8,29 +8,127 @@ browser-compat: api.HTMLScriptElement.text
 
 {{APIRef("HTML DOM")}}
 
-The **`text`** property of the {{domxref("HTMLScriptElement")}} interface is a string that reflects the text content inside the {{HTMLElement("script")}} element. It acts the same way as the {{domxref("Node.textContent")}} property.
+> [!WARNING]
+> This property represents the text content of a script element, which may be executable depending on the script type.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
```
hamishwillee, line 13: Note, slightly different warning as this isn't DOM injection. Also because a script might not be executable.

files/en-us/web/api/htmlscriptelement/text/index.md
```diff
@@ -8,29 +8,127 @@ browser-compat: api.HTMLScriptElement.text
...
+
+### text vs textContent vs innerText
+
+The `text` and {{domxref("HTMLScriptElement.textContent", "textContent")}} properties of `HTMLScriptElement` are equivalent: both can be set with a string or a `TrustedScript` type and both return a string representing the content of the script element.
+The main difference is that {{domxref("Node.textContent", "textContent")}} is also defined on {{domxref("Node")}} and can be used with other elements to set their content with a string.
+
+{{domxref("HTMLScriptElement.innerText", "innerText")}} will generally set and execute the text in the same as the other methods, but may return a slightly different value.
+The reason for this is that this property is designed for getting the rendered text of a string of HTML markup.
+When setting the value the text is treated as a text node, which normalizes the string as if it were visible text (collapsing spaces and converting `\n` to line breaks).
+This does not change the execution of the text, but it does alter the text that is stored and returned.
```
hamishwillee: `text` used to be the script-specific way to set/get the text. There was also inherited innerText and textContent. These two are now implemented in the interface https://w3c.github.io/trusted-types/dist/spec/#enforcement-in-scripts - they are all very similar but there is a subtle difference for innerText. I thought this might clarifya bit for those that are interested.

files/en-us/web/api/htmlscriptelement/text/index.md
```diff
@@ -8,29 +8,127 @@ browser-compat: api.HTMLScriptElement.text
...
+
+```js
+const untrustedCode = "alert('Potentially evil code!');";
+scriptElement.text = untrustedCode; // shows the alert
+```
+
+You can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) or reject the text before it is injected.
+
+The behavior of the transformation function will depend on the specific use case that requires a user provided script.
```
hamishwillee, line 61: Wasn't entirely sure what to say here. The fact is that you shouldn't generally be taking untrusted strings and injecting them directly into scripts. If you do then using trustedScripts is a good idea, but it is impossible to provide general guidance for something that you would only be doing in extreme circumstances. I've tried to lock it down a bit and give some flavour of the risk in the avoidance suggestions.

files/en-us/web/api/htmlscriptelement/text/index.md
```diff
@@ -8,29 +8,127 @@ browser-compat: api.HTMLScriptElement.text
...
+## Description
+
+The **`text`** property of the {{domxref("HTMLScriptElement")}} interface represents the text content inside the {{HTMLElement("script")}} element.
+
+For an executable script {{domxref('HTMLScriptElement/type','type')}}, such as a module or classic script, this text is inline executable code.
+For other types it might represent an import map, speculation rules, or some other kind of data block.
+
+Note that if the {{domxref('HTMLScriptElement/src','src')}} property is set the content of the `text` property is ignored.
+
+### text vs textContent vs innerText
```
wbamberg:
```suggestion
### `text` vs `textContent` vs `innerText`
```

it's ugly but I think good practice to do this, especially for the benefit of localizers.

files/en-us/web/api/htmlscriptelement/text/index.md
```diff
@@ -8,29 +8,127 @@ browser-compat: api.HTMLScriptElement.text
...
+
+The `text` and {{domxref("HTMLScriptElement.textContent", "textContent")}} properties of `HTMLScriptElement` are equivalent: both can be set with a string or a `TrustedScript` type and both return a string representing the content of the script element.
+The main difference is that {{domxref("Node.textContent", "textContent")}} is also defined on {{domxref("Node")}} and can be used with other elements to set their content with a string.
+
+{{domxref("HTMLScriptElement.innerText", "innerText")}} will generally set and execute the text in the same as the other methods, but may return a slightly different value.
```
wbamberg:
```suggestion
{{domxref("HTMLScriptElement.innerText", "innerText")}} will generally set and execute the text in the same way as the other methods, but may return a slightly different value.
```

files/en-us/web/api/htmlscriptelement/text/index.md
```diff
@@ -8,29 +8,127 @@ browser-compat: api.HTMLScriptElement.text
...
+const untrustedCode = "alert('Potentially evil code!');";
+scriptElement.text = untrustedCode; // shows the alert
+```
+
+You can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) or reject the text before it is injected.
+
+The behavior of the transformation function will depend on the specific use case that requires a user provided script.
+If possible you should lock the allowed scripts to exactly the code that you trust to run.
+If that is not possible, you might whitelist or blacklist the use of certain functions within the provided string.
```
wbamberg:
```suggestion
If that is not possible, you might allow or block the use of certain functions within the provided string.
```

? maybe? -> https://developers.google.com/style/word-list#blacklist

files/en-us/web/api/htmlscriptelement/text/index.md
```diff
@@ -8,29 +8,127 @@ browser-compat: api.HTMLScriptElement.text
...
 
-el.text = "console.log(10);";
+// Create a TrustedScript instance using the policy
+const trustedScript = policy.createScript(untrustedScriptOne);
+
+// Inject the TrustedScript (which contains a trusted string)
+el.text = trustedScript;
+```
+
+### text and textContent properties are equivalent
```
wbamberg:
```suggestion
### Comparing `text` and `textContent`
```

idk why, but headings whose text tells me things irk me.

wbamberg (commented):
looks great, just some little things.

files/en-us/web/api/htmlscriptelement/text/index.md
hamishwillee: This is better thanks.

hamishwillee:
Thanks @wbamberg - I accepted all the suggestions.

wbamberg (approved):
👍 thank you Hamish!

PR #41147, opened 2025-09-15
CSS invert function

estelle:
made the page more like blur() function

### DIFF
--- a/files/en-us/web/css/filter-function/invert/index.md
+++ b/files/en-us/web/css/filter-function/invert/index.md
@@ -60,7 +60,7 @@ invert(60%)
 
 
 - {{cssxref("&lt;number&gt;")}} or {{cssxref("&lt;percentage&gt;")}} {{Optional_Inline}}
-  - : The amount of the conversion, specified as a {{cssxref("&lt;number&gt;")}} or a {{cssxref("&lt;percentage&gt;")}}. A value of `100%` is completely inverted, while a value of `0%` leaves the input unchanged. Values between `0%` and `100%` are linear multipliers on the effect. The initial value for {{Glossary("interpolation")}} is `0`. The default value is `1`.
+  - : Specifies the amount of the conversion. A value of `100%` is completely inverted, while a value of `0%` leaves the input unchanged. Values between `0%` and `100%` are linear multipliers on the effect. The initial value for {{Glossary("interpolation")}} is `0`. The default value is `1`.
 
 ## Formal syntax
 
--- a/files/en-us/web/css/filter-function/invert/index.md
+++ b/files/en-us/web/css/filter-function/invert/index.md
@@ -60,7 +60,7 @@ invert(60%)
 
 The `invert( <amount> )` function's parameter is:
 
-- `amount` {{Optional_Inline}}
+- {{cssxref("&lt;number&gt;")}} or {{cssxref("&lt;percentage&gt;")}} {{Optional_Inline}}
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/filter-function/invert/index.md
```diff
@@ -43,31 +43,134 @@ filter: invert(1);
...
+invert(100%)
+
+/* 60% inversion */
+invert(.6)
+invert(60%)
 ```
 
 ### Parameters
 
+The `invert( <amount> )` function's parameter is:
```
dipikabh:
```suggestion

```

files/en-us/web/css/filter-function/invert/index.md
```diff
@@ -43,31 +43,134 @@ filter: invert(1);
...
+
 - `amount` {{Optional_Inline}}
```
dipikabh: I think, to be closer to the Formal syntax, we should not use "amount", but instead:
```suggestion
- {{cssxref("&lt;number&gt;")}} or {{cssxref("&lt;percentage&gt;")}} {{Optional_Inline}}
```

files/en-us/web/css/filter-function/invert/index.md
```diff
@@ -43,31 +43,134 @@ filter: invert(1);
...
   - : The amount of the conversion, specified as a {{cssxref("&lt;number&gt;")}} or a {{cssxref("&lt;percentage&gt;")}}. A value of `100%` is completely inverted, while a value of `0%` leaves the input unchanged. Values between `0%` and `100%` are linear multipliers on the effect. The initial value for {{Glossary("interpolation")}} is `0`. The default value is `1`.
```
dipikabh: And then the description could read:
```suggestion
  - : Specifies the amount of the conversion. A value of `100%` is completely inverted, while a value of `0%` leaves the input unchanged. Values between `0%` and `100%` are linear multipliers on the effect. The initial value for {{Glossary("interpolation")}} is `0`. The default value is `1`.
```

files/en-us/web/css/filter-function/invert/index.md
```diff
@@ -43,31 +43,134 @@ filter: invert(1);
...
 
 ## Formal syntax
 
 {{CSSSyntax}}
 
+## SVG filter
```
dipikabh, line 68: all the same comments for the `blur()` function page PR (https://github.com/mdn/content/pull/41006) apply here as well, including the comments in the "Examples" section

dipikabh (changes requested):

files/en-us/web/css/filter-function/invert/index.md
dipikabh: since this is outside the scope of this PR, let's take it up in a follow up

files/en-us/web/css/filter-function/invert/index.md
```diff
@@ -43,31 +43,132 @@ filter: invert(1);
...
+filter: invert(0.9);
+filter: invert(90%);
+filter: url("#invert90"); /* with embedded SVG */
+filter: url("fileName.svg#invert90"); /* external SVG */
+```
+
 ## Examples
 
-### Examples of correct values for invert()
+This example shows three images: the image with a `blur()` filter function applied, the image with the equivalent SVG blur function applied, and the original images for comparison:
```
dipikabh:
```suggestion
This example shows three images for comparison: an image with an `invert()` filter function applied, an image with the equivalent SVG function applied, and the original image:
```

dipikabh (approved):
Thanks, @estelle! Just some last nits and copy/paste issue

PR #41156, opened 2025-09-16
remove duplicated paragraph

leander-behr-blag:
### Description

There were two versions of the same paragraph, I kept the one I think is clearer.


wbamberg (commented):
Thanks for catching this! But it looks like we want to keep the first paragraph, not the second: https://github.com/mdn/content/pull/39642/files#r2122235373.

But if you have suggestions for how to make it clearer, happy to discuss :).

wbamberg (approved):
👍 thanks!

PR #41163, opened 2025-09-17
remove deleted docs from CSSRef macro

OnkarRuikar:
The following links in the CSSRef macro needs to be removed
      - /Learn_web_development/Core/Styling_basics/Fundamental_CSS_comprehension
      - /Learn_web_development/Core/Styling_basics/Fancy_letterheaded_paper
      - /Learn_web_development/Core/Styling_basics

The first two docs have been removed and the links are now in redirects' file.

### DIFF
--- a/files/sidebars/cssref.yaml
+++ b/files/sidebars/cssref.yaml
@@ -29,9 +29,6 @@ sidebar:
       - /Learn_web_development/Core/Styling_basics/Images_media_forms
       - /Learn_web_development/Core/Styling_basics/Tables
       - /Learn_web_development/Core/Styling_basics/Debugging_CSS
-      - /Learn_web_development/Core/Styling_basics/Fundamental_CSS_comprehension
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
Thanks for the fix, @OnkarRuikar. I'll merge this, and then I'll get some more updates done to this sidebar. There is quite a lot of divergence now between it and then learn sidebar. My bad.

OnkarRuikar:
I think now, with the sidebars defined in yaml files, it is possible to have a precommit hook to check broken/redirected links in the sidebars when a commit has deleted files. Will that be helpful?

chrisdavidmills:
> I think now, with the sidebars defined in yaml files, it is possible to have a precommit hook to check broken/redirected links in the sidebars when a commit has deleted files. Will that be helpful?

It would be great to extend the broken link checking to the sidebars, yes.

PR #41171, opened 2025-09-17
Use percentages instead of absolute units for view progress timeline …

radiantly:
### Bug description

The view progress scroll timeline example is broken on MDN playground because it uses absolute units (stumped me for quite a bit!). Here's a screenshot of how it looks like before my changes on Chrome v140:

Screenshot before changes

### Fix

I've updated the example to use percentage units instead of absolute units. Here's a screen recording with my fix:

Screen recording with my changes

### DIFF
--- a/files/en-us/web/css/animation-timeline/view/index.md
+++ b/files/en-us/web/css/animation-timeline/view/index.md
@@ -165,13 +165,13 @@ To aid the understanding of the result, extra elements `subject-container`, `top
 
 .top {
   top: 0;
-  height: 244px;
+  height: 50%;
   align-items: end;
 }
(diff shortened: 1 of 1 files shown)


dipikabh (approved):
Hi @radiantly, it seems this is your first PR here. If so, welcome to MDN! 🎉 
The change looks good - thanks a lot for sending the fix.

Just a small tip for next time: adding the exact section link (like [this one](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timeline/view#setting_an_anonymous_view_progress_timeline)) can make it even easier for reviewers to check what you're updating.

PR #41173, opened 2025-09-17
Make Learn sidebar entries consistent across learn and css sidebars

chrisdavidmills:
### Description

The Learn area tutorial entries in the learn sidebar have become inconsistent with those in the cssref sidebar. This PR makes them consistent.

I've not included the test your skills index articles in the cssref sidebar, as I didn't think it was necessary to list them there.

### DIFF
--- a/files/sidebars/cssref.yaml
+++ b/files/sidebars/cssref.yaml
@@ -20,13 +20,23 @@ sidebar:
       - /Learn_web_development/Core/Styling_basics/Attribute_selectors
       - /Learn_web_development/Core/Styling_basics/Pseudo_classes_and_elements
       - /Learn_web_development/Core/Styling_basics/Combinators
+      - /Learn_web_development/Core/Styling_basics/Test_your_skills/Selectors
       - /Learn_web_development/Core/Styling_basics/Box_model
+      - /Learn_web_development/Core/Styling_basics/Test_your_skills/Box_model
       - /Learn_web_development/Core/Styling_basics/Handling_conflicts
+      - /Learn_web_development/Core/Styling_basics/Test_your_skills/Cascade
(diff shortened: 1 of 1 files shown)


files/sidebars/cssref.yaml
```diff
@@ -20,13 +20,23 @@ sidebar:
       - /Learn_web_development/Core/Styling_basics/Attribute_selectors
       - /Learn_web_development/Core/Styling_basics/Pseudo_classes_and_elements
       - /Learn_web_development/Core/Styling_basics/Combinators
+      - /Learn_web_development/Core/Styling_basics/Test_your_skills/Selectors
       - /Learn_web_development/Core/Styling_basics/Box_model
+      - /Learn_web_development/Core/Styling_basics/Test_your_skills/Box_model
       - /Learn_web_development/Core/Styling_basics/Handling_conflicts
+      - /Learn_web_development/Core/Styling_basics/Test_your_skills/Cascade
+      - /Learn_web_development/Core/Styling_basics/Fixing_blog_styles
```
dipikabh, line 28: While this one and `Size_decorate_content_panel` are okay to add, should we avoid adding the "Test_your_skills" links to the CSS sidebar?

Perhaps that's what you also meant in your description in the PR if I'm not mistaken

dipikabh (approved):

files/sidebars/cssref.yaml
chrisdavidmills: Nah, I meant just not to add the test your skills index article, i.e. this one https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Test_your_skills in the case of Styling basics.

If I end up implementing the solution described in https://github.com/orgs/mdn/discussions/828, these will go away anyway. But we'll merge this for now ;-)

Issue #30931, opened 2023-12-11
update Notifications API content to better explain persistent events and not persistent events


skyclouds2001:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API

### What information was incorrect, unhelpful, or incomplete?

see https://github.com/mdn/content/pull/30717#pullrequestreview-1773990291
reference to https://notifications.spec.whatwg.org/

> Our documentation for the Notifications API has a big flaw: we don't explain the difference between persistent events and not persistent events.
>
> This may be done in a follow-up PR, and we would need to update:
>
> * the overview page (Notifications API) to explain these two concepts.
> * Rework _Using the notifications API", maybe dividing it into two guides, one for persistent notifications (associated with service workers, and one for non-persistent notifications.

### What did you expect to see?

ditto.

BHUVANSH855:
The overview now covers the persistent vs non-persistent distinction. For the remaining guide work, would you prefer restructuring the existing “Using the Notifications API” guide to cover both models, or still split it into separate guides as originally suggested in #30717?

Josh-Cena:
@wbamberg perhaps may have some ideas

wbamberg:
I have a half-written PR on this that I still haven't got around to finishing: https://github.com/mdn/content/pull/41267. Maybe I will pick it up when the privacy docs are done.

PR #41177, opened 2025-09-17
Create notifications from service worker so it works on mobile

### LINKED ISSUE #9167: Content bug: PWA Notification guide only works on desktop browsers.
## What page(s) did you find the problem on?

https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Re-engageable_Notifications_Push#create_a_notification

## Specific page section or heading?

#create_a_notification

## What is the problem?

I followed the guide in order to create a simple PWA notification, however it only works on desktop version.

## What did you expect to see?

I expect it to work on android.

## Did you test this? If so, how?

Tested on Ubuntu 20.04 and Windows 10, works on both as expected.

Tested it on my real device Android 9 with Chrome version (93.0.4577.82).
Tested it on my android emulator running Android 5 with latest chrome version.

Both test on mobile shows no notification.

# How I solved it

I manage to solve by changing the code below.

```diff
function randomNotification() {
  const randomItem = Math.floor(Math.random() * games.length);
  const notifTitle = games[randomItem].name;
  const notifBody = `Created by ${games[randomItem].author}.`;
  const notifImg = `data/img/${games[randomItem].slug}.jpg`;
  const options = {
    body: notifBody,
    icon: notifImg,
  };
-  new Notification(notifTitle, options);
+ navigator.serviceWorker.getRegistration()
+  .then((reg) => {
+    reg.showNotification(notifTitle, options);
+  });
  setTimeout(randomNotification, 30000);
}
```

Now it works both on desktop(windows, linux) and mobile(android).

Josh-Cena:
Fix https://github.com/mdn/content/issues/9167. I did not change the other occurrences of `new Notification` because they look like quick examples.

### DIFF
--- a/files/en-us/web/api/notification/notification/index.md
+++ b/files/en-us/web/api/notification/notification/index.md
@@ -14,7 +14,7 @@ Trying to create a notification inside the {{domxref("ServiceWorkerGlobalScope")
 
 You must first get permission before being able to display notifications, using {{domxref("Notification.requestPermission()")}}. The permission may not be grantable, for example if the page is in private browsing mode.
 
-This constructor throws a {{jsxref("TypeError")}} when called in nearly all mobile browsers and this is unlikely to change, because web pages on mobile devices almost never "run in the background", which is the main use case for notifications. Instead, you need to make the web page persistent, such as installing it to the home screen or registering a service worker and using {{domxref("ServiceWorkerRegistration.showNotification()")}}. See [Chrome issue](https://crbug.com/481856) for more information.
+This constructor throws a {{jsxref("TypeError")}} when called in nearly all mobile browsers and this is unlikely to change, because web pages on mobile devices almost never "run in the background", which is the main use case for notifications. Instead, you need to register a service worker and use {{domxref("ServiceWorkerRegistration.showNotification()")}}. See [Chrome issue](https://crbug.com/481856) for more information.
 
 ## Syntax
 
--- a/files/en-us/web/api/notification/notification/index.md
+++ b/files/en-us/web/api/notification/notification/index.md
@@ -8,11 +8,13 @@ browser-compat: api.Notification.Notification
 
 {{APIRef("Web Notifications")}}{{securecontext_header}} {{AvailableInWorkers}}
 
-The **`Notification()`** constructor creates a new
-{{domxref("Notification")}} object instance, which represents a user notification.
+The **`Notification()`** constructor creates a new {{domxref("Notification")}} object instance, which represents a user notification.
 
-> [!NOTE]
-> Trying to create a notification inside the {{domxref("ServiceWorkerGlobalScope")}} using the `Notification()` constructor will throw a `TypeError`. Use {{domxref("ServiceWorkerRegistration.showNotification()")}} instead.
+Trying to create a notification inside the {{domxref("ServiceWorkerGlobalScope")}} using the `Notification()` constructor will throw a `TypeError`. Use {{domxref("ServiceWorkerRegistration.showNotification()")}} instead.
+
+You must first get permission before being able to display notifications, using {{domxref("Notification.requestPermission()")}}. The permission may not be grantable, for example if the page is in private browsing mode.
+
+This constructor throws a {{jsxref("TypeError")}} when called in nearly all mobile browsers and this is unlikely to change, because web pages on mobile devices almost never "run in the background", which is the main use case for notifications. Instead, you need to make the web page persistent, such as installing it to the home screen or registering a service worker and using {{domxref("ServiceWorkerRegistration.showNotification()")}}. See [Chrome issue](https://crbug.com/481856) for more information.
 
 ## Syntax
 
@@ -88,15 +90,6 @@ if (Notification.permission === "granted") {
 ## Browser compatibility
 
 {{Compat}}
-
-### Chrome notes
-
-Starting in Chrome 49, notifications don't work in incognito mode.
-
-Chrome for Android will throw a {{jsxref("TypeError")}} when calling the
-`Notification` constructor. It only supports creating
-notifications from a service worker. See the
-[Chromium issue tracker](https://crbug.com/481856) for more details.
 
 ## See also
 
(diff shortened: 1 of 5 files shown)


files/en-us/web/api/notification/notification/index.md
```diff
@@ -8,11 +8,13 @@ browser-compat: api.Notification.Notification
...
 
-> [!NOTE]
-> Trying to create a notification inside the {{domxref("ServiceWorkerGlobalScope")}} using the `Notification()` constructor will throw a `TypeError`. Use {{domxref("ServiceWorkerRegistration.showNotification()")}} instead.
+Trying to create a notification inside the {{domxref("ServiceWorkerGlobalScope")}} using the `Notification()` constructor will throw a `TypeError`. Use {{domxref("ServiceWorkerRegistration.showNotification()")}} instead.
+
+You must first get permission before being able to display notifications, using {{domxref("Notification.requestPermission()")}}. The permission may not be grantable, for example if the page is in private browsing mode.
+
+This constructor throws a {{jsxref("TypeError")}} when called in nearly all mobile browsers and this is unlikely to change, because web pages on mobile devices almost never "run in the background", which is the main use case for notifications. Instead, you need to make the web page persistent, such as installing it to the home screen or registering a service worker and using {{domxref("ServiceWorkerRegistration.showNotification()")}}. See [Chrome issue](https://crbug.com/481856) for more information.
```
wbamberg: I personally would make this simpler and more direct, and I would subsume the first para into this as they are kind of connected:

```suggestion
In most mobile browsers, this constructor throws a {{jsxref("TypeError")}} when called, and you should instead show a notification by calling {{domxref("ServiceWorkerRegistration.showNotification()")}}, putting the notification under control of an active service worker.

In service worker contexts, this constructor always throws a {{jsxref("TypeError")}}, and you should always use {{domxref("ServiceWorkerRegistration.showNotification()")}} to show a notification.
```

Reasoning:

- in general I don't think it's worth it or useful to make forward-looking statements. People just need to know what to do now.
- "you need to make the web page persistent, such as installing it to the home screen or registering a service worker" is confusing: I'm not sure what a persistent web page is, and "installing it ... **or** registering a SW..." suggests that you can do it without a SW, which I don't think you can, and installing it is beside the point: all you need AFAICS is a SW, which is an implicit requirement of being able to call `ServiceWorkerRegistration.showNotification()`.

I also don't understand why BCD for this marks it as partially supported on Android.

files/en-us/web/api/notifications_api/using_the_notifications_api/index.md
```diff
@@ -28,6 +28,9 @@ Because of abuses of push notifications in the past, web browsers and developers
 
 In addition, In Chrome and Firefox you cannot request notifications at all unless the site is a secure context (i.e., HTTPS), and you can no longer allow notification permissions to be requested from cross-origin {{htmlelement("iframe")}}s.
 
+> [!NOTE]
+> The examples in this article uses the {{domxref("Notification/Notification", "Notification()")}} constructor to create notifications. This is fine for desktop, but on most mobile browsers this will throw a {{jsxref("TypeError")}}. If you are targeting mobile devices, you should use {{domxref("ServiceWorkerRegistration.showNotification()")}} from a service worker instead.
```
wbamberg: Is it necessarily _from a SW_? I thought you could call this from the web page (or the SW), but it was handled in the SW context?

files/en-us/web/progressive_web_apps/tutorials/js13kgames/app_structure/index.md
```diff
@@ -114,8 +114,14 @@ document.getElementById("content").innerHTML = content;
 Next, it registers a service worker:
 
 ```js
+let swRegistration = null;
```
wbamberg, line 117: These changes look fine but I haven't tested it. I guess the corresponding changes in https://github.com/mdn/pwa-examples/tree/main/js13kpwa need to be made.

files/en-us/web/api/notification/notification/index.md
Josh-Cena: I think it's important to say the "why" in addition to the "how". For example, if it's due to some technical limitation and may change in the next few months or years, then I may either do feature testing or leave a TODO comment saying to change to the simpler version when the browser bug is fixed; but if it's never going to work then I would just give up and use the long form and pretend that `Notification()` doesn't exist.

By "persistent", I meant that the process keeps running even when the page is unloaded. Mobile browsers aggressively background tabs (or even the browser itself can gets backgrounded), so the `Notification()` constructor _by definition_ cannot fire unless you are on the page itself, whereas on desktop it's likely to have the tab open while you are doing something else.

files/en-us/web/api/notifications_api/using_the_notifications_api/index.md
Josh-Cena: Yeah true, this is what I meant:

```suggestion
> The examples in this article uses the {{domxref("Notification/Notification", "Notification()")}} constructor to create notifications. This is fine for desktop, but on most mobile browsers this will throw a {{jsxref("TypeError")}}. If you are targeting mobile devices, you should register a service worker and use {{domxref("ServiceWorkerRegistration.showNotification()")}} instead.
```

files/en-us/web/api/notification/notification/index.md
wbamberg: About the why. Maybe it is useful for web developers to know whether a thing will ever be supported (although predicting the future is hard). I don't really agree but don't mind too much.

But the second point still stands.

wbamberg:
The spec says that notifications created using the constructor are "non-persistent notifications", and notifications created by `ServiceWorkerRegistration.showNotification()` are "persistent notifications", and that they have quite different behaviors.

It would be great to talk about this all explicitly in these docs, and that could also cover the use cases for the different types, and why persistent notifications are not so applicable on mobile.

But I suppose that's out of scope for this PR.

files/en-us/web/api/notification/notification/index.md
Josh-Cena:
```suggestion
This constructor throws a {{jsxref("TypeError")}} when called in nearly all mobile browsers and this is unlikely to change, because web pages on mobile devices almost never "run in the background", which is the main use case for notifications. Instead, you need to register a service worker and use {{domxref("ServiceWorkerRegistration.showNotification()")}}. See [Chrome issue](https://crbug.com/481856) for more information.
```

Josh-Cena:
> But I suppose that's out of scope for this PR.

Indeed, and we already have https://github.com/mdn/content/issues/30931

wbamberg (approved):

wbamberg:
> > But I suppose that's out of scope for this PR.
> 
> Indeed, and we already have #30931

Ah, great. I think that extra context will really help it all make sense.

PR #41178, opened 2025-09-17
Mention that text fragments are less durable

### LINKED ISSUE #39342: Improve wording concerning text fragments durability
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments

Concepts and usage

This section includes the following text:

<blockquote>If the ID is changed or removed, the document fragment is ignored, and the link just links through to the top of the page. This is reasonable in terms of graceful degradation, but it would arguably be better if the author of the link had full control over where they link to, without needing to rely on the page author.

Text fragments make this a reality — they allow link authors to specify text content to link to, rather than document fragments, in a flexible manner.</blockquote>

If the ID is changed or removed, the document fragment is ignored, and the link just links through to the top of the page. This is reasonable in terms of graceful degradation, but it would arguably be better if the author of the link had full control over where they link to, without needing to rely on the page author.

Text fragments make this a reality — they allow link authors to specify text content to link to, rather than document fragments, in a flexible manner. (With limits: Text on a destination page can be edited after the link is created, creating the same graceful degradation issue.)

_No response_

_No response_

Josh-Cena:
Fix https://github.com/mdn/content/issues/39342

### DIFF
--- a/files/en-us/web/uri/reference/fragment/text_fragments/index.md
+++ b/files/en-us/web/uri/reference/fragment/text_fragments/index.md
@@ -9,7 +9,7 @@ browser-compat:
 sidebar: urlsidebar
 ---
 
-**Text fragments** link directly to specific text in a web page, without requiring the page author to add an ID. They use a special syntax in the URL fragment. This feature lets you create deep links to content you don't control, which may not have associated IDs. It also makes sharing links more useful by directly point others to specific words. Browsers may differ in how they draw attention to the linked text—usually, the text is scrolled into view and highlighted with color.
+**Text fragments** link directly to specific text in a web page, without requiring the page author to add an ID. They use a special syntax in the URL fragment. This feature lets you create deep links to content that you don't control and may not have IDs associated. It also makes sharing links more useful by directly pointing others to specific words. Browsers may differ in how they draw attention to the linked text—usually, the text is scrolled into view and highlighted with color.
 
 ## Concepts and usage
 
--- a/files/en-us/web/uri/reference/fragment/text_fragments/index.md
+++ b/files/en-us/web/uri/reference/fragment/text_fragments/index.md
@@ -9,7 +9,7 @@ browser-compat:
 sidebar: urlsidebar
 ---
 
-**Text fragments** allow linking directly to a specific portion of text in a web document, without requiring the author to annotate it with an ID, using particular syntax in the URL fragment. Supporting browsers are free to choose how to draw attention to the linked text, e.g., with a color highlight and/or scrolling to the content on the page. This is useful because it allows web content authors to deep-link to other content they don't control, without relying on the presence of IDs to make that possible. Building on top of that, it could be used to generate more effective content-sharing links for users to pass to one another.
+**Text fragments** link directly to specific text in a web page, without requiring the page author to add an ID. They use a special syntax in the URL fragment. This feature lets you create deep links to content you don't control, which may not have associated IDs. It also makes sharing links more useful by directly point others to specific words. Browsers may differ in how they draw attention to the linked text—usually, the text is scrolled into view and highlighted with color.
 
 ## Concepts and usage
 
--- a/files/en-us/web/uri/reference/fragment/text_fragments/index.md
+++ b/files/en-us/web/uri/reference/fragment/text_fragments/index.md
@@ -28,9 +28,9 @@ The issue with linking to specific document fragments is that the author of the 
 </h2>
 ```
 
-Not all documents have such anchors, and even if they do, depending on the context, linking to a heading may be much less obvious than linking directly to the piece of text you are citing. This is where text fragments come in: it allows the link author to have full control over where in the document they want to link to, without requiring any special markup in the document itself. For example, a search engine may refer to a specific sentence in a page in its search results, and clicking the link will take you directly to that sentence.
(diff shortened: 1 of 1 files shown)


files/en-us/web/uri/reference/fragment/text_fragments/index.md
```diff
@@ -28,9 +28,9 @@ The issue with linking to specific document fragments is that the author of the
 </h2>
 ```
 
-If the ID is changed or removed, the document fragment is ignored, and the link just links through to the top of the page. This is reasonable in terms of graceful degradation, but it would arguably be better if the author of the link had full control over where they link to, without needing to rely on the page author.
+Not all documents have such anchors, and even if they do, depending on the context, linking to a heading may be much less obvious than linking directly to the piece of text you are citing. This is where text fragments come in: it allows the link author to have full control over where in the document they want to link to, without requiring any special markup in the document itself. For example, a search engine may refer to a specific sentence in a page in its search results, and clicking the link will take you directly to that sentence.
```
dipikabh: - Let's bring back the deleted para to maintain continuity with the previous text. We can follow that up with your additions

> If the ID is changed or removed, the browser will ignore the fragment and will open the top of the page. This is reasonable in terms of graceful degradation, but it also means that you need to depend on the page's markup.

- Some suggestions to remove few words to make prose more compact
```suggestion
Not all documents have such anchors, and even if they do, linking to a heading might be much less obvious than linking directly to the specific text you are citing. This is where text fragments help: they allow the link author to have full control over what text to link to, without requiring any special markup in the target document. For example, a search engine may refer to a specific sentence in its search results, and clicking the link will take you directly to that sentence.
```

files/en-us/web/uri/reference/fragment/text_fragments/index.md
```diff
@@ -28,9 +28,9 @@ The issue with linking to specific document fragments is that the author of the
...
 
-**Text fragments** make this a reality — they allow link authors to specify text content to link to, rather than document fragments, in a flexible manner.
+Of course, text fragments also have the limitation that text in a document is less stable than document structure. If the linked document is updated, the fragment would not navigate to the correct point, but would stay at the top of the page. This is fine for transient links such as those in search results, but if you intend for the link to work over time, document fragments may be more reliable.
```
dipikabh:
```suggestion
However, text fragments also have a limitation: text in a document is less stable than document structure. If the text in the linked document is updated, the fragment no longer matches, and the browser navigates to the top of the page. This is fine for transient links such as those in search results, but if you intend for the link to work over time, document fragments may be more reliable.
```

files/en-us/web/uri/reference/fragment/text_fragments/index.md
dipikabh, line 1: while here, can we also improve the intro para - the 1st sentence is long and convoluted and the flow of sentences is breaking up by the browser sentence.

> **Text fragments** let you link directly to specific text in a web page, without requiring the page author to add an ID. They use a special syntax in the URL fragment. This feature lets you create deep links to content you don't control, without relying on the presence of IDs. It also makes sharing links more useful, since you can directly point others to specific words. Browsers may differ in how they draw attention to the linked text (for example, with a color highlight) and/or scroll it into view.

dipikabh (approved):
Thanks a lot! It's looking good overall

files/en-us/web/uri/reference/fragment/text_fragments/index.md
Josh-Cena: > Let's bring back the deleted para to maintain continuity with the previous text

My intention is for the current version to fully subsume the original meaning but without the incorrect connotation that "text fragments are more stable than ID fragments", which the point about changing IDs is alluding to. The point about author having full control still exists.

files/en-us/web/uri/reference/fragment/text_fragments/index.md
```diff
@@ -9,7 +9,7 @@ browser-compat:
 sidebar: urlsidebar
 ---
 
-**Text fragments** allow linking directly to a specific portion of text in a web document, without requiring the author to annotate it with an ID, using particular syntax in the URL fragment. Supporting browsers are free to choose how to draw attention to the linked text, e.g., with a color highlight and/or scrolling to the content on the page. This is useful because it allows web content authors to deep-link to other content they don't control, without relying on the presence of IDs to make that possible. Building on top of that, it could be used to generate more effective content-sharing links for users to pass to one another.
+**Text fragments** link directly to specific text in a web page, without requiring the page author to add an ID. They use a special syntax in the URL fragment. This feature lets you create deep links to content you don't control, which may not have associated IDs. It also makes sharing links more useful by directly point others to specific words. Browsers may differ in how they draw attention to the linked text—usually, the text is scrolled into view and highlighted with color.
```
dipikabh: looks like i introduced a typo with "point"
```suggestion
**Text fragments** link directly to specific text in a web page, without requiring the page author to add an ID. They use a special syntax in the URL fragment. This feature lets you create deep links to content that you don't control and may not have IDs associated. It also makes sharing links more useful by directly pointing others to specific words. Browsers may differ in how they draw attention to the linked text—usually, the text is scrolled into view and highlighted with color.
```

files/en-us/web/uri/reference/fragment/text_fragments/index.md
dipikabh: Actually, that's true - we don't need that para ("If the ID is changed or removed...") since we explain it later as a limitation of text fragments. My concern was that we might losing a detail, but we're not :)

dipikabh (approved):
Just one nit, and then it's good to go. Thanks a lot!

Josh-Cena:
Thanks for the review!

PR #41179, opened 2025-09-17
[htmlsidebar]: Replace manual HTML guides list with `listSubPages`

dipikabh:
### Description

With https://github.com/mdn/rari/issues/201 fixed and the items in the sidebar now automatically sorted based on `short-title` as well, we can replace the manually curated list for `HTML/Guides` in the sidebar file with `listSubPages`.

Also added a tiny tweak to the `short-title` in the cheatsheet so it surfaces to the top in the sidebar.

### Additional details

| With this fix | Presently on prod |
| --- | --- | 
| <kbd>html sidebar with listSubPages</kbd> | <kbd>html sidebar using manual listing</kbd> |

### DIFF
--- a/files/en-us/web/html/guides/cheatsheet/index.md
+++ b/files/en-us/web/html/guides/cheatsheet/index.md
@@ -1,6 +1,6 @@
 ---
 title: HTML cheatsheet for syntax and common tasks
-short-title: HTML cheatsheet
+short-title: Cheatsheet
 slug: Web/HTML/Guides/Cheatsheet
 page-type: guide
 sidebar: htmlsidebar
--- a/files/sidebars/htmlsidebar.yaml
+++ b/files/sidebars/htmlsidebar.yaml
@@ -3,19 +3,11 @@
 sidebar:
   - type: section
     link: /Web/HTML
-  - type: section
+  - type: listSubPages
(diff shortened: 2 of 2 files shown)


wbamberg:
In this particular case it doesn't seem to matter, but as a general rule, I don't think listing guides alphabetically is the right call. Often guides follow a curated order, usually from introductory and general to advanced and niche, and it's independent of the title.

Similarly, while I think "Cheatsheet" is a better short title than "HTML cheatsheet", renaming guides so as to affect their order seems like a bad practice: the name of a guide and its order are independent things.

This has been discussed a lot, and one option that some websites do is to have "weight" as a front matter item (https://gohugo.io/methods/page/weight/).

pepelsbey (approved):
Looks good! Thank you :)

dipikabh:
I agree 100% @wbamberg and so we shouldn't be replacing all our other hand curated orderings elsehwere with auto alphabetical order. In this specific case, I thought it was okay to do the alpha sorting.
I also realize that my little cheat with the Cheatsheet might not hold up if in the future we happen to add a guide that precedes "Cheatsheet" alphabetically.

Thanks for reviewing, @pepelsbey!

PR #41181, opened 2025-09-17
Remove unnecessary useCapture parameter from addEventListener (1)

Josh-Cena:
There hasn't been a need to specify `useCapture=false` since IE, and almost all cases where we specify this parameter, we are setting it to `false`. Removing it helps makes our `addEventListener` syntax more uniform and more easily greppable (for example, when I was removing `load` handlers, I missed a few because those have `"load"` on a separate line). This also removes one level of indentation which is nice. In order to review it, you probably want to hide whitespace.


bsmth (approved):
Nice, thank you 👍🏻

PR #41189, opened 2025-09-18
Dispel ambiguity in `Using classes`

### LINKED ISSUE #41188: Missing context on implicit continuity of code in `Using classes`
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_classes#accessor_fields

Extends and inheritance

After the parent class is done with modifying this, the derived class can do its own logic. Here we added a private field called #alpha, and also provided a pair of getter/setters to interact with them.

A derived class inherits all methods from its parent. For example, although ColorWithAlpha doesn't declare a get red() accessor itself, you can still access red because this behavior is specified by the parent class:

```js
const color = new ColorWithAlpha(255, 0, 0, 0.5);
console.log(color.red); // 255
```

A small investigation finds all other examples which explicitly ignore some code written has `// ...` to signify, but not in this case in [Extends and inheritance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_classes#extends_and_inheritance), where `class Color` is defined with just:

```js
class Color {
  #values;
  constructor(r, g, b, a = 1) {
    this.#values = [r, g, b, a];
  }
  get alpha() {
    return this.#values[3];
  }
  set alpha(value) {
    if (value < 0 || value > 1) {
      throw new RangeError("Alpha value must be between 0 and 1");
    }
    this.#values[3] = value;
  }
}
```

and it follows the statement we consider to have ambiguity - which `class Color` definition is the passage about?

- the example above in the current section?
  - ah NO `get red` accessor is defined there, nor does it gives hint that the accessor is defined the same as the example far-more above
- the example far-more above
  - in this case, rephrasing the passage into the following may be more helpful for the readers:

  ```markdown
  A derived class inherits all methods from its parent. For example, consider the get red() accessor we added to the Color in the [Accessor fields](#accessor_fields) section, even though we haven't declared one in the ColorWithAlpha, we can still access red because this behavior is specified by the parent class:
  ```

_No response_

originally posted in mdn/translated-content#29023, and investigated by @yin1999

PassionPenguin:
### Description

make it clear that we are refering to the `accessor fields` section's example, instead of the example just above the current sentence.

### Motivation

see #41188 for motivation

Fixes #41188 and hopefully, mdn/translated-content#29023

### DIFF
--- a/files/en-us/web/javascript/guide/using_classes/index.md
+++ b/files/en-us/web/javascript/guide/using_classes/index.md
@@ -637,7 +637,7 @@ There are a few things that have immediately come to attention. First is that in
 
 After the parent class is done with modifying `this`, the derived class can do its own logic. Here we added a private field called `#alpha`, and also provided a pair of getter/setters to interact with them.
 
-A derived class inherits all methods from its parent. For example, although `ColorWithAlpha` doesn't declare a `get red()` accessor itself, you can still access `red` because this behavior is specified by the parent class:
+A derived class inherits all methods from its parent. For example, consider the get red() accessor we added to the Color in the [Accessor fields](#accessor_fields) section, even though we haven't declared one in the ColorWithAlpha, we can still access red because this behavior is specified by the parent class:
 
 ```js
 const color = new ColorWithAlpha(255, 0, 0, 0.5);
--- a/files/en-us/web/javascript/guide/using_classes/index.md
+++ b/files/en-us/web/javascript/guide/using_classes/index.md
@@ -637,7 +637,7 @@ There are a few things that have immediately come to attention. First is that in
 
 After the parent class is done with modifying `this`, the derived class can do its own logic. Here we added a private field called `#alpha`, and also provided a pair of getter/setters to interact with them.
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/javascript/guide/using_classes/index.md
```diff
@@ -637,7 +637,7 @@ There are a few things that have immediately come to attention. First is that in
 
 After the parent class is done with modifying `this`, the derived class can do its own logic. Here we added a private field called `#alpha`, and also provided a pair of getter/setters to interact with them.
 
-A derived class inherits all methods from its parent. For example, although `ColorWithAlpha` doesn't declare a `get red()` accessor itself, you can still access `red` because this behavior is specified by the parent class:
+A derived class inherits all methods from its parent. For example, consider the get red() accessor we added to the Color in the [Accessor fields](#accessor_fields) section, even though we haven't declared one in the ColorWithAlpha, we can still access red because this behavior is specified by the parent class:
```
Josh-Cena:
```suggestion
A derived class inherits all methods from its parent. For example, consider the `get red()` accessor we added to the `Color` in the [Accessor fields](#accessor_fields) section—even though we haven't declared one in `ColorWithAlpha`, we can still access `red` because this behavior is specified by the parent class:
```

Josh-Cena (approved):
Nice, thanks

PR #41195, opened 2025-09-18
Add top-level self-link to the learnsidebar

chrisdavidmills:
### Description

I noticed that all the other major sections of MDN (HTML, CSS, JavaScript, etc.) have a top-level self-link at the top of their sidebars. This PR does the same thing for Learn web development.


Josh-Cena (approved):
:+1:

Issue #39393, opened 2025-05-05
Learning docs have pages not included in the sidebar


ZimCodes:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Multiple-column_Layout

### What specific section or headline is this issue about?

Learn web development > Core Modules > CSS Layouts

### What information was incorrect, unhelpful, or incomplete?

The [Multi-column layout](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Multiple-column_Layout) page is not included in the CSS Layout section of the Core Modules when learning web development. 

Not sure if it was intended or not. 

### What did you expect to see?

Multi-column layout added to the list of CSS Layouts when learning web development.

### Do you have any supporting links, references, or citations?

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Multiple-column_Layout

### Do you have anything more you want to share?

![Image](https://github.com/user-attachments/assets/419ca930-ec19-47a3-9d6d-5ab1205e7259)

Josh-Cena:
Within learning, we have the following docs not in any sidebar:

[/en-us/learn_web_development](https://developer.mozilla.org/en-US/docs/Learn_web_development) - adding in https://github.com/mdn/content/pull/41195
[/en-us/learn_web_development/core/challenges](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Challenges) - removing in https://github.com/mdn/content/pull/41194
[/en-us/learn_web_development/core/css_layout/legacy_layout_methods](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Legacy_Layout_Methods) - PR for including the `css_layout` ones: https://github.com/mdn/content/pull/41196
[/en-us/learn_web_development/core/css_layout/multiple-column_layout](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Multiple-column_Layout)
[/en-us/learn_web_development/core/css_layout/practical_positioning_examples](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Practical_positioning_examples)
[/en-us/learn_web_development/core/css_layout/supporting_older_browsers](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Supporting_Older_Browsers)
[...]

PR #41196, opened 2025-09-18
[Experiment]: Add CSS layout additional tutorials links to learnsidebar

chrisdavidmills:
### Description

There are still [several Learn articles](https://github.com/mdn/content/issues/39393#issuecomment-2851420219) not linked from the learnsidebar. The ones we want to keep should really be linked in some way. This PR provides an experimental suggestion of what we could do in each case — an "Additional tutorials" subsection at the end of each module sidebar subsection.

### DIFF
--- a/files/sidebars/learnsidebar.yaml
+++ b/files/sidebars/learnsidebar.yaml
@@ -128,6 +128,13 @@ sidebar:
       - /Learn_web_development/Core/CSS_layout/Test_your_skills/Responsive_design
       - /Learn_web_development/Core/CSS_layout/Fundamental_Layout_Comprehension
       - /Learn_web_development/Core/CSS_layout/Test_your_skills
+      - title: Additional_tutorials
+        details: closed
+        children:
+          - /Learn_web_development/Core/CSS_layout/Multiple-column_Layout
+          - /Learn_web_development/Core/CSS_layout/Practical_positioning_examples
+          - /Learn_web_development/Core/CSS_layout/Legacy_Layout_Methods
+          - /Learn_web_development/Core/CSS_layout/Supporting_Older_Browsers
   - link: /Learn_web_development/Core/Scripting
     title: Dynamic_scripting_with_JavaScript
     details: closed
@@ -394,6 +401,7 @@ l10n:
     About: About
     Resources_for_educators: Resources for educators
     Changelog: Changelog
+    Additional_tutorials: Additional tutorials
   pt-BR:
     CSS_layout: Esquemas CSS
     Asynchronous_JavaScript: Asynchronous JavaScript


Josh-Cena:
There isn't a chance we could just incorporate them into the normal flow instead of stashing them away? No strong opinions, this does solve the reported problem.

chrisdavidmills:
> There isn't a chance we could just incorporate them into the normal flow instead of stashing them away? No strong opinions, this does solve the reported problem.

They _were_ in the normal flow, but during my big audit and restructuring of the Learn area, these were the articles that I deemed potentially useful, but not suitable for the normal flow. CSS multi-col is not an essential layout technique that you should advise all beginners to learn, in the same way that flexbox or grid is, for example.

files/sidebars/learnsidebar.yaml
```diff
@@ -128,6 +128,13 @@ sidebar:
       - /Learn_web_development/Core/CSS_layout/Test_your_skills/Responsive_design
       - /Learn_web_development/Core/CSS_layout/Fundamental_Layout_Comprehension
       - /Learn_web_development/Core/CSS_layout/Test_your_skills
+      - title: Additional_tutorials
+        details: closed
+        children:
+          - /Learn_web_development/Core/CSS_layout/Multiple-column_Layout
+          - /Learn_web_development/Core/CSS_layout/Practical_positioning_examples
```
bsmth, line 135: Positioning one is still okay for "normal flow", IMO. Would you consider the others "legacy" techniques instead of "additional"? 

Overall, I think it's alright to have them on the same level as the other guides, but maybe down lower in the sidebar (more niche) with their status / context in the article intro.

chrisdavidmills: > Positioning one is still okay for "normal flow", IMO. Would you consider the others "legacy" techniques instead of "additional"?

I'm not sure what you mean by "Positioning one is still okay for "normal flow"". Can you rephrase?

They are not all legacy techniques.
 
> Overall, I think it's alright to have them on the same level as the other guides, but maybe down lower in the sidebar (more niche) with their status / context in the article intro.

I disagree. I think having them at the same level in the nav will make people think they are part of the normal flow, and surprise them when they don't get to them via prev/next.

bsmth: >I disagree. I think having them at the same level in the nav will make people think they are part of the normal flow, and surprise them when they don't get to them via prev/next.

I forgot about prev/next continuity expectation -> you have the best feel for that in terms of the pathway, so deferring to you on that.

>Positioning one is still okay for "normal flow" Can you rephrase?

I thought this guide has some good info, it could be useful to have it up with the others, IMO, but up to you 👍🏻

bsmth (approved):
+1 for adding these docs into the sidebars, tnx

Josh-Cena:
Yeah also a +1 for me

PR #41209, opened 2025-09-19
FF144 Relnote: ScreenOrientation.lock() and unlock()

hamishwillee:
FF144 supports [`ScreenOrientation.lock()`](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock) and [`ScreenOrientation.unlock()`](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/unlock) in https://bugzilla.mozilla.org/show_bug.cgi?id=1983483

This adds a release note.

Related docs work can be tracked in #41142


pepelsbey (approved):
Looks good! Thank you :)

PR #41215, opened 2025-09-19
Alphabetizing roles that cannot be named

### LINKED ISSUE #41214: aria-label list of prohibited roles not correctly sorted
The aria-label page includes a list of [prohibited roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label#associated_roles) that is meant to be alphabetically sorted but is not.

[code lines](https://github.com/mdn/content/blob/5e3308146f872a133a3221e612ea8c6ee85dd77d/files/en-us/web/accessibility/aria/reference/attributes/aria-label/index.md?plain=1#L87-L102)

aardrian:
Closes https://github.com/mdn/content/issues/41214

Also matches the sorting of https://w3c.github.io/aria/#namefromprohibited


Josh-Cena (approved):
Nice, thank you

PR #41218, opened 2025-09-19
Remove more load event handlers

Josh-Cena:
I missed these in my last PR because of the format being different


bsmth (approved):
Thanks 👍🏻

PR #41223, opened 2025-09-19
Add deprecation notice for audioprocess event

sczesny:
### Description

Added deprecation notice for audioprocess event

### Motivation

It's declared as deprecated in the video element reference but not in the audio one


hamishwillee (approved):
@sczesny Thanks very much, and welcome to MDN. FYI I took the opportunity to turn these tables into definition lists as well, as this is "on balance" preferred on MDN

PR #41225, opened 2025-09-20
Fix "element" to "event"

mfuji09:
### Description

Fix "element" to "event".

### Motivation

The `cancel` is an event, not an element.


wbamberg (approved):
👍 thanks @mfuji09 !

PR #41226, opened 2025-09-20
Fix "the grid or row" with "the column or row"

mfuji09:
### Description

Fix "the grid or row" with "the column or row"

### Motivation

Perhaps this wants to say "column or row", directions which a grid spans.


chrisdavidmills (approved):
Yes! This is exactly what it meant to say. Thanks for the fix, @mfuji09.

PR #41227, opened 2025-09-20
remove `text-autospace` from a list of properties not supported.

mfuji09:
### Description

remove `text-autospace` from a list of properties not supported.

### Motivation

`text-autospace` is already supported by some browsers and is documented, so it should be removed from the list.


Josh-Cena (approved):
Good catch, thanks

PR #41228, opened 2025-09-20
Fix typo in Web Permissions API

meduzen:
### Description

Replace an underscore by a space in _Clipboard_API_.

Now [online](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API#permission-aware_apis):

> - [Clipboard_API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations): `clipboard-read`, `clipboard-write`

This PR:

> - [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations): `clipboard-read`, `clipboard-write`


wbamberg (approved):
👍 thank you!

PR #41231, opened 2025-09-21
Fix - Added missing demo choices for `font-variant-numeric`

### LINKED ISSUE #41198: [CSS] demo for font-variant-numeric does not show all values
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric

Try it

Examples are missing for the variants for the display of fractions and for `ordinal`

All relevant variants should have a demo.

_No response_

_No response_

ragul1697:
### Description

Fixes #41198 - Added missing `font-variant-numeric` choices in interactive example


chrisdavidmills (changes requested):
Thanks for adding these, @ragul1697! I don't have any problem with the new choices you've added. Just a couple of thoughts about what we could do to improve the look of the interactive demo (it currently scrolls a lot, which looks slightly unwieldy):

- Maybe you could try applying the "taller" keyword to the macro, in the same way "shorter" is applied [here](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Code_examples#interactive_examples).
- I think it makes sense to have "normal" and "ordinal" at the top of the list. Maybe you could put the others in alphabetical order? Or did you have a particular reason for the order you chose?

Let me know what you think. Thanks again!

ragul1697:
Hi @chrisdavidmills, Sorry it's been a while. I have kept the order followed in the syntax section. Other than "normal", "ordinal" and "slashed-zero", rest all falls under categories. So, if we arrange them alphabetically, there will be some confusion.

Josh-Cena:
The example text doesn't distinguish between some of these values. You at least need `1st` and `3/4`. Very possibly you even need to select another font.

chrisdavidmills:
> The example text doesn't distinguish between some of these values. You at least need `1st` and `3/4`. Very possibly you even need to select another font.

@ragul1697 Yeah, this is a good point. I am happy with the other points I brought up earlier, but you should really consider updating the example text to cover fractions/ordinals.

ragul1697:
@chrisdavidmills I have made the changes. One thing is we need to switch font families to show all the differences. The value "ordinal" has no effect with "Fira Sans" font family.

chrisdavidmills:
> @chrisdavidmills I have made the changes. One thing is we need to switch font families to show all the differences. The value "ordinal" has no effect with "Fira Sans" font family.

This article says that Source Sans Pro supports these features: https://www.petermekhaeil.com/til/css-font-variant-numeric/. I've checked out his demo, and it seems to support most of them.

ragul1697:
@chrisdavidmills I have updated the font family to "Source Sans Pro". Is this good or is there anything needs to be changed?

chrisdavidmills:
> @chrisdavidmills I have updated the font family to "Source Sans Pro". Is this good or is there anything needs to be changed?

@ragul1697 I've just managed to get the live preview page to update (there has been a problem with this functionality, but it is fixed now).

It doesn't seem to be working very well, and I'm not sure why.

If you go to the example in the "Examples" section, edit it to add some zeros, fractions, etc., and then change the `font-variant-numeric` value, many of the values work.

Some more testing needed, I think.

files/en-us/web/css/font-variant-numeric/index.md
```diff
@@ -39,23 +59,29 @@ font-variant-numeric: oldstyle-nums;
...
 
 ```css interactive-example
 @font-face {
-  font-family: "Fira Sans";
+  font-family: "Source Sans Pro";
   src:
-    local("FiraSans-Regular"),
-    url("/shared-assets/fonts/FiraSans-Regular.woff2") format("woff2");
+    local("SourceSansPro-Regular"),
+    url("/shared-assets/fonts/SourceSansPro-Regular.otf") format("otf");
```
Josh-Cena:
```suggestion
    url("/shared-assets/fonts/SourceSansPro-Regular.otf") format("opentype");
```

Josh-Cena:
@chrisdavidmills This LGTM; PTAL

chrisdavidmills:
> @chrisdavidmills This LGTM; PTAL

That does work better; well spotted, @Josh-Cena.

The interactive example now works better. I wonder if we should remove the examples that have no visible effect, namely `tabular-nums` and `stacked-fractions`? We don't need to include all values, and there is no point in including them if they don't demonstrate anything.

Josh-Cena:
It is nearly always the case that `normal` clashes with one of the specific variants. Usually numbers don't have the same width so `pnum` is the default. However SSP makes `tnum` the default. We still should show both `proportional-nums` and `tabular-nums` though because they are supposed to contrast, and IMO `normal` should stay too since it's the starting point. The fact that they  are the same is just a coincidence of the font choice and doesn't pertain to the example. Same stuff with `lnum` and `onum`. For `afrac` it does appear that the font doesn't support it, so I'm fine to remove.

chrisdavidmills:
I'm not disputing any of that, but that is the kind of stuff that ought to be explained in the article, if we are going to explain it. For people who don't understand this stuff, it may still seem weird if three of the options appear to do the same thing (or as an alternative viewport, two of them don't seem to work).

But I don't feel that strongly about it. It's still much better than it was. Let's remove the one option and get it merged.

chrisdavidmills (approved):
Approving so we can merge it once the change is done.

Josh-Cena:
Thanks @chrisdavidmills for the review

PR #41237, opened 2025-09-22
FF144 Relnote+Doc: Cross origin iframe requires user interaction for top navigation

hamishwillee:
FF144 aligns with Chrome and Safari such that a cross origin iframe now requires sticky activation in order to redirect the top level page via `window.top.location` (in https://bugzilla.mozilla.org/show_bug.cgi?id=1419501).

This PR adds a release note and also a section in the iframe docs explaining the expected behavior as it is now. 

Related docs work can be tracked in https://github.com/mdn/content/issues/41138

### DIFF
--- a/files/en-us/mozilla/firefox/releases/144/index.md
+++ b/files/en-us/mozilla/firefox/releases/144/index.md
@@ -61,7 +61,7 @@ Firefox 144 is the current [Beta version of Firefox](https://www.firefox.com/en-
 #### Media, WebRTC, and Web Audio
 
 - Cross-origin {{htmlelement("iframe")}}s now require either user interaction ({{glossary("sticky activation")}}) or explicit permission to redirect the top-level page using `window.top.location`.
-  See [`Window.top` navigation](/en-US/docs/Web/HTML/Reference/Elements/iframe#window.top_navigation) for more information. ([Firefox bug 1419501](https://bugzil.la/1419501)).
+  See [`Window.top` navigation](/en-US/docs/Web/HTML/Reference/Elements/iframe#top_navigation_in_cross-origin_frames) for more information. ([Firefox bug 1419501](https://bugzil.la/1419501)).
 - {{domxref("RTCDataChannel")}} instances are now [transferrable objects](/en-US/docs/Web/API/Web_Workers_API/Transferable_objects), and hence can be passed to [workers](/en-US/docs/Web/API/Worker). ([Firefox bug 1209163](https://bugzil.la/1209163)).
 - The [`closing` event](/en-US/docs/Web/API/RTCDataChannel/closing_event) and the `onclosing()` event handler are now supported on the {{domxref("RTCDataChannel")}} interface. ([Firefox bug 1611953](https://bugzil.la/1611953)).
 - The {{domxref("MediaDevices/getUserMedia","getUserMedia()")}} and {{domxref("MediaDevices/getDisplayMedia","getDisplayMedia()")}} methods of the {{domxref("MediaDevices")}} interface now support the [`resizeMode`](/en-US/docs/Web/API/MediaTrackConstraints#resizemode) constraint.
--- a/files/en-us/mozilla/firefox/releases/144/index.md
+++ b/files/en-us/mozilla/firefox/releases/144/index.md
@@ -60,7 +60,8 @@ Firefox 144 is the current [Beta version of Firefox](https://www.firefox.com/en-
 
 #### Media, WebRTC, and Web Audio
 
-- Cross-origin {{htmlelement("iframe")}}s require either user interaction or explicit permission in order to redirect to another page using `window.top.location`. See [`Window.top` navigation](/en-US/docs/Web/HTML/Reference/Elements/iframe#window.top_navigation) in the `<iframe>` topic for more information. ([Firefox bug 1419501](https://bugzil.la/1419501)).
+- Framebusting Intervention: Cross-origin {{htmlelement("iframe")}}s require either user interaction or explicit permission in order to redirect the top level page using `window.top.location`, aligning behavior with Chrome and Safari.
+  See [`Window.top` navigation](/en-US/docs/Web/HTML/Reference/Elements/iframe#window.top_navigation) for more information. ([Firefox bug 1419501](https://bugzil.la/1419501)).
 - {{domxref("RTCDataChannel")}} instances are now [transferrable objects](/en-US/docs/Web/API/Web_Workers_API/Transferable_objects), and hence can be passed to [workers](/en-US/docs/Web/API/Worker). ([Firefox bug 1209163](https://bugzil.la/1209163)).
 - The [`closing` event](/en-US/docs/Web/API/RTCDataChannel/closing_event) and the `onclosing()` event handler are now supported on the {{domxref("RTCDataChannel")}} interface. ([Firefox bug 1611953](https://bugzil.la/1611953)).
 - The {{domxref("MediaDevices/getUserMedia","getUserMedia()")}} and {{domxref("MediaDevices/getDisplayMedia","getDisplayMedia()")}} methods of the {{domxref("MediaDevices")}} interface now support the [`resizeMode`](/en-US/docs/Web/API/MediaTrackConstraints#resizemode) constraint.
--- a/files/en-us/mozilla/firefox/releases/144/index.md
+++ b/files/en-us/mozilla/firefox/releases/144/index.md
@@ -60,6 +60,7 @@ Firefox 144 is the current [Beta version of Firefox](https://www.firefox.com/en-
 
 #### Media, WebRTC, and Web Audio
 
+- Cross-origin {{htmlelement("iframe")}}s require either user interaction or explicit permission in order to redirect to another page using `window.top.location`. See [`Window.top` navigation](/en-US/docs/Web/HTML/Reference/Elements/iframe#window.top_navigation) in the `<iframe>` topic for more information. ([Firefox bug 1419501](https://bugzil.la/1419501)).
 - {{domxref("RTCDataChannel")}} instances are now [transferrable objects](/en-US/docs/Web/API/Web_Workers_API/Transferable_objects), and hence can be passed to [workers](/en-US/docs/Web/API/Worker). ([Firefox bug 1209163](https://bugzil.la/1209163)).
 - The [`closing` event](/en-US/docs/Web/API/RTCDataChannel/closing_event) and the `onclosing()` event handler are now supported on the {{domxref("RTCDataChannel")}} interface. ([Firefox bug 1611953](https://bugzil.la/1611953)).
 - The {{domxref("MediaDevices/getUserMedia","getUserMedia()")}} and {{domxref("MediaDevices/getDisplayMedia","getDisplayMedia()")}} methods of the {{domxref("MediaDevices")}} interface now support the [`resizeMode`](/en-US/docs/Web/API/MediaTrackConstraints#resizemode) constraint.
--- a/files/en-us/mozilla/firefox/releases/144/index.md
+++ b/files/en-us/mozilla/firefox/releases/144/index.md
@@ -60,7 +60,7 @@ Firefox 144 is the current [Beta version of Firefox](https://www.firefox.com/en-
 
 #### Media, WebRTC, and Web Audio
 
-- Framebusting Intervention: Cross-origin {{htmlelement("iframe")}}s require either user interaction or explicit permission in order to redirect the top level page using `window.top.location`, aligning behavior with Chrome and Safari.
+- Cross-origin {{htmlelement("iframe")}}s now require either user interaction ({{glossary("sticky activation")}}) or explicit permission to redirect the top-level page using `window.top.location`.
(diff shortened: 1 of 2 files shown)


files/en-us/web/html/reference/elements/iframe/index.md
```diff
@@ -186,7 +186,21 @@ With the DOM {{domxref("HTMLIFrameElement")}} object, scripts can access the {{d
...
 
-Script access to a frame's content is subject to the [same-origin policy](/en-US/docs/Web/Security/Same-origin_policy). Scripts cannot access most properties in other `window` objects if the script was loaded from a different origin, including scripts inside a frame accessing the frame's parent. Cross-origin communication can be achieved using {{domxref("Window.postMessage()")}}.
+Script access to a frame's content is subject to the [same-origin policy](/en-US/docs/Web/Security/Same-origin_policy).
+Scripts cannot access most properties in other `window` objects if the script was loaded from a different origin, including scripts inside a frame accessing the frame's parent.
+Cross-origin communication can be achieved using {{domxref("Window.postMessage()")}}.
+
+### `Window.top` navigation
+
+Scripts running in a same-origin frame can access the {{domxref("Window.top")}} property and set {{domxref("Window.location","window.top.location")}} to redirect the top level page to a new location.
+This is referred to as a "top-navigation" or "framebusting".
```
dipikabh:
```suggestion
Scripts running in a same-origin frame can access the {{domxref("Window.top")}} property and set {{domxref("Window.location","window.top.location")}} to redirect the top-level page to a new location.
This behavior is referred to as "top navigation".
```
- Should it be "top navigation" without the hyphen?
- should we differentiate between "top navigation" and "framebusting"?
  - From https://en.wikipedia.org/wiki/Framekiller it seems that framebusting is a defensive security technique where a site forces itself out of an iframe "to avoid being embedded", that is, it's a special case of top navigation?
  - should we clarify that framebusting is an "attempt to force" a top navigation (and that the "intervention" is browser's behavior for preventing cross-origin framebusting)?

That is to say, AFAICT the same-origin top nav default behavior is not framebusting. Can you you check?

files/en-us/web/html/reference/elements/iframe/index.md
```diff
@@ -186,7 +186,21 @@ With the DOM {{domxref("HTMLIFrameElement")}} object, scripts can access the {{d
...
+
+A cross-origin frame is only allowed to redirect the page using `top` if the frame has {{glossary("Sticky activation")}}.
+If a top-navigation is blocked the browser may prompt for user permission to redirect, or it may simply report the error in the developer console.
+What this means is that you won't be able to load a cross-origin frame and immediately redirect to a new page — the user must first (or previously) have interacted with the frame or granted permission to redirect.
```
dipikabh: Would it be accurate to call this part "framebusting intervention"?
```suggestion
A cross-origin frame is allowed to redirect the top-level page using `top` only if the frame has {{glossary("sticky activation")}}.
If a top navigation is blocked, browsers may either prompt for user permission to redirect or report the error in the developer console (or both).
This restriction by browsers is called a _framebusting intervention_.
What this means is that a cross-origin frame can't immediately redirect the top-level page — the user must have previously interacted with the frame or granted permission to redirect.
```

files/en-us/web/html/reference/elements/iframe/index.md
```diff
@@ -186,7 +186,21 @@ With the DOM {{domxref("HTMLIFrameElement")}} object, scripts can access the {{d
...
+
+A sandboxed frame will block all top-navigations unless the values [`allow-top-navigation`](#allow-top-navigation) or [`allow-top-navigation-by-user-activation`](#allow-top-navigation-by-user-activation) are set.
+Note that top-navigation permissions are inherited, so a nested frame will only be allowed to perform top-navigation if its parents are able to.
```
dipikabh: - first para rephrasing is mostly stylistic
- second para edit is for changing the placement of "only", plus parents -> parent frames just to be clearer
```suggestion
A sandboxed frame blocks all top navigation unless the `sandbox` attribute values are set to [`allow-top-navigation`](#allow-top-navigation) or [`allow-top-navigation-by-user-activation`](#allow-top-navigation-by-user-activation).
Note that top-navigation permissions are inherited, so a nested frame can perform a top navigation only if its parent frames are also allowed to.
```

files/en-us/web/html/reference/elements/iframe/index.md
```diff
@@ -186,7 +186,21 @@ With the DOM {{domxref("HTMLIFrameElement")}} object, scripts can access the {{d
 
 From the inside of a frame, a script can get a reference to its parent window with {{domxref("window.parent")}}.
 
-Script access to a frame's content is subject to the [same-origin policy](/en-US/docs/Web/Security/Same-origin_policy). Scripts cannot access most properties in other `window` objects if the script was loaded from a different origin, including scripts inside a frame accessing the frame's parent. Cross-origin communication can be achieved using {{domxref("Window.postMessage()")}}.
+Script access to a frame's content is subject to the [same-origin policy](/en-US/docs/Web/Security/Same-origin_policy).
+Scripts cannot access most properties in other `window` objects if the script was loaded from a different origin, including scripts inside a frame accessing the frame's parent.
+Cross-origin communication can be achieved using {{domxref("Window.postMessage()")}}.
+
+### `Window.top` navigation
```
dipikabh: Wondering if this title is intuitive enough. Or if we can call it something like "Top navigation in cross-origin frames" or any other phrase that devs are likely to search for to land on this section

files/en-us/mozilla/firefox/releases/144/index.md
```diff
@@ -60,6 +60,8 @@ Firefox 144 is the current [Beta version of Firefox](https://www.firefox.com/en-
 
 #### Media, WebRTC, and Web Audio
```
dipikabh, line 61: should this update go under HTML?

files/en-us/mozilla/firefox/releases/144/index.md
```diff
@@ -60,6 +60,8 @@ Firefox 144 is the current [Beta version of Firefox](https://www.firefox.com/en-
...
 
+- Framebusting Intervention: Cross-origin {{htmlelement("iframe")}}s require either user interaction or explicit permission in order to redirect the top level page using `window.top.location`, aligning behavior with Chrome and Safari.
```
dipikabh: - we don't usually include such info ", aligning behavior with Chrome and Safari"...but may be okay to keep in this case
- should we add the glossary link to sticky activation after "user interaction"
```suggestion
- Framebusting intervention is now supported, where cross-origin {{htmlelement("iframe")}}s require either user interaction or explicit permission to redirect the top-level page using `window.top.location`, aligning behavior with Chrome and Safari.
```
OR
```suggestion
- Cross-origin {{htmlelement("iframe")}}s now require either user interaction (sticky activation) or explicit permission to redirect the top-level page using `window.top.location`. This restriction is called framebusting intervention.
```

files/en-us/mozilla/firefox/releases/144/index.md
```diff
@@ -60,6 +60,8 @@ Firefox 144 is the current [Beta version of Firefox](https://www.firefox.com/en-
...
+  See [`Window.top` navigation](/en-US/docs/Web/HTML/Reference/Elements/iframe#window.top_navigation) for more information. ([Firefox bug 1419501](https://bugzil.la/1419501)).
```
dipikabh: would need to be updated if we update the section title

dipikabh (changes requested):
I've left some questions and suggestions - let me know if any feel off the mark

files/en-us/mozilla/firefox/releases/144/index.md
hamishwillee: I decided to modify this in line with your comments - i.e. not aligning stuff. Remove framebusting - its a new term for me and I don't think it will help most users.


```suggestion
- Cross-origin {{htmlelement("iframe")}}s now require either user interaction ({{glossary("sticky activation")}}) or explicit permission to redirect the top-level page using `window.top.location`.
```

files/en-us/web/html/reference/elements/iframe/index.md
hamishwillee: Yeah, that's better



```suggestion
### Top navigation in cross-origin frames
```

files/en-us/web/html/reference/elements/iframe/index.md
hamishwillee: Thanks. TL;DR I'm going to take your text then perhaps come back and explain frame busting. I'm not sure how much needs to be said about it really - I only mentioned it originally to make the behaviour easier to find in search.

---

> Should it be "top navigation" without the hyphen? 

Yes. I didn't understand the grammar rule :-0

> should we differentiate between "top navigation" and "framebusting"?

Yes, they are different.

Framebusting is indeed a "legacy" security technique that allows a page to prevent itself from being embedded. Essentially it checks if it is the location of `Window.top.location` and if not it is a frame, so it forces itself to the top level. So framebusting is a technique that may use top navigation.

A framebusting intervention is something that prevents or restricts framebusting. IN this case it is forcing the user to have interacted with the page or acknowledged a prompt.

SO what confused me is why you would intervene in something designed for a good purpose.
The first reason is that there are better ways to prevent yourself being embedded now - e.g. CORS.
The second is that framebusting gives a false sense of security - a malicious page can (apparently) hijack the redirect of top navigation to redirect to whatever page it likes - say a near duplicate of the embedded page - the user is then on a phishing site. With this intervention the redirect can't just happen. The user will need to have given permission or clicked in the frame.


> That is to say, AFAICT the same-origin top nav default behavior is not framebusting. Can you you check?

Same-origin or cross origin don't affect framebusting - framebusting is just a page setting itself as the main window.

files/en-us/web/html/reference/elements/iframe/index.md
hamishwillee: Yes, you got it. I made some tweaks, but that's good.

files/en-us/mozilla/firefox/releases/144/index.md
hamishwillee: I don't _think_ so. While it is associated with an element, the thing that is relevant to it being used is the use of the API.

files/en-us/mozilla/firefox/releases/144/index.md
hamishwillee:
```suggestion
  See [`Window.top` navigation](/en-US/docs/Web/HTML/Reference/Elements/iframe#top_navigation_in_cross-origin_frames) for more information. ([Firefox bug 1419501](https://bugzil.la/1419501)).
```

hamishwillee:
OK @dipikabh I think this is good to look at again. Decided not to go into the detail of framebusting :-)

files/en-us/mozilla/firefox/releases/144/index.md
```diff
@@ -60,6 +60,8 @@ Firefox 144 is the current [Beta version of Firefox](https://www.firefox.com/en-
 
 #### Media, WebRTC, and Web Audio
 
+- Cross-origin {{htmlelement("iframe")}}s now require either user interaction ({{glossary("sticky activation")}}) or explicit permission to redirect the top-level page using `window.top.location`.
+  See [`Window.top` navigation](/en-US/docs/Web/HTML/Reference/Elements/iframe#top_navigation_in_cross-origin_frames) for more information. ([Firefox bug 1419501](https://bugzil.la/1419501)).
```
dipikabh: nit with the link text

```suggestion
  See [Top navigation in cross-origin frames](/en-US/docs/Web/HTML/Reference/Elements/iframe#top_navigation_in_cross-origin_frames) for more information. ([Firefox bug 1419501](https://bugzil.la/1419501)).
```

files/en-us/web/html/reference/elements/iframe/index.md
```diff
@@ -186,7 +186,22 @@ With the DOM {{domxref("HTMLIFrameElement")}} object, scripts can access the {{d
...
+Cross-origin communication can be achieved using {{domxref("Window.postMessage()")}}.
+
+### Top navigation in cross-origin frames
+
+Scripts running in a same-origin frame can access the {{domxref("Window.top")}} property and set {{domxref("Window.location","window.top.location")}} to redirect the top-level page to a new location.
+This behavior is referred to as "top navigation".
+
+A cross-origin frame is allowed to redirect the top-level page using `top` only if the frame has {{glossary("sticky activation")}}.
+If a top navigation is blocked, browsers may either prompt for user permission to redirect or report the error in the developer console (or both).
+This restriction by browsers is called a _framebusting intervention_.
```
dipikabh: on re-reading, i think we may not need those "a"s

```suggestion
A cross-origin frame is allowed to redirect the top-level page using `top` only if the frame has {{glossary("sticky activation")}}.
If top navigation is blocked, browsers may either prompt for user permission to redirect or report the error in the developer console (or both).
This restriction by browsers is called _framebusting intervention_.
```

dipikabh (approved):
Thanks a lot, Hamish! This is looking better now, just a few nits

hamishwillee:
Thanks @dipikabh !

PR #41248, opened 2025-09-22
Consistently mention that Map/Set compare object keys by reference

chronoDave:
### Description

The documentation does not clarify that `has()` uses object reference for comparison despite `WeakMap` only accepting objects and non-registered symbols.

Object references are counter-intuitive to those unfamiliar with the language.

### Motivation

The WeakMap documentation page highlights [caching](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#caching) as an use-case. Naively caching objects might result in unexpected behaviour, e.g.:

```TS
import fsp from 'fs/promises';

const cache = new WeakMap();

for await (const file of fsp.glob('images/**/*')) {
	const raw = await fsp.readFile(file); // Uint8Array

	let cached = cache.get(raw);
	if (!cached) {
		cached = file;
		cache.set(raw, cached);
	} else {
		console.log(cached); // Will never reach
	}
}
```

This implementation will never cache, even if the images are duplicates.

### Additional details

- [WeakMap.prototype.has() (MDN)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap/has)
- [3.4 WeakMap.prototype.has (key) (ECMAScript® 2026 Language Specification)](https://tc39.es/ecma262/multipage/keyed-collections.html#sec-weakmap.prototype.has)
- [7.2.9 SameValue (x, y) (ECMAScript® 2026 Language Specification)](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-samevalue)


Josh-Cena (approved):
Thanks. Your idea makes total sense to me, but we would need to do this for every single collection's CRUD interface, so I've done that.

PR #41251, opened 2025-09-22
Tweak operator precedence example

### LINKED ISSUE #41243: Javascript Guide: Short-circuiting example is a bit confusing
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence

Short-circuiting

Short-circuiting is illustrated using the following two example code blocks:

> ```js
> function A() { console.log('called A'); return false; }
> function B() { console.log('called B'); return false; }
> function C() { console.log('called C'); return true; }
> 
> console.log(C() || B() && A());
> 
> // Logs:
> // called C
> // true
> ```
> 
> Only C() is evaluated, despite && having higher precedence. This does not mean that || has higher precedence in this case — it's exactly because (B() && A()) has higher precedence that causes it to be neglected as a whole. If it's re-arranged as:
> 
> ```js
> console.log(A() && C() || B());
> // Logs:
> // called A
> // called B
> // false
> ```
> 
> Then the short-circuiting effect of && would only prevent C() from being evaluated, but because A() && C() as a whole is false, B() would still be evaluated.
> 
> However, note that short-circuiting does not change the final evaluation outcome. [...]

While it is technically correct, it can be a bit confusing for the reader: The text between those two code blocks says that the expression that is passed to `console.log()` is being "re-arranged", and in a mathematical sense, when "re-arranging" terms etc., the value is not changed. However in this example here, the value of the "re-arranged" expressioin is changed. Why not leave the expression mathematically the same, like in the following code block?

(This would also go nicely with the last sentence in the quoted text above, that says that "short-circuiting does not change the final evaluation outcome".)

> ```js
> console.log(A() && B() || C());
> // Logs:
> // called A
> // called C
> // true
> ```
> 
> Then the short-circuiting effect of && would only prevent B() from being evaluated. Because A() && B() as a whole is false, C() would still be evaluated.
> 
> However, note that short-circuiting does not change the final evaluation outcome. [...]

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/javascript/reference/operators/operator_precedence`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/operators/operator_precedence/index.md

</details>

jeetsingh008:
## Description
This PR updates a confusing code example used to explain JavaScript's logical operator short-circuiting. The previous "re-arranged" example has been replaced with a clearer one that more accurately demonstrates the concept.

## Motivation
The previous example was confusing for learners because the "re-arranged" expression resulted in a different boolean outcome, which obscured the core concept of short-circuiting. The new example (A() && B() || C()) directly and clearly illustrates how the && operator short-circuits without this confusion, improving the article's educational value.

## Additional details
N/A

## Related issues and pull requests
Fixes #41243

### DIFF
--- a/files/en-us/web/javascript/reference/operators/operator_precedence/index.md
+++ b/files/en-us/web/javascript/reference/operators/operator_precedence/index.md
@@ -221,11 +221,11 @@ console.log(C() || B() && A());
 Only `C()` is evaluated, despite `&&` having higher precedence. This does not mean that `||` has higher precedence in this case — it's exactly _because_ `(B() && A())` has higher precedence that causes it to be neglected as a whole. If it's re-arranged as:
 
 ```js-nolint
-console.log(A() && C() || B());
+console.log(A() && B() || C());
 // Logs:
 // called A
-// called B
-// false
+// called C
+// true
 ```
 
 Then the short-circuiting effect of `&&` would only prevent `C()` from being evaluated, but because `A() && C()` as a whole is `false`, `B()` would still be evaluated.
--- a/files/en-us/web/javascript/reference/operators/operator_precedence/index.md
+++ b/files/en-us/web/javascript/reference/operators/operator_precedence/index.md
@@ -228,7 +228,7 @@ console.log(A() && B() || C());
 // true
 ```
 
-Then the short-circuiting effect of `&&` would only prevent `C()` from being evaluated, but because `A() && C()` as a whole is `false`, `B()` would still be evaluated.
+Then the short-circuiting effect of `&&` would only prevent `B()` from being evaluated, but because `A() && B()` as a whole is `false`, `C()` would still be evaluated.
 
 However, note that short-circuiting does not change the final evaluation outcome. It only affects the evaluation of _operands_, not how _operators_ are grouped — if evaluation of operands doesn't have side effects (for example, logging to the console, assigning to variables, throwing an error), short-circuiting would not be observable at all.
 


files/en-us/web/javascript/reference/operators/operator_precedence/index.md
```diff
@@ -221,11 +221,11 @@ console.log(C() || B() && A());
...
+console.log(A() && B() || C());
 // Logs:
 // called A
-// called B
-// false
+// called C
+// true
 ```
 
 Then the short-circuiting effect of `&&` would only prevent `C()` from being evaluated, but because `A() && C()` as a whole is `false`, `B()` would still be evaluated.
```
Josh-Cena:
```suggestion
Then the short-circuiting effect of `&&` would only prevent `B()` from being evaluated, but because `A() && B()` as a whole is `false`, `C()` would still be evaluated.
```

Josh-Cena (approved):
Thanks. Congrats on your first PR here.

Next time, please read the issue more carefully because you didn't change the text explaining this example. The issue clearly suggested what to change for the paragraph below.

jeetsingh008:
Thank you so much for merging my first PR here! I really appreciate you pointing that out. You're right, I focused only on the code and overlooked the necessary changes to the paragraph. I'll make sure to be more thorough and read the entire issue description carefully on future contributions. Thanks again!

PR #41252, opened 2025-09-22
Reference for stage 3 upsert

Josh-Cena:
Part of https://github.com/mdn/content/issues/41134

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/map/getorinsert/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/map/getorinsert/index.md
@@ -11,7 +11,9 @@ sidebar: jsref
 
 {{SeeCompatTable}}
 
-The **`getOrInsert()`** method of {{jsxref("Map")}} instances returns the value corresponding to the key in this `Map`, or inserts and returns a default value if there is none.
+The **`getOrInsert()`** method of {{jsxref("Map")}} instances returns the value corresponding to the specified key in this `Map`. If the key is not present, it inserts a new entry with the key and a given default value, and returns the inserted value.
+
+If the computation of the default value is expensive, consider using {{jsxref("Map.prototype.getOrInsertComputed()")}} instead, which takes a callback to compute the default value only if it's actually needed.
 
 {{InteractiveExample("JavaScript Demo: Map.prototype.getOrInsert()")}}
 
@@ -54,7 +56,7 @@ if (map.has(key)) {
 }
 ```
 
-Alternatively, you can also see patterns like the following (which is slightly less reliable if `null` or `undefined` are valid values in your map):
+It is also similar to the following pattern (which is slightly less reliable if `null` or `undefined` are valid values in your map):
 
 ```js
 map.set(key, map.get(key) ?? defaultValue);
@@ -87,6 +89,8 @@ document.body.dataset.theme = options.get("theme");
 
 ## See also
 
+- [Polyfill of `Map.prototype.getOrInsert` in `core-js`](https://github.com/zloirock/core-js#map-upsert)
+- [es-shims polyfill of `Map.prototype.getOrInsert`](https://www.npmjs.com/package/map.prototype.getorinsert)
 - {{jsxref("Map")}}
 - {{jsxref("Map.prototype.get()")}}
 - {{jsxref("Map.prototype.set()")}}
--- a/files/en-us/web/javascript/reference/global_objects/map/getorinsert/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/map/getorinsert/index.md
@@ -1,0 +1,94 @@
+---
+title: Map.prototype.getOrInsert()
+short-title: getOrInsert()
+slug: Web/JavaScript/Reference/Global_Objects/Map/getOrInsert
+page-type: javascript-instance-method
+status:
+  - experimental
+browser-compat: javascript.builtins.Map.getOrInsert
+sidebar: jsref
+---
+
+{{SeeCompatTable}}
+
+The **`getOrInsert()`** method of {{jsxref("Map")}} instances returns the value corresponding to the key in this `Map`, or inserts and returns a default value if there is none.
+
+{{InteractiveExample("JavaScript Demo: Map.prototype.getOrInsert()")}}
+
+```js interactive-example
+const map = new Map([["bar", "foo"]]);
+console.log(map.getOrInsert("bar", "default"));
+// Expected output: "foo"
+
+console.log(map.getOrInsert("baz", "default"));
+// Expected output: "default"
+```
+
+## Syntax
+
+```js-nolint
+getOrInsert(key, defaultValue)
+```
+
+### Parameters
+
+- `key`
+  - : The key of the value to return from the `Map` object. Object keys are compared by [reference](/en-US/docs/Glossary/Object_reference), not by value.
+- `defaultValue`
+  - : The value to insert and return if the key is not already present in the `Map` object.
+
+### Return value
+
+The value associated with the specified key in the `Map` object. If the key can't be found, `defaultValue` is inserted and returned.
+
+## Description
+
+The `getOrInsert()` method is equivalent to the following:
+
+```js
+if (map.has(key)) {
+  return map.get(key);
+} else {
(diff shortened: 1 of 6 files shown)


files/en-us/web/javascript/reference/global_objects/map/index.md
```diff
@@ -353,6 +353,10 @@ These properties are defined on `Map.prototype` and shared by all `Map` instance
   - : Calls `callbackFn` once for each key-value pair present in the `Map` object, in insertion order. If a `thisArg` parameter is provided to `forEach`, it will be used as the `this` value for each callback.
 - {{jsxref("Map.prototype.get()")}}
   - : Returns the value corresponding to the key in this `Map`, or `undefined` if there is none.
+- {{jsxref("Map.prototype.getOrInsert()")}}
+  - : Returns the value corresponding to the key in this `Map`, or inserts and returns a default value if there is none.
+- {{jsxref("Map.prototype.getOrInsertComputed()")}}
+  - : Returns the value corresponding to the key in this `Map`, or inserts and returns a default value computed from a callback if there is none.
```
hamishwillee: Maybe addition of specified and given as show. The reason is the "default" - if we don't specify that it is "given" it might be a global default. Something to think about, and if you agree, roll pattern out everywhere.

```suggestion
  - : Returns the value corresponding to the specified key in this `Map`, or inserts and returns the given default value if the key is not present.
- {{jsxref("Map.prototype.getOrInsertComputed()")}}
  - : Returns the value corresponding to the specified key in this `Map`, or inserts and returns a default value computed from a callback if the key is not present.
```

files/en-us/web/javascript/reference/global_objects/weakmap/getorinsert/index.md
```diff
@@ -0,0 +1,77 @@
...
+page-type: javascript-instance-method
+status:
+  - experimental
+browser-compat: javascript.builtins.WeakMap.getOrInsert
+sidebar: jsref
+---
+
+{{SeeCompatTable}}
+
+The **`getOrInsert()`** method of {{jsxref("WeakMap")}} instances returns the value corresponding to the key in this `WeakMap`, or inserts and returns a default value if there is none.
```
hamishwillee:
```suggestion
The **`getOrInsert()`** method of {{jsxref("WeakMap")}} instances returns the value corresponding to a specified key in this `WeakMap`, or inserts and returns a given default value if the key is not present.
```

files/en-us/web/javascript/reference/global_objects/weakmap/getorinsert/index.md
```diff
@@ -0,0 +1,77 @@
...
+## Syntax
+
+```js-nolint
+getOrInsert(key, defaultValue)
+```
+
+### Parameters
+
+- `key`
+  - : The key of the value to return from the `WeakMap` object. Must be either an object or a [non-registered symbol](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#shared_symbols_in_the_global_symbol_registry). Object keys are compared by [reference](/en-US/docs/Glossary/Object_reference), not by value.
```
hamishwillee, line 38: Should we link to https://pr41252.review.mdn.allizom.net/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#key_equality as well since the comparison also covers things like NaN?

files/en-us/web/javascript/reference/global_objects/weakmap/getorinsert/index.md
hamishwillee: Or maybe.... (very pedantic, but this version make it clear that the default value is inserted "associated with the key)

```suggestion
The **`getOrInsert()`** method of {{jsxref("WeakMap")}} instances returns the value corresponding to the key in this `WeakMap`. If the key is not present it inserts a new element with the key and the given default value, and returns the inserted value.
```

files/en-us/web/javascript/reference/global_objects/map/getorinsert/index.md
```diff
@@ -0,0 +1,94 @@
...
+```js
+if (map.has(key)) {
+  return map.get(key);
+} else {
+  map.set(key, defaultValue);
+  return defaultValue;
+}
+```
+
+Alternatively, you can also see patterns like the following (which is slightly less reliable if `null` or `undefined` are valid values in your map):
```
hamishwillee: The "see patterns" thing doesn't quite work. Maybe.

```suggestion
It is also similar to the following pattern (which is slightly less reliable if `null` or `undefined` are valid values in your map):
```

files/en-us/web/javascript/reference/global_objects/map/getorinsertcomputed/index.md
```diff
@@ -0,0 +1,81 @@
...
+- `key`
+  - : The key of the element to return from the `Map` object. Object keys are compared by [reference](/en-US/docs/Glossary/Object_reference), not by value.
+- `callback`
+  - : A function that returns the value to insert and return if the key is not already present in the `Map` object. The function is called with the following argument:
+    - `key`
+      - : The same key that was passed to `getOrInsertComputed()`.
+
+### Return value
+
+The value associated with the specified key in the `Map` object. If the key can't be found, `callback(key)` is inserted and returned.
```
hamishwillee: Nit. (everywhere)
```suggestion
The value associated with the specified key in the `Map` object. If the key can't be found, the result of `callback(key)` is inserted and returned.
```

files/en-us/web/javascript/reference/global_objects/map/getorinsertcomputed/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Examples
+
+### Avoiding unnecessary default computation
```
hamishwillee, line 52: So it never occurred to me seeing `getOrInsert()` that the `defaultValue` might be a function, and hence something that could be calculated.

I think this is good, but we should mention in  `getOrInsert()` description that it is possible for the value to be calculated, but not recommended because it is always calculated - "if you need to calculate the value instead use `getInsert...`. Then you could link here.

files/en-us/web/javascript/reference/global_objects/map/getorinsert/index.md
```diff
@@ -0,0 +1,94 @@
...
+## Specifications
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
```
zloirock, line 91: Please, add links to `core-js` polyfills of this, https://github.com/zloirock/core-js#map-upsert

files/en-us/web/javascript/reference/global_objects/weakmap/getorinsert/index.md
Josh-Cena: This is standard verbiage for every `Map` method so I'm not super motivated to entertain alternative wording, although it's not a bad idea.

files/en-us/web/javascript/reference/global_objects/map/getorinsert/index.md
```diff
@@ -11,7 +11,9 @@ sidebar: jsref
 
 {{SeeCompatTable}}
 
-The **`getOrInsert()`** method of {{jsxref("Map")}} instances returns the value corresponding to the key in this `Map`, or inserts and returns a default value if there is none.
+The **`getOrInsert()`** method of {{jsxref("Map")}} instances returns the value corresponding to the specified key in this `Map`. If the key is not present, it inserts a new entry with the key and a given default value, and returns the inserted value.
+
+If the computation of the default value is expensive, consider using {{jsxref("Map.prototype.getOrInsertComputed()")}} instead, which takes a callback to compute the default value only if it's actually needed.
```
hamishwillee, line 16: Yeah, this all much better

files/en-us/web/javascript/reference/global_objects/weakmap/getorinsert/index.md
hamishwillee: OK. Perhaps something to do at some later point

hamishwillee (approved):
Thanks very much!

PR #41254, opened 2025-09-23
FF144 CSSStyleProperties docs

hamishwillee:
FF144 (and Safari) implement `CSSStyleProperties`. This is a derived class of `CSSStyleDeclaration` that represents the inline or computed styles on an element. It exists because  `CSSStyleDeclaration` is now going to be a base class from which specific types of style and rule information will be applied, and only include the relevant information for that context.

Anyway, this adds the new class and fixes up all the places that return it. It also updates  `CSSStyleDeclaration`  to note that this is now a base class but didn't used to be.

There is quite a lot of duplication. Mostly because there was before I started too. I wasn't sure how much to include in say `HTMLElement.style` and how much to have in the `CSSStyleProperties` as common. Generally I have kept about the same level in each doc as originally.

Related docs work can be tracked in #41133


files/en-us/web/api/cssstyleproperties/index.md
```diff
@@ -0,0 +1,169 @@
...
+```js
+function getPopulatedProperties(styleProperties) {
+  const camelProps = Object.getOwnPropertyNames(
+    Object.getPrototypeOf(styleProperties),
+  ).filter((prop) => prop !== "constructor" && styleProperties[prop] !== "");
+
+  for (const prop of camelProps) {
+    try {
+      log(`${prop} = ${styleProperties[prop]}`);
+    } catch {// error}
```
saschanaz: Syntax error:

```suggestion
    } catch {/* error */}
```

But enumerating computed properties freezes the demo for a few seconds...

files/en-us/web/api/cssstyleproperties/index.md
```diff
@@ -0,0 +1,169 @@
...
+  clearLog();
+});
+```
+
+#### JavaScript
+
+The code first defines the function we're going to use to get the style properties of our element "elt".
+This uses [`Object.prototype.getOwnPropertyNames()`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames) to get all properties defined on the object and then filters those that have a value of `""` or are named "constructor".
+
+Note that we need to use `getOwnPropertyNames()` rather than {{domxref("CSSStyleDeclaration.getPropertyValue()")}} because we want to also list the camel case properties, which are non-enumerable.
```
saschanaz: But camel case properties would have the same values, why would the caller want to get the same values multiple times?

files/en-us/web/api/cssstyleproperties/cssfloat/index.md
```diff
@@ -0,0 +1,47 @@
...
+title: "CSSStyleProperties: cssFloat property"
+short-title: cssFloat
+slug: Web/API/CSSStyleProperties/cssFloat
+page-type: web-api-instance-property
+browser-compat: api.CSSStyleProperties.cssFloat
+---
+
+{{APIRef("CSSOM")}}
+
+The **`cssFloat`** property of the {{domxref("CSSStyleProperties")}} interface returns the result of invoking {{DOMxRef("CSSStyleDeclaration.getPropertyValue()","getPropertyValue()")}} on the object with `float` as an argument.
```
saschanaz: Which, as a result, gets the value of the CSS `float` property. Maybe that's easier to understand 👀 (Calling `getPropertyValue("float")` is more of a detail, IMO)

files/en-us/web/api/cssstyleproperties/index.md
hamishwillee: They wouldn't - it was mostly to show that you can. Perhaps we should just list the `getPropertyValue()`?

saschanaz: Perhaps something like this would be more interesting in case of `.style`:

```js
// Iterates only the assigned properties
for (const property of element.style) {
    // needs to use getPropertyValue in case `property` is a custom property
    console.log(property, element.style.getPropertyValue(property))
}
```

I don't find it very interesting to iterate over the result of getComputedStyle because it will iterate every single existing CSS property, but could be interesting as a way to see how many CSS properties exist?

files/en-us/web/api/cssstyleproperties/cssfloat/index.md
```diff
@@ -0,0 +1,47 @@
+---
+title: "CSSStyleProperties: cssFloat property"
```
saschanaz, line 2: BTW, we already have CSSStyleDeclaration.cssFloat doc, should we just move that to this new directory?

hamishwillee: No, because `CSSStyleDeclaration.cssFloat` is a different property - it was added in different versions to a different interface. Doing it this way makes it possible for users to track the movement of the property and where it is present and not present in various types.

saschanaz: In that case CSSStyleDeclaration.cssFloat should be marked as removed and the page should be eventually removed too.

saschanaz: (the current table there looks wrong too, it exists in Firefox..)

files/en-us/web/api/cssstyleproperties/index.md
hamishwillee: Yeah, I modified it a bit like this. Essentially the same example as before but using `getPropertyValue` to get the enumerated styles on an element (of each type). It's faster now we're not including the camelcase versions.

files/en-us/web/api/cssstyleproperties/cssfloat/index.md
hamishwillee: Correct. I have got a pending BCD update for updates which includes remove of  `CSSStyleDeclaration.cssFloat` (see https://github.com/mdn/browser-compat-data/pull/27975)

> (the current table there looks wrong too, it exists in Firefox..)

That is marked like this
```

            "firefox": {
              "version_added": "1",
              "version_removed": "17"
            },
```

Can you tell me the "right" informaiton.

Note that this won't disappear until chrome migrates AND no one has used it for 2 years.

files/en-us/web/api/cssstyledeclaration/index.md
```diff
@@ -7,20 +7,27 @@ browser-compat: api.CSSStyleDeclaration
...
+- {{domxref("CSSPageDescriptors")}} — Styles for CSS at-rules.
 
-- Via {{DOMxRef("HTMLElement.style")}}, which deals with the inline styles of a single element (e.g., `<div style="…">`).
-- Via the {{DOMxRef("CSSStyleSheet")}} API. For example, `document.styleSheets[0].cssRules[0].style` returns a `CSSStyleDeclaration` object on the first CSS rule in the document's first stylesheet.
-- Via {{DOMxRef("Window.getComputedStyle()")}}, which exposes the `CSSStyleDeclaration` object as a **read-only** interface.
+The interface exposes style information and various style-related methods and properties.
+For example, it provides {{DOMxRef("CSSStyleDeclaration/getPropertyValue","getPropertyValue()" )}} for getting the value of a dash-named CSS property, such as `border-top`, which can't be directly accesses using dot notation because of the hyphens in its name.
+
+> ![!NOTE]
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
> [!NOTE]
```

files/en-us/web/api/cssstyleproperties/cssfloat/index.md
hamishwillee: Thanks. Doing this.


```suggestion
The **`cssFloat`** property of the {{domxref("CSSStyleProperties")}} interface returns CSS `float` property.

This is the result of invoking {{DOMxRef("CSSStyleDeclaration.getPropertyValue()","getPropertyValue()")}} on the object with `float` as an argument.
```

files/en-us/web/api/window/getcomputedstyle/index.md
```diff
@@ -27,45 +22,54 @@ getComputedStyle(element, pseudoElt)
...
+
+However, the two objects have different purposes:
+
+- The object from `getComputedStyle()` is read-only, and should be used to inspect the element's style — including those set by a `<style>` element or an external stylesheet.
+- The `element.style` object can be used to **set** styles on that element, or inspect styles directly added to it from JavaScript manipulation or the global `style` attribute.
+
+Note that:
+
+- The returned {{DOMxRef("CSSStyleProperties")}} object contains active values for CSS property _longhand_ names as well as shorthand names.
+  For example, the returned object contains entries for {{cssxref("border-bottom-width")}} in addition to the {{cssxref("border-width")}} and {{cssxref("border")}} [shorthand property names](/en-US/docs/Web/CSS/CSS_cascade/Shorthand_properties).
```
hamishwillee: @saschanaz Thanks for all your help on this. Its now "first draft" meaning I have touched every document I expect to, but might still move some things around. 

IN particular, this document had lots of information about `CSSStyleDeclaration` which wasn't in the other ones, but might apply to all. 

I tested for computed and inline styles and if you enumerate the properties you get all longhand properties only - not shorhand. However if you do `getPropertyValue` on a shorthand property then you will get a result. Is that expected? How would you describe that/why aren't they enumerated?

files/en-us/web/api/window/getcomputedstyle/index.md
```diff
@@ -27,45 +22,54 @@ getComputedStyle(element, pseudoElt)
...
+- During [CSS transitions](/en-US/docs/Web/CSS/CSS_transitions), `getComputedStyle()` returns the original property value in Firefox, but the final property value in WebKit.
```
hamishwillee: This is compatibility information. It should be in the browser compatibility data unless it is so out of date as to be irrelevant.
1. Which of these two things original/final is expected by the spec?
2. Do we know the versions that this affects?

hamishwillee: Or is it no longer relevant?

files/en-us/web/api/window/getcomputedstyle/index.md
```diff
@@ -27,45 +22,54 @@ getComputedStyle(element, pseudoElt)
...
+- In Firefox, properties with the value `auto` return the used value, not the value `auto`.
```
hamishwillee: This is compatibility information. It should be in the browser compatibility data unless it is so out of date as to be irrelevant.
1. So this is incorrect behaviour. Is it still true? If so, what version did it come in. If not, which versions was it true for.

files/en-us/web/api/window/getcomputedstyle/index.md
```diff
@@ -27,45 +22,54 @@ getComputedStyle(element, pseudoElt)
...
+  So if you apply `top:auto` and `bottom:0` on an element with `height:30px` and a containing block of `height:100px`, Firefox's computed style for `top` returns `70px`, as 100 − 30 = 70.
+- For compatibility reasons, serialized color values are expressed as [`rgb()`](/en-US/docs/Web/CSS/color_value/rgb) colors if the alpha channel value is exactly `1`, and `rgba()` colors otherwise.
+  In both cases, legacy syntax is used, with commas as separators (for example `rgb(255, 0, 0)`).
```
hamishwillee: Same compat questions for this. I'm really asking if we can delete this text, and if so, what notes we might add in BCD to capture it. Also if all of this is true of the `CSSStyleProperties` returned by all interfaces, or JUST for the computed styles?

files/en-us/web/api/cssstylerule/style/index.md
```diff
@@ -8,40 +8,40 @@ browser-compat: api.CSSStyleRule.style
...
+A {{domxref("CSSStyleProperties")}}.
 
-- computed flag
-  - : Unset.
-- declarations
-  - : The declared declarations in the rule, in the order they were specified, shorthand properties expanded to longhands.
-- parent CSS rule
-  - : The context object, which is an alias for [`this`](https://heycam.github.io/webidl/#this).
-- owner node
-  - : Null.
```
hamishwillee, line 24: FYI, none of this is relevant to third parties. I have captured it as "inline".

This is effectively identical to the element.style right? My understanding is that the shorthand is expanded for inline styles but NOT for computed styles, so I should probably copy that information here right. I do wonder if perhaps I should put it in `CSSStyleProperties` "for inline styles shorthand is expanded but not for computed styles" or whatever.

files/en-us/web/api/cssstyledeclaration/index.md
```diff
@@ -7,20 +7,27 @@ browser-compat: api.CSSStyleDeclaration
 
 {{APIRef("CSSOM")}}
 
-The **`CSSStyleDeclaration`** interface represents an object that is a CSS declaration block, and exposes style information and various style-related methods and properties.
+The **`CSSStyleDeclaration`** interface is the base class for objects that represent CSS declaration blocks with different supported sets of CSS style information:
```
hamishwillee, line 10: Updated to reflect this is now a base class. Does it make sense (entirely?)

I was trying to capture that this is now a base class, that it might still be returned on some browsers, but that your code doesn't need to change - the result of this would be similar - just the older browser would return an object with some spurious properties in some contexts.

hamishwillee (commented):
@saschanaz I think this is ready for another look "in total" since I have touched all files now.
Does it make sense? Is there common information that should go in `CSSStyleProperties`? Or is it "about right"?

The whole change affected more APIs than I expected at first!

files/en-us/web/api/cssstyleproperties/cssfloat/index.md
saschanaz: version_removed should be the same version when CSSStyleProperties.cssFloat is added.

files/en-us/web/api/window/getcomputedstyle/index.md
saschanaz: This is getting out of my expertise, but per my spec reading, setProperty [only sets the longhand values](https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setproperty) and then getPropertyValue [computes the value from longhand values](https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-getpropertyvalue). Thus the shorthands are never set directly as CSS declaration.

cc @emilio or @dshin-moz who know better than me.

files/en-us/web/api/window/getcomputedstyle/index.md
saschanaz: TBH I have zero idea. Delegating to Emilio or David.

files/en-us/web/api/window/getcomputedstyle/index.md
emilio: Yes, that's expected. Enumeration is over longhands (as otherwise you enumerate over the same property multiple times). When you query or set a shorthand, it does get synthesized from the longhands or expanded to them, respectively. The CSSOM spec should be pretty clear about how that happens.

files/en-us/web/api/window/getcomputedstyle/index.md
emilio: I think it's pretty interoperable now. Transitions return the interpolating value everywhere iirc?

files/en-us/web/api/window/getcomputedstyle/index.md
saschanaz: Huh. I get the opposite, when setting `margin: auto` and doing `getComputedStyle(document.body).margin`:

* Blink gives `0px`
* Gecko gives `auto`.

emilio: This is only true for some properties, and should happen across browsers. Maybe it should read:

 * Some properties will not return the computed but the used style, see https://drafts.csswg.org/cssom/#resolved-values.

Or something like that.

files/en-us/web/api/window/getcomputedstyle/index.md
emilio: I think this is true for most used colors if they use rgb or so, but not if they use the newer color() function.

saschanaz: That's described in the spec: https://drafts.csswg.org/css-color/#serializing-alpha-values, but only applies to (traditional) sRGB color. In 2025 there are a bunch of other color syntax that are not traditional sRGB - lab()/lch()/oklab()/oklch()/color().

(Edit: `color(srgb 0 0 0)` would not be serialized to rgb() as color() is treated separately)

files/en-us/web/api/cssstylerule/style/index.md
saschanaz: > FYI, none of this is relevant to third parties. I have captured it as "inline".

It is in the spec: https://drafts.csswg.org/cssom/#the-cssstylerule-interface

But also not sure it's worth in MDN. Probably not?

>My understanding is that the shorthand is expanded for inline styles but NOT for computed styles

AFAICT it's all expanded, `getComputedStyle(document.body).borderLeftWidth` is `1px` if you set `border: 1px solid black`.

files/en-us/web/api/window/getcomputedstyle/index.md
hamishwillee: Thanks. I've removed this point, but linked to https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Value_processing#resolved_value elsewhere.

files/en-us/web/api/cssstylerule/style/index.md
hamishwillee: > But also not sure it's worth in MDN. Probably not?

MDN documents a human readable and website developer-centric version of the spec. Those flags are in the spec, but they aren't in the IDL or visible to web developers directly. So we don't document them like this, but instead comment on their effects. In this case, that the object is readonly and inline. 

   Upshot, no, I don't think we should capture them directly, but if the owner node and parent CSS rule had website developer direct impact, we'd capture those effects.

> AFAICT it's all expanded, getComputedStyle(document.body).borderLeftWidth is 1px if you set border: 1px solid black.

Thanks. Of course you're right. Everything expanded everywhere.

files/en-us/web/api/window/getcomputedstyle/index.md
hamishwillee: OK, I'm going to delete this then on the assumption that we document the spec. IFF you know the versions at which we became compliant they should go in the compatibility data, but I don't plan try hunt that down for this job.

Note, I did add a comment above in the text though

> The distinction that the object contains [resolved values](/en-US/docs/Web/CSS/CSS_cascade/Value_processing#resolved_value) is important.
For most properties, in particular those that depend on layout such as `display`, `font-size`, or `line-height`, the resolved value is the computed value.
For properties that do depend on layout, the used value may differ slightly from the computed value, and this is what is returned as the resolved value.
_For an animating property value it will be the computed value at the current point in the animation._

files/en-us/web/api/window/getcomputedstyle/index.md
hamishwillee: Thanks, I have tried to capture this in a new section above

```
### Color values

For compatibility reasons, serialized color values s...
```

If it is not quite right, then can you suggest an improvement?

I think this `getComputedStyle()` is now good.

bsmth:
How's this looking @hamishwillee - are you still waiting on review from @saschanaz?

hamishwillee:
@bsmth This is ready for your review. @saschanaz May wish to comment further, but has been of great help and I think his issues are addressed.

files/en-us/web/api/cssstyledeclaration/index.md
```diff
@@ -7,20 +7,27 @@ browser-compat: api.CSSStyleDeclaration
...
+- {{domxref("CSSPageDescriptors")}} — Styles for CSS at-rules.
 
-- Via {{DOMxRef("HTMLElement.style")}}, which deals with the inline styles of a single element (e.g., `<div style="…">`).
-- Via the {{DOMxRef("CSSStyleSheet")}} API. For example, `document.styleSheets[0].cssRules[0].style` returns a `CSSStyleDeclaration` object on the first CSS rule in the document's first stylesheet.
-- Via {{DOMxRef("Window.getComputedStyle()")}}, which exposes the `CSSStyleDeclaration` object as a **read-only** interface.
+The interface exposes style information and various style-related methods and properties.
+For example, it provides {{DOMxRef("CSSStyleDeclaration/getPropertyValue","getPropertyValue()" )}} for getting the value of a dash-named CSS property, such as `border-top`, which can't be directly accesses using dot notation because of the hyphens in its name.
```
bsmth:
```suggestion
For example, it provides {{DOMxRef("CSSStyleDeclaration/getPropertyValue","getPropertyValue()" )}} for getting the value of a dash-named CSS property, such as `border-top`, which can't be directly accessed using dot notation because of the hyphens in its name.
```

files/en-us/web/api/cssstyleproperties/index.md
```diff
@@ -0,0 +1,239 @@
...
+```js
+log("\nChange color to green and log borderTop");
+elementStyle.borderTopColor = "green";
+log(`"borderTop" = '${elementStyle.borderTop}'`);
+```
+
+#### Results
+
+The results are shown below.
+Note how changing the longhande property is reflected in the corresponding shorthand property.
```
bsmth:
```suggestion
Note how changing the longhand property is reflected in the corresponding shorthand property.
```

files/en-us/web/api/cssstyleproperties/index.md
```diff
@@ -0,0 +1,239 @@
...
+const reload = document.querySelector("#reset");
+
+reload.addEventListener("click", () => {
+  clearLog();
+});
+```
+
+#### JavaScript
+
+The code first defines the function we're going to use to enumerate the properties of our element `elt`.
```
bsmth:
```suggestion
The code first defines the function we're going to use to enumerate the properties of our element with the `elt` ID.
```

files/en-us/web/api/cssstylerule/style/index.md
```diff
@@ -8,40 +8,40 @@ browser-compat: api.CSSStyleRule.style
 
 {{ APIRef("CSSOM") }}
 
-The read-only **`style`** property is the {{ domxref("CSSStyleDeclaration") }} interface for the declaration block of the {{ DOMXref("CSSStyleRule") }}.
+The read-only **`style`** property is a {{domxref("CSSStyleProperties")}} object represents the inline styles of a style rule ({{DOMXref("CSSStyleRule")}}).
```
bsmth:
```suggestion
The read-only **`style`** property is a {{domxref("CSSStyleProperties")}} object that represents the inline styles of a style rule ({{DOMXref("CSSStyleRule")}}).
```

files/en-us/web/api/window/getcomputedstyle/index.md
```diff
@@ -27,45 +22,63 @@ getComputedStyle(element, pseudoElt)
...
+
+The method returns a live read-only {{DOMxRef("CSSStyleProperties")}} object containing the [resolved values](/en-US/docs/Web/CSS/CSS_cascade/Value_processing#resolved_value) of all CSS properties of an element, after applying active stylesheets and resolving any computation those values may contain.
+
+The returned object can be used to inspect the element's styles — including those set inline, using a `<style>` element, or via an external stylesheet.
+Since the object read-only you can't use it to set the styles of an element.
+However because it is "live", if you update the element styles using another API (such as {{domxref("HTMLElement.style")}}), the returned object will be updated with the corresponding resolved value.
+
+The distinction that the object contains [resolved values](/en-US/docs/Web/CSS/CSS_cascade/Value_processing#resolved_value) is important.
+For most properties, in particular those that depend on layout such as `display`, `font-size`, or `line-height`, the resolved value is the computed value.
+For properties that do depend on layout, the used value may differ slightly from the computed value, and this is what is returned as the resolved value.
```
bsmth: Could you check?

```suggestion
For properties that do not depend on layout, the used value may differ slightly from the computed value, and this is what is returned as the resolved value.
```

files/en-us/web/api/window/getcomputedstyle/index.md
```diff
@@ -27,45 +22,63 @@ getComputedStyle(element, pseudoElt)
...
+For an animating property value it will be the computed value at the current point in the animation.
```
bsmth:
```suggestion
For an animating property value, it will be the computed value at the current point in the animation.
```

files/en-us/web/api/cssstyledeclaration/index.md
```diff
@@ -7,20 +7,27 @@ browser-compat: api.CSSStyleDeclaration
...
+
+> [!NOTE]
+> Earlier versions of the specification used `CSSStyleDeclaration` to represent all CSS declaration blocks, and some browsers and browser versions may still do so (check the browser compatibility tables for the above APIs).
+> Generally the same website code will "work" in both old and new versions, but some of the properties returned in a `CSSStyleDeclaration` may not be relevant in a particular context.
```
bsmth:
```suggestion
> Generally the same website code will be functional in both old and new versions, but some properties returned in a `CSSStyleDeclaration` may not be relevant in a particular context.
```

files/en-us/web/api/cssstyledeclaration/index.md
```diff
@@ -7,20 +7,27 @@ browser-compat: api.CSSStyleDeclaration
 
 {{APIRef("CSSOM")}}
 
-The **`CSSStyleDeclaration`** interface represents an object that is a CSS declaration block, and exposes style information and various style-related methods and properties.
+The **`CSSStyleDeclaration`** interface is the base class for objects that represent CSS declaration blocks with different supported sets of CSS style information:
 
-A `CSSStyleDeclaration` object can be exposed using three different APIs:
+- {{domxref("CSSStyleProperties")}} — CSS styles declared in stylesheet ({{domxref("CSSStyleRule.style")}}), inline styles for an element such as {{DOMxRef("HTMLElement/style","HTMLElement")}}, {{domxref("SVGElement/style","SVGElement")}}, and {{domxref("MathMLElement/style","MathMLElement")}}, or the computed style for an element returned by {{DOMxRef("Window.getComputedStyle()")}}.
+- {{domxref("CSSPageDescriptors")}} — Styles for CSS at-rules.
```
bsmth:
```suggestion
- {{domxref("CSSPageDescriptors")}} — Styles for CSS [at-rules](/en-US/docs/Web/CSS/CSS_syntax/At-rule).
```

files/en-us/web/api/cssstyleproperties/cssfloat/index.md
```diff
@@ -0,0 +1,49 @@
...
+short-title: cssFloat
+slug: Web/API/CSSStyleProperties/cssFloat
+page-type: web-api-instance-property
+browser-compat: api.CSSStyleProperties.cssFloat
+---
+
+{{APIRef("CSSOM")}}
+
+The **`cssFloat`** property of the {{domxref("CSSStyleProperties")}} interface returns CSS `float` property.
```
bsmth: "the" here? Also maybe link this one instead of the later one on L16?
```suggestion
The **`cssFloat`** property of the {{domxref("CSSStyleProperties")}} interface returns the CSS {{cssxref("float")}} property.
```

files/en-us/web/api/cssstyleproperties/index.md
```diff
@@ -0,0 +1,239 @@
...
+The HTML defines a {{htmlelement("div")}} with a number of styles set, nested within another that sets the `font-weight`.
+There are also buttons to get the inline styles and computed styles for the element (and hidden code for a reset button and logging).
+
+```html
+<div style="font-weight: bold;">
+  <div style="border-top: 1px solid blue; color: red;" id="elt">
+    An example div
+  </div>
+</div>
+<button id="inline_style" type="button">Inline Style</button>
```
bsmth: Not seeing much use for the buttons in this example or have I missed something?

bsmth (approved):
Thanks, Hamish! I've applied some obvious fixes, but will leave other suggestions for you to check first. Looking good otherwise, so a +1 for when you're ready 👍🏻

files/en-us/web/api/window/getcomputedstyle/index.md
hamishwillee: What I had is correct, but I am going to tidy slightly. 

```suggestion
For properties that depend on layout, the used value may differ slightly from the computed value, and this is what is returned as the resolved value.
```

From https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Value_processing#resolved_value :

> The used value is the property's value after all calculations have been performed on the [computed value](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Value_processing#computed_value) and it has been refined with layout-specific details (e.g., percentages resolved to actual pixel values).

So used value depends on layout, and this may differ from the computed value. It is also mentioned in the spec.

files/en-us/web/api/cssstyleproperties/index.md
hamishwillee: No, this was rubbish. I have rewritten it. Thank you.

hamishwillee:
Thanks so much for the review @bsmth - in particular for catching that example, which I have fixed up. Happy enough with it to merge.

PR #41261, opened 2025-09-23
Declared missing variable i in Example 1.

GordonRamsay-689:
In "Example 1: height and width", in the for loop iterating over arrImages an undeclared variable "i" was being used, breaking the example.

### Description

A declaration for the variable "i" has  been added just before it is used, as well as appropriate incrementation after each use in the loop.

### Motivation

A broken example creates confusion.


Josh-Cena (approved):
Great catch, thank you

anazoka2019-cmyk (commented):
Ghsj cknvfjn

PR #41262, opened 2025-09-23
Deprecate onunload event in body element documentation

sczesny:
Mark the 'onunload' event as deprecated.

### Description

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/body#onunload leads to https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event which is deprecated.

### Motivation

Deprecated event should be marked as such.


chrisdavidmills (approved):
@sczesny seems reasonable; thanks!

Issue #32884, opened 2024-03-29
Add onfocusin and onfocusout to event handler content attributes


josepharhar:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes

### What specific section or headline is this issue about?

"The event handler attributes:"

### What information was incorrect, unhelpful, or incomplete?

This list does not include "onfocusin" and "onfocusout"

### What did you expect to see?

"onfocusin" and "onfocusout" to be in the list like "onfocus" is

### Do you have any supporting links, references, or citations?

I am adding these event handler content attributes to the spec here: https://github.com/whatwg/html/pull/10235

### Do you have anything more you want to share?

I'm not sure if there's anywhere else on MDN that these event handler content attributes should be enumerated, but when I searched for "onfocus" this was the only page I found.

hamishwillee:
Looking at Chrome (at least) the events exist ([`focusout`](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusout_event), [`focusin`](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusin_event)), but the associated `onXxx()` handlers do not - which makes sense, because that is what you're adding in https://github.com/whatwg/html/pull/10235.

MDN does not document "stuff" until there is at least one browser implementation, so this needs to be on hold for a bit.

When this is implemented, yes then https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes would need to be updated. The two events would also need to be updated in their syntax boxes to add the `onXzzz()` handler. You can see what this would look like by comparing with the [`focus` event syntaxt box.](https://developer.mozilla.org/en-US/docs/Web/API/Element/focus_event#syntax)

Josh-Cena:
Here are a list of places to update: https://developer.mozilla.org/en-US/docs/Web/API/Window#listening_for_events_on_window, https://developer.mozilla.org/en-US/docs/Web/API/HTMLBodyElement#event_handlers, https://developer.mozilla.org/en-US/docs/Web/API/Element/focusin_event, https://developer.mozilla.org/en-US/docs/Web/API/Element/focusout_event, because we are adding the IDL properties as well

Josh-Cena:
https://github.com/mdn/content/pull/41264 takes care of the HTML attributes, ~~but the IDL properties are still unimplemented anywhere~~ actually they are in Safari.

PR #41264, opened 2025-09-24
Update event handler attributes in global attributes

sczesny:
### Description

Removed: `onautocomplete`, `onautocompleteerror`, `onshow`, `onsort` due to not being documented anywhere (only in w3 drafts)
Removed: `onabort` due to not being a global event listener (https://html.spec.whatwg.org/multipage/dom.html#global-attributes)

Added: `onafterscriptexecute`, `onanimationcancel`, `onanimationend`, `onanimationiteration`, `onanimationstart`, `onauxclick`, `onbeforeinput`, `onbeforematch`, `onbeforescriptexecute`, `onbeforetoggle`, `onbeforexrselect`, `oncommand`, `oncompositionend`, `oncompositionstart`, `oncompositionupdate`, `oncontentvisibilityautostatechange`, `oncontextlost`, `oncontextrestored`, `oncopy`, `oncut`, `onfocusin`, `onfocusout`, `onformdata`, `onfullscreenchange`, `onfullscreenerror`, `ongesturechange`, `ongestureend`, `ongesturestart`, `ongotpointercapture`, `onlostpointercapture`, `onpaste`, `onpointercancel`, `onpointerdown`, `onpointerenter`, `onpointerleave`, `onpointermove`, `onpointerout`, `onpointerover`, `onpointerrawupdate`, `onpointerup`, `onscrollend`, `onscrollsnapchange`, `onscrollsnapchanging`, `onsecuritypolicyviolation`, `onselectstart`, `onslotchange`, `ontouchcancel`, `ontouchend`, `ontouchmove`, `ontouchstart`, `ontransitioncancel`, `ontransitionend`, `ontransitionrun`, `ontransitionstart`, `onwebkitmouseforcechanged`, `onwebkitmouseforcedown`, `onwebkitmouseforceup`, `onwebkitmouseforcewillbegin`, `onwheel`

Kept: `onblur`, `oncancel`, `oncanplay`, `oncanplaythrough`, `onchange`, `onclick`, `onclose`, `oncontextmenu`, `oncuechange`, `ondblclick`, `ondrag`, `ondragend`, `ondragenter`, `ondragleave`, `ondragover`, `ondragstart`, `ondrop`, `ondurationchange`, `onemptied`, `onended`, `onerror`, `onfocus`, `oninput`, `oninvalid`, `onkeydown`, `onkeypress`, `onkeyup`, `onload`, `onloadeddata`, `onloadedmetadata`, `onloadstart`, `onmousedown`, `onmouseenter`, `onmouseleave`, `onmousemove`, `onmouseout`, `onmouseover`, `onmouseup`, `onmousewheel`, `onpause`, `onplay`, `onplaying`, `onprogress`, `onratechange`, `onreset`, `onresize`, `onscroll`, `onseeked`, `onseeking`, `onselect`, `onstalled`, `onsubmit`, `onsuspend`, `ontimeupdate`, `ontoggle`, `onvolumechange`, `onwaiting`

Moved everything to a list with a new section called **List of global event handler attributes**, which is sorted alphabetically.

### Motivation

The current list is out of date. It contains non-existent and deprecated events.

### Additional details

The new list includes all attributes from html.spec.whatwg.org in addition to deprecated `onmousewheel`, `afterscriptexecute` and `beforescriptexecute` event handlers.


Josh-Cena:
Many of the attributes you added are not global attributes. They are listeners for `window`, so they can be used on `<body>`. So they are documented in https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/body#event_attributes instead.

sczesny:
@Josh-Cena You're right. Removed these listeners.

files/en-us/web/html/reference/global_attributes/index.md
```diff
@@ -14,7 +14,7 @@ In addition to the basic HTML global attributes, the following global attributes
...
 - The ARIA [`role`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles) attribute and the multiple [`aria-*`](/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes) states and properties, used for ensuring accessibility.
-- The [event handler](/en-US/docs/Web/HTML/Reference/Attributes#event_handler_attributes) attributes: `onabort`, `onautocomplete`, `onautocompleteerror`, `onblur`, `oncancel`, `oncanplay`, `oncanplaythrough`, `onchange`, `onclick`, `onclose`, `oncontextmenu`, `oncuechange`, `ondblclick`, `ondrag`, `ondragend`, `ondragenter`, `ondragleave`, `ondragover`, `ondragstart`, `ondrop`, `ondurationchange`, `onemptied`, `onended`, `onerror`, `onfocus`, `oninput`, `oninvalid`, `onkeydown`, `onkeypress`, `onkeyup`, `onload`, `onloadeddata`, `onloadedmetadata`, `onloadstart`, `onmousedown`, `onmouseenter`, `onmouseleave`, `onmousemove`, `onmouseout`, `onmouseover`, `onmouseup`, `onmousewheel`, `onpause`, `onplay`, `onplaying`, `onprogress`, `onratechange`, `onreset`, `onresize`, `onscroll`, `onseeked`, `onseeking`, `onselect`, `onshow`, `onsort`, `onstalled`, `onsubmit`, `onsuspend`, `ontimeupdate`, `ontoggle`, `onvolumechange`, `onwaiting`.
```
Josh-Cena, line 17: This list is almost directly taken from https://html.spec.whatwg.org/multipage/dom.html#global-attributes. While some of them are deprecated, it's important to include all of them because they indicate whether your HTML is syntactically conforming or not. You also need to include things like `onvolumechange`, which, although only make sense on certain elements, can be applied to all of them.

sczesny: The deprecated events I let out are: `onDOMActivate`, `onDOMMouseScroll`, `onMozMousePixelScroll`, `onafterscriptexecute`, `onbeforescriptexecute`, `onkeypress`, `onmousewheel`

Isn't `onvolumechange` an event from the `HTMLMediaElement` Interface, that only applies to `<audio>` and `<video>` elements?

Josh-Cena: You can check the spec link above. `onvolumechange` is a valid attribute (conformance-wise) on all elements, although it only makes sense on media elements.

`onDOMActivate`, `onDOMMouseScroll`, `onMozMousePixelScroll` aren't real attributes AFAIK, but correct me if I'm wrong. Definitely fine to leave out since they aren't in the HTML spec. The `afterscriptexecute` and `beforescriptexecute` events are being removed altogether. `onkeypress` and `onmousewheel` need to stay because they are valid HTML attributes, but we can turn this into a list so you can add `{{deprecated_inline}}` if you really wish (I personally don't think it matters).

sczesny:
@Josh-Cena I moved everything to a **List of global event handler attributes** below and included attributes from [html.spec.whatwg.org](https://html.spec.whatwg.org/multipage/dom.html#global-attributes), `onmousewheel`, `afterscriptexecute` and `beforescriptexecute`.
Let me know if I'm missing something.

Josh-Cena:
I am surveying the whole domain of this problem. I assume that HTML attributes always have IDL counterparts, so I can query "all event attributes" in JS, with ``console.log(Object.getOwnPropertyNames(HTMLElement.prototype).filter((x) => x.startsWith("on")).sort().join("\n"))``. I query this in all browsers and merge their results. This list is, unsurprisingly, a strict superset of the HTML spec. I then get the list on MDN from before this PR and after this PR. There are a total of 132 entries to look at.

- For those in the HTML spec, plus `onmousewheel` since all other mouse events are here:
  - 57 are unchanged in the PR, therefore no additional investigation is needed.
  - 15 are added in the PR, which is great.
- For those not in browsers:
  - 4 are removed in the PR: `onautocompleteerror`, `onautocomplete`, `onshow`, `onsort`
  - 16 are added in the PR, and very careful review is needed:
    - `onafterscriptexecute`
    - `onbeforescriptexecute`
    - `oncompositionend`
    - `oncompositionstart`
    - `oncompositionupdate`
    - `onfocusin`
    - `onfocusout`
    - `onfullscreenchange`
    - `onfullscreenerror`
    - `ongesturechange`
    - `ongestureend`
    - `ongesturestart`
    - `ontouchcancel`
    - `ontouchend`
    - `ontouchmove`
    - `ontouchstart`
- For those in browsers but not in the HTML spec:
  - `onabort` is removed in the PR. Need to check.
  - 28 are added in the PR. They mainly consist of animation/transition events and pointer events, which presumably aren't in the HTML spec for good reasons. Will need to look closer to see if they are real HTML attributes:
    - `onanimationcancel`
    - `onanimationend`
    - `onanimationiteration`
    - `onanimationstart`
    - `onbeforexrselect`
    - `oncontentvisibilityautostatechange`
    - `ongotpointercapture`
    - `onlostpointercapture`
    - `onpointercancel`
    - `onpointerdown`
    - `onpointerenter`
    - `onpointerleave`
    - `onpointermove`
    - `onpointerout`
    - `onpointerover`
    - `onpointerrawupdate`
    - `onpointerup`
    - `onscrollsnapchange`
    - `onscrollsnapchanging`
    - `onselectstart`
    - `ontransitioncancel`
    - `ontransitionend`
    - `ontransitionrun`
    - `ontransitionstart`
    - `onwebkitmouseforcechanged`
    - `onwebkitmouseforcedown`
    - `onwebkitmouseforceup`
    - `onwebkitmouseforcewillbegin`
  - 11 are not in the PR before or after. They look non-standard but would be worth investigating:
    - `onbeforecopy`
    - `onbeforecut`
    - `onbeforepaste`
    - `ondragexit`
    - `onmozfullscreenchange`
    - `onmozfullscreenerror`
    - `onselectionchange`
    - `onwebkitanimationend`
    - `onwebkitanimationiteration`
    - `onwebkitanimationstart`
    - `onwebkittransitionend`

(Btw I didn't check `Element.prototype` which also have some event listeners)

Josh-Cena:
Found a great way to test it:

```html
<!doctype html>
<html>
  <head>
    <title></title>
    <meta charset="utf-8" />
    <style>
      body {
        width: 1400px;
        display: flex;
        flex-wrap: wrap;
      }
      div {
        width: 150px;
        border: 1px solid black;
        text-align: center;
        margin: 0.2em;
        padding: 0.2em;
      }
      [data-fired] {
        background: lightgreen;
      }
    </style>
  </head>
  <body>
    <script>
      function testEvent(type) {
        const div = document.createElement("div");
        div.id = `event-${type}`;
        div.textContent = type;
        document.body.appendChild(div);
        div.setAttribute(`on${type}`, "this.setAttribute('data-fired', 'yes')");
        div.dispatchEvent(new Event(type));
      }
      for (const type of [
        ...Object.getOwnPropertyNames(HTMLElement.prototype),
        ...Object.getOwnPropertyNames(Element.prototype),
      ]
        .filter((p) => p.startsWith("on"))
        .sort()) {
        testEvent(type.slice(2));
      }
    </script>
  </body>
</html>
```

Chrome:

image

Safari:

image

Firefox:

image

Josh-Cena:
Going back to my previous list:

Added in this PR but not in browsers (as IDL):

- `onafterscriptexecute`, `onbeforescriptexecute` - do exist in FF 143 as an attribute, but already removed in FF 144 per https://github.com/mdn/content/issues/39304.
- `oncompositionend`, `oncompositionstart`, `oncompositionupdate` - aren't in any browser as attributes
- `onfocusin`, `onfocusout` - all supporting browsers have the attributes. More content work in https://github.com/mdn/content/issues/32884 to add the IDL.
- `onfullscreenchange`, `onfullscreenerror` - only in FF as attributes. Chrome supports the event but not the attributes.
- `ongesturechange`, `ongestureend`, `ongesturestart` - all supporting browsers (Safari) have the attributes.
- `ontouchcancel`, `ontouchend`, `ontouchmove`, `ontouchstart` - attributes supported in Chrome and Safari. Weirdly BCD says Safari doesn't support these events, so there's something fishy.

So `onafterscriptexecute` and `onbeforescriptexecute` should be removed, just to be forward looking. `oncompositionend`, `oncompositionstart`, `oncompositionupdate` also need to be removed unless I'm missing something. The other changes are good.

Removed in this PR but in browsers: `onabort` is in all browsers. Please restore it.

Added in this PR and in browsers:

- `onanimationcancel` - all supporting browsers (FF, Safari) have the attributes.
- `onanimationend`, `onanimationiteration`, `onanimationstart` - all supporting browsers have the attributes.
- `onbeforexrselect` - not supported in any browser as attribute.
- `oncontentvisibilityautostatechange`, `ongotpointercapture`, `onlostpointercapture`, `onpointercancel`, `onpointerdown`, `onpointerenter`, `onpointerleave`, `onpointermove`, `onpointerout`, `onpointerover`, `onpointerup`, `onpointerrawupdate` - all supporting browsers have the attributes.
- `onscrollsnapchange`, `onscrollsnapchanging` - all supporting browsers (Chrome) have the attributes.
- `onselectstart` - all supporting browsers have the attributes.
- `ontransitioncancel`, `ontransitionend`, `ontransitionrun`, `ontransitionstart` - not in Chrome as attributes but in the other two. All three support the event.
- `onwebkitmouseforcechanged`, `onwebkitmouseforcedown`, `onwebkitmouseforceup`, `onwebkitmouseforcewillbegin` - all supporting browsers (Safari) have the attributes.

So `onbeforexrselect` should be removed.

Never documented but in browsers:

- `onbeforecopy`, `onbeforecut`, `onbeforepaste` - in Chrome and Safari as attributes
- `ondragexit`, `onmozfullscreenchange`, `onmozfullscreenerror` - in FF as attribute
- `onselectionchange` - in FF and Safari as attributes
- `onwebkitanimationend`, `onwebkitanimationiteration`, `onwebkitanimationstart`, `onwebkittransitionend` - in none of the browsers as attributes

So `onselectionchange` needs to be added, while `onbeforecopy`, `onbeforecut`, `onbeforepaste` could be documented if you could find a spec.

That seems about it. Thanks!

files/en-us/web/html/reference/global_attributes/index.md
```diff
@@ -14,7 +14,7 @@ In addition to the basic HTML global attributes, the following global attributes
...
+- The [event handler attributes](/en-US/docs/Web/HTML/Reference/Global_attributes#list_of_global_event_handler_attributes) listed below.
```
Josh-Cena:
```suggestion
- The [event handler attributes](#list_of_global_event_handler_attributes) listed below.
```

files/en-us/web/html/reference/global_attributes/index.md
```diff
@@ -120,6 +120,293 @@ In addition to the basic HTML global attributes, the following global attributes
...
+- [`onwebkitmouseforcechanged`](/en-US/docs/Web/API/Element/webkitmouseforcechanged_event) {{non-standard_inline}}
+  - : Fired each time the amount of pressure changes on the trackpad touch screen.
+- [`onwebkitmouseforcedown`](/en-US/docs/Web/API/Element/webkitmouseforcedown_event) {{non-standard_inline}}
+  - : Fired after the mousedown event as soon as sufficient pressure has been applied to qualify as a "force click".
+- [`onwebkitmouseforceup`](/en-US/docs/Web/API/Element/webkitmouseforceup_event) {{non-standard_inline}}
+  - : Fired after the [`webkitmouseforcedown`](/en-US/docs/Web/API/Element/webkitmouseforcedown_event) event as soon as the pressure has been reduced sufficiently to end the "force click".
+- [`onwebkitmouseforcewillbegin`](/en-US/docs/Web/API/Element/webkitmouseforcewillbegin_event) {{non-standard_inline}}
+  - : Fired before the [`mousedown`](/en-US/docs/Web/API/Element/mousedown_event) event.
+- [`onwheel`](/en-US/docs/Web/API/Element/wheel_event)
+  - : Fired when the user rotates a wheel button on a pointing device (typically a mouse).
```
Josh-Cena: IMO we should remove these `- :` so it's just a simple UL. This is a bit hard to look at and frankly not very useful because we really want to discourage the usage of these attributes. Totally fine if you think otherwise though.

sczesny: I've copied the format from the List of global attributes above. As far as I've seen, it's the usual format for lists.

Josh-Cena: Yeah, both are common styles. I'm saying this one doesn't really deserve this much explanation and is crowding out the actually valuable docs a bit. We want to stash away these attributes as if they don't exist, so a plain UL is about the attention that should be devoted.

sczesny: Updated the format 👍

Josh-Cena (approved):
This looks really good now. Thanks for sending this! I just removed all the extra explanation (I probably didn't make this instruction clear) so it's just a simple UL, and added some more intro text.

PR #41271, opened 2025-09-24
DNM: test writer toolbar

LeoMcA:
Do not merge, apologies for the noise: testing if the writer toolbar now appears on PRs.


LeoMcA:
Lovely stuff, toolbar appears: closing.

PR #41275, opened 2025-09-24
Fix typo in http/reference/status/502

Jamesllllllllll:
Hey there!

I assume this is meant to be "catch-all" - I searched and could not find any usage of the phrase "catch-call" in case I was unfamiliar with the term.

Cheers

### Description

Fix a typo.

### Motivation

Avoid confusion.


bsmth (approved):
Nice thanks! This is probably my fault 😄

PR #41276, opened 2025-09-24
fix incorrect aria-label usage for required labels

### LINKED ISSUE #41245: misuse of aria-label in "How to structure a web form" doc
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/How_to_structure_a_web_form

Multiple labels (and others)

The doc indicates that people should use `<span aria-label=required>*</span>`, but this is an invalid practice that doesn't always get exposed to screen readers (though the mdn doc incorrectly claims otherwise).

not all screen readers will expose the aria-label on a span element when reading through a web page.   

For instance, the example of `<p>Required fields are followed by <span aria-label=required>*</span>.</p>` will be read by most (if not all) screen readers as "required fields are followed by star" - "asterisk" may be read instead of "star" depending on the platform.

Where it _can_ expose the aria-label is when used within a `label` element, as the browser will try to collect as much information as it can to provide the accName to the `input`.

However, even this caveat is not a good reason to be using it.  As per one of the examples in the doc:
```
<label for="username">
    <span>Name:</span>
    <input id="username" type="text" name="username" required />
    <span aria-label="required">*</span>
  </label>
```

while the `aria-label=required` will be used in the calculation of the `input`'s accName, along with "Name" - its use is unnecessary as the `input` also has the `required` attribute - resulting in "required" being announced as part of the accessible name, AND as the state of the input.

A better recommendation would be to _not_ use `aria-label` in a way that the [ARIA](https://w3c.github.io/aria/#namefromprohibited) and [ARIA in HTML](https://w3c.github.io/html-aria/#el-span) specs indicate are prohibited/invalid.

Rather, a better practice would be to introduce a form with a message like:
"Please complete all required(*) fields."

Again, as the `input` elements already have the `required` attribute on them, this will expose they are required to AT.  The asterisk can be left alone to ensure that it will either be announced the same (e.g., "star") or _ignored_ for those who have user settings that purposefully do not announce all special characters.

The ARIA / ARIA in HTML specs that have already been referenced.

_No response_

### MDN metadata

<details>

* Folder: `en-us/learn_web_development/extensions/forms/how_to_structure_a_web_form`
* MDN URL: https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/How_to_structure_a_web_form
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/learn_web_development/extensions/forms/how_to_structure_a_web_form/index.md
* Last commit: https://github.com/mdn/content/commit/1ff044ac87e406eb23ae7181dd171bad87421b79
* Document last modified: 2025-09-10T16:21:26.000Z

</details>

chrisdavidmills:
### Description

https://github.com/mdn/content/issues/41245 explains why the `<span aria-label=required>*</span>` pattern in forms is bad. This PR aims to remove it from the places I could find it and replace it with a simpler, more effective pattern.

Fixes https://github.com/mdn/content/issues/41245

### DIFF
--- a/files/en-us/learn_web_development/extensions/forms/form_validation/index.md
+++ b/files/en-us/learn_web_development/extensions/forms/form_validation/index.md
@@ -158,7 +158,7 @@ Add a `required` attribute to your input, as shown below.
 
 ```html live-sample___the-required-attribute
 <form>
-  <label for="choose">Would you prefer a banana or cherry? (*)</label>
+  <label for="choose">Would you prefer a banana or cherry? *</label>
   <input id="choose" name="i-like" required />
   <button>Submit</button>
 </form>
@@ -220,7 +220,7 @@ Update your HTML to add a [`pattern`](/en-US/docs/Web/HTML/Reference/Attributes/
 
 ```html live-sample___validate-regular-expression
 <form>
-  <label for="choose">Would you prefer a banana or a cherry? (*)</label>
+  <label for="choose">Would you prefer a banana or a cherry? *</label>
   <input id="choose" name="i-like" required pattern="[Bb]anana|[Cc]herry" />
   <button>Submit</button>
 </form>
@@ -279,7 +279,7 @@ Now delete the contents of the `<body>` element, and replace it with the followi
 ```html live-sample___constraining-values
 <form>
   <div>
-    <label for="choose">Would you prefer a banana or a cherry? (*)</label>
+    <label for="choose">Would you prefer a banana or a cherry? *</label>
     <input
       type="text"
       id="choose"
@@ -335,7 +335,7 @@ First, some HTML:
 <form>
   <p>Please complete all required (*) fields.</p>
   <fieldset>
-    <legend>Do you have a driver's license? (*)</legend>
+    <legend>Do you have a driver's license? *</legend>
     <input type="radio" required name="driver" id="r1" value="yes" />
     <label for="r1">Yes</label>
     <input type="radio" required name="driver" id="r2" value="no" />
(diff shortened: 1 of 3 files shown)


files/en-us/learn_web_development/extensions/forms/form_validation/index.md
```diff
@@ -158,13 +158,14 @@ Add a `required` attribute to your input, as shown below.
 
 ```html live-sample___the-required-attribute
 <form>
-  <label for="choose">Would you prefer a banana or cherry? (required)</label>
+  <label for="choose">Would you prefer a banana or cherry? (*)</label>
```
scottaohara: in my issue, i wrote (*) because it was paired with the word "required" in the description of the form.

the parenthesis aren't typically needed or used outside of that context.

files/en-us/learn_web_development/extensions/forms/form_validation/index.md
```diff
@@ -332,27 +333,20 @@ First, some HTML:
 
 ```html
 <form>
+  <p>Please complete all required (*) fields.</p>
```
scottaohara, line 336: so per my previous comment, using the parenthesis here makes sense because it's used in the sentence to indicate that required fields will be visually marked with the *.

but in the legend, it'd be more common to just have the * at the end of the text, without the () surrounding it.

scottaohara (approved):
thanks for taking care of all these while addressing my original issue.

my comment about the (*) is also meant more as a clarification to what i was expecting, rather than a "need to fix".  so update it if you like, or keep it as you have it.  nothing wrong with it, just wasn't what I was expecting per how I typically see required fields visually identified.

files/en-us/learn_web_development/extensions/forms/form_validation/index.md
chrisdavidmills: @scottaohara This makes sense. I've changed it as suggested. Thx for the review!

PR #41283, opened 2025-09-24
Element.prototype.{onfocusin,onfocusout} now exist

### LINKED ISSUE #32884: Add onfocusin and onfocusout to event handler content attributes

Josh-Cena:
Fix https://github.com/mdn/content/issues/32884

### DIFF
--- a/files/en-us/web/api/element/focusin_event/index.md
+++ b/files/en-us/web/api/element/focusin_event/index.md
@@ -16,14 +16,13 @@ The `focusin` event is not cancelable.
 
 ## Syntax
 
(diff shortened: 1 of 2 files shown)


wbamberg (approved):
👍

PR #41289, opened 2025-09-25
Fix link to 2D rotate function in documentation

jasonren0403:
origin link was linking to rotate() function

### Description

Fix a possible typo, link text saying `rotate3D` but link is going to `rotate`.


Josh-Cena (approved):
Good catch, thanks

PR #41301, opened 2025-09-26
fix(css): add warning messages to the livesamples

### LINKED ISSUE #41113: corner-shape examples should show notification when the feature is not supported
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/corner-shape

corner-shape

The demos require the feature to be supported in the user's browser

Either a rendered image of the demo, or a banner saying that the feature is not supported thus might not be rendered correctly

_No response_

_No response_

OnkarRuikar:
- fix https://github.com/mdn/content/issues/41113

The PR does the following:
- Adds warnings to the live sample outputs if the `corner-shape` is not supported by the browsers. We have used the `@supports not (corner-shape: ...)` technique on other pages.
- Simplifies the last example. Many non-technical users may not know how to hover or focus on the div on touch devices. So to make the animation starting point prominent, we can simply extend the initial state for a while.

You'll have to test this on both Firefox and Chrome browsers.

### DIFF
--- a/files/en-us/web/css/corner-shape/index.md
+++ b/files/en-us/web/css/corner-shape/index.md
@@ -175,7 +175,7 @@ div {
 
 @supports not (corner-shape: scoop) {
   body {
-    width: 100%;
+    all: unset !important;
   }
 
   body::before {
@@ -185,6 +185,7 @@ div {
     display: block;
     width: 100%;
     text-align: center;
+    padding: 1rem 0;
   }
 
   body > * {
@@ -259,8 +260,7 @@ div {
 
 @supports not (corner-shape: scoop notch) {
   body {
-    width: 100%;
-    background: initial !important;
+    all: unset !important;
   }
 
   body::before {
@@ -270,6 +270,7 @@ div {
     display: block;
     width: 100%;
     text-align: center;
+    padding: 1rem 0;
   }
 
   body > * {
@@ -397,7 +398,7 @@ section {
 
 @supports not (corner-shape: scoop) {
   body {
-    width: 100%;
+    all: unset !important;
   }
 
   body::before {
@@ -407,6 +408,7 @@ section {
     display: block;
     width: 100%;
     text-align: center;
+    padding: 1rem 0;
   }
 
   body > * {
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/corner-shape/index.md
```diff
@@ -172,6 +172,25 @@ div {
     rgb(255 255 255 / 0.5)
   );
 }
+
+@supports not (corner-shape: scoop) {
+  body {
```
chrisdavidmills, line 177: This functionality works OK, but it looks slightly jarring as the not supported banner appears at a different vertical position on the embedded viewport in each demo. To fix this, maybe set the HTML and body height to 100% and use flexbox to vertically center the body in each case, inside the @supports block?

Also, it might be nice to give the not supported banner some vertical padding in each case so that it looks a bit nicer.

files/en-us/web/css/corner-shape/index.md
```diff
@@ -540,14 +634,18 @@ div {
     corner-shape: square;
   }
 
+  /* To make the starting point apparent, let us keep the shape the same for a small duration. */
```
chrisdavidmills:
```suggestion
  /* To make the starting point apparent, let us keep
  the shape the same for a small duration. */
```

Or similar, so that the comment doesn't break untidily across the two lines.

chrisdavidmills (changes requested):
Thanks @OnkarRuikar, I like how you've handled these issues. I'll use this techniques on future property pages.

Just a couple of suggestions to check out.

files/en-us/web/css/corner-shape/index.md
OnkarRuikar: doen

files/en-us/web/css/corner-shape/index.md
OnkarRuikar: My bad. Now all the warning banners are uniform with 1rem top and bottom padding:
>

chrisdavidmills (approved):
Awesome, this looks great, thanks @OnkarRuikar!

PR #41317, opened 2025-09-27
Fixed incorrect stylesheet selected in setproperty() example

djsmall:
### Description

Fixed the following "TypeError: can't access property "style", boxParaRule is undefined" by updating stylesheet used: const stylesheet = document.styleSheets[2];

### Motivation

The example in the setProperty() article was not functioning properly when I tested it on several different browsers and devices, I think this update will help other users understand setProperty() better with a functioning example.


Josh-Cena:
Thanks for the fix! This will be fixed via https://github.com/mdn/yari/issues/12809, and in the meantime we'd rather not do anything about it since there are many other similar problems, and without a stable contract of which stylesheet we should manipulate, it's prone to change again.

PR #41319, opened 2025-09-28
Fix typo

SalamehAn:
This was a typo in the following article:
https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Tools_and_setup/set_up_a_local_testing_server

There are three ways to correcting this:
1. "The Node.js http-server module is the easiest way to host HTML files in any directory."
2. "The Node.js http-server module is an easy way to host HTML files in any directory."
3. "The Node.js http-server module is one of the easiest ways to host HTML files in any directory."

I've opted for the second change, to adopt the least authoritative tone on the topic. Three works as well, but these are just my opinions.

### Description
Fixed typo in the following article: https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Tools_and_setup/set_up_a_local_testing_server

It was in the `Using Node.js` subsection, under the `Running a simple local HTTP server` section.

### Additional details
Article edited: https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Tools_and_setup/set_up_a_local_testing_server

### DIFF
--- a/files/en-us/learn_web_development/howto/tools_and_setup/set_up_a_local_testing_server/index.md
+++ b/files/en-us/learn_web_development/howto/tools_and_setup/set_up_a_local_testing_server/index.md
@@ -58,7 +58,7 @@ For VS Code, try out the following free extensions:
 
 ### Using Node.js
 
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
Nice, thanks

PR #41327, opened 2025-09-29
Fix "Using relative colors" syntax example

caugner:
### Description

Fixes the "General syntax" example on the "Using relative colors" page.

### Motivation

The example was not consistent with the following text.

### Additional details

Reported by @Mossop on Slack.


chrisdavidmills (approved):
LGTM; thanks @caugner!

PR #41331, opened 2025-09-29
Bug-1976509 cookie.set method rejects invalid cookies

rebloor:
### Description

Addresses the dev-docs-needed requirements of [Bug 1976509](https://bugzilla.mozilla.org/show_bug.cgi?id=1976509) Reject invalid cookies in the webExt cookie API with:
- a release note describing the change
- an update to the cookies.set method documentation

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/api/cookies/set/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/api/cookies/set/index.md
@@ -62,7 +62,7 @@ If there is more than one cookie with the same name for a URL, the cookie with t
 If the requested cookie is invalid or the call otherwise fails, the promise is rejected with an error message.
 
 > [!NOTE]
-> Before Firefox 145 , invalid cookies were created.
+> Before Firefox 145, invalid cookies were created.
 
 ## Examples
 
--- a/files/en-us/mozilla/add-ons/webextensions/api/cookies/set/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/api/cookies/set/index.md
@@ -6,11 +6,9 @@ browser-compat: webextensions.api.cookies.set
 sidebar: addonsidebar
 ---
 
-The **`set()`** method of the {{WebExtAPIRef("cookies")}} API sets a cookie containing the specified cookie data. This method is equivalent to issuing an HTTP `Set-Cookie` header during a request to a given URL.
+Sets a cookie. This method is equivalent to issuing an HTTP `Set-Cookie` header during a request to a URL.
 
 To use this method, an extension must have the `"cookies"` permission and relevant host permissions. See [`cookie` permissions](/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies#permissions) for more details.
-
-This is an asynchronous function that returns a [`Promise`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).
 
 ## Syntax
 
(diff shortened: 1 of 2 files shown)


files/en-us/mozilla/add-ons/webextensions/api/cookies/set/index.md
```diff
@@ -6,12 +6,10 @@ browser-compat: webextensions.api.cookies.set
 sidebar: addonsidebar
 ---
 
-The **`set()`** method of the {{WebExtAPIRef("cookies")}} API sets a cookie containing the specified cookie data. This method is equivalent to issuing an HTTP `Set-Cookie` header during a request to a given URL.
+Sets a cookie. This method is equivalent to issuing an HTTP `Set-Cookie` header during a request to a URL.
```
bsmth, line 9: The first sentence follows the convention where we have:

```
The **`{method_name}`** method of the {API_Name} API {usage}…
```

So I think we have to stick with that here, something like:

```suggestion
The **`set()`** method of the {{WebExtAPIRef("cookies")}} API sets a cookie.
This method is equivalent to including an HTTP {{HTTPHeader("Set-Cookie")}} header in a request to a URL.
```

rebloor: @bsmth this convention is very much the exception for the web extension API. It's only used on six methods in the bookmarks and cookies. Everything else uses the more straightforward approach.

dotproto: I suggest we use the current WebExt doc convention for this PR and schedule a discussion about this in our editorial sync meeting? @bsmth, would you be amenable to that?

EDIT: I've added it to the agenda for our next meeting.

dotproto (approved):
Approving pending anticipated resolution of the discussion in https://github.com/mdn/content/pull/41331/files#r2390199923

files/en-us/mozilla/add-ons/webextensions/api/cookies/set/index.md
bsmth: Thanks, both!

> this convention is very much the exception for the web extension API

I'm not familiar with the WebExt conventions, but I don't see other places we use shorter sentences, or have I misunderstood? Just highlighting that the rationale for this in other docs areas is that the first sentence(s) are used by the platform for creating meta description and other summary-type elements, e.g. this comes from the first sentence here https://github.com/mdn/content/blob/09cc618e47efb40e15a3b6ff60557a455cf76ef8/files/en-us/mozilla/add-ons/webextensions/api/cookies/remove/index.md?plain=1#L10

```html
<meta name="description" content="The remove() method of the cookies API deletes a cookie, given its name and URL.">
```

>I've added it to the agenda for our next meeting.

Thanks a lot!

rebloor: @bsmth As far as I can tell, and I haven't checked all of the examples, our guidance is much less clear. [For example](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Write_an_api_reference#structure_of_an_interface_page), "the first paragraph of the interface page should provide a short (sic) concise description of the interface's overarching purpose." [API method page template](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Write_an_api_reference) is more specific but not prescriptive, "Begin the content on the page with an introductory paragraph — start by naming the method, saying what interface it is part of, and saying what it does. This should ideally be one or two short sentences." (We don't have similar templates for the web extension API pages.) I can't see any mention that the sentence is used for any particular purpose.

The more concise form seems perfectly adequate for search results, for example:

image

Whereas the alternative seems repetitive

image

(One thing we probably do need is to identify Web extension API in the page title, for example, "scripting.executeScript() - WebExtension APIs – Mozilla - MDN

I also noticed that Chrome doesn't use meta-descriptions in their developer documentation.

PR #41339, opened 2025-09-30
Remove example of using sibling-index with content

### LINKED ISSUE #41290: Wrong code sample
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/sibling-index

https://developer.mozilla.org/en-US/docs/Web/CSS/sibling-index#css_2

Incorrect sample 
```css
div {
  --list-index: sibling-index();
  display: flex;
  gap: 1ch;
}

div::before {
  content: var(--list-index);
}
```
`sibling-index` returns number value, and `content ` property avail only string, so this sample not working. element `::before` not created.

I expect fix this, for example use counters

https://developer.mozilla.org/en-US/play?id=zjs%2BiXkQOJQx8fdwOlDSH56IjDGx8QUbl%2FafbhGwBrluzJLPBBDNq8hw174KjEQIfrN2CPe5FEyaFG15

```css
div {
  --list-index: sibling-index();
  display: flex;
  gap: 1ch;
  
}

div::before {
  counter-set: c var(--list-index);
  content: counter(c);
}

```

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/css/sibling-index`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/sibling-index
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/sibling-index/index.md
* Last commit: https://github.com/mdn/content/commit/9ccdb2ddd0d46e6d203044c3fffb50a784528c8c
* Document last modified: 2025-08-22T14:19:45.000Z

</details>

sagarp2901:
### Description

This PR fixes the “Ordered list” example on the [sibling-index()](https://developer.mozilla.org/en-US/docs/Web/CSS/sibling-index?utm_source=chatgpt.com) documentation page.

### Motivation

The bug: The old example set content: var(--list-index); where --list-index was just an integer. The CSS content property does not accept raw numbers, so the pseudo-element would render nothing.

The fix: By introducing a CSS counter (counter-set + counter()), the numeric value from sibling-index() is converted into a valid, displayable string.

Fixes https://github.com/mdn/content/issues/41290


Josh-Cena (commented):
While this fixes the issue, I think it's an antipattern since you should just use `counter()` in this case. @ramiy WDYT? We should remove this example and/or replace it with something more reasonable.

Josh-Cena (approved):
I opted to just remove the example, because for generated content one should use `counter()`.

files/en-us/web/css/sibling-index/index.md
```diff
@@ -10,7 +10,13 @@ sidebar: cssref
...
+The **`sibling-index()`** [CSS](/en-US/docs/Web/CSS) [function](/en-US/docs/Web/CSS/CSS_values_and_units/CSS_value_functions) returns an integer representing the position of the current element in the DOM tree relative to all its sibling elements. The returned value is the index number of the contextual child's position among all the sibling elements within a parent element, with the first child returning `1` and the last child, returning the `length` of {{domxref("Element.children")}}.
+
+> [!NOTE]
+> Like the {{CSSxRef(":nth-child()")}} pseudo-class, `sibling-index()` starts from 1, not 0.
+
+> [!NOTE]
+> The {{CSSxRef("counter()")}} function provides a similar result but it returns a `<string>` (which is more suitable for [generated content](/en-US/docs/Web/CSS/CSS_generated_content), while `sibling-count()` returns an `<integer>` (which can be used for calculations).
```
Josh-Cena:
```suggestion
> The {{CSSxRef("counter()")}} function provides a similar result but it returns a `<string>` (which is more suitable for [generated content](/en-US/docs/Web/CSS/CSS_generated_content), while `sibling-index()` returns an `<integer>` (which can be used for calculations).
```

sagarp2901:
Thanks for your review @Josh-Cena!

PR #41341, opened 2025-09-30
HTMLScriptElement.textContent/innerText - docs for TrustedTypes

hamishwillee:
This updates the TrustedType documentation for `HTMLScriptElement.textContent` so that it includes boilerplate and security considerations.

These was inherited from node but is now in IDL - see https://w3c.github.io/trusted-types/dist/spec/#enforcement-in-scripts

Note that this essentially copies [`HTMLScriptElement.text`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement/text) which was updated in #41128. The only minor difference is the note about the fact it is also defined on Node and so can be used with other elements.

Project tracking in #37518

### DIFF
--- a/files/en-us/web/api/htmlelement/innertext/index.md
+++ b/files/en-us/web/api/htmlelement/innertext/index.md
@@ -8,6 +8,13 @@ browser-compat: api.HTMLElement.innerText
 
 {{APIRef("HTML DOM")}}
 
+> [!WARNING]
+> This property represents the text content of a script element, which may be executable depending on the script type.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
+>
+> You can mitigate this risk by always assigning {{domxref("TrustedScript")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](#security_considerations) for more information.
+
 The **`innerText`** property of the {{domxref("HTMLElement")}} interface represents the rendered text content of a node and its descendants.
 
 As a getter, it approximates the text the user would get if they highlighted the contents of the element with the cursor and then copied it to the clipboard.
@@ -24,8 +31,7 @@ A string representing the rendered text content of an element.
 If the element itself is not [being rendered](https://html.spec.whatwg.org/multipage/rendering.html#being-rendered) (for example, is detached from the document or is hidden from view), the returned value is the same as the {{domxref("Node.textContent")}} property.
 
 > [!WARNING]
-> Setting `innerText` on a node removes _all_ of the node's children
-> and replaces them with a single text node with the given string value.
+> Setting `innerText` on a node removes _all_ of the node's children and replaces them with a single text node with the given string value.
 
 ## Examples
 
@@ -53,9 +59,9 @@ Note how `innerText` is aware of things like {{htmlElement("br")}} elements, and
   <span style="display:none">HIDDEN TEXT</span>
 </p>
 <h3>Result of textContent:</h3>
-<textarea id="textContentOutput" rows="6" cols="30" readonly>…</textarea>
+<textarea id="textContentOutput" rows="18" cols="40" readonly>…</textarea>
 <h3>Result of innerText:</h3>
-<textarea id="innerTextOutput" rows="6" cols="30" readonly>…</textarea>
+<textarea id="innerTextOutput" rows="6" cols="40" readonly>…</textarea>
 ```
 
 ### JavaScript
@@ -71,7 +77,7 @@ innerTextOutput.value = source.innerText;
 
 ### Result
 
-{{EmbedLiveSample("Examples", 700, 450)}}
+{{EmbedLiveSample("Examples", 700, 650)}}
 
 ## Specifications
 
@@ -83,5 +89,6 @@ innerTextOutput.value = source.innerText;
 
 ## See also
 
+- {{domxref("HTMLScriptElement.innerText")}}
 - {{domxref("HTMLElement.outerText")}}
 - {{domxref("Element.innerHTML")}}
--- a/files/en-us/web/api/htmlelement/innertext/index.md
+++ b/files/en-us/web/api/htmlelement/innertext/index.md
@@ -8,13 +8,6 @@ browser-compat: api.HTMLElement.innerText
 
 {{APIRef("HTML DOM")}}
 
-> [!WARNING]
-> This property represents the text content of a script element, which may be executable depending on the script type.
-> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
->
-> You can mitigate this risk by always assigning {{domxref("TrustedScript")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
-> See [Security considerations](#security_considerations) for more information.
-
 The **`innerText`** property of the {{domxref("HTMLElement")}} interface represents the rendered text content of a node and its descendants.
 
 As a getter, it approximates the text the user would get if they highlighted the contents of the element with the cursor and then copied it to the clipboard.
--- a/files/en-us/web/api/htmlscriptelement/index.md
+++ b/files/en-us/web/api/htmlscriptelement/index.md
@@ -51,7 +51,6 @@ _Inherits properties from its parent, {{domxref("HTMLElement")}}._
 - {{domxref("HTMLScriptElement.text")}}
   - : A property that represents the inline text content of the {{HTMLElement("script")}} element.
     The property accepts either a {{domxref("TrustedScript")}} object or a string.
-    It returns a string that joins the contents of all {{domxref("Text")}} nodes inside the {{HTMLElement("script")}} element (ignoring other nodes like comments) in tree order.
     It acts the same way as the [textContent](#textcontent) property.
 - {{domxref("HTMLScriptElement.textContent")}}
   - : A property that represents the inline text content of the {{HTMLElement("script")}} element.
--- a/files/en-us/web/api/htmlscriptelement/index.md
+++ b/files/en-us/web/api/htmlscriptelement/index.md
@@ -36,6 +36,9 @@ _Inherits properties from its parent, {{domxref("HTMLElement")}}._
   - : A string; an obsolete way of registering event handlers on elements in an HTML document.
 - {{domxref("HTMLScriptElement.fetchPriority")}}
   - : An optional string representing a hint given to the browser on how it should prioritize fetching of an external script relative to other external scripts. If this value is provided, it must be one of the possible permitted values: `high` to fetch at a high priority, `low` to fetch at a low priority, or `auto` to indicate no preference (which is the default). It reflects the `fetchpriority` attribute of the {{HTMLElement("script")}} element.
+- {{domxref("HTMLScriptElement.innerText")}}
+  - : A property that represents the inline text content of the {{HTMLElement("script")}} element as though it were rendered text.
+    The property accepts either a {{domxref("TrustedScript")}} object or a string.
 - {{domxref("HTMLScriptElement.integrity")}}
   - : A string that contains inline metadata that a browser can use to verify that a fetched resource has been delivered without unexpected manipulation. It reflects the `integrity` attribute of the {{HTMLElement("script")}} element.
 - {{domxref("HTMLScriptElement.noModule")}}
--- a/files/en-us/web/api/htmlscriptelement/index.md
+++ b/files/en-us/web/api/htmlscriptelement/index.md
@@ -11,6 +11,9 @@ HTML {{HTMLElement("script")}} elements expose the **`HTMLScriptElement`** inter
 
 JavaScript files should be served with the `text/javascript` [MIME type](/en-US/docs/Web/HTTP/Guides/MIME_types), but browsers are lenient and block them only if the script is served with an image type (`image/*`), video type (`video/*`), audio type (`audio/*`), or `text/csv`. If the script is blocked, its element receives an {{domxref("HTMLElement/error_event", "error")}} event; otherwise, it receives a {{domxref("Window/load_event", "load")}} event.
 
+> [!NOTE]
+> When inserted using the {{domxref("Document.write()")}} method, {{HTMLElement("script")}} elements execute (typically synchronously), but when inserted using {{domxref("Element.innerHTML")}} or {{domxref("Element.outerHTML")}}, they do not execute at all.
+
 {{InheritanceDiagram}}
 
 ## Instance properties
@@ -40,15 +43,20 @@ _Inherits properties from its parent, {{domxref("HTMLElement")}}._
 - {{domxref("HTMLScriptElement.referrerPolicy")}}
   - : A string that reflects the [`referrerPolicy`](/en-US/docs/Web/HTML/Reference/Elements/script#referrerpolicy) HTML attribute indicating which referrer to use when fetching the script, and fetches done by that script.
 - {{domxref("HTMLScriptElement.src")}}
-  - : A string representing the URL of an external script; this can be used as an alternative to embedding a script directly within a document. It reflects the `src` attribute of the {{HTMLElement("script")}} element.
+  - : A string representing the URL of an external script; this can be used as an alternative to embedding a script directly within a document.
+    It reflects the `src` attribute of the {{HTMLElement("script")}} element.
 - {{domxref("HTMLScriptElement.text")}}
-  - : A string that joins and returns the contents of all {{domxref("Text")}} nodes inside the {{HTMLElement("script")}} element (ignoring other nodes like comments) in tree order. On setting, it acts the same way as the {{domxref("Node.textContent")}} property.
-
-    > [!NOTE]
-    > When inserted using the {{domxref("Document.write()")}} method, {{HTMLElement("script")}} elements execute (typically synchronously), but when inserted using {{domxref("Element.innerHTML")}} or {{domxref("Element.outerHTML")}}, they do not execute at all.
-
+  - : A property that represents the inline text content of the {{HTMLElement("script")}} element.
+    The property accepts either a {{domxref("TrustedScript")}} object or a string.
+    It returns a string that joins the contents of all {{domxref("Text")}} nodes inside the {{HTMLElement("script")}} element (ignoring other nodes like comments) in tree order.
+    It acts the same way as the [textContent](#textcontent) property.
+- {{domxref("HTMLScriptElement.textContent")}}
+  - : A property that represents the inline text content of the {{HTMLElement("script")}} element.
+    The property is redefined from {{domxref("Node/textContent","Node")}} to support {{domxref("TrustedScript")}} as an input.
+    On this element it behaves exactly like the [text](#text) property.
 - {{domxref("HTMLScriptElement.type")}}
-  - : A string representing the type of the script. It reflects the `type` attribute of the {{HTMLElement("script")}} element.
+  - : A string representing the type of the script.
+    It reflects the `type` attribute of the {{HTMLElement("script")}} element.
 
 ## Static methods
 
--- a/files/en-us/web/api/htmlscriptelement/innertext/index.md
+++ b/files/en-us/web/api/htmlscriptelement/innertext/index.md
@@ -1,9 +1,9 @@
 ---
-title: "HTMLScriptElement: text property"
-short-title: text
-slug: Web/API/HTMLScriptElement/text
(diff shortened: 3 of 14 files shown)


files/en-us/web/api/htmlscriptelement/textcontent/index.md
```diff
@@ -0,0 +1,82 @@
...
+>
+> You can mitigate this risk by always assigning {{domxref("TrustedScript")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](#security_considerations) for more information.
+
+The **`textContent`** property of the {{domxref("HTMLScriptElement")}} interface represents the inline text content of the script element.
+It acts the same way as the {{domxref("HTMLScriptElement.text","text")}} property.
+
+The `textContent` property is also defined on {{domxref("Node.textContent","Node")}} and can hence be used with other elements.
+When used with other elements it does not expect or enforce the assignment of a {{domxref("TrustedScript")}}.
+It is also different in that a {{domxref("HTMLScriptElement")}} is not expected to contain other elements, and will parse input as plain text rather than as HTML markup.
```
hamishwillee: This is the main difference compared to the `text` docs from which this was largely copied.

files/en-us/web/api/htmlscriptelement/textcontent/index.md
```diff
@@ -0,0 +1,82 @@
...
+You can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) or reject the text before it is injected.
+
+The behavior of the transformation function will depend on the specific use case that requires a user provided script.
+If possible you should lock the allowed scripts to exactly the code that you trust to run.
+If that is not possible, you might allow or block the use of certain functions within the provided string.
+
+## Examples
+
+See [Examples](/en-US/docs/Web/API/HTMLScriptElement/text) in {{domxref("HTMLScriptElement")}} (the behaviour is the same).
```
hamishwillee: Note that I didn't duplicate the example, but referred to the `text` page.

Should I do the same for all or the other parts of the description? I was thinking maybe keeping the headings "### `text` vs `textContent` vs `innerText`" and `Security consideration` (so the boilerplate can link to it) and following the same link approach as in Examples.
Or I could in the intro just link to the Description in the other doc.

Not sure. I'm OK with this level of duplication but also with linking since we're saying the behaviour is the same.

files/en-us/web/api/node/textcontent/index.md
```diff
@@ -8,52 +8,33 @@ browser-compat: api.Node.textContent
 
 {{APIRef("DOM")}}
```
hamishwillee, line 9: FYI all the changes in this doc are layout except for the "See also" addition of cross link to the Script properties.

files/en-us/web/api/htmlelement/innertext/index.md
```diff
@@ -8,6 +8,13 @@ browser-compat: api.HTMLElement.innerText
...
 {{APIRef("HTML DOM")}}
 
+> [!WARNING]
+> This property represents the text content of a script element, which may be executable depending on the script type.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
+>
+> You can mitigate this risk by always assigning {{domxref("TrustedScript")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](#security_considerations) for more information.
+
 The **`innerText`** property of the {{domxref("HTMLElement")}} interface represents the rendered text content of a node and its descendants.
```
hamishwillee, line 11: Injected  wrong place.

```suggestion
{{APIRef("HTML DOM")}}

The **`innerText`** property of the {{domxref("HTMLElement")}} interface represents the rendered text content of a node and its descendants.
```

files/en-us/web/api/htmlscriptelement/innertext/index.md
```diff
@@ -0,0 +1,167 @@
...
+If the element is not [being rendered](https://html.spec.whatwg.org/multipage/rendering.html#being-rendered) (for example, is detached from the document or is hidden from view), the returned value is the same as the {{domxref("HTMLScriptElement.textContent","textContent")}} property.
+
+Setting the property accepts either a {{domxref("TrustedScript")}} object or a string.
+
+## Description
+
+The **`innerText`** property of the {{domxref("HTMLScriptElement")}} interface represents the text content inside the {{HTMLElement("script")}} element as though it were rendered text.
+
+When the property is set, the input is normalized the same way as any other element's `innerText` — whitespace is collapsed and `\n` is converted into line breaks.
+This produces the textual representation a user would obtain if a `<script>` element had visible rendering and they were to select its contents and copy them to the clipboard.
```
hamishwillee: This does not appear to be true. If you look at my second example below the output for both methods is the same, which would means that the section on `text` vs `textContent` vs `innerText` is also wrong. I will check

files/en-us/web/api/htmlscriptelement/innertext/index.md
```diff
@@ -0,0 +1,167 @@
...
+This example compares `innerText` with {{domxref("HTMLScriptElement.textContent")}}.
+The result should be that `innerText` squashes any whitespace.
+
+#### HTML
+
+```html
+<h3>Source element:</h3>
+<script id="source">
+const num = 10;       console.log(num);
+const x = 5;
```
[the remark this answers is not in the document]
hamishwillee: Not changing yet. Intentional

files/en-us/web/api/htmlscriptelement/index.md
```diff
@@ -33,22 +36,30 @@ _Inherits properties from its parent, {{domxref("HTMLElement")}}._
...
 - {{domxref("HTMLScriptElement.referrerPolicy")}}
   - : A string that reflects the [`referrerPolicy`](/en-US/docs/Web/HTML/Reference/Elements/script#referrerpolicy) HTML attribute indicating which referrer to use when fetching the script, and fetches done by that script.
 - {{domxref("HTMLScriptElement.src")}}
-  - : A string representing the URL of an external script; this can be used as an alternative to embedding a script directly within a document. It reflects the `src` attribute of the {{HTMLElement("script")}} element.
+  - : A string representing the URL of an external script; this can be used as an alternative to embedding a script directly within a document.
+    It reflects the `src` attribute of the {{HTMLElement("script")}} element.
 - {{domxref("HTMLScriptElement.text")}}
-  - : A string that joins and returns the contents of all {{domxref("Text")}} nodes inside the {{HTMLElement("script")}} element (ignoring other nodes like comments) in tree order. On setting, it acts the same way as the {{domxref("Node.textContent")}} property.
```
hamishwillee, line 45: Not really relevant here.

files/en-us/web/api/htmlscriptelement/innertext/index.md
hamishwillee: @fred-wang I'm trying to work out the differences between `text` vs `textContent` vs `innerText` on `HTMLScriptElement`.
There is a section on this below, but my assumption based on comparing the base classes for `textContent` and `innerText` was that `innerText` would collapse whitespace.

However in the example shows that it does not: https://pr41341.review.mdn.allizom.net/en-US/docs/Web/API/HTMLScriptElement/innerText#comparing_innertext_to_textcontent

So is there any difference between these properties on `HTMLScriptElement` or are they just new versions to allow the TrustedType to be supplied, and the behaviour is the same?

fred-wang: @hamishwillee I'm not aware of any difference. Normally the `HTMLScriptElement` specialization is just to do the TT enforcement before. Do you mean enabling TT causes a behavior change here?

hamishwillee: > Do you mean enabling TT causes a behavior change here?

@fred-wang No. Definitely not!

What I'm asking is whether the `innerText` returns exactly the same value as `text` and `textContent`.

In other elements the returned value may be different because the `innerText` is supposed to be like a rendered version of the DOM, with whitespace removed. But for script elements you can't have inner elements. So my suspicion is that _for script elements_ the result is the same as the other properties and always has been.

The test certainly shows that the output seems the same.

I think you are saying is yes, as far as you know.

fred-wang: OK, I don't recall any specific difference while I was working on Trusted Types, so I would need to check the spec/code to be able to reply. The new versions on HTMLScriptElement are basically to change the setter so that it performs TT enforcement (if not a TrustedScript either throw or convert the input string to the return value of the default policy) then do whatever the parent class was already doing.

hamishwillee: Thanks. The spec is the same for textContent and innerText. I'll adjust all these.

files/en-us/web/api/htmlscriptelement/index.md
```diff
@@ -33,22 +36,30 @@ _Inherits properties from its parent, {{domxref("HTMLElement")}}._
...
-
-    > [!NOTE]
-    > When inserted using the {{domxref("Document.write()")}} method, {{HTMLElement("script")}} elements execute (typically synchronously), but when inserted using {{domxref("Element.innerHTML")}} or {{domxref("Element.outerHTML")}}, they do not execute at all.
```
hamishwillee, line 48: Not relevant at this level. Moved up to description ^^

files/en-us/web/api/htmlscriptelement/index.md
```diff
@@ -33,22 +36,30 @@ _Inherits properties from its parent, {{domxref("HTMLElement")}}._
...
-
+  - : A property that represents the inline text content of the {{HTMLElement("script")}} element.
+    The property accepts either a {{domxref("TrustedScript")}} object or a string.
+    It returns a string that joins the contents of all {{domxref("Text")}} nodes inside the {{HTMLElement("script")}} element (ignoring other nodes like comments) in tree order.
```
hamishwillee: Not relevant for script nodes, which can only have a text node


```suggestion
```

hamishwillee:
@wbamberg This should be good to look at now.

It turns out that I was wrong about the properties `textContent`, `innerText`, and  `text`, being different - testing shows that setting and getting has the same result in all of them. The spec is also the same for `innerText` and `textContent`. The spec for  `text` is very similar (it lacks steps for getter, which I think is an omission)
They aren't different in testing prior to TT either on Firefox.
I think that this is because it is a script element - the script element isn't rendered and can only contain a text element, so most of the differences in definition simply wouldn't apply even in the old spec wording.

Anyway, because of this I have
- removed the comparison section and added a statement they are the same 
- Security considerations and Examples sections all refer to the version in textContent doc
- Descriptions are almost the same. The difference is, if relevant, a comment about the version in the base class.

files/en-us/web/api/htmlscriptelement/textcontent/index.md
```diff
@@ -0,0 +1,146 @@
...
+The **`textContent`** property of the {{domxref("HTMLScriptElement")}} interface represents the inline text content of the {{HTMLElement("script")}} element.
+It behaves in the same way as the {{domxref("HTMLScriptElement.text","text")}} and {{domxref("HTMLScriptElement.innerText","innerText")}} properties.
+
+## Value
+
+Getting the property returns a string containing the script's text.
+
+Setting the property accepts either a {{domxref("TrustedScript")}} object or a string.
+
+## Description
```
hamishwillee, line 32:
```suggestion
## Value

Getting the property returns a string containing the script's text.

Setting the property accepts either a {{domxref("TrustedScript")}} object or a string.

### Exceptions

- `TypeError`
  - : Thrown if the property is set to a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.

## Description
```

files/en-us/web/api/htmlscriptelement/innertext/index.md
```diff
@@ -0,0 +1,62 @@
...
+### Exceptions
+
+- `TypeError`
+  - : Thrown if the property is set to a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
+
+## Description
+
+The **`innerText`** property of the {{domxref("HTMLScriptElement")}} interface represents the text content inside the {{HTMLElement("script")}} element.
+
+For an executable script {{domxref('HTMLScriptElement/type','type')}}, such as a module or classic script, this text is inline executable code.
```
wbamberg:
```suggestion
For an executable script (that is, a script whose {{domxref('HTMLScriptElement/type','type')}} indicates that it is a module or classic script), this text is inline executable code.
```

Picky, but it is grating to me for the property name `type` to be used for the abstract "script type". I'm probably the only person in the world who cares about this.

files/en-us/web/api/htmlscriptelement/innertext/index.md
```diff
@@ -0,0 +1,62 @@
...
+The `innerText` property is also defined on {{domxref("HTMLElement.innerText","HTMLElement")}} and can hence be used with other elements.
+When used with other elements, the property does not expect or enforce the assignment of a {{domxref("TrustedScript")}}.
+
+### Security considerations
+
+See [security considerations](/en-US/docs/Web/API/HTMLScriptElement/textContent#security_considerations) in {{domxref("HTMLScriptElement.textContent")}} (the considerations are the same for `text`, `textContent` and `innerText` properties).
+
+## Examples
+
+See the [Examples](/en-US/docs/Web/API/HTMLScriptElement/textContent#examples) in {{domxref("HTMLScriptElement.textContent")}}.
```
wbamberg:
```suggestion
See the [examples](/en-US/docs/Web/API/HTMLScriptElement/textContent#examples) in {{domxref("HTMLScriptElement.textContent")}}.
```

(or "See the Examples section")

files/en-us/web/api/htmlscriptelement/text/index.md
wbamberg: Same comment here.

files/en-us/web/api/htmlscriptelement/text/index.md
```diff
@@ -33,103 +38,13 @@ For other types it might represent an import map, speculation rules, or some oth
...
-el.text = "const num = 10;\nconsole.log(num)";
-console.log(el.text); // Output: "const num = 10;\nconsole.log(num);"
-console.log(el.textContent); // Output: "const num = 10;\nconsole.log(num);"
-
-// Set the textContent property
-el.textContent = "console.log(10);";
-console.log(el.text); // Output: "console.log(10);"
-console.log(el.textContent); // Output: "console.log(10);"
-```
+See the [Examples](/en-US/docs/Web/API/HTMLScriptElement/textContent#examples) in {{domxref("HTMLScriptElement.textContent")}}.
```
wbamberg: Same comment.

files/en-us/web/api/htmlscriptelement/textcontent/index.md
```diff
@@ -0,0 +1,151 @@
...
+### Exceptions
+
+- `TypeError`
+  - : Thrown if the property is set to a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
+
+## Description
+
+The **`textContent`** property of the {{domxref("HTMLScriptElement")}} interface represents the text content inside the {{HTMLElement("script")}} element.
+
+For an executable script {{domxref('HTMLScriptElement/type','type')}}, such as a module or classic script, this text is inline executable code.
```
wbamberg: Same comment.

files/en-us/web/api/htmlscriptelement/textcontent/index.md
```diff
@@ -0,0 +1,151 @@
...
+
+The `textContent` property — and identical `text` and `innerText` properties — are a possible vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, where potentially unsafe strings provided by a user are executed.
+For example, the following example assumes the `scriptElement` is an executable `<script>` element, and that `untrustedCode` was provided by a user:
+
+```js
+const untrustedCode = "alert('Potentially evil code!');";
+scriptElement.textContent = untrustedCode; // shows the alert
+```
+
+You can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
```
wbamberg:
```suggestion
You can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
```

files/en-us/web/api/htmlscriptelement/textcontent/index.md
```diff
@@ -0,0 +1,151 @@
...
+If that is not possible, you might allow or block the use of certain functions within the provided string.
+
+## Examples
+
+### Using TrustedScript
+
+To mitigate the risk of XSS, we should always assign `TrustedScript` instances to the `textContent` property.
+
+Trusted types are not yet supported on all browsers, so first we define the [trusted types tinyfill](/en-US/docs/Web/API/Trusted_Types_API#trusted_types_tinyfill).
+This acts as a transparent replacement for the trusted types JavaScript API:
```
wbamberg:
```suggestion
This acts as a transparent replacement for the Trusted Types JavaScript API:
```

? it's used as a name here so should get caps?

files/en-us/web/api/htmlscriptelement/textcontent/index.md
```diff
@@ -0,0 +1,151 @@
...
+el.text = "const num = 10;\nconsole.log(num)";
+console.log(`textContent: ${el.textContent}`); // textContent: const num = 10; console.log(num)"
+console.log(`text: ${el.text}`); // "text: const num = 10; console.log(num)"
+console.log(`innerText: ${el.innerText}`); // "innerText: const num = 10; console.log(num)"
+
+// Set the innerText property
+el.innerText = "const num = 10;alert('Help')";
+console.log(`textContent: ${el.textContent}`); // textContent: const num = 10;alert('Help')"
+console.log(`text: ${el.text}`); // "text: const num = 10;alert('Help')"
+console.log(`innerText: ${el.innerText}`); // "innerText: const num = 10;alert('Help')"
```
wbamberg: probably better to put the comment on the next line, so the comment doesn't doesn't wrap weirdly, and have a blank line after each bit, for readability:


```suggestion
el.textContent = "console.log(10);";

console.log(`textContent: ${el.textContent}`);
// "textContent: console.log(10);"

console.log(`text: ${el.text}`);
// "text: console.log(10);"

console.log(`innerText: ${el.innerText}`);
// "innerText: console.log(10);"

// Set the text property
el.text = "const num = 10;\nconsole.log(num)";

console.log(`textContent: ${el.textContent}`);
// textContent: const num = 10; console.log(num)"

console.log(`text: ${el.text}`);
// "text: const num = 10; console.log(num)"

console.log(`innerText: ${el.innerText}`);
// "innerText: const num = 10; console.log(num)"

// Set the innerText property
el.innerText = "const num = 10;alert('Help')";

console.log(`textContent: ${el.textContent}`);
// textContent: const num = 10;alert('Help')"

console.log(`text: ${el.text}`);
// "text: const num = 10;alert('Help')"

console.log(`innerText: ${el.innerText}`);
// "innerText: const num = 10;alert('Help')"
```

files/en-us/web/api/htmlscriptelement/text/index.md
hamishwillee:
```suggestion
See the [examples](/en-US/docs/Web/API/HTMLScriptElement/textContent#examples) in {{domxref("HTMLScriptElement.textContent")}}.
```

files/en-us/web/api/htmlscriptelement/textcontent/index.md
hamishwillee: Thanks, resolved across the TT docs (i.e. including the other doc I copied this from)

hamishwillee:
Thanks @wbamberg - all changes accepted.

wbamberg (approved):
👍 thanks, and sorry to be slow...

PR #41343, opened 2025-09-30
Add hyperlink in page intro

T34-active:
### Description
as title


files/en-us/web/css/unset/index.md
bsmth, line 1: I guess these should all be `zh-CN` -> `en-US`

T34-active: wrong fix

bsmth (approved):
Thanks! I was going to wait for Estelle to have a look, but seeing as I've already looked through, I think we're good to go to save extra reviewer time. Thank you 👍🏻

PR #41345, opened 2025-09-30
Editorial review: Add CSS typed arithmetic docs

### LINKED ISSUE #40988: Update `calc()` documentation based on browser support for typed arithmetic
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/calc

"Input considerations"

Both [Webkit](https://bugs.webkit.org/show_bug.cgi?id=278244) and [Chrome (v140)](https://chromestatus.com/feature/4740780497043456) now have support for typed arithmetic in CSS `calc()`, for example this will work: `calc(2rem / 1px)`. Currently MDN still mentions:

> Current implementations require that, when using the * and / operators, one of the operands must be unitless. For /, the right operand must be unitless. For example font-size: calc(1.25rem / 1.25) is valid but font-size: calc(1.25rem / 125%) is invalid.

- Update the documentation to reflect the latest browser support
- Add a new row under "Browser compatibility" to document this new capability

- https://bugs.webkit.org/show_bug.cgi?id=278244
- https://chromestatus.com/feature/4740780497043456
- https://github.com/mozilla/standards-positions/issues/1252

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/css/calc`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/calc
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/calc/index.md
* Last commit: https://github.com/mdn/content/commit/39a17e10bc078c6e76717683b26a5b20d9d9c574
* Document last modified: 2025-08-09T02:43:04.000Z

</details>

chrisdavidmills:
### Description

Chrome 140 adds support for [CSS typed arithmetic](https://www.w3.org/TR/css-values-4/#calc-type-checking); see https://chromestatus.com/feature/4740780497043456.

It is not very clear what this is from the spec, but https://css-tricks.com/css-typed-arithmetic/ provides a good explanation.

This PR specifically:

- [x] Adds a guide under the CSS values and units module explaining what it is, and showing some examples.
- [x] Adds short sections to the function pages that this new behavior is relevant to (`calc()`. `abs()` and `sign()` for now, although other functions might be relevant too), and edits any mentions of typed arithmetic rules that are rendered out-of-date by this update. 

Fixes https://github.com/mdn/content/issues/40988


danielsakhapov:
I'd say it looks great overall, just two questions:

1) Will "typed arithmetic" be confusing? We used it mostly as a project name, not sure if it's correct to call it like this in the docs. I mean you already can do like calc(1px + 1px) also has types? This is mostly about typed multiplication and division. But on the other hand, the name looks cool :)

2) I don't think we need refs to typed arithmetics from abs()? sign() probably makes sense, but not sure why it can be helpful for abs().

3) Probably add that ref to tan() and atan() to hint that the common de-unit-ication trick can now be done via simple division of units.

4) Heads-up. If you'll work on progress(), probably makes sense to add the ref from there, as progress also allows you to do de-unit-ication.

chrisdavidmills:
> I'd say it looks great overall, just two questions:
> 
>     1. Will "typed arithmetic" be confusing? We used it mostly as a project name, not sure if it's correct to call it like this in the docs. I mean you already can do like calc(1px + 1px) also has types? This is mostly about typed multiplication and division. But on the other hand, the name looks cool :)

@danielsakhapov In the spirit of other hard-to-summarize-succinctly features and techniques (Ajax, DOM scripting, and Houdini also come to mind), it is useful to have a term to familiarize people with, so I think we should keep it. However, it is also true that we might get criticized for accuracy, as typed arithmetic already existed, and really we are referring to a behavioral change to that existing feature.

I've tried to mitigate this by still using the term, but changing the wording to make it clear that it generally refers to existing calculations, but it is also used to refer to this behavioral update.

> 
>     2. I don't think we need refs to typed arithmetics from abs()? sign() probably makes sense, but not sure why it can be helpful for abs().

Ah, good call. I think I got confused and thought that `abs()` accepted calculations directly, but really you'd be putting a `calc()` call inside an `abs()` call for that...removed.

> 
>     3. Probably add that ref to tan() and atan() to hint that the common de-unit-ication trick can now be done via simple division of units.

Good idea. I've added a see also link to both pages.

> 
>     4. Heads-up. If you'll work on progress(), probably makes sense to add the ref from there, as progress also allows you to do de-unit-ication.

True, but I wonder if this is needed because it is the `tan(atan2())` hack that was the most publicized. Progress was also added at a similar time to the typed arithmetic updates, so would people have used it for this purpose as much? Happy to get more input on this if you've got more thoughts.

chrisdavidmills:
@danielsakhapov comments answered. Let me know if you've got any other concerns, or comment with "LGTM" if you are happy ;-)

danielsakhapov:
LGTM!

P.S. About `progress()`. I think when you will describe it you can just mention:
`progress(value, start, end)` is equal to `(value - start) / (end - start)` and can be used to make `value` unitless, but that also can be achieved with [type arithemtic link here], so `progress(10px, 0px, 1px)` is equal to `(10px - 0px) / (1px - 0px)` or just `10px / 1px` and you get unitless `10`.

chrisdavidmills:
> `progress(value, start, end)` is equal to `(value - start) / (end - start)`

@danielsakhapov I added a small section covering this to https://github.com/mdn/content/pull/40159.

files/en-us/web/css/atan2/index.md
```diff
@@ -113,3 +113,4 @@ div.box-5 {
 - {{CSSxRef("asin")}}
 - {{CSSxRef("acos")}}
 - {{CSSxRef("atan")}}
+- [Using CSS typed arithmetic](/en-US/docs/Web/CSS/CSS_Values_and_Units/Using_CSS_typed_arithmetic): Explains how the unitless values achieved via the [`tan(atan2())` hack](https://dev.to/janeori/css-type-casting-to-numeric-tanatan2-scalars-582j) can now be achieved via simple division.
```
estelle:
```suggestion
- [Using CSS typed arithmetic](/en-US/docs/Web/CSS/CSS_Values_and_Units/Using_CSS_typed_arithmetic)
```

files/en-us/web/css/calc/index.md
```diff
@@ -104,10 +104,17 @@ When an {{cssxref("&lt;integer&gt;")}} is expected, the `calc()` expression can
...
 - The `*` and `/` operators do not require whitespace, but adding it for consistency is recommended.
 - It is permitted to nest `calc()` functions, in which case, the inner ones are treated as simple parentheses.
-- Current implementations require that, when using the `*` and `/` operators, one of the operands must be unitless. For `/`, the right operand must be unitless. For example `font-size: calc(1.25rem / 1.25)` is valid but `font-size: calc(1.25rem / 125%)` is invalid.
 - Math expressions involving percentages for widths and heights on table columns, table column groups, table rows, table row groups, and table cells in both auto and fixed layout tables _may_ be treated as if `auto` is specified.
 - See {{cssxref("calc-sum", "&lt;calc-sum&gt;")}} for more information on the syntax of `+` and `-` expressions.
 
+### CSS typed arithmetic
+
+Implementations historically required that, when using the `*` and `/` operators, one of the operands had to be unitless. This is still the case for `*` — calculations like `200px * 4px` don't make sense, as 800px<sup>2</sup> is meaningless in CSS.
```
estelle:
```suggestion
When using `calc()` to multiply values, using the `*` operator, only one value can contain a unit. Calculations like `200px * 4px` are not supported as 800px<sup>2</sup> is meaningless in CSS.
```

files/en-us/web/css/calc/index.md
```diff
@@ -104,10 +104,17 @@ When an {{cssxref("&lt;integer&gt;")}} is expected, the `calc()` expression can
...
+
+For `/`, it used to be the case that the right operand had to be unitless. However, typed arithmetic behavioral changes mean that [supporting browsers](#browser_compatibility) now support calculations such as `100vw / 1px`, which result in unitless values. These can then be used in the values of unitless properties, or converted into a different type altogether by multiplying them by another typed value.
```
estelle:
```suggestion
Inversely, `200px / 4px` resolves to `50`, which does make sense in CSS. For this reason, CSS typed arithmetic does allow for dividing a value by another value if both are of a similar data type. When using the `calc()` function to divide numbers, using the `/` operator, [supporting browsers](#browser_compatibility) allow units on both sides of the operand. For example, two {{cssxref("length")}} values, such as `100vw / 1px`, results in a unitless value. The quotient can then be used in the values of properties or CSS functions that accept a {{cssxref("number")}} as a value or parameter. Alternatively, the quotient can be converted into a different data type by multiplying it by a dimension value or other value with a unit.
```

files/en-us/web/css/calc/index.md
```diff
@@ -104,10 +104,17 @@ When an {{cssxref("&lt;integer&gt;")}} is expected, the `calc()` expression can
...
+
+Such calculations enable interesting relationships to be created between different values on the page, and interesting new UI features to be implemented. For a full explanation and examples, see [Using CSS typed arithmetic](/en-US/docs/Web/CSS/CSS_Values_and_Units/Using_CSS_typed_arithmetic).
```
estelle:
```suggestion
For a full explanation of typed arithmetic in CSS, along with examples, see [Using CSS typed arithmetic](/en-US/docs/Web/CSS/CSS_Values_and_Units/Using_CSS_typed_arithmetic).
```

we can give an example of "interesting relationships" but we shouldn't market one feature as interesting as that means all the other CSS features aren't :P

files/en-us/web/css/css_values_and_units/index.md
```diff
@@ -204,6 +204,9 @@ The CSS values and units module also introduces the [`<frequency>`](/en-US/docs/
 - [Value definition syntax](/en-US/docs/Web/CSS/CSS_Values_and_Units/Value_definition_syntax)
   - : The formal grammar used to define the set of valid values for CSS properties and functions.
 
+- [Using CSS typed arithmetic](/en-US/docs/Web/CSS/CSS_Values_and_Units/Using_CSS_typed_arithmetic)
+  - : An explanation of CSS typed arithmetic behavior, updates to that behavior, and the kinds of features enabled by it.
```
estelle:
```suggestion
  - : An explanation of CSS typed arithmetic behavior and use cases enabled by it.
```
in 10 years will we still want it to read "updates". I assume not, but haven't read the entire thing yet,

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
+---
+title: Using CSS typed arithmetic
+short-title: Using typed arithmetic
+slug: Web/CSS/CSS_Values_and_Units/Using_CSS_typed_arithmetic
```
estelle:
```suggestion
slug: Web/CSS/CSS_values_and_units/Using_CSS_typed_arithmetic
```
i submitted a PR that changes this.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+page-type: guide
+sidebar: cssref
+---
+
+**CSS typed arithmetic** refers generally to permitted calculations performed with typed CSS values via functions such as {{cssxref("calc()")}}. The term is also used to refer specifically to updates to the [CSS values and units](/en-US/docs/Web/CSS/CSS_Values_and_Units) module — and subsequent browser behavior — around these calculations. These updates mean that you can divide a value with one unit by a value with a different unit of the same overall type, resulting in unitless quotients. These quotients can then be used in the values of unitless properties, or converted into a different type altogether by multiplying them by another typed value.
```
estelle:
```suggestion
**CSS typed arithmetic** refers to permitted calculations performed with typed CSS values via functions such as {{cssxref("calc()")}}, and specifically, to the behavior defined in the [CSS values and units](/en-US/docs/Web/CSS/CSS_values_and_units) module. CSS typed arithmetic enables dividing a value with one unit by a value with a different unit of the same data type, resulting in unitless quotients.

These quotients can then be used as a {{cssxref("number")}} in the values of unitless properties, or converted into any numeric data type (such as a {{cssxref("length")}},  {{cssxref("percentage")}}, or  {{cssxref("angle")}}) by multiplying them by a numeric typed value.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+This updated typed arithmetic behavior enables interesting relationships to be created between different values on a page, and interesting new UI features to be rendered in [supporting browsers](/en-US/docs/Web/CSS/calc#browser_compatibility).
+
+This article explores what the updates are in more detail, and presents some examples that make use of them.
```
estelle:
```suggestion
This typed arithmetic behavior enables relationships to be created between different values on a page. This article explores typed arithmetic and presents several examples that make use of it.

>[!NOTE]
> Originally, the [CSS values and units](/en-US/docs/Web/CSS/CSS_Values_and_Units) module limited the arguments multiplication and division operations could take to make dividing by zero detectable at parse time and avoid issues such as squared units. Typed arithmetic relaxes some of those restrictions in [supporting browsers](/en-US/docs/Web/CSS/calc#browser_compatibility).
```

estelle: I am trying to "future proof" the page. Let's put all the history and browser compact in a note that can easily be removed when baseline widely available is achieved.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+## CSS typed arithmetic rules
+
+When performing calculations on values in CSS, there are some rules around compatibility of values with different data types.
+
+### Addition and subtraction
+
+When adding or subtracting values, their units must be the same, or different units within the same overall data type. The following examples all produce valid results:
```
estelle:
```suggestion
When adding or subtracting values, all the values must be within the same overall data type. The following examples all produce valid results:
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+calc(100% - 50px)
+calc(50vw + 2rem)
+calc(25cqw + 3in)
+
+/* <angle> units */
+calc(40deg + 2rad)
+calc(420deg - 1turn)
+```
+
+However, the following do not. The results of such mixed-type calculations are meaningless:
```
estelle:
```suggestion
However, the following are invalid. The results of such mixed-type calculations would be meaningless:
```

changed "are" to "would be" as there are no results, as they're invalid.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+```css example-bad
+calc(200px + 100ms)
+calc(50% + 90deg)
+```
+
+## Multiplication
+
+When multiplying values in CSS, one of the values must always be a unitless {{cssxref("&lt;number>")}} value. This is because you want a product that is a greater amount of the same unit, for example:
```
estelle:
```suggestion
When multiplying values in CSS, only one of the values can have a unit. All other values must be unitless {{cssxref("number")}} values. This is because you want a product that is a greater or lesser amount of the same unit, and do not want to create squared units, which no CSS properties accept. 
```

estelle (commented):
restarting my computer so committing what i have thus far

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+calc(60deg * 3) /* 180deg */
+```
+
+If you try to multiply two values with units — even the same units — normal mathematical rules dictate that the units should be multiplied as well. For example:
+
+```css example-bad
+calc(200px * 4px)
+```
+
+However, the previous function call is invalid: <code>800px<sup>2</sup></code> is meaningless in CSS.
```
estelle:
```suggestion
In this case, the function is invalid as <code>800px<sup>2</sup></code> is meaningless in CSS.
```

"however" didn't quite sync with "For example".

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+## Division
+
+When dividing values in CSS, you can divide a value with a unit by a unitless value:
+
+```css example-good
+calc(1000px / 2) /* 500px */
+calc(360deg / 4) /* 90deg */
+```
+
+However, you can't divide a unitless value by a value with a unit — that doesn't make sense:
```
estelle:
```suggestion
However, you can't divide a unitless value by a value with a unit as that doesn't make sense:
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+```css example-bad
+calc(1000 / 2px) /* ?!? */
+```
+
+In addition — **and this is where it gets interesting** — It used to be the case that you couldn't divide one value with a unit by another, even different units of the same overall type. However, the specification updates mentioned earlier mean **this is now possible**.
```
estelle:
```suggestion
```

suggesting moving lower and switching to a removable note for future proofing

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+`100vw` is equal to `100%` of the width of the viewport. If the viewport is currently `1000px` wide, the above calculation will return the unitless value `1000`. However, if the viewport is `500px` wide, the above calculation will return the unitless value `500`.
+
+If we replace the `1px` divisor by `1em`, we get very different results:
+
+```css
+calc(100vw / 1em)
+```
+
+If the viewport is set to `1000px` wide, and `1em` is equal to the browser default of `16px` at the point of calculation, the previous calculation will return `1000px / 16px` = `62.5`.
+
```
estelle, line 96:
```suggestion

> [!NOTE]
>It used to be the case that you couldn't divide one value with a unit by another, even if the two values has units of the same exact type. However, the specification was updated to enable dividing by units of the same data type. Check for [browser compatibility](#browser_compatibility).
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+However, you can't divide a unitless value by a value with a unit — that doesn't make sense:
+
+```css example-bad
+calc(1000 / 2px) /* ?!? */
+```
+
+In addition — **and this is where it gets interesting** — It used to be the case that you couldn't divide one value with a unit by another, even different units of the same overall type. However, the specification updates mentioned earlier mean **this is now possible**.
+
+When two values with units of the same type are fed into a divison operation, the units cancel each other out, and you are left with a unitless value. Behind the scenes, the two value's computed values are divided by one another.
```
estelle:
```suggestion
When a value of any numeric data type is divided by a value of the same type, the units cancel each other out, and you are left with a unitless value. Behind the scenes, the two values are computed before being divided by one another.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+As a result, the same calculation can have very different results depending on the context it is used in and the units of the divisor.
+
+Take the following example:
+
+```css
+calc(100vw / 1px)
+```
+
+`100vw` is equal to `100%` of the width of the viewport. If the viewport is currently `1000px` wide, the above calculation will return the unitless value `1000`. However, if the viewport is `500px` wide, the above calculation will return the unitless value `500`.
```
estelle:
```suggestion
The `100vw` is equal to `100%` of the width of the viewport. If the viewport is currently `1000px` wide, the calculation will return the unitless value `1000`. However, if the viewport is `500px` wide, the calculation will return the unitless value `500`.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+If we replace the `1px` divisor by `1em`, we get very different results:
+
+```css
+calc(100vw / 1em)
+```
+
+If the viewport is set to `1000px` wide, and `1em` is equal to the browser default of `16px` at the point of calculation, the previous calculation will return `1000px / 16px` = `62.5`.
```
estelle:
```suggestion
If the viewport is `1000px` wide, and `1em` is equal to the browser default of `16px` at the point of calculation, the previous calculation will return `1000px / 16px` = `62.5`.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Why is the behavioral update useful?
+
+The above behavioral change doesn't sound very significant at first, but it enables all kinds of useful associations between different values, resulting in an interesting set of responsive UI features.
```
estelle:
```suggestion
The fact that we can divide a data type value by another value of the same data type may not sound very significant at first, but it enables useful associations between different values, that can be used anywhere numeric data type can be used, including handling responsive UIs. The key to all of this is being able to represent values in a unitless form via divisions like this one:
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+The key to all of this is being able to represent values in a unitless form via divisions like this one (which is already achievable in a limited manner via the [`tan(atan2())` hack](https://dev.to/janeori/css-type-casting-to-numeric-tanatan2-scalars-582j)):
```
estelle:
```suggestion
```
moved first part up. Removed the second part as it won't be relevant now that we don't need a hack.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+```css
+--viewport-width-in-pixels: calc(100vw / 1px);
+```
+
+You can then dynamically vary other property values based on this value, regardless of what units they have. For example:
```
estelle:
```suggestion
The result is a {{cssxref("number")}} that can be used anywhere a number is valid, including other `calc()` functions. With the size of the viewport, you can dynamically vary other property values based on this value, regardless of what units they have. For example:
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+```css
+/* Unitless value can be immediately transferred to opacity */
+opacity: calc(var(--viewport-width-in-pixels) / 1000 - 0.5);
+
+/* Multiply by 1deg to create an <angle> value */
+rotate: calc(var(--viewport-width-in-pixels) * 1deg);
+
+/* Multiply by 1rem to create a <length> value */
+font-size: calc(var(--viewport-width-in-pixels) * 1rem / 200);
+```
```
estelle, line 127: Prefer to not include useful context with comments as those are not as likely to be read. , Let's create three paragraphs, with a css block between each, so the comments are text.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+Let's walk through some examples to show how this technique can be useful.
```
estelle:
```suggestion
Let's walk through some examples to show how CSS typed arithmetic can be useful.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Basic example: responsive background opacity
+
+Our [Responsive background opacity](https://mdn.github.io/dom-examples/css-typed-arithmetic/responsive-background-opacity) example ([see source code](https://github.com/mdn/dom-examples/tree/main/css-typed-arithmetic/responsive-background-opacity)) shows how to vary the opacity of a background image as the viewport width is changed. Having the background image fade out as the screen gets narrower means the image is less likely to affect the readability of the text content as they start to overlap.
```
estelle:
```suggestion
The [responsive background opacity](https://mdn.github.io/dom-examples/css-typed-arithmetic/responsive-background-opacity) example ([see source code](https://github.com/mdn/dom-examples/tree/main/css-typed-arithmetic/responsive-background-opacity)) shows how to vary the opacity of a background image as the viewport width is changed. Having the background image fade out as the screen gets narrower means the image is less likely to affect the readability of the text content as they start to overlap.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+    aliquam dolor, eu lacinia lorem placerat vulputate. Duis felis orci,
+    pulvinar id metus ut, rutrum luctus orci. Cras porttitor imperdiet nunc, at
+    ultricies tellus laoreet sit amet.
+  </p>
+</div>
+```
+
+### CSS
+
+In the CSS, we start by defining a [CSS custom property](/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties) called `--width-percentage` on the {{cssxref(":root")}} element, which contains the result of the calculation `100vw / 2000px`. The aim here is to represent the viewport width as a percentage that we can later use as an alpha channel value. When the viewport is `2000px` wide, the calculation will return `1`, which is equivalent to `100%` alpha. Anything less than a `2000px` viewport width will result in a smaller value.
```
estelle:
```suggestion
We start by defining a [CSS custom property](/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties) called `--width-percentage` on the {{cssxref(":root")}} element, which contains the result of the calculation `100vw / 2000px`. This value represents the viewport width as a percentage of `2000px` that we will later use as an alpha channel value. When the viewport is `2000px` wide, the calculation will return `1`, which is equivalent to `100%` alpha. Anything less than a `2000px` viewport width will result in a smaller value.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+```css
+:root {
+  --width-percentage: calc((100vw / 2000px));
```
estelle, line 157: Should we clamp to `1` or add a sentence that anything greater than 1 as an alpha value will be treated as one?

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+.wrapper {
+  width: 480px;
+  padding: 20px;
+  margin: 0 auto;
+}
+```
+
+Finally, we set multiple backgrounds on the {{htmlelement("body")}} element. We'll go through these in reverse order:
+
+1. The bottom {{cssxref("background")}} is a non-repeating background image that sits up near the top-right corner.
```
estelle:
```suggestion
1. The last and therefore bottom-most {{cssxref("background")}} value is a non-repeating background image located near the top-right corner.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+2. The top background is a white semi-transparent overlay created using a {{cssxref("linear-gradient()")}} function with both color stops set to the same color. The color's alpha channel value is calculated as `1` minus the `--width-percentage` custom property we set earlier: As the viewport width gets narrower, `--width-percentage` will get smaller, therefore the white gradient will become more opaque. This has the effect of reducing the opacity of the background image.
```
estelle:
```suggestion
2. The top background is a white semi-transparent overlay created using a {{cssxref("linear-gradient()")}} function with both color stops set to the same color. The color's alpha channel value is calculated as `1` minus the `--width-percentage` custom property we set earlier: As the viewport width gets narrower, `--width-percentage` will get smaller, therefore the white gradient will become more opaque, reducing the opacity of the background image.
```

files/en-us/web/css/calc/index.md
chrisdavidmills: I like the idea here, but it felt a bit verbose to me, so I've had a go at editing it locally and pushed the result in a separate commit.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+  background:
+    linear-gradient(
+      rgb(255 255 255 / calc(1 - var(--width-percentage))),
+      rgb(255 255 255 / calc(1 - var(--width-percentage)))
+    ),
+    url("https://mdn.github.io/shared-assets/images/examples/colorful-heart.png")
+      no-repeat top 50px right 50px;
+}
+```
+
```
estelle, line 190:
```suggestion

### Results

{{}}

To view the results in a responsive window, see the [responsive background opacity live](https://mdn.github.io/dom-examples/css-typed-arithmetic/responsive-background-opacity).
```

add the results, and link to the live version as well.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+## Varying values of different types based on a single value
+
+Our [Different type variations](https://mdn.github.io/dom-examples/css-typed-arithmetic/different-type-variations) example ([see source code](https://github.com/mdn/dom-examples/tree/main/css-typed-arithmetic/different-type-variations)) works in a very similar way to the previous example, except that in this case we demonstrate that you can use a single variable value in the definition of multiple other values with completely different data types.
```
estelle:
```suggestion
The [different type variations](https://mdn.github.io/dom-examples/css-typed-arithmetic/different-type-variations) example ([see source code](https://github.com/mdn/dom-examples/tree/main/css-typed-arithmetic/different-type-variations)) works in a very similar way to the previous example, except that in this case, we demonstrate the use of a single variable value in the definition of multiple other values with different data types.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+The markup for this example contains a paragraph of text:
+
+```html
+<p>Hello there!</p>
+```
+
+### CSS
+
+We start off our CSS in a similar way to the previous demo, creating a unitless value based on the viewport width. This time, we calculate it by dividing `100vw` by `1px` and store it in a custom property called `--viewport-in-pixels`. This gives us a value that is the current viewport width in pixels, but with no unit.
```
estelle:
```suggestion
We start off our CSS in a similar way to the previous demo, creating a unitless value based on the viewport width. This time, we calculate it by dividing `100vw` by `1px` and store the resulting number in a custom property called `--viewport-in-pixels`. The value represents the current viewport width in pixels, without any units.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+```css
+:root {
+  --viewport-in-pixels: calc(100vw / 1px);
+}
+```
+
+Now on to the styling of the paragraph itself. We first give it some basic styles (a {{cssxref("border")}} and {{cssxref("text-align")}} of `center`), then set two values based on the `--viewport-in-pixels` property we created earlier:
```
estelle:
```suggestion
Now on to the styling of the paragraph itself. We first give it some basic styles (a {{cssxref("border")}} and {{cssxref("text-align")}} of `center`), then set two values based on the `--viewport-in-pixels` property created earlier:
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+1. We set the {{cssxref("font-size")}} of the paragraph to a value equal to `--viewport-in-pixels` divided by `200`, multplied by `1em` to convert it into `em`s.
```
estelle:
```suggestion
1. We set the {{cssxref("font-size")}} of the paragraph to a value equal to `--viewport-in-pixels` divided by `200`, multplied by `1em` to convert the numeric quotient into `em`s.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+2. We set the {{cssxref("background-color")}} of the paragraph to equal an [`lch()`](/en-US/docs/Web/CSS/color_value/lch) color. The lightness and chroma components are constant values (`75%` and `50%`, respectively) whereas the hue component is set to `--viewport-in-pixels` divided by `10`, plus `100`. We then multiply the result of that by `1deg` to specify the value in degrees.
```
estelle:
```suggestion
2. We set the {{cssxref("background-color")}} of the paragraph to an [`lch()`](/en-US/docs/Web/CSS/color_value/lch) color value. The lightness and chroma components are constant values (`75%` and `50%`, respectively), whereas the hue component is set to `--viewport-in-pixels` divided by `10`, plus `100`. We then multiply the result of that by `1deg` to ensure the value is an {{cssxref("angle")}}.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+   > [!NOTE]
+   > This last step is not strictly necessary, as `lch()` also accepts unitless hue values. However, we think a degree value is more intuitive, plus we wanted to show another example of how the unitless value can be cast into a different data type.
```
estelle:
```suggestion
   > This last step is not strictly necessary, as `lch()` also accepts unitless hue values. However, a degree value may be more intuitive, and we wanted to show another example of how the unitless value can be converted into a different data type.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+p {
+  border: 5px solid black;
+  text-align: center;
+  font-size: calc(1em * (var(--viewport-in-pixels) / 200));
+  background-color: lch(
+    75% 50% calc((100 + (var(--viewport-in-pixels) / 10)) * 1deg)
+  );
+}
+```
+
```
estelle, line 236: add the ### result title, macro, and link to live standalone demo.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+## An animated story circle
+
+Our final example — [Animated story circle](https://mdn.github.io/dom-examples/css-typed-arithmetic/animated-story-circle) example ([see source code](https://github.com/mdn/dom-examples/tree/main/css-typed-arithmetic/animated-story-circle)) is a little bit more complicated than the others, and serves to demonstrate the kind of fun effects you can start to create with the new CSS typed arithmetic rules. In this case, we have laid out the different paragraphs of a body of text in a circle, emanating from the center. What's more, the circle of paragraphs closes up into a fan shape when the viewport is made narrower, and opens up again when it is made wider.
```
estelle:
```suggestion
The [animated story circle](https://mdn.github.io/dom-examples/css-typed-arithmetic/animated-story-circle) example ([see source code](https://github.com/mdn/dom-examples/tree/main/css-typed-arithmetic/animated-story-circle)) demonstrates a more complex effect created by leveraging CSS typed arithmetic rules. In this case, the different paragraphs of a body of text are laid out in a circle emanating from a center point. What's more, the circle of paragraphs closes up into a fan shape when the viewport is made narrower, and opens up again when it is made wider.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+:root {
+  height: 100%;
+}
+```
+
+Next, we set several properties on the {{htmlelement("body")}} element:
+
+- We start by setting a `height` of `inherit`, meaning the `<body>` will inherit the `:root` element's `100%` height and therefore span the full height of the viewport.
+- Next, we horizontally center the `<body>` using {{cssxref("margin")}}, and give it a {{cssxref("max-width")}}. As you'll see later on, this upper bound is important for controlling the maximum rotation of the fan/circle shape.
+- We center the child element of the `<body>` (the `<div>` with `class="story-circle"`) horizontally and vertically inside it using [flexbox](/en-US/docs/Web/CSS/CSS_flexible_box_layout).
```
estelle:
```suggestion
- We center all the child elements of the `<body>`, in this case it is just the `<div>` with `class="story-circle"`, horizontally and vertically inside it using [flexbox](/en-US/docs/Web/CSS/CSS_flexible_box_layout).
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+- We use the {{cssxref("container-type")}} property to declare the `<body>` as an inline [size query container](/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_size_queries). This is important because we want to vary the shape rotation based on the `<body>` width, and not the viewport width as we did in previous examples. Setting it as a size query container allows us to reference its size in calculations.
```
estelle, line 278:
```suggestion
- We use the {{cssxref("container-type")}} property to declare the `<body>` as an inline [container size query](/en-US/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_size_queries). This is important because we want to vary the shape rotation based on the `<body>` width, and not the viewport width as we did in previous examples. Setting it as a container size query allows us to reference its size in calculations.
```
i am pretty sure this is the correct name of the feature.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+  margin: 0 auto;
+  max-width: 1600px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  container-type: inline-size;
+}
+```
+
+We now come to styling the `story-circle` `<div>`. We give it a `width` and `height` of `1px`, meaning it acts as a reference point for the paragraphs to be positioned around in the center of the screen (we don't even need to set positioning on it, as it is fine for the paragraphs to be positioned relative to the `<body>`).
```
estelle:
```suggestion
We now come to styling the `story-circle` `<div>`. We set the centered container's `width` and `height` to `1px`, meaning the container of all the paragraphs is just a tiny reference point being used as the origin point for the paragraphs that will be positioned relative to it, positioning the container and its descendats around in the center of the screen (we don't even need to set positioning on it, as it is fine for the paragraphs to be positioned relative to the `<body>`).
```

i had a hard time parsing this paragraph, so tried to clarify via elaboration

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+```css
+.story-circle {
+  width: 1px;
+  height: 1px;
+  --width-percentage: calc((100cqw / 1200px) - 0.33333);
+}
+```
+
+We are using the container query width rather than the viewport width to control the rotation because we want to set a maximum bound for it, which is controlled by the `<body>` `max-width`. If the rotation value becomes greater than a full circle, the last paragraph starts to overlap the first one, which spoils the effect.
```
estelle:
```suggestion
We are using the container query width rather than the viewport width to control the rotation because we want to set a maximum bound for the container, which is controlled by the `<body>` `max-width`. If the rotation value becomes greater than a full circle, the last paragraph will start to overlap the first one, which spoils the effect.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+Since the `max-width` is `1600px`, you might have expected the calculation to be `100cqw / 1600px)`. This would work, but we've instead gone for `(100cqw / 1200px) - 0.33333` (the `0.33333` comes from `1600px/1200px - 1`). This still causes the maximum rotation to occur at a `<body>` width of `1600px`, but the fan now has a smaller minimum rotation, which gives a better effect at narrow viewport widths.
```
estelle:
```suggestion
Since the `max-width` is `1600px`, you might have expected the calculation to be `100cqw / 1600px)`. This would work, but we've instead gone for `(100cqw / 1200px) - 0.33333` (the `0.33333` comes from `1600px/1200px - 1`). Both would cause the maximum rotation to occur at a `<body>` width of `1600px`, but the fan now has a smaller minimum rotation, which gives a better effect at narrow viewport widths.
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+The final style rule selects the paragraphs themselves. Most of this styling is rudimentary. It is worth pointing out that we have set {{cssxref("position")}} to `absolute` to cause all of the paragraphs to sit on top of one another, and we've set a {{cssxref("transform-origin")}} value of `center left` to cause paragraphs to rotate around the center of their left edge, so they will all fan out from a center point over their containing `<div>`.
+
+Now onto the interesting bit — we define a custom property called `--angle` that contains the unitless rotation angle of the paragraph, before then setting the {{cssxref("rotate")}} property to that value multiplied by `1deg` to convert it to a degree value. The `--angle` custom property is equal to:
```
estelle:
```suggestion
Now onto the interesting bit — we define a custom property called `--angle` that contains the unitless rotation angle of the paragraph, before setting the {{cssxref("rotate")}} property to the resulting value. We multiply the resulting number by `1deg` to convert it to a degree value. The `--angle` custom property is the product of three values:
```

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+
+- The paragraph's {{cssxref("sibling-index()")}} - `1` (which causes the first paragraph to have a rotation angle of `0`, as we want it to be horizontal)
+- multiplied by
+- `360` divided by the paragraph's {{cssxref("sibling-count()")}} (which causes all the paragraphs to be equally spaced around the circle, meaning the design will still work if paragraphs are added or removed)
+- multiplied by
+- our `--width-percentage` custom property (this causes the rotation of the paragraphs around the circle to vary as the viewport width is changed). Remember that this property has a maximum value of `1`, which will be achieved when the `<body>` element hits its `max-width` of `1600px`.
```
estelle:
```suggestion
1. The paragraph's {{cssxref("sibling-index()")}} - `1`, which causes the first paragraph to have a rotation angle of `0`, as we want it to be horizontal.
2. `360` divided by the paragraph's {{cssxref("sibling-count()")}}, which causes all the paragraphs to be equally spaced around the circle, meaning the design will still work if the number of paragraphs changes.
3. our `--width-percentage` custom property, which causes the rotation of the paragraphs around the circle to vary as the viewport width changes. Remember that this property has a maximum value of `1`, which will be achieved when the `<body>` element hits its `max-width` of `1600px`.
```
by changing the previous paragraph, we can get rid of the "multiplied by"

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,319 @@
...
+  );
+  rotate: calc(var(--angle) * 1deg);
+}
+```
+
+## See also
+
+- {{cssxref("calc()")}}, {{cssxref("abs()")}}
+- [CSS values and units](/en-US/docs/Web/CSS/CSS_Values_and_Units) module
+- [CSS Typed Arithmetic](https://css-tricks.com/css-typed-arithmetic/) on css-tricks.com (2025)
```
estelle, line 342:
```suggestion
- [CSS typed arithmetic](https://css-tricks.com/css-typed-arithmetic/) via css-tricks.com (2025)
```

files/en-us/web/css/sign/index.md
```diff
@@ -73,3 +73,4 @@ div {
 ## See also
 
 - {{CSSxRef("abs")}}
+- [Using CSS typed arithmetic](/en-US/docs/Web/CSS/CSS_Values_and_Units/Using_CSS_typed_arithmetic)
```
estelle:
```suggestion
- [Using CSS typed arithmetic](/en-US/docs/Web/CSS/CSS_values_and_units/Using_CSS_typed_arithmetic)
```

files/en-us/web/css/tan/index.md
```diff
@@ -107,3 +107,4 @@ body {
 - {{CSSxRef("acos")}}
 - {{CSSxRef("atan")}}
 - {{CSSxRef("atan2")}}
+- [Using CSS typed arithmetic](/en-US/docs/Web/CSS/CSS_Values_and_Units/Using_CSS_typed_arithmetic): Explains how the unitless values achieved via the [`tan(atan2())` hack](https://dev.to/janeori/css-type-casting-to-numeric-tanatan2-scalars-582j) can now be achieved via simple division.
```
estelle:
```suggestion
- [Using CSS typed arithmetic](/en-US/docs/Web/CSS/CSS_values_and_units/Using_CSS_typed_arithmetic)
```

estelle (commented):
Future proofing: opting for notes and no "new" or currently, etc.

estelle (commented):
Future proofing: opting for notes and no "new" or currently, etc.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: The name of the actual article is cased in that way, plus I don't like "via" in this context. No change here.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: I don't agree with this change. The feature is called a "container size query", yes, but the element the query is applied to is called a "size query container", and that's what we are referring to.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: done

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: I've added the note but played with the wording a bit. Update will be in my next commit.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: done in next commit; simplified the wording a bit.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: done

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: done, but simplified in next commit

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: done in next commit

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: added note; I don't want to change the code at this point if at all possible.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: Done. I'm not adding live examples in, as you can't really see the point of any of them in a small embedded frame.

chrisdavidmills: But I have added live links and an explanation of what you should see.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: I'm not convinced this helps. We already explained what the value is earlier on. No change.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: added

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: I think this sentence and your suggestion are both a bit too complicated. I had a play with it and ended up with:

> We center the `<div>` with `class="story-circle"` horizontally and vertically inside the `<body>` using [flexbox](/en-US/docs/Web/CSS/CSS_flexible_box_layout).

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: Nice try, but I think this has become a bit verbose and meandering. I have end up with the following in my next commit:

> We now come to styling the `story-circle` `<div>`. We set its `width` and `height` to `1px`: it will act as a reference point with its child paragraphs positioned in a circle around it. (We don't even need to set positioning on it, as it is fine for the paragraphs to be positioned relative to the `<body>`).

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
estelle: the "it" can be unclear

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
```diff
@@ -0,0 +1,342 @@
...
+
+The final style rule selects the paragraphs themselves. Most of this styling is rudimentary. It is worth pointing out that we have set {{cssxref("position")}} to `absolute` to cause all of the paragraphs to sit on top of one another, and we've set a {{cssxref("transform-origin")}} value of `center left` to cause paragraphs to rotate around the center of their left edge, so they will all fan out from a center point over their containing `<div>`.
+
+Now onto the interesting bit — we define a custom property called `--angle` that contains the unitless rotation angle of the paragraph, before setting the {{cssxref("rotate")}} property to the resulting value. We multiply the resulting number by `1deg` to convert it to a degree value. The `--angle` custom property is the product of three values:
+
+1. The paragraph's {{cssxref("sibling-index()")}} - `1`, which causes the first paragraph to have a rotation angle of `0`, as we want it to be horizontal.
```
estelle: either change the "-" to the word "minus" (or something similar) or make the whole thing code, maybe {{cssxref("sibling-index()", ""sibling-index() - 1")}}

estelle (approved):
just two nits: the "it" and the "-". Approving.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: OK, good point. Applying.

files/en-us/web/css/css_values_and_units/using_css_typed_arithmetic/index.md
chrisdavidmills: Done! Thanks for the review!

PR #41347, opened 2025-09-30
tab widget example - update color of focused tabpanel border

scottaohara:
Updated the focus style for when a tabpanel receives focus.  the previous color was orange, but that did not pass minimum non-text contrast.  i just arbitrarily picked the new color because it does pass contrast against the white background, and it had more than a 3:1 ratio compared to the non-focused border color, so it'd still generally be a noticeable change.


Josh-Cena (commented):
Thank you!

Josh-Cena (approved):

PR #41354, opened 2025-09-30
Mention how to immediately remove cookies

### LINKED ISSUE #25324: “Using HTTP cookies” needs a note on how to delete/unset a cookie
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies

The Set-Cookie and Cookie headers

Section explains how to create a cookie, but not how to delete one.

There's lots of articles on the web with different opinions about how to remove a cookie. I would like MDN to provide some authoritative guidance on the recommended way to immediately remove a cookie that was previously created.

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/http/cookies`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/cookies/index.md
* Last commit: https://github.com/mdn/content/commit/eb9eef29f1ccdaf1c8a464dbe4483c78f7a13b2a
* Document last modified: 2023-03-03T04:45:34.000Z

</details>

Josh-Cena:
Fix https://github.com/mdn/content/issues/25324

### DIFF
--- a/files/en-us/web/http/guides/cookies/index.md
+++ b/files/en-us/web/http/guides/cookies/index.md
@@ -92,6 +92,12 @@ You can specify an expiration date or time period after which the cookie should 
   > [!NOTE]
   > If your site authenticates users, it should regenerate and resend session cookies, even ones that already exist, whenever a user authenticates. This approach helps prevent [session fixation](https://owasp.org/www-community/attacks/Session_fixation) attacks, where a third-party can reuse a user's session.
 
+To immediately remove a cookie, set the cookie again with the same name, path, and domain (if specified), and set its `Expires` attribute to a date in the past or its `Max-Age` attribute to `0` or negative. This instructs the browser to delete the cookie right away. For example:
+
+```http
+Set-Cookie: id=a3fWa; Max-Age=0
+```
+
 There are some techniques designed to recreate cookies after they're deleted. These are known as "zombie" cookies. These techniques violate the principles of user [privacy](#privacy_and_tracking) and control, may violate [data privacy regulations](#cookie-related_regulations), and could expose a website using them to legal liability.
 
 ### Updating cookie values
--- a/files/en-us/web/http/guides/cookies/index.md
+++ b/files/en-us/web/http/guides/cookies/index.md
@@ -98,7 +98,7 @@ To immediately remove a cookie, set the cookie again with the same name, path, a
 Set-Cookie: id=a3fWa; Max-Age=0
 ```
 
-It is also possible to clear all cookies associated with a registrable domain using the {{httpheader("Clear-Site-Data")}} response header.
+You can also clear all cookies associated with a registrable domain using the {{httpheader("Clear-Site-Data")}} response header.
 For example, the following header sent from `https://foo.example.com/` would clear all cookies sent by `example.com` and all of its subdomains, such as `all.bar.example.com`.
 
 ```http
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/guides/cookies/index.md
```diff
@@ -92,6 +92,12 @@ You can specify an expiration date or time period after which the cookie should
   > [!NOTE]
   > If your site authenticates users, it should regenerate and resend session cookies, even ones that already exist, whenever a user authenticates. This approach helps prevent [session fixation](https://owasp.org/www-community/attacks/Session_fixation) attacks, where a third-party can reuse a user's session.
 
+To immediately remove a cookie, set the cookie again with the same name, path, and domain (if specified), and set its `Expires` attribute to a date in the past or its `Max-Age` attribute to `0` or negative. This instructs the browser to delete the cookie right away. For example:
+
+```http
+Set-Cookie: id=a3fWa; Max-Age=0
+```
+
```
hamishwillee, line 100: @Josh-Cena Given that it was mentioned in the associated issue, I think we should also mention `Clear-Site-Data`. This could be in a separate section "Clearing all cookies". Thoughts?


````suggestion

It is also possible to clear all cookies associated with a registrable domain using the {{httpheader("Clear-Site-Data")}} response header.
For example, the following header sent from `https://foo.example.com/` would clear all cookies sent by `example.com` and all of its subdomains, such as `all.bar.example.com`.

```http
Clear-Site-Data: "cookies"
```

````

Comment (Josh-Cena) [reply files/en-us/web/http/guides/cookies/index.md:100]: I am unsure of the scope of CSD. I'm pretty sure it's only limited to subdomains (the spec says it uses "domain-match"), but both your edit here and the comment in the original issue suggest that it's to everything under the same TLD+1.

Comment (hamishwillee) [reply files/en-us/web/http/guides/cookies/index.md:100]: You are right to be unsure. I believe the comment to be true based on this part of the spec

> When delivered with a response from https://example.com/clear, the following header will cause cookies associated with the origin https://example.com to be cleared, as well as cookies on any origin in the same **registered domain** (e.g. https://www.example.com/ and https://more.subdomains.example.com/).

A registered domain is the top level domain, which matches my text.

@Jxck Can you confirm this point I am adding in https://github.com/mdn/content/pull/41354#discussion_r2400462543 - i.e that clear affects the whole domain irrespective of which subdomain it is served from.

Comment (Jxck) [reply files/en-us/web/http/guides/cookies/index.md:100]: Hmm I'm not sure about the scope it affects.
@annevk is a better person to answer correctly.

Comment (annevk) [reply files/en-us/web/http/guides/cookies/index.md:100]: It's called a "registrable domain" (see URL standard). I see that Clear-Site-Data says "registered domain", but that's wrong. In any event, in terms of scope it's correct.

https://w3c.github.io/webappsec-clear-site-data/#clear-cookies
https://url.spec.whatwg.org/#host-registrable-domain

Comment (hamishwillee) [reply files/en-us/web/http/guides/cookies/index.md:100]: Thanks! @Josh-Cena I have updated the suggestion to correctly spell "registrable domain"

File: files/en-us/web/http/guides/cookies/index.md [hunk 2]
```diff
@@ -92,6 +92,19 @@ You can specify an expiration date or time period after which the cookie should
...
+It is also possible to clear all cookies associated with a registrable domain using the {{httpheader("Clear-Site-Data")}} response header.
```
Josh-Cena:
```suggestion
You can also clear all cookies associated with a registrable domain using the {{httpheader("Clear-Site-Data")}} response header.
```

hamishwillee (approved):
Thanks @Josh-Cena !

PR #41360, opened 2025-10-01
Add missing quotes in console log

genseirin:
### Description

The quotes were only delimiting the string, rather than being part of it.

### Motivation

Beginners might be confused about the possibility of "nested double quotes".


Josh-Cena (approved):
Thanks, good catch! We should keep the quotes because the example in `indexOf` has them.

PR #41365, opened 2025-10-01
Update index.md

iacore:
### Description

Fix "html5 is only in utf-8" bullshit.

Not sure if the link I put is ok.

### Motivation

Remove hallucination

### Additional details

https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding
https://html.spec.whatwg.org/multipage/semantics.html#charset

### DIFF
--- a/files/en-us/web/html/reference/elements/meta/index.md
+++ b/files/en-us/web/html/reference/elements/meta/index.md
@@ -24,7 +24,7 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
 > The [`itemprop`](/en-US/docs/Web/HTML/Reference/Global_attributes/itemprop) attribute must not be set on a `<meta>` element that includes a [`name`](/en-US/docs/Web/HTML/Reference/Elements/meta/name), [`http-equiv`](/en-US/docs/Web/HTML/Reference/Elements/meta/http-equiv), or [`charset`](#charset) attribute.
 
 - `charset`
-  - : This attribute declares the document's character encoding. If the attribute is present, its value must be an ASCII case-insensitive match for the string `"utf-8"`, because UTF-8 is the only valid encoding for HTML5 documents. `<meta>` elements which declare a character encoding must be located entirely within the first 1024 bytes of the document.
+  - : This attribute declares the document's [character encoding](https://html.spec.whatwg.org/multipage/semantics.html#charset). `<meta>` elements which declare a character encoding must be located entirely within the first 1024 bytes of the document.
 - [`content`](/en-US/docs/Web/HTML/Reference/Attributes/content)
   - : This attribute contains the value for the [`http-equiv`](/en-US/docs/Web/HTML/Reference/Elements/meta/http-equiv) or [`name`](/en-US/docs/Web/HTML/Reference/Elements/meta/name) attribute, depending on which is used.
 - [`http-equiv`](/en-US/docs/Web/HTML/Reference/Elements/meta/http-equiv)


Josh-Cena:
You linked to the spec yourself, which says this:

> If the attribute is present, its value must be an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "utf-8".

and this:

> The Encoding standard requires use of the [UTF-8](https://encoding.spec.whatwg.org/#utf-8) [character encoding](https://encoding.spec.whatwg.org/#encoding) and requires use of the "utf-8" [encoding label](https://encoding.spec.whatwg.org/#label) to identify it. Those requirements necessitate that the document's [character encoding declaration](https://html.spec.whatwg.org/multipage/semantics.html#character-encoding-declaration), if it exists, specifies an [encoding label](https://encoding.spec.whatwg.org/#label) using an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for "utf-8". Regardless of whether a [character encoding declaration](https://html.spec.whatwg.org/multipage/semantics.html#character-encoding-declaration) is present or not, the actual [character encoding](https://dom.spec.whatwg.org/#concept-document-encoding) used to encode the document must be [UTF-8](https://encoding.spec.whatwg.org/#utf-8).

So I'm not sure where you got the impression that the current content is wrong.

Please be mindful of your language next time. This constitutes a CoC violation.

iacore:
Sorry about the language, but are these page not real? Even Firefox doesn't obey the spec.


https://mbays.freeshell.org/intricacy/

image

image


```
<!doctype html>
<meta charset="big5">
測試
```

image



> A character encoding declaration is required (either in the [Content-Type metadata](https://html.spec.whatwg.org/#content-type) or explicitly in the file) even when all characters are in the ASCII range, because a character encoding is needed to process non-ASCII characters entered by the user in forms, in URLs generated by scripts, and so forth.
> 
> Using non-UTF-8 encodings can have unexpected results on form submission and URL encodings, which use the [document's character encoding](https://dom.spec.whatwg.org/#concept-document-encoding) by default.

unexpected, but allowed.

Again,

> 13.2.3.3 Character encodings

User agents must support the encodings defined in Encoding, including, but not limited to, [UTF-8](https://encoding.spec.whatwg.org/#utf-8), [ISO-8859-2](https://encoding.spec.whatwg.org/#iso-8859-2), [ISO-8859-7](https://encoding.spec.whatwg.org/#iso-8859-7), [ISO-8859-8](https://encoding.spec.whatwg.org/#iso-8859-8), [windows-874](https://encoding.spec.whatwg.org/#windows-874), [windows-1250](https://encoding.spec.whatwg.org/#windows-1250), [windows-1251](https://encoding.spec.whatwg.org/#windows-1251), [windows-1252](https://encoding.spec.whatwg.org/#windows-1252), [windows-1254](https://encoding.spec.whatwg.org/#windows-1254), [windows-1255](https://encoding.spec.whatwg.org/#windows-1255), [windows-1256](https://encoding.spec.whatwg.org/#windows-1256), [windows-1257](https://encoding.spec.whatwg.org/#windows-1257), [windows-1258](https://encoding.spec.whatwg.org/#windows-1258), [GBK](https://encoding.spec.whatwg.org/#gbk), [Big5](https://encoding.spec.whatwg.org/#big5), [ISO-2022-JP](https://encoding.spec.whatwg.org/#iso-2022-jp), [Shift_JIS](https://encoding.spec.whatwg.org/#shift_jis), [EUC-KR](https://encoding.spec.whatwg.org/#euc-kr), [UTF-16BE](https://encoding.spec.whatwg.org/#utf-16be), [UTF-16LE](https://encoding.spec.whatwg.org/#utf-16le), [UTF-16BE/LE](https://encoding.spec.whatwg.org/#utf-16be-le), and [x-user-defined](https://encoding.spec.whatwg.org/#x-user-defined). User agents must not support other encodings.

https://html.spec.whatwg.org/multipage/parsing.html#character-encodings

Again,

Point 9 of https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding says, the guess encoding could not be UTF-8

Josh-Cena:
The page you shared is in quirks mode, so any behavior could ensue. It's also very possible that browsers do support more things, but those are non-standard and deprecated. Only the things HTML acknowledge should be valid HTML; browsers try to be lenient but those things won't be documented.

iacore:
Then I think the wording is at best "should", not "must". Should MDN not reflect what is actually used in the wild?

Josh-Cena:
"Must" is also a normative requirement. We document what we believe you should write, and if you deviate then you should expect to see anything: either your app still works as intended or it doesn't, but unfortunately we can't consistently describe these non-happy paths.

PR #41368, opened 2025-10-02
Add y8.com to independent game portals section

agejevasv:
### Description

Add y8.com to independent game portals/publishers section.


hamishwillee:
Thanks, but I'm going to close this.

MDN doesn't try to be exhaustive when we list external resources - these kinds of links are hard to keep current. In fact we might be better off reducing the current list.

PR #41369, opened 2025-10-02
Reference more articles about font smoothing

pioug:
### Description

Add 2 articles about font smoothing.

### Motivation

These two articles helped me decide whether font smoothing is still relevant. Not only they are more recent than the other articles linked right now, but they also offer a different perspective.


Josh-Cena (approved):
Sure, thank you!

PR #41372, opened 2025-10-02
Add information about Chrome DevTools extensibility API

tunetheweb:
### Description

[Chrome DevTool's extensibility API](https://developer.chrome.com/docs/devtools/performance/extension) extends some standard (and non-standard!) Web APIs with either new additions, or in the use of structured versions of existing fields.

Therefore I think it is helpful for developers who may see these being used in the wild to understand them.

At the same time, as these are non-standard extensions (at least for now!) I don't think it should be documented on MDN other than a small note. I'm also conscious that MDN is not a documentation site for specific browsers, but rather the web platform which is another reason to keep this light. Happy to take feedback on this and iterate if you don't think I've struck the right balance here.

### Motivation

A developer approached me confused that the `console.timeStamp` MDN docs suggested this API did not allow for the extensions Chrome uses.

Additionally, with the likes of popular libraries like [React](https://react.dev/reference/dev-tools/react-performance-tracks) and [Angular](https://angular.dev/best-practices/profiling-with-chrome-devtools) uses the extensibility API, this is becoming a more popular API.

Firefox have also expressed interest to potentially implementing these extensions (at which point we can document this more on MDN as it's more cross-browser).

### Additional details

https://developer.chrome.com/docs/devtools/performance/extension


files/en-us/web/api/console/timestamp_static/index.md
```diff
@@ -25,6 +25,9 @@ console.timeStamp(label)
 - `label` {{Optional_Inline}}
   - : Label for the timestamp.
 
+> [!NOTE]
+> Chrome has extended this `console.timeStamp()` method to allow additional, optional parameters to be provided as part of its extensibility API that surfaces these in performances traces. See the [Chrome's extensibility API documentation](https://developer.chrome.com/docs/devtools/performance/extension#inject_your_data_with_consoletimestamp) for more information.
+
```
wbamberg: The normal thing to do here would be to document the extra parameters in the MDN page, and update the BCD for this method to indicate that they are Chromium-only.

files/en-us/web/api/performance/mark/index.md
```diff
@@ -80,6 +80,10 @@ performance.mark("login-button-pressed", {
 });
 ```
 
+### Chrome DevTools Extensibility API
+
+Chrome DevTools uses `performance.mark()` and in particular a structured `detail` property as part of its extensibility API that surfaces these in custom tracks in performance traces. See the [Chrome's extensibility API documentation](https://developer.chrome.com/docs/devtools/performance/extension#inject_your_data_with_consoletimestamp) for more information and examples.
+
```
wbamberg: This should not be in examples. Examples should never add new factual information about an API. Examples should be about illustrating information that's already given in the page. Someone should be able to skip Examples entirely and still know everything about the API.

In this cases, as I understand things, it might be best to add a paragraph to the description of the `detail` parameter, linking to its special usage in Chrome.

It also appears that the link here isn't right - it only talks about `timeStamp`. Should the link be https://developer.chrome.com/docs/devtools/performance/extension#inject_your_data_with_the_user_timings_api ?

files/en-us/web/api/performancemark/index.md
```diff
@@ -20,7 +20,12 @@ Entries of this type are typically created by calling {{domxref("Performance.mar
 
 ## Instance properties
 
-This interface extends the following {{domxref("PerformanceEntry")}} properties by qualifying/constraining the properties as follows:
+This interface defines:
+
+- {{domxref("PerformanceMeasure.detail")}}
+  - : Contains arbitrary metadata about the measure.
+
+In addition, it extends extends the following {{domxref("PerformanceEntry")}} properties by qualifying/constraining the properties as follows:
```
wbamberg:
```suggestion
In addition, it extends the following {{domxref("PerformanceEntry")}} properties by qualifying/constraining the properties as follows:
```

I would prefer this to say "This interface also", rather than "In addition, it", but am not making this suggestion so as to stay consistent with the language in PerformanceMeasure. But if you'd like to use that wording in both places I am +1 to that.

files/en-us/web/api/performancemark/index.md
```diff
@@ -20,7 +20,12 @@ Entries of this type are typically created by calling {{domxref("Performance.mar
 
 ## Instance properties
 
-This interface extends the following {{domxref("PerformanceEntry")}} properties by qualifying/constraining the properties as follows:
+This interface defines:
+
```
wbamberg: I would prefer not to have this, as we never do usually, but am not making this suggestion so as to stay consistent with the language in PerformanceMeasure. But if you'd like to remove it in both places I am +1 to that.

files/en-us/web/api/console/timestamp_static/index.md
tunetheweb: Opened BCD PR here: https://github.com/mdn/browser-compat-data/pull/28270

Will mark this PR as draft until that's merged and return to it later.

files/en-us/web/api/performance/mark/index.md
tunetheweb: Fixed.

files/en-us/web/api/performancemark/index.md
tunetheweb: Fixed in both.

files/en-us/web/api/performancemark/index.md
tunetheweb: Removed from both.

files/en-us/web/api/console/timestamp_static/index.md
tunetheweb: Actually I think now this CAN be reviewed independently.

tunetheweb:
Thanks @wbamberg . I've now done this the proper way, rather than trying to wedge it in as a note.

wbamberg (approved):
👍 thank you.

files/en-us/web/api/console/timestamp_static/index.md
```diff
@@ -14,21 +14,94 @@ The **`console.timeStamp()`** static method adds a single marker to the browser'
 
 You can optionally supply an argument to label the timestamp, and this label will then be shown alongside the marker.
 
+Some browsers have further extended this `console.timeStamp()` method to allow additional, optional parameters to be provided as part of its extensibility API that surfaces these in performances traces. See the [Chrome's extensibility API documentation](https://developer.chrome.com/docs/devtools/performance/extension#inject_your_data_with_consoletimestamp) for more information.
```
Penya1313, line 17: []()

PR #28569, opened 2023-08-15
Add `firstInterimResponseStart` and `deliveryType` to PerformanceResourceTiming

tunetheweb:
### Description

Add `deliveryType` and `firstInterimResponseStart` to PerformanceResourceTiming

I also added experimental status to `renderBlockingStatus` as currently Chromium only.

### Motivation

New additions landing in Chrome

### Additional details

- firstInterimResponseStart in Chrome 115 - https://chromestatus.com/feature/5086730938482688
- deliveryType in Chrome 117 - https://chromestatus.com/feature/6347141115543552

### Related issues and pull requests

These have already been added to browser-compat-data by @Elchi3 in https://github.com/mdn/browser-compat-data/pull/20518 and https://github.com/mdn/browser-compat-data/pull/20519

PR #41373, opened 2025-10-02
Remove descrepencies and duplication between Resource Timing  and PerformanceResourceTiming docs

tunetheweb:
### Description

There's two Resource Timing docs:
- https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Resource_timing
- https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming

At present the timings are duplicated in both but some new timings were only added to one doc in previous PRs (#28569 and #37740).

This PR removes duplication to remove this inconsistency and adds references too each. We agreed below that we'd have:

- [Resource timing overview](https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Resource_timing) would be the general guide and usage.
- [PerformanceResourceTiming reference](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) would detail all the specific fields.

### Motivation

Fix inconsistencies.

We maybe should consider if it's really that useful to duplicate this information in two different places as I'm sure this is bound to happen again.

Previous PRs:
- https://github.com/mdn/content/pull/28569
- https://github.com/mdn/content/pull/37740


files/en-us/web/api/performance_api/resource_timing/index.md
```diff
@@ -30,8 +30,10 @@ An application can get timestamps for the various stages used to load a resource
...
-12. {{domxref('PerformanceResourceTiming.responseStart','responseStart')}}: timestamp immediately after the browser receives the first byte of the response from the server, cache, or local resource.
-13. {{domxref('PerformanceResourceTiming.responseEnd','responseEnd')}}: timestamp immediately after the browser receives the last byte of the resource or immediately before the transport connection is closed, whichever comes first.
+12. {{domxref('PerformanceResourceTiming.firstInterimResponseStart','firstInterimResponseStart')}}: timestamp immediately after the browser receives the first byte of the first [interim (or informational) response](/en-US/docs/Web/HTTP/Reference/Status#informational_responses) (for example [103 Early Hints](/en-US/docs/Web/HTTP/Reference/Status/103)).
+13. {{domxref('PerformanceResourceTiming.responseStart','responseStart')}}: timestamp immediately after the browser receives the first byte of the response from the server, cache, or local resource. Which may be an interim response.
```
wbamberg:
```suggestion
13. {{domxref('PerformanceResourceTiming.responseStart','responseStart')}}: timestamp immediately after the browser receives the first byte of the response from the server, cache, or local resource. This may be an interim response.
```

wbamberg (commented):
> We maybe should consider if it's really that useful to duplicate this information in two different places as I'm sure this is bound to happen again.

Yes, I agree. Of the two I would choose to keep https://pr41373.review.mdn.allizom.net/en-US/docs/Web/API/Performance_API/Resource_timing, partly because it's guide-y cvontent I think. But it would be good to cross-link from https://pr41373.review.mdn.allizom.net/en-US/docs/Web/API/PerformanceResourceTiming#description, we could say something like "To learn about the resource timing metrics that are typically calculated using this API, see ...".

What do you think?

tunetheweb:
> > We maybe should consider if it's really that useful to duplicate this information in two different places as I'm sure this is bound to happen again.
> 
> Yes, I agree. Of the two I would choose to keep https://pr41373.review.mdn.allizom.net/en-US/docs/Web/API/Performance_API/Resource_timing, partly because it's guide-y cvontent I think. But it would be good to cross-link from https://pr41373.review.mdn.allizom.net/en-US/docs/Web/API/PerformanceResourceTiming#description, we could say something like "To learn about the resource timing metrics that are typically calculated using this API, see ...".
> 
> What do you think?

I think we should go the other way actually.

The [Resource Timing page](https://developer.mozilla.org//en-US/docs/Web/API/Performance_API/Resource_timing) as as you say more of a guide — which I see as a high-level explanation, rather than a detailed reference of each field. In fact it already delegates the non-timing sections to [the other page](https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Resource_timing#other_properties)

The [PerformanceResourceTiming page](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) on the other had IS a reference page for this interface. So IMHO, as the reference page, it SHOULD explicitly list all entries. In addition each entry that has a subpage (like [`fetchStart`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/fetchStart) to pick one example) has the subpage under the `PerformanceResourceTiming` page in the hierarchy. Giving even more reason for the parent page to list these all.

So will update this PR to trim down the [Resource Timing page](https://developer.mozilla.org//en-US/docs/Web/API/Performance_API/Resource_timing). Let me know if you still disagree.

wbamberg:
I'm happy with your reasoning :).

tunetheweb:
OK updated now. Went through each section and moved most of the details to only be in one or the other. There's usually a stripped down version of that section in the doc with a link.

I've also updated the PR title and initial comment description.

PTAL.

tunetheweb:
Gentle ping @wbamberg, whenever you get a chance.

wbamberg:
So sorry, I thought I had looked at this one again. Give me a minute.

files/en-us/web/api/performanceresourcetiming/index.md
```diff
@@ -114,41 +115,26 @@ Additionally, this interface exposes the following properties containing more in
...
-
-Example using {{domxref("Performance.getEntriesByType()")}}, which only shows `resource` performance entries present in the browser's performance timeline at the time you call this method:
-
-```js
-const resources = performance.getEntriesByType("resource");
-resources.forEach((entry) => {
-  console.log(entry);
-});
-```
+For examples, see the [examples section of Resource Timing](/en-US/docs/Web/API/Performance_API/Resource_timing#examples).
```
wbamberg: Examples should be in this page (the reference page for the feature) not the guide.

files/en-us/web/api/performanceresourcetiming/index.md
```diff
@@ -114,41 +115,26 @@ Additionally, this interface exposes the following properties containing more in
...
 
+The properties which are returned as `0` by default when loading a resource from an origin other than the one of the web page itself: `redirectStart`, `redirectEnd`, `domainLookupStart`, `domainLookupEnd`, `connectStart`, `connectEnd`, `secureConnectionStart`, `requestStart`, and `responseStart`.
+
 For example, to allow `https://developer.mozilla.org` to see resource timing information, the cross-origin resource should send:
 
 ```http
 Timing-Allow-Origin: https://developer.mozilla.org
 ```
 
+## Managing resource buffer sizes
```
wbamberg: This should not be an H2 after "Examples". It should be an H3 under the existing "Description" section.

files/en-us/web/api/performance_api/resource_timing/index.md
```diff
@@ -12,26 +12,16 @@ Every resource on a document will be represented by a {{domxref("PerformanceReso
 
 For each `PerformanceResourceTiming` entry, a _resource loading timeline_ will be recorded, with {{domxref("DOMHighResTimeStamp","high-resolution timestamps", "", 1)}} for network events such as redirect start and end times, DNS lookup start and end times, request start, response start and end times, and so on. Besides the timestamps, other properties that provide information about the resource are included as well, such the size of the fetched resource, or the type of resource that initiated the fetch.
 
+[Typical resource timing metrics](/en-US/docs/Web/API/PerformanceResourceTiming#typical_resource_timing_metrics) this API is used for are detailed in the {{domxref("PerformanceResourceTiming")}}.
```
wbamberg:
```suggestion
See [Typical resource timing metrics](/en-US/docs/Web/API/PerformanceResourceTiming#typical_resource_timing_metrics) in the reference page for the {{domxref("PerformanceResourceTiming")}} interface.
```

(I think it is odd for the sentence to link to the info and _then_ say it's a link, and "the `PerformanceResourceTiming`" isn't really grammatical)

files/en-us/web/api/performance_api/resource_timing/index.md
```diff
@@ -12,26 +12,16 @@ Every resource on a document will be represented by a {{domxref("PerformanceReso
...
-11. {{domxref('PerformanceResourceTiming.requestStart','requestStart')}}: timestamp of the time immediately before the browser starts requesting the resource from the server, cache, or local resource.
-12. {{domxref('PerformanceResourceTiming.responseStart','responseStart')}}: timestamp immediately after the browser receives the first byte of the response from the server, cache, or local resource.
-13. {{domxref('PerformanceResourceTiming.responseEnd','responseEnd')}}: timestamp immediately after the browser receives the last byte of the resource or immediately before the transport connection is closed, whichever comes first.
+An application can get timestamps for the various stages used to load a resource. For example the {{domxref('PerformanceEntry.startTime','startTime')}}, DNS timestamps, connection set up times and then various resource download times.
+
+The [timestamps](/en-US/docs/Web/API/PerformanceResourceTiming#timestamps) this API provides are detailed in the {{domxref("PerformanceResourceTiming")}}.
```
wbamberg:
```suggestion
See [timestamps](/en-US/docs/Web/API/PerformanceResourceTiming#timestamps) in the reference page for the {{domxref("PerformanceResourceTiming")}} interface.
```

files/en-us/web/api/performance_api/resource_timing/index.md
```diff
@@ -43,49 +33,45 @@ The {{domxref('PerformanceResourceTiming.encodedBodySize','encodedBodySize')}} p
...
+
+Many of the resource timing properties are restricted to return `0` or an empty string when the resource is a cross-origin request. To expose cross-origin timing information, the {{HTTPHeader("Timing-Allow-Origin")}} HTTP response header needs to be set.
 
 For example, to allow `https://developer.mozilla.org` to see resource timing information, the cross-origin resource should send:
 
 ```http
 Timing-Allow-Origin: https://developer.mozilla.org
 ```
 
+For more information on the fields affected, see the [relevant section of PerformanceResourceTiming](/en-US/docs/Web/API/PerformanceResourceTiming#cross-origin_timing_information).
```
wbamberg:
```suggestion
For more information on the fields affected, see [Cross-origin timing information](/en-US/docs/Web/API/PerformanceResourceTiming#cross-origin_timing_information) in the reference page for the {{domxref("PerformanceResourceTiming")}} interface.
```

files/en-us/web/api/performance_api/resource_timing/index.md
```diff
@@ -43,49 +33,45 @@ The {{domxref('PerformanceResourceTiming.encodedBodySize','encodedBodySize')}} p
...
 
-When {{Glossary("CORS")}} is in effect, many of the timing properties' values are returned as zero unless the server's access policy permits these values to be shared. This requires the server providing the resource to send the {{httpheader("Timing-Allow-Origin")}} HTTP response header with a value specifying the origin or origins which are allowed to get the restricted timestamp values.
+```js
+const resources = performance.getEntriesByType("resource");
+resources.forEach((entry) => {
+  console.log(entry);
+});
+```
+
+## Security requirements
```
wbamberg: It would be better not to have this heading, and promote "Cross-origin timing information" to H2. "Security requirements" sections as quasi-formal sections for reference pages.

wbamberg: But also this section just feels redundant given the section in the ref page.

files/en-us/web/api/performanceresourcetiming/index.md
```diff
@@ -114,41 +115,26 @@ Additionally, this interface exposes the following properties containing more in
...
+
+By default only 250 resource timing entries are buffered. For more information see the [resource buffer sizes section of Resource Timing](/en-US/docs/Web/API/Performance_API/Resource_timing#managing_resource_buffer_sizes).
```
wbamberg:
```suggestion
By default only 250 resource timing entries are buffered. For more information see the [resource buffer sizes](/en-US/docs/Web/API/Performance_API/Resource_timing#managing_resource_buffer_sizes) of the Resource Timing guide.
```

files/en-us/web/api/performanceresourcetiming/index.md
tunetheweb: Moved to here

files/en-us/web/api/performanceresourcetiming/index.md
tunetheweb: Done

files/en-us/web/api/performance_api/resource_timing/index.md
tunetheweb: Moved it to a H2 and shorted it down to mostly just a link to the ref page.

tunetheweb:
Thanks for the review @wbamberg ! Think I confused myself cause both are under the /web/api path rather than a Guide so wasn't as obvious what was API version guide.

Anyway should all be sorted no so PTAL.

files/en-us/web/api/performance_api/resource_timing/index.md
```diff
@@ -12,26 +12,16 @@ Every resource on a document will be represented by a {{domxref("PerformanceReso
 
 For each `PerformanceResourceTiming` entry, a _resource loading timeline_ will be recorded, with {{domxref("DOMHighResTimeStamp","high-resolution timestamps", "", 1)}} for network events such as redirect start and end times, DNS lookup start and end times, request start, response start and end times, and so on. Besides the timestamps, other properties that provide information about the resource are included as well, such the size of the fetched resource, or the type of resource that initiated the fetch.
 
+See [Typical resource timing metrics](/en-US/docs/Web/API/PerformanceResourceTiming#typical_resource_timing_metrics) in the reference page for the {{domxref("PerformanceResourceTiming")}} interface
```
wbamberg:
```suggestion
See [Typical resource timing metrics](/en-US/docs/Web/API/PerformanceResourceTiming#typical_resource_timing_metrics) in the reference page for the {{domxref("PerformanceResourceTiming")}} interface.
```

wbamberg (approved):
👍 thank you!

wbamberg:
> Thanks for the review @wbamberg ! Think I confused myself cause both are under the /web/api path rather than a Guide so wasn't as obvious what was API version guide.

In Web/API, with very few exceptions, there are 2 sorts of pages at the top level:
- API overview pages, like https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API, that document some abstract collection of web platform features that usually but not always maps to a spec. 
- Interface pages, like https://developer.mozilla.org/en-US/docs/Web/API/Request, that's a concrete WP feature (i.e. a thing that code can directly interact with).

Any pages under API overview pages should always be guide pages, and that's the only place under Web/API that guide pages are allowed. Pages under interface pages are always constructor, method, property, or event pages.

But you can always tell what kind of page a page is by looking at the `page-type` front matter key: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types/Page_type_key.

PR #21946, opened 2022-10-30
Update Transfer-Encoding to hex numbers

Schniz:
### Description

The current chunked encoded example is streaming strings that their length is lower than 10 (Mozilla, Developer, Network), therefore, it is easy to miss that the chunked length should be in hexadecimal form.

By merging the 2nd and 3rd line into `Developer Network`, we reach length of 17 which is 11. This marks it much clearer that it is hex.

### Motivation

Making the usage of hex numbers stand out.

PR #41385, opened 2025-10-03
Fix Transfer-Encoding example with correct length

TheCodingDutchman:
### Description
Fixes the example where the length of the "Developer Network" chunk was incorrectly set to 11 instead of 17.

### Motivation
I was using the example to test my own HTTP server, only to realize that the chunk length was incorrect in the example.

### Additional details
None

None


wbamberg:
See https://github.com/mdn/content/pull/21946, it looks like these are supposed to be hex, so 11===17. Might be even clearer to use a chunk in the example whose hex length included non-decimal characters, like `1A`.

PR #41396, opened 2025-10-05
Improve description of Chrome's active tab in collapsed group behavior

l10nelw:
### Description

More concise and yet also more thorough.


macktony497-dev:
stop

PR #41402, opened 2025-10-06
Update index.md - Finding A Job

ghost:
"Keep passionate..." does not read smoothly.
Use "stay passionate," instead.

Screenshot 2025-10-05 234858

[https://developer.mozilla.org/en-US/docs/Learn_web_development/Getting_started/Soft_skills/Finding_a_job#try_try_and_try_again](url)


chrisdavidmills (approved):
Yup, this reads more smoothly. Thanks for the update!

PR #41403, opened 2025-10-06
setInterval(code) - update with TrustedScript

hamishwillee:
This updates the TrustedScript documentation for [`Window/setInterval()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/setInterval) to include information about using it with trusted types.

A lot of this was to update the doc to current template - such as moving Usage notes under description and so on.
I also thought it worthwhile to purge mention of getting things to work with IE, and focus on how you get the "this" right rather than not being able to get it right on legacy JavaScript.

The security considerations mostly mention CSP since generally the `code` option is disabled by default. 

I will look at `setTimeout()` after this, since the changes would be mostly the same.

Project tracking in #37518

### DIFF
--- a/files/en-us/web/api/window/setinterval/index.md
+++ b/files/en-us/web/api/window/setinterval/index.md
@@ -160,7 +160,7 @@ const id = setInterval(untrustedCode, 1000);
 
 Websites with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) will prevent such code running by default; if you need to use the method with `code` then you will first need to allow the [`unsafe-eval`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#unsafe-eval) in your CSP [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src).
 
-If you must allow the scripts to run you can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+If you must allow the scripts to run you can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
 This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) or reject the text before it is injected.
 
 The behavior of the transformation function will depend on the specific use case that requires a user provided script.
--- a/files/en-us/web/api/window/setinterval/index.md
+++ b/files/en-us/web/api/window/setinterval/index.md
@@ -8,6 +8,13 @@ browser-compat: api.setInterval
 
 {{APIRef("HTML DOM")}}
 
+> [!WARNING]
+> This method dynamically executes input passed to the `code` parameter as JavaScript.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
+>
+> You can mitigate this risk by always assigning {{domxref("TrustedScript")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](#security_considerations) for more information.
+
 The **`setInterval()`** method of the {{domxref("Window")}} interface repeatedly calls a function or executes a code snippet, with a fixed time delay between each call.
 
 ## Syntax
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/window/setinterval/index.md
```diff
@@ -26,29 +33,140 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
+If you must allow the scripts to run you can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) or reject the text before it is injected.
+
+The behavior of the transformation function will depend on the specific use case that requires a user provided script.
+If possible you should lock the allowed scripts to exactly the code that you trust to run.
+If that is not possible, you might allow or block the use of certain functions within the provided string.
+
 ## Examples
 
 ### Example 1: Basic syntax
```
hamishwillee, line 171: Note, I didn't add a trusted type example because to my mind it isn't the main use case with this. I can add one if you think it useful.

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -26,29 +33,140 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
+For example, the following example assumes the `scriptElement` is an executable `<script>` element, and that `untrustedCode` was provided by a user:
+
+```js
+const untrustedCode = "alert('Potentially evil code!');";
+const id = setInterval(untrustedCode, 1000);
+```
+
+Websites with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) will prevent such code running by default; if you need to use the method with `code` then you will first need to allow the [`unsafe-eval`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#unsafe-eval) in your CSP [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src).
+
+If you must allow the scripts to run you can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
```
hamishwillee:
```suggestion
If you must allow the scripts to run you can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
```

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -8,6 +8,13 @@ browser-compat: api.setInterval
 
 {{APIRef("HTML DOM")}}
 
+> [!WARNING]
+> This method dynamically executes input passed to the `code` parameter as JavaScript.
```
wbamberg:
```suggestion
> When the `code` parameter is used, this method dynamically executes its value as JavaScript.
```

Just because otherwise it may not be obvious that this isn't always the case, with this method.

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -26,29 +33,140 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
 
 - `func`
-  - : A {{jsxref("function")}} to be executed every `delay` milliseconds. The first execution happens after `delay` milliseconds.
+  - : A {{jsxref("function")}} to be executed every `delay` milliseconds.
+    The first execution happens after `delay` milliseconds.
 - `code`
-  - : An optional syntax allows you to include a string instead of a function, which is compiled and executed every `delay` milliseconds.
-    This syntax is _not recommended_ for the same reasons that make using {{jsxref("Global_Objects/eval", "eval()")}} a security risk.
+  - : A {{domxref("TrustedScript")}} or a string of arbitrary code that is compiled and executed every `delay` milliseconds.
+    This can be used instead of passing a function, but is _highly discommended_ for the same reasons that make using {{jsxref("Global_Objects/eval", "eval()")}} a security risk.
```
wbamberg:
```suggestion
    This can be used instead of passing a function, but is _strongly discouraged_ for the same reasons that make using {{jsxref("Global_Objects/eval", "eval()")}} a security risk.
```

This is the only use of "discommended" in all of MDN, so consider the more commonly used "discouraged".

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -26,29 +33,140 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
 - `delay` {{optional_inline}}
-  - : The time, in milliseconds (thousandths of a second), the timer should delay in between executions of the specified function or code. Defaults to 0 if not specified.
+  - : The time, in milliseconds (thousandths of a second), the timer should delay in between executions of the specified function or code.
```
wbamberg: idk if we have to say what a millisecond is.

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -26,29 +33,140 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
+    Defaults to 0 if not specified.
     See [Delay restrictions](#delay_restrictions) below for details on the permitted range of `delay` values.
 - `arg1`, …, `argN` {{optional_inline}}
   - : Additional arguments which are passed through to the function specified by _func_ once the timer expires.
 
 ### Return value
 
-The `setInterval()` method returns a positive integer (typically within the range of 1 to 2,147,483,647) that uniquely identifies the interval timer created by the call. This identifier, often referred to as an "interval ID", can be passed to {{domxref("Window.clearInterval", "clearInterval()")}} to stop the repeated execution of the specified function.
+A positive integer (typically within the range of 1 to 2,147,483,647) that uniquely identifies the interval timer created by the call.
```
wbamberg, line 50: I think it is worth keeping in the bit about being able to pass the value into `clearInterval()` even though you do say it later. The details about uniqueness I agree to defer to the description.

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -26,29 +33,140 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
+setTimeout(() => myArray.myMethod(1), 1500); // Alert "one" after 1.5 seconds
+setTimeout(() => myArray.myMethod(2), 3000); // Alert "one" after 3 seconds
+```
+
+You might also use the [`Function.prototype.bind()`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) method, which lets you specify the value that should be used as `this` for all calls to a given function.
+That lets you bypass problems where it's unclear what `this` will be, depending on the context from which your function was called.
+
+### Security considerations
+
+When used with the function parameter, the method is no more or less dangerous than any other method that can trigger code execution.
```
wbamberg:
```suggestion
When used with the `function` parameter, the method is no more or less dangerous than any other method that can trigger code execution.
```

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -26,29 +33,140 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
+While specific mechanisms are not required to address this case, you should always consider deploying a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) to harden your site.
```
wbamberg: I don't understand how a CSP is relevant to the `function` version.

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -26,29 +33,140 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
+
+The method can also execute arbitrary input passed in the `code` parameter.
+If the input is a potentially unsafe string provided by a user, this is a possible vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
+For example, the following example assumes the `scriptElement` is an executable `<script>` element, and that `untrustedCode` was provided by a user:
+
+```js
```
wbamberg: example-bad?

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -26,29 +33,140 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
+const untrustedCode = "alert('Potentially evil code!');";
+const id = setInterval(untrustedCode, 1000);
+```
+
+Websites with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) will prevent such code running by default; if you need to use the method with `code` then you will first need to allow the [`unsafe-eval`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#unsafe-eval) in your CSP [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src).
+
+If you must allow the scripts to run you can mitigate these issues by always assigning {{domxref("TrustedScript")}} objects instead of strings, and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) or reject the text before it is injected.
```
wbamberg: As we've said recently, the link to sanitization is not a good one for scripts. Not sure what we should say. Since we follow up by giving examples, maybe we could just say:

```suggestion
This ensures that the input is passed through a transformation function.
```

files/en-us/web/api/window/setinterval/index.md
hamishwillee: We don't :-)
```suggestion
  - : The delay time between executions of the specified function or code, in milliseconds.
```

files/en-us/web/api/window/setinterval/index.md
hamishwillee:
```suggestion
A positive integer (typically within the range of 1 to 2,147,483,647) that uniquely identifies the interval timer created by the call.

This identifier, often referred to as an "interval ID", can be passed to {{domxref("Window.clearInterval", "clearInterval()")}} to stop the repeated execution of the specified function.
```

files/en-us/web/api/window/setinterval/index.md
hamishwillee: I guess so.
````suggestion
```js example-bad
````

File: files/en-us/web/api/window/setinterval/index.md (hunk 10 above)
Comment (hamishwillee) [reply files/en-us/web/api/window/setinterval/index.md]: I have nothing better to offer.

File: files/en-us/web/api/window/setinterval/index.md [hunk 11]
```diff
@@ -26,29 +33,142 @@ setInterval(func, delay, arg1, arg2, /* …, */ argN)
...
+
+You might also use the [`Function.prototype.bind()`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) method, which lets you specify the value that should be used as `this` for all calls to a given function.
+That lets you bypass problems where it's unclear what `this` will be, depending on the context from which your function was called.
+
+### Security considerations
+
+When used with the `function` parameter, the method is no more or less dangerous than any other method that can trigger code execution.
+While specific mechanisms are not required to address this case, you should always consider deploying a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) to harden your site.
+
+The method can also execute arbitrary input passed in the `code` parameter.
```
hamishwillee:
```suggestion
The method can be used to execute arbitrary input passed in the `code` parameter.
```

files/en-us/web/api/window/setinterval/index.md
hamishwillee: Makes no sense. Don't know what I was thinking. Deleted in https://github.com/mdn/content/pull/41403/commits/142857e68acdc870cec74ba44bc9613c2458e67d

hamishwillee:
Thanks @wbamberg - I have effectively accepted all.

wbamberg (approved):
👍 thank you!

PR #41405, opened 2025-10-06
Add live final version to styling tables article

chrisdavidmills:
### Description

I decided to update the [Styling tables](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Tables) article to make a few small styling updates and add a live version of the finished design for readers to compare their work against.

### DIFF
--- a/files/en-us/learn_web_development/core/styling_basics/tables/index.md
+++ b/files/en-us/learn_web_development/core/styling_basics/tables/index.md
@@ -38,7 +38,7 @@ Styling an HTML table isn't the most glamorous job in the world, but sometimes w
 
 Let's start by looking at a typical HTML table. Well, I say typical — most HTML table examples are about shoes, or the weather, or employees; we decided to make things more interesting by making it about famous punk bands from the UK. The markup looks like so:
 
-```html live-sample___unstyled live-sample___styled
+```html live-sample___unstyled live-sample___punk-style live-sample___best-practice-style
 <table>
   <caption>
     A summary of the UK's most famous punk bands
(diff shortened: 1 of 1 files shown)


files/en-us/learn_web_development/core/styling_basics/tables/index.md
hamishwillee: You changed this above.

I think this is the only "have to" fix in the doc.

files/en-us/learn_web_development/core/styling_basics/tables/index.md
```diff
@@ -222,7 +222,7 @@ caption {
 }
 ```
 
-{{embedlivesample("styled", "", "500")}}
+{{embedlivesample("punk-style", "", "500")}}
 
 However, this design is rather garish. In this article, we'll get you to mark it up using some best practices for table design — as outlined in [Web Typography: designing tables to be read not looked at](https://alistapart.com/article/web-typography-tables/).
```
hamishwillee, line 227: FWIW I'm not really sure why this example exists at all, except perhaps as a segue to the how to design article.

files/en-us/learn_web_development/core/styling_basics/tables/index.md
hamishwillee: This is not a challenge or test, but in those you have somewhat standardized the explanation of the workflow where they copy-paste the example. If you want to do it this way then fine but noting that it is a different approach.

chrisdavidmills: Good call; I've updated that text to fit with the new code. I've removed the bit about mobile responsiveness, as it didn't seem that relevant to the main thrust of the article, and it made the live example look crappy.

files/en-us/learn_web_development/core/styling_basics/tables/index.md
chrisdavidmills: That was the design I originally went with on this article, but I didn't want to throw it away completely. I decided to keep it as an example to show that you _can_ do quite a lot when styling a table if you want to, and then show the reader how to build up a slightly more normal "best practice" example.

files/en-us/learn_web_development/core/styling_basics/tables/index.md
chrisdavidmills: There isn't much harm in a little bit of variation, but I would prefer to keep the source code all available on the page rather than linking out to the learning-area repo, which we are trying to reduce our reliance on.

To fix this, I've replaced the external repo link with a relative link to the section earlier on that lists the source HTML.

hamishwillee: Works for me. Thanks!

hamishwillee (approved):
LGTM!

PR #37917, opened 2025-02-02
Expand trusted types guidance

wbamberg:
I had thought to defer updates to the trusted types docs until it ships cross-browser, but given conversations like https://github.com/w3c-cg/swag/pull/9#discussion_r1931309886 and our general desire to talk about this API, I thought it was worth expanding on things a little now.

We should think of this PR as a stop-gap though: I still think we should revisit these docs, write a proper guide, and refresh all the API pages, when the API ships in Firefox or Safari.

This PR expands the [Trusted Types API overview page](https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API) in two ways:
- explain better about enforcement using a CSP
- explain about polyfills, since people can get the benefit of this API cross-browser now

I also updated the rest of the section to be clearer about what the API actually does and how it connects with XSS - I thought things like "lock down the insecure parts of the DOM API" weren't very meaningful.

I hope I understood the bit about polyfills!

@aaronshim, @lukewarlow

PR #41413, opened 2025-10-07
HTMLScriptElement.src - update for TrustedScriptURL

hamishwillee:
This updates the TrustedScriptURL documentation for [`HTMLScriptElement/src`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement/src) to include information about using it with trusted types.

I don't know why/when it would ever make sense to accept an  untrusted URL, and if you did, to use trusted types. I've presented this primarily as "Use CSP to restrict the script options but if you have to take and untrusted script then you can use TT".

Project tracking in #37518

Also part of #41507

### DIFF
--- a/files/en-us/web/api/htmlscriptelement/src/index.md
+++ b/files/en-us/web/api/htmlscriptelement/src/index.md
@@ -10,7 +10,7 @@ browser-compat: api.HTMLScriptElement.src
 
 > [!WARNING]
 > This property represents the URI of an external script loaded into the script element, which may be executable depending on the script {{domxref("HTMLScriptElement/type","type")}}.
-> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) and other types of [website attacks](/en-US/docs/Web/Security/Attacks).
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
 >
 > You can mitigate this risk by having a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that restricts the locations from which scripts can be loaded, and by always assigning {{domxref("TrustedScriptURL")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
 > See [Security considerations](#security_considerations) for more information.
@@ -40,7 +40,7 @@ If set, scripts provided via the text properties {{domxref("HTMLScriptElement.te
 
 The `src` property is used to load and run external scripts.
 The fetched script is run in the context of the current page, and can hence do anything that your own website code can do (even if the URL is not same-origin with your site).
-If the input is provided by a user, this is a possible vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
+If the input is provided by a user, this is a possible vector for [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
 Even if the resource is trusted by your website, it may still be compromised in a supply chain attack.
 
 A website should control what scripts that are allowed to run using a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) with the [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src) directive (or a fallback defined in [`default-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src)).
--- a/files/en-us/web/api/htmlscriptelement/src/index.md
+++ b/files/en-us/web/api/htmlscriptelement/src/index.md
@@ -8,15 +8,102 @@ browser-compat: api.HTMLScriptElement.src
 
 {{APIRef("HTML DOM")}}
 
+> [!WARNING]
+> This property represents the URI of an external script loaded into the script element, which may be executable depending on the script {{domxref("HTMLScriptElement/type","type")}}.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) and other types of [website attacks](/en-US/docs/Web/Security/Attacks).
+>
+> You can mitigate this risk by having a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that restricts the locations from which scripts can be loaded, and by always assigning {{domxref("TrustedScriptURL")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](#security_considerations) for more information.
+
 The **`src`** property of the {{domxref("HTMLScriptElement")}} interface is a string representing the URL of an external script; this can be used as an alternative to embedding a script directly within a document.
 
-It reflects the `src` attribute of the {{HTMLElement("script")}} element.
+It reflects the [`src`](/en-US/docs/Web/HTML/Reference/Elements/script#src) attribute of the {{HTMLElement("script")}} element.
 
 ## Value
 
-A string.
+Getting the property returns a string containing the element's script URI.
+
+Setting the property accepts either a {{domxref("TrustedScriptURL")}} object or a string.
+
+### Exceptions
+
+- `TypeError`
+  - : Thrown if the property is set with a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
+    This is also thrown if the fetched URL cannot be successfully parsed as its indicated type, such as a module or importmap.
+
+## Description
+
+The **`src`** property represents the URL of an external script.
+If set, scripts provided via the text properties {{domxref("HTMLScriptElement.text","text")}}, {{domxref("HTMLScriptElement.textContent","textContent")}}, or {{domxref("HTMLScriptElement.textContent","innerText")}}, are ignored.
+
+### Security considerations
+
+The `src` property is used to load and run external scripts.
+The fetched script is run in the context of the current page, and can hence do anything that your own website code can do (even if the URL is not same-origin with your site).
+If the input is provided by a user, this is a possible vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
+Even if the resource is trusted by your website, it may still be compromised in a supply chain attack.
+
+A website should control what scripts that are allowed to run using a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) with the [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src) directive (or a fallback defined in [`default-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src)).
+This can restrict scripts to those from the current origin, or a specific set of origins, or even particular files.
+To mitigate against supply chain attacks you can even restrict files to those that have a specific hash or nonce: the browser won't load a file that does not match the condition even if it is sent by a compromised server.
+
+It is extremely risky to accept and execute arbitrary URLs from untrusted origins.
+If you find a situation where this is necessary, you should always assign {{domxref("TrustedScriptURL")}} objects instead of strings, and [enforce trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) or reject the URL before it is injected.
+
+The behavior of the transformation function will depend on the specific use case that requires a user provided script.
 
 ## Examples
+
+### Using TrustedScriptURL
+
+To mitigate the risk of XSS, we should always assign `TrustedScriptURL` instances to the `src` property.
+
+Trusted types are not yet supported on all browsers, so first we define the [trusted types tinyfill](/en-US/docs/Web/API/Trusted_Types_API#trusted_types_tinyfill).
+This acts as a transparent replacement for the trusted types JavaScript API:
+
+```js
+if (typeof trustedTypes === "undefined")
+  trustedTypes = { createPolicy: (n, rules) => rules };
+```
+
+Next we create a {{domxref("TrustedTypePolicy")}} that defines a {{domxref("TrustedTypePolicy/createScriptURL", "createScriptURL()")}} method for transforming input strings into {{domxref("TrustedScriptURL")}} instances.
+For the purpose of this example we'll just log the script URL:
+
+```js
+const policy = trustedTypes.createPolicy("script-url-policy", {
+  createScriptURL(input) {
+    console.log(`Log TT script-url-policy: ${input}`);
+    return input; // allow the script
+  },
+});
+```
+
+Next we'll create the script element to which we will assign the value and get a handle to the element.
+
+```html
+<script id="el" type="text/javascript"></script>
+```
+
+```js
+// Get the script element we're injecting the code into
+const el = document.getElementById("el");
+```
+
+Then we use the `policy` object to create a `trustedScript` object from the potentially unsafe input string, and assign the result to the element:
+
+```js
+// The potentially malicious string
+const untrustedScript = "https://evil.example.com/naughty.js";
+
+// Create a TrustedScriptURL instance using the policy
+const trustedScriptURL = policy.createScriptURL(untrustedScript);
+
+// Inject the TrustedScriptURL (which contains a trusted URL)
+el.src = trustedScriptURL;
+```
+
+## Reading the `src` element
 
 Assume the code is running on a website whose URL is `https://example.com`.
 
--- a/files/en-us/web/api/htmlscriptelement/src/index.md
+++ b/files/en-us/web/api/htmlscriptelement/src/index.md
@@ -103,7 +103,7 @@ const trustedScriptURL = policy.createScriptURL(untrustedScript);
 el.src = trustedScriptURL;
 ```
 
-## Reading the `src` element
+### Reading the `src` element
 
 Assume the code is running on a website whose URL is `https://example.com`.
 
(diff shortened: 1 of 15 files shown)


files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,102 @@ browser-compat: api.HTMLScriptElement.src
...
 
-A string.
+Getting the property returns a string containing the element's script URI.
+
+Setting the property accepts either a {{domxref("TrustedScriptURL")}} object or a string.
+
+### Exceptions
+
+- `TypeError`
+  - : Thrown if the property is set with a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
```
hamishwillee, line 31:
```suggestion
  - : Thrown if the property is set with a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
    This is also thrown if the fetched URL cannot be successfully parsed as its indicated type, such as a module or importmap.
  
```

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,104 @@ browser-compat: api.HTMLScriptElement.src
...
+    This is also thrown if the fetched URL cannot be successfully parsed as its indicated type, such as a module or importmap.
```
hamishwillee, line 32: This isn't quite right, but I'm not sure how to be precise.
I believe it is a NetworkError on fetching. TypeError is thrown for an importmap during parsing (https://html.spec.whatwg.org/multipage/webappapis.html#import-map-authoring-requirements), and I think it is a Syntax error you get normally on parsing JavaScript.

Thoughts?

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,103 @@ browser-compat: api.HTMLScriptElement.src
 
 {{APIRef("HTML DOM")}}
 
+> [!WARNING]
+> This property represents the URI of an external script loaded into the script element, which may be executable depending on the script {{domxref("HTMLScriptElement/type","type")}}.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) and other types of [website attacks](/en-US/docs/Web/Security/Attacks).
```
wbamberg: I should have noticed this before but it's not "cross-site-scripting", it's "cross-site scripting".

Also, which other types of website attacks?

files/en-us/web/api/htmlscriptelement/src/index.md
wbamberg: Not sure I understand the question here.

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,103 @@ browser-compat: api.HTMLScriptElement.src
...
+## Description
+
+The **`src`** property represents the URL of an external script.
+If set, scripts provided via the text properties {{domxref("HTMLScriptElement.text","text")}}, {{domxref("HTMLScriptElement.textContent","textContent")}}, or {{domxref("HTMLScriptElement.textContent","innerText")}}, are ignored.
+
+### Security considerations
+
+The `src` property is used to load and run external scripts.
+The fetched script is run in the context of the current page, and can hence do anything that your own website code can do (even if the URL is not same-origin with your site).
+If the input is provided by a user, this is a possible vector for [Cross-site-scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
```
wbamberg:
```suggestion
If the input is provided by a user, this is a possible vector for [Cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
```

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,103 @@ browser-compat: api.HTMLScriptElement.src
...
+Even if the resource is trusted by your website, it may still be compromised in a supply chain attack.
```
wbamberg: At first I thought that we should not talk about supply chain attacks at all in this page. After all,  if "the resource" (which I guess means basically the URL) is trusted, then obviously checking the value of `src` will never be a fix. And checking the value is the only thing we can do when trying to secure assignment to `src`.

But maybe we should, but disentangle it from the TT/CSP stuff, which is about XSS. So maybe first cover XSS, and CSP/TT as mitigations, then say that supply chain attacks are also a risk, and you should use SRI to mitigate that?

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,103 @@ browser-compat: api.HTMLScriptElement.src
...
+
+A website should control what scripts that are allowed to run using a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) with the [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src) directive (or a fallback defined in [`default-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src)).
+This can restrict scripts to those from the current origin, or a specific set of origins, or even particular files.
+To mitigate against supply chain attacks you can even restrict files to those that have a specific hash or nonce: the browser won't load a file that does not match the condition even if it is sent by a compromised server.
```
wbamberg: So as said before, a nonce doesn't protect against supply chain attacks. Recommendation here is to use SRI (https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/Supply_chain_attacks#using_subresource_integrity) which is not part of CSP.

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,103 @@ browser-compat: api.HTMLScriptElement.src
...
+
+It is extremely risky to accept and execute arbitrary URLs from untrusted origins.
+If you find a situation where this is necessary, you should always assign {{domxref("TrustedScriptURL")}} objects instead of strings, and [enforce trusted type](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive.
+This ensures that the input is passed through a transformation function, which has the chance to [sanitize](/en-US/docs/Web/Security/Attacks/XSS#sanitization) or reject the URL before it is injected.
+
+The behavior of the transformation function will depend on the specific use case that requires a user provided script.
```
wbamberg: I don't think [sanitize](https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/XSS#sanitization) is a very helpful link for script URLs (as opposed to HTML). We've talked before, I think, about what the TT transformation is supposed to do with scripts and script URLs, and the best we came up with, I think, was "block them entirely" or "allow only a predefined set". That's also what we say in https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage:

> For JavaScript and JavaScript URL sinks, the policy may turn off the sinks entirely or allow certain predefined inputs (for example, specific URLs).

So I suppose it might be not so much "a situation where this is necessary" as "a situation where this is not ever what you want to do, and you want to be sure that nowhere in your codebase is it happening".

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,103 @@ browser-compat: api.HTMLScriptElement.src
...
+```
+
+Next we create a {{domxref("TrustedTypePolicy")}} that defines a {{domxref("TrustedTypePolicy/createScriptURL", "createScriptURL()")}} method for transforming input strings into {{domxref("TrustedScriptURL")}} instances.
+For the purpose of this example we'll just log the script URL:
+
+```js
+const policy = trustedTypes.createPolicy("script-url-policy", {
+  createScriptURL(input) {
+    console.log(`Log TT script-url-policy: ${input}`);
+    return input; // allow the script
```
wbamberg: So relatedly then, maybe the policy should block them or only allow a predefined set of values?

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,103 @@ browser-compat: api.HTMLScriptElement.src
...
+
+```js
+// The potentially malicious string
+const untrustedScript = "https://evil.example.com/naughty.js";
+
+// Create a TrustedScriptURL instance using the policy
+const trustedScriptURL = policy.createScriptURL(untrustedScript);
+
+// Inject the TrustedScriptURL (which contains a trusted URL)
+el.src = trustedScriptURL;
```
wbamberg, line 111: I mean it seems unfortunate that this TT is going to allow that naughty JS.

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,103 @@ browser-compat: api.HTMLScriptElement.src
...
+```
+
+## Reading the `src` element
```
wbamberg:
```suggestion
### Reading the `src` element
```

files/en-us/web/api/htmlscriptelement/src/index.md
hamishwillee: 1. Fixed cross-site-scripting globally by search and replace in https://github.com/mdn/content/pull/41413/commits/e267f568feabfb145b3066b52f2787ace95675c0
2. I removed "the other types of website attacks" here. I think I was using weasel words because I wasn't sure.

files/en-us/web/api/htmlscriptelement/src/index.md
hamishwillee: Sorry. I think the text is correct. There are however other errors you might get such as NetworkError and SyntaxError.
I am not going to capture those here on the grounds that this fix is "better than it was" and I have lots to do in this release cycle. Resolving.

files/en-us/web/api/htmlscriptelement/src/index.md
hamishwillee: @wbamberg EDITED (a number of times)

I removed "sanitized" and replaced with "This ensures that the input is passed through a transformation function, which has the chance to reject or modify the URL before it is injected.".
The _modify_ is to catch cases where you might want to allow but redirect the URL or perhaps strip off URL parameters or whatever - I don't really have a practical case.

> For JavaScript and JavaScript URL sinks, the policy may turn off the sinks entirely or allow certain predefined inputs (for example, specific URLs).

The problem here is that if you want to block scripts or allow only specified scripts to execute then you would use  CSP that defines a `script-src` - much is more effective and less error prone than using a transformation.

I don't see any "real" reason why you would **ever** use trusted types, other than you're already enforcing them for other things so you need to.

I have modified the section to say that - so this is not so much "use TT to block stuff" as "use TT because otherwise your already controlled resource will be blocked".

Does that make sense?

files/en-us/web/api/htmlscriptelement/src/index.md
hamishwillee: Yes, removed the nonce stuff and moved the supply chain stuff to the end. It is just links - no deep flavour.

files/en-us/web/api/htmlscriptelement/src/index.md
wbamberg: > @wbamberg
> 
> I removed "sanitized" and replaced with "This ensures that the input is passed through a transformation function, which has the chance to reject or modify the URL before it is injected.". The _modify_ is to catch cases where you might want to allow but redirect the URL or perhaps strip off URL parameters or whatever - I don't really have a practical case.
> 
> > For JavaScript and JavaScript URL sinks, the policy may turn off the sinks entirely or allow certain predefined inputs (for example, specific URLs).
> 
> The problem here is that if you want to block scripts or allow only specified scripts to execute then you would use CSP that defines a `script-src` - much is more effective and less error prone than using a transformation.
> 

I would like to know what @aaronshim thinks about this. Because we recommend that developers use a strict CSP, that uses nonces or hashes (depending on whether you can generate responses dynamically or not) rather than locations. But if you use nonces, then setting `src` is dangerous, because the script will be allowed to run if it also has the correct nonce (which it probably will, if the `<script>` element we are modifying was properly served in the first place). So at least in that case I could see a use for a TT policy that bans assignment to `src`.

I mean, as a developer, say you don't ever want to allow assignment to `script.src`. Is that alone a strong enough reason to use a location-based CSP (with all its known fragility) instead of a nonce-based one?

> I don't see any "real" reason why you would **ever** use trusted types, other than you're already enforcing them for other things so you need to (perhaps I should say that). Hence the "where it is necessary" words.
> 
> Does that make sense?

files/en-us/web/api/htmlscriptelement/src/index.md
hamishwillee: Yes, generally we should have an allowlist I think, or use a CSP:script-src and allow everything. As above, its all a bit unclear. I think though we now have at least a valid use case for TT here.

Anyway, updated example in https://github.com/mdn/content/pull/41413/commits/e4ecd7a4c903ca8f87f184b82b16b6c25d69688b

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,113 @@ browser-compat: api.HTMLScriptElement.src
...
+A website should control what scripts that are allowed to run using a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) with the [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src) directive (or a fallback defined in [`default-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src)).
+This can restrict scripts to those from the current origin, or a specific set of origins, or even particular files.
+
+If you're using this property and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) (using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive), you will need to always assign {{domxref("TrustedScriptURL")}} objects instead of strings.
+This ensures that the input is passed through a transformation function, which has the chance to reject or modify the URL before it is injected.
+If you're permitting some sources using `CSP: script-src`, the transformation function would use the sources as an allow-list.
```
wbamberg:
```suggestion
```

I would omit this, because not necessarily: you might be happy to allow some scripts in the markup you serve but not want to allow their dynamic assignment.

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,113 @@ browser-compat: api.HTMLScriptElement.src
...
+
+```js
+if (typeof trustedTypes === "undefined")
+  trustedTypes = { createPolicy: (n, rules) => rules };
+```
+
+Next we create a {{domxref("TrustedTypePolicy")}} that defines a {{domxref("TrustedTypePolicy/createScriptURL", "createScriptURL()")}} method for transforming input strings into {{domxref("TrustedScriptURL")}} instances.
+
+For the purpose of this example we'll assume that we want to allow a set URLs in the `scriptAllowList` array and log any other scripts.
+Note that if we were using a `CSP: script-src` we might just be able to pass through all URLs.
```
wbamberg:
```suggestion
```

...this is not good advice AFAICT. If we have `script-src` with a nonce, which is our recommendation, then it is surely dangerous to pass through all URLs.

files/en-us/web/api/htmlscriptelement/src/index.md
```diff
@@ -8,16 +8,113 @@ browser-compat: api.HTMLScriptElement.src
...
+This acts as a transparent replacement for the trusted types JavaScript API:
+
+```js
+if (typeof trustedTypes === "undefined")
+  trustedTypes = { createPolicy: (n, rules) => rules };
+```
+
+Next we create a {{domxref("TrustedTypePolicy")}} that defines a {{domxref("TrustedTypePolicy/createScriptURL", "createScriptURL()")}} method for transforming input strings into {{domxref("TrustedScriptURL")}} instances.
+
+For the purpose of this example we'll assume that we want to allow a set URLs in the `scriptAllowList` array and log any other scripts.
```
wbamberg:
```suggestion
For the purpose of this example we'll assume that we want to allow a predefined set of URLs in the `scriptAllowList` array and log any other scripts.
```

wbamberg (commented):
I made a couple of suggestions to downplay the stuff about CSP here.

files/en-us/web/api/htmlscriptelement/src/index.md
hamishwillee: Given the following CSP, HTML and JS the HTML fetch would be allowed but the API fetch would not - because it doesn't have the nonce. 

1. What I think you're saying is that to allow the API fetch you would have to relax the strict CSP with the nonce, and you shouldn't do that. 
   - Is that right?
   -  If so, yes, need comment from @aaronshim for the recommendation.

```http
Content-Security-Policy: script-src-elem https://some-cdn/file.js 'nonce-416d1177-4d12-4e3b-b7c9-f6c409789fb8';
```

```html
<script src="https://some-cdn/file.js" nonce=416d1177-4d12-4e3b-b7c9-f6c409789fb8>, 
```

```js
document.querySelector("script").src="https://some-cdn/file.js" 
```

Note that you still could use the `integrity` on the HTMLScriptElement, because that exists.

2. `CSP  script-src-elem` controls script fetch for both the HTML and JS API. It seems reasonable that you would want to allow via HTML but not via JS (as you noted). 
   - Using nonces would prevent the JS in this case 
   - With a less restrictive CSP that allows an address then either fetch would be possible. So if we don't support nonces then yes, we might want to block access via a TT policy.

wbamberg: > Given the following CSP, HTML and JS the HTML fetch would be allowed but the API fetch would not - because it doesn't have the nonce.
> 

It's a bit confusing, because your nonces don't match, but I assume they were intended to?

```http
Content-Security-Policy: script-src 'nonce-123456';
```

(But note that your example also includes the script location, which I would not expect to see in a strict CSP.)

```
<script src="https://some-cdn/file.js" nonce="123456">, 
```

...given that, I would expect that this:

```js
document.querySelector("script").src="https://evil.com/file.js" 
```

WILL work, because the script element still has the nonce attribute on it, and the `nonce` attribute is still present and correct. We've just set the `src` attribute.

I haven't tested this though.

But if this is true, it seems that if you have a nonce-based CSP, which we recommend, then you CAN'T say "it's OK to allow assignment to `src` because the CSP will protect you".

That's what I mean by:

>  But if you use nonces, then setting src is dangerous, because the script will be allowed to run if it also has the correct nonce (which it probably will, if the <script> element we are modifying was properly served in the first place). So at least in that case I could see a use for a TT policy that bans assignment to src.

hamishwillee: Ah, yes I meant matching nonces (updated), yes that sounds like a risk, thanks, an interesting use case.

wbamberg: FWIW Aaron had a related useful comment at https://github.com/mdn/content/pull/37917#discussion_r1958668942.

files/en-us/web/api/htmlscriptelement/src/index.md
hamishwillee: What do you mean? `trustedScriptURL` is either an URL we are OK with in scriptAllowList  or the empty string.

hamishwillee:
Thanks for the review @wbamberg 

> I made a couple of suggestions to downplay the stuff about CSP here.

Thanks. They made sense.

I have merged all suggestions, but left https://github.com/mdn/content/pull/41413/files#r2548532278 open because I don't understand it - I might perhaps have updated the example since you commented.

This should be good to go I think.

files/en-us/web/api/htmlscriptelement/src/index.md
wbamberg: Yes, I think this was from before you updated the policy to have an allowlist!

wbamberg (approved):
👍 thanks Hamish.

hamishwillee:
No, thanks for reviewing. Getting there.

PR #41414, opened 2025-10-07
Add challenge to test media, form, and table styling

chrisdavidmills:
### Description

Specifically, this PR adds a new challenge that tests the reader's knowledge of form, table, and media styling.

> [!NOTE]
> This PR still needs to be updated to include the correct image URLs, once https://github.com/mdn/shared-assets/pull/71 has been merged.


files/en-us/learn_web_development/core/styling_basics/home_color_scheme_search/index.md
```diff
@@ -0,0 +1,385 @@
...
+1. No border.
+2. A `black` background color and `white` text color.
+3. `5px` rounded corners.
+4. Vertical padding of `10px` and horizontal padding of `2em`.
+5. A background color of `#666666` when hovered or focused.
+6. A background color of `#aaaaaa` when disabled.
+
+### Style the table
+
+You should now add some best practice stylinf to the table, as learned earlier in the module, plus a few extras.
```
hamishwillee:
```suggestion
You should now add some best practice styling to the table, as learned earlier in the module, plus a few extras.
```

files/en-us/learn_web_development/core/styling_basics/home_color_scheme_search/index.md
```diff
@@ -0,0 +1,385 @@
...
+1. Give the table a fixed layout, a width of `100%`, and collapsed borders.
+2. Make the table's top and bottom borders `1px` thick, solid, and colored `#999999`.
+3. Give the table header cells and normal cells `0.6em` of padding, and make their content vertically aligned to the top of the cells.
+4. Give the table header cells a bottom border that is `1px` thick, solid, and colored `#999999`.
+5. Give all of the table rows a width of `20%`, except for the fourth row, which should have a width of `40%`.
+6. Inside the table body, there are four rows. The second cell inside each of these rows has an `rgb()` color as its content. Give each one of these cells the same background color as the color it contains.
+7. Create zebra stripes: Give each odd-numbered row a background color of `#eeeeee`, inside the table body only.
+8. Give the caption padding of `1em`, an italic font style, and letter spacing of `1px`.
+
+### Fixing the image display
```
hamishwillee, line 213: This part of the fix is not highlighted in the solution that I can see.

files/en-us/learn_web_development/core/styling_basics/home_color_scheme_search/index.md
```diff
@@ -0,0 +1,385 @@
...
+You should now add some best practice stylinf to the table, as learned earlier in the module, plus a few extras.
+
+Specifically:
+
+1. Give the table a fixed layout, a width of `100%`, and collapsed borders.
+2. Make the table's top and bottom borders `1px` thick, solid, and colored `#999999`.
+3. Give the table header cells and normal cells `0.6em` of padding, and make their content vertically aligned to the top of the cells.
+4. Give the table header cells a bottom border that is `1px` thick, solid, and colored `#999999`.
+5. Give all of the table rows a width of `20%`, except for the fourth row, which should have a width of `40%`.
+6. Inside the table body, there are four rows. The second cell inside each of these rows has an `rgb()` color as its content. Give each one of these cells the same background color as the color it contains.
```
hamishwillee: I found this hard to understand, probably because I skipped over the HTML.

```suggestion
6. Inside the table body, there are four rows. The second cell inside each of these rows contains text for an `rgb()` color. Give each one of these cells a background color that corresponds to its text.
```

hamishwillee: FWIW it _might_ be nice to also highlight the parts of this one in the solution, rather than just the top label.

hamishwillee (commented):
The style is different than the others we have reviewed - in that it is all done in the play area rather than downloading the files and creating an example. Is there any particular reason for this/does it matter?

I've had a quick look but I'm finding it hard to get a real final feel without the images rendering. Can you ping me when that's working?

chrisdavidmills:
> The style is different than the others we have reviewed - in that it is all done in the play area rather than downloading the files and creating an example. Is there any particular reason for this/does it matter?

Several challenges do it this way; you've mostly reviewed the ones that do it the other way. This way works for smaller UI features and examples that don't take up much space. The other approach works better for entire pages, where having a full browser to test in matters, such as media query-dependent features, etc.

I don't think this is a problem. 
 
> I've had a quick look but I'm finding it hard to get a real final feel without the images rendering. Can you ping me when that's working?

Yup, sure thing.

files/en-us/learn_web_development/core/styling_basics/home_color_scheme_search/index.md
chrisdavidmills: It is here, right underneath the `/* Table styling */` comment:

```css
table img {
  width: 100%;
  height: 150px;
  /* 5. Get images displaying at intrinsic aspect ratio, from bottom */
  object-fit: cover;
  object-position: bottom;
}
```

files/en-us/learn_web_development/core/styling_basics/home_color_scheme_search/index.md
chrisdavidmills: That makes sense. I have added a comment `/* Provide background colors for the "Raw color" cells */` right above those rules.

files/en-us/learn_web_development/core/styling_basics/home_color_scheme_search/index.md
hamishwillee: OK. I was looking for a note like "Fixes the image display". The heading isn't numbered, so the mapping requires some mental gymnastics on scan. I'll leave you to think about whether that's a problem.

chrisdavidmills: I have updated the note to what you suggested, as it is more intuitive.

I think, given the choice of adding numbers to the headings or removing numbers from the comments, I'd rather remove numbers from the comments. I really don't like numbered headings, stylistically.

So, I've removed the numbers and replaced them with "Solution: ". I've also updated the comments so they pretty much exactly match the challenge headings they provide solutions for. 

Let me know what you think. If you like that, I'll roll out the same style to all of the challenges.

chrisdavidmills:
@hamishwillee OK, the images are now merged into `shared-assets`, so you can see what the demo is supposed to look like, before and after.

files/en-us/learn_web_development/core/styling_basics/home_color_scheme_search/index.md
hamishwillee: I think it is much better. Thanks.

files/en-us/learn_web_development/core/styling_basics/home_color_scheme_search/index.md
```diff
@@ -0,0 +1,387 @@
...
+
+/* Table styling */
+
+table img {
+  width: 100%;
+  height: 150px;
+}
+```
+
+{{embedlivesample("app-start", "100%", 500)}}
```
hamishwillee: @chrisdavidmills You might want to make this even taller, exposing more of the example. Ideally full size, though if that looks stupid, I would break it inside one of the cells so that it is clear there is more below the fold.

Ditto for the final version.

hamishwillee (approved):
@chrisdavidmills Looks good. I'd consider making all your solutions in all learn topics follow a similar "Solution" markup. Something to think about.

This is good. I have a nit in https://github.com/mdn/content/pull/41414/files#r2427550252 but approving in case you wish to ignore that.

Nice one.

chrisdavidmills:
> @chrisdavidmills Looks good. I'd consider making all your solutions in all learn topics follow a similar "Solution" markup. Something to think about.

Cool; I'll do that in a separate PR.

> This is good. I have a nit in https://github.com/mdn/content/pull/41414/files#r2427550252 but approving in case you wish to ignore that.

I like this suggestion. I've increased the height a little bit in each case to give more of an idea of what is going on, but not so much that it makes them too unwieldy.

> Nice one.

Thanks for the review, man!

PR #41418, opened 2025-10-07
Document the Umm al-Qura range currently known in browsers in terms of ISO/Gregorian

hsivonen:
### Description

Mentions what the range of KACST-provided Umm al-Qura month length data in browser (as of October 2025) is in terms of the ISO / Gregorian calendar.

### Motivation

This helps readers who aren't used to working with Hijri calendars but are familiar with the Gregorian calendar get an idea of the covered time span.

### Additional details

For verifying:

* Browsers use ICU4C and/or ICU4X.
* The year range being inclusive of 1600 AH can be seen for ICU4X at https://github.com/unicode-org/icu4x/blob/329aa64d04e229c47f04879204a811156eff9576/components/calendar/src/cal/hijri/ummalqura_data.rs and for ICU4C at https://github.com/unicode-org/icu/blob/1bf6bf774dbc8c6c2051963a81100ea1114b497f/icu4c/source/i18n/islamcal.cpp#L264
* The Gregorian correspondence can be checked by running (10 instead of 11 for month due to the oddity of `Date`):

```js
var fmt = new Intl.DateTimeFormat("en-US-u-ca-islamic-umalqura");
console.log(fmt.format(new Date(Date.UTC(1882, 10, 11))));
console.log(fmt.format(new Date(Date.UTC(1882, 10, 12))));
console.log(fmt.format(new Date(Date.UTC(2174, 10, 25))));
console.log(fmt.format(new Date(Date.UTC(2174, 10, 26))));
```


Josh-Cena (approved):
Nice, thank you!

PR #41419, opened 2025-10-07
Add note for anchor positioning fallbacks behavior

djmarland:
### Description

Adding a note for position-try-fallbacks to explain that the element does not revert to its original position when available space is restored.

### Motivation

Developer's intuition for how anchor positioning fallbacks works does not match the specced behaviour, as demonstrated in [this poll](https://mastodon.social/@Meyerweb/115106404124445073)

The spec is not expected to change - https://github.com/w3c/csswg-drafts/issues/12682 - so a note to developers can help prevent the pitfall of not understanding why it's not going back to the starting position.


files/en-us/web/css/position-try-fallbacks/index.md
```diff
@@ -199,6 +199,9 @@ Try scrolling so the anchor nears the edges:
 - If you move the anchor near the top of the viewport, you will see the positioned element flip to the bottom-left of the anchor to avoid overflowing.
 - If you move the anchor near the left of the viewport, you will see the positioned element flip to the top-right of the anchor to avoid overflowing.
 
+> [!NOTE]
+> The positioned element will not return to its original position if the anchor is moved back. As long as the fallback position is still valid (not overflowing), the positioned element will remain in the fallback position.
```
estelle:
```suggestion
As long as the fallback position is still valid (not overflowing), the positioned element will remain in the fallback position, even if the anchor itself is moved back.
```
1. This doesn't need to be a note. 
2. I am not sure what you mean by "if the anchor is moved back".

djmarland: It is referencing the "anchor" image that is being moved around in the interactive demo above it and the two lines above that are instructing the reader to move the anchor around

djmarland: But I acknowledge that is confusing with the concept of an "element marked as an anchor" in the webpage

djmarland: If not a note, should it be a third bullet point since it is a related to the instructions?

estelle: Maybe:

```suggestion
Depending on the browser, once the positioned element moves to the fallback position, it may remain in the fallback position even if the fallback positioning is no longer necessary, such as if updates to the user interface mean the space is available for it to return to the position defined by the {{cssxref("position-area")}}.
```

files/en-us/web/css/reference/properties/position-try-fallbacks/index.md
```diff
@@ -199,6 +199,8 @@ Try scrolling so the anchor nears the edges:
 - If you move the anchor near the top of the viewport, you will see the positioned element flip to the bottom-left of the anchor to avoid overflowing.
 - If you move the anchor near the left of the viewport, you will see the positioned element flip to the top-right of the anchor to avoid overflowing.
 
+Depending on the browser, once the positioned element moves to the fallback position, it may remain in the fallback position even if the fallback positioning is no longer necessary, such as if updates to the user interface mean the space is available for it to return to the position defined by the {{cssxref("position-area")}}.
```
djmarland: I'm a little worried with it not being a note that the "However" below is now disjointed from the bullet points.

djmarland: Maybe there is somewhere else on the page it is more appropriate to highlight this common developer pitfall

Josh-Cena (approved):
I'm happy with this and IMO it flows well enough. We say that "it avoids overflowing" and "it _lazily_ avoids overflowing" and finally "it might not avoid overflowing if it can't".

PR #41420, opened 2025-10-07
Document that the Japanese calendar is currently broken in browsers for dates before 1868-10-23

hsivonen:
### Description

Adds a note that the Japanese calendar doesn't actually at present work correctly in browser for dates before 1868-10-23.

### Motivation

Helps readers avoid triggering a cross-browser bug.

### Additional details

See https://unicode-org.atlassian.net/browse/CLDR-11375 (linked from the text of the PR) for the bug that causes all browsers to have incorrect data for the start of the Meiji era.

Also, see https://tc39.es/proposal-intl-era-monthcode/ , which requires browsers to use Gregorian eras for the Japanese calendar for dates before the Meiji era.


hsivonen:
Side note: This calendar was taken [into use on 1873-01-01](https://en.wikipedia.org/wiki/Adoption_of_the_Gregorian_calendar#Countries_that_used_lunisolar_calendars), so the bug only affects proleptic dates.

files/en-us/web/javascript/reference/global_objects/intl/supportedvaluesof/index.md
```diff
@@ -76,7 +76,7 @@ Below are all values that are commonly supported by browsers for the `calendar`
...
 | `iso8601`          | ISO calendar (variant of the Gregorian calendar with week rules and formatting parameters made region-independent)                                                                                                                                                                                                                                               |
-| `japanese`         | Japanese Imperial calendar (this calendar adds an era for each new emperor, so the output year and era for a future date may not match the input year and era when your code runs on a future engine version)                                                                                                                                                    |
+| `japanese`         | Japanese Imperial calendar (this calendar adds an era for each new emperor, so the output year and era for a future date may not match the input year and era when your code runs on a future engine version. **Note:** As of October 2025, dates prior to 1868-10-23 ISO don't work correctly in browsers due to a CLDR bug.)                                   |
```
Josh-Cena: IMO this isn't particularly helpful (what does "don't work" mean?) and I understand you are trying to make the text succinct. Can we move this into a note after the table, and inline the link rather than presenting it as a reference?

hsivonen: I added longer remarks after the table and said more specifically what the behaviors that don't match the current spec proposal are.

files/en-us/web/javascript/reference/global_objects/intl/supportedvaluesof/index.md
```diff
@@ -76,10 +76,12 @@ Below are all values that are commonly supported by browsers for the `calendar`
...
 | `roc`              | Republic of China calendar                                                                                                                                                                                                                                                                                                                                       |
 
+As of October 2025, in the `japanese` calendar dates prior to 1868-10-23 ISO don't work the way they are supposed to in browsers for two reasons: First, [CLDR had the wrong start date for the Meiji era](https://unicode-org.atlassian.net/browse/CLDR-11375), which makes the calendar implementations treat the Meiji era as extending further to the past than it actually did. Second, the upcoming [Intl era and monthCode Proposal](https://tc39.es/proposal-intl-era-monthcode/) specifies that dates prior to the Meiji era use Gregorian eras, but browsers have shipped behavior that uses approximations of prior Japanese eras instead. This caledar was taken into use on January 1, 6 Meiji / 1873-01-01 ISO, so the problem only affects proleptic dates.
```
Josh-Cena:
```suggestion
As of October 2025, in the `japanese` calendar, dates prior to 1868-10-23 ISO don't work as expected in browsers in two ways. First, [CLDR had the wrong start date for the Meiji era](https://unicode-org.atlassian.net/browse/CLDR-11375), which causes calendar implementations to treat the Meiji era as extending further to the past than it actually did. Second, the upcoming [Intl era and monthCode Proposal](https://tc39.es/proposal-intl-era-monthcode/) specifies that dates prior to the Meiji era should use Gregorian eras, but browsers have traditionally used approximations of prior Japanese eras instead. The `japanese` calendar was taken into use on January 1, 6 Meiji / 1873-01-01 ISO, so these problems only affect proleptic dates.
```

Also I'd like to understand the first problem better: if the starting point of Meiji is wrong, does that mean that all Meiji dates are also offset by that difference, so dates after 1873-01-01, up to the start of Taisho, are affected as well?

hsivonen: The incorrect Meiji start date 1868-09-08 is in the same year as the correct start date 1868-10-23, so the error does not cascade further than affecting 1868-09-08 to and including 1868-10-22. Earlier dates are then affected by the Gregorian era vs. approximation of historical Japanese era issue.

files/en-us/web/javascript/reference/global_objects/intl/supportedvaluesof/index.md
```diff
@@ -76,10 +76,12 @@ Below are all values that are commonly supported by browsers for the `calendar`
...
 | `roc`              | Republic of China calendar                                                                                                                                                                                                                                                                                                                                       |
 
+As of October 2025, in the `japanese` calendar, dates prior to 1868-10-23 ISO don't work as expected in browsers in two ways. First, [CLDR had the wrong start date for the Meiji era](https://unicode-org.atlassian.net/browse/CLDR-11375), which causes calendar implementations to treat the Meiji era as extending further to the past than it actually did. Second, the upcoming [Intl era and monthCode Proposal](https://tc39.es/proposal-intl-era-monthcode/) specifies that dates prior to the Meiji era should use Gregorian eras, but browsers have traditionally used approximations of prior Japanese eras instead. The `japanese` calendar was taken into use on January 1, 6 Meiji / 1873-01-01 ISO, so these problems only affect proleptic dates.
```
Josh-Cena:
```suggestion
As of October 2025, in the `japanese` calendar, dates prior to 1868-10-23 ISO (the start date of the year 1 Meiji) don't work as expected in browsers in two ways. First, [CLDR had the wrong start date for the Meiji era](https://unicode-org.atlassian.net/browse/CLDR-11375), which causes calendar implementations to treat the Meiji era as extending further to the past than it actually did. Second, the upcoming [Intl era and monthCode Proposal](https://tc39.es/proposal-intl-era-monthcode/) specifies that dates prior to the Meiji era should use Gregorian eras, but browsers have traditionally used approximations of prior Japanese eras instead. The `japanese` calendar was taken into use on January 1, 6 Meiji / 1873-01-01 ISO, so these problems only affect proleptic dates.
```

files/en-us/web/javascript/reference/global_objects/intl/supportedvaluesof/index.md
```diff
@@ -76,10 +76,12 @@ Below are all values that are commonly supported by browsers for the `calendar`
...
 | `roc`              | Republic of China calendar                                                                                                                                                                                                                                                                                                                                       |
 
+As of October 2025, in the `japanese` calendar, dates prior to 1868-10-23 ISO (the start date of the year 1 Meiji) don't work as expected in browsers in two ways. First, [CLDR had the wrong start date for the Meiji era](https://unicode-org.atlassian.net/browse/CLDR-11375), which causes calendar implementations to treat the Meiji era as extending further to the past than it actually did. Second, the upcoming [Intl era and monthCode Proposal](https://tc39.es/proposal-intl-era-monthcode/) specifies that dates prior to the Meiji era should use Gregorian eras, but browsers have traditionally used approximations of prior Japanese eras instead. The `japanese` calendar was taken into use on January 1, 6 Meiji / 1873-01-01 ISO, so these problems only affect proleptic dates.
```
Josh-Cena:
```suggestion
As of October 2025, in the `japanese` calendar, dates prior to 1868-10-23 ISO (the start date of the year 1 Meiji) don't work as expected in browsers in two ways. First, [CLDR had the wrong start date for the Meiji era](https://unicode-org.atlassian.net/browse/CLDR-11375), which causes calendar implementations to extend the Meiji era further to the past than it actually did. Second, the upcoming [Intl era and monthCode Proposal](https://tc39.es/proposal-intl-era-monthcode/) specifies that dates prior to the Meiji era should use Gregorian eras, but browsers have traditionally used approximations of prior Japanese eras instead. The `japanese` calendar was taken into use on January 1, 6 Meiji / 1873-01-01 ISO, so these problems only affect proleptic dates.
```

Josh-Cena (approved):
Great, thanks!

PR #41428, opened 2025-10-07
Only use example.com as example domain

### LINKED ISSUE #39865: example.com is a better choice for an example domain
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security

Description

foo.com is a real domain, so I think it's better to use example.com when writing examples.

(Sorry, but it is a machine translation)

`example.com` is better.

https://en.wikipedia.org/wiki/Example.com

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/http/reference/headers/strict-transport-security`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/reference/headers/strict-transport-security/index.md

</details>

Josh-Cena:
Remove a lot of example domains that may happen to be valid, to fix https://github.com/mdn/content/issues/39865. It is impossible to find all such examples, but I found a good bunch.


files/en-us/web/http/reference/headers/access-control-allow-headers/index.md
```diff
@@ -85,7 +85,7 @@ The preflight request below tells the server that we want to send a CORS `GET` r
 OPTIONS /resource/foo
 Access-Control-Request-Method: GET
 Access-Control-Request-Headers: content-type,x-requested-with
-Origin: https://foo.bar.org
+Origin: https://example.com
```
bsmth: Do you want to keep a subdomain here and the one below?

```suggestion
Origin: https://subdomain.example.com
```

files/en-us/mozilla/add-ons/webextensions/api/cookies/index.md
```diff
@@ -39,10 +39,10 @@ For an extension to use this API, it must specify the `"cookies"` [API permissio
...
 
-1. You visit `a-shopping-site.com`, which uses `ad-tracker.com` to deliver its adverts on the web. `ad-tracker.com` sets a cookie associated with the `ad-tracker.com` domain. While you are on `a-shopping-site.com`, `ad-tracker.com` receives information about the products you browse.
-2. You now visit `a-news-site.com` that uses `ad-tracker.com` to deliver adverts. `ad-tracker.com` read its cookie and use the information collected from `a-shopping-site.com` to decide which adverts to display to you.
+1. You visit `a-shopping-site.example`, which uses `ad-tracker.example` to deliver its adverts on the web. `ad-tracker.example` sets a cookie associated with the `ad-tracker.example` domain. While you are on `a-shopping-site.example`, `ad-tracker.example` receives information about the products you browse.
+2. You now visit `a-news-site.example` that uses `ad-tracker.example` to deliver adverts. `ad-tracker.example` read its cookie and use the information collected from `a-shopping-site.example` to decide which adverts to display to you.
```
rebloor: @dotproto any thoughts on this? Does it make the examples to unrealistic?

wbamberg: FWIW I don't much like `.example` as a TLD, as it is so unexpected (same applies in the XSS guide). But if it's the consensus, I would go along with it.

bsmth: +1, this was also my initial read here, but I think it's the right call.

Josh-Cena:
Hi everyone! Since we can't agree on using the .example TLD I'm removing it from this PR, and only working on the cases where the domain name itself has no meaning and can be replaced with example.com.

wbamberg (approved):

PR #41430, opened 2025-10-07
Feat: data-* attribute page for MathMLElement documentation

yashrajbharti:
### Description

- Point to MathML reference
- Created `data-*` for MathML

### DIFF
--- a/files/en-us/web/mathml/reference/global_attributes/data-_star_/index.md
+++ b/files/en-us/web/mathml/reference/global_attributes/data-_star_/index.md
@@ -1,0 +1,55 @@
+---
+title: data-*
+slug: Web/MathML/Reference/Global_attributes/data-*
+page-type: mathml-attribute
+browser-compat: mathml.global_attributes.data_attributes
+sidebar: mathmlref
+---
+
+The **`data-*`** MathML attributes are called custom data attributes. They let MathML markup and its resulting DOM share information that standard attributes can't, usually for scripting purposes. Their custom data are available via the {{domxref("MathMLElement")}} interface of the element the attributes belong to, with the {{domxref("MathMLElement.dataset")}} property.
+
+The `*` can be replaced by any characters allowed in [XML's rules for names](https://www.w3.org/TR/xml/#NT-Name), with the following restrictions:
+
+- Can't start with `xml`.
+- No semicolons (`;`, `U+003A`).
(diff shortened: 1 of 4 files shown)


Josh-Cena:
Each markup language defines its own `data-*` attribute. For MathML, it's normatively defined in https://w3c.github.io/mathml/#parsing_wellformed, specifically `MathMLDataAttributes`. SVG does the same and MDN has a separate SVG `data-*` page: <https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/data-*>. So the bug here is that the page for MathML is unwritten, not that the link is incorrect.

Josh-Cena:
Actually, perhaps you could update the parent page to also point to this hypothetical page as well?

yashrajbharti:
> Actually, perhaps you could update the parent page to also point to this hypothetical page as well?

Oh Understood!, doing that right now

yashrajbharti:
> Actually, perhaps you could update the parent page to also point to this hypothetical page as well?

I copied SVG data-_star_ so we can have this unwritten page

files/en-us/web/mathml/reference/global_attributes/data-_star_/index.md
```diff
@@ -0,0 +1,55 @@
...
+
+You can use this attribute with any MathML element.
+
+## Example
+
+```html
+<math>
+  <msup data-formula="euler" data-value="-1">
+    <mi>e</mi>
+    <mi>iπ</mi>
```
fred-wang: This would be written as `<mrow><mi>i</mi> <mi>π</mi></mrow>` in MathML.

fred-wang (dismissed):

files/en-us/web/mathml/reference/global_attributes/data-_star_/index.md
yashrajbharti: Fixed, and also changed at another place!

fred-wang (commented):
OK thanks. This LGTM for the MathML part but I'll let someone else approve.

yashrajbharti:
Thanks @fred-wang , also @Josh-Cena  wonder if you could take a look at it again 🙂

Josh-Cena (approved):
Thanks, I think this is reasonable. Great addition

PR #41438, opened 2025-10-08
change case of slug to lower

estelle:
all css slugs are sentence case. or will be. 

We should also update the formal syntax with the lowercase link


dipikabh (approved):
Looks good, thanks!

chrisdavidmills (approved):
LGTM2. Leaving it unmerged in case anyone else wants to look at this, since it touches such a lot of pages.

Issue #37526, opened 2025-01-06
Unnecessary use of roles in aria basics article


scottaohara:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility/WAI-ARIA_basics

### What specific section or headline is this issue about?

Signposts/Landmarks

### What information was incorrect, unhelpful, or incomplete?

this section / example should be rethought. per my other comments, there's actually very little that was improved here - since aside from the search role, all other roles added were redundant.

it is also sending a confusing message, because the header, footer and main elements did not have roles added to them - so is this trying to imply that the elements that had the roles added somehow lacked support? or were the redundant roles not added to these elements because they were forgotten?

### What did you expect to see?

if redundant roles were to be added, i'd expect them to be added to all elements with a note specifically stating this was done to support legacy browsers (Internet explorer) but that no modern browser needs any of these roles to be added.

OR, the inaccessible demo should instead just use `div`s with IDs that match the landmark names - and then put roles onto those `div` elements instead - so that the whole section doesn't need to be rewritten.

ideally, this would also mention that the `search` element exists and that could be used in tandem with the `form` element (or instead of it if this was a JS powered search).  

a better job identifying that all the mentioned elements have implicit roles so developers _don't_ need to explicitly add them could be another way to handle this / provide a separate demo (if my other suggestion to make all the elements divs for the initial example was used) to show that HTML is all that's needed here.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/learn_web_development/core/accessibility/wai-aria_basics`
* MDN URL: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility/WAI-ARIA_basics
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/learn_web_development/core/accessibility/wai-aria_basics/index.md
* Last commit: https://github.com/mdn/content/commit/5b20f5f4265f988f80f513db0e4b35c7e0cd70dc
* Document last modified: 2024-12-19T15:37:45.000Z

</details>

scottaohara:
it seems this issue has been resolved, so this can probably be closed?

Josh-Cena:
Yes, I think the current content makes sense

PR #41439, opened 2025-10-08
Prefer <search> over role="search"

### LINKED ISSUE #37799: Accessibility section of <input type='search'> should mention <search> rather than <form role='search'>
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

#search_form_labels_and_accessibility

It encourages you to use `<form role='search'>` for accessibility. It seems to me that this is no longer (and maybe was never) recommended.

It seems to me that this should instead highlight the use of the `<search>` tag, maybe even by just linking, since the article about the search tag covers everything in detail:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/search#accessibility

https://adrianroselli.com/2015/08/where-to-put-your-search-role.html

https://www.scottohara.me/blog/2023/03/24/search-element.html

Thanks!

### MDN metadata

<details>

* Folder: `en-us/web/html/element/input/search`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/element/input/search/index.md
* Last commit: https://github.com/mdn/content/commit/5b20f5f4265f988f80f513db0e4b35c7e0cd70dc
* Document last modified: 2024-12-19T15:37:45.000Z

</details>

Josh-Cena:
Fix https://github.com/mdn/content/issues/37799


files/en-us/learn_web_development/core/accessibility/wai-aria_basics/index.md
```diff
@@ -365,7 +365,7 @@ If you go to VoiceOver's landmarks menu (accessed using VoiceOver key + U and th
 
 However, we could do better here. The search form is a really important landmark that people will want to find, but it is not listed in the landmarks menu or treated like a notable landmark beyond the actual input being called out as a search input (`<input type="search">`).
 
-We could improve it by the use of the ARIA `role="search"`, but using the {{htmlelement("search")}} element implicitly gives that role to the form.
+To mark the form as a landmark, you can either give it ARIA `role="search"`, or wrap it with the {{htmlelement("search")}} element. It is generally better to use HTML semantics where possible and only use ARIA where there is no HTML equivalent.
```
hamishwillee: Nit, if we prefer search element, let's mention it first.


```suggestion
To mark the form as a landmark, you can either wrap it with the {{htmlelement("search")}} element or give it ARIA `role="search"`. It is generally better to use the HTML element where possible as you should only use ARIA where there is no HTML equivalent.
```

files/en-us/web/accessibility/aria/reference/roles/search_role/index.md
```diff
@@ -41,7 +41,9 @@ When a {{HTMLElement('form')}} is a search form, use the `search` role instead o
 
 ### Prefer HTML
 
-Using the {{HTMLElement('form')}} element in conjunction with a declaration of `role="search"` will provide the largest amount of support.
+Using the {{HTMLElement('form')}} element in conjunction with a declaration of `role="search"` will provide the largest amount of support. Alternatively, you can wrap the `<form>` element with the {{HTMLElement('search')}} element. There aren't many reasons to prefer one over the other, other than that `role="search"` may be fewer characters while `<search>` uses native HTML semantics.
+
+However, if you are adding `role="search"` to generic containers like `<div>`, consider using `<search>` instead, just like `role="navigation"` should be replaced with `<nav>`.
```
hamishwillee: This doesn't really match the heading, which is "Prefer HTML". The reason being really that WAI-ARIA say so. It is great you've captured the reality that the role of search has the greatest rollout via `role="search"`. 
I think we should concentrate on depth of browser support, not the character thing.

Maybe

```suggestion
For accessibility you can use either  `role="search"` to mark search elements, or wrap them with the  {{HTMLElement('search')}} element.
Using the element is preferred, if only because this is what is recommended by the relevant accessibility standards.

The main reason you might choose to use `role="search"` is that it has deeper browser support.
```

files/en-us/web/html/reference/elements/input/search/index.md
```diff
@@ -189,25 +189,27 @@ You can see how the placeholder is rendered below:
 
 One problem with search forms is their accessibility; a common design practice is not to provide a label for the search field (although there might be a magnifying glass icon or similar), as the purpose of a search form is normally fairly obvious for sighted users due to placement ([this example shows a typical pattern](https://mdn.github.io/learning-area/accessibility/aria/website-aria-roles/)).
 
-This could, however, cause confusion for screen reader users, since they will not have any verbal indication of what the search input is. One way around this that won't impact on your visual design is to use [WAI-ARIA](/en-US/docs/Learn_web_development/Core/Accessibility/WAI-ARIA_basics) features:
+This could, however, cause confusion for screen reader users, since they will not have any verbal indication of what the search input is. One way around this that won't impact on your visual design is to use landmark elements.
```
hamishwillee: Landmark elements needs a definition link/glossary

files/en-us/web/html/reference/elements/input/search/index.md
```diff
@@ -189,25 +189,27 @@ You can see how the placeholder is rendered below:
...
 
-- A `role` attribute of value `search` on the `<form>` element will cause screen readers to announce that the form is a search form.
+- If your `<input>` is already in a `<form>`, add [`role="search"`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/search_role) to the `<form>` element. Alternatively, consider putting the `<form>` (or the `<input>` if it is not already in a form) inside a {{HTMLElement("search")}} element. Both create a landmark which will cause screen readers to announce that the form is a search form. The `<search>` element uses native HTML semantics, while `role="search"` has more support and may be terser to type if you already have a `<form>` wrapper.
```
hamishwillee: My takeaway was that we should recommend using the element first - so I'd structure to present that option first. I'm not sure how much the terseness matters

hamishwillee (commented):
This looks reasonable with the caveat that we should recommend search first (after all, that's the title) and this focusses on the role first. I understand the reasoning (depth of support) but that will become less significant over time, and I'm not particularly worried about relative terseness.

Josh-Cena:
I think terseness is a big factor when judging two pieces of code that do the same thing. Someone telling me "prefer A over B" doesn't really matter, even if that someone is WAI, if there's no tangible reason for me as a developer or for my users. Also `<search><form></form></search>` provides more "freedom" because now extra content can either go inside the `<form>` or go outside the form but inside `<search>`, and I don't really like introducing unnecessary degrees of freedom, so in a sense terseness means restrictiveness which is nice.

hamishwillee:
> I think terseness is a big factor when judging two pieces of code that do the same thing.

I'm a bit meh. I'd probably just use the search element because its purpose is clear and obvious. But either way, sure, you're the author. Still worth wording search first.

Josh-Cena:
Yeah just did

files/en-us/web/accessibility/aria/reference/roles/landmark_role/index.md
```diff
@@ -18,15 +18,26 @@ A `landmark` is an abstract role for a section of content that is important enou
...
+Do not use `role="landmark"`. Do use HTML and subclass landmark roles. It is also good practice to include the role redundantly with the associated semantic element. It is not as good practice to use non-semantic elements, such as {{HTMLElement('div')}}, adding semantics with landmark roles. But do include one or the other or both. Otherwise, your content is no longer as navigable for screen reader users.
```
hamishwillee: Not sure about this "stylistically" - i.e. it says what you want but does not flow all that well. Maybe.

I would move `Do not use `role="landmark"` to the end because it was just said a few paragraphs before and in any case, I think it is more or less obvious that you should use the specific type.

I would use concrete rather than subclass

```suggestion
It is important to use either or both HTML semantic elements and the concrete landmark roles to mark up your navigable elements, as this ensures that your content will be navigable by screen readers.

The HTML semantic elements are preferred but it is also good practice to include the role redundantly, as roles are supported in older browser versions.
It is not as recommended to use non-semantic elements like {{HTMLElement('div')}} and add semantics with landmark roles, but this is better than nothing!

Remember not to use the abstract role `role="landmark"`.
```

hamishwillee (commented):
@Josh-Cena Nice. I'd quite like a review from @scottohara or @estelle too.

files/en-us/web/accessibility/aria/reference/roles/landmark_role/index.md
estelle:
```suggestion
Do not use `role="landmark"`; rather, use semantic HTML and include subclass landmark roles as appropriate. Although no longer necessary, it is considered good practice to redundantly include the subclass landmark roles with the associated semantic element for legacy browsers. This is preferable to including the `role` attribute to create landmark semantics for non-semantic elements, such as {{HTMLElement('div')}}. But do include one or the other or both. Otherwise, your content will be less navigable for screen reader users.
```

files/en-us/web/accessibility/aria/reference/roles/search_role/index.md
```diff
@@ -41,7 +41,9 @@ When a {{HTMLElement('form')}} is a search form, use the `search` role instead o
 
 ### Prefer HTML
 
-Using the {{HTMLElement('form')}} element in conjunction with a declaration of `role="search"` will provide the largest amount of support.
+Using the {{HTMLElement('search')}} element will automatically communicate a section has a role of `search`. If at all possible, prefer using it instead.
```
estelle:
```suggestion
Using the {{HTMLElement('search')}} element will automatically communicate that the section has a role of `search`. If possible, prefer using the semantic `<search>` element instead of the `search` role.
```

files/en-us/web/accessibility/aria/reference/roles/search_role/index.md
```diff
@@ -41,7 +41,9 @@ When a {{HTMLElement('form')}} is a search form, use the `search` role instead o
...
+
+If your search input is already contained within a {{HTMLElement("form")}}, then wrapping the form in another `<search>` element may be too much markup. In this case, using `role="search"` on the form itself is also acceptable.
```
estelle:
```suggestion
If your `<input>` of type `search` is already contained within a {{HTMLElement("form")}}, then wrapping the form in another `<search>` element may be unnecessary markup. In this case, using `role="search"` on the `<form>` itself is acceptable.
```

files/en-us/web/html/reference/elements/input/search/index.md
```diff
@@ -189,25 +189,27 @@ You can see how the placeholder is rendered below:
 
 One problem with search forms is their accessibility; a common design practice is not to provide a label for the search field (although there might be a magnifying glass icon or similar), as the purpose of a search form is normally fairly obvious for sighted users due to placement ([this example shows a typical pattern](https://mdn.github.io/learning-area/accessibility/aria/website-aria-roles/)).
 
-This could, however, cause confusion for screen reader users, since they will not have any verbal indication of what the search input is. One way around this that won't impact on your visual design is to use [WAI-ARIA](/en-US/docs/Learn_web_development/Core/Accessibility/WAI-ARIA_basics) features:
+This could, however, cause confusion for screen reader users, since they will not have any verbal indication of what the search input is. One way around this that won't impact on your visual design is to use [landmark elements](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/landmark_role).
```
estelle:
```suggestion
This could, however, cause confusion for screen reader users, as they will not receive any verbal indication of what the search input is. One way around this that won't impact your visual design is to use [landmark elements](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/landmark_role).
```
not your edits, but since we're editing....

files/en-us/web/html/reference/elements/input/search/index.md
```diff
@@ -189,25 +189,27 @@ You can see how the placeholder is rendered below:
...
 
-- A `role` attribute of value `search` on the `<form>` element will cause screen readers to announce that the form is a search form.
+- Wrap the whole search functionality in a {{HTMLElement("search")}} element, which creates a landmark region that assistive technologies can announce and quickly navigate to. If your `<input>` is already in a `<form>`, you can alternatively add [`role="search"`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/search_role) to the `<form>` element, which also makes the `<form>` a search landmark. The `<search>` element uses native HTML semantics, while `role="search"` has more support and may be terser to type if you already have a `<form>` wrapper.
```
estelle, line 194:
```suggestion
- Wrap the whole search functionality in a {{HTMLElement("search")}} element, which creates a landmark region that assistive technologies can announce and quickly navigate to. If your `<input>` is already in a `<form>`, you can alternatively add [`role="search"`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/search_role) to the `<form>` element, which also makes the `<form>` a search landmark. The `<search>` element uses native HTML semantics, while `role="search"` is well supported and may be terser to type if you already have a `<form>` wrapper.
```
To future-proof, since they may soon be equally well supported (if not already the case)

estelle (commented):
a few nits....

files/en-us/web/html/reference/elements/input/search/index.md
Josh-Cena: While they can both be "well-supported", it is always the case that `role="search"` has "more support". Whenever we talk about AT, old browsers and ATs are still relevant.

files/en-us/web/accessibility/aria/reference/roles/search_role/index.md
Josh-Cena: This is standard verbiage for all landmark roles so I'm not inclined to change it in this PR

estelle: Avoid using "it" whenever possible, opting to clarify what "it" refers to instead, as readers may not know what "it" refers to.

estelle (approved):
Thanks! 🎉

files/en-us/web/accessibility/aria/reference/roles/landmark_role/index.md
```diff
@@ -18,15 +18,26 @@ A `landmark` is an abstract role for a section of content that is important enou
...
+Do not use `role="landmark"`; rather, use subclass landmark roles as appropriate, or semantic HTML if possible. Although no longer necessary, it is considered good practice to redundantly include the subclass landmark roles with the associated semantic element for legacy browsers. This is preferable to using landmark roles on non-semantic elements, such as {{HTMLElement('div')}}, but do use at least one of `role` or semantic elements for creating landmarks. Otherwise, your content will be less navigable for screen reader users.
```
scottaohara, line 40: the ARIA basics article had redundant roles declared in one of the sections - prompting me to open #37526.  but it looks like those redundant roles were removed (yay!)

i would suggest that if this advice is kept at all, it's front-loaded to indicate this is _only necessary_ if one needs to support legacy browsers - like IE11 and older.

for instance:
>Although no longer necessary, if one needs to support legacy browsers it may be beneficial to include the subclass landmark roles with the associated semantic element for legacy browsers.

---


IMO, the following could do better to explain _why_ it's preferrable to use the semantic HTML elements vs `div role=name-of-landmark` - since they'll effectively be the same thing in the a11y tree, and generally speaking - outside of the `form` element, none of the other HTML landmarks have any functionality or notable styling.

>This is preferable to using landmark roles on non-semantic elements, such as `div`, but do use at least one of `role` or semantic elements for creating landmarks. Otherwise, your content will be less navigable for screen reader users.

for instance, might it be preferable because once legacy browser support is dropped, the redundant role can be dropped without needing to change the HTML element?  That it allows for easier styling hooks than relying on the `role=...`.  other?

files/en-us/web/html/reference/elements/input/search/index.md
scottaohara: it should be more of a legacy browser issue than an AT (screen reader, really) issue.

if role=search is supported by the AT, then any browser that exposes the `search` element with that landmark should treat it just fine.

it's the browsers that don't support the search landmark mapping for the element (browsers pre-its 2023 release).

per my previous comment - _this_ would be the most valid reason to double up the `role=search` on the search element, or use `form role=search`, since this landmark is the newest and thus has the greatest potential for `<search>` to not be respected by anyone's browser that hasn't been updated for the last 2.5 years.

scottaohara (commented):
unfortunately i had a review going and forgot to hit submit...

Josh-Cena:
Thanks @scottaohara for the reviews. This is a bit hard to unpack and I'm not sure what concrete suggestions they map to, other than to frontload the suggestion with its applicable scenario (which is a bit minor to me, but I agree). I disagree that there are any tangible differences between `<div role="navigation">` and `<nav>`, other than terseness and "because WAI said so"—regarding styling, I think there are so many ways in which markup can change that people should just learn to use class names instead. But if you have ideas about how the current text can be rewritten, I will be very glad to review any follow up PRs.

scottaohara:
>I disagree that there are any tangible differences between `<div role="navigation">` and `<nav>`,

how do you disagree? because your response seems to indicate that you _do_ agree with my critique of the following text from the doc:
>This is preferable to using landmark roles on non-semantic elements, such as div, but do use at least one of role or semantic elements for creating landmarks. Otherwise, your content will be less navigable for screen reader users.

my comment on styling was trying to provide a reason as to _why_ that text might be in there. But, as you noted you think people should be using classes - so then what would be justification for that line other than "terseness and because WAI said so"?

does that help clarify?

Josh-Cena:
Perhaps I misread. So you are suggesting that the existing text about "This is preferable to..." is unjustified? To me, this is just a prescriptive suggestion like "you should do A instead of B because we and important people said so", so while I personally can't see it tangibly matters, I think it's acceptable—stylistic suggestions are a big part of MDN and we do have subjective recommendations about how people ought to code.

PR #41443, opened 2025-10-09
Docs: Clarify <progress> element attributes

dogxii:
### Description

Clarify the <progress> element attributes, delete the `min` attribute which is not allowed for the <progress> element.

### Motivation

In the [`<progress>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress) MDN Web, have this note:

> [!NOTE] 
> Note: Unlike the [`<meter>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meter) element, the minimum value is always 0, and the min attribute is not allowed for the \<progress\> element.

so I delete the `min` attribute in `<progress>` and the description in these pages.

### Additional details

[\<progress\> - HTML | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress)


Josh-Cena (approved):
Thanks, LGTM

PR #41446, opened 2025-10-09
Correction to the use of `textTrack` referring to `TextTrack`

Chukwujiobi-Canon:
The documentation has a slight typographical error which makes a huge difference in reading and processing by automated systems. This commit aims to correct this typographical error.


chrisdavidmills (approved):
LGTM; thanks for the fix, @drizzle042!

PR #41449, opened 2025-10-09
Add note on convention

smith558:
### Description

Add a note on the convention used in some examples in the article.

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/object/create/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/object/create/index.md
@@ -9,6 +9,10 @@ sidebar: jsref
 
 The **`Object.create()`** static method creates a new object, using an existing object as the prototype of the newly created object.
 
+> [!NOTE]
+> By convention, capitalised names are often used for objects that represent prototypes, and lowercase names for regular instances.
+> 
+
 {{InteractiveExample("JavaScript Demo: Object.create()", "taller")}}
 
 ```js interactive-example


Josh-Cena:
Thanks, but I don't think this note is worthwhile. First this note is incorrect: capitalized names are used for _constructors_, not _prototypes_: constructors are functions. Second this convention is already universal, both on MDN and in JavaScript in general. It's not a specific quirk of this page.

smith558:
> Thanks, but I don't think this note is worthwhile. First this note is incorrect: capitalized names are used for _constructors_, not _prototypes_: constructors are functions. Second this convention is already universal, both on MDN and in JavaScript in general. It's not a specific quirk of this page.

Then many other MDNs articles (e.g. [this one](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_objects#using_the_object.create_method)) do this incorrectly because **they capitalise many prototypes** (not constructors).

smith558:
> Second this convention is already universal, both on MDN and in JavaScript in general. It's not a specific quirk of this page.

Yes, but this would have been the (possibly) 1st time in logical order of the articles a reader could be exposed to these.

*Which is not relevant if the note is wrong (as you say) to begin with, though.

smith558:
@Josh-Cena, I have to say, I feel like (some maintainers) close PRs prematurely and thus lose out on valuable contributions/insights. Like in this case, that the other MDN articles often incorrectly capitalise prototypes.

Josh-Cena:
?? The article you are linking is fully consistent: uppercase for constructor (`new Car()`), lowercase for instances (`car1`). I don't know what you mean by "prototype"—the rule for capitalization is simple:

- Does its `typeof` return `function`? If not, it's lowercase
- If yes, is it used as a constructor (i.e., called with `new`)? If not, it's lowercase
- If yes, it's uppercase

Josh-Cena:
OK I see what you mean with that page: there are a few places where we capitalize plain objects like `Animal` and `Manager`. That's incorrect. You are welcome to fix that separately, but correct cases do not warrant a note.

smith558:
> OK I see what you mean with that page: there are a few places where we capitalize plain objects like `Animal` and `Manager`. That's incorrect. You are welcome to fix that separately, but correct cases do not warrant a note.

Might look into it later.

pransh15:
Hi @smith558 — Thanks for opening this Pull Request! I really appreciate your effort to help improve MDN through your contribution.

PR #41456, opened 2025-10-09
Add label to English captions track in interactive example

jacob-willden:
### Description
Adds the label "English" to the captions track in the interactive example for [the `<track>` element page](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/track).

### Motivation

This will help readers using the interactive example to select or deselect the English captions track.

### Additional details

Not applicable.

None found.


estelle (approved):
Thanks! 🎉

PR #41458, opened 2025-10-10
Intl.DurationFormat Example Code: fix space error

JangAyeon:
### Description

The spacing in the output of the example code in the [original text](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat) does not match the actual output when the code is executed.

**output of the example code from original MDN document : `"1h 46min 40s"`**
image

**actual output when code is excuted : `"1 h 46 min 40 s"`**
image


Josh-Cena (approved):
Good catch, thank you

PR #41469, opened 2025-10-11
Lint: quote non-generic font-family values (1)

Josh-Cena:
After https://github.com/mdn/content/pull/40228, we are a lot more explicit on some CSS stylistic decisions. In particular, we now require all non-generic font family names to be quoted, no exceptions—not even `"Arial"`. In addition, I noticed that the weird `Arial, Helvetica, sans-serif` combination is used a lot—which is pointless because every computer that supports Helvetica is also going to support Arial, so I reordered these two.

### DIFF
--- a/files/en-us/learn_web_development/core/frameworks_libraries/angular_styling/index.md
+++ b/files/en-us/learn_web_development/core/frameworks_libraries/angular_styling/index.md
@@ -46,7 +46,7 @@ In `src/styles.css`, paste the following styles:
 
 ```css
 body {
-  font-family: Helvetica, Arial, sans-serif;
+  font-family: "Helvetica", "Arial", sans-serif;
 }
 
 .btn-wrapper {
--- a/files/en-us/learn_web_development/core/frameworks_libraries/react_todo_list_beginning/index.md
+++ b/files/en-us/learn_web_development/core/frameworks_libraries/react_todo_list_beginning/index.md
@@ -487,7 +487,7 @@ body {
   box-sizing: border-box;
   clear: left;
   display: block;
-  font-family: Arial, sans-serif;
+  font-family: "Arial", sans-serif;
   font-size: 1.6rem;
   font-weight: 400;
   line-height: 1.25;
--- a/files/en-us/learn_web_development/core/styling_basics/handling_conflicts/index.md
+++ b/files/en-us/learn_web_development/core/styling_basics/handling_conflicts/index.md
@@ -314,7 +314,7 @@ This behavior helps avoid repetition in your CSS. A common practice is to define
 h2 {
   font-size: 2em;
   color: black;
-  font-family: Georgia, "Times New Roman", Times, serif;
+  font-family: "Georgia", serif;
 }
 
 .small {
--- a/files/en-us/learn_web_development/core/text_styling/fundamentals/index.md
+++ b/files/en-us/learn_web_development/core/text_styling/fundamentals/index.md
@@ -91,11 +91,11 @@ To set a different font for your text, you use the {{cssxref("font-family")}} pr
 
 ```css
 p {
-  font-family: Arial;
+  font-family: "Arial";
 }
 ```
 
-This would make all paragraphs on a page adopt the arial font, which is found on any computer.
+This would make all paragraphs on a page adopt the Arial font, which is found on any computer.
 
 > [!NOTE]
 > Scrimba's [Web-safe fonts](https://scrimba.com/learn-html-and-css-c0p/~01r?via=mdn) <sup>[_MDN learning partner_](/en-US/docs/MDN/Writing_guidelines/Learning_content#partner_links_and_embeds)</sup> scrim provides an interactive guide to why fonts are important, web-safe fonts, and how to specify fonts in CSS — along with a challenge to test your knowledge.
@@ -272,7 +272,7 @@ Since you can't guarantee the availability of the fonts you want to use on your 
 
 ```css
 p {
-  font-family: "Trebuchet MS", Verdana, sans-serif;
+  font-family: "Trebuchet MS", "Verdana", sans-serif;
 }
 ```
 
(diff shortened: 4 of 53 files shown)


files/en-us/learn_web_development/core/frameworks_libraries/angular_styling/index.md
```diff
@@ -46,7 +46,7 @@ In `src/styles.css`, paste the following styles:
 
 ```css
 body {
-  font-family: Helvetica, Arial, sans-serif;
+  font-family: "Helvetica", "Arial", sans-serif;
```
hamishwillee, line 49: Do we need sans-serif - Arial is web safe https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide/CSS#specifying_font_families

Ditto for all other Arial cases (this might be the only one)

files/en-us/learn_web_development/core/frameworks_libraries/react_todo_list_beginning/index.md
```diff
@@ -487,7 +487,7 @@ body {
   box-sizing: border-box;
   clear: left;
   display: block;
-  font-family: Arial, sans-serif;
+  font-family: "Arial", sans-serif;
```
hamishwillee, line 490: Do we need sans-serif - Arial is web safe https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide/CSS#specifying_font_families

files/en-us/learn_web_development/core/styling_basics/handling_conflicts/index.md
```diff
@@ -314,7 +314,7 @@ This behavior helps avoid repetition in your CSS. A common practice is to define
 h2 {
   font-size: 2em;
   color: black;
-  font-family: Georgia, "Times New Roman", Times, serif;
+  font-family: "Georgia", serif;
```
hamishwillee, line 317: Assume that this removal is OK in the context of other docs in page

files/en-us/learn_web_development/core/text_styling/fundamentals/index.md
```diff
@@ -272,7 +272,7 @@ Since you can't guarantee the availability of the fonts you want to use on your
 
 ```css
 p {
-  font-family: "Trebuchet MS", Verdana, sans-serif;
+  font-family: "Trebuchet MS", "Verdana", sans-serif;
```
hamishwillee, line 275: Do we need sans-serif - Verdana  and Trebuchet MS are both web safe https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide/CSS#specifying_font_families

Ditto for all other cases (this might be the only one)

files/en-us/learn_web_development/core/frameworks_libraries/angular_styling/index.md
hamishwillee: PS I put these kind of comments in all the PRs. I'm sure you'll globally fix these (if I am correct) so sorry about the spam

files/en-us/web/svg/tutorials/svg_from_scratch/using_fonts/index.md
```diff
@@ -20,7 +20,7 @@ Note that the CSS here is nested within an SVG {{SVGElement("style")}} element,
   <style>
     text {
       /* Specify the system or custom font to use */
-      font-family: "Courier New", sans-serif;
+      font-family: "Courier New", monospace;
```
hamishwillee, line 23: Courier New is web safe for monospace

files/en-us/learn_web_development/core/styling_basics/handling_conflicts/index.md
Josh-Cena: Georgia is web safe, so the other fonts will never be used

files/en-us/learn_web_development/core/styling_basics/handling_conflicts/index.md
hamishwillee, line 305: Given change below, shouldn't this be

```suggestion
This behavior helps avoid repetition in your CSS. A common practice is to define generic styles for the basic elements, and then create classes for those that are different. For example, in the stylesheet below, we have defined a generic style for level 2 headings, and then created some classes that change only some of the properties and values. The values defined initially are applied to all headings, then the more specific values are applied to the headings with the classes.
```

Josh-Cena: You just changed "generic styles" to "a generic style" right? I think the former is more correct because "styles" just refer to abstractly all the CSS stylings, not a specific declaration block.

hamishwillee: What's up with GitHub at the moment??? Anyway, I _think_ this is more correct because we are referring to "for level 2 headings", which is a specific declaration block. Right?

> For example, in the stylesheet below, we have defined **a** generic style~s~ for level 2 headings

Josh-Cena: To me "styles" refer to each property inside the declaration and not the declaration as a whole. For example I would treat the size, font, color, etc. as different "styles".

hamishwillee: I personally like to use styles as you do. 
FWIW though, the last time I looked at the specs they seem to refer to a style block in the singular - so the style of an element is the whole block, not the set of styles. 

But anyhow, since this doesn't really apply to your changes lets ignore this. Thanks

hamishwillee (approved):
Thanks!

Josh-Cena:
Thanks for all the reviews!

hamishwillee:
You're welcome. Personally if I've done the work, I like to get the PRs off my mental queue as soon as possible.

PR #41470, opened 2025-10-11
Lint: quote non-generic font-family values (2)

Josh-Cena:
After https://github.com/mdn/content/pull/40228, we are a lot more explicit on some CSS stylistic decisions. In particular, we now require all non-generic font family names to be quoted, no exceptions—not even `"Arial"`. In addition, I noticed that the weird `Arial, Helvetica, sans-serif` combination is used a lot—which is pointless because every computer that supports Helvetica is also going to support Arial, so I reordered these two.

### DIFF
--- a/files/en-us/web/css/animation-range-end/index.md
+++ b/files/en-us/web/css/animation-range-end/index.md
@@ -110,7 +110,7 @@ The `subject` element and its containing `content` element are styled minimally,
 
 p,
 h1 {
-  font-family: Arial, Helvetica, sans-serif;
+  font-family: "Helvetica", "Arial", sans-serif;
 }
 
 h1 {
--- a/files/en-us/web/css/css_transforms/index.md
+++ b/files/en-us/web/css/css_transforms/index.md
@@ -358,8 +358,8 @@ Use the sliders in the example below to modify the translation, rotation, scale,
   margin-bottom: 4px;
   accent-color: blue; /* or any color */
   font-family:
-    "Inter", "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell",
-    "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
+    "Inter", "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans",
+    "Droid Sans", "Helvetica Neue", sans-serif;
 }
 
 #allTransformFieldset > legend {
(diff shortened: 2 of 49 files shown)


files/en-us/web/css/animation-range-end/index.md
```diff
@@ -110,7 +110,7 @@ The `subject` element and its containing `content` element are styled minimally,
 
 p,
 h1 {
-  font-family: Arial, Helvetica, sans-serif;
+  font-family: "Helvetica", "Arial", sans-serif;
```
hamishwillee, line 113: Do we need sans - Arial is web safe https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide/CSS#specifying_font_families

hamishwillee: Ditto below, all cases with Arial

files/en-us/web/css/css_transforms/index.md
```diff
@@ -358,8 +358,8 @@ Use the sliders in the example below to modify the translation, rotation, scale,
   margin-bottom: 4px;
   accent-color: blue; /* or any color */
   font-family:
-    Inter, "system-ui", "Segoe UI", Roboto, Oxygen, Ubuntu, Cantarell,
-    "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
+    "Inter", "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans",
```
hamishwillee, line 361: Removal of "system-ui" because? (specifically, because I haven't checked this is referenced elsewhere in the doc)

files/en-us/web/css/animation-range-end/index.md
Josh-Cena: I thought about removing it, but opted to keep it for now because that would be a separate stylistic issue, one that we don't have guidelines for (we say it's "optional" to include generic families for web-safe fonts, but we don't say it's "not allowed).

files/en-us/web/css/css_transforms/index.md
Josh-Cena: `"system-ui"` is not a generic family; it's a specific font called `system-ui` that likely doesn't exist. Since this is likely a bug anyway I just removed it. Also we are discouraging `system-ui`: https://developer.mozilla.org/en-US/docs/Web/CSS/generic-family#system-ui

files/en-us/web/css/animation-range-end/index.md
hamishwillee: Fair enough!

files/en-us/web/css/css_transforms/index.md
hamishwillee: Works for me - it isn't mentioned elsewhere in this doc.

hamishwillee (approved):
Thanks @Josh-Cena

PR #41520, opened 2025-10-14
Fix method name

srdrtmy:
This page describes Date.prototype.toLocaleDateString() method but in the second paragraph it refers to toLocaleString() method instead of toLocaleDateString. It is not clear that toLocaleDateString() method performs a search in a big database of localization strings or not.


Josh-Cena (approved):
Good catch, thanks

PR #41524, opened 2025-10-14
Editorial review: Add precommitHandler() and other info to Navigation API docs

chrisdavidmills:
### Description

Chromium 141 adds support for the `precommitHandler` option of the [`NavigateEvent.intercept()`](https://developer.mozilla.org/en-US/docs/Web/API/NavigateEvent/intercept) method. See https://chromestatus.com/feature/5134734612496384.

This PR adds documentation for it.

It also overhauls the `intercept()` method page in general, adding a bunch of details in a large "Description" section that was missing before. I think more work needs to be done here, but it is out of the scope of this PR for me to try to fix everything. For now, I've added a bit more information about how the `intercept()` features work so that the page is a bit more informative.


files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -25,7 +25,9 @@ intercept(options)
 - `options` {{optional_inline}}
   - : An options object containing the following properties:
     - `handler` {{optional_inline}}
-      - : A callback function that defines what the navigation handling behavior should be. This generally handles resource fetching, and returns a promise.
+      - : A callback function that defines what the navigation handling behavior should be; it returns a promise. This function will run after the {{domxref("Navigation.navigate_event", "navigate")}} event has fired and the {{domxref("Navigation.currentEntry", "currentEntry")}} property has been updated.
+    - `precommitHandler` {{optional_inline}}
+      - : A callback function that defines any behavior that should occur just before the navigation has finished; it accepts a controller object as an argument and returns a promise. This function will run after the {{domxref("Navigation.navigate_event", "navigate")}} event has fired but before the {{domxref("Navigation.currentEntry", "currentEntry")}} property has been updated.
```
noamr: It's a bit strange to say "after navigate event has fired" because you can only intercept once the event is fired.

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,145 @@ None (`undefined`).
 - `InvalidStateError` {{domxref("DOMException")}}
   - : Thrown if the current {{domxref("Document")}} is not yet active, or if the navigation has been cancelled.
 - `SecurityError` {{domxref("DOMException")}}
-  - : Thrown if the event was dispatched by a {{domxref("EventTarget.dispatchEvent", "dispatchEvent()")}} call, rather than the user agent, or if the navigation cannot be intercepted (i.e., {{domxref("NavigateEvent.canIntercept")}} is `false`).
+  - : Thrown if:
+    - The event was dispatched by a {{domxref("EventTarget.dispatchEvent", "dispatchEvent()")}} call, rather than the user agent.
+    - The navigation cannot be intercepted ({{domxref("NavigateEvent.canIntercept")}} is `false`).
+    - A `precommitHandler()` callback was called on a non-cancelable event ({{domxref("Event.cancelable")}} is `false`).
```
noamr: s/called/provided

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,145 @@ None (`undefined`).
...
+
+## Description
+
+The `intercept()` method is used to implement custom navigation behavior when a link is clicked in an application. It does this via a couple of different callbacks.
```
noamr: Not only links. Also form submissions and any type of programmatic navigation like calling `pushState`.

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,145 @@ None (`undefined`).
...
+
+### Handling immediate navigations with `handler()`
+
+The `handler()` callback is generally used to handle resource fetching and updating the UI to show the content for the new page, which will typically look like this, at a basic level:
```
noamr: Don't know about "at a basic level"...

Technically, a way to describe this is that `handler` occurs as a response to a committed navigation, and `precommitHandler` occurs as a way to modify or cancel an in-flight navigation, or to perform operations while the navigation is ongoing and before it is committed.

This also works with the browser UI - e.g. the precommit handlers work before the current URL is updated in the URL bar / omnibox.

I think this distinction was a bit missing for me throughout.

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,145 @@ None (`undefined`).
...
+
+  // Include multiple conditions for different page types here, as needed
+});
+```
+
+`handler()` will run after the {{domxref("Navigation.navigate_event", "navigate")}} event has fired and the {{domxref("Navigation.currentEntry", "currentEntry")}} property has been updated, meaning that it should be used to implemnent navigation behavior where the navigation is committed to. The new URL is shown in the browser, and history is updated, and the user should be shown something new.
+
+### Handling precommit actions with `precommitHandler()`
+
+However, depending on the URL, you might not always want to commit the navigation immediately. For example, if the user navigates to a restricted page and the user is not signed in, you may want to redirect the browser to a sign-in page. This kind of scenario can be dealt with using the `precommitHandler()` callback, which will run after the {{domxref("Navigation.navigate_event", "navigate")}} event has fired but before the {{domxref("Navigation.currentEntry", "currentEntry")}} property has been updated.
```
noamr: Not only depending on the URL. It can depend on all kinds of conditions, e.g. the source element.

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,145 @@ None (`undefined`).
...
+  }
+});
+```
+
+This pattern is simpler than the alternative of canceling the original navigation and starting a new one to the redirect location, because it avoids exposing the intermediate state. For example, only one `navigatesuccess` or `navigateerror` event fires, and if the navigation was triggered by a call to {{domxref("Navigation.navigate()")}}, the promise only fulfills once the redirect destination is reached.
+
+The `precommitHandler()` callback takes a `controller` object as an argument, which contains a `redirect()` method. The `redirect()` method takes two parameters — a string representing the URL to redirect to, and an options object containing two parameters:
+
+- `state` {{optional_inline}}
+  - : Contains any state information you want to pass along with the navigation, for example for logging or tracking purposes.
```
noamr: This state is the same as the state in https://developer.mozilla.org/en-US/docs/Web/API/NavigationHistoryEntry/getState

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,145 @@ None (`undefined`).
...
+
+When the promises returned by the `intercept()` handler functions fulfill, the `Navigation` object's {{domxref("Navigation/navigatesuccess_event", "navigatesuccess")}} event fires, allowing you to run cleanup code after a successful navigation has completed. If they reject, meaning the navigation has failed, {{domxref("Navigation/navigateerror_event", "navigateerror")}} fires instead, allowing you to gracefully handle the failure case.
+
+There is also a `finished` property on the return value of navigation methods (such as {{domxref("Navigation.navigate()")}}), which fulfills or rejects at the same time as the aforementioned events are fired, providing another path for handling the success and failure cases.
+
+### Interaction between `precommitHandler()` and `handler()`
+
+Both `precommitHandler()` and `handler()` callbacks can be included inside the same `intercept()` call.
+
+- When the `precommitHandler()` promise fulfills, the browser moves on to run any `handler()` callback present. If `precommitHandler()` rejects, `navigateerror` fires, the `committed` and `finished` promises reject, and the navigation is cancelled.
```
noamr: Note that multiple intercept calls are possible, which means something like "all precommit handlers are called first, and when all of them resolved, the navigation commits and the handler callbacks are called".

files/en-us/web/api/navigation_api/index.md
```diff
@@ -28,7 +28,9 @@ The `navigation` interface has several associated events, the most notable being
 
 The `NavigationEvent` object also provides two methods:
 
-- {{domxref("NavigateEvent.intercept", "intercept()")}} takes as an argument a callback handler function returning a promise. It allows you to control what happens when the navigation is initiated. For example, in the case of an SPA, it can be used to load relevant new content into the UI based on the path of the URL navigated to.
+- {{domxref("NavigateEvent.intercept", "intercept()")}} allows you to specify custom behavior for navigations, and can take the following as arguments:
+  - Callback handler functions allowing you to specify what happens both _when_ the navigation is initiated and _just before_ the navigation is initiated. For example, it could be used to load relevant new content into the UI based on the path of the URL navigated to, or redirect the browser to a sign-in page if the URL points to a restricted page and the user is not signed in.
```
noamr: The word "initiated" a bit inaccurate. It's before/when the navigation is committed.
The navigate event itself correspond to when the navigation was "initiated".

noamr (commented):
Thanks for working on this, see comments

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: I'm pretty sure I stole that wording from the explainer ;-)

But I see what you mean — the whole point is that it is run after the event is fired, so it isn't worth saying. I've updated this, and similar wording, to remove that bit, so for example:

> This function will run before the {{domxref("Navigation.currentEntry", "currentEntry")}} property has been updated

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: Updated; thanks.

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: Good point. I've updated this bit to

> The `intercept()` method is used to implement custom SPA navigation behavior when a navigation occurs, for example when a link is clicked, a form is submitted, or a programmatic navigation is initiated (using {{domxref("History.pushState()")}}, {{domxref("Window.location")}}, etc.).
>
> It does this via a couple of different callbacks, `handler()` and `precommitHandler()`.

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: This makes sense, yeah. I've attempted to update the descriptive text around the two handlers to follow this pattern a bit more. Let me know what you think.

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: I changed this bit anyway, as per the previous comment.

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: Good point; I've added a mention and a link to the description:

> Contains any state information you want to pass along with the navigation, for example, for logging or tracking purposes. The state for the navigation can subsequently be retrieved via {{domxref("NavigationHistoryEntry.getState()")}}.

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: OK, this is good to know. I've rewritten the section like so:

---

Both `precommitHandler()` and `handler()` callbacks can be included inside the same `intercept()` call.

1. First, the `precommitHandler()` handler runs.
   - When the `precommitHandler()` promise fulfills, the navigation commits.
   - If the `precommitHandler()` rejects, `navigateerror` fires, the `committed` and `finished` promises reject, and the navigation is cancelled.

2. When the navigation commits, a new {{domxref("NavigationHistoryEntry")}} is created for the navigation, and its `committed` promise fulfills.
3. Next, the `handler()` promise runs.
   - When the `handler()` promise fulfills and the `navigatesuccess` event fires, the navigation `finished` promise fulfills as well, to indicate the navigation is finished.
   - If `handler()` rejects, `navigateerror` fires, the `finished` promise rejects, and the navigation is canceled.

Note that the above process is upheld even across multiple `intercept()` calls on the same `NavigateEvent`. All `precommitHandler()` callbacks are called first, and when all of them resolve, the navigation commits, and all the `handler()` callbacks are called.

files/en-us/web/api/navigation_api/index.md
chrisdavidmills: Got it. I've changed these instances to "committed".

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -25,7 +25,9 @@ intercept(options)
 - `options` {{optional_inline}}
   - : An options object containing the following properties:
     - `handler` {{optional_inline}}
-      - : A callback function that defines what the navigation handling behavior should be. This generally handles resource fetching, and returns a promise.
+      - : A callback function that defines what the navigation handling behavior should be; it returns a promise. This function will run after the {{domxref("Navigation.currentEntry", "currentEntry")}} property has been updated.
+    - `precommitHandler` {{optional_inline}}
+      - : A callback function that defines any behavior that should occur just before the navigation has finished; it accepts a controller object as an argument and returns a promise. This function will run before the {{domxref("Navigation.currentEntry", "currentEntry")}} property has been updated.
```
noamr:
```suggestion
      - : A callback function that defines any behavior that should occur just before the navigation has committed; it accepts a controller object as an argument and returns a promise. This function will run before the {{domxref("Navigation.currentEntry", "currentEntry")}} property has been updated.
```

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,156 @@ None (`undefined`).
...
 - `SecurityError` {{domxref("DOMException")}}
-  - : Thrown if the event was dispatched by a {{domxref("EventTarget.dispatchEvent", "dispatchEvent()")}} call, rather than the user agent, or if the navigation cannot be intercepted (i.e., {{domxref("NavigateEvent.canIntercept")}} is `false`).
+  - : Thrown if:
+    - The event was dispatched by a {{domxref("EventTarget.dispatchEvent", "dispatchEvent()")}} call, rather than the user agent.
+    - The navigation cannot be intercepted ({{domxref("NavigateEvent.canIntercept")}} is `false`).
+    - A `precommitHandler()` callback was provided on a non-cancelable event ({{domxref("Event.cancelable")}} is `false`).
+
+## Description
+
+The `intercept()` method is used to implement custom SPA navigation behavior when a navigation occurs, for example when a link is clicked, a form is submitted, or a programmatic navigation is initiated (using {{domxref("History.pushState()")}}, {{domxref("Window.location")}}, etc.).
```
noamr: I think "SPA navigation behavior" is a bit of an awkward wording, but I'll leave that to the editorial review.

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,156 @@ None (`undefined`).
...
+
+The `precommitHandler()` callback takes a `controller` object as an argument, which contains a `redirect()` method. The `redirect()` method takes two parameters — a string representing the URL to redirect to, and an options object containing two parameters:
+
+- `state` {{optional_inline}}
+  - : Contains any state information you want to pass along with the navigation, for example, for logging or tracking purposes. The state for the navigation can subsequently be retrieved via {{domxref("NavigationHistoryEntry.getState()")}}.
+- `history` {{optional_inline}}
+  - : An enumerated value that specifies how this redirect should be added to the navigation history. It can take one of the following values:
+    - `auto`
+      - : The default value, which lets the browser decide how to handle it. Usually the value used is `push`, but it will become `replace` if the redirect points to the same URL as the pre-navigation URL.
```
noamr: It can also be whatever was passed to `navigate()`. The main point of `auto` is that the `redirect` call doesn't change history handling.

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,156 @@ None (`undefined`).
...
+});
+```
+
+This pattern is simpler than the alternative of canceling the original navigation and starting a new one to the redirect location, because it avoids exposing the intermediate state. For example, only one `navigatesuccess` or `navigateerror` event fires, and if the navigation was triggered by a call to {{domxref("Navigation.navigate()")}}, the promise only fulfills once the redirect destination is reached.
+
+The `precommitHandler()` callback takes a `controller` object as an argument, which contains a `redirect()` method. The `redirect()` method takes two parameters — a string representing the URL to redirect to, and an options object containing two parameters:
+
+- `state` {{optional_inline}}
+  - : Contains any state information you want to pass along with the navigation, for example, for logging or tracking purposes. The state for the navigation can subsequently be retrieved via {{domxref("NavigationHistoryEntry.getState()")}}.
+- `history` {{optional_inline}}
```
noamr, line 119: Something about this: `redirect` can change the behavior between auto/push/replace, but it cannot turn a traverse navigation to a push/replace navigation or vice versa.

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: I've changed it to "...implement same-document (SPA) navigation behavior..." for now, but we can iterate on it further if required.

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: OK, I _think_ I know what you mean here. Is the following correct? (ignore the formatting)

---

- `auto`
  - : The default value, which lets the browser decide how to handle it:
    - If the original navigation occurred as a result of a {{domxref("Navigation.navigate()")}} call, the value will be whatever was specified in the `navigate()` call's [`history`](/en-US/docs/Web/API/Navigation/navigate#history) option.
    - Otherwise, the value used is usually `push`, but it will become `replace` if the redirect points to the same URL as the pre-navigation URL.

files/en-us/web/api/navigateevent/intercept/index.md
chrisdavidmills: Cool; I've added this info as a note at the bottom of the bit about `replace()`:

> [!NOTE]
> The `redirect()` method can can convert the history behavior between `auto`, `push`, and `replace`, but it cannot turn a `traverse` navigation into a `push`/`replace` navigation and vice versa.

chrisdavidmills:
@noamr thanks for the reviews. This one looks about ready to go on to the editorial review stage now, but feel free to add more comments if you think anything else still needs attention.

noamr:
> @noamr thanks for the reviews. This one looks about ready to go on to the editorial review stage now, but feel free to add more comments if you think anything else still needs attention.

Yes, technical LGTM from me!

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,161 @@ None (`undefined`).
 - `InvalidStateError` {{domxref("DOMException")}}
   - : Thrown if the current {{domxref("Document")}} is not yet active, or if the navigation has been cancelled.
 - `SecurityError` {{domxref("DOMException")}}
-  - : Thrown if the event was dispatched by a {{domxref("EventTarget.dispatchEvent", "dispatchEvent()")}} call, rather than the user agent, or if the navigation cannot be intercepted (i.e., {{domxref("NavigateEvent.canIntercept")}} is `false`).
+  - : Thrown if:
+    - The event was dispatched by a {{domxref("EventTarget.dispatchEvent", "dispatchEvent()")}} call, rather than the user agent.
+    - The navigation cannot be intercepted ({{domxref("NavigateEvent.canIntercept")}} is `false`).
+    - A `precommitHandler()` callback was provided on a non-cancelable event ({{domxref("Event.cancelable")}} is `false`).
```
sideshowbarker:
```suggestion
    - A `precommitHandler()` callback is provided on a non-cancelable event ({{domxref("Event.cancelable")}} is `false`).
```

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,161 @@ None (`undefined`).
...
+
+## Description
+
+The `intercept()` method is used to implement same-document (SPA) navigation behavior when a navigation occurs, for example when a link is clicked, a form is submitted, or a programmatic navigation is initiated (using {{domxref("History.pushState()")}}, {{domxref("Window.location")}}, etc.).
```
sideshowbarker:
```suggestion
The `intercept()` method is used to implement same-document (SPA) navigation behavior when a navigation occurs; for example, when a link is clicked, a form is submitted, or a programmatic navigation is initiated (using {{domxref("History.pushState()")}}, {{domxref("Window.location")}}, etc.).
```

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,161 @@ None (`undefined`).
...
+  }
+});
+```
+
+This pattern is simpler than the alternative of canceling the original navigation and starting a new one to the redirect location, because it avoids exposing the intermediate state. For example, only one `navigatesuccess` or `navigateerror` event fires, and if the navigation was triggered by a call to {{domxref("Navigation.navigate()")}}, the promise only fulfills once the redirect destination is reached.
+
+The `precommitHandler()` callback takes a `controller` object as an argument, which contains a `redirect()` method. The `redirect()` method takes two parameters — a string representing the URL to redirect to, and an options object containing two parameters:
+
+- `state` {{optional_inline}}
+  - : Contains any state information you want to pass along with the navigation, for example, for logging or tracking purposes. The state for the navigation can subsequently be retrieved via {{domxref("NavigationHistoryEntry.getState()")}}.
```
sideshowbarker:
```suggestion
  - : Contains any state information you want to pass along with the navigation; for example, for logging or tracking purposes. The state for the navigation can subsequently be retrieved via {{domxref("NavigationHistoryEntry.getState()")}}.
```

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,161 @@ None (`undefined`).
...
+    - `push`
+      - : Adds a new {{domxref("NavigationHistoryEntry")}} to the navigation history, and clears any available forward navigation (that is, if the user previously navigated to other locations, then used the back button to return back through the history before then initiating the navigation that caused the redirect).
+    - `replace`
+      - : Replaces the {{domxref("Navigation.currentEntry")}} with the `NavigationHistoryEntry`.
+
+> [!NOTE]
+> The `redirect()` method can can convert the history behavior between `auto`, `push`, and `replace`, but it cannot turn a `traverse` navigation into a `push`/`replace` navigation and vice versa.
+
+`precommitHandler()` generally handles any modifications to the navigation behavior that are required before the destination URL is actually displayed in the browser, cancelling or redirecting it somewhere else as required. Because `precommitHandler()` can be used to cancel navigations, they can only be run when the event's {{domxref("Event.cancelable")}} property is `true`. Calling `intercept()` with a `precommitHandler()` on a non-cancelable event results in a `SecurityError` being thrown.
```
sideshowbarker:
```suggestion
`precommitHandler()` generally handles any modifications to the navigation behavior that are required before the destination URL is actually displayed in the browser, cancelling or redirecting it somewhere else as required. Because `precommitHandler()` can be used to cancel navigations, it will only work as expected when the event's {{domxref("Event.cancelable")}} property is `true`. Calling `intercept()` with a `precommitHandler()` on a non-cancelable event results in a `SecurityError` being thrown.
```

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,161 @@ None (`undefined`).
...
+
+### Responding to navigation success or failure
+
+When the promises returned by the `intercept()` handler functions fulfill, the `Navigation` object's {{domxref("Navigation/navigatesuccess_event", "navigatesuccess")}} event fires, allowing you to run cleanup code after a successful navigation has completed. If they reject, meaning the navigation has failed, {{domxref("Navigation/navigateerror_event", "navigateerror")}} fires instead, allowing you to gracefully handle the failure case.
```
sideshowbarker:
```suggestion
When the promises returned by the `intercept()` handler functions fulfill, the `Navigation` object's {{domxref("Navigation/navigatesuccess_event", "navigatesuccess")}} event fires, allowing you to run cleanup code after a successful navigation has completed. If those promises reject, meaning the navigation has failed, {{domxref("Navigation/navigateerror_event", "navigateerror")}} fires instead, allowing you to gracefully handle the failure case.
```

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,161 @@ None (`undefined`).
...
+
+### Interaction between `precommitHandler()` and `handler()`
+
+Both `precommitHandler()` and `handler()` callbacks can be included inside the same `intercept()` call.
+
+1. First, the `precommitHandler()` handler runs.
+   - When the `precommitHandler()` promise fulfills, the navigation commits.
+   - If the `precommitHandler()` rejects, `navigateerror` fires, the `committed` and `finished` promises reject, and the navigation is cancelled.
+
+2. When the navigation commits, a new {{domxref("NavigationHistoryEntry")}} is created for the navigation, and its `committed` promise fulfills.
```
sideshowbarker, line 149:
```suggestion
2. When the navigation commits, a new {{domxref("NavigationHistoryEntry")}} is created for the navigation, and its `committed` promise fulfills.

```
(added a line of vertical space before the next numbered item)

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,161 @@ None (`undefined`).
...
+3. Next, the `handler()` promise runs.
+   - When the `handler()` promise fulfills and the `navigatesuccess` event fires, the navigation `finished` promise fulfills as well, to indicate the navigation is finished.
+   - If `handler()` rejects, `navigateerror` fires, the `finished` promise rejects, and the navigation is canceled.
+
+Note that the above process is upheld even across multiple `intercept()` calls on the same `NavigateEvent`. All `precommitHandler()` callbacks are called first, and when all of them resolve, the navigation commits, and all the `handler()` callbacks are called.
+
+### Controlling focus behavior
+
+By default, after a navigation handled using `intercept()` has occurred, the document focus will reset to the first element in the DOM with an [`autofocus`](/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) attribute set, or the {{htmlelement("body")}} element, if no `autofocus` attribute is set. If you want to override this behavior, to manually implement a more accessible focus position on navigation (for example, the new top-level heading), you can do so by setting the `focusReset` option to `manual`.
```
sideshowbarker:
```suggestion
By default, after a navigation handled using `intercept()` has occurred, the document focus will reset to the first element in the DOM with an [`autofocus`](/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) attribute set, or otherwise to the {{htmlelement("body")}} element, if no `autofocus` attribute is set. If you want to override this behavior, to manually implement a more accessible focus position on navigation (for example, the new top-level heading), you can do so by setting the `focusReset` option to `manual`.
```

files/en-us/web/api/navigateevent/intercept/index.md
```diff
@@ -48,7 +50,161 @@ None (`undefined`).
...
+  }
+});
+```
+
+### Controlling scroll behavior
+
+After an `intercept()` navigation occurs, the following scrolling behavior occurs:
+
+- For `push` and `replace` navigations (see {{domxref("Navigation.navigate()")}}), the browser will attempt to scroll to the fragment given by `event.destination.url`. If there is no fragment available, it will reset the scroll position to the top of the page.
+- For {{domxref("Navigation.traverseTo", "traverse")}} and {{domxref("Navigation.reload", "reload")}} navigations, the browser behaves similarly to the previous bullet description, but delays its scroll restoration logic until the `intercept()` promise fulfills. It will perform no scroll restoration if the promise rejects. If the user has scrolled during the transition then no scroll restoration will be performed.
```
sideshowbarker:
```suggestion
- For {{domxref("Navigation.traverseTo", "traverse")}} and {{domxref("Navigation.reload", "reload")}} navigations, the browser behaves similarly to the description in the previous item above in this list, but delays its scroll restoration logic until the `intercept()` promise fulfills. It will perform no scroll restoration if the promise rejects. If the user has scrolled during the transition then no scroll restoration will be performed.
```

files/en-us/web/api/navigation_api/index.md
```diff
@@ -28,7 +28,9 @@ The `navigation` interface has several associated events, the most notable being
 
 The `NavigationEvent` object also provides two methods:
 
-- {{domxref("NavigateEvent.intercept", "intercept()")}} takes as an argument a callback handler function returning a promise. It allows you to control what happens when the navigation is initiated. For example, in the case of an SPA, it can be used to load relevant new content into the UI based on the path of the URL navigated to.
+- {{domxref("NavigateEvent.intercept", "intercept()")}} allows you to specify custom behavior for navigations, and can take the following as arguments:
+  - Callback handler functions allowing you to specify what happens both _when_ the navigation is committed and _just before_ the navigation is committed. For example, it could be used to load relevant new content into the UI based on the path of the URL navigated to, or redirect the browser to a sign-in page if the URL points to a restricted page and the user is not signed in.
```
sideshowbarker:
```suggestion
  - Callback handler functions allowing you to specify what happens both _when_ the navigation is committed and _just before_ the navigation is committed. For example, you could load relevant new content into the UI based on the path of the URL navigated to, or redirect the browser to a sign-in page if the URL points to a restricted page and the user is not signed in.
```

sideshowbarker (approved):

chrisdavidmills:
Looking good; thanks for the review, @sideshowbarker!

PR #41526, opened 2025-10-14
Reference "origin" glossary entry in `X-Frame-Options` header reference

Viicos:
This is referenced in the `'self'` CSP directive (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#self), so figured it wouldn't hurt to have it here as well.

### DIFF
--- a/files/en-us/web/http/reference/headers/x-frame-options/index.md
+++ b/files/en-us/web/http/reference/headers/x-frame-options/index.md
@@ -39,7 +39,7 @@ X-Frame-Options: SAMEORIGIN
 - `DENY`
   - : The page cannot be displayed in a frame, regardless of the site attempting to do so. Not only will the browser attempt to load the page in a frame fail when loaded from other sites, attempts to do so will fail when loaded from the same site.
 - `SAMEORIGIN`
-  - : The page can only be displayed if all ancestor frames are same {{glossary("origin")}} to the page itself. You can still use the page in a frame as long as the site including it in a frame is the same as the one serving the page.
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/reference/headers/x-frame-options/index.md
```diff
@@ -39,7 +39,7 @@ X-Frame-Options: SAMEORIGIN
 - `DENY`
   - : The page cannot be displayed in a frame, regardless of the site attempting to do so. Not only will the browser attempt to load the page in a frame fail when loaded from other sites, attempts to do so will fail when loaded from the same site.
 - `SAMEORIGIN`
-  - : The page can only be displayed if all ancestor frames are same origin to the page itself. You can still use the page in a frame as long as the site including it in a frame is the same as the one serving the page.
+  - : The page can only be displayed if all ancestor frames are same {{glossary("origin")}} to the page itself. You can still use the page in a frame as long as the site including it in a frame is the same as the one serving the page.
```
wbamberg:
```suggestion
  - : The page can only be displayed if all ancestor frames have the same {{glossary("origin")}} as the page itself. You can still use the page in a frame as long as the site including it in a frame is the same as the one serving the page.
```

(grammar like "are same origin to..." suggest we might prefer to make it "are same-origin to...", and link "same-origin" to https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy. But let's keep the link to https://developer.mozilla.org/en-US/docs/Glossary/Origin).

wbamberg (commented):
Thank you!

wbamberg (approved):
👍 thank you!

PR #41527, opened 2025-10-14
Make challenge solution comments more consistent, remove numbering

chrisdavidmills:
### Description

As discussed with @hamishwillee, I wanted to update the comments in the Learn area challenge solutions to:

- Be more consistent
- Use the same or similar wording to the project brief headings and tasks so that it is more intuitive which solution relates to which challenge
- Remove the numbers, as most of the time they are a red herring and are not related to any numbered steps in the challenges.

This PR aims to fix these issues for the current challenges I've recently updated.


hamishwillee (approved):
Much better way of doing things. Thank you.

Issue #37701, opened 2025-01-17
Document Signed HTTP exchanges


Josh-Cena:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

### What information was incorrect, unhelpful, or incomplete?

Follow up from: https://github.com/mdn/content/issues/1458

Need pages for the following:

- `Accept-Signature`
- `Signed-Headers`
- `Signature`

### What did you expect to see?

New pages

### Do you have any supporting links, references, or citations?

Also see Brian's resources:

> For signed exchanges, some info here:
> 
> - https://github.com/WICG/webpackage?tab=readme-ov-file#specifications
> - Blogs:
>   - https://web.dev/articles/signed-exchanges
>   - https://developers.google.com/search/docs/appearance/signed-exchange
>   - https://developers.cloudflare.com/speed/optimization/other/signed-exchanges/

aamuley:
Hey, I came across this issue and I was wondering why the headers point towards a WICG draft instead of RFC9421, would you be in favor of switching out the links to point to the RFC? 

cc @jyasskin, if https://github.com/WICG/webpackage/issues/713 hasn't reached a resolution, is the WICG still working on/maintaining an HTTP Signature protocol? [The draft](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html) is expired so I assumed not.

Josh-Cena:
Yes, it looks like we should really be documenting RFC 9421 instead, at least for the purpose of the HTTP reference. SXG might also need documentation but that would be a separate topic, and it's an entirely separate protocol in its own right, so I'm not sure where to put it.

PR #41530, opened 2025-10-14
update signature headers to point to rfc9421

aamuley:
### Description

Update the `Signature`, `Accept-Signature`, and `Signature-Input` headers to point to the RFC9421 rather than a WICG draft. Also `Signed-Headers` seems to to not be used, so changing in favor of `Signature-Input` which is defined. Based on [this issue](https://github.com/WICG/webpackage/issues/713#issue-1142375682), the WICG http message signatures seems somewhat obsoleted? 

### Motivation

Points to the adopted RFC rather than draft documents

### Additional details

n/a

(https://github.com/WICG/webpackage/issues/713)
(https://github.com/mdn/content/issues/37701)


Josh-Cena:
@aamuley Is this ready for review?

Josh-Cena (approved):
I think it's better that we get this in. You are right that the WICG version is confusing because it did not ship anywhere, but Chrome recently shipped RFC 9421. I moved them to a separate section so I can more conveniently add a note explaining the situation.

Josh-Cena:
Thanks for sending this in!

PR #41547, opened 2025-10-15
Make Defense summary checklist always an H2

wbamberg:
The "attacks" pages are not consistent about which level to put the "Defense summary checklist" at. I think probably H2 is better so it's accessible from the ToC.

I also added one for clickjacking, as we didn't have one.

### DIFF
--- a/files/en-us/web/security/attacks/clickjacking/index.md
+++ b/files/en-us/web/security/attacks/clickjacking/index.md
@@ -63,6 +63,16 @@ If `frame-ancestors` and `X-Frame-Options` are both set, then browsers that supp
 
 As an additional partial mitigation, sites should set the [`SameSite`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) cookie attribute for session cookies to `Lax` or `Strict`. Requests from embedded contexts such as `<iframe>` elements that are not {{glossary("Site", "same-site")}} with the top-level document will not include these cookies, and the server will therefore not treat the request as coming from a logged-in user.
 
+## Defense summary checklist
+
+As a primary defense:
+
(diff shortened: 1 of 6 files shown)


chrisdavidmills (approved):
@wbamberg the heading level change makes sense to me — good to be consistent, and nice to see these sections in the TOC.

I'm going to approve this, as the next comments are out of scope for this PR, and you might action them in a follow-up PR if you want to, but:

- The style of these sections is inconsistent. The bulleted lists seem good, but some of them have a lead-up of one or more paragraphs (more of a traditional summary), and some do not.
- I'm also not sure what "primary defense" and "defense in depth" mean. Is this like "mandatory, definitely do this" and "optional if you want to go further"?
- You can probably remove the "We can summarize the defenses above as follows:" lines, as the section heading says "summary".

wbamberg:
> @wbamberg the heading level change makes sense to me — good to be consistent, and nice to see these sections in the TOC.
> 
> I'm going to approve this, as the next comments are out of scope for this PR, and you might action them in a follow-up PR if you want to, but:

I think these are legit comments and have applied most of them.

> * The style of these sections is inconsistent. The bulleted lists seem good, but some of them have a lead-up of one or more paragraphs (more of a traditional summary), and some do not.

The only one AFAICS that is still different is XS-leaks, and that's tricky because unlike the others it's not a single attack with a single set of defenses, but multiple attacks with multiple overlapping defenses, including some attacks with no real defenses. So I don't really see a good alternative to explaining that.

> * I'm also not sure what "primary defense" and "defense in depth" mean. Is this like "mandatory, definitely do this" and "optional if you want to go further"?

Fair enough. Really I think the message is "do both".

> * You can probably remove the "We can summarize the defenses above as follows:" lines, as the section heading says "summary".

Done, and agreed it is unnecessary.

chrisdavidmills:
All makes sense, cheers @wbamberg! Nice work.

PR #41548, opened 2025-10-15
Use shared asset link in using_object-view-box content

OnkarRuikar:
Use an image from the shared-assets repo.


estelle (approved):
Thanks! 🎉

PR #41556, opened 2025-10-16
Fix Glossary/CIA Wikipedia link

JackDunnCodes:
Headers have changed, so link didn't quite lead to the right place

### Description

Just a very small fix I spotted. Change the header that the link linked to, as the one it led to right now doesn't exist - but this seems to be where it's meant to go

### Motivation

Improves usability of documentation/further reading links

### Additional details

N/A

None known to me


chrisdavidmills (approved):
LGTM; thanks for the fix, @JackDunnCodes!

PR #41558, opened 2025-10-16
Technical review: Add details and example for null bindGroupLayouts entries

chrisdavidmills:
### Description

As of Chrome 135, [`GPUDevice.createPipelineLayout`](https://developer.mozilla.org/en-US/docs/Web/API/GPUDevice/createPipelineLayout) supports `null` entries in the `bindGroupLayouts` array. See https://developer.chrome.com/blog/new-in-webgpu-135#allow_creating_pipeline_layout_with_null_bind_group_layout for details.

This PR adds details for this feature, plus a quick example snippet.

### DIFF
--- a/files/en-us/web/api/gpudevice/createpipelinelayout/index.md
+++ b/files/en-us/web/api/gpudevice/createpipelinelayout/index.md
@@ -22,7 +22,9 @@ createPipelineLayout(descriptor)
 - `descriptor`
   - : An object containing the following properties:
     - `bindGroupLayouts`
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
cc @beaufortfrancois

beaufortfrancois (commented):
LGTM

files/en-us/web/api/gpudevice/createpipelinelayout/index.md
```diff
@@ -79,6 +81,21 @@ const pipelineLayout = device.createPipelineLayout({
 // …
 ```
 
+### Specifying an empty bind group layout
+
+In this snippet, we create three bind group layouts, with bind group layout 1 representing fragment data and bind group layout 2 representing vertex data. If we want to create a pipeline that uses only bind group layouts 0 and 2, we can pass `null` for bind group layout 1 and then render without a fragment shader.
+
+```js
+const bgl0 = myDevice.createBindGroupLayout({ entries: myGlobalEntries });
```
beaufortfrancois: Nit: maybe use `device` to match the example above

chrisdavidmills: Good call, done.

pepelsbey (approved):
Looks good! Thank you :)

PR #41559, opened 2025-10-16
Fix Kitchensink live example

caugner:
### Description

Fixes unexpected encoding in a Kitchensink live sample.

### Motivation

Resolves a console error:

> SyntaxError: private names aren't valid in this context

Noticed in https://github.com/mdn/fred/pull/922.


LeoMcA (approved):
I'm relieved this isn't an escaping error in play runner!

PR #41561, opened 2025-10-16
Fix inheritance of Crypto interface

Zenkoina:
### Description

Removes mention of the RandomSource interface.

The last mention of the RandomSource interface I was able to find is in the [Last Call Working Draft of the Specification](https://www.w3.org/TR/2014/WD-WebCryptoAPI-20140325/) which is from 25 March 2014, and it wasn't included in the [Candidate Recommendation Snapshot](https://www.w3.org/TR/2014/CR-WebCryptoAPI-20141211/) from 11 December 2014 or any future versions of the specification.

### Motivation

Keeping the documentation up to date.

### DIFF
--- a/files/en-us/web/api/crypto/index.md
+++ b/files/en-us/web/api/crypto/index.md
@@ -14,14 +14,10 @@ The `Crypto` is available in windows using the {{domxref("Window.crypto")}} prop
 
 ## Instance properties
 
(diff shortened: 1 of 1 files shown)


wbamberg (approved):
👍 thanks for digging through the old specs so we don't have to :)

PR #41577, opened 2025-10-17
Add browser support considerations to code guidelines

pepelsbey:
### Description

Adds browser support considerations to code guidelines:

- Use feature detection
- Don’t specify supporting browsers

### Motivation

To improve maintenance.


hamishwillee (approved):
Good one

PR #41579, opened 2025-10-17
Fix auto ID for 3d specific CSS properties heading

nmleuz:
Added a custom ID to the 3d specific CSS properties heading to fix the issue with auto-generated ID stripping the leading number
Related to mdn/fred#1007


marjorieazoulay04-cyber (commented):
https://github.com/mdn/content/pull/41579

Josh-Cena:
This is not a syntax recognized by our Markdown infra, and this is a platform bug, not a content bug.

PR #41583, opened 2025-10-18
Update SVG attributes page

estelle:
added some text and links. deduped some events


files/en-us/web/svg/reference/attribute/index.md
```diff
@@ -413,73 +421,80 @@ Below is a list of all of the attributes available in SVG along with links to re
...
-- **`onshow`**
-- **`onstalled`**
-- **`onsubmit`**
-- **`onsuspend`**
-- **`ontimeupdate`**
-- **`ontoggle`**
-- **`onunload`**
-- **`onvolumechange`**
-- **`onwaiting`**
+ The event handler content attributes, prefixed with the `on` are supported by all {{domxref("Document")}} and {{domxref("Window")}} objects, as event handler event types:
```
Josh-Cena: This is in direct contradiction with this page which is about attributes on SVG elements. Perhaps you want to adopt the verbiage from https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes#list_of_global_event_handler_attributes instead.

estelle: The wording comes from the specs, but I agree with not including it. Would you be ok with removing all of them and linking to the page instead? The original version, without links, wasn't useful

files/en-us/web/svg/reference/attribute/index.md
```diff
@@ -413,73 +421,80 @@ Below is a list of all of the attributes available in SVG along with links to re
...
-- **`onsubmit`**
-- **`onsuspend`**
-- **`ontimeupdate`**
-- **`ontoggle`**
-- **`onunload`**
-- **`onvolumechange`**
-- **`onwaiting`**
+Event handler attributes are discouraged. While the attributes listed below apply to all HTML and SVG elements, they are not useful on all elements. For example, the `oncuechange` attribute is accepted by all elements and attaches an event listener for `cuechange`, but only {{htmlelement("track")}} elements will ever receive a `cuechange` event fired by the browser. For other elements, you can only use {{domxref("EventTarget.dispatchEvent()")}} to manually dispatch one. [Some attributes](/en-US/docs/Web/HTML/Reference/Elements/body#event_attributes) can be specified on {{HTMLElement("svg")}}, but they would instead listen to events on `window`.
+
+- {{domxref("HTMLMediaElement/abort_event", "onabort")}}
```
Josh-Cena: Where does this list come from? I checked https://svgwg.org/svg2-draft/attindex.html, and they don't match (granted, the previous list don't match perfectly either). Furthermore not all of these attributes are global.

This doesn't look right:

> [Some attributes](/en-US/docs/Web/HTML/Reference/Elements/body#event_attributes) can be specified on {{HTMLElement("svg")}}, but they would instead listen to events on `window`.

I don't think SVG has the same window-reflecting event attribute concept like HTML. Even when the SVG is opened as a separate document, I'm not sure if this can happen.

estelle: I updated the list that was there and created links out of them, since the list was not helpful as it was, plus it had duplicates. I am all for getting rid of the whole list.

The "doesn't look right" part:  the `<SVG>` element, which is the HTML `<svg>` element.

estelle: I reverted the section, so we can merge the rest (if approved), as that was the important part. The rest was just to be hopefully helpful, but not required, so reverted. See #41602 which we can edit, merge or close.

files/en-us/web/svg/reference/attribute/index.md
```diff
@@ -291,37 +291,44 @@ Below is a list of all of the attributes available in SVG along with links to re
...
+The XLink attributes can reference resources.
+
+- {{SVGAttr("xlink:href")}} {{deprecated_inline}}
 - {{SVGAttr("xlink:type")}}
 - {{SVGAttr("xlink:role")}}
 - {{SVGAttr("xlink:arcrole")}}
-- {{SVGAttr("xlink:title")}}
+- {{SVGAttr("xlink:title")}} {{deprecated_inline}}
 - {{SVGAttr("xlink:show")}}
 - {{SVGAttr("xlink:actuate")}}
```
Josh-Cena: All of these should be deprecated

files/en-us/web/svg/reference/attribute/index.md
```diff
@@ -291,37 +291,44 @@ Below is a list of all of the attributes available in SVG along with links to re
...
 
-### Generic attributes
-
-- Core attributes
-  - {{SVGAttr("id")}}
-  - {{SVGAttr("class")}}
-  - {{SVGAttr("style")}}
-  - {{SVGAttr("lang")}}
-  - {{SVGAttr("tabindex")}}
-  - {{SVGAttr("autofocus")}}
```
Josh-Cena, line 304: You need to restore `autofocus` in the new list

Josh-Cena (approved):
Two things, but mergeable afterwards

files/en-us/web/svg/reference/attribute/index.md
```diff
@@ -291,37 +291,44 @@ Below is a list of all of the attributes available in SVG along with links to re
...
-
-- Conditional processing attributes
-  - {{SVGAttr("requiredExtensions")}}
-  - {{SVGAttr("requiredFeatures")}}
-  - {{SVGAttr("systemLanguage")}}
+### Core attributes
+
+The core attributes are global attributes.
+
+- {{SVGAttr("id")}}
```
estelle, line 301:
```suggestion
- {{SVGAttr("autofocus")}}
- {{SVGAttr("id")}}
```

PR #41586, opened 2025-10-18
SVG tutorial introd SVG 2.0 history

estelle:
Updated the SVG tutorial introduction to include details about SVG 2.0 and its components.


hamishwillee (approved):
@estelle Looks great. 

Approving, but FMI how does a "full" SVG version differ from "the current standard"?

estelle:
> Approving, but FMI how does a "full" SVG version differ from "the current standard"?

My interpretation was that "full" meant "recommendation", because the "current standard" is still at "candidate recommendation".

hamishwillee:
> > Approving, but FMI how does a "full" SVG version differ from "the current standard"?
> 
> My interpretation was that "full" meant "recommendation", because the "current standard" is still at "candidate recommendation".

Thanks very much. Not your fault or problem, but it would be good if we could tidy that up since it's a pretty unclear recommendation about what people should target.

PR #41598, opened 2025-10-20
Add await to getFileHandle call in example

eight04:
### Description

Add `await` to `getFileHandle` call in example.

### Motivation

`getFileHandle` returns a promise.


Josh-Cena (approved):
Sure, thank you

PR #41600, opened 2025-10-20
Editorial review: Update property info to mention GPUTexture as shortcut for default GPUTextureView

chrisdavidmills:
### Description

In Chrome 140+, there are multiple places in WebGPU methods where a `GPUTextureView` was expected, but now you can specify a `GPUTexture` as a shorthand, provided a default `GPUTextureView` is desired (it is equivalent to a `GPUTextureView` created using a call to `GPUTexture.createView()` with no descriptor specified).

See https://developer.chrome.com/blog/new-in-webgpu-140#shorthand_for_using_texture_where_texture_view_is_used.

This PR updates the content to mention this available alternative value in the appropriate places.


chrisdavidmills:
cc @beaufortfrancois

files/en-us/web/api/gpucommandencoder/beginrenderpass/index.md
```diff
@@ -70,9 +70,13 @@ Color attachment objects can have the following properties:
     - `"discard"`: Discards the resulting value of the render pass for this attachment.
     - `"store"`: Stores the resulting value of the render pass for this attachment.
 - `resolveTarget` {{optional_inline}}
-  - : A {{domxref("GPUTextureView")}} object representing the texture subresource that will receive the resolved output for this color attachment if `view` is multisampled.
+  - : An object representing the texture subresource that will receive the resolved output for this color attachment if `view` is multisampled. This can be one of the following:
+    - {{domxref("GPUTextureView")}}
+    - {{domxref("GPUTexture")}}: Can be used in place of a `GPUTextureView`, provided a default view is desired. When used in this context, `GPUTexture` is equivalent to a `GPUTextureView` object created using a {{domxref("GPUTexture.createView()")}} call with no `descriptor` object specified.
```
beaufortfrancois: Since it's not clear where `descriptor` comes from, we should add it there and in other places
 ```suggestion
     - {{domxref("GPUTexture")}}: Can be used in place of a `GPUTextureView`, provided a default view is desired. When used in this context, `GPUTexture` is equivalent to a `GPUTextureView` object created using a {{domxref("GPUTexture.createView(descriptor)")}} call with no `descriptor` object specified.
```

**Review by beaufortfrancois — COMMENTED** [review]
LGTM with nit

File: files/en-us/web/api/gpucommandencoder/beginrenderpass/index.md (hunk 1 above)
Comment (chrisdavidmills) [reply files/en-us/web/api/gpucommandencoder/beginrenderpass/index.md]: I'm not sure I agree with that. If you don't know what it means, you can click the link and look at the reference page to find out. We don't tend to write links to method ref pages with arguments specified, even if they have mandatory arguments.

Comment (beaufortfrancois) [reply files/en-us/web/api/gpucommandencoder/beginrenderpass/index.md]: Maybe replacing `no descriptor object` with `no parameter` then?

```sugggestion
    - {{domxref("GPUTexture")}}: Can be used in place of a `GPUTextureView`, provided a default view is desired. When used in this context, `GPUTexture` is equivalent to a `GPUTextureView` object created using a {{domxref("GPUTexture.createView()")}} call with no parameter specified.
```

chrisdavidmills: Fair, that is more intuitive. I've made the change, although I've used "argument" in place of "parameter". I think this is the more correct term when you are calling the function as opposed to defining it, although most people tend to use them interchangeably.

beaufortfrancois: LGTM

pepelsbey (approved):
Looks good! Thank you :)

PR #41615, opened 2025-10-21
Fix comment to use correct `line-height`

kovsu:
Replaced "line height" with "line-height"


chrisdavidmills (approved):
LGTM; thanks for the fix, @kovsu!

PR #41619, opened 2025-10-21
Fix 'pager' spelling in autocomplete attributes

tscpp:
### Description

There's a mismatch between MDN and the HTML living standard on the recipient token used for pagers and beepers in the autocomplete attribute. I fixed the spelling from "page" to "pager".

### Motivation

The HTML specification states:

> "pager", meaning the field describes a pager's or beeper's contact details

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-pager


Josh-Cena (approved):
Nice catch, thanks

PR #41621, opened 2025-10-22
Link to individual transform properties

### LINKED ISSUE #41599: Emphasize individual transform properties (`translate`, `scale`, and `rotate`)
Or a reference way to the top. As `scale: n m` is equivalent to `transform: scale(n m)` but way shorter and easier understandable it may want to be pushed more.

Josh-Cena:
Fix https://github.com/mdn/content/issues/41599


estelle (approved):
Thanks! 🎉

PR #41624, opened 2025-10-22
fix: Add perms to calling workflow

bsmth:
### Description

Adding permissions to add or remove labels from PRs. I think the fix should be here and not in the called workflow at https://github.com/mdn/workflows/blob/main/.github/workflows/pr-rebase-needed.yml

### Motivation

This workflow is passing when no action needed and failing when it needs to add or remove a label: https://github.com/mdn/content/actions/workflows/pr-rebase-needed.yml

### DIFF
--- a/.github/workflows/pr-rebase-needed.yml
+++ b/.github/workflows/pr-rebase-needed.yml
@@ -8,6 +8,8 @@ on:
 permissions:
   # Label pull requests.
   pull-requests: write
+  contents: read
+  issues: write
 
 concurrency:
   group: pr-rebase-needed-${{ github.event.pull_request.number || github.ref_name }}


bsmth:
See also https://github.com/mdn/workflows/pull/171

.github/workflows/pr-rebase-needed.yml
```diff
@@ -8,6 +8,8 @@ on:
 permissions:
   # Label pull requests.
   pull-requests: write
+  contents: read
```
caugner, line 11: Afaik `contents: read` is only needed in private repos.

.github/workflows/pr-rebase-needed.yml
```diff
@@ -8,6 +8,8 @@ on:
...
+  issues: write
```
caugner, line 12: IIRC there was a time when PR label changes required `issues: write`, but I don't think this is still the case.

caugner (commented):
The [upstream changes](https://github.com/mdn/workflows/pull/171) should have already resolved the issue.

https://github.com/mdn/content/actions/workflows/pr-rebase-needed.yml

bsmth:
Looks good, thanks, closing now. 🙏🏻

PR #41626, opened 2025-10-22
Add example to 'inert' attribute

estelle:
Added example of using the 'inert' attribute in HTML. 
Remove links to external articles


chrisdavidmills (approved):
LGTM. Thanks @estelle!

PR #41627, opened 2025-10-22
fix broken text-fragment for example.com

k-yle:
### Description

[example.{com,org,net}](https://example.com) have recently had their first update [in over 10 years](https://web.archive.org/web/20140228235225/http://example.com/). The wording on the page has changed: the phrase `asking for` no longer exists, which broke one of the examples for [text fragments](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments).

### Motivation

it's broken. which is a bit ironic given [the warning a few lines above](https://github.com/k-yle/content/blob/8109821f5642d9200b376a0f974f9176af22997b/files/en-us/web/uri/reference/fragment/text_fragments/index.md?plain=1#L33) 😆

### Additional details

&shy;-

&shy;-

### DIFF
--- a/files/en-us/web/uri/reference/fragment/text_fragments/index.md
+++ b/files/en-us/web/uri/reference/fragment/text_fragments/index.md
@@ -76,7 +76,7 @@ Supporting browsers will scroll to and highlight the first text fragment in the 
 
 ### Text fragment with textStart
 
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
@k-yle LGTM. Good spot, and thanks for the fix!

PR #41628, opened 2025-10-22
Add missing words to incomplete sentence

ghost:
### Description

Original sentence incomplete; added the two missing words that were very likely intended by the original author.

### Motivation

To improve the reader's comprehension of the sentence.


Josh-Cena:
Hello, we use "throw" as an intransitive verb, which is short for "throw an error".

ghost:
Thanks for the explanation Josh!
I'd never come across this usage before.

PR #41629, opened 2025-10-22
Clarify highWaterMark description

cdoublev:
### Description

The current text defines the `highWaterMark` property of the `queuingStrategy` argument object of `ReadableStream.constructor(underlyingSource, queuingStrategy)`, as a *"total number of chunks"* while it is specified as a *"total size of chunks"* in the WHATWG Stream specification.

### Motivation

I make this change because I am learning about streams and this material is incorrect.

### Additional details

https://streams.spec.whatwg.org/#high-water-mark


wbamberg (approved):
👍 thank you!

PR #41630, opened 2025-10-22
Link to definition of default "XSS-safe" elements and attributes

untitaker:
### Description

Link to the specification to define what "XSS-safe" means.

"XSS-safe" is not well-defined within MDN so let's link to the spec.

### Motivation

The current description of the default sanitizer's behavior is ultimately not meaningful.

### Additional details

I don't know if "XSS-safe" is a well-defined term within the spec, but if it is then the definition should be linked in MDN (or translated to it)

### DIFF
--- a/files/en-us/web/api/element/sethtml/index.md
+++ b/files/en-us/web/api/element/sethtml/index.md
@@ -50,7 +50,7 @@ The **`setHTML()`** method provides an XSS-safe method to parse and sanitize a s
 It then removes any HTML entities that aren't allowed by the sanitizer configuration, and further removes any XSS-unsafe elements or attributes — whether or not they are allowed by the sanitizer configuration.
 
 If no sanitizer configuration is specified in the `options.sanitizer` parameter, `setHTML()` is used with the default {{domxref("Sanitizer")}} configuration.
-This configuration allows all elements and attributes that are considered XSS-safe, thereby disallowing entities that are considered unsafe.
+This configuration allows all elements and attributes that are [considered XSS-safe](https://wicg.github.io/sanitizer-api/#built-in-safe-default-configuration), thereby disallowing entities that are considered unsafe.
 A custom sanitizer or sanitizer configuration can be specified to choose which elements, attributes, and comments are allowed or removed.
 Note that even if unsafe options are allowed by the sanitizer configuration, they will still be removed when using this method (which implicitly calls {{domxref('Sanitizer.removeUnsafe()')}}).
 


evilpie:
I wish we had a better explanation for this that we could link to. Maybe this something https://github.com/WICG/sanitizer-api/issues/287 should aspire to define.
Actually https://wicg.github.io/sanitizer-api/#built-in-safe-baseline-configuration might be better. That is actually the list that is used by removeUnsafe, to remove everything that is script-ty.

wbamberg:
We should explain this but I don't think this link is very helpful. Apart from being quite hard to read, I'm not sure what it means. For instance when it says:

```json
{
  "name": "abbr",
  "namespace": "http://www.w3.org/1999/xhtml",
  "attributes": []
}
```

...does that mean that by default no attributes are allowed on `<abbr>`?

It would be great to have a page under https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API that describes the default configuration. I also don't understand the difference between the "built-in safe default configuration" and the "built-in safe baseline configuration" and would prefer not to have to pick through the spec to try to work it out.

@hamishwillee , what do you think?

hamishwillee:
> We should explain this but I don't think this link is very helpful. Apart from being quite hard to read, I'm not sure what it means. For instance when it says:
> 
> ```json
> {
>   "name": "abbr",
>   "namespace": "http://www.w3.org/1999/xhtml",
>   "attributes": []
> }
> ```
> 
> ...does that mean that by default no attributes are allowed on `<abbr>`?
> 
> It would be great to have a page under https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API that describes the default configuration. I also don't understand the difference between the "built-in safe default configuration" and the "built-in safe baseline configuration" and would prefer not to have to pick through the spec to try to work it out.
> 
> @hamishwillee , what do you think?

I agree with you - generally we don't link to spec and I don't think this is all that useful a link.

The TLDR IMO is that 99% of people don't need to know the list - this should be used as a drop in replacement for innerHTML. If you're injecting any of those untrusted things you should be talking a long hard look at yourself :-)

That said, I guess it might help people who are wondering why something in the input disappeared.  Since the text is huge and may change we could do this:

> This configuration allows all elements and attributes that are considered XSS-safe, thereby disallowing entities that are considered unsafe (you can test what elements that are considered unsafe by running the [`Sanitizer()` constructor example](/en-US/docs/Web/API/Sanitizer/Sanitizer#results)).

----

As I understand it, the "built-in safe default configuration" is the very restrictive configuration you get by default. This excludes anything that can cause an XSS attack AND some other bits and bobs that can in theory lead to other less worrying attacks such as clickjacking.
The  "built-in safe baseline configuration" is a more permissive configuration that just restricts the XSS subset of attributes and elements.

I believe that removeUnsafe() is the only thing that uses the more permissive version. That is not documented, and possibly should be.

> > ...does that mean that by default no attributes are allowed on `<abbr>`?

It means that `<abbr>` is not allowed. If their were attributes it would mean that the element is allowed, but if those attributes are present they should be filtered.

untitaker:
I'm honestly surprised that such a trivial patch generates this much discussion and holdup.

I don't think that asking people to run the example to figure out how the web platform works is acceptable. You can take that approach with anything if you're unwilling to document it.

So far it still seems to me this part of the spec is the best document to link to, despite things like `"attributes": []` not being immediately clear. It can always be replaced with a better link once one is actually available.

Josh-Cena:
FYI, https://github.com/mdn/content/pull/41677 added a link to `Sanitizer()`, which does define the default options to some degree.

hamishwillee:
Thanks @untitaker - I think https://github.com/mdn/content/pull/41695 fixes this to the point that is useful for most developers. I know it is not a precise definition, but it is a useful and maintainable one. Closing.

PR #41636, opened 2025-10-22
Make CSS module titles uniform

estelle:
1. If only glossary terms, the header is "Glossary terms", otherwise "Glossary terms and definitions" or  "Glossary terms, definitions, and keywords" 
2. All at-rules and all descriptors now merged as "At-rules and descriptors"
3. All glossary terms and definitions start with an uppercase. Keywords don't


chrisdavidmills (approved):
LGTM

PR #41637, opened 2025-10-22
Add realistic reportError() use case

### LINKED ISSUE #35670: reportError could mention other error reporting use cases
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/reportError

The second paragraph starting "This feature is primarily intended for..."

> This feature is primarily intended for custom event-dispatching or callback-manipulating libraries. Libraries can use this feature to catch errors in callback code and re-throw them to the top level handler. This ensures that an exception in one callback will not prevent others from being handled, while at the same time ensuring that stack trace information is still readily available for debugging at the top level.

I think reportError is generally useful for reporting errors in application code, particularly if you have telemetry tools like Sentry installed. Or even without telemetry, it could be a reasonable coding style to use it for any case where an error occurs but you don't want to halt the current function. But this paragraph seems to nudge people away from using it unless you are writing a library or some low-level/framework code.

Either include a broader set of possible use cases, or don't prescribe use cases at all. I think it would be enough to merely document behavior.

_No response_

_No response_

Josh-Cena:
Fix https://github.com/mdn/content/issues/35670

### DIFF
--- a/files/en-us/web/api/window/reporterror/index.md
+++ b/files/en-us/web/api/window/reporterror/index.md
@@ -23,7 +23,7 @@ reportError(throwable)
 ### Parameters
 
 - `throwable`
-  - : An error object such as a {{jsxref("TypeError")}}.
+  - : Any JavaScript value, but preferably an error object such as a {{jsxref("TypeError")}}.
 
 ### Return value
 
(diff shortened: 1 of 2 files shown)


wbamberg (approved):
👍

PR #41644, opened 2025-10-22
Synchronize with BCD v7.1.15

OnkarRuikar:
[BCD release v7.1.15](https://github.com/mdn/browser-compat-data/releases/tag/v7.1.15)

/cc cc/ @bsmth @wbamberg

### DIFF
--- a/files/en-us/web/api/cookiestore/get/index.md
+++ b/files/en-us/web/api/cookiestore/get/index.md
@@ -42,28 +42,28 @@ A {{jsxref("Promise")}} that resolves with an object representing the first cook
 
 The object returned for a match contains the following properties:
 
-- `domain`
+- `domain` {{experimental_inline}} {{non-standard_inline}}
   - : A string containing the domain of the cookie.
 
-- `expires`
+- `expires` {{experimental_inline}} {{non-standard_inline}}
   - : A timestamp, given as {{glossary("Unix time")}} in milliseconds, containing the expiration date of the cookie.
 
-- `name`
+- `name` {{experimental_inline}} {{non-standard_inline}}
   - : A string containing the name of the cookie.
 
-- `partitioned`
+- `partitioned` {{experimental_inline}} {{non-standard_inline}}
   - : A boolean indicating whether the cookie is a partitioned cookie (`true`) or not (`false`). See [Cookies Having Independent Partitioned State (CHIPS)](/en-US/docs/Web/Privacy/Guides/Privacy_sandbox/Partitioned_cookies) for more information.
 
-- `path`
+- `path` {{experimental_inline}} {{non-standard_inline}}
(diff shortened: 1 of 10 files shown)


Josh-Cena:
This is very frustrating—yesterday I discovered some BCD status was not synced properly to content (a non-standard thing that was not marked, IIRC) but now I can't for the life of me recall what it is...

Josh-Cena:
Ah! Found it. @OnkarRuikar Is is safe to add `{{non-standard_inline}}` to the return value fields on https://developer.mozilla.org/en-US/docs/Web/API/CookieStore/get, or would your script insist to remove them?

hamishwillee (approved):
Possibly a little early for Integrity policy - it ships in 145 which isn't released yet. Approved so you can merge when needed.

OnkarRuikar:
> Is is safe to add {{non-standard_inline}} to the return value fields on [developer.mozilla.org/en-US/docs/Web/API/CookieStore/get](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore/get), or would your script insist to remove them?

I've added the statuses to the return values of `CookieStore.get()`, and the script won't revert them in subsequent runs.

The script doesn't touch the definition list items that are not in the BCD. In this case, the script doesn't know what `*_return_property` suffix means, so it thinks the definition list items are not covered by the BCD and it doesn't touch them.

I can cover the return values in the script, but it looks like there is no convention for how the returned values/properties are marked in BCD. I find different encodings :(
- [`domain_return_property`](https://github.com/mdn/browser-compat-data/blob/9f09d9d98784c722761f42d8e56f61ec71ebf1aa/api/CookieStore.json#L181)
- [`returns_CSSStyleProperties`](https://github.com/mdn/browser-compat-data/blob/9f09d9d98784c722761f42d8e56f61ec71ebf1aa/api/HTMLElement.json#L2550)
- [`returns_userSettings_isOnToolbar_property`](https://github.com/mdn/browser-compat-data/blob/9f09d9d98784c722761f42d8e56f61ec71ebf1aa/webextensions/api/action.json#L324)
- [`returns_promise`](https://github.com/mdn/browser-compat-data/blob/9f09d9d98784c722761f42d8e56f61ec71ebf1aa/docs/data-guidelines/api.md?plain=1#L167)
- [`return_object_property_sampleSize`](https://github.com/mdn/browser-compat-data/blob/9f09d9d98784c722761f42d8e56f61ec71ebf1aa/api/MediaDevices.json#L1035C10-L1035C43)
- ...

***

> Possibly a little early for Integrity policy - it ships in 145 which isn't released yet. Approved so you can merge when needed.

The content and the corresponding BCD data are both live, so there shouldn't be any issues merging this, right?

Josh-Cena (approved):
Thanks. No need to automate return values; just catching them as we go.

PR #41645, opened 2025-10-23
Use getter syntax for static observedAttributes

KonnorRogers:
So, after losing my marbles in a codepen:

https://codepen.io/paramagicdev/pen/xbZWabx?editors=1010

it turns out FF 144, Safari 17.6, and Chrome 141.0 do not support static class properties for `observedAttributes`.

I was trying to figure out why `attributeChangedCallback()` wasn't firing, and turns out its because the browser expect the getter and not a static property 🙃


KonnorRogers:
False Alarm. Thanks to @jaredcwhite its because static properties need to be assigned before a static setup block.

```js
  // Breaks
  static {}
  static observedAttributes = [
       "value"
  ]
  
  // Works
  static observedAttributes = [
       "value"
  ]
  static {}
```

PR #41647, opened 2025-10-23
Give an example of a property that do not accept percentage value

youssef-el-atmani:
### Motivation

To give an idea about a property that do not accept percentage value as width

### DIFF
--- a/files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
+++ b/files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
@@ -342,7 +342,7 @@ li {
 
 {{EmbedLiveSample("percentage-fonts")}}
 
-Note that, while many value types accept a length or a percentage, there are some that only accept length. You can see which values are accepted on the MDN property reference pages. If the allowed value includes {{cssxref("length-percentage")}} then you can use a length or a percentage. If the allowed value only includes `<length>`, it is not possible to use a percentage.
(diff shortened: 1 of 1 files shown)


files/en-us/learn_web_development/core/styling_basics/values_and_units/index.md
```diff
@@ -342,7 +342,7 @@ li {
 
 {{EmbedLiveSample("percentage-fonts")}}
 
-Note that, while many value types accept a length or a percentage, there are some that only accept length. You can see which values are accepted on the MDN property reference pages. If the allowed value includes {{cssxref("length-percentage")}} then you can use a length or a percentage. If the allowed value only includes `<length>`, it is not possible to use a percentage.
+Note that, while many value types accept a length or a percentage, there are some that only accept length; `border-width` is an example of such a property. You can see which values are accepted on the MDN property reference pages. If the allowed value includes {{cssxref("length-percentage")}} then you can use a length or a percentage. If the allowed value only includes `<length>`, it is not possible to use a percentage.
```
chrisdavidmills:
```suggestion
While many properties accept a length or a percentage as a value, some only accept a length, for example {{cssxref("border-width")}}. MDN's property reference pages detail which value types they accept. If the allowed value includes {{cssxref("length-percentage")}}, then you can use a length or a percentage. If the allowed value only includes `<length>`, it is not possible to use a percentage.
```

chrisdavidmills (changes requested):
@youssef-el-atmani this is a valid suggestion, for sure. Thanks for submitting it.

I have a suggestion about the wording, which I went slightly overboard on ;-)

chrisdavidmills (approved):
Perfect. Let's get this in. Thanks again @youssef-el-atmani!

PR #41648, opened 2025-10-23
Editorial review: Information about User-Agent reduction

chrisdavidmills:
### Description

Chromium browsers support [User-Agent reduction](https://privacysandbox.google.com/protections/user-agent), which means that features such as the `User-Agent` HTTP header provide less-specific information in an effort to reduce their effectiveness for fingerprinting/tracking.

This PR adds information on User-Agent reduction to appropriate places on MDN, as well as a new guide to cover it, and some extra information on client-side hints.

This PR is part of the final set of privacy sandbox information to be added to MDN. Much of it is being removed, but this feature is stable and will be kept, so it should be added.


files/en-us/web/api/navigator/useragent/index.md
```diff
@@ -8,46 +8,37 @@ browser-compat: api.Navigator.userAgent
...
 
 ## Examples
 
 ```js
-alert(window.navigator.userAgent);
-// alerts "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:124.0) Gecko/20100101 Firefox/124.0"
+console.log(navigator.userAgent);
+// On Chrome, logs something like "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36" (reduced UA string)
+
+// On Firefox, logs something like "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:124.0) Gecko/20100101 Firefox/124.0"
```
miketaylr: nit: these are two different platforms, so maybe write "On Chrome on macOS" and "On Firefox on Windows"

files/en-us/web/http/guides/client_hints/index.md
```diff
@@ -13,7 +13,22 @@ The set of "hint" headers are listed in the topic [HTTP Headers](/en-US/docs/Web
 
 ## Overview
 
-A server must announce that it supports client hints, using the {{HTTPHeader("Accept-CH")}} header to specify the hints that it is interested in receiving.
+When the browser first makes a request to load a webpage, it will send a {{httpheader("User-Agent")}} header and a default set of `Sec-CH-UA-*` headers along with the initial request. An Android device, for example, might send the following:
+
+```http
+User-Agent: Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Mobile Safari/537.36
+Sec-CH-UA: "Chrome"; v="143"
```
miketaylr: `Sec-CH-UA` is a bit more verbose. It would look like `"Google Chrome";v="143", "Chromium";v="143", "Not A(Brand";v="24"` (ye olde trusty echo.opera.com is useful here if you don't want to have to look at dev tools).

There's a few places in this PR that this should be updated, but I won't comment on any others.

files/en-us/web/http/guides/client_hints/index.md
```diff
@@ -13,7 +13,22 @@ The set of "hint" headers are listed in the topic [HTTP Headers](/en-US/docs/Web
...
+Sec-CH-UA-Mobile: ?1
+```
+
+These `CH-UA` headers provide the following information:
+
+- {{httpheader("Sec-CH-UA")}}: The major browser version.
+- {{httpheader("Sec-CH-UA-Platform")}}: The platform.
+- {{httpheader("Sec-CH-UA-Mobile")}}: A boolean that indicates whether the browser is on a mobile device (`?1`) or not (`?0`).
+
+In response, the server can announce that it supports client hints and specify the hints that it is interested in receiving using the {{HTTPHeader("Accept-CH")}} header.
```
miketaylr: ...or Critical-CH.

files/en-us/web/http/guides/client_hints/index.md
```diff
@@ -141,10 +156,35 @@ Headers include: {{HTTPHeader("Device-Memory")}}, {{HTTPHeader("Width")}}, {{HTT
...
+
+An Android tablet on the other hand would send the following:
+
+```http
+Sec-CH-UA: "Chrome"; v="143"
+Sec-CH-UA-Platform: "Android"
+Sec-CH-UA-Mobile: ?0
+```
+
+The {{httpheader("Sec-CH-UA-Mobile")}} header can be used to determine whether the device is a mobile device or not. For hardware-specific use cases, the device model name can be requested via the high-entropy {{httpheader("Sec-CH-UA-Model")}} hint.
```
miketaylr: Probably worth mentioning the Sec-CH-UA-Form-Factors CH here as well. So something like "the device model and form factor can be requested..."

files/en-us/web/http/guides/client_hints/index.md
```diff
@@ -141,10 +156,35 @@ Headers include: {{HTTPHeader("Device-Memory")}}, {{HTTPHeader("Width")}}, {{HTT
...
+
+For many responsive design needs, [media queries](/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries) would be more convenient and flexible. However, there may be cases where you don't have control over the individual stylesheets of a site, and need to vary the stylesheet served based on the device signature.
```
miketaylr: There are some "MQ-ish" client hints though, https://web.dev/articles/user-preference-media-features-headers. Not sure how in depth you want to go there.

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,174 @@
...
+
+Also see [Browser detection using the user agent](/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent) for more information on why serving different content to different browsers is usually a bad idea.
+
+## Requesting detailed UA information via client hints
+
+You may still have code that relies on detailed UA string data, which can't be coverted to use feature detection or progressive enhancement. Examples include fine-grained logging, fraud prevention measures, or a software help site that serves different content based on the user's device type.
+
+If this is the case, you can still access detailed UA string data via [`Sec-CH-UA-*`](/en-US/docs/Web/HTTP/Reference/Headers#user_agent_client_hints) headers (aka **User-Agent client hints**) and other headers besides. The headers provide a safer, more privacy-preserving way to send such information because servers have to opt in to the pieces of information they want, rather it being sent all the time through the `User-Agent` string. It also provides access to a wider selection of information.
```
miketaylr: "and other headers besides" - is there a missing word? You could also mention the JS API here.

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,174 @@
...
+   - {{httpheader("Sec-CH-UA-Platform")}}: The platform.
+   - {{httpheader("Sec-CH-UA-Mobile")}}: A boolean that indicates whether the browser is running on a mobile device (`?1`) or not (`?0`).
+
+3. The server can request additional client hints using the {{httpheader("Accept-CH")}} response header, which contains a comma-delimited list of the headers it would like to receive in subsequent requests. For example:
+
+   ```http
+   Accept-CH: Sec-CH-UA, Sec-CH-UA-Platform, Sec-CH-UA-Mobile, Sec-CH-UA-Model, Sec-CH-UA-Form-Factors
+   ```
+
+   Here we've requested the default set of headers. We do this because each subsequent `Accept-CH` setting overrides the previous one, and we want to continue to receive the default information _in addition_ to the newly-requested information:
```
miketaylr: > We do this because each subsequent `Accept-CH` setting overrides the previous one,

This is true for high-entropy hints but _not_ the default, low-entropy ones (those will get sent no matter what you request, or don't request). So we can either remove UA, Platform, and Mobile or rewrite this.

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,174 @@
...
+   Sec-CH-UA: "Chrome"; v="143"
+   Sec-CH-UA-Platform: "Android"
+   Sec-CH-UA-Mobile: ?1
+   Sec-CH-UA-Model: "Pixel 9"
+   Sec-CH-UA-Form-Factors: "Mobile"
+   ```
+
+Since these are hints, the browser may choose to ignore some of all of the server's requests for more information. In addition, the browser may be blocked from sending some or all of the information by various security features such as {{httpheader("Permissions-Policy")}}.
+
+[EDITORIAL: IS THIS TRUE? I READ THIS SOMEWHERE, BUT I CAN'T FIND ANY INFO ANYWHERE ON WHAT CAN BLOCK BROWSERS FROM SENDING CLIENT HINTS. WHAT CLIENT FEATURES CAN BLOCK SENDING CLIENT HINTS?]
```
miketaylr: Kinda true - see https://wicg.github.io/client-hints-infrastructure/#policy-controlled-features

Most client hints can only be sent by default for the top-level site (except the default, low-entropy ones). In order to send them across cross-site frame boundaries, you have to opt in specific origins via Permissions Policy.

And besides disabling them via commandline parameters, there's no other way to prevent them from getting sent in Chrome (maybe possible via an extension? I haven't kept up with Manifest v3 changes, so not sure).

files/en-us/web/http/reference/headers/user-agent/index.md
```diff
@@ -91,27 +148,55 @@ Older, Presto-based Opera releases used:
...
 Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.124 Safari/537.36 Edg/91.0.864.59
 ```
 
 ## Safari UA string
 
-In this example, the user agent string is mobile Safari's version. It contains the word `"Mobile"`.
+Safari is based on the WebKit engine, but its UA string is also similar to the Blink-based browsers. It tends to include a `Version/xxx` string before the actual engine build version to indicate the browser release version, which unlike Blink-based browsers is different. In the case of iPhone (Mobile) Safari, the string also includes `Mobile`.
+
+> [!NOTE]
+> At the time of writing, non-Apple iPhone browsers (such as Firefox, Chrome, and Edge) are still based on WebKit, therefore their UA strings are the same or similar to the Safari UA string.
```
miketaylr: Similar is more correct than same here. I think these all follow a similar pattern, which is to include some identififer like `CriOS/version` or `FxiOS/version`

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent/Firefox#firefox_for_ios
https://chromium.googlesource.com/chromium/src.git/+/master/docs/ios/user_agent.md

hamishwillee:
@chrisdavidmills Can you ping me directly when ready for editorial reveiw.

files/en-us/web/api/navigator/useragent/index.md
chrisdavidmills: Good call; updated in next commit.

files/en-us/web/http/guides/client_hints/index.md
chrisdavidmills: Ah, good catch. Updated in all appropriate places.

I'd completely forgotten about echo.opera.com. Livin' in the past, man! Still very useful tho.

files/en-us/web/http/guides/client_hints/index.md
chrisdavidmills: Added, along with a link to the "Critical client hints" section later on that covers it.

files/en-us/web/http/guides/client_hints/index.md
chrisdavidmills: Added

files/en-us/web/http/guides/client_hints/index.md
chrisdavidmills: Good call. I don't think this invalidates what I am saying here, but it should be expanded to cover these cases. I've updated the text to:

> For many responsive design needs, [media queries](/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries) may be more convenient and flexible. However, there may be cases where you don't have control over the individual stylesheets of a site, and need to vary the stylesheet served based on the device signature or some kind of user preference. There are client hints that mirror some of the "user preference" media queries, such as {{httpheader("Sec-CH-Prefers-Color-Scheme")}}, {{httpheader("Sec-CH-Prefers-Reduced-Motion")}}, and {{httpheader("Sec-CH-Prefers-Reduced-Transparency")}}.

I don't think it needs much more detail on this page. There are examples and other details on the linked ref pages.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: I think I can probably just delete it. I was thinking of headers that aren't `Sec-CH-UA-*` but can still be requested like client hints, like `Save-Data`. But it did sound wishy washy.

I've included a note mentioning the JavaScript option and linked to the section that covers it in more detail.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: Ah, I misunderstood this. I've rewritten it to make it clear that the defaults are always sent. I've also added a section that defines low- and high-entropy hints in a note, as it is useful to understand.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: I've removed this bit, and added info about permissions policy to the new section mentioned in the previous comment.

Looking at the link you sent gives me a sinking feeling — we don't document _any_ of those permission-policy directives on MDN. For now, I'm going to link to the spec (which will be frowned upon, as generally we don't link to the spec if we can avoid it), and then suggest adding docs for those directives in a follow-on PR.

files/en-us/web/http/reference/headers/user-agent/index.md
chrisdavidmills: I've updated it to "UA strings are similar to".

I've also added a bit explaining the extra identifiers to each section, like I did for the Chrome section. Except for the Firefox section, as it has a whole dedicated article that explains it already.

files/en-us/web/http/guides/client_hints/index.md
miketaylr: Looks good, thanks.

files/en-us/web/http/guides/user-agent_reduction/index.md
miketaylr: (I figured it might be some wild English from the Isle of Man, but it would be rude to write that).

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,182 @@
...
+   ```
+
+For more information, see [User-Agent client hints](/en-US/docs/Web/HTTP/Guides/Client_hints).
+
+### Low- and high-entropy hints
+
+Client hints are divided in low-entropy and high-entropy hints:
+
+- The default hints are considered low-entropy hints because they don't give away much information that could be used to fingerprint a user.
+- All other hints are considered high-entropy — they could potentially be used for fingerprinting, so they are controlled by user preferences or by preferences such as {{httpheader("Permissions-Policy")}}.
```
miketaylr: One thing to note here: I'm not aware of any browser that ties a user preference to sending high-entropy hints or not, but it would be a cool feature and totally allowed by the spec.

miketaylr (commented):
LGTM - thanks Chris. Just one note, but not sure it requires any edits.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: Heh, a Manx Gaelic MDN locale would be wild. Gonna push for that.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: It's not really a good idea to include it in MDN if it's currently only theoretical. I've updated it to

> "...for fingerprinting, so they are controlled by {{httpheader("Permissions-Policy")}} headers.

chrisdavidmills:
> LGTM - thanks Chris. Just one note, but not sure it requires any edits.

Super cool, thanks for the review, @miketaylr! I'll move this one on to the editorial review stage, but feel free to chime in if you notice anything else that needs attention.

files/en-us/web/api/navigator/appversion/index.md
```diff
@@ -8,37 +8,50 @@ browser-compat: api.Navigator.appVersion
...
+- You really have no guarantee that the user agent advertised by this property is really the one your site is loaded in. Browser vendors can basically do what they like with the UA string, and some browsers enable users to change the value of this field if they want (**UA spoofing**).
+- Browser detection lead to a situation where browsers had to return fake values from such properties in order not to be locked out of some websites.
 
-The `window.navigator.userAgent` property may also contain the version
-number (for example
-`"Mozilla/5.0 (Windows; U; Win98; en-US; rv:0.9.2) Gecko/20010725 Netscape 6/6.1"`),
-but you should be aware of how easy it is to change the user agent string and "spoof"
-other browsers, platforms, or user agents, and also how cavalier the browser vendor
-themselves are with these properties.
+See [Browser detection using the user agent](/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent) for more information on why serving different content to different browsers is usually a bad idea, and recommendations for what you should do instead.
```
hamishwillee: This is almost a duplicate of https://pr41648.review.mdn.allizom.net/en-US/docs/Web/API/Navigator/userAgent#description (which lacks the - desirable - structure of the "Browser detection heading. Problem is that this refers to the UA string, but there is no linkage to UA string in this topic - so that seems weird. 

I wouldn't duplicate "in full". For this sections something like below (assuming you add a browser detection heading in the other doc).


```suggestion
Theoretically this information is useful for detecting the browser and serving code to work around browser-specific bugs or lack of feature support. However, this is **unreliable** and **is not recommended** for the same reasons given in [User-Agent reduction]/en-US/docs/Web/HTTP/Guides/User-agent_reduction#alternatives_to_ua_string_browser_sniffing) and [Browser detection using the user agent](/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent) .

[Feature detection](/en-US/docs/Learn_web_development/Extensions/Testing/Feature_detection) is a much more reliable strategy.
```

files/en-us/web/api/navigator/appversion/index.md
```diff
@@ -8,37 +8,50 @@ browser-compat: api.Navigator.appVersion
...
+A string representing version information about the browser.
 
-## Examples
+## Description
 
-```js
-alert(`Your browser version is reported as ${navigator.appVersion}`);
+The `appVersion` property returns some information indicating the browser version.
+
+In some browsers, such as Chrome, this is nearly the same as the value returned by {{domxref("Navigator.userAgent")}}, with the `Mozilla/` prefix removed. For example:
```
hamishwillee: Perhaps


```suggestion
Note that the information returned varies significantly by browser. In some browsers, such as Chrome, this is nearly the same as the value returned by {{domxref("Navigator.userAgent")}}, with the `Mozilla/` prefix removed. For example:
```

files/en-us/web/api/navigator/appversion/index.md
```diff
@@ -8,37 +8,50 @@ browser-compat: api.Navigator.appVersion
...
 
-The `window.navigator.appVersion`, `window.navigator.appName` and
-`window.navigator.userAgent` properties have been used in "browser sniffing"
-code: scripts that attempt to find out what kind of browser you are using and adjust
-pages accordingly. This lead to the current situation, where browsers had to return fake
-values from these properties in order not to be locked out of some websites.
+### User-Agent reduction
+
+The information exposed in the `appVersion` property has historically raised [privacy](/en-US/docs/Web/Privacy) concerns — it can be used to identify a particular user agent, and can therefore be used for {{glossary("fingerprinting")}}. To mitigate such concerns, [supporting browsers](#browser_compatibility) provide a reduced set of information in their {{HTTPHeader("User-agent")}} header, the {{domxref("Navigator.userAgent", "userAgent")}} property, and other related properties. For more information, see [User-Agent reduction](/en-US/docs/Web/HTTP/Guides/User-agent_reduction).
```
hamishwillee: Is their an associated BCD update? Right now that link takes me to the browser compat which links all browsers - but the implication of this link is that I can get a list of those that support user-agent reduction.

Further it might be worth clarifying "reduced compared to the historical amount".

files/en-us/web/api/navigator/appversion/index.md
hamishwillee: @chrisdavidmills Great minds, just reviewed down to `navigator.platform` which is very similar to this. I think you can spell it out once in https://pr41648.review.mdn.allizom.net/en-US/docs/Web/HTTP/Guides/User-agent_reduction and cross link the same way in each. But they certainly should be the same for most of these.

files/en-us/web/api/navigator/appversion/index.md
hamishwillee: Same comment for the other docs.

files/en-us/web/api/navigator/platform/index.md
```diff
@@ -10,29 +10,43 @@ browser-compat: api.Navigator.platform
...
 
 ## Examples
 
-`navigator.platform` should almost always be avoided in favor of [feature detection](/en-US/docs/Learn_web_development/Extensions/Testing/Feature_detection). But there is one case where, among the options you could use, `navigator.platform` may be the least-bad option: When you need to show users advice about whether the modifier key for keyboard shortcuts is the `⌘` command key (found on Apple systems) rather than the `⌃` control key (on non-Apple systems):
+One case where `navigator.platform` can be useful is when you need to show users advice about whether the modifier key for keyboard shortcuts is the `⌘` command key (found on Apple systems) rather than the `Ctrl` control key (on non-Apple systems):
```
hamishwillee, line 36: Web APIs should have a level 3 heading with specific info about the example too. For the other cases the examples are so trivial there is no real point, but it makes sense in this one.

files/en-us/web/api/navigator/useragent/index.md
```diff
@@ -8,46 +8,37 @@ browser-compat: api.Navigator.userAgent
...
-Browser identification based on detecting the user agent string is
-**unreliable** and **is not recommended**, as the user agent
-string is user configurable. For example:
-
-- In Firefox, you can change the preference `general.useragent.override` in
-  `about:config`. Some Firefox extensions do that; however, this only changes
-  the HTTP header that gets sent and that is returned by `navigator.userAgent`.
-  There might be other methods that utilize JavaScript code to identify the browser.
-- Opera 6+ allows users to set the browser identification string via a menu.
+The **`Navigator.userAgent`** read-only property of the {{domxref("Navigator")}} interface returns the `User-Agent` (UA) string for the current browser.
```
hamishwillee, line 11: Worth linking IMO to show where it comes from.
```suggestion
The **`Navigator.userAgent`** read-only property of the {{domxref("Navigator")}} interface returns the {{httpheader("User-Agent")}} (UA) string for the current browser.
```

hamishwillee: Hmm, actually you have linked immediately below. I was going to argue linking on first-instance is better, but perhaps ignore this one.

files/en-us/web/api/navigator/useragent/index.md
```diff
@@ -8,46 +8,37 @@ browser-compat: api.Navigator.userAgent
...
-several pieces of info. Each of these pieces of info comes from other navigator
-properties which are also settable by the user. For more information about the
-form of the user agent string, see the {{HTTPHeader("User-agent")}} HTTP header.
+The browser also provides this via the {{HTTPHeader("User-agent")}} HTTP header. Parts of this information are also available in {{Glossary("HTTP")}} headers such as [User-Agent client hints](/en-US/docs/Web/HTTP/Guides/Client_hints) and other related API features such as {{domxref("Navigator.appVersion")}} and {{domxref("Navigator.platform")}}.
+
+The UA string is built on a formal structure, which can be decomposed into several pieces of information. For more information about the form of the UA string, see the {{HTTPHeader("User-agent")}} HTTP header.
+
+## Description
+
+The `userAgent` property provides the current browser's UA string. Theoretically this is useful for detecting the browser and serving code to work around browser-specific bugs or lack of feature support. However, browser identification based on detecting the UA string is **unreliable** and **is not recommended**:
```
hamishwillee: The stuff in Value would be better in description

```suggestion
The **`Navigator.userAgent`** read-only property of the {{domxref("Navigator")}} interface returns the {{HTTPHeader("User-agent")}} (UA) string for the current browser.

## Value

A string specifying the browser's complete UA string.

## Description

The `userAgent` property provides the current browser's UA string. 
The string is built on a formal structure, which can be decomposed into several pieces of information.

The browser also provides this via the {{HTTPHeader("User-agent")}} HTTP header.
Parts of this information are also available in {{Glossary("HTTP")}} headers such as [User-Agent client hints](/en-US/docs/Web/HTTP/Guides/Client_hints) and other related API features such as {{domxref("Navigator.appVersion")}} and {{domxref("Navigator.platform")}}.

### Browser detection

Theoretically this is useful for detecting the browser and serving code to work around browser-specific bugs or lack of feature support. However, browser identification based on detecting the UA string is **unreliable** and **is not recommended**:

```

files/en-us/web/http/reference/headers/user-agent/index.md
```diff
@@ -46,6 +46,30 @@ User-Agent: Mozilla/5.0 (<system-information>) <platform> (<platform-details>) <
...
+
+```plain
+Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Mobile Safari/537.36
+```
+
+- The platform version is always a fixed value, in this case, `Android 10`.
+- The device model is always a fixed value, in this case, `K`.
+- The Chrome major version number shows correctly, but the minor version numbers are always shown as zeros — `0.0.0`.
+
+Servers that want to request more information can do so using [User-Agent client hints](/en-US/docs/Web/HTTP/Guides/Client_hints). After the initial connection, the server can send an {{httpheader("Accept-CH")}} response header detailing the data items they want, and the client can then send the data back via [`Sec-CH-UA-*`](/en-US/docs/Web/HTTP/Reference/Headers#user_agent_client_hints) headers. This information can also be accessed via the [User-Agent Client Hints API](/en-US/docs/Web/API/User-Agent_Client_Hints_API).
```
hamishwillee: Perhaps worth clarifying "need" is desirable.
```suggestion
Servers that need more information can request it via [User-Agent client hints](/en-US/docs/Web/HTTP/Guides/Client_hints). After the initial connection, the server can send an {{httpheader("Accept-CH")}} response header detailing the data items they want, and the client can then send the data back via [`Sec-CH-UA-*`](/en-US/docs/Web/HTTP/Reference/Headers#user_agent_client_hints) headers. This information can also be accessed via the [User-Agent Client Hints API](/en-US/docs/Web/API/User-Agent_Client_Hints_API).
```

files/en-us/web/http/reference/headers/user-agent/index.md
```diff
@@ -64,25 +88,58 @@ Mozilla/5.0 (platform; rv:gecko-version) Gecko/gecko-trail Firefox/firefox-versi
...
+
+```plain
+Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Mobile Safari/537.36
+```
+
+#### Pre-User-Agent reduction example
+
 ```plain
 Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36
 ```
```
hamishwillee: Odd rendering because this is a heading. I'd perhaps do it like


````suggestion
> [!NOTE]
> Prior to user-agent reduction the UA string might have looked like this:
> ```plain
>  Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36
>  ```
````

Comment (hamishwillee) [reply files/en-us/web/http/reference/headers/user-agent/index.md]: Actually, just a suggestion, but you could put ALL the pre-user reduction examples in one section for all browsers. I think this might age better because over time the whole user agent reduction thing is going to be business as usual.

File: files/en-us/web/http/guides/user-agent_reduction/index.md [hunk 19]
```diff
@@ -0,0 +1,182 @@
+---
+title: User-Agent reduction
+slug: Web/HTTP/Guides/User-agent_reduction
+page-type: guide
+sidebar: http
+---
+
+The information exposed in the {{httpheader("User-Agent")}} HTTP header has historically raised [privacy](/en-US/docs/Web/Privacy) concerns — it can be used to identify a particular user agent, and can therefore be used for {{glossary("fingerprinting")}}. To mitigate such concerns, **User-Agent reduction** provides a reduced set of information in the browser `User-Agent` header, and in related API features such as {{domxref("Navigator.userAgent")}}, {{domxref("Navigator.appVersion")}}, and {{domxref("Navigator.platform")}}.
+
+This article explains the differences in user agent (UA) strings as a result of User-Agent reduction, and explains modern strategies for accessing further UA information if required.
```
hamishwillee:
```suggestion
This article explains the differences in user agent (UA) strings as a result of User-Agent reduction, and explains how you can access both the redacted and additional UA information if it is needed.
```

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,182 @@
...
+
+Also see [Browser detection using the user agent](/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent) for more information on why serving different content to different browsers is usually a bad idea.
+
+## Requesting detailed UA information via client hints
+
+You may still have code that relies on detailed UA string data, which can't be coverted to use feature detection or progressive enhancement. Examples include fine-grained logging, fraud prevention measures, or a software help site that serves different content based on the user's device type.
+
+If this is the case, you can still access detailed UA string data via [`Sec-CH-UA-*`](/en-US/docs/Web/HTTP/Reference/Headers#user_agent_client_hints) headers (aka **User-Agent client hints**). The headers provide a safer, more privacy-preserving way to send such information because servers have to opt in to the pieces of information they want, rather it being sent all the time through the `User-Agent` string. It also provides access to a wider selection of information.
```
hamishwillee:
```suggestion
If this is the case, you can still access detailed UA string data via [`Sec-CH-UA-*`](/en-US/docs/Web/HTTP/Reference/Headers#user_agent_client_hints) headers (also known as **User-Agent client hints**). The headers provide a safer, more privacy-preserving way to send such information because servers have to opt in to the pieces of information they want, rather it being sent all the time through the `User-Agent` string. It also provides access to a wider selection of information.
```

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,182 @@
...
+
+> [!NOTE]
+> You can also [access client hint information via JavaScript](#accessing_client_hints_via_javascript).
+
+Client hints are used like so:
+
+1. When the browser first makes a request to load a webpage, it will send the reduced `User-Agent` header (as detailed earlier) to the server.
+2. Additionally, it will send the server a default set of `Sec-CH-UA-*` headers. The Android example we looked at earlier would send the following:
```
hamishwillee:
```suggestion
2. Additionally, it will send the server a default set of `Sec-CH-UA-*` headers this set of hints are referred to as the [low entropy hints](/en-US/docs/Web/HTTP/Guides/Client_hints#low_entropy_hints)). The Android example we looked at earlier would send the following:
```

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,182 @@
...
+   - {{httpheader("Sec-CH-UA-Platform")}}: The platform.
+   - {{httpheader("Sec-CH-UA-Mobile")}}: A boolean that indicates whether the browser is running on a mobile device (`?1`) or not (`?0`).
+
+3. The server can request additional client hints using the {{httpheader("Accept-CH")}} response header, which contains a comma-delimited list of the additional headers it would like to receive in subsequent requests. For example:
+
+   ```http
+   Accept-CH: Sec-CH-UA-Model, Sec-CH-UA-Form-Factors
+   ```
+
+   The default set of headers are always sent. In addition to those, we've also requested:
```
hamishwillee:
```suggestion
   The default "low entropy" headers are always sent. In addition to those, we've also requested:
```

files/en-us/web/http/guides/user-agent_reduction/index.md
hamishwillee: Hmmm, I see you refer to low entropy below, so you might link down to that section (not the section I point to above). I do think it is worth at least refrering to these as low entropy hints on first mention as well as "default".

files/en-us/web/http/guides/user-agent_reduction/index.md
hamishwillee: Given you document the low entropy shortly below we don't need a second reference here sorry.

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,182 @@
...
+   Sec-CH-UA: "Google Chrome";v="143", "Chromium";v="143", "Not A(Brand";v="24"
+   Sec-CH-UA-Platform: "Android"
+   Sec-CH-UA-Mobile: ?1
+   Sec-CH-UA-Model: "Pixel 9"
+   Sec-CH-UA-Form-Factors: "Mobile"
+   ```
+
+For more information, see [User-Agent client hints](/en-US/docs/Web/HTTP/Guides/Client_hints).
+
+### Low- and high-entropy hints
```
hamishwillee: Link to the high/low entropy sections in https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Client_hints#low_entropy_hints

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,182 @@
...
+
+Client hints are divided in low-entropy and high-entropy hints:
+
+- The default hints are considered low-entropy hints because they don't give away much information that could be used to fingerprint a user.
+- All other hints are considered high-entropy — they could potentially be used for fingerprinting, so they are controlled by {{httpheader("Permissions-Policy")}} headers.
+
+By default, high-entropy hints can only be sent for the top-level site (you have to opt-in to send them across cross-site frame boundaries). See [Policy-controlled features](https://wicg.github.io/client-hints-infrastructure/#policy-controlled-features) for a list of the associated `Permissions-Policy` directives.
+
+### Critical client hints
```
hamishwillee: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Client_hints#critical_client_hints

files/en-us/web/http/guides/user-agent_reduction/index.md
hamishwillee: Actually much of this is duplication. I think this whole document should be reduced and updated where necessary in https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Client_hints#critical_client_hints. The might be missing stuff in that too - such as more on the JavaScript side.

hamishwillee (commented):
Looks good. Some comments, the most important one is probably https://github.com/mdn/content/pull/41648/files#r2464158868 - lots of duplication with the Client Hints guide.

files/en-us/web/api/navigator/appversion/index.md
chrisdavidmills: This is helpful, thanks. I wasn't sure how much repetition would be useful, so I decided to overdo it a bit and see what you thought. I have rewritten the user-agent reduction article so that it presents the information on browser detection and privacy concerns up-front, as background for why user-agent reduction came to be.

I then reduced the repetitive info in the API articles to text similar to your suggestion. 

I think the actual `User-Agent` header page needs a bit more info on it, plus an example of the changes due to the User-Agent reduction, so I've left that intact for now.

files/en-us/web/api/navigator/appversion/index.md
chrisdavidmills: Good call; added.

files/en-us/web/api/navigator/appversion/index.md
chrisdavidmills: Yes, there is: https://github.com/mdn/browser-compat-data/pull/28275.

I don't think the "reduced compared to the historical amount" text is on these pages anymore, now that I have linked them all to the user-agent reduction page (which explains what it means in detail anyway).

files/en-us/web/api/navigator/platform/index.md
chrisdavidmills: H3 added.

files/en-us/web/api/navigator/useragent/index.md
chrisdavidmills: Agreed. I've updated the page.

files/en-us/web/http/reference/headers/user-agent/index.md
chrisdavidmills: updated

files/en-us/web/http/reference/headers/user-agent/index.md
chrisdavidmills: This is a good idea. I have updated the page to put all the examples in one section. 

I also removed the "Examples" headings, as they seemed repetitive.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: updated

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: updated

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: Update applied; yup, I've applied the link to the section below in the same article.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: This is a good point. I have put the good stuff into the client hints article, and cut down the client hints and JS API sections to summarize the info, with links to the appropriate places for more information.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: I removed this section from the user-agent reduction article.

files/en-us/web/api/navigator/appversion/index.md
```diff
@@ -16,9 +16,9 @@ A string representing version information about the browser.
```
hamishwillee, line 16: Just a FWIW, but I'm never sure what to have in the overview, value, description to avoid irritating duplication. By way of example, here we have:
- ... returns a string representing version information about the browser.
- A string representing version information about the browser.
- ... returns information indicating the browser version.

I don't know the answer in general, but for this case I would consider making the value into "A string". Not so in the other cases though.

files/en-us/web/api/navigator/appversion/index.md
```diff
@@ -30,19 +30,9 @@ In other browsers, such as Firefox, this is cut down to a short string that hint
 5.0 (Macintosh)
 ```
 
-### Browser detection
+Theoretically this information is useful for detecting the browser and serving code to work around browser-specific bugs or lack of feature support. However, this is **unreliable** and **is not recommended** for the reasons given in [User-Agent reduction](/en-US/docs/Web/HTTP/Guides/User-agent_reduction) and [Browser detection using the user agent](/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent).
```
hamishwillee, line 33: FWIW, I like the new reduced structure and links here.

files/en-us/web/http/reference/headers/user-agent/index.md
```diff
@@ -199,6 +165,36 @@ iPhone example:
 Mozilla/5.0 (iPhone; CPU iPhone OS 18_6 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.0 Mobile/15E148 Safari/604.1
 ```
 
+## Pre-user-agent reduction examples
+
+This section provides some examples of UA strings prior to user-agent reduction:
```
hamishwillee: Maybe
```suggestion
This section provides some examples of UA strings in earlier browser versions, prior to the introduction of user-agent reduction:
```

files/en-us/web/http/guides/client_hints/index.md
```diff
@@ -13,19 +13,41 @@ The set of "hint" headers are listed in the topic [HTTP Headers](/en-US/docs/Web
...
+   - {{httpheader("Sec-CH-UA-Platform")}}: The platform.
+   - {{httpheader("Sec-CH-UA-Mobile")}}: A boolean that indicates whether the browser is running on a mobile device (`?1`) or not (`?0`).
+
+3. The server can announce that it supports client hints and request additional client hints using the {{httpheader("Accept-CH")}} response header, which contains a comma-delimited list of the additional headers it would like to receive in subsequent requests. For example:
+
+   ```http
+   Accept-CH: Sec-CH-UA-Model, Sec-CH-UA-Form-Factors
+   ```
+
+   The default set of headers are always sent. In addition to those, we've also requested:
```
hamishwillee: Perhaps useful to say "in this case", even though it is obvious.

```suggestion
   The default set of headers are always sent. In this case we've also requested:
```

files/en-us/web/http/guides/client_hints/index.md
```diff
@@ -13,19 +13,41 @@ The set of "hint" headers are listed in the topic [HTTP Headers](/en-US/docs/Web
...
+   - {{httpheader("Sec-CH-UA-Model")}}: The device model the platform is running on.
+   - {{httpheader("Sec-CH-UA-Form-Factors")}}: The device's form factor(s), which indicate how the user interacts with the user-agent — the screen size, controls, etc.
 
-This approach is efficient in that the server only requests the information that it is able to usefully handle.
-It is also relatively "privacy-preserving", in that it is up to the client to decide what information it can safely share.
+4. If the browser is permitted to send the server all the requested information, it will do so along with all subsequent requests until the browser or tab is closed. For example, our example Android phone might send the following updated headers with subsequent requests:
```
hamishwillee:
```suggestion
4. If the browser is permitted, it will send the requested headers in all subsequent requests, until the browser or tab is closed. For example, our example Android phone might send the following updated headers with subsequent requests:
```

files/en-us/web/http/guides/client_hints/index.md
```diff
@@ -63,7 +85,7 @@ For example, to stop requesting any hints it would send `Accept-CH` with an empt
 ## Low entropy hints
 
 Client hints are broadly divided into high and low entropy hints.
-The low entropy hints are those that don't give away much information that might be used to create a [fingerprinting](/en-US/docs/Glossary/Fingerprinting) for a user.
+The low entropy hints are those that don't give away much information that might be used to [fingerprint](/en-US/docs/Glossary/Fingerprinting) a user.
 They may be sent by default on every client request, irrespective of the server `Accept-CH` response header, depending on the permission policy.
```
hamishwillee: This isn't on you, but we should define what we mean by permission policy here, because normally this means a specific permission, and there is no client hints permission.

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,89 @@
+---
+title: User-Agent reduction
+slug: Web/HTTP/Guides/User-agent_reduction
+page-type: guide
+sidebar: http
+---
+
+This article explains the differences in user agent (UA) strings as a result of **User-Agent reduction**, and explains how you can access both the redacted and additional UA information if it is needed.
```
hamishwillee: The double use of explains followed by explains reads oddly. More important is that we introduce the term we are going to describe without defining it. That's OK if it is hard to do so, but in this case it isn't. 

```suggestion
User-Agent reduction is a broadly accepted browser initiative to reduce the amount of privacy-sensitive information provided in the user agent (UA) strings.

This article shows the differences in UA strings as a result of **User-Agent reduction**, and explains how you can access both the redacted and additional UA information if it is needed.
```

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Browser detection
+
+Theoretically the UA string is useful for detecting the browser and serving code to work around browser-specific bugs or lack of feature support. However, this is **unreliable** and **is not recommended**:
+
+- Future browsers will fix bugs and add support for new features, so your browser detection code will need to be regularly updated to avoid locking out browsers that do actually support the features you are testing for. [Feature detection](/en-US/docs/Learn_web_development/Extensions/Testing/Feature_detection) is a much more reliable strategy.
+- You really have no guarantee that the user agent advertised by this property is really the one your site is loaded in. Browser vendors can basically do what they like with the UA string, and some browsers enable users to change the value of this field if they want (**UA spoofing**).
+- Browser detection lead to a situation where browsers had to return fake values from such properties in order not to be locked out of some websites.
```
hamishwillee: The tense and tone is different in the final bullet than the others. Maybe merge them.

Note, one reason I love sentence breaks is that it makes reviewing long line updates much easier.


```suggestion
- You really have no guarantee that the user agent advertised by this property is really the one your site is loaded in. Browser vendors can basically do what they like with the UA string, and historically would return fake values from such properties in order not to be locked out of some websites.
  Some browsers enable users to change the value of this field if they want (**UA spoofing**).
```

files/en-us/web/http/guides/user-agent_reduction/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+Also see [Browser detection using the user agent](/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent) for more information on why serving different content to different browsers is usually a bad idea.
+
+### Privacy concerns
+
+In addition, the information exposed in the UA string has historically raised [privacy](/en-US/docs/Web/Privacy) concerns — it can be used to identify a particular user agent, and can therefore be used for {{glossary("fingerprinting")}}.
+
+To mitigate such concerns, [supporting browsers](/en-US/docs/Web/HTTP/Reference/Headers/User-Agent#browser_compatibility) implement user-agent reduction, which updates the `User-agent` header and related API features to provide a reduced set of information.
+
+## UA string differences resulting from User-Agent reduction
```
hamishwillee: This heading wraps - can we simplify it? Maybe


```suggestion
## UA string changes after reduction
```

hamishwillee (commented):
Much better IMO. Few more nittish things.

files/en-us/web/http/guides/client_hints/index.md
miketaylr: Note: permissions and permissions policies are related, but different. And client hints do define a number of [policy controlled features](https://wicg.github.io/client-hints-infrastructure/#policy-controlled-features) (which is what a permission policy controls), but you're correct that they're not powerful features (which is what permissions control).

https://w3c.github.io/webappsec-permissions-policy/
https://w3c.github.io/permissions/

hamishwillee: G'day @miketaylr 

Thanks. So to be clear, you're saying that I might set a policy for access to the various features using https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy with one of the linked permission names? 
So by default `Width` is available to the current origin but if I want to to be available to some other origin in a nested context I could set `Permissions-Policy: ch-width=*` (say) to get permission?

The reason I ask is that the existence of this level of permissions is unknown to me, and likely unknown to the rest of MDN. The things I/we know about are listed here https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy#directives

Off hand do you know of many other similar lists of permissions. I can make sure they get added if we know about them.

files/en-us/web/api/navigator/appversion/index.md
chrisdavidmills: Yeah, they are a little repetitive. It makes sense to say "A string" for `appVersion`/`userAgent`. For `platform`, I ended up going with "A string indicating a platform, for example:"

files/en-us/web/api/navigator/appversion/index.md
chrisdavidmills: Awesome.

files/en-us/web/http/reference/headers/user-agent/index.md
chrisdavidmills: Like it; updated.

files/en-us/web/http/guides/client_hints/index.md
chrisdavidmills: Yeah, makes sense. Updated.

files/en-us/web/http/guides/client_hints/index.md
chrisdavidmills: updated

files/en-us/web/http/guides/client_hints/index.md
chrisdavidmills: I've added links to the list of permission polices in the spec for now. I'd previously added a link to the User-agent reduction article, but that section then got removed.

It isn't a perfect solution, but I don't want to document those 22 new permissions policy directives as part of this PR. I think that's a follow-up job.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: Yeah, good call. I have used your text, with a few tweaks.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: Updated, except that I made the last sentence into its own bullet. I think it works as its own reason why browser detection is unreliable and should be avoided.

files/en-us/web/http/guides/user-agent_reduction/index.md
chrisdavidmills: done

files/en-us/web/http/guides/client_hints/index.md
hamishwillee: I agree. Not closing this as resolved yet, because I'm hoping @miketaylr might respond to my question above on "other lists".

hamishwillee (approved):
Thanks @chrisdavidmills - this holds together well IMO.

chrisdavidmills:
> Thanks @chrisdavidmills - this holds together well IMO.

Lovely, thanks @hamishwillee, and thanks to @miketaylr for the tech review.

files/en-us/web/http/guides/client_hints/index.md
miketaylr: @hamishwillee I'm sorry to have missed the ping here!

> So by default Width is available to the current origin but if I want to to be available to some other origin in a nested context I could set Permissions-Policy: ch-width=* (say) to get permission?

Yep, that's exactly correct. And yes, the discoverability for each of these permissions policy is quite low. 

We (the permissions and permissions policy editors) had hoped to improve that via https://w3c.github.io/permissions-registry/, which would eventually cover permissions policies in addition to powerful features (aka, permissions)... but some folks had strong feelings against it and that was enough stop energy to just give up. For now, https://dontcallmedom.github.io/webdex/p.html#policy-controlled%20feature%40%40permissions-policy%25%25dfn is probably the best source of truth (modulo spec bugs that don't link things correctly).

hamishwillee: Thanks. This is still on my list. I'll have to tidy up some links like those mentioned in https://github.com/mdn/content/pull/42880/files#r2726383362

PR #41651, opened 2025-10-24
Set-Cookie: Domain directive default value clarification

### LINKED ISSUE #41646: Set-Cookie without Domain attribute
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#domaindomain-value

Domain=<domain-value>

The sentence

    If omitted, this attribute defaults to
    the host of the current document URL,
    not including subdomains.

could be improved.

A reader could assume that sending a cookie from `a.example.com` would also be valid for `example.com`, because `example.com` is a URL `not including subdomains`.

However, a cookie sent from `a.example.com` is only valid for that particular domain, neither `aa.a.example.com`, nor `b.example.com`, nor `example.com`.

    If omitted, this attribute defaults to
    the host of the current document URL.
    (This is more restrictive than sending the host name,
    as subdomains of the host are not included.)

Reference:
https://datatracker.ietf.org/doc/html/rfc6265#page-20

    If the server omits the Domain attribute, the user
    agent will return the cookie only to the origin server.

And tests in Chrome, Firefox, Safari.

_No response_

hamishwillee:
Fixes #41646

### DIFF
--- a/files/en-us/web/http/reference/headers/set-cookie/index.md
+++ b/files/en-us/web/http/reference/headers/set-cookie/index.md
@@ -74,7 +74,7 @@ Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnl
 
     Only the current domain can be set as the value, or a domain of a higher order, unless it is a public suffix. Setting the domain will make the cookie available to it, as well as to all its subdomains.
 
-    If omitted, the cookie is returned only to the host that sent them (i.e. it becomes a "host-only cookie").
+    If omitted, the cookie is returned only to the host that sent them (i.e., it becomes a "host-only cookie").
     This is more restrictive than setting the host name, as the cookie is not made available to subdomains of the host.
 
     Contrary to earlier specifications, leading dots in domain names (`.example.com`) are ignored.
--- a/files/en-us/web/http/reference/headers/set-cookie/index.md
+++ b/files/en-us/web/http/reference/headers/set-cookie/index.md
@@ -74,7 +74,7 @@ Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnl
 
     Only the current domain can be set as the value, or a domain of a higher order, unless it is a public suffix. Setting the domain will make the cookie available to it, as well as to all its subdomains.
 
-    If omitted, the cookie is returned only to the origin of the current document URL.
+    If omitted, the cookie is returned only to the host that sent them.
     This is more restrictive than setting the host name, as the cookie is not made available to subdomains of the host.
 
     Contrary to earlier specifications, leading dots in domain names (`.example.com`) are ignored.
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/reference/headers/set-cookie/index.md
```diff
@@ -74,7 +74,8 @@ Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnl
 
     Only the current domain can be set as the value, or a domain of a higher order, unless it is a public suffix. Setting the domain will make the cookie available to it, as well as to all its subdomains.
 
-    If omitted, this attribute defaults to the host of the current document URL, not including subdomains.
```
hamishwillee, line 77: Note that I didn't like this, because "the attribute default to" is not quite correct. That implies that this is the same as setting the current document host, and the "not including subdomains" doesn't quite capture that it is not.

bsmth: I think we're a little off, the "origin server" language in the spec is not the `origin` term, which is `scheme+host+port`. I think we're looking at "host only":

* https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/Cookie#hostonly
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#cookie_prefixes
* https://stackoverflow.com/questions/12387338/what-is-a-host-only-cookie

files/en-us/web/http/reference/headers/set-cookie/index.md
```diff
@@ -74,7 +74,8 @@ Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnl
...
+    If omitted, the cookie is returned only to the origin of the current document URL.
+    This is more restrictive than setting the host name, as the cookie is not made available to subdomains of the host.
```
hamishwillee:
```suggestion
    If omitted, the cookie is returned only to the host that sent them.
    This is more restrictive than setting the host name, as the cookie is not made available to subdomains of the host.
```

files/en-us/web/http/reference/headers/set-cookie/index.md
```diff
@@ -74,7 +74,8 @@ Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnl
 
     Only the current domain can be set as the value, or a domain of a higher order, unless it is a public suffix. Setting the domain will make the cookie available to it, as well as to all its subdomains.
 
-    If omitted, this attribute defaults to the host of the current document URL, not including subdomains.
+    If omitted, the cookie is returned only to the host that sent them.
```
hamishwillee:
```suggestion
    If omitted, the cookie is returned only to the host that sent them (i.e. it becomes a "host-only cookie").
```

files/en-us/web/http/reference/headers/set-cookie/index.md
hamishwillee: Thanks @bsmth - that is certainly the intent. I've modified.

As an aside, you can't determine if a cookies is host-only using the cookiestore API https://developer.mozilla.org/en-US/docs/Web/API/CookieStore/get#return_value - though as you link you can with the old-style API https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/Cookie#hostonly

bsmth (approved):
This is looking good, thank you!

PR #41663, opened 2025-10-24
relnote: fix mismatched bugzilla URLs

bsmth:
### Description

There's multiple cases of `[Firefox bug X](https://bugzil.la/Y)` where the URL doesn't match the link text.

I found them using ripgrep and fixed them up. Some are 404s, most are just copy-paste / typos.

```bash
rg -P '\[Firefox bug (\d+)\]\(https://bugzil\.la/(?!\1)\d+\)'
```

### Motivation

Making sure the references are correct.


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #40966, opened 2025-09-01
added experimental release note for text-autospace

dletorey:
### Description

- Added experimental Release note for `text-autospace`
- Added experimental note to 143 page for `text-autospace`

### Motivation

- Working on [MDN issue #40776](https://github.com/mdn/content/issues/40776)

#### Related issues

- [Content PR](https://github.com/mdn/content/pull/40912)
- [Previous BCD PR](https://github.com/mdn/browser-compat-data/pull/27672)
  - [New BCD PR](https://github.com/mdn/browser-compat-data/pull/27726)
- [Data PR](https://github.com/mdn/data/pull/1005)

PR #41665, opened 2025-10-24
Add text-decoration-trim experimental feature entry

chrisdavidmills:
### Description

The `text-decoration-trim` property is supported behind a preference as of Firefox 145. See https://bugzilla.mozilla.org/show_bug.cgi?id=1979915.

I've verified that it does work to a degree, although it currently seems buggy. As a result, I don't think we should add docs or BCD for this feature yet, but it does warrant an experimental features entry.


files/en-us/mozilla/firefox/experimental_features/index.md
```diff
@@ -336,6 +336,20 @@ The {{CSSXRef(":heading")}} pseudo-class allows you to style all [heading elemen
 - `layout.css.heading-selector.enabled`
   - : Set to `true` to enable.
 
+### `text-decoration-trim`
```
bsmth, line 353: We also have to add a para at the end of the 145 docs here: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/145#experimental_web_features

bsmth: (see https://github.com/mdn/content/pull/40966/files)

chrisdavidmills: Ok, done. Thanks.

bsmth (approved):
Great, ty

PR #41679, opened 2025-10-25
Fix typo in chemical formula example

joshavanier:
### Description

Fixing a tiny typo: `H2O` not `H20`

### Motivation

I was just going through the element docs and it caught my eye.


Josh-Cena (approved):
Nice catch, thank you

PR #41680, opened 2025-10-26
Adds `graphics-document`, `graphics-object`, and `graphics-symbol` `aam-graphics` roles.

### LINKED ISSUE #28707: Document W3C WAI-ARIA Graphics Module
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

N/A

[W3C WAI-ARIA Graphics Module](https://www.w3.org/TR/graphics-aria-1.0/) introduces three ARIA roles that has been [implemented in Chrome](https://wpt.fyi/results/graphics-aria/graphics-roles.html?label=master&label=experimental&aligned&q=label%3Aaccessibility):

* `graphics-document`
* `graphics-object`
* `graphics-symbol`

They should be noted on MDN.

These ARIA roles are documented.

_No response_

_No response_

younisdev:
### Description

Add`graphics-document`, `graphics-object`, and `graphics-symbol` `aam-graphics` roles.
### Motivation

These changes ensure that the reader are aware of the 3 graphical am-graphics roles. And ensures that MDN is up to date!
### Additional details

None.

Fixes #28707


Josh-Cena:
Reminder that you need to update the parent page, Web/Accessibility/ARIA/Reference/Roles

younisdev:
> Reminder that you need to update the parent page, Web/Accessibility/ARIA/Reference/Roles

Should I add these as a separate "Graphics roles" section?

Josh-Cena:
Not sure

files/en-us/web/accessibility/aria/reference/roles/graphics-document_role/index.md
```diff
@@ -0,0 +1,56 @@
...
+### How it differs from similar roles
+
+- `img`
+  - : Unlike an image, a `graphics-document` role may include structured, semantic content such as links and interactive regions.
+- `document`
+  - : The `graphics-document` role emphasizes the visual layout and relationships, rather than textual or structural content.
+- `graphics-object`
+  - : The `graphics-document` role is self-contained and retains its meaning when separated from the surrounding content.
+
+## Examples
```
scottaohara, line 37: this probably isn't the best example, since these roles shouldn't even be needed for these svg elements.

files/en-us/web/accessibility/aria/reference/roles/graphics-document_role/index.md
```diff
@@ -0,0 +1,56 @@
...
+  <g
+    role="graphics-object"
+    aria-label="Map compass"
+    transform="translate(50,50)"></g>
+</svg>
+```
+
+## Accessibility concerns
+
+To ensure backward compatibility with ARIA 1.0 user agents and assistive technologies, include the [`document`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/document_role) role as a fallback value.
```
scottaohara: these roles aren't defined in ARIA.  they're from the extension specs graphics aam and svg aam.

i'm not sure the accessibility concern is necessary... really only the newest of ARIA roles would need fallback roles (and these graphics roles are not new).

scottaohara (changes requested):
just leaving a quick review.

i don't have time to dig into these further, but i'm mostly concerned that the narrative here is telling people to use these roles on elements that don't need them, and including fallback roles unnecessarily.

younisdev:
> just leaving a quick review.
> 
> i don't have time to dig into these further, but i'm mostly concerned that the narrative here is telling people to use these roles on elements that don't need them, and including fallback roles unnecessarily.

Thanks for the review! I've added a note to all the aam-graphics that clearly states these docs aren't a part of the core WAI-ARIA spec. I also removed the fallback note , and  reworded sections to be less prescriptive.

younisdev:
I plan to rework Examples and the document section of these roles soon.

younisdev:
I've updated the example for the `aam-graphics` roles.
However, I'm not sure if these should remain alongside the ARIA roles on the main page, or if they should have a standalone section maybe something like `/AAM/Reference/Roles`?

pepelsbey:
@AmeliaBR since you are the co-author of the spec (it’s been a while, sorry), could you please recommend good examples for these roles? I agree with Scott, this doesn’t sit right with me:

```html
<div role="graphics-document" aria-label="Example map">
  <div role="graphics-object" aria-label="Europe region"></div>
  <div role="graphics-object" aria-label="Asia region"></div>
</div>
```

younisdev:
I've updated all the three aam-roles examples based on further research into what these roles are meant to be used for. I came across the original blog post back in 2015 https://www.w3.org/blog/2015/wai-aria-graphics-module-published/. Which quotes "Users may wish to explore connections between content in a two-dimensional manner that does not follow the linear structure of the document source code". So according to that I crafted 3 examples that the meaning and the direction of said element covey semantic meaning. 

So in summary of my recent changes:

- I added a `System Authentication flow chart` example to graphics-document_role.
- I added a  `accurate positions Simplified world map` to graphics-object_role.
- I added a `Parking lot` example to graphics-symbol role.
- Did some modification for the overall tone.
- Added a live demo for all these roles.

PR #41691, opened 2025-10-26
Fix the font-format example

pepelsbey:
### Description

- [x] Rewrites the `@supports font-format` example to fix the mistake.
- [x] Fixes the color font example URL to link to a font file, not CSS.

### Motivation

- The `src` is a descriptor exclusive to `@font-face`, not a property that can be set.
- Google Fonts serves CSS files at these URL, not font files.

### Additional details

Initially [reported on Mastodon](https://typo.social/@pixelambacht/115441236943309738) by @RoelN


files/en-us/web/css/@supports/index.md
```diff
@@ -302,17 +302,45 @@ If a browser doesn't support the font technology, a fallback font (`Bungee-fallb
 
 ### Testing for the support of a font format
 
-The following example applies the CSS style if the browser supports the `woff2` font format:
+The following example uses the WOFF 2 version of the font if the browser supports this font format, otherwise it falls back to the previously specified WOFF 1 version:
```
RoelN: I don't see the two versions being referenced as "WOFF 2" and "WOFF 1" elsewhere, just WOFF2 and WOFF, for example here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/WOFF So maybe tie into that?

files/en-us/web/css/@supports/index.md
```diff
@@ -302,17 +302,45 @@ If a browser doesn't support the font technology, a fallback font (`Bungee-fallb
...
+}
+
+@font-face {
+  font-family: "Open Sans WOFF 2";
+  src: url("open-sans.woff2") format("woff2");
+}
+
+body {
+  font-family: "Open Sans WOFF 1", sans-serif;
+}
```
RoelN, line 317: Just a note: technically these `@font-face` rules don't need the `format`, as we do the the "if WOFF2 is supported, load it instead of the WOFF" ourselves in the `@supports`.

files/en-us/web/css/@supports/index.md
pepelsbey: I took the inspiration to call them WOFF 1 and WOFF 2 from the spec, but you're probably right about consistency.

files/en-us/web/css/@supports/index.md
```diff
@@ -275,44 +275,69 @@ ul:has(> li li) {
...
-  p {
+  body {
     font-family: "Bungee Spice", fantasy;
   }
 }
 ```
 
 It's also possible to test for the support of a font technology by using the `tech` function inside the {{CSSxRef("@font-face")}} at-rule.
-If a browser doesn't support the font technology, a fallback font (`Bungee-fallback.otf`) can be used instead.
+If a browser doesn't support the color font technology, a regular [Bungee](https://fonts.google.com/specimen/Bungee) font can be used instead.
```
RoelN: A nitpick perhaps, but it reads as if "a regular Bungee" is the alternative to use when any color font isn't supported :-)

Perhaps something like the original sentence, "...a regular font ([`bungee.woff2`](https://fonts.google.com/specimen/Bungee)) can be used instead."

files/en-us/web/css/@supports/index.md
```diff
@@ -275,44 +275,69 @@ ul:has(> li li) {
 
 ### Testing for the support of a font technology
 
-The following example applies the CSS style if the browser supports the `COLRv1` font technology:
+The following example applies the [Bungee Spice](https://fonts.google.com/specimen/Bungee+Spice) color font family if the browser supports the `COLRv1` font technology:
```
RoelN: Very nitpick, but you reference the _font_, not the entire _family_, so I'd probably omit "family" here.

files/en-us/web/css/@supports/index.md
pepelsbey: Well, it kinda is the point: if the color version is not supported, then use the monochrome one. No? :)

RoelN: Yeah, but this says, "if a color font is not supported, use Bungee" :-) Which I'm sure @djrrb wouldn't mind!

But it's a nitpick—feel free to just leave as-is!

pepelsbey: Ah, I see. I’ll try to make it more clear. Thanks!

RoelN:
@pepelsbey Thanks for these changes, and your patience! :-) I can't approve this PR via the Github interface, but this looks good to me 👍

files/en-us/web/css/@supports/index.md
```diff
@@ -275,44 +275,69 @@ ul:has(> li li) {
 
 ### Testing for the support of a font technology
 
-The following example applies the CSS style if the browser supports the `COLRv1` font technology:
+The following example applies the [Bungee Spice](https://fonts.google.com/specimen/Bungee+Spice) color font if the browser supports the `COLRv1` font technology:
```
bsmth, line 278: For later: might be interesting to look at using something from shared assets repo instead.

bsmth (approved):
TY!

PR #41703, opened 2025-10-27
docs: clarify responseStart behavior for canceled requests in Performance API

fengyinxqy:
### Description
Clarifies in the Performance API documentation that canceled HTTP requests are still recorded as performance entries, but their `responseStart` timestamp will be 0. This update helps developers avoid calculation errors when analyzing performance metrics for canceled requests.


sideshowbarker (approved):

sideshowbarker:
肖祺彦, 👍 nice update, and congrats on landing your first docs change here — welcome aboard 🎉

PR #41710, opened 2025-10-27
ci(workflows): pin 3rd party actions

caugner:
### Description

Pins all 3rd party GitHub Actions to specific commit hashes instead of version tags.

Each pinned action includes an inline comment with the resolved version number for reference.

### Motivation

Security best practice to pin actions to immutable commit hashes, preventing potential supply chain attacks from compromised action versions or tag hijacking.

### Additional details

See: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Part of https://github.com/mdn/fred/issues/1005.

### DIFF
--- a/.github/workflows/pr-review-companion.yml
+++ b/.github/workflows/pr-review-companion.yml
@@ -39,7 +39,7 @@ jobs:
       STATUS_TARGET: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
     steps:
       - name: Download artifact
-        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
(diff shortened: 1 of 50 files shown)


.github/workflows/auto-cleanup-bot.yml
```diff
@@ -5,7 +5,7 @@ on:
     - cron: "0 0 * * *"
   workflow_dispatch:
 
-# No GITHUB_TOKEN permissions, as we only use it to increase API limit.
+    # No GITHUB_TOKEN permissions, as we only use it to increase API limit.
```
caugner:
```suggestion
# No GITHUB_TOKEN permissions, as we only use it to increase API limit.
```

PR #41724, opened 2025-10-28
Fix: Updated `null` operator documentation

### LINKED ISSUE #41717: `null`: mention unusual `typeof` value on page
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null

 Just a suggestion (I could see arguments that this isn't really relevant), but the null page could have a `> [!NOTE]` mentioning that despite being a primitive value, null has a `typeof` value of `'object'` and linking to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null.

Not sure if this would make more sense by the `undefined` comparison (as this *is* a notable difference between the two values) or above.

### MDN metadata

<details>

* Folder: `en-us/web/javascript/reference/operators/null`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/operators/null/index.md

</details>

ragul1697:
### Description
Fixes #41717 - Added a note mentioning the `typeof null` is `object`.


files/en-us/web/javascript/reference/operators/null/index.md
```diff
@@ -33,30 +31,22 @@ null
...
+- The [`typeof null`](/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null) result is `"object"`. This is a bug in JavaScript that cannot be fixed due to backward compatibility.
+- Unlike `undefined`, {{jsxref("JSON.stringify()")}} can represent `null` faithfully.
+
+JavaScript is unique to have two nullish values: `null` and `undefined`. Semantically, their difference is very minor: `undefined` represents the absence of a value, while `null` represents the absence of an _object_. For example, the end of the [prototype chain](/en-US/docs/Web/JavaScript/Guide/Inheritance_and_the_prototype_chain) is `null` because the prototype chain is composed of objects; {{domxref("document.querySelector()")}} returns `null` if it doesn't match, because had it matched, the result would be an object. If you are designing an API, you should likely treat `null` and `undefined` as equivalent, because many codebases have stylistic rules about when to use `null` or `undefined` by default.
```
Josh-Cena:
```suggestion
JavaScript is unique to have two nullish values: `null` and `undefined`. Semantically, their difference is very minor: `undefined` represents the absence of a value, while `null` represents the absence of an _object_. For example, the end of the [prototype chain](/en-US/docs/Web/JavaScript/Guide/Inheritance_and_the_prototype_chain) is `null` because the prototype chain is composed of objects; {{domxref("document.querySelector()")}} returns `null` if it doesn't match, because had it matched, the result would be an object. If you are designing an API, you should likely accept `null` and `undefined` as equivalent inputs, because many codebases have stylistic rules about when to use `null` or `undefined` by default.
```

PR #41725, opened 2025-10-28
feature detection example with get

addsoupbase:
Added feature detection example for colorType parameter support in HTMLCanvasElement.getContext()

### Description

Added an example to detect features using a getter.

### Motivation

They can help developers see what is and is not supported in the browser.

### DIFF
--- a/files/en-us/web/javascript/reference/functions/get/index.md
+++ b/files/en-us/web/javascript/reference/functions/get/index.md
@@ -182,7 +182,7 @@ const obj = {
 
 ### Feature detection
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/javascript/reference/functions/get/index.md
```diff
@@ -198,15 +180,26 @@ const obj = {
 };
 ```
 
-### get vs. defineProperty
+### Feature detection
+
+Many functions accept an object and retrieve individual properties from it as separate parameters (this object parameter is known as an _options bag_). You can detect whether a specific option is supported by detecting if the property has been retrieved with a getter. This example checks if the `colorType` option is supported on the {{domxref("HTMLCanvasElement.getContext()")}} method.
```
Josh-Cena:
```suggestion
Many functions accept an object and retrieve individual properties from it as separate parameters (this object parameter is known as an _options bag_). You can detect whether a specific option is supported by using a getter to track if the property has been retrieved. This example checks if the `colorType` option is supported by the {{domxref("HTMLCanvasElement.getContext()")}} method.
```

PR #41730, opened 2025-10-29
Add note for exiting the text input example editor with keyboard

brysonbw:
#### Changes

- Add note for exiting the text input example editor with keyboard

### Before

image

### After

image

- https://github.com/mdn/fred/issues/983

### DIFF
--- a/files/en-us/web/html/reference/elements/input/text/index.md
+++ b/files/en-us/web/html/reference/elements/input/text/index.md
@@ -37,6 +37,9 @@ label {
 }
 ```
 
+> [!NOTE]
+> To exit the editor above using your keyboard, press `Esc`, then press the `Tab` key.
+
 ## Value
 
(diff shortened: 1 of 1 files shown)


Josh-Cena:
This is a platform issue, not a content issue. This issue applies to every single editor, so we can't just add notes in content.

brysonbw:
@Josh-Cena  Ok, smooth, thanks for the heads up. Has there been a change like this before? If so, what are the steps to apply such change? Is there a PR I could see for reference?

Josh-Cena:
You would need to edit https://github.com/mdn/fred, where the issue is posted. I don't know exactly how the dev team would like it—probably not as visible as an actual note, and definitely only visible when focused. You can prototype the code and ask under the issue.

brysonbw:
Ah I see, I should have known since it was in fred mybad lol. Thanks

PR #41736, opened 2025-10-29
Document ARIA attributes for SVG element

MarcusCaspeco:
### Description

Documents that various ARIA-attributes are valid for the `<svg>` element.

### Motivation

It is not clear on this page if the `<svg>` element supports any other attributes than the ones listed. Most other pages have something along the lines of:

> This element's attributes include the [global attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes).

(see e.g. [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#attributes),  [`<div>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/div#attributes))

I saw some examples on MDN (e.g. on the [`aria-label` page](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label)) use ARIA labels on the `<svg>` element, but it's not clear on this page if this is supported or not. I had to go into the specification to find out.

### Additional details

The ARIA-attributes are listed as supported in the [specification](https://svgwg.org/svg2-draft/struct.html#SVGElement).

None

### DIFF
--- a/files/en-us/web/svg/reference/element/svg/index.md
+++ b/files/en-us/web/svg/reference/element/svg/index.md
@@ -17,7 +17,7 @@ The **`<svg>`** [SVG](/en-US/docs/Web/SVG) element is a container that defines a
 
 ## Attributes
 
-This element's attributes include the ARIA [role](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles) attribute and various [aria-*](/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes) attributes.
+This element's attributes include the ARIA [role](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles) attribute and various [aria-\*](/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes) attributes.
 
 - {{SVGAttr("baseProfile")}} {{deprecated_inline}}
   - : The minimum SVG language profile that the document requires.


Josh-Cena:
Hello, ARIA attributes are allowed on all SVG elements, not just `<svg>`. The problem here is we don't have a list of global attributes in the style of https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes, which is linked for every element—but for example, would you believe that `<svg>` does not accept `id` or `class`, just because we didn't list them?

I think the correct approach here is to either keep the content as-is, or write up that section in https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute, and link to it for every element.

MarcusCaspeco:
> but for example, would you believe that <svg> does not accept id or class, just because we didn't list them?

Honestly speaking, I would definitely doubt my intuition (that they are allowed), since it is so clearly listed on the HTML element pages!

You're completely right that this applies to all SVG elements of course, I was a bit lazy. This is probably better as an issue instead of a PR, perhaps someone with more time on their hands can make a proper change. Feel free to close this PR, I've opened https://github.com/mdn/content/issues/41747

Josh-Cena:
Thanks! I'm going to close now. You are definitely welcome to work on that issue if/when you get the time.

PR #41737, opened 2025-10-29
Module landing page and guide: scroll-driven animations

estelle:
Part of https://github.com/openwebdocs/project/issues/147

### DIFF
--- a/files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
+++ b/files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
@@ -26,7 +26,7 @@ CSS scroll-driven animations are performant. JavaScript scroll-driven animations
 
 ## Foundations
 
-Scroll-driven animations build upon [CSS animations](/en-US/docs/Web/CSS/CSS_animations) and the [Web Animations API](/en-US/docs/Web/API/Web_Animations_API). Prior to creating scroll-driven animations, you must have an understanding of CSS {{cssxref("@keyframes")}} animations. See the [using CSS animations guide](/en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations) to learn more.
+Scroll-driven animations build upon [CSS animations](/en-US/docs/Web/CSS/CSS_animations) and the [Web Animations API](/en-US/docs/Web/API/Web_Animations_API). Before creating scroll-driven animations, you must have an understanding of CSS {{cssxref("@keyframes")}} animations. See the [using CSS animations guide](/en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations) to learn more.
 
 In CSS, animations are created by attaching keyframe animations to an element using the {{cssxref("animation-name")}} property (or {{cssxref("animation")}} shorthand). By default, animations run on the default document timeline, moving from the `from` keyframe to the `to` keyframe as time passes by, with the animation lasting as long as the time defined by the {{cssxref("animation-duration")}} property value. When set to run on the default document timeline, animations play through to completion unless prevented from doing so, for example, by having the {{cssxref("animation-play-state")}} set to `paused` or by removing the `animation-name` from the element.
 
@@ -100,7 +100,7 @@ When the checkbox is checked, the `action` animation is applied to the element. 
 
 {{EmbedLiveSample("regular", "100%", "150")}}
 
-Try checking the checkbox. Nothing will happen during the half-second animation delay. Then, once the animation starts, the box will jump to a 45 degree rotation, and then it will take 3 seconds to rotate an additional 720 degrees, or two additional full rotations. After a total of three and half seconds, the animation concludes, and the `<div>` will return to its default non-rotated state.
+Try checking the checkbox. Nothing will happen during the half-second animation delay. Then, once the animation starts, the box will jump to a 45-degree rotation, and then it will take 3 seconds to rotate an additional 720 degrees, or two additional full rotations. After a total of three and a half seconds, the animation concludes, and the `<div>` will return to its default non-rotated state.
 
 > [!NOTE]
 > The `animation-timeline` is reset to the default `auto` value by the {{cssxref("animation")}} shorthand property, but cannot be set using the shorthand. Therefore, when creating scroll-driven animations, always declare the `animation-timeline` after any `animation` shorthand declarations to achieve the desired effect.
@@ -115,7 +115,7 @@ To create a scroll progress timeline, the `animation-timeline` value must refere
 
 A _named scroll progress timeline_ is one where the scroller is explicitly named using the {{cssxref("scroll-timeline-name")}} property (or the {{cssxref("scroll-timeline")}} shorthand). The name is a {{cssxref("dashed-ident")}}. The scroller is linked to the element to be animated by specifying its `scroll-timeline-name` as the value of that element's `animation-timeline` property.
 
-Our HTML includes three elements: the `item`, which we will be animating; its `container`, which we will be scrolling; and the scroller. The `container` needs to be large enough to overflow its `scroller` parent: If there is no scrolling, there will be no scroll timeline.
+Our HTML includes three elements: the `item`, which we will animate; its `container`, which we will scroll; and the scroller. The `container` needs to be large enough to overflow its `scroller` parent: If there is no scrolling, there will be no scroll timeline.
 
 ```html live-sample___named_scroll live-sample___anon_scroll
 <main class="scroller">
@@ -180,7 +180,7 @@ Before any scrolling occurs, the container's position is at the top of the scrol
 
 You may have noticed that the {{cssxref("animation-duration")}} component of the `animation` shorthand was set to `1ms`. When creating [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations), specifying an `animation-duration` value does not affect the duration of the animation, and shouldn't be necessary. However, durations can affect non-linear view progress animation timelines, and Firefox requires a non-zero `animation-duration` to apply an animation to an element. For these reasons, it is common practice to set `animation-duration` to `1ms`.
 
-Setting `animation-durations: 1ms` ensures that the animation works in Firefox, that the animation effect is consistent across all browsers, and that the animation is not visible if a browser doesn't support view progress animation timelines. If the browser supports `@keyframe` animations, the animation will not be visible to the user, but the animation still happens, and animation events are fired.
+Setting `animation-duration: 1ms` ensures that the animation works in Firefox, the animation effect is consistent across all browsers, and the animation is hidden if a browser doesn't support view progress animation timelines. If the browser supports keyframe animations, the animation will not be visible to the user. However, the animation still happens, and animation events are fired.
 
 ### Anonymous scroll progress timelines
 
@@ -201,7 +201,7 @@ This example uses all the same CSS as the previous example, except for the `anim
 
 {{EmbedLiveSample("anon_scroll", "100%", "150")}}
 
-We set an {{cssxref("inline-size")}} on the container so that it overflows in the inline direction, and set the {{cssxref("block-size")}} to `100%`, so it no longer overflows in the block direction. Try scrolling in the inline direction.
+We set an {{cssxref("inline-size")}} on the container so that it overflows in the inline direction, and set the {{cssxref("block-size")}} to `100%` so it no longer overflows in the block direction. Try scrolling in the inline direction.
 
 ## View progress timelines
 
@@ -235,7 +235,7 @@ Here we use the {{cssxref("view-timeline-name")}} property to name an element, i
 
 We applied the animation **before** the animation timeline, as the `animation` resets the `animation-timeline` to `auto`.
 
-The animation is slightly different from the previous animations in that the spinning effect starts 20% through the animation and ends 80% of the way through the animation; this means the element will not be actively spinning when it first comes into view and will stop spinning before it is completely out of view.
+The animation is slightly different from the previous examples in that the spinning effect starts at `20%` and ends at `80%` of the way through the animation; this means the element will not be actively spinning when it first comes into view and will stop spinning before it is completely out of view.
 
 ```css live-sample___named_view live-sample___anon_view
 @keyframes action {
@@ -287,13 +287,13 @@ The animation is slightly different from the previous animations in that the spi
 
 {{EmbedLiveSample("named_view", "100%", "250")}}
 
-Scroll the element into view. Note that the element animates through the `@keyframe` animation as it moves through the visible area of its ancestor scroller.
+Scroll the element into view. Note that the element animates through the `@keyframes` animation as it moves through the visible area of its ancestor scroller.
 
 ### Anonymous view progress timeline: the `view()` function
 
 Alternatively, a {{cssxref("animation-timeline/view", "view()")}} function can be set as the value of the `animation-timeline` property to specify that an element's animation timeline is an _anonymous view progress timeline_. This causes the element to be animated based on its position inside its nearest parent scroller.
 
-The `view()` function is used to create a view timeline. You attach the timeline to the element you want to animate using the `animation-timeline` property. The function creates a view timeline for each element matched by the selector.
+The `view()` function creates a view timeline. You attach the timeline to the element you want to animate using the `animation-timeline` property. The function creates a view timeline for each element matched by the selector.
 
 In this example, we again define the `animation` before the `animation-timeline`, so the timeline is not reset. We then include an argument-less `view()` function. We don't specify a scroller, as, by definition, the subject's visibility is tracked by its nearest ancestor scroller.
 
@@ -311,7 +311,7 @@ In this example, we again define the `animation` before the `animation-timeline`
 The `view()` function takes up to three optional values as arguments:
 
 - Zero or one `<axis>` parameters. If set, this specifies the scroll axis along which the animation progresses.
-- Either the keyword `auto` or zero, one, or two {{cssxref("length-percentage")}} inset values. If set, these values specify offsets for the start and/or end of the scrollport.
+- Either the keyword `auto` or zero, one, or two {{cssxref("length-percentage")}} inset values. If set, these values specify offsets for the scrollport start and/or end.
 
 Declaring `view()` is equivalent to `view(block auto)`, which defines `block` as the axis of the parent element that supplies the timeline and the {{cssxref("scroll-padding")}}, which generally defaults to `0`, as the insets within the visible area at which the animation starts and ends.
 
@@ -357,7 +357,7 @@ Setting `animation-timeline: none` disassociates the element from all animation 
 }
 ```
 
-Because the [`animation` shorthand sets the `animation-timeline` to `auto`](#animation-timeline-and-the-animation-shorthand), use a selector with great enough specificity to ensure your `animation-timeline` isn't overridden by your `animation` shorthand declarations.
+Because the [`animation` shorthand sets the `animation-timeline` to `auto`](#animation-timeline-and-the-animation-shorthand), use a selector with enough specificity to ensure your `animation-timeline` isn't overridden by your `animation` shorthand declarations.
 
 ## See also
 


files/en-us/web/css/css_scroll-driven_animations/index.md
```diff
@@ -2,83 +2,169 @@
...
 
-- {{domxref("Element.animate()")}}
-- {{domxref("AnimationTimeline")}}
 - {{domxref("ScrollTimeline")}}
 - {{domxref("ViewTimeline")}}
 
+## Guides
+
+- [Scroll-driven animation timelines](/en-US/docs/Web/CSS/CSS_scroll-driven_animations/Timelines)
+  - : Scroll-driven animation timelines and creating scroll-driven animations
```
chrisdavidmills:
```suggestion
  - : Scroll-driven animation timelines and creating scroll-driven animations.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
+---
+title: Scroll-driven animation timelines
+slug: Web/CSS/CSS_scroll-driven_animations/Timelines
+page-type: guide
+spec-urls: https://drafts.csswg.org/scroll-animations-1/
+sidebar: cssref
+---
+
+A common UI pattern involves elements that are animated as the user scrolls up and down (or left and right) across a page. These animations are called _scroll-driven animations_ as the effects are in direct response to scrolling; be it scrolling a page, or just an overflowing scroll container within a page.
```
chrisdavidmills:
```suggestion
A common UI pattern involves elements that are animated as the user scrolls up and down (or left and right) across a page. These animations are called _scroll-driven animations_ because they occur in direct response to the scrolling of a page or an overflowing scroll container within a page.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+The properties defined in the [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module expand upon [CSS animations](/en-US/docs/Web/CSS/CSS_animations) by enabling animating property values defined in {{cssxref("@keyframes")}} animations in response to user interaction.
+
+This guide provides an overview of using CSS to scroll-driven animation timelines and creating scroll-driven animations.
```
chrisdavidmills:
```suggestion
This guide provides an overview of using CSS to create scroll-driven animation timelines and animations.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+## What is scroll-driven animation
```
chrisdavidmills:
```suggestion
## What is scroll-driven animation?
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+The [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module defines properties that enable linking [CSS keyframe animations](/en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations#defining_an_animation_sequence_using_keyframes) to scrolling.
```
chrisdavidmills:
```suggestion
The [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module defines properties that enable [CSS keyframe animations](/en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations#defining_an_animation_sequence_using_keyframes) to be linked to scrolling.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+The properties defined in the [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module expand upon [CSS animations](/en-US/docs/Web/CSS/CSS_animations) by enabling animating property values defined in {{cssxref("@keyframes")}} animations in response to user interaction.
+
+This guide provides an overview of using CSS for scroll-driven animation timelines and creating scroll-driven animations.
+
+## What is scroll-driven animation
+
+The [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module defines properties that enable linking [CSS keyframe animations](/en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations#defining_an_animation_sequence_using_keyframes) to scrolling.
+
+Animations can be set to progress along a scroll-based timeline instead of the default time-based document timeline, without needing JavaScript. This means that, by [defining which animation timeline](#animation_timelines) to use, you can animate an element by scrolling a scrollable element, rather than just by the passing of time with CSS.
```
chrisdavidmills: This whole section feels repetitive, wordy, and hard to follow. It needs to be cut down to explain what scroll-driven animations are in terms of the effect created, without including all the properties and other code features required to implement them. You explain those later on.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+Scroll-driven animations are CSS animations that are not run on the default [DocumentTimeline](/en-US/docs/Web/API/DocumentTimeline); rather, on a scroll-progress or view-progress timeline, which is driven by user scrolling.
+
+In scroll-driven animations, there's a direct link between the user's scrolling action and the animation's progress along the `@keyframe` keyframes. Instead of progressing through animation keyframes over time, the scroll-driven animation progresses based on the scroll position of overflowing content within its scroll port. As the user scrolls up, down, left, or right, the animation moves forward or backward through the keyframe progression. When scrolling is paused, the animation pauses, as if `animation-play-state` were set to `pause`.
+
```
chrisdavidmills, line 34: Maybe add a section that summarizes what you need to create a scroll-driven animation, and then walk through each step in the subsequent sections. You kind of do this below, but it would be nice to help the reader understand all that's involved first with a summary. It is pretty complex stuff, after all.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+## Animation timelines
+
+All animations run on a timeline. In CSS, you can have regular CSS animations that progress over time, or you can set animations to progress based on scrolling, using either a _scroll progress timeline_ or a _view progress timeline_.
```
chrisdavidmills: It feels like you make this point several times earlier on.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+The {{cssxref("animation-timeline")}} property, defined in the [CSS animations](/en-US/docs/Web/CSS/CSS_animations) module, is used to set the timeline used for the animation. The [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module extends the concept of animation timelines, defining properties that enable animating elements based on scroll position and element visibility.
+
+You can define an element to drive another element's animation progression by explicitly [naming an element as a timeline controller](#named_timelines) using the `scroll-timeline-*` and `view-timeline-*` properties, then setting that name as the `animation-timeline` of a descendant element. You can also define _anonymous scroll progress timelines_ and _anonymous view progress timelines_ using the [`scroll()`](#scroll-progress_timelines) and [`view()`](#view-progress_timelines)) functions.
```
chrisdavidmills: This stuff makes sense, but then I'd expect to jump straight into building examples of these. Instead, you then show how to create a normal CSS animation. I think you should cut this down to briefly making the point about the default document timeline, then go on to say that scroll-driven animations don't use this, and move on to building up an example.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+  left: calc(50% - (var(--size) / 2));
+  top: calc(50% - (var(--size) / 2));
+}
+```
+
+When the checkbox is checked, the `action` animation is applied to the element. When unchecked, the animation is not applied to the `<div>`.
+
+{{EmbedLiveSample("regular", "100%", "150")}}
+
+Try checking the checkbox. Nothing will happen during the 3-second animation delay. Then, once the animation starts, the box will jump to a 45 degree rotation, and then it will take 3 seconds to rotate an additional 720 degrees, or two additional full rotations. After a total of six seconds, the animation concludes, and the `<div>` will return to its default non-rotated state.
```
chrisdavidmills: I'd prefer to get rid of this example altogether, but as a sidenote, the example appears broken because nothing happens when you click the checkbox. Given that you've got a control to start/stop it, you don't need the animation delay.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+### `animation-timeline` and the `animation` shorthand
+
+The `animation-timeline` is reset to the default `auto` value by the {{cssxref("animation")}} shorthand property, but cannot be set using the shorthand. Therefore, when creating scroll-driven animations, always declare the `animation-timeline` after any `animation` shorthand declarations to achieve the desired effect.
```
chrisdavidmills: This seems like a weird point to start with. I'd expect to see you jump straight into building up an example, and then mention this as an aside later on to draw attention to it.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Scroll progress timelines
+
+With _scroll progress timeline_, the timeline progresses based on the scrolling of the scrollable element (_scroller_) from top to bottom (or left to right) and back again. By default, the position in the scroll range is converted into a percentage of progress — `0%` at the start and `100%` at the end. This [animation range can be controlled](#controlling_the_animation_range) via the {{cssxref("animation-range")}} properties.
```
chrisdavidmills: It isn't very clear what the animation range is, or what controlling it means. Maybe omit that detail for now, so the reader can just concentrate on seeing a basic example up and running?

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+To create a scroll progress timeline, the `animation-timeline` value must reference the scroller. The _scroller_, the element that provides the scroll progress timeline, can be named or anonymous.
+
+#### Named scroll progress timelines
+
+A _named scroll progress timeline_ is one where the scroller is explicitly named using the {{cssxref("scroll-timeline-name")}} property (or the {{cssxref("scroll-timeline")}} shorthand). The name is a {{cssxref("dashed-ident")}}. This name is applied to the scroller, the {{glossary("scroll container")}} whose scrolling controls the progress of the animation timeline. The scroller is linked to the element to be animated by specifying its name as the value of that element's `animation-timeline` property.
```
chrisdavidmills:
```suggestion
A _named scroll progress timeline_ is one where the scroller (the {{glossary("scroll container")}} whose scrolling controls the progress of the animation timeline) is explicitly named using the {{cssxref("scroll-timeline-name")}} property (or the {{cssxref("scroll-timeline")}} shorthand). The name is a {{cssxref("dashed-ident")}}. The scroller is linked to the element to be animated by specifying its `scroll-timeline-name` as the value of that element's `animation-timeline` property.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+Our HTML includes three elements: the `item`, which we will be animating; its `container`, which we will be scrolling; and the' scroller', which will need to be large enough to overflow its `scroller` parent.
```
chrisdavidmills:
```suggestion
Our HTML includes three elements: the `item`, which we will be animating; its `container`, which we will be scrolling; and the scroller. The `container` needs to be large enough to overflow its `scroller` parent: If there is no scrolling, there will be no scroll timeline.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+```html live-sample___named_scroll live-sample___anon_scroll
+<main class="scroller">
+  <div class="container">
+    <span class="item"></span>
+  </div>
+</main>
+```
+
+We provide some basic styles, which we've hidden for brevity. The important styles include setting a height on the container that is taller than the scroller, and then setting the overflow to allow scrolling (if there is no scrolling, there will be no scroll timeline):
```
chrisdavidmills:
```suggestion
We provide some basic styles, which we've hidden for brevity. The important styles include setting a height on the container that is taller than the scroller, and then setting the overflow to allow scrolling:
```

Moved to the other suggestion above.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+  width: var(--size);
+  background-color: magenta;
+  border: 1px solid;
+  position: absolute;
+  left: calc(50% - (var(--size) / 2));
+  top: calc(50% - (var(--size) / 2));
+}
+```
+
+In this case, we don't have a checkbox, as the `action` animation progression is controlled by the scrolling of the overflowing scroller, which, unlike time, does not expire.
```
chrisdavidmills, line 173: If you remove the regular animation example, you won't need to say this.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+{{EmbedLiveSample("named_scroll", "100%", "150")}}
+
+Before any scrolling occurs, the container's position is at the top of the scroller, and the animation is at the 0% keyframe. Try scrolling down. As you scroll, the animation progresses through the timeline, rotating an additional 720 degrees. When you can no longer scroll, the animation's progression is at the 100%, or `to`, keyframe. The animated item never returns to its default rotation.
+
+##### Animation duration
+
+You may have noticed that the {{cssxref("animation-duration")}} component of the `animation` shorthand was set to `1ms`. When creating [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations), specifying an `animation-duration` value has no effect on the duration of the animation, and shouldn't be necessary. However, while a duration has no effect on scroll progress timeline animations, durations can have an effect on non-linear view progress animation timelines, and Firefox requires a non-zero `animation-duration` to apply an animation to an element. For these reasons, it is common practice to set `animation-duration` to `1ms`.
```
chrisdavidmills:
```suggestion
You may have noticed that the {{cssxref("animation-duration")}} component of the `animation` shorthand was set to `1ms`. When creating [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations), specifying an `animation-duration` value does not affect the duration of the animation, and shouldn't be necessary. However, durations can affect non-linear view progress animation timelines, and Firefox requires a non-zero `animation-duration` to apply an animation to an element. For these reasons, it is common practice to set `animation-duration` to `1ms`.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+Setting `animation-durations: 1ms` ensures that the animation works in Firefox, that the animation effect is consistent across all browsers, and that the animation is not visible if a browser doesn't support view progress animation timelines. If the browser supports `@keyframe` animations, the animation will not be visible to the user, but the animation still happens, and animation events are fired.
+
+#### Anonymous scroll progress timelines
+
+You don't have to name your scroll progress timeline. Instead, you can associate an _anonymous scroll progress timeline_. In this case, the `animation-timeline` of the element to animate is set to a {{cssxref("animation-timeline/scroll", "scroll()")}} function. The function selects the scroller that provides the scroll progress timeline and the scroll axis to use based on the optional arguments you pass to it. One parameter is a [`<scroller>`](/en-US/docs/Web/CSS/animation-timeline/scroll#scroller) keyword defining the relationship of the scroller element to the current element (`nearest`, `root`, or `self`). The other is the scrollbar [`<axis>`](/en-US/docs/Web/CSS/animation-timeline/scroll#axis) value (`block`, `inline`, `y`, or `x`).
```
chrisdavidmills:
```suggestion
You don't have to name your scroll progress timeline. Instead, you can associate an _anonymous scroll progress timeline_ with the animation. In this case, the `animation-timeline` of the element to animate is set to a {{cssxref("animation-timeline/scroll", "scroll()")}} function. The function selects the scroller that provides the scroll progress timeline and the scroll axis to use based on the optional arguments you pass to it. One parameter is a [`<scroller>`](/en-US/docs/Web/CSS/animation-timeline/scroll#scroller) keyword defining the relationship of the scroller element to the current element (`nearest`, `root`, or `self`). The other is the scrollbar [`<axis>`](/en-US/docs/Web/CSS/animation-timeline/scroll#axis) value (`block`, `inline`, `y`, or `x`).
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+We use all the same CSS as in the previous examples, except for the `animation-timeline`, which we set to a `scroll()` function. We are also overriding the size of the container to change the direction of the scroll:
```
chrisdavidmills:
```suggestion
This example uses all the same CSS as the previous example, except for the `animation-timeline`, which we set to a `scroll()` function. We are also overriding the size of the container to change the direction of the scroll:
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+}
+.container {
+  inline-size: 800px;
+  block-size: 100%;
+}
+```
+
+{{EmbedLiveSample("anon_scroll", "100%", "150")}}
+
+We set an {{cssxref("inline-size")}} on the container so that in overflows in the inline direction, and set the {{cssxref("block-size")}} to `100%`, so it no longer overflows the container. Try scrolling in the inline direction.
```
chrisdavidmills:
```suggestion
We set an {{cssxref("inline-size")}} on the container so that it overflows in the inline direction, and set the {{cssxref("block-size")}} to `100%`, so it no longer overflows in the block direction. Try scrolling in the inline direction.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+### View progress timelines
+
+You can also progress an animation based on the change in visibility of an element inside a scroller. The _view progress timeline_ is a type of scroll progress timeline; it responds to scroll, but instead of tracking the scroll offset of a scroll container, it tracks the relative position of an element, called the _subject_, within a scrollport. The progression of the animation is based on the _visibility_ of the subject inside the scroller. With view progress timelines, you can't specify the scroller — the subject's visibility is always tracked within its nearest ancestor scroller, which is different from scroll progress timelines.
```
chrisdavidmills:
```suggestion
You can also progress an animation based on the change in visibility of an element inside a scroller — this is done via _view progress timelines_. Instead of tracking the scroll offset of a scroll container, view progress timelines track the relative position of an element, called the _subject_, within a scrollport. The progression of the animation is based on the _visibility_ of the subject inside the scroller. Unlike scroll progress timelines, with view progress timelines, you can't specify the scroller — the subject's visibility is always tracked within its nearest ancestor scroller.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+When a view progress timeline is applied to an animated element, the animation progresses along that timeline rather than the default time-based one, with the animation's keyframe progression reflecting the subject's position in the scroller. The animation only occurs when the element is visible within its scrollport, with the current point in the keyframe animation being determined by the element's scroll position within that scrollport. Timeline progress starts at `0%` when the tracked subject starts intersecting the scrollport at the block or inline end edge. The `100%` occurs when the subject exits the scrollport at the block or inline start edge.
```
chrisdavidmills:
```suggestion
When a view progress timeline is applied to an animated element, the animation's keyframe progression reflects the subject's position in the scroller. The animation only occurs when the element is visible within its scrollport. Timeline progress starts at `0%` when the tracked subject starts intersecting the scrollport at the block or inline end edge. The `100%` occurs when the subject exits the scrollport at the block or inline start edge.
```

Cutting down on repetition. Also, the last couple of sentences are not necessarily true — you can vary the range of view progress timeline animations too.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+By default, the timeline is at `0%` when the subject is first visible at the start edge of the scroller, and `100%` when the element's end edge reaches the end edge of its nearest scroll container ancestor. Because the `100%` is only reached when the element leaves the viewport, when creating keyframe animations for view-timed animations, you likely want to set the final effect of your animation in a keyframe block that occurs well before the end of the animation. You can set your completed effect within the 20%, 50%, or 80% keyframe block rather than using the `to` or `100%` keyframe to ensure the element finishes animating while still in view.
```
chrisdavidmills:
```suggestion
Because the `100%` is generally reached when the element leaves the viewport, you likely want to set the final effect of your animation in a keyframe block that occurs well before the end of the animation. You can set your completed effect within the 20%, 50%, or 80% keyframe block rather than using the `to` or `100%` keyframe to ensure the element finishes animating while still in view.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+With view progress timelines, you can adjust the view progress visibility range.
+Use {{cssxref("view-timeline-inset")}}, part of the {{cssxref("view-timeline")}} shorthand, to adjust when the subject is considered to be in view. The default value is `auto`. The effect of any non-`auto` inset value is as if you moved the edges of the scroll port: a positive inset value creates an inward adjustment and a negative value an outward adjustment.
```
chrisdavidmills:
```suggestion
Use {{cssxref("view-timeline-inset")}}, part of the {{cssxref("view-timeline")}} shorthand, to adjust when the subject is considered to be in view. The default value is `auto`. The effect of any non-`auto` inset value is as if you moved the edges of the scroll port: a positive inset value creates an inward adjustment, and a negative value creates an outward adjustment.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+  <div class="item"></div>
+  <p>&nbsp;</p>
+  <p>&nbsp;</p>
+  <p>&nbsp;</p>
+  <p>&nbsp;</p>
+  <p>Scroll up to view the animation</p>
+</main>
+```
+
+{{EmbedLiveSample("named_view", "100%", "150")}}
```
chrisdavidmills: For these examples, I would scroll in the inline direction and adjust the animation to make it a bit smoother. At present, the animation looks really choppy, and the window of animation is so small that it is hard to clearly see where the animation starts and finishes.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+Scroll the element into view. Note that the element animates through the `@keyframe` animation as it moves through the visible area of its ancestor scroller.
+
+#### Anonymous view progress timeline: the `view()` function
+
+Alternatively, the CSS {{cssxref("animation-timeline/view", "view()")}} function can be used to specify that the `animation-timeline` for an element is a view progress timeline. An _anonymous view progress timeline_ is one in which the subject is given a `view()` function as an `animation-timeline` value, causing it to be animated based on its position inside its nearest parent scroller.
```
chrisdavidmills:
```suggestion
Alternatively, a {{cssxref("animation-timeline/view", "view()")}} function can be set as the value of the  `animation-timeline` property to specify that an element's animation timeline is an _anonymous view progress timeline_. This causes the element to be animated based on its position inside its nearest parent scroller.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+The `view()` function is used to create a view timeline. You attach the timeline to the element you want to animate using the `animation-timeline` property. The function creates a view timeline for each element matched by the selector.
```
chrisdavidmills: Do we need this paragraph? We already said all this in the previous paragraph.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+Again, we define the `animation` before the `animation-timeline`, so as to not reset the timeline. We then include an argument-less `view()` function. We don't specify a scroller as, by definition, the subject's visibility is tracked by its nearest ancestor scroller.
```
chrisdavidmills:
```suggestion
In this example, we again define the `animation` before the `animation-timeline`, so the timeline is not reset. We then include an argument-less `view()` function. We don't specify a scroller, as, by definition, the subject's visibility is tracked by its nearest ancestor scroller.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+  animation: action 1ms linear;
+  animation-timeline: view();
+}
+```
+
+{{EmbedLiveSample("anon_view", "100%", "150")}}
+
+#### Parameters of the `view()` function
+
+The `view()` function takes up to three optional values as arguments: zero or one `<axis>` parameters and either the keyword `auto` or zero, one, or two {{cssxref("length-percentage")}} inset values. Declaring `view()` is equivalent to `view(block auto)`, which defines `block` as the axis of the parent element that supplies the timeline and the {{cssxref("scroll-padding")}}, which generally defaults to `0`, as the insets within the visible area at which the animation starts and ends. The function sets the values of the {{cssxref("view-timeline-axis")}} and {{cssxref("view-timeline-inset")}} properties.
```
chrisdavidmills:
```suggestion
The `view()` function takes up to three optional values as arguments:

- Zero or one `<axis>` parameters. If set, this specifies the scroll axis along which the animation progresses.
- Either the keyword `auto` or zero, one, or two {{cssxref("length-percentage")}} inset values. If set, these values specify offsets for the start and/or end of the scrollport.

Declaring `view()` is equivalent to `view(block auto)`, which defines `block` as the axis of the parent element that supplies the timeline and the {{cssxref("scroll-padding")}}, which generally defaults to `0`, as the insets within the visible area at which the animation starts and ends.

The function sets the values of the {{cssxref("view-timeline-axis")}} and {{cssxref("view-timeline-inset")}} properties.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+The {{cssxref("view-timeline-inset")}} argument can either be set to `auto` or up to two {{cssxref("length-percentage")}} values. It specifies an inset (positive) or outset (negative) adjustment of the scrollport. The inset is used to determine whether the element is in view, which determines the length of the animation timeline. In other words, instead of starting at the start edge and ending at the end edge of the scrollport, the animation lasts as long as the element is in the inset-adjusted view.
```
chrisdavidmills:
```suggestion
The {{cssxref("view-timeline-inset")}} arguments specify insets (if positive) or outsets (if negative) that adjust the start and end of the scrollport. They are used to determine whether the element is in view, which determines the length of the animation timeline. In other words, instead of starting at the start edge and ending at the end edge of the scrollport, the animation lasts as long as the element is in the inset-adjusted view.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+Unlike the scroll timeline's `scroll()` function, there is no `<scroller>` argument in the `view()` function, as the view timeline always tracks the subject within its nearest ancestor scroll container.
+
+In this case, as we are using inset values, we can use the `from` and `to` keyframe selectors.
```
chrisdavidmills:
```suggestion
In this example, as we are using inset values, we can use the `from` and `to` keyframe selectors.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+}
+```
+
+{{EmbedLiveSample("anon_view_args", "100%", "150")}}
+
+## Accessibility concerns
+
+As with all animations and transitions, always take any user's [`prefers-reduced-motion`](/en-US/docs/Web/CSS/@media/prefers-reduced-motion) preference into account.
+
+### Removing an animation's timeline
```
chrisdavidmills, line 348:
```suggestion
## Removing an animation's timeline
```

chrisdavidmills: I don't see this as just an accessibility topic — you might want to remove the timeline for other purposes, e.g. if the UI simply requires it.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+Setting `animation-timeline: none` disassociates the element from all animation timelines, including the default time-based document timeline, meaning the element will not animate. While some animations may be necessary, you can remove animations based on the user's `prefers-reduced-motion` setting with:
+
+```css
+@media (prefers-reduced-motion: reduce) {
```
chrisdavidmills, line 353: Maybe expand this to show an example without `prefers-reduced-motion` first, making the point that if required, you can remove an animation by adding such a class, and then go on to say that it makes sense to stick it inside a `prefers-reduced-motion` MQ for a11y purposes.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+> [!NOTE]
+> Scroll-driven animations should not be confused with _scroll-triggered animations_, which are regular animations programmed to apply to elements based on scrolling. These document timeline-based animations are initially paused, inactive, or non-existent, and are initiated by scroll interactions that add an `animation-name` to an element, thereby initiating the animation, changing the `animation-play-state` to a value other than `paused`, or otherwise triggering an animation on an element. Scroll-triggered animations do not use features of the [scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module. Rather, once the user reaches a certain scroll offset, the animation is triggered and, as with regular CSS animations, the animation progresses through the keyframes as a timed animation normally does. Scrolling only initiates, or triggers, the animation; it has no effect on the animation progression once triggered. For example, scrolling back will not stop or reverse the scroll-triggered animation.
```
chrisdavidmills: This feels very verbose, and I'm not sure if it is quite accurate. Scroll-triggered animations are triggered via JS, and often they ARE programmed so that they will reverse if the user scrolls back. JS is used to set the position of an element based on the scroll position, rather than triggering a CSS animation in such cases. Maybe just say that scroll-triggered animations are controlled by JS, and don't use CSS scroll-driven animations...?

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
chrisdavidmills: Update — you don't actually cover the range stuff at all in this article. I think that's OK, as it is quite long already. But update it to remove dead references/links for now. I'd cover these in a separate article, later on.

chrisdavidmills (changes requested):
@estelle A good opening effort to tackle a very complex subject. I've got a bunch of suggestions for you, but this is going in the right direction.

Major things you don't cover:

- The animation range stuff. You'll need to update some links and references in light of this, as it looks like you meant to cover it, but then didn't.
- The WAA/JS API equivalent of this functionality.

I think both of these should be presented in follow-on articles at a later date. The existing article has enough in its scope as-is, for now.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle: Removing the timeline is part of the a11y example, so it fits under this section.

It is mentioned in animation-timeline page under `none`, and isn't the point of this page, so doesn't need to be promoted above all other timelines. There are very few h2s in this page, so making this an H2 would make it the only

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle: it's in the a11y section.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle:
```suggestion
```

estelle: removed. It wasn't really relevant to the page.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle: intentionally keeping the examples in teh same direction to show two ways of doing the same thing

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle: keeping regular animation example

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle: the article was so long i cut out the animation range stuff to put in a second guide which i'll create when i get back from Japan.

estelle:
```suggestion
With _scroll progress timeline_, the timeline progresses based on the scrolling of the scrollable element (_scroller_) from top to bottom (or left to right) and back again. By default, the position in the scroll range is converted into a percentage of progress — `0%` at the start and `100%` at the end. <!--This [animation range can be controlled](#controlling_the_animation_range) via the {{cssxref("animation-range")}} properties.-->
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle:
```suggestion
>[!NOTE]
>The `animation-timeline` is reset to the default `auto` value by the {{cssxref("animation")}} shorthand property, but cannot be set using the shorthand. Therefore, when creating scroll-driven animations, always declare the `animation-timeline` after any `animation` shorthand declarations to achieve the desired effect.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle:
```suggestion
Try checking the checkbox. Nothing will happen during the half-second animation delay. Then, once the animation starts, the box will jump to a 45 degree rotation, and then it will take 3 seconds to rotate an additional 720 degrees, or two additional full rotations. After a total of three and half seconds, the animation concludes, and the `<div>` will return to its default non-rotated state.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+By default, CSS animation progression is based on the time-based document timeline.
+
+Setting `animation-timeline` explicitly to `auto`, or omitting the property and allowing it to default to `auto`, sets the timeline to be the default document timeline. When set to this default value, the animation's progress is determined by the {{cssxref("animation-duration")}}, the {{cssxref("animation-delay")}}, and how much time has passed since the animation was associated with the element via the `animation-name` property. The time-based document timeline is the timeline traditionally associated with CSS animations.
+
+```css live-sample___regular
+:checked ~ .container > .item {
+  animation-name: action;
+  animation-duration: 3s;
+  animation-delay: 3s;
```
estelle:
```suggestion
  animation-delay: 500ms;
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle: keeping the delay as it makes it clearer that the animation is based on time.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle:
```suggestion
By default, CSS animations follow the time-based document timeline; but animation progression doesn't need to be based only on time. You can define an element to drive another element's animation progression by explicitly [naming an element as a timeline controller](#named_timelines) using the `scroll-timeline-*` and `view-timeline-*` properties, then setting that name as the `animation-timeline` of a descendant element. You can also define _anonymous scroll progress timelines_ and _anonymous view progress timelines_ using the [`scroll()`](#scroll-progress_timelines) and [`view()`](#view-progress_timelines)) functions.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,364 @@
+---
+title: Scroll-driven animation timelines
+slug: Web/CSS/CSS_scroll-driven_animations/Timelines
+page-type: guide
+spec-urls: https://drafts.csswg.org/scroll-animations-1/
+sidebar: cssref
+---
+
+A common UI pattern involves animating elements in direct response to scrolling, as the user scrolls across a page. These _scroll-driven animations_ have been implemented in many places, including effects that respond to user scrolling up and down a page, or left and right on overflowing content.
```
estelle:
```suggestion
A common UI pattern involves elements that animate as the user scrolls vertically or horizontally across a page. These _scroll-driven animations_ occur in direct response to page scrolling or an overflowing scroll container within a page.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,364 @@
...
+
+The properties defined in the [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module expand upon [CSS animations](/en-US/docs/Web/CSS/CSS_animations) by enabling animating property values defined in {{cssxref("@keyframes")}} animations in response to user interaction.
+
+This guide provides an overview of using CSS for scroll-driven animation timelines and creating scroll-driven animations.
```
estelle:
```suggestion
This guide provides an overview of using CSS to create scroll-driven animation timelines and animations.
```

files/en-us/web/css/animation-timeline/scroll/index.md
```diff
@@ -39,7 +39,7 @@ animation-timeline: scroll(x self);
 
 ### Parameters
 
-- scroller
+- `<scroller>`
```
estelle:
```suggestion
- scroller
```

files/en-us/web/css/animation-timeline/scroll/index.md
```diff
@@ -48,7 +48,7 @@ animation-timeline: scroll(x self);
     - `self`
       - : The current element itself.
 
-- axis
+- `<axis>`
```
estelle:
```suggestion
- axis
```

files/en-us/web/css/css_scroll-driven_animations/index.md
```diff
@@ -2,83 +2,169 @@
...
+  - {{cssxref("view-timeline-axis")}}
+  - {{cssxref("view-timeline-inset")}}
+  - {{cssxref("view-timeline-name")}}
 
-### At-rules and descriptors
+### Data types and values
 
-CSS scroll-driven animations adds the ability to include `<timeline-range-name>`s in {{cssxref("@keyframes")}} blocks, to place keyframes at specific positions inside named timeline ranges.
+- [`<axis>`](/en-US/docs/Web/CSS/animation-timeline/scroll#axis)
+- [`<scroller>`](/en-US/docs/Web/CSS/animation-timeline/scroll#scroller)
```
estelle:
```suggestion
<!--
- {{cssxref("axis")}}
- {{cssxref("scroller")}} -->
```

files/en-us/web/css/css_scroll-driven_animations/index.md
```diff
@@ -2,83 +2,170 @@
...
+  - {{cssxref("view-timeline-inset")}}
+  - {{cssxref("view-timeline-name")}}
 
-### At-rules and descriptors
+### Data types and values
 
-CSS scroll-driven animations adds the ability to include `<timeline-range-name>`s in {{cssxref("@keyframes")}} blocks, to place keyframes at specific positions inside named timeline ranges.
+<!--
+- {{cssxref("axis")}}
+- {{cssxref("scroller")}} -->
```
estelle, line 132: This is commented out as these are going to be added in the next PR

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+This guide provides an overview of using CSS to create scroll-driven animation timelines and animations.
+
+## What is scroll-driven animation?
+
+The [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module defines properties that enable [CSS keyframe animations](/en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations#defining_an_animation_sequence_using_keyframes) to be linked to scrolling.
+
+### Timeline progression
+
+Animations can be set to progress along a _scroll-based timeline_ instead of the default time-based document timeline, without needing JavaScript. CSS enables us to [define which animation timeline](#animation_timelines) to use, including animating elements by scrolling a scrollable element, rather than just by the passing of time.
```
chrisdavidmills:
```suggestion
Animations can be set to progress along a _scroll-based timeline_ instead of the default time-based document timeline, without needing JavaScript. CSS enables us to [define which animation timeline](#animation_timelines) to use, including animating elements by scrolling a scrollable element rather than by the passing of time.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+### Performance benefits
+
+CSS scroll-driven animations are performant. JavaScript scroll driven animations require [`scroll`](/en-US/docs/Web/API/Document/scroll_event) event listeners and {{domxref("IntersectionObserver")}} objects on the {{glossary("main thread")}} to track elements across the [scrollport](/en-US/docs/Glossary/Scroll_container#scrollport). Any time you rely on the main thread to render effects with JavaScript, you run the risk of blocking the main thread,hich can lead to an unresponsive page and a bad user experience, or {{glossary("jank")}}. Since CSS scroll-driven animations are native to the web platform, JavaScript isn't necessary.
```
chrisdavidmills:
```suggestion
CSS scroll-driven animations are performant. JavaScript scroll-driven animations require [`scroll`](/en-US/docs/Web/API/Document/scroll_event) event listeners and {{domxref("IntersectionObserver")}} objects on the {{glossary("main thread")}} to track elements across the [scrollport](/en-US/docs/Glossary/Scroll_container#scrollport). Any time you rely on the main thread to render effects with JavaScript, you run the risk of blocking the main thread, which can lead to an unresponsive page and a bad user experience, or {{glossary("jank")}}.
```

We already said "without needing JavaScript" earlier on.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+### Foundations
+
+Scroll driven animations build upon [CSS animations](/en-US/docs/Web/CSS/CSS_animations) and the [Web Animations API](/en-US/docs/Web/API/Web_Animations_API). Prior to creating scroll-driven animations, you must have an understanding of CSS {{cssxref("@keyframes")}} animations. See the [using CSS animations guide](/en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations) to learn more.
+
+In CSS, animations created by attaching keyframe animations to an element using the {{cssxref("animation-name")}} property (or {{cssxref("animation")}} shorthand). By default, animations run on the default document timeline, moving from the `from` keyframe to the `to` keyframe as time passes by, with the animation lasting as long as the time defined by the {{cssxref("animation-duration")}} property value. When set to run on the default document timeline, animations play through completion unless prevented, such as by having the {{cssxref("animation-play-state")}} set to `paused` or by removing the `animation-name` from the element.
```
chrisdavidmills:
```suggestion
In CSS, animations are created by attaching keyframe animations to an element using the {{cssxref("animation-name")}} property (or {{cssxref("animation")}} shorthand). By default, animations run on the default document timeline, moving from the `from` keyframe to the `to` keyframe as time passes by, with the animation lasting as long as the time defined by the {{cssxref("animation-duration")}} property value. When set to run on the default document timeline, animations play through to completion unless prevented from doing so, for example, by having the {{cssxref("animation-play-state")}} set to `paused` or by removing the `animation-name` from the element.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+Scroll-driven animations are CSS animations that are not run on the default [DocumentTimeline](/en-US/docs/Web/API/DocumentTimeline); rather, on a scroll-progress or view-progress timeline, which is driven by user scrolling.There's a direct link between the user's scrolling action and the animation's progress along the `@keyframe` keyframes. Instead of progressing through animation keyframes over time, the scroll-driven animation progresses based on the scroll position of overflowing content within its scroll port. As the user scrolls up, down, left, or right, the animation moves forward or backward through the keyframe progression. When scrolling is paused, the animation pauses, as if `animation-play-state` were set to `pause`.
```
chrisdavidmills:
```suggestion
Scroll-driven animations are CSS animations that are not run on the default [DocumentTimeline](/en-US/docs/Web/API/DocumentTimeline). Instead, they run on a scroll-progress or view-progress timeline, which is driven by the scrolling of an element's contents. There's a direct link between the user's scrolling action and the animation's progress along the `@keyframe` keyframes. As the user scrolls up, down, left, or right, the animation moves forward or backward through the keyframe progression. When scrolling is paused, the animation pauses, as if `animation-play-state` were set to `pause`.
```

chrisdavidmills: I am suggesting removing that sentence because it seems repetitive.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Animation timelines
+
+The {{cssxref("animation-timeline")}} property, defined in the [CSS animations](/en-US/docs/Web/CSS/CSS_animations) module, is used to set the timeline used for the animation. By default, CSS animations follow the time-based document timeline. The [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module extends the concept of animation timelines, defining properties that enable animating elements based on scroll position and element visibility, called _scroll progress timelines_ and _view progress timelines_, respectively.
```
chrisdavidmills: Again, seems repetitive. You've already said much of this before, multiple times. How about:

```suggestion
The {{cssxref("animation-timeline")}} property, defined in the [CSS animations](/en-US/docs/Web/CSS/CSS_animations) module, is used to set the timeline used for the animation.
```

Then add some of this module definitions stuff to the next para (see next suggestion).

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+You can define an element to drive another element's animation progression by explicitly [naming an element as a timeline controller](#named_timelines) using the `scroll-timeline-*` and `view-timeline-*` properties, then setting that name as the `animation-timeline` of a descendant element. You can also define _anonymous scroll progress timelines_ and _anonymous view progress timelines_ using the [`scroll()`](#scroll-progress_timelines) and [`view()`](#view-progress_timelines)) functions.
```
chrisdavidmills:
```suggestion
The [CSS scroll-driven animations](/en-US/docs/Web/CSS/CSS_scroll-driven_animations) module defines features for setting the `animation-timeline` as a scroll-progress or view-progress timeline. You can explicitly [name an element as a timeline controller](#named_timelines) using the `scroll-timeline-*` and `view-timeline-*` properties, then set that name as the `animation-timeline` of a descendant element. You can also define _anonymous scroll progress timelines_ and _anonymous view progress timelines_ using the [`scroll()`](#scroll-progress_timelines) and [`view()`](#view-progress_timelines)) functions.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+Alternatively, the `animation-timeline` property can be used to explicitly state that the [default document timeline be used](#regular_css_animations_default_document_timeline) or to specify that the [animation doesn't have a timeline](#removing_an_animations_timeline), and therefore shouldn't occur at all.
+
+### Regular CSS animations: default document timeline
+
+By default, CSS animation progression is based on the time-based document timeline.
```
chrisdavidmills:
```suggestion
```

You've already said this so many times.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+  animation-timeline: auto;
+}
+```
+
+We create a rotation keyframe animation called `action`:
+
+```css live-sample___regular live-sample___named_scroll live-sample___anon_scroll
+@keyframes action {
+  from {
+    rotate: 45deg;
```
chrisdavidmills, line 61: Start this from `0deg`, so it looks smoother when applied? It currently looks jerky.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+> [!NOTE]
+> The `animation-timeline` is reset to the default `auto` value by the {{cssxref("animation")}} shorthand property, but cannot be set using the shorthand. Therefore, when creating scroll-driven animations, always declare the `animation-timeline` after any `animation` shorthand declarations to achieve the desired effect.
+
+### Scroll progress timelines
+
+With _scroll progress timeline_, the timeline progresses based on the scrolling of the scrollable element (_scroller_) from top to bottom (or left to right) and back again. By default, the position in the scroll range is converted into a percentage of progress — `0%` at the start and `100%` at the end. <!--This [animation range can be controlled](#controlling_the_animation_range) via the {{cssxref("animation-range")}} properties.-->
+
+To create a scroll progress timeline, the `animation-timeline` value must reference the scroller. The _scroller_, the element that provides the scroll progress timeline, can be named or anonymous.
```
chrisdavidmills:
```suggestion
To create a scroll progress timeline, the `animation-timeline` value must reference the scroller, which can be named or anonymous.
```

You already defined "scroller" above.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+#### Named scroll progress timelines
+
+A _named scroll progress timeline_ is one where the scroller (the {{glossary("scroll container")}} whose scrolling controls the progress of the animation timeline) is explicitly named using the {{cssxref("scroll-timeline-name")}} property (or the {{cssxref("scroll-timeline")}} shorthand). The name is a {{cssxref("dashed-ident")}}. The scroller is linked to the element to be animated by specifying its `scroll-timeline-name` as the value of that element's `animation-timeline` property.
```
chrisdavidmills:
```suggestion
A _named scroll progress timeline_ is one where the scroller is explicitly named using the {{cssxref("scroll-timeline-name")}} property (or the {{cssxref("scroll-timeline")}} shorthand). The name is a {{cssxref("dashed-ident")}}. The scroller is linked to the element to be animated by specifying its `scroll-timeline-name` as the value of that element's `animation-timeline` property.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+```html live-sample___named_scroll live-sample___anon_scroll
+<main class="scroller">
+  <div class="container">
+    <span class="item"></span>
+  </div>
+</main>
+```
+
+We provide some basic styles, which we've hidden for brevity. The important styles include setting a height on the container that is taller than the scroller, and then setting the overflow to allow scrolling:
```
chrisdavidmills: This reads weirdly, as you are introducing some styles but then not showing them. Maybe just show them all? They don't take up that much space.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+  left: calc(50% - (var(--size) / 2));
+  top: calc(50% - (var(--size) / 2));
+}
+```
+
+In this case, we don't have a checkbox, as the `action` animation progression is controlled by the scrolling of the overflowing scroller, which, unlike time, does not expire.
+
+{{EmbedLiveSample("named_scroll", "100%", "150")}}
+
+Before any scrolling occurs, the container's position is at the top of the scroller, and the animation is at the 0% keyframe. Try scrolling down. As you scroll, the animation progresses through the timeline, rotating an additional 720 degrees. When you can no longer scroll, the animation's progression is at the 100%, or `to`, keyframe. The animated item never returns to its default rotation.
```
chrisdavidmills:
```suggestion
Before any scrolling occurs, the container's position is at the top of the scroller, and the animation is at the 0% keyframe. Try scrolling down. As you scroll, the animation progresses through the timeline, rotating an additional 720 degrees. When you can no longer scroll, the animation's progression is at the 100%, or `to`, keyframe. The animated item doesn't return to its default rotation unless the scroller is scrolled back up to the top.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+}
+```
+
+{{EmbedLiveSample("anon_scroll", "100%", "150")}}
+
+We set an {{cssxref("inline-size")}} on the container so that it overflows in the inline direction, and set the {{cssxref("block-size")}} to `100%`, so it no longer overflows in the block direction. Try scrolling in the inline direction.
+
+### View progress timelines
+
+You can also progress an animation based on the change in visibility of an element inside a scroller — this is done via _view progress timelines_. Instead of tracking the scroll offset of a scroll container, view progress timelines track the relative position of an element, called the _subject_, within a scrollport. The progression of the animation is based on the _visibility_ of the subject inside the scroller. Unlike scroll progress timelines, with view progress timelines, you can't specify the scroller — the subject's visibility is always tracked within its nearest ancestor scroller.
```
chrisdavidmills:
```suggestion
You can also progress an animation based on the change in visibility of an element inside a scroller — this is done via _view progress timelines_. Instead of tracking the scroll offset of a scroll container, view progress timelines track the relative position of an element, called the _subject_, within a scrollport. The progression of an animation's keyframes is based on the _visibility_ of the subject inside the scroller. Unlike scroll progress timelines, with view progress timelines, you can't specify the scroller — the subject's visibility is always tracked within its nearest ancestor scroller.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+When a view progress timeline is applied to an animated element, the animation's keyframe progression reflects the subject's position in the scroller. The animation only occurs when the element is visible within its scrollport. Timeline progress starts at `0%` when the tracked subject starts intersecting the scrollport at the block or inline end edge. The `100%` occurs when the subject exits the scrollport at the block or inline start edge.
```
chrisdavidmills:
```suggestion
A view progress timeline animation only occurs when the element is visible within its scrollport. Timeline progress starts at `0%` when the tracked subject starts intersecting the scrollport at the block or inline end edge. The `100%` occurs when the subject exits the scrollport at the block or inline start edge.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+Because the `100%` is generally reached when the element leaves the viewport, you likely want to set the final effect of your animation in a keyframe block that occurs well before the end of the animation. You can set your completed effect within the 20%, 50%, or 80% keyframe block rather than using the `to` or `100%` keyframe to ensure the element finishes animating while still in view.
```
chrisdavidmills:
```suggestion
Because the `100%` is generally reached when the element leaves the viewport, you likely want to set the final effect of your animation in a keyframe block that occurs well before the end of the animation. You can set your completed effect within the `20%`, `50%`, or `80%` keyframe block rather than using the `to` or `100%` keyframe to ensure the element finishes animating while still in view.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+Similar to scroll progress timelines, the view progress timeline can be named or anonymous.
+
+#### Named view progress timeline
+
+A _named view progress timeline_ is one where the subject is explicitly named using the {{cssxref("view-timeline-name")}} property, a component of the `view-timeline` shorthand. The `<dashed-ident>` name is then linked to the element to animate by specifying it as the value of that element's `animation-timeline` property.
+
+With named view progress timelines, the element to animate does not have to be the same as the subject. In other words, the element controlling the timeline doesn't have to be the same as the element being animated. This means you can animate one element based on another element's movement in the viewport, and animations can depend on a separate element's visibility or position within its scrollable container.
```
chrisdavidmills:
```suggestion
With named view progress timelines, the element to animate does not have to be the same as the subject. In other words, the element controlling the timeline doesn't have to be the same as the element being animated. This means you can animate one element based on another element's movement within its scrollable container.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
chrisdavidmills: I don't really understand your point here. You could change _both_  view progress timeline examples to work in the line direction. I really do think it is hard to see what is going on in these examples.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,366 @@
...
+
+- Zero or one `<axis>` parameters. If set, this specifies the scroll axis along which the animation progresses.
+- Either the keyword `auto` or zero, one, or two {{cssxref("length-percentage")}} inset values. If set, these values specify offsets for the start and/or end of the scrollport.
+
+Declaring `view()` is equivalent to `view(block auto)`, which defines `block` as the axis of the parent element that supplies the timeline and the {{cssxref("scroll-padding")}}, which generally defaults to `0`, as the insets within the visible area at which the animation starts and ends.
+
+The function sets the values of the {{cssxref("view-timeline-axis")}} and {{cssxref("view-timeline-inset")}} properties.
+
+The {{cssxref("view-timeline-inset")}} arguments specify insets (if positive) or outsets (if negative) that adjust the start and end of the scrollport. They are used to determine whether the element is in view, which determines the length of the animation timeline. In other words, instead of starting at the start edge and ending at the end edge of the scrollport, the animation lasts as long as the element is in the inset-adjusted view.
```
chrisdavidmills:
```suggestion
The {{cssxref("view-timeline-inset")}} arguments specify insets (if positive) or outsets (if negative) that adjust the start and end of the scrollport. They are used to determine the scroll positions at which the element is considered "in view", which determines the length of the animation timeline. In other words, instead of starting at the start edge and ending at the end edge of the scrollport, the animation occurs at the start and end of the inset-adjusted view.
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
chrisdavidmills: OK. Saying that, I do think the menu is a bit sparse, and could do with more H2s for people to find the scroll progress and view progress info more easily.

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle: I'll make the animation taller

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle: intentional. The jerk helps the reader when that the animation is applied versus when it isn't

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,365 @@
...
+
+```html live-sample___named_scroll live-sample___anon_scroll
+<main class="scroller">
+  <div class="container">
+    <span class="item"></span>
+  </div>
+</main>
+```
+
+We provide some basic styles, which we've hidden for brevity. The important styles include setting a height on the container that is taller than the scroller, and then setting the overflow to allow scrolling:
```
estelle:
```suggestion
We provide some basic styles. The important ones include setting a height on the container that is taller than the scroller, and then setting the overflow to allow scrolling:
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,365 @@
...
+
+```css hidden live-sample___named_scroll live-sample___anon_scroll
```
estelle:
```suggestion
```css live-sample___named_scroll live-sample___anon_scroll
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,365 @@
...
+  to {
+    rotate: 720deg;
+  }
+}
+```
+
+```css hidden live-sample___named_view live-sample___anon_view live-sample___anon_view_args
+.scroller {
+  width: 400px;
+  height: 100px;
```
estelle:
```suggestion
  height: 200px;
  line-height: 2;
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
estelle: promoted the headings so there are more h2s

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,365 @@
...
+  <div class="item"></div>
+  <p>&nbsp;</p>
+  <p>&nbsp;</p>
+  <p>&nbsp;</p>
+  <p>&nbsp;</p>
+  <p>Scroll up to view the animation</p>
+</main>
+```
+
+{{EmbedLiveSample("named_view", "100%", "150")}}
```
estelle:
```suggestion
{{EmbedLiveSample("named_view", "100%", "250")}}
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,365 @@
...
+In this example, we again define the `animation` before the `animation-timeline`, so the timeline is not reset. We then include an argument-less `view()` function. We don't specify a scroller, as, by definition, the subject's visibility is tracked by its nearest ancestor scroller.
+
+```css live-sample___anon_view
+.item {
+  animation: action 1ms linear;
+  animation-timeline: view();
+}
+```
+
+{{EmbedLiveSample("anon_view", "100%", "150")}}
```
estelle:
```suggestion
{{EmbedLiveSample("anon_view", "100%", "250")}}
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
```diff
@@ -0,0 +1,365 @@
...
+  }
+}
+
+.item {
+  animation: action 1ms linear;
+  animation-timeline: view(block 20% 20%);
+}
+```
+
+{{EmbedLiveSample("anon_view_args", "100%", "150")}}
```
estelle:
```suggestion
{{EmbedLiveSample("anon_view_args", "100%", "250")}}
```

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
chrisdavidmills: Much better, thanks!

files/en-us/web/css/css_scroll-driven_animations/timelines/index.md
chrisdavidmills: Perfect.

chrisdavidmills (approved):
@estelle looking good to go! I decided to commit directly to the PR to fix a couple of remaining typos and minor grammar issues I noticed rather than throw it back to you a third time; hope you don't mind.

PR #41754, opened 2025-10-31
Lint: HTML conformance/style fixes

Josh-Cena:
- Lowercase DOCTYPE per style guide
- Buttons must have accessible name
- Scripts don't need `type="text/javascript"`
- Boolean attributes should be specified without value
- `id`s should be valid CSS identifiers for convenience
- Avoid deprecated attributes like `align` and `frameborder`
- Avoid invalid attributes, especially with `<source>` elements: they have different valid attributes depending on the parent, so the parent element should be explicit
- `` elements require `src` and `alt`; `<a>` elements require `href`
- Fix some invalid nesting involving `<label>` and `<button>` permitted contents and `<th>` permitted parents

### DIFF
--- a/files/en-us/web/api/htmlsourceelement/sizes/index.md
+++ b/files/en-us/web/api/htmlsourceelement/sizes/index.md
@@ -25,7 +25,7 @@ A string.
     srcset="medium-pic.jpg"
     type="image/jpeg"
     sizes="(50em <= width <= 60px) 50em,
-         (30em <= width < 50em) 30em" />
+           (30em <= width < 50em) 30em" />
 </picture>
 ```
 


files/en-us/web/api/node/previoussibling/index.md
```diff
@@ -50,18 +50,18 @@ document.getElementById("b2").previousSibling.id; // "b1"
 
 ### Second example
 
-In this example, there are whitespace text nodes (line breaks) between the `img` elements.
+In this example, there are whitespace text nodes (line breaks) between the `span` elements.
```
bsmth:
```suggestion
In this example, there are whitespace text nodes (line breaks) between the {{htmlelement("span")}} elements.
```

files/en-us/web/api/htmlsourceelement/sizes/index.md
```diff
@@ -19,12 +19,14 @@ A string.
...
-  src="mediumVideo.mov"
-  type="video/quicktime"
-  sizes="(50em <= width <= 60px) 50em,
+<picture>
+  <source
+    id="el"
+    srcset="medium-pic.jpg"
+    type="image/jpeg"
+    sizes="(50em <= width <= 60px) 50em,
          (30em <= width < 50em) 30em" />
```
Josh-Cena:
```suggestion
           (30em <= width < 50em) 30em" />
```

PR #41757, opened 2025-10-31
Add note for embedding SVG files

antareepsarkar:
### Description

Readers may think that the `svg` element is used for putting svg files. But, the `img` element is used for that. To avoid this confusion, a note is added. 

### Motivation

I got confused and tried put a svg file using the `svg` element.

### DIFF
--- a/files/en-us/web/svg/reference/element/svg/index.md
+++ b/files/en-us/web/svg/reference/element/svg/index.md
@@ -8,8 +8,7 @@ sidebar: svgref
 
 The **`<svg>`** [SVG](/en-US/docs/Web/SVG) element is a container that defines a new coordinate system and [viewport](/en-US/docs/Web/SVG/Reference/Attribute/viewBox). It is used as the outermost element of SVG documents, but it can also be used to embed an SVG fragment inside an SVG or HTML document.
 
-> [!NOTE]
-> The `svg` element is not required to put svg files in the document. Instead, the `img` element can be used for that. The `src` attribute of the `img` element will be equal to the path of the file.
+This element is for creating _new_ SVG documents. If you have an existing SVG document to embed in another document via URL, use {{HTMLElement("img")}}, {{HTMLElement("object")}}, or {{SVGElement("image")}}.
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/svg/reference/element/svg/index.md
```diff
@@ -8,6 +8,9 @@ sidebar: svgref
 
 The **`<svg>`** [SVG](/en-US/docs/Web/SVG) element is a container that defines a new coordinate system and [viewport](/en-US/docs/Web/SVG/Reference/Attribute/viewBox). It is used as the outermost element of SVG documents, but it can also be used to embed an SVG fragment inside an SVG or HTML document.
 
+> [!NOTE]
+> The `svg` element is not required to put svg files in the document. Instead, the `img` element can be used for that. The `src` attribute of the `img` element will be equal to the path of the file.
```
Josh-Cena: MDN only says what something does, not what something doesn't do, because there are infinitely many things something can't do.

```suggestion
This element is for creating _new_ SVG documents. If you have an existing SVG document to embed in another document via URL, use {{HTMLElement("img")}}, {{HTMLElement("object")}}, or {{SVGElement("image")}}.
```

antareepsarkar: Thanks for the conversation. Actually, I am quite new to open source. I thought that some may confuse that they need the `svg` element to embed an existing svg document.

Josh-Cena: I believe what I proposed here conveys the same meaning as yours, just written in an affirmative manner rather than negative

antareepsarkar: Please tell me if I need to commit the suggestion to finalize it. If yes, then give a commit message.

Josh-Cena:
Thanks. You don't need to worry about commit messages because PRs are squash merged. You can click on "commit suggestion" if you agree with it.

Josh-Cena (approved):

PR #41769, opened 2025-11-01
Fix syntax example for device-cmyk function

tylerjmorg:
### Description

Add an example to include real numbers and `none` in syntax.

### Motivation

Developers should have alternative examples of the syntax if they choose not to use percentages.

### Additional details

https://drafts.csswg.org/css-color-5/#device-cmyk


Josh-Cena (approved):
Thanks

PR #41771, opened 2025-11-01
Lint: HTML conformance/style fixes (2)

Josh-Cena:
Follow up of https://github.com/mdn/content/pull/41754.

- Tag `vue` code blocks properly
- Remove illicit `<iframe>` children; I know this is for fallback but there's no reason to care about lack of `iframe` support
- Buttons and links should have accessible names
- Images should have `src`
- Fix invalid nesting, such as `<menu>` having non-`li` children
- `<map>` need a name, and if it has an id they need to be equal. Also slightly updated the docs for `HTMLImageElement.useMap` to reflect this
- `id` should be valid CSS identifiers
- Fix a case where we do `<legend role="button">` to use an actual button instead
- Properly escape `<` and `>`; properly add trailing slashes for void elements
- Boolean attributes have no value
- Remove deprecated attributes
- `street-address` is not a valid `autocomplete` token for `<input>`; replaced with `<textarea>`
- Labels cannot label multiple controls: there are a few cases where we use the pattern `<label for="a"><output></output></label><input id="a">` where we don't actually intend for the `output` to be a form control, so replaced it with `<span>` instead

### DIFF
--- a/files/en-us/web/api/htmlimageelement/usemap/index.md
+++ b/files/en-us/web/api/htmlimageelement/usemap/index.md
@@ -67,6 +67,8 @@ You can learn more about client-side image maps in our learning article [Add a h
 
 ## Examples
 
+### Using useMap
+
 Consider a `<map>` that looks like this:
 
 ```html


files/en-us/learn_web_development/core/structuring_content/including_vector_graphics_in_html/index.md
```diff
@@ -171,16 +171,14 @@ You can open SVG images in your browser just like webpages. So embedding an SVG
...
-</iframe>
+<iframe src="triangle.svg" width="500" height="500" sandbox></iframe>
 ```
 
 This is definitely not the best method to choose:
 
 #### Cons
 
-- `iframe`s do have a fallback mechanism, as you can see, but browsers only display the fallback if they lack support for `iframe`s altogether.
+- `iframe`s do have a fallback mechanism by including content between the opening and closing tags, but browsers only display the fallback if they lack support for `iframe`s altogether.
```
chrisdavidmills: Strange phrasing, how about this:

```suggestion
- `<iframe>` elements can include fallback content between their opening and closing tags, but this is only displayed in browsers that don't support `<iframe>`s, not when the image fails to load.
```

files/en-us/web/api/htmlformcontrolscollection/nameditem/index.md
```diff
@@ -32,7 +32,12 @@ namedItem(name)
 
 ### Return value
 
-A {{domxref("RadioNodeList")}}, {{domxref("Element")}}, or [`null`](/en-US/docs/Web/JavaScript/Reference/Operators/null).
+- A {{domxref("RadioNodeList")}} if there are multiple elements with the given `name` or `id`
+- An {{domxref("Element")}} if there is exactly one element with the given `name` or `id`
+- [`null`](/en-US/docs/Web/JavaScript/Reference/Operators/null) if no element has the given `name` or `id`
+
+> [!NOTE]
+> The returned {{domxref("RadioNodeList")}} is live, meaning that its contents automatically update if elements that match the given name are added to or removed from collection. Furthermore, it may contain non-radio input elements, despite the name of the collection.
```
chrisdavidmills:
```suggestion
> The returned {{domxref("RadioNodeList")}} is live, meaning that its contents automatically update if elements that match the given name are added to or removed from the collection. Furthermore, it may contain non-radio input elements despite the collection's name.
```

files/en-us/web/api/htmlformcontrolscollection/nameditem/index.md
```diff
@@ -32,7 +32,12 @@ namedItem(name)
 
 ### Return value
 
-A {{domxref("RadioNodeList")}}, {{domxref("Element")}}, or [`null`](/en-US/docs/Web/JavaScript/Reference/Operators/null).
+- A {{domxref("RadioNodeList")}} if there are multiple elements with the given `name` or `id`
```
chrisdavidmills: Add full stops to the ends of these?

files/en-us/web/api/htmlimageelement/usemap/index.md
```diff
@@ -98,6 +58,38 @@ img {
...
 
+## Value
+
+A string containing the hash `#` symbol followed by the [`name`](/en-US/docs/Web/HTML/Reference/Elements/map#name) of the {{HTMLElement("map")}} element
+which defines the image map to apply to the image.
+
+You can learn more about client-side image maps in our learning article [Add a hitmap on top of an image](/en-US/docs/Web/HTML/How_to/Add_a_hit_map_on_top_of_an_image).
+
+## Examples
+
```
chrisdavidmills, line 69: Include an H3 that describes the example under here, along the lines of "### Basic `usemap` usage"?

files/en-us/web/http/guides/permissions_policy/index.md
```diff
@@ -183,9 +183,8 @@ Several features can be controlled at the same time by including a semi-colon-se
 It is worth giving the `src` value a special mention. We mentioned above that using this allowlist value will mean that the associated feature will be allowed in this `<iframe>`, as long as the document loaded into it comes from the same origin as the URL in its {{HTMLElement('iframe','src','#Attributes')}} attribute. This value is the _default_ `allowlist` value for features listed in `allow`, so the following are equivalent:
 
 ```html
-<iframe src="https://example.com" allow="geolocation 'src'">
-  <iframe src="https://example.com" allow="geolocation"></iframe
-></iframe>
+<iframe src="https://example.com" allow="geolocation 'src'"></iframe>
+<iframe src="https://example.com" allow="geolocation"></iframe>
```
chrisdavidmills, line 187: I agree that the nesting is a bit weird, but I'm not sure if this example is now equivalent to the previous one, as the preceding text says. Would it be better to include just the first iframe in the block?

Also maybe update the preceding text to "...so the following is equivalent:"?

files/en-us/web/http/reference/headers/permissions-policy/index.md
```diff
@@ -267,9 +267,8 @@ Several features can be controlled at the same time by including a semi-colon-se
 It is worth giving the `src` value a special mention. We mentioned above that using this allowlist value will mean that the associated feature will be allowed in this `<iframe>`, as long as the document loaded into it comes from the same origin as the URL in its {{HTMLElement('iframe','src','#Attributes')}} attribute. This value is the _default_ `allowlist` value for features listed in `allow`, so the following are equivalent:
 
 ```html
-<iframe src="https://example.com" allow="geolocation 'src'">
-  <iframe src="https://example.com" allow="geolocation"></iframe
-></iframe>
+<iframe src="https://example.com" allow="geolocation 'src'"></iframe>
```
chrisdavidmills, line 270: Same comment as for the Permissions Policy guide.

dipikabh:
Just a heads up - we'll most likely move the at-rule pages under `css / reference` tomorrow,  so this PR might be impacted if it's not merged before then.

Josh-Cena:
Thank for the heads up. I'm busy this morning but I'll get my recent PRs merged this afternoon.

files/en-us/web/api/htmlformcontrolscollection/nameditem/index.md
```diff
@@ -32,7 +32,12 @@ namedItem(name)
...
+- An {{domxref("Element")}} if there is exactly one element with the given `name` or `id`
+- [`null`](/en-US/docs/Web/JavaScript/Reference/Operators/null) if no element has the given `name` or `id`
```
Josh-Cena:
```suggestion
- A {{domxref("RadioNodeList")}} if there are multiple elements with the given `name` or `id`,
- An {{domxref("Element")}} if there is exactly one element with the given `name` or `id`, or
- [`null`](/en-US/docs/Web/JavaScript/Reference/Operators/null) if no element has the given `name` or `id`.
```

files/en-us/web/api/htmlimageelement/usemap/index.md
Josh-Cena, line 69:
```suggestion

### Using useMap

```

files/en-us/web/api/htmlimageelement/usemap/index.md
```diff
@@ -98,6 +58,38 @@ img {
...
+  <area shape="rect" coords="25, 25, 100, 150" href="/index.html" alt="Shop" />
+</map>
+```
+
+Given the image map named `mainmenu-map`, the image which uses it should
+look something like the following:
+
+```html
+<img src="menubox.png" usemap="#mainmenu-map" />
+```
```
Josh-Cena:
````suggestion
Given the image map named `mainmenu-map`, you can dynamically construct images that reference the image map as follows:

```js
const image = new Image();
image.src = "menubox.png";
image.alt = "";
image.useMap = "#mainmenu-map";
```
````

files/en-us/web/http/guides/permissions_policy/index.md
Josh-Cena: This is actually the originally intended meaning. Before #24058, the example was written without closing tags and Prettier gets confused. The original text also had two "elements" juxtaposed, and I believe the intention is clear enough: we are comparing two elements within one code block.

files/en-us/web/http/reference/headers/permissions-policy/index.md
Josh-Cena: Same as above ^

Josh-Cena:
Thanks for the reviews @chrisdavidmills! A lot of these HTML errors are caused by the Prettier formatting, so I've just restored them to the intended, error-free meaning.

PR #41772, opened 2025-11-01
DOM scripting introduction: Remove redundant instruction in "Creating a dynamic shopping list" exercise

zunaid-hassan:
Clarified instructions for building a dynamic shopping list by removing redundant phrasing about item deletion.

### Description

Removes a duplicated requirement (item 2 and 3 were identical) to prevent confusion in the exercise.

**Changes:**
> - The item should appear in the list.
> - Each item should be given a button that can be pressed to delete that item off the list.
> - Each item should have a button next to it that removes the item from the list when clicked.
> - The input fields should be cleared and focused, ready for the next item entry.

**To:**
> - The item should appear in the list.
> - Each item should have a button next to it that removes the item from the list when clicked.
> - The input fields should be cleared and focused, ready for the next item entry.

### Motivation

Removes unnecessary repetition and makes it easier for learners to progress. 

### Additional details

Article: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/DOM_scripting#creating_a_dynamic_shopping_list

N/A


Josh-Cena (approved):
Good catch, thanks

PR #41779, opened 2025-11-02
Add a federated identity guide

wbamberg:
This PR adds a guide to federated identity.

- an overview of what it is
- a moderately close look at the main flow in OpenID Connect
- a short but sufficient (I think) discussion of the use of third-party cookies in federated identity systems
- an overview of FedCM
- a short section on considerations around choosing IdPs
- summary of strengths and weaknesses

**Note that preview images are broken because of https://github.com/mdn/fred/issues/904.**

### DIFF
--- a/files/en-us/web/security/authentication/federated_identity/index.md
+++ b/files/en-us/web/security/authentication/federated_identity/index.md
@@ -213,7 +213,7 @@ In particular, it finds that:
 
 - The next most secure pattern is one in which the website uses a web server to handle all OAuth/OIDC interactions, but then returns the access token to the front end, and the front end then makes API requests directly. In this scenario the website can be a confidential client but malicious code running in the browser (for example, through an XSS attack) can potentially steal access tokens. However, the front end doesn't have to store access tokens long-term: it can retrieve them from the backend when it needs them.
 
-- The least secure pattern is one in which OAuth/OIDC interactions and interactions with APIs both take place in the front end. This, for example, would be the natural architecture for a {{glossary("SPA", "Single-page app")}}, where the entire application executes in the browser. In this architecture the RP can't be a confidential client, because it can't reliably keep a client secret. This means that it can't authenticate itself to the IdP. It also has to persistently store tokens, which increases the risk of malicious code stealing them.
+- The least secure pattern is one in which OAuth/OIDC interactions and interactions with APIs both take place in the front end. This, for example, would be the natural architecture for an application that executes entirely in the browser. In this architecture the RP can't be a confidential client, because it can't reliably keep a client secret. This means that it can't authenticate itself to the IdP. It also has to persistently store tokens, which increases the risk of malicious code stealing them.
 
 The specification also includes detailed recommendations for security practices to follow in each of these three scenarios.
 


hamishwillee:
>  I think it is most helpful to name specific providers, such as Google and GitHub and (in a slightly different way I suppose) Auth0 but I don't like the idea of seeming to pick winners or entrench monopolies.

Is there any benefit to using a particular specific provider or are they all "much of a muchness". If there is a real benefit then it is absolutely necessary to mention them.
You imply that you have to choose just one provider - I thought you could offer several (perhaps I am imagining things).
Anyway, I imagine that if nothing else, the fact that more users are likely to already be authenticated with an entrenched service would seem compelling.

So depending on the answer above you might say something like.

> There are numerous identity providers, and all of them provide similar services. There is some benefit to using established providers, such as Google, Github, and Facebook, as it increase the likelihood that your users will already have an account they can authenticate with.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+The standard most commonly used for federated identity on the web is [OpenID Connect (OIDC)](https://openid.net/developers/how-connect-works/), which is an authentication protocol built on top of the [OAuth 2.0 authorization framework](https://datatracker.ietf.org/doc/html/rfc6749).
+
+### Authentication flow
+
+In this section we'll start by walking through the main authentication flow defined in OIDC. There are many options within the OIDC authentication flow: in this walkthrough we'll present the recommended options, and we will talk later about alternatives.
+
+This flow is defined in the [OpenID Connect Core](https://openid.net/specs/openid-connect-core-1_0.html) specification.
+
+As a prerequisite, the RP needs to be known to the IdP: the IdP needs to have an ID for the RP, which is called a client ID, and the RP must be able to authenticate itself to the IdP, for example using a shared secret called a client secret or using some other mechanism such as TLS client authentication.
```
hamishwillee: Nit. "known to the IdP: ". That colon feels like it would be better as an emdash or a full stop. Actually split this whole paragraph as much as you can, also on "For example".

hamishwillee: If you're reading these docs fresh - i.e. you're not in the fedcm space - it is easy to lose track of what IdP and RP are. I know that the point of providing the acronyms is to make it easier to refer to them, but it might be worth occasionally spelling it out again. 

So here you might, for example change "the IdP needs to have an ID for the RP," to "the IdP needs to have an ID for your site (the Relying Party),".

I'd understand if you objected but read the sentence again - it is acronym soup.

hamishwillee: It might be worth noting that Client ID is `client_id` in the flow even this early.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,213 @@
...
+- The RP generates a value that is hard to guess and is specific to this authentication request. This value is called the _code verifier_.
+- The RP creates a {{glossary("hash", "cryptographic hash")}} of the code verifier, and uses it as the `code_challenge` parameter in the authentication request.
+- The IdP stores the code challenge, and associates it with the access code that it returns to the RP.
+
+In the token request:
+
+- The RP passes the code verifier in the _code_verifier_ parameter.
+- The IdP hashes the code verifier, and compares the result with the stored code challenge: if they do not match, then the token request is denied.
+
+This defends against two attacks: CSRF against the RP's redirect URL, and authorization code injection.
```
hamishwillee: Link to attack page for CSRF

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+The CSRF attack works as follows:
```
hamishwillee:
```suggestion
A CSRF attack works as follows:
```

hamishwillee: Consider also having the attacks as sub headings. There's a lot going on here and breaking it up might make it more digestible.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+3. The IdP authenticates the user. The protocol does not specify a particular method for this: the IdP could use a password, a one-time password, a biometric, or any other appropriate method.
+
+4. If authentication is successful, the IdP generates the authorization code. It also stores the `code_challenge` value, and associates it with the authorization code. The IdP then redirects the browser to the RP's redirect URL, passing the authorization code as a parameter.
+
+In the token request:
+
+1. The RP makes a {{httpmethod("POST")}} request to the token endpoint. This request includes the following parameters:
+   - `client_id`: Identifies this RP to the IdP.
+   - `client_secret`: The secret used to authenticate the RP to the token endpoint. The RP could use some alternative mechanism for client authentication, such as TLS client authentication.
```
hamishwillee: Presumably this is just a number right? Presumably this is all done via HTTPs so it can never leak?

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+2. The attacker tricks the user's browser into making an HTTP request to the RP's redirect URL, including the attacker's authorization code. To the RP, this looks like a response from the IdP to an authentication request originating from the user.
+
+3. The RP makes a token request to the IdP, including the attacker's authorization code.
+
+4. The IdP responds with the attacker's tokens.
+
+5. The RP signs the user into the attacker's account: now any information or instructions they provide are under the attacker's control.
+
+Essentially, the attack succeeds because the RP doesn't know that the authentication response is not a response to a request made on behalf of the user. PKCE prevents this attack because the IdP would have a stored code challenge for the attacker's access code, and this would not match the code verifier that the RP passes into the token request.
```
hamishwillee: Consider rewording, because the attack doesn't succeed.

hamishwillee: Let me put it another way. After reading this I am not sure how PKCE prevented the attack - I'm just too tired. I might be able to understand if the place at which the request was prevented was in the preceding bullet points. Even if just a note or highlight at the point where success is thwarted.

Same problem with the next case.

I might be wrong. I'll try look at it with fresh eyes next time, not at the end of the day.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,213 @@
...
+- The RP passes the code verifier in the _code_verifier_ parameter.
+- The IdP hashes the code verifier, and compares the result with the stored code challenge: if they do not match, then the token request is denied.
+
+This defends against two attacks: CSRF against the RP's redirect URL, and authorization code injection.
+
+The CSRF attack works as follows:
+
+1. The attacker makes an authentication request to the IdP for themselves, and gets back an authorization code for their own tokens.
+
+2. The attacker tricks the user's browser into making an HTTP request to the RP's redirect URL, including the attacker's authorization code. To the RP, this looks like a response from the IdP to an authentication request originating from the user.
```
hamishwillee: Might be worth numbering the point of injection on the preceding flow so it is easier to see what is happening (just thinking aloud)

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+2. The attacker starts its own OIDC authentication flow, but intercepts the IdP's authentication response, replacing the authorization code with the code it stole from the user. This is straightforward, because the authentication response is for the attacker, so it passes through the attacker's device.
+
+3. The RP then continues its authentication flow for the attacker by making the token request to the IdP, including the user's authorization code that the attacker has injected.
+
+4. The IdP responds with the user's tokens.
+
+5. The RP signs the attacker into the user's account.
+
+Note that this is really the inverse of the CSRF attack. Note also that the `state` parameter doesn't help here, because the authentication request and response really do belong to the same flow - the attacker's.
```
hamishwillee: "Note that this is really the inverse of the CSRF attack." I don't understand in what way.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+PKCE protects against this attack, because the IdP's token endpoint will check the `code_verifier` in the token request against the code challenge that it received in the authentication request.
+
+- Because the IdP's stored code challenge is associated with the authorization code, it will be the challenge for the user.
+- Because the `code_verifier` in the token request is part of the attacker's flow, it will be the verifier for the attacker.
```
hamishwillee: What is meant by ", it will be the challenge for the user." and " it will be the verifier for the attacker.". I can tell this is "the point", but I don't get it.

I _think_ you're saying that
- when the authorization request was made for the stolen auth code it included a code_challenge, (which is a hash of the RP code verifier) is stored by the idp and associated with the code.
- the auth code gets passed into the flow just before the token request. My understanding is that the RP would include the code_verifier which is a hash of its original challenge
- But what happens now is a mystery, because the RP only knows its original request, so presumably that is the correct  code_verifier for the original challenge.

I suspect that  "t will be the challenge for the user" means that somehow the RP associated the code_verifier that it sends with the token with the user, so when the attacker injects itself into the flow it will have some (or no) code_verifier associated with it.

Whatever the case, I'm not sure the relationships are clear enough.

wbamberg: Perhaps it's clearer if we start earlier, with the user's flow. And lets look at how it works without PKCE, first.

1. the legitimate user tries to sign in
2. the RP makes an authentication request to the IdP
3. the IdP authenticates the user and sends the authorization code
4. BUT the user has installed a malicious extension, that steals the code, sends it to the attacker, and terminates the user's flow. The user just thinks "something went wrong".
5. now the attacker has the authorization code. But they can't just make a token request directly, because of client authentication (the attacker can't impersonate the RP).
6. so instead, the attacker starts their own authentication flow with the RP.
7. when the IdP sends them the authorization code, they intercept the redirect and replace their code with the the user's code that they stole. Remember, the user's code is for the user's tokens.
8. next, the RP processes the next step of the attacker's flow, which is the token request. But they're sending the user's code, so it returns the user's tokens to the attacker's flow. Result: the attacker is signed into the user's account

Basically the attack is just the attacker signing in, but splicing in the user's auth code.

(It's the inverse of CSRF, because with CSRF it's the user signing in, but splicing in the attacker's auth code.)

hamishwillee: Ah, so in the token request the code_verifier cannot be hashed to the code challenge, because the code challenge was produced in the users original authoriseation flow thingy?

Thanks you. Makes sense.

Now you're going to have to think "did Hamish need me to go through that because it was unclear or because he is an idiot". I think there is room for improvement. But if you want me to suggest something might need to wait until Friday.

wbamberg: With PKCE:

1. the legitimate user tries to sign in
2. the RP makes an authentication request to the IdP. **-> the RP generates a code verifier, and sends the hashed verifier (i.e. the challenge) to the IdP.**
3. the IdP authenticates the user and sends the authorization code **->the IdP stores the code challenge with the user's code**
4. BUT the user has installed a malicious extension, that steals the code, sends it to the attacker, and terminates the user's flow. The user just thinks "something went wrong".
5. now the attacker has the authorization code.
6. the attacker starts their own authentication flow with the RP. **->the RP generates a new code verifier, and the IdP stores this new code challenge next to the attacker's code**
7. When the IdP sends them the authorization code, they intercept the redirect and replace their code with the the user's code that they stole. Remember, the user's code is for the user's tokens.
8. next, the RP processes the next step of the attacker's flow, which is the token request. **->the RP sends the attacker's code verifier and the user's code. The IdP looks up the user's code, finds the user's verifier, and it doesn't match the code verifier it was passed. So it returns an error.** 

Why can't the attacker steal the user's code verifier too? Because it's never exposed to the front end, it only ever lives in the RP, server-side.

wbamberg: > "did Hamish need me to go through that because it was unclear or because he is an idiot"

It's hard. It took me a while. One challenge is that using more words and taking it step by step helps, but I don't want to spend too many words on this. Another is that once you understand it, it makes sense, and it's hard to know if _your explanation_ makes sense to someone else who doesn't understand it yet! I do think I have probably tried to compress it too much here.

wbamberg: However, I'm pretty sure that omitting the user's initial flow is a mistake, and perhaps including that is enough for this to make sense?

wbamberg: OK I have tried in https://github.com/mdn/content/pull/41779/commits/412b8efd124a78fc10cec87c9a9ea19cdf33822f to clarify the auth code injection attack and defense.

BTW there is a great presentation on this attack here: https://www.youtube.com/watch?v=moQidjdV5cw , probably I should include this too. But I'd appreciate if you can see if the words make sense before watching this :).

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: I had not done this because are already at H4(!) but you're right, done in https://github.com/mdn/content/pull/41779/commits/412b8efd124a78fc10cec87c9a9ea19cdf33822f.

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: -> https://github.com/mdn/content/pull/41779/commits/412b8efd124a78fc10cec87c9a9ea19cdf33822f

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: I've sort of had a go at this in https://github.com/mdn/content/pull/41779/commits/e84a7dc84f54c73a4dda9029dfff2987d9f2e3d0. I've broken it into bullets, which hopefully makes it more readable, and changed ID->identifier which removes at least one acronym.

I have not touched RP/IdP, because I'm not that sure it is helpful: you do need to get RP and IdP in your head for this, no option, and I feel like once you have, doing things like "your site (the Relying Party)" sometimes might be more confusing (is this the same as the thing we call RP elsewhere? Why are we using something different here?).

I also didn't have `client_id` yet because I think we re not ready.

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: Yeah, just an opaque value. And yes, everything over HTTPS. I didn't say that because it feels like the default. Worth spelling out do you think?

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: Removed, I don't think this helps (although I do think it is interesting).

hamishwillee: This caught my eye (i.e. this is just flyby response). I thought this was interesting too. You could add it back, but do it more fully as you did in your extra info:

> (It's the inverse of CSRF, because with CSRF it's the user signing in, but splicing in the attacker's auth code.)

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: I've updated the CSRF section in https://github.com/mdn/content/pull/41779/commits/8570f9b0efd104310e06e5d564c519754aa28354 - there are still some points I'm not very clear on (specifically, how does the RP associate code verifier values with requests, so it can say "this is the code verifier value for _that_ request"?) so we might need some more updates here.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -116,51 +121,77 @@ In the authentication request:
 
 - The RP generates a value that is hard to guess and is specific to this authentication request. This value is called the _code verifier_.
 - The RP creates a {{glossary("hash", "cryptographic hash")}} of the code verifier, and uses it as the `code_challenge` parameter in the authentication request.
-- The IdP stores the code challenge, and associates it with the access code that it returns to the RP.
+- The IdP stores the code challenge, and associates it with the authorization code that it returns to the RP.
```
hamishwillee, line 126: Mabye "user's authorization code" or similar. Just so that later we already have in our minds that this is associated with the user

files/en-us/web/security/authentication/federated_identity/index.md
hamishwillee: It does make it more readable. I accept your argument on RP, IdP. It's taken me lots of work to get my head around the terms. Really, the problem is that Relying Party is not a really intuitive name, even though it is the standard.

files/en-us/web/security/authentication/federated_identity/index.md
hamishwillee: Worth implying it can be any value maybe, but not that it needs to be HTTPs.

So maybe "The secret used to authenticate the RP to the token endpoint; this can be any value agreed between the RP and IdP."

In "The RP could use some alternative mechanism for client authentication, such as TLS client authentication." The "RP could use" reads as though the RP is doing the authentication and deciding what the authentication is. If that is correct then cool, but I assumed that this was to identify the RP to the IdP - so it is the IdP doing the authentication of the client.

Does this make sense as a question/ambiguity?

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -116,51 +121,77 @@ In the authentication request:
...
 
-5. The RP signs the user into the attacker's account: now any information or instructions they provide are under the attacker's control.
+Essentially, the attack succeeds because the RP doesn't know that the request to the redirect URL is not a response to a request made on behalf of the user.
 
-Essentially, the attack succeeds because the RP doesn't know that the authentication response is not a response to a request made on behalf of the user. PKCE prevents this attack because the IdP would have a stored code challenge for the attacker's access code, and this would not match the code verifier that the RP passes into the token request.
+When PKCE is used:
+
+- In step 1, the RP generates a code verifier for the attacker's request, and sends the hashed code verifier (the code challenge) to the IdP.
+- In step 2, the IdP stores the code challenge alongside the attacker's authorization code.
+- In step 5, the RP won't be able to find a code verifier for the user that matches the challenge the IdP stored, so the token request will fail.
```
hamishwillee, line 162: If possible, it might be nice to capture a flavour of your comment below here:

> Why can't the attacker steal the user's code verifier too? Because it's never exposed to the front end, it only ever lives in the RP, server-side.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -116,51 +121,77 @@ In the authentication request:
...
+5. The RP makes a token request to the IdP, including the attacker's authorization code, which it took from the redirect URL.
 
-3. The RP makes a token request to the IdP, including the attacker's authorization code.
+6. The IdP responds with the attacker's tokens.
 
-4. The IdP responds with the attacker's tokens.
+7. The RP signs the user into the attacker's account: now any information or instructions they provide are under the attacker's control.
 
-5. The RP signs the user into the attacker's account: now any information or instructions they provide are under the attacker's control.
+Essentially, the attack succeeds because the RP doesn't know that the request to the redirect URL is not a response to a request made on behalf of the user.
```
hamishwillee, line 155: The whole section is clear/much better. 

FWIW this line still grates because it says the attack succeeds, and isn't it that PKCE is part of the standard and you will never actually see this - i.e. its an explanatory device?

If so, maybe "this attack would succeed"? I'm being pedantic.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -116,51 +121,77 @@ In the authentication request:
...
+- In step 1, the RP generates a code verifier for the attacker's request, and sends the hashed code verifier (the code challenge) to the IdP.
+- In step 2, the IdP stores the code challenge alongside the attacker's authorization code.
+- In step 5, the RP won't be able to find a code verifier for the user that matches the challenge the IdP stored, so the token request will fail.
 
 An alternative protection against this attack is the `state` parameter defined in OAuth 2.0. In this defense, the RP provides an unpredictable value as a parameter in the authentication request, and the IdP includes the same value in the response: the RP checks that they match. Because the attacker can't predict the value of `state`, they can't pass a matching value to the RP's redirect URL.
 
-In the authorization code injection attack:
+##### Authorization code injection
+
+In an authorization code injection attack, the attacker steals an authorization code from the target user, and is able to inject it into the attacker's own sign-in flow. The result is that the attacker is signed into the user's account.
```
hamishwillee, line 167: As per https://github.com/mdn/content/pull/41779/files/ea59a88d0088a5926070966854f7eae344ad2e5b..8570f9b0efd104310e06e5d564c519754aa28354#r2501736705 - maybe prefer the phrasing that doesn't imply this can succeed - ", and ~is able~tries to".

files/en-us/web/security/authentication/federated_identity/index.md
hamishwillee: I've made some whiny comments, that don't take away from the fact that this is IMO crystal clear now.

wbamberg:
Added a section on FedCM -> https://github.com/mdn/content/pull/41779/commits/afe404f00d5daded08fc54e3b1cbd0e13542242c

Added a bit on PKCE downgrade -> https://github.com/mdn/content/pull/41779/commits/ac6e5ec467823db3659815b8d1061e2616d63426

wbamberg:
Recommend public-key crypto for client authentication -> https://github.com/mdn/content/pull/41779/commits/ad5a597cc04f55efcba3000f02ca343b002962bb

wbamberg:
Choosing IdPs, and a summary listing strengths and weaknesses -> https://github.com/mdn/content/pull/41779/commits/443cec780645f2ef96103525464a3e377531dce5

chrisdavidmills:
I've removed my review flag and requested a review from Hamish, as it looks like he is handling this nicely. Let me know if you do, in fact, need input from me.

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: I have tried using different/more words in https://github.com/mdn/content/pull/41779/commits/f97f7693eac726d5768a8633eda81887ab17d4fc, I'm not sure if it is clearer or not.

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: I'm not sure exactly what change is needed here.

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: Not sure. The flow here (steps 1-7 as described) are what would happen without PKCE. If you are assuming PKCE, then you don't get as far as 6, right? It stops at 5. So without PKCE, "the attack succeeds". I think if you do it all conditionally, it gets complicated, because you have to make 6 and 7 conditional too.

Hm. I pushed https://github.com/mdn/content/pull/41779/commits/7f542f3caf8f19c483e197b1750be21bbe8bdce6 to clarify that these attacks are without PKCE.

> isn't it that PKCE is part of the standard and you will never actually see this

Actually no. PKCE is not part of the standard yet although it will I think be in [OAuth 2.1](https://oauth.net/2.1/). But even then there will be clients around that are not using it, which is partly why (a) it's so complicated and (b) it's so important to have best practice guidelines.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+PKCE protects against this attack, because:
+
+- In step 2, the RP generates a code verifier and sent the hashed code challenge to the IdP, which stores the challenge alongside the user's code.
+- In step 6, the RP's token request contains the _attacker's_ code verifier but the _user's_ code. The IdP looks up the code challenge for the user's code: it won't match the attacker's code verifier, and the token request will be denied.
+
+An alternative to PKCE, specified in OIDC, is the [`nonce`](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#name-nonce) value. The RP includes this as another parameter in the authentication request: the IdP stores it, and the token endpoint returns it to the RP along with the tokens. The RP then checks that the returned value is the same as the original value.
+
+##### Ensuring that PKCE is used
+
+Although PKCE provides effective protection against the attacks described here, it's an optional part of the protocol. This means that an RP must ensure that its chosen IdP not only supports PKCE, but that it mandates the use of PKCE, refusing the token request if a valid code verifier is not included.
```
hamishwillee: I very much like this section.

1. Are there any major IdP that unexpectedly do/do not mandate the use of PKCE? It might be good to hint if there are particular concerns or "no concerns for the major players'.
2. Very nittish, but maybe even more direct - the fact that it is optional is interesting, but the most important bit here is that they need to check that it is mandatory - worth a highlight.

```suggestion
To ensure that PKCE is used, the RP must confirm that its chosen IdP not only supports PKCE, but that it also _mandates_ the use of PKCE — refusing the token request if a valid code verifier is not included.
```

wbamberg: > Are there any major IdP that unexpectedly do/do not mandate the use of PKCE? It might be good to hint if there are particular concerns or "no concerns for the major players'.

I'm not familiar enough with OIDC providers to know this unfortunately - and I expect that this is the kind of information that would go out of date quickly.

One very interesting and relevant difference between PKCE and state/nonce is that with PKCE it's the IdP that does the checking, so as an RP you must trust them to do it, and do it properly. Whereas with state/nonce it's the RP that does the checking, so you are more in control of your own destiny.

In some ways that makes PKCE better, because as a rule IdPs are probably going to be more security-minded than random websites. On the other hand it opens this up downgrade attack.

I considered adding this, but I've already spent a _lot_ of words on PKCE.

hamishwillee: That's cool. Leaving this unresolved in case you want to consider my suggestion in 2.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+### FedCM and federated identity protocols
+
+FedCM does not itself implement a federated identity protocol such as OIDC. You can think of it as providing transport between the RP, the user, and the IdP, but it is agnostic about the items that are exchanged or their interpretation.
+
+For example, in an implementation of OIDC using FedCM, the token returned by `CredentialsContainer.get()` may be an authorization code, and the RP will then have to retrieve the identity token from the IdP's token endpoint. That is, FedCM takes care of only the first part of the [authentication flow](#authentication_flow). The [FedCM for OAuth](https://github.com/aaronpk/oauth-fedcm-profile) document describes how OAuth and OIDC could be implemented using FedCM.
+
+In general, when an RP decides to use a particular IdP for federated login, the RP will register with the IdP, and as part of this process the IdP should explain to the RP exactly which arguments it expected to be given, how it should handle the objects that the IdP returns, and any other behavior it expects the RP to implement.
+
+## Choosing IdPs
```
hamishwillee, line 278: I accidentally added some comments to one of your linked commits - so linking in case that doesn't notify properly https://github.com/mdn/content/commit/443cec780645f2ef96103525464a3e377531dce5#r171001071

The main useful point was to consider adding info about the costs of having more idps.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+It's also common practice to provide a fallback option for users who can't use any of your chosen IdPs. In this option, users authenticate directly with your site using a method such as a [password](/en-US/docs/Web/Security/Authentication/Passwords) or an [OTP](/en-US/docs/Web/Security/Authentication/OTP).
+
+Whichever identity providers you choose will provide detailed instructions and tools for integrating your site with their system, and this is likely to take care of many of the complexities inherent in a protocol like OIDC. However, it's still extremely helpful to understands what's happening under the surface.
+
+## Strengths and weaknesses
+
+For web developers, the biggest benefit of using federated identity is reducing sign-up friction for those users who already have an account with one of the chosen IdPs. Additionally, their chosen IdPs can help websites to securely implement federated identity.
+
+From a security perspective, the biggest benefit is that because users don't have to create new credentials for each account, there's a lower risk of them choosing passwords that are easy to remember (that is, weak) or of them reusing passwords across sites.
```
hamishwillee:
```suggestion
From a security perspective, the biggest benefit is that because users don't have to create new credentials for each account, there's a lower risk of them choosing weak passwords that are easier to remember, or of them reusing passwords across sites.
```

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+We can say that federated identity is a more secure option than just passwords, but it still has problems:
```
hamishwillee:
```suggestion
Using federated identity is a more secure option than just passwords, but it still has problems:
```

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+   - `code_challenge`: A cryptographic {{glossary("hash")}} of a secret specific to this authorization request, which will be used by the token endpoint to ensure that the token request is really the counterpart of this authorization request.
+   - `scope`: A list of strings that specify which sets of user data the RP wishes to access.
+
+3. The IdP authenticates the user. The protocol does not specify a particular method for this: the IdP could use a password, a one-time password, a biometric, or any other appropriate method.
+
+4. If authentication is successful, the IdP generates the authorization code. It also stores the `code_challenge` value, and associates it with the authorization code. The IdP then redirects the browser to the RP's redirect URL, passing the authorization code as a parameter.
+
+In the token request:
+
+1. The RP makes a {{httpmethod("POST")}} request to the token endpoint. This request includes the following parameters:
```
hamishwillee:
```suggestion
1. The RP makes a {{httpmethod("POST")}} request to the IdP's token endpoint. This request includes the following parameters:
```

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+   - `client_id`: Identifies this RP to the IdP.
+   - `client_secret`: The secret used to authenticate the RP to the IdP: this could be any value previously agreed between the RP and the IdP. Instead of a shared secret, the RP and the IdP could use some alternative mechanism for client authentication, such as TLS client authentication.
+   - `grant_type`: This should be `"authorization_code"`.
+   - `code`: The authorization code.
+   - `code_verifier`: This is the original secret that was used to produce the `code_challenge` parameter in the authentication request.
+
+2. The IdP validates the request:
+   - It authenticates the RP using client authentication.
```
hamishwillee: Perhaps something along the lines of

```suggestion
   - It authenticates that the request is from the particular RP using the client secret or some other form of client authentication.
```

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+In the authentication request:
+
+1. The user asks to sign into the RP.
+
+2. The RP redirects the browser to the IdP's authorization endpoint, asking it to authenticate the user. The RP can provide various parameters along with the request, including:
+   - `client_id`: Identifies this RP to the IdP.
+   - `response_type`: Always `"code"` if we are using the two-part flow described here, which is the recommended option.
+   - `redirect_uri`: The URL in the RP, that the IdP should redirect to once it has attempted to authenticate the user. This is the URL to which the IdP will deliver the authorization code.
+   - `code_challenge`: A cryptographic {{glossary("hash")}} of a secret specific to this authorization request, which will be used by the token endpoint to ensure that the token request is really the counterpart of this authorization request.
+   - `scope`: A list of strings that specify which sets of user data the RP wishes to access.
```
hamishwillee, line 71: "Practically" what might be in the scope is not clear. I assume this perhaps means email addresses and such. Not sure whether it matters.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+#### Client authentication
+
+In the flow we have described, the RP authenticates itself to the token endpoint when it makes the token request. This means that if an attacker does manage to steal the authorization code, it still has to successfully impersonate the RP to get the tokens from the IdP.
+
+The OAuth specification distinguishes between [_confidential_ and _public_ clients](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1). Confidential clients are essentially clients that can keep a secret, and public clients are those that can't.
+
+The specification considers that clients running on the user's browser are public clients, for the same reason we've already encountered: it's too easy for an attacker to access secrets in a browser via attacks such as XSS. Clients running on a web _server_ are confidential clients.
+
+In OIDC, only confidential clients may use client authentication, because only confidential clients can be trusted to maintain the security of the client's credentials.
```
hamishwillee, line 114: I believe here we are using the term client authentication to mean "RP authenticating its identity to the IdP. It feels like at this point we're using the term client without having ever defined what a client is in this context.

This is kind of what I was getting to in https://github.com/mdn/content/pull/41779/files#r2548032506

I don't know if you agree this needs to be f"ixed", but perhaps here.

```suggestion
In OIDC, only confidential clients, such as the RP server, may can authenticate themselves as clients of the IdP, because only confidential clients can be trusted to maintain the security of the client's credentials.
```

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+- The RP generates a value that is hard to guess and is specific to this authentication request. This value is called the _code verifier_.
+- The RP creates a {{glossary("hash", "cryptographic hash")}} of the code verifier, and uses it as the `code_challenge` parameter in the authentication request.
+- The IdP stores the code challenge, and associates it with the authorization code that it returns to the RP.
+
+In the token request:
+
+- The RP passes the code verifier in the _code_verifier_ parameter.
+- The IdP hashes the code verifier, and compares the result with the stored code challenge: if they do not match, then the token request is denied.
+
+This defends against two attacks: [CSRF against the RP's redirect URL](#csrf_against_the_redirect_url), and [authorization code injection](#authorization_code_injection).
```
hamishwillee: "This" in "This defends against", is associated with the text above not the feature.

```suggestion
PKCE defends against two attacks: [CSRF against the RP's redirect URL](#csrf_against_the_redirect_url), and [authorization code injection](#authorization_code_injection).
```

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+The specification considers that clients running on the user's browser are public clients, for the same reason we've already encountered: it's too easy for an attacker to access secrets in a browser via attacks such as XSS. Clients running on a web _server_ are confidential clients.
+
+In OIDC, only confidential clients may use client authentication, because only confidential clients can be trusted to maintain the security of the client's credentials.
+
+The RP can authenticate itself to the IdP using a shared secret, but [it is better to use a method based on public-key cryptography](https://www.rfc-editor.org/rfc/rfc9700.html#name-client-authentication), such as {{glossary("TLS")}} client authentication.
+
+#### Proof Key for Code Exchange (PKCE)
+
+The `code_challenge` and `code_verifier` values that the RP provides in the authentication request and token request, respectively, are part of a mechanism called _Proof Key for Code Exchange_ (PKCE), specified in {{rfc("7636")}}.
```
hamishwillee, line 120: _Perhaps_ pull the sentence from the end https://github.com/mdn/content/pull/41779/files#r2548065911 to be  clear up front why users should care about PKCE. 

```suggestion
The `code_challenge` and `code_verifier` values that the RP provides in the authentication request and token request, respectively, are part of a mechanism called _Proof Key for Code Exchange_ (PKCE), specified in {{rfc("7636")}}.
This mechanism defends against two attacks: [CSRF against the RP's redirect URL](#csrf_against_the_redirect_url), and [authorization code injection](#authorization_code_injection).
```

You might then have to have "In the PKCE authentication request"

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: Yes, in OAuth/OIDC, "client" always means RP.

Maybe I could extend the note at the start of https://pr41779.review.mdn.allizom.net/en-US/docs/Web/Security/Authentication/Federated_identity#authentication_flow, to mention this as well as IdP->OP. And maybe in this section, when I say:

> The OAuth specification distinguishes between [confidential and public clients](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1). Confidential clients are essentially clients that can keep a secret, and public clients are those that can't.

I could instead say

> The OAuth specification distinguishes between [confidential and public clients](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1). Confidential clients are essentially **RPs** that can keep a secret, and public clients are those that can't.

...just to reinforce the usage in this section. 

I don't particularly want to say things like "client (which means the RP) everywhere", it will get tortured.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+5. The attacker starts its own OIDC authentication flow, but intercepts the IdP's authentication response, replacing the authorization code with the code it stole from the user. This is straightforward, because the authentication response is for the attacker, so it passes through the attacker's device.
+
+6. The RP then continues its authentication flow for the attacker by making the token request to the IdP, including the user's authorization code that the attacker has injected.
+
+7. The IdP responds with the user's tokens.
+
+8. The RP signs the attacker into the user's account.
+
+Note that the `state` parameter doesn't help here, because the authentication request and response really do belong to the same flow - the attacker's.
```
hamishwillee: Maybe "using the OIDC `state`..." or 
```suggestion
Note that using a `state` parameter doesn't help in this case, because the authentication request and response really do belong to the same flow - the attacker's.
```

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+However, because third-party cookies are widely used for [tracking users](/en-US/docs/Web/Privacy/Guides/Third-party_cookies#what_is_the_problem_with_third-party_cookies), browsers have taken steps to deprecate and remove support for them, and they are now not supported by default in some browsers.
+
+As such, we recommend not implementing federated identity features in a way that depends on third-party cookies.
+
+## The FedCM API
+
+The [Federated Credential Management API (FedCM API)](/en-US/docs/Web/API/FedCM_API) provides built-in browser support for federated identity. The API does not yet have cross-browser support and is still being actively developed, so we can't fully recommend its use, but it promises several benefits over implementing a protocol like OpenID Connect directly:
```
hamishwillee: Maybe chop up a little

```suggestion
The [Federated Credential Management API (FedCM API)](/en-US/docs/Web/API/FedCM_API) provides built-in browser support for federated identity. We can't fully recommend its use, because the API does not yet have cross-browser support and is still being actively developed. However it promises several benefits over implementing a protocol like OpenID Connect directly:
```

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+- In the OIDC flow we've previously described, the website using OIDC (that is, the RP) has to coordinate the interactions between itself, the user, and the IdP. As we've seen, this is complicated and error-prone. With FedCM, the browser takes care of this interaction: as an RP, you call a browser API, and the browser locates the IdP, asks the user to authenticate, and returns a token from the IdP that the RP can use to sign the user in.
```
hamishwillee:
```suggestion
- In the OIDC flow we've previously described, the website using OIDC (that is, the RP) has to coordinate the interactions between itself, the user, and the IdP. As we've seen, this is complicated, and hence error-prone. With FedCM, the browser takes care of this interaction: as an RP, you call a browser API, and the browser locates the IdP, asks the user to authenticate, and returns a token from the IdP that the RP can use to sign the user in.
```

hamishwillee: This means the token itself is available in the browser before being returned to the RP. Do we need to worry about the token be stolen or otherwise being misused?

hamishwillee: Ah, I see further down the token is not "the token" we already discussed - it might be the authentication code. This overloading of terms feels a bit confusing.

files/en-us/web/security/authentication/federated_identity/index.md
hamishwillee: Maybe add it just once in the first sentence - styled as code to highlight

> In the flow we have described, the RP `("client")` authenticates itself to the `IdP's` token endpoint when it makes the token request. This means that if an attacker does manage to steal the authorization code, it still has to successfully impersonate the RP to get the tokens from the IdP.

That makes it very clear what the client is for client authentication. I would also be happy with "as well", your suggestion:

> The OAuth specification distinguishes between [confidential and public clients](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1). Confidential clients are essentially RPs that can keep a secret, and public clients are those that can't.

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: Yeah it is confusing. These are unfortunately the terms used in the spec (I feel like when the spec was written it was assumed that this might be an actual ID token). We could use something other than "token" here though - 'object", "value". I don't think we have to use the spec language.

wbamberg: FWIW the spec says:

> The content of the [token](https://w3c-fedid.github.io/FedCM/#dom-identityassertionresponse-token) is opaque to the user agent and can contain anything that the [IDP](https://w3c-fedid.github.io/FedCM/#idp) would like to pass to the [RP](https://w3c-fedid.github.io/FedCM/#rp) to facilitate the login.

see also my related complaints at https://github.com/mdn/content/issues/42025.

hamishwillee: Yuck. I'd be tempted to keep token but make it clear that it is not the same token - because the terminology is both accurate and also is probably used in the API description and would be hard to purge. So "something like"

```suggestion
- In the OIDC flow we've previously described, the website using OIDC (that is, the RP) has to coordinate the interactions between itself, the user, and the IdP. As we've seen, this is complicated and error-prone. With FedCM, the browser takes care of this interaction: as an RP, you call a browser API, and the browser locates the IdP, asks the user to authenticate, and returns an opaque token from the IdP that the RP can use to sign the user in. In an OIDC implementation this token might be the authentication code from the IdP.
```

hamishwillee: Got to get onto my queue, so pretending I didn't see your FWIW in https://github.com/mdn/content/pull/41779#discussion_r2548115328

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: I think I felt like exactly what can go in here is too much detail for this guide, and this is the point where you can go and check the spec (or do what your IdP asked you to do).

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: Well I went with mine: https://github.com/mdn/content/pull/41779/commits/7b7aadf3494ae0006e7f3053cc7b1b723e704f29.

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: I did like this at first, but then I thought it lost the flow into the sections describing the attacks. So I didn't do it in the end.

files/en-us/web/security/authentication/federated_identity/index.md
wbamberg: I think I've addressed this with my "If PKCE were not used," edits.

files/en-us/web/security/authentication/federated_identity/index.md
hamishwillee: A matter of taste :-)

hamishwillee:
That last round of fixes LGTM. Who else are you going to get look at this?

wbamberg:
> That last round of fixes LGTM. Who else are you going to get look at this?

Thank you Hamish! I'm hoping to get some [SWAG CG](https://www.w3.org/community/swag/)-affiliated people to have a look.

files/en-us/web/security/authentication/federated_identity/index.md
```diff
@@ -0,0 +1,302 @@
...
+
+- The next most secure pattern is one in which the website uses a web server to handle all OAuth/OIDC interactions, but then returns the access token to the front end, and the front end then makes API requests directly. In this scenario the website can be a confidential client but malicious code running in the browser (for example, through an XSS attack) can potentially steal access tokens. However, the front end doesn't have to store access tokens long-term: it can retrieve them from the backend when it needs them.
+
+- The least secure pattern is one in which OAuth/OIDC interactions and interactions with APIs both take place in the front end. This, for example, would be the natural architecture for a {{glossary("SPA", "Single-page app")}}, where the entire application executes in the browser. In this architecture the RP can't be a confidential client, because it can't reliably keep a client secret. This means that it can't authenticate itself to the IdP. It also has to persistently store tokens, which increases the risk of malicious code stealing them.
```
Josh-Cena: > This, for example, would be the natural architecture for a {{glossary("SPA", "Single-page app")}}

I would avoid making this claim, because a lot of SPAs, if they have user data at all, do have a server side, in the form of REST APIs. SPA, CSR, etc., exclusively refer to the frontend architecture and don't really say much beyond "how the page is rendered". Perhaps: "This, for example, would be the natural architecture for an application that executes entirely in the browser."

wbamberg: -> https://github.com/mdn/content/pull/41779/commits/8e7a48f3a3003a490f7fbb734e3e248fd860d4de

wbamberg:
This has been open for a while and has had some review, so I'm merging it. Happy to handle any more comments in follow-ups. Thanks for your work on the review @hamishwillee , much appreciated.

PR #41794, opened 2025-11-03
Add missing child page links

Josh-Cena:
The new CSS landing pages should link to all its subpages. Also fixed some other link issues.

Draft to wait for the remaining CSS moves.

### DIFF
--- a/files/en-us/web/css/guides/colors/index.md
+++ b/files/en-us/web/css/guides/colors/index.md
@@ -87,8 +87,6 @@ The CSS color module also introduces the `CSSColorProfileRule` interface. Curren
   - : Color theory and resources, including finding the right colors to create an accessible color palette, contrast, and printing in color.
 - [Using relative colors](/en-US/docs/Web/CSS/Guides/Colors/Using_relative_colors)
   - : This article explains relative CSS color syntax, shows what the different options are, and looks at some illustrative examples.
-- [Color format converter](/en-US/docs/Web/CSS/Guides/Colors/Color_format_converter)
-  - : A tool that lets you enter or pick a color and copy its corresponding value in any CSS [color format](/en-US/docs/Web/CSS/Reference/Values/color_value).
 - [Understanding color and luminance](/en-US/docs/Web/Accessibility/Guides/Colors_and_Luminance)
   - : Color perception and using colors with color insensitive (color blind) users, reduced vision users and users with vestibular disorders or other neurological disorders in mind.
 - [WCAG 1.4.1: Color contrast](/en-US/docs/Web/Accessibility/Guides/Understanding_WCAG/Perceivable/Color_contrast)
@@ -137,4 +135,5 @@ The CSS color module also introduces the `CSSColorProfileRule` interface. Curren
 - The [`VideoColorSpace`](/en-US/docs/Web/API/VideoColorSpace) interface
 - The SVG [`<feColorMatrix>`](/en-US/docs/Web/SVG/Reference/Element/feColorMatrix) element
 - [Canvas API: applying styles and colors](/en-US/docs/Web/API/Canvas_API/Tutorial/Applying_styles_and_colors#colors)
+- [Color format converter](/en-US/docs/Web/CSS/Guides/Colors/Color_format_converter)
 - [Color mixer](/en-US/docs/Web/CSS/Guides/Colors/Color_mixer)


files/en-us/web/css/guides/pseudo-elements/index.md
```diff
@@ -12,6 +12,9 @@ Pseudo-elements are prefixed with a double colon (`::`). You add pseudo-elements
 
 Pseudo-elements enable targeting entities not included in HTML and sections of content that cannot be targeted otherwise without adding extra markup. Consider the placeholder of an {{HTMLelement("input")}} element. This is an abstract element and not a distinct node in the document tree. You can select this placeholder by using the {{CSSXref("::placeholder")}} pseudo-element. As another example, the {{CSSXref("::selection")}} pseudo-element matches the content currently highlighted by a user, allowing you to style what is matched as the user interacts with the content and changes the selection. Similarly, the {{CSSXref("::first-line")}} pseudo-element targets the first line of an element, updating automatically if the character count of the first line changes, without having to query the element's line length.
 
+> [!NOTE]
+> This page introduces a CSS module. To find an exhaustive list of all pseudo-elements defined by CSS specifications, see the [pseudo-elements](/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-elements) reference page.
+
```
dipikabh: Nice!

Do you want to move it even earlier on the page, after the first para?

files/en-us/web/css/guides/pseudo-elements/index.md
```diff
@@ -114,6 +117,7 @@ The specification also defines the `::details-content` and `::search-text` pseud
 ## See also
 
 - [Specificity](/en-US/docs/Web/CSS/CSS_cascade/Specificity)
+- [List of pseudo-elements](/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-elements)
```
dipikabh, line 120: Nice one. It should probably be the first item in this list.

files/en-us/web/css/reference/properties/index.md
dipikabh, line 1: We should now be able to replace this manual listing with `{{ListSubpages}}`.

files/en-us/web/css/reference/selectors/pseudo-classes/index.md
```diff
@@ -401,6 +401,28 @@ W
 
 - {{CSSxRef(":where", ":where()")}}
 
+### Non-standard pseudo-classes
+
+Non-standard, or vendor-prefixed pseudo-classes include the following:
+
+#### `-moz-` prefix
```
dipikabh, line 408: this h4 isn't needed

files/en-us/web/css/reference/selectors/pseudo-classes/index.md
```diff
@@ -401,6 +401,28 @@ W
...
+- {{CSSxRef(":-moz-locale-dir_rtl")}}
+- {{CSSxRef(":-moz-only-whitespace")}}
+- {{CSSxRef(":-moz-first-node")}}
+- {{CSSxRef(":-moz-drag-over")}}
+- {{CSSxRef(":-moz-loading")}}
+- {{CSSxRef(":-moz-suppressed")}}
+- {{CSSxRef(":-moz-submit-invalid")}}
+- {{CSSxRef(":-moz-user-disabled")}}
+- {{CSSxRef(":-moz-handler-blocked")}}
+- {{CSSxRef(":-moz-handler-crashed")}}
```
dipikabh: There's a more complete list at https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Mozilla_extensions#pseudo-classes.
Might be better to just add a link (in "See also") than to have to maintain this list of non-standard entities in two places.

files/en-us/web/css/reference/selectors/index.md
```diff
@@ -38,6 +38,16 @@ selector {
 - [Pseudo-element selectors](/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-elements)
```
dipikabh, line 42: This is a mistake, we should remove these two.
This list should ideally exactly match the list in sidebar under "Selectors"
The sidebar lists all child pages with this page-type:
```
  - type: listSubPages
    path: /Web/CSS/Reference/Selectors
    tags: css-selector
```

files/en-us/web/css/reference/selectors/index.md
```diff
@@ -38,6 +38,16 @@ selector {
...
 - [Type selectors](/en-US/docs/Web/CSS/Reference/Selectors/Type_selectors)
 - [Universal selectors](/en-US/docs/Web/CSS/Reference/Selectors/Universal_selectors)
+- [Namespace separator](/en-US/docs/Web/CSS/Reference/Selectors/Namespace_separator)
```
dipikabh: This is listed under combinators, `page-type: css-combinator`.

files/en-us/web/css/reference/selectors/index.md
```diff
@@ -38,6 +38,16 @@ selector {
...
+
+## Index of combinators
+
+- [Child combinator (`>`)](/en-US/docs/Web/CSS/Reference/Selectors/Child_combinator)
+- [Column combinator (`||`)](/en-US/docs/Web/CSS/Reference/Selectors/Column_combinator) {{Experimental_Inline}}
+- [Descendant combinator (whitespace)](/en-US/docs/Web/CSS/Reference/Selectors/Descendant_combinator)
+- [Next-sibling combinator (`+`)](/en-US/docs/Web/CSS/Reference/Selectors/Next-sibling_combinator)
+- [Subsequent-sibling combinator (`~`)](/en-US/docs/Web/CSS/Reference/Selectors/Subsequent-sibling_combinator)
+- [Selector list (`,`)](/en-US/docs/Web/CSS/Reference/Selectors/Selector_list)
```
dipikabh: I didn’t include this list here because the `CSS/Reference/Selectors` directory now includes selectors, combinators, pseudo-classes, and pseudo-elements. Adding one would mean we'd need to add the others as well. (It's a pity our sidebar doesn't currently support showing the hierarchy, such as under `Selectors`.)

However, to reflect the relationship between selectors and combinators, pseudo-classes, pseudo-elements, we could probably add links to their respective listing pages (instead of repeating those lists here) - [index of combinators](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Combinators#index_of_combinators), [index of pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-classes#alphabetical_index), and the list of [pseudo-elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-elements).


It's a similar situation with [Values](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values).

files/en-us/web/css/reference/index.md
dipikabh, line 1: Thanks for these updates!

files/en-us/web/css/reference/selectors/pseudo-elements/index.md
```diff
@@ -181,6 +184,39 @@ V
 - {{cssxref("::view-transition-new()")}}
 - {{cssxref("::view-transition-old()")}}
 
+### Non-standard pseudo-elements
```
dipikabh, line 187: Should we instead add links to https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Mozilla_extensions#pseudo-elements and https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Webkit_extensions#pseudo-elements

files/en-us/web/css/reference/properties/index.md
Josh-Cena: I'm unsure about this because it makes the list substantially harder to read without H3 headings. I wish ListSubpages can alphabetize the list automatically.

files/en-us/web/css/reference/selectors/pseudo-classes/index.md
Josh-Cena: This is consistent with how the `Reference/Properties` page works, and there's no guarantee that we won't document `-webkit-` pseudo-classes

files/en-us/web/css/reference/selectors/pseudo-classes/index.md
Josh-Cena: This list is trivial to update because I can automatically find all children of `Pseudo-classes` that are unlinked on this page. OTOH it's actually harder to do so for the `Mozilla_extensions` page.

files/en-us/web/css/reference/selectors/index.md
Josh-Cena: Hmmm, I think that's a mistake. The namespace separator is more like a part of a type selector and is definitely not a combinator because you don't use it to compose multiple selectors. I would change the page type instead.

files/en-us/web/css/reference/selectors/pseudo-elements/index.md
Josh-Cena: My checker requires all `css-pseudo-element` pages to be listed here, so this list is easy to update

files/en-us/web/css/reference/selectors/index.md
Josh-Cena: I removed the "Combinators" list because I didn't realize this page existed.

Josh-Cena:
I'm piggybacking some other changes related to links as my PR description already says so. I hope it's fine :)

files/en-us/web/css/guides/text/index.md
```diff
@@ -41,8 +42,10 @@ The specification also defines the `hyphenate-limit-last`, `hyphenate-limit-line
 
 ### Guides
```
dipikabh:
```suggestion
## Guides
```

files/en-us/web/css/guides/colors/index.md
```diff
@@ -131,8 +132,9 @@ The CSS color module also introduces the `CSSColorProfileRule` interface. Curren
...
 ## See also
 
-- [CSS color adjustment](/en-US/docs/Web/CSS/CSS_color_adjustment) module and the {{cssxref("print-color-adjust")}} property.
-- [CSS images](/en-US/docs/Web/CSS/CSS_images) module, which is where CSS [`<gradient>`](/en-US/docs/Web/CSS/Reference/Values/gradient) images are defined.
+- [CSS color adjustment](/en-US/docs/Web/CSS/Guides/Color_adjustment) module and the {{cssxref("print-color-adjust")}} property.
+- [CSS images](/en-US/docs/Web/CSS/Guides/Images) module, which is where CSS [`<gradient>`](/en-US/docs/Web/CSS/Reference/Values/gradient) images are defined.
 - The [`VideoColorSpace`](/en-US/docs/Web/API/VideoColorSpace) interface
 - The SVG [`<feColorMatrix>`](/en-US/docs/Web/SVG/Reference/Element/feColorMatrix) element
 - [Canvas API: applying styles and colors](/en-US/docs/Web/API/Canvas_API/Tutorial/Applying_styles_and_colors#colors)
+- [Color mixer](/en-US/docs/Web/CSS/Guides/Colors/Color_mixer)
```
dipikabh, line 139: This can be moved under https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Colors#guides after the "Color format converter" entry

files/en-us/web/css/reference/properties/index.md
dipikabh: Noted, will pass on this requirement to the engineering team

files/en-us/web/css/reference/selectors/pseudo-classes/index.md
dipikabh: Okay to keep the h4 if we also add a "#### `-webkit-` prefix" section: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Webkit_extensions#pseudo-classes

files/en-us/web/css/reference/selectors/pseudo-classes/index.md
dipikabh: Sure, let's keep. Could we make this list alphabetical?

files/en-us/web/css/reference/selectors/pseudo-elements/index.md
```diff
@@ -181,6 +184,39 @@ V
...
+
+Non-standard, or vendor-prefixed pseudo-elements include the following:
```
dipikabh:
```suggestion
Non-standard vendor-prefixed pseudo-elements include:
```

files/en-us/web/css/reference/selectors/pseudo-classes/index.md
```diff
@@ -401,6 +401,28 @@ W
 
 - {{CSSxRef(":where", ":where()")}}
 
+### Non-standard pseudo-classes
+
+Non-standard, or vendor-prefixed pseudo-classes include the following:
```
dipikabh:
```suggestion
Non-standard vendor-prefixed pseudo-classes include:
```

files/en-us/web/css/reference/selectors/index.md
dipikabh: Thanks. I can open a follow-up PR to add the links I was suggesting

files/en-us/web/css/reference/selectors/index.md
dipikabh: You're right, it's a selector. The page-type needs updating.
Similarly, [Selector list](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Selector_list) needs to be fixed: `css-combinator` -> `css-selector`.

Would you mind including the page-type fixes in this PR? The sidebar will update automatically since both Selectors and Combinators are filtered and listed by their page types.
If you do, you could reflect this change in the PR title as well ("...and update page-type") :)

The alpha index of selectors and combinators would need to be manually updated though in their respective listing pages.

files/en-us/web/css/guides/colors/index.md
Josh-Cena: This is not a guide though, it's a tool. The border tools are listed under "See also" as well. I feel like the "Color format converter" really ought to be moved here, WDYT?

files/en-us/web/css/reference/selectors/index.md
dipikabh: Thanks

files/en-us/web/css/guides/colors/index.md
dipikabh: Agree! Looking at other instances of tools, it should be the other way, that is, "Color format converter" should be in "See also" and removed from Guides listing on the page.  
(Another example is our Shape generator tool, which is also listed in the [See also](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Shapes#see_also).)

PR #41806, opened 2025-11-04
Update Firefox 145 release notes for WebDriver conforming changes

juliandescottes:
Release notes update based on the following list of [relnote worthy bugs](https://bugzilla.mozilla.org/buglist.cgi?v1=fixed&resolution=FIXED&f2=cf_status_firefox144&query_format=advanced&o1=equals&f1=cf_status_firefox145&o2=notequals&status_whiteboard=%5Bwebdriver%3Arelnote%5D&columnlist=component%2Cresolution%2Cassigned_to%2Cshort_desc%2Cbug_type%2Cstatus_whiteboard&status_whiteboard_type=substring&v2=fixed&list_id=17748021).

### DIFF
--- a/files/en-us/mozilla/firefox/releases/145/index.md
+++ b/files/en-us/mozilla/firefox/releases/145/index.md
@@ -81,13 +81,17 @@ Firefox 145 is the current [Beta version of Firefox](https://www.firefox.com/en-
 
 <!-- #### Removals -->
 
-<!-- ### WebDriver conformance (WebDriver BiDi, Marionette) -->
+### WebDriver conformance (WebDriver BiDi, Marionette)
 
-<!-- #### General -->
+#### WebDriver BiDi
 
-<!-- #### WebDriver BiDi -->
-
-<!-- #### Marionette -->
+- Implemented the `emulation.setUserAgentOverride` command, which allows to override the user-agent string used by the browser either for a set of contexts, user contexts or globally. ([Firefox bug 1987935](https://bugzil.la/1987935)).
+- Implemented the `browsingContext.downloadEnd` event, which is emitted when a download finishes (whether it is successful or canceled) ([Firefox bug 1970293](https://bugzil.la/1970293)).
+- Updated the `destination` property of the `network.beforeRequestSent` event to `document` for top-level navigations. ([Firefox bug 1985552](https://bugzil.la/1985552)).
+- Updated the `browsingContext` download events to reuse the same navigation id as the previous `browsingContext.navigationStarted` event. ([Firefox bug 1986938](https://bugzil.la/1986938)).
+- Fixed a bug for network data collection, where non-ASCII characters in response bodies were not properly encoded. ([Firefox bug 1986022](https://bugzil.la/1986022)).
+- Fixed a bug with the `network.getData` command, which would fail requests with an empty response body. ([Firefox bug 1986025](https://bugzil.la/1986025)).
+- Fixed a bug where some `network` events could be flagged as blocked even if they were not. ([Firefox bug 1989919](https://bugzil.la/1989919)).
 
 ## Changes for add-on developers
 


files/en-us/mozilla/firefox/releases/145/index.md
```diff
@@ -81,13 +81,17 @@ Firefox 145 is the current [Beta version of Firefox](https://www.firefox.com/en-
...
-<!-- ### WebDriver conformance (WebDriver BiDi, Marionette) -->
+### WebDriver conformance (WebDriver BiDi, Marionette)
 
-<!-- #### General -->
+#### WebDriver BiDi
 
-<!-- #### WebDriver BiDi -->
-
-<!-- #### Marionette -->
+- Implemented the "emulation.setUserAgentOverride" command, which allows to override the user-agent string used by the browser either for a set of contexts, user contexts or globally. ([Firefox bug 1987935](https://bugzil.la/1987935))
```
whimboo: Please fix the following for all the bullet points:

```suggestion
- Implemented the `emulation.setUserAgentOverride` command, which allows to override the user-agent string used by the browser either for a set of contexts, user contexts or globally. ([Firefox bug 1987935](https://bugzil.la/1987935)).
```

files/en-us/mozilla/firefox/releases/145/index.md
```diff
@@ -81,13 +81,17 @@ Firefox 145 is the current [Beta version of Firefox](https://www.firefox.com/en-
...
+- Implemented the "browsingContext.downloadEnd" event, which is emitted when a download finishes (whether it is successful or canceled) ([Firefox bug 1970293](https://bugzil.la/1970293))
+- Updated the "destination" property of the "network.beforeRequestSent" event to "document" for top-level navigations. ([Firefox bug 1985552](https://bugzil.la/1985552))
+- Fixed a bug for network data collection, where non-ASCII characters in response bodies were not properly encoded. ([Firefox bug 1986022](https://bugzil.la/1986022))
+- Fixed a bug with the "network.getData" command, which would fail requests with an empty response body. ([Firefox bug 1986025](https://bugzil.la/1986025))
+- Updated the "browsingContext" download events to reuse the same navigation id as the previous browsingContext.navigationStarted event. ([Firefox bug 1986938](https://bugzil.la/1986938))
```
whimboo: Wasn't that only necessary for the started event given that the implementation of dowload end had it immediately included?

Otherwise a nit:

```suggestion
- Updated the "browsingContext" download events to reuse the same navigation id as the previous `browsingContext.navigationStarted` event. ([Firefox bug 1986938](https://bugzil.la/1986938))
```

juliandescottes: When I fixed it we only had dowloadStarted implemented, but I don't want to call this out as a fix specific for download started.

whimboo: Feel free to leave as is then. I think it's subtle.

whimboo: Actually when you don't want to have it listed as a fix can you please move it up higher so that it is not in the middle of other fixes?

juliandescottes: done (but note that it is a bug fix, I'm just not writing "Fixed XYZ" every time to vary things up a bit).

juliandescottes: > Actually when you don't want to have it listed as a fix can you please move it up higher so that it is not in the middle of other fixes?

I never said I didn't want to list it as a fix, just that I didn't want to make it specific to downloadStarted (because that somehow assumes readers know when we landed a specific patch compared to the other which doesn't make sense IMO)

whimboo: Maybe then best to always use `Fix: ` as prefix for those lines that are about fixes? It might make it clearer then if MDN folks are fine with it. But nothing which blocks us now.

whimboo (approved):

PR #41809, opened 2025-11-05
Update index.md - counter-increment css example enhanced.

woutervancgi:
### Description
Made the hardcoded 100 more dynamic based on 7 and made 7 a variable easy to change, showing that numbers don't need hardcoding due to calc.

### Motivation
When playing with it, the 7 was coded at two places so needed two places to change it and 100 can't be divided by 7, so though how to make it dynamic based on the root-number.

### Additional details
https://developer.mozilla.org/en-US/docs/Web/CSS/calc
https://developer.mozilla.org/en-US/docs/Web/CSS/var

### DIFF
--- a/files/en-us/mozilla/firefox/releases/145/index.md
+++ b/files/en-us/mozilla/firefox/releases/145/index.md
@@ -81,13 +81,17 @@ Firefox 145 is the current [Beta version of Firefox](https://www.firefox.com/en-
 
 <!-- #### Removals -->
 
-<!-- ### WebDriver conformance (WebDriver BiDi, Marionette) -->
+### WebDriver conformance (WebDriver BiDi, Marionette)
 
-<!-- #### General -->
+#### WebDriver BiDi
 
-<!-- #### WebDriver BiDi -->
-
-<!-- #### Marionette -->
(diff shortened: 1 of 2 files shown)


Josh-Cena:
Thanks. I don't think this is an improvement. The original example clearly does what it's described to do: start from 100, go down by 7 each time. What you changed here is completely irrelevant to the `counter-increment` property: you are just demonstrating how `calc` works which is barely the focus. We can't make our example more complicated in a way that's not relevant to the feature being demonstrated, because that distracts the reader. You are definitely welcome to use your pattern in your own code.

woutervancgi:
@Josh-Cena I care to differ. I think it's good to have the number 7 only one time in the example. So having it as a variable is a good improvement. As the multiple places they are now in, are actually related. I think showing best practices everywhere in our examples will help people (and AI ;) ) to mimic that behavior.

Josh-Cena:
It's not mentioned multiple times? It's mentioned a single time, which is `counter-increment: sevens -7;`. It's only mentioned multiple times if you try to change the initial counter value, which again, is totally not the point of the example.

PR #41801, opened 2025-11-04
chore(ia): Reorganize CSS pages - part 5 (moving at-rules and related files under reference/at-rules)

dipikabh:
See the proposed structure at [css-tree-proposed.json](https://github.com/dipikabh/mdn-css-reorg/blob/main/css-tree-proposed.json).

### Steps performed

1. Created the `css/reference/at-rules` directory and a listing page for "At-rules" (`files/en-us/web/css/reference/at-rules/index.md`).
2. Moved the existing `css_syntax/at-rule_functions` directory under `css/reference/at-rules`.
    <details>
    <summary>Breadcrumb screenshot for at-rule functions before the reorg:</summary>
    <kbd>breadcrumb for at-rule functions before the reorganization</kbd>
    </details>
    <details>
    <summary>Breadcrumb screenshot for at-rule functions after the reorg:</summary>
    <kbd>breadcrumb for at-rule functions after the reorganization</kbd>
    </details>
3. Moved all pages with `page-type: css-at-rule` and their child directories (e.g., `@page/page-orientation`) under `css/reference/at-rules`.

4. Updated paths in the sidebar (`files/sidebars/cssref.yaml`) for "At-rules" and added a link for the new listing page.
    Also removed the entry `Web/CSS/CSS_syntax/At-rule_functions` ([current link](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_syntax/At-rule_functions)) from the `Syntax` guide grouping ([link to code on GH](https://github.com/mdn/content/blob/f336c5b6795a562c64fe859aa9ee2becf223ad8a/files/sidebars/cssref.yaml#L258)) to correctly reflect the hierarchy in breadcrumbs (as shown in previous screenshot).

 > [!NOTE]
 > - After the reorg, [At-rule functions](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_syntax/At-rule_functions) will no longer be linked from the sidebar. This needs to be addressed in the future.
 > - Also note that at-rule descriptors (such as the [font-family](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-palette-values/font-family) descriptor) currently don't appear in the sidebar. Using "`depth: 2` with `tags: css-at-rule-descriptor`" doesn't solve this because descriptors get listed at the same level as their parent at-rules, rather than being nested as child nodes under respective at-rules. A proper hierarchical solution is needed.

[...]

PR #41792, opened 2025-11-03
chore(ia): Reorganize CSS pages - part 4 (moving selector files under reference/selectors)

dipikabh:
### Description

See proposed structure at [css-tree-proposed.json](https://github.com/dipikabh/mdn-css-reorg/blob/main/css-tree-proposed.json).

1. Created `css/reference/selectors` and `css/reference/selectors/combinators` directories.
2. Created listing pages for Selectors (`files/en-us/web/css/reference/selectors/index.md`) and Combinators (`files/en-us/web/css/reference/selectors/combinators/index.md`).
3. Moved the existing `css/pseudo_classes` directory under `css/reference/selectors`.
4. Moved the existing `css/pseudo_elements` directory under `css/reference/selectors`.
5. Moved all pages with `page-type: css-selector`, `page-type: css-combinator`, `page-type: css-pseudo-class`, and `page-type: css-pseudo-element` under `css/reference/selectors`.

 > [!NOTE]
 > I've retained the "_selectors" suffix in directory names despite the redundancy with the parent `selectors/` directory. It'll help to quickly distinguish selectors from combinators when browsing the filesystem.

6. Updated paths in the sidebar (`files/sidebars/cssref.yaml`) for Selectors, Combinators, Pseudo-classes, and Pseudo-elements, and added links for the new listing pages.

7. Small follow-up to previous [PR](https://github.com/mdn/content/pull/41746): Updated `page-type` in `files/en-us/web/css/reference/properties/index.md` page. Changed from `landing-page` to `listing-page`.

----
### File structure before and after

<details>
<summary>Current structure:</summary>

```
files/en-us/web/css/
.
.
├── class_selectors
.
├── column_combinator
.
.
├── _colon_-moz-drag-over
.
├── _doublecolon_after
.
.
```

</details>

<details>
<summary>After the move:</summary>

```
files/en-us/web/css/reference/selectors/
.
.
├── class_selectors
.
├── column_combinator
.
.
├── _colon_-moz-drag-over
.
├── _doublecolon_after
.
.
```

</details>

### Breadcrumbs after the update

<kbd>Screenshot 2025-10-30 at 9 49 32 PM</kbd>

<details>
<summary>More breadcrumb screenshots</summary>

<kbd>Screenshot 2025-10-30 at 9 49 11 PM</kbd>

<kbd>Screenshot 2025-10-30 at 9 50 00 PM</kbd>

<kbd>Screenshot 2025-10-30 at 9 50 28 PM</kbd>

[...]

PR #41746, opened 2025-10-30
chore(ia): Reorganize CSS pages - part 3 (moving properties and shorthand under reference/properties)

dipikabh:
### Description

Proposed new structure: https://github.com/dipikabh/mdn-css-reorg/blob/main/css-tree-proposed.json

1. Moved the existing `css/properties` directory under `css/reference`.
2. Moved all pages with `page-type: css-property` to the new `css/reference/properties`.
    For sample: `yarn content move Web/CSS/animation-direction Web/CSS/Reference/Properties/animation-direction`

3. Moved all pages with `page-type: css-shorthand-property` to the new `css/reference/properties`.
    For sample: `yarn content move Web/CSS/animation Web/CSS/Reference/Properties/animation`

5. Added short-titles to Reference (`files/en-us/web/css/reference/index.md`) and Properties (`files/en-us/web/css/reference/properties/index.md`) landing pages so we don't repeat "CSS" in the breadcrumb:

    Currently:
    <kbd>property breadcrumb before the move</kbd>

    After the update:
    <kbd>property breadcrumb after the move</kbd>

8. Updated links in the sidebar (`files/sidebars/cssref.yaml`) for the new path to Properties.

### Before and after

Current structure:
```
files/en-us/web/css/
.
.
├── animation
├── animation-composition
.
.
├── block-size
├── border
├── border-block
.
.
```

After the move:
```
files/en-us/web/css/reference/properties
.
.
├── animation
├── animation-composition
.
.
├── block-size
├── border
├── border-block
.
.
```

### Related pull requests

- https://github.com/mdn/content/pull/41720
- https://github.com/mdn/content/pull/41735

### Related issues

Discussion: https://github.com/orgs/mdn/discussions/256
Tracker issue: https://github.com/mdn/mdn/issues/662

PR #41735, opened 2025-10-29
 chore(ia): Reorganize CSS pages - part 2 (moving layout_cookbook to how_to)

dipikabh:
### Description

Performed the following steps:

1. Created `css/how_to` and a `how_to/index.md`.
2. Moved `layout_cookbook` under `how_to`:
    `yarn content move Web/CSS/Layout_cookbook Web/CSS/How_to/Layout_cookbook`

3. To update the slug casing from "Layout_cookbook/Breadcrumb_Navigation" to "Layout_cookbook/Breadcrumb_navigation":
   - Updated the slug casing in the front matter.
   - Updated the redirect using:
     `yarn content add-redirect /en-US/docs/Web/CSS/How_to/Layout_cookbook/Breadcrumb_Navigation /en-US/docs/Web/CSS/How_to/Layout_cookbook/Breadcrumb_navigation`
4. Same steps to update the slug casing from "Layout_cookbook/Split_Navigation" to "Layout_cookbook/Split_navigation".
5. Updated all links using `node scripts/update-moved-file-links.js`.
6. Updated page-types to `how-to` (except for contribute and cookbook template).
7. Updated `cssref.yaml` to:
   - Add the "How to" section to the sidebar.
   - Add contribute and cookbook template to the sidebar.

### Before and after

Current structure:
```
files/en-us/web/css/
├── layout_cookbook
         ├── breadcrumb_navigation
         ├── card
         ├── center_an_element
         ├── column_layouts
         ├── contribute_a_recipe
         |    └── cookbook_template
         ├── grid_wrapper
         ├── list_group_with_badges
         ├── media_objects
         ├── pagination
         ├── split_navigation
         └── sticky_footers
```

After the move:
```
files/en-us/web/css/how_to
├── layout_cookbook
         ├── breadcrumb_navigation
         ├── card
         ├── center_an_element
         ├── column_layouts
         ├── contribute_a_recipe
         |    └── cookbook_template
         ├── grid_wrapper
         ├── list_group_with_badges
         ├── media_objects
         ├── pagination
         ├── split_navigation
         └── sticky_footers
```

| Sidebar(currently) | Sidebar (after the move) | 
| --- | --- |
| <kbd>layout cookbook sidebar before the move</kbd> | <kbd>layout cookbook sidebar after the move</kbd> | 

### Related issues and pull requests

Discussion: https://github.com/orgs/mdn/discussions/256
Tracker issue: https://github.com/mdn/mdn/issues/662

[...]

PR #41810, opened 2025-11-05
chore(ia): Reorganize CSS pages - part 6 (moving keywords, types, functions under reference/values)

dipikabh:
See the proposed structure at [css-tree-proposed.json](https://github.com/dipikabh/mdn-css-reorg/blob/main/css-tree-proposed.json).

### Steps performed

1. Created the `css/reference/values` directory and a listing page for "Values". (`files/en-us/web/css/reference/values/index.md`).
2. Moved these pages into `css/reference/values`:
    - Pages with `page-type: css-keyword`
    - Pages with `page-type: css-type` and their child directories (e.g., `basic-shape/circle`)
    - Pages with `page-type: css-functions` (not already inside `values`, `properties`, or `at-rules`)

3. Moved the existing `css_values_and_units/css_value_functions` directory under and as `css/reference/values/functions`.
4. Moved the existing `css_values_and_units/css_data_types` directory under and as `css/reference/values/data_types`.

6. Updated the sidebar (`files/sidebars/cssref.yaml`):
    - Added an entry for a link to "Values" with a listing of all keywords.
    - Swapped the position of Functions and Types in the sidebar, moving Types before Functions.

----
### File structure before and after

<details>
<summary>Current structure:</summary>

```
files/en-us/web/css/
.
.
├── important
├── inherit
.
.
├── atan2
├── attr
.
├── blend-mode
├── box-edge
.
├── basic-shape
│   ├── circle
│   ├── ellipse
.
.
```

</details>

<details>
<summary>After the move:</summary>

```
files/en-us/web/css/reference/values
.
.
├── important
├── inherit
.
.
├── atan2
├── attr
.
├── blend-mode
├── box-edge
.
├── basic-shape
│   ├── circle
│   ├── ellipse
.
.
```

</details>

----

### Related pull requests

- https://github.com/mdn/content/pull/41720
- https://github.com/mdn/content/pull/41735
- https://github.com/mdn/content/pull/41746
- https://github.com/mdn/content/pull/41792
- https://github.com/mdn/content/pull/41801

### Related issues

Discussion: https://github.com/orgs/mdn/discussions/256
Tracker issue: https://github.com/mdn/mdn/issues/662


files/en-us/web/css/reference/values/color_value/index.md
```diff
@@ -1,12 +1,12 @@
 ---
 title: <color>
-slug: Web/CSS/color_value
+slug: Web/CSS/Reference/Values/color_value
```
Josh-Cena, line 3: I thought you would want to remove `_value` here?

Josh-Cena:
I know we already reached "agreement" on this so not a blocker (though it's because I didn't really put much thought into it at first), but is there a reason to put functions, keywords, and types in the same folder? Types are not code entities in CSS but functions and keywords are, and to me the difference is extremely unambiguous, so I would have separated `types/` and `values/`. If you think this doesn't work for things like `<color>` which has subpages, I would suggest lifting the subpages out of `<color>` just like `url_function` is not nested under `url_value`.

dipikabh:
I am in favor of keeping the relationships intact in the filesystem. So we're keeping:
- `animation-timeline` (property) / `scroll()`
- `@import` (at-rule) / `layer()`

And similarly, `<basic-shape>` (data type) / `ellipse()` and `<color>` (data type) / `lab()`.
The nesting helps keep the context intact instead of scattering entities across directories. Moving types and functions to their respective directories will break the inherent connection between them - and it would then also be tricky to represent the relationship on the site (we're able to leverage the breadcrumbs to display the correct context for `image()`:  `CSS > Reference > Values > <image> > image()`; `image()` outside of `<image>` and plainly under `Values` will seem untethered). Additionally, data types provide a nice parent grouping in the directory structure for all the related functions, so all functions under `basic-shape` are grouped together and all functions under `easing-function` stay together.

The relationship between `url_function` and `url_value` is very different from say `<color>`/`rgb()` or `<transform-function>`/`matrix()`. Since `url()` is also a general purpose function that can be used in different contexts apart from being used as `<url>` type, they're kept as sibling directories rather than nested. 

About dropping the `_value`  identifier, I agree - we can drop it from `color_value`, `flex_value`, `position_value`, `position-area_value`, and `overflow_value` now that the same-named properties and values are in separate directories.
However, our `cssxref` macro will first need to be updated to handle the distinction or perhaps we need a new macro for values that can take the new slug `css/reference/values` (without the `_value`, `{{cssxref("position", "&lt;position&gt;")}}` or even `{{cssxref("&lt;position&gt;")}}` is currently linking to the `position` property instead of the data type).
I'll flag this to the engineering team. (see [cssxref in rari](https://github.com/mdn/rari/blob/29efb996fd19ba7e686d3028a45a358567a749f8/crates/rari-doc/src/templ/templs/links/cssxref.rs#L61) and https://github.com/mdn/rari/issues/361).

dipikabh:
Thanks for reviewing 🙌

PR #41811, opened 2025-11-05
fix(docs): Add inline technology status in the at-rules page

dipikabh:
### Description

This PR follows up on https://github.com/mdn/content/pull/41801 and adds the missing inline technology statuses to the newly added listing page.


pepelsbey (approved):
Looks good! Thank you :)

PR #41825, opened 2025-11-06
Editorial review: GPUAdapter.requestDevice() consumes adapter

chrisdavidmills:
### Description

From Chrome 140 onwards, calling `GPUAdapter.requestDevice()` multiple times on the same adapter causes the returned promise to reject. This is because the adapter is consumed by the first call. See https://developer.chrome.com/blog/new-in-webgpu-140#device_requests_consume_adapter.

Previously, you could call `GPUAdapter.requestDevice()` multiple times on the same adapter, but multiple calls would result in lost devices.

This PR documents this new behavior.

### DIFF
--- a/files/en-us/web/api/gpuadapter/requestdevice/index.md
+++ b/files/en-us/web/api/gpuadapter/requestdevice/index.md
@@ -39,12 +39,14 @@ Not all features and limits will be available to WebGPU in all browsers that sup
 
 A {{jsxref("Promise")}} that fulfills with a {{domxref("GPUDevice")}} object instance.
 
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
FYI @beaufortfrancois

beaufortfrancois:
LGTM

chrisdavidmills:
Are you OK to check and merge some of these little WebGPU bits once they've been tech reviewed, @sideshowbarker?

sideshowbarker (approved):

chrisdavidmills:
@sideshowbarker super, cheers!

PR #13965, opened 2022-03-16
Delete HTTP headers Cookie2 and Set-Cookie2

bershanskiy:
#### Summary

Delete HTTP headers Cookie2 and Set-Cookie2

#### Motivation

These headers are obsolete and have not been used anywhere in a while now. For context, the proposed standard [RFC 6265](https://datatracker.ietf.org/doc/html/rfc6265) which obsoleted them was published in April 2011.

#### Supporting details

See MDN tables for [`Cookie2`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie2#browser_compatibility) and [`Set-Cookie2`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie2#browser_compatibility).
Chrome and Firefox do not process these headers, they just treat them as any other nonsensical "magic bytes" passing on the wire.

#### Related issues

https://github.com/mdn/browser-compat-data/pull/15405

#### Metadata

- [ ] Adds a new document
- [ ] Rewrites (or significantly expands) a document
- [x] Fixes a typo, bug, or other error

PR #41830, opened 2025-11-06
Update forbidden request headers in glossary

verhovsky:
### Description

The spec says `Cookie2` and `Set-Cookie` are forbidden. Looking at the Chrome source code it also checks for `Access-Control-Request-Private-Network`, which is added to the list in [this draft spec](https://wicg.github.io/private-network-access/#forbidden-header-names).

You can try it out by opening https://example.com/ and doing

```js
await fetch('https://example.com', {
  headers: {
    'access-control-request-private-network': 'aaaaaaaaaaaaaa',
    'cookie2': 'bbbbbbbbbbbbbb',
    'set-cookie': 'ccccccccccccc',
    'Permissions-Policy': 'this one actually gets sent',
  }
});
```

and looking in the Network tab.

### Additional details

Chrome source code: https://source.chromium.org/chromium/chromium/src/+/main:net/http/http_util.cc;l=324-348;drc=c8dc70b538f1bb0862f1be58237d6e945ee81819

Spec: https://fetch.spec.whatwg.org/#forbidden-request-header

### DIFF
--- a/files/en-us/glossary/forbidden_request_header/index.md
+++ b/files/en-us/glossary/forbidden_request_header/index.md
@@ -28,7 +28,7 @@ Forbidden headers are one of the following:
 - {{HTTPHeader("Connection")}}
 - {{HTTPHeader("Content-Length")}}
 - {{HTTPHeader("Cookie")}}
-- {{HTTPHeader("Cookie2")}}
+- `Cookie2`
 - {{HTTPHeader("Date")}}
 - {{HTTPHeader("DNT")}}
 - {{HTTPHeader("Expect")}}


files/en-us/glossary/forbidden_request_header/index.md
```diff
@@ -28,16 +28,17 @@ Forbidden headers are one of the following:
...
 - {{HTTPHeader("DNT")}}
 - {{HTTPHeader("Expect")}}
 - {{HTTPHeader("Host")}}
 - {{HTTPHeader("Keep-Alive")}}
 - {{HTTPHeader("Origin")}}
-- {{HTTPHeader("Permissions-Policy")}}
 - `Proxy-` headers
 - `Sec-` headers
 - {{HTTPHeader("Referer")}}
+- {{HTTPHeader("Set-Cookie")}}
```
hamishwillee, line 41: Response header right?

files/en-us/glossary/forbidden_request_header/index.md
```diff
@@ -28,16 +28,17 @@ Forbidden headers are one of the following:
 - {{HTTPHeader("Connection")}}
 - {{HTTPHeader("Content-Length")}}
 - {{HTTPHeader("Cookie")}}
+- `Cookie2`
```
hamishwillee, line 31: What browser is this implemented in? MDN doesn't document things that have no implementation.

files/en-us/glossary/forbidden_request_header/index.md
```diff
@@ -53,6 +54,9 @@ Forbidden headers are one of the following:
 > [!NOTE]
 > While the {{HTTPHeader("Referer")}} header is listed as a forbidden header [in the spec](https://fetch.spec.whatwg.org/#forbidden-request-header), the user agent does not retain full control over it and the header can be programmatically modified. For example, when using [`fetch()`](/en-US/docs/Web/API/Window/fetch), the {{HTTPHeader("Referer")}} header can be programmatically modified via the [`referrer` option](/en-US/docs/Web/API/RequestInit#referrer).
 
+> [!NOTE]
+> Chrome also forbids `Access-Control-Request-Private-Network`
```
hamishwillee, line 58: FWIW We should document this header and add it to BCD https://github.com/mdn/browser-compat-data/issues/25467.

Because we have not done so I think this is good (because there is no way for anyone to work out the support story on this) but when BCD/docs exist this would just be in the main list.

files/en-us/glossary/forbidden_request_header/index.md
verhovsky: All 3 of them. I linked the Chrome source code and I also ran the snippet I posted above that sends the 4 mentioned headers on macOS:

Chrome
Screenshot 2025-11-13 at 17 35 18

Safari
Screenshot 2025-11-13 at 17 35 27

Firefox
Screenshot 2025-11-13 at 17 35 34

files/en-us/glossary/forbidden_request_header/index.md
verhovsky: Even though it's a response header, it's forbidden if you send it in a request too. It's listed in the `fetch` spec under both forbidden "request" and "response" headers.

files/en-us/glossary/forbidden_request_header/index.md
hamishwillee: Thanks

hamishwillee (approved):

files/en-us/glossary/forbidden_request_header/index.md
bsmth: FYI, some context on removal: 

- [x] https://github.com/mdn/content/pull/13965

hamishwillee: Thanks for that clarification @bsmth - I should have searched the history. 

I have reverted this in #41964 because Cookie2 is not part of the web platform. I suspect this will come up again until the spec gets updated to remove the mention of this header.

PR #41831, opened 2025-11-06
Remove irrelevant link

verhovsky:
Hop-by-hop compression doesn't seem related to hop-by-hop headers.

### DIFF
--- a/files/en-us/web/http/reference/headers/connection/index.md
+++ b/files/en-us/web/http/reference/headers/connection/index.md
@@ -11,19 +11,10 @@ The HTTP **`Connection`** header controls whether the network connection stays o
 If the value sent is `keep-alive`, the connection is persistent and not closed, allowing subsequent requests to the same server on the same connection.
 
 > [!WARNING]
-> Connection-specific header fields such as
-> `Connection` and {{HTTPHeader("Keep-Alive")}} are prohibited
-> in [HTTP/2](https://httpwg.org/specs/rfc9113.html#ConnectionSpecific) and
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/reference/headers/connection/index.md
```diff
@@ -18,7 +18,7 @@ If the value sent is `keep-alive`, the connection is persistent and not closed,
 > Firefox ignore them in HTTP/2 responses, but Safari conforms to the HTTP/2
 > spec requirements and does not load any response that contains them.
 
-All [hop-by-hop headers](/en-US/docs/Web/HTTP/Guides/Compression#hop-by-hop_compression), including the standard hop-by-hop headers ({{HTTPHeader("Keep-Alive")}},
+All hop-by-hop headers, including the standard hop-by-hop headers ({{HTTPHeader("Keep-Alive")}},
```
hamishwillee: I agree with you. I think we'll do this link - it is somewhat cyclic since this header is the only one listed - but if more get added later it will be useful.

```suggestion
All hop-by-hop headers, including the standard [hop-by-hop](/en-US/docs/Web/HTTP/Reference/Headers#hop-by-hop_headers) headers ({{HTTPHeader("Keep-Alive")}},
```

hamishwillee (approved):
Thanks very much @verhovsky - note, I replaced the link in https://github.com/mdn/content/pull/41831#discussion_r2501028434 with something that is not "wrong" and which might be more useful in future.

Issue #41837, opened 2025-11-07
Unprecise section about favicons


Odysseus-GH24:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Webpage_metadata

### What specific section or headline is this issue about?

Adding custom icons to your site

### What information was incorrect, unhelpful, or incomplete?

> A favicon can be added to your page by:
> 
>     Saving it in the same directory as the site's index page, saved in .ico format

### What did you expect to see?

This made me understand that the favicon had to be saved alongside the typical 'index.html', so in 'templates'.
I'm sorry if this is normally very clear, but I'm learning to code and this confused me.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/learn_web_development/core/structuring_content/webpage_metadata`
* MDN URL: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Webpage_metadata
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/learn_web_development/core/structuring_content/webpage_metadata/index.md
* Last commit: https://github.com/mdn/content/commit/f33de00c56ac53878eb2cb7cb5849df1f9ab8db7
* Document last modified: 2025-08-22T09:28:31.000Z

</details>

chrisdavidmills:
Thanks for reporting, @Odysseus-GH24; I agree that this is a bit vague. I've submitted https://github.com/mdn/content/pull/41838 to fix it; let me know what you think!

PR #41838, opened 2025-11-07
Clarify favicon addition instructions in documentation

chrisdavidmills:
### Description

Updated instructions for adding favicons to make them clearer.

Fixes https://github.com/mdn/content/issues/41837

### DIFF
--- a/files/en-us/learn_web_development/core/structuring_content/creating_links/index.md
+++ b/files/en-us/learn_web_development/core/structuring_content/creating_links/index.md
@@ -44,7 +44,7 @@ This article shows the syntax required to make a link, and discusses link best p
 
 ## What is a hyperlink?
 
-Hyperlinks are regions in a document that, when clicked or otherwise activated, cause the browser to navigate to other documents or resources, sometimes specific parts of documents.
+Hyperlinks are features of an HTML document that, when clicked or otherwise activated, cause the browser to navigate to other documents or resources, sometimes specific parts of documents.
 Hyperlinks are one of the most exciting innovations the Web has to offer.
 They've been a feature of the Web since the beginning, and are what makes the Web _a web._
 Each resource on the web has an address, known as a {{glossary("URL")}} (Uniform Resource Locator), which hyperlinks point to.
@@ -222,7 +222,7 @@ Link targets are URLs. A URL, or Uniform Resource Locator, is a string of text t
 A [web server](/en-US/docs/Learn_web_development/Howto/Web_mechanics/What_is_a_web_server) receives requests for URLs and responds with the appropriate resource. Most resources are stored as files on the server's file system, so the URLs for these resources often resemble file paths.
 
 > [!NOTE]
-> For now, pretend file paths and URLs are the same thing. We'll discuss more about the nuances in the [how do URLs translate into file paths?](#how_do_urls_translate_into_file_paths) section.
+> File paths and URLs are not the same thing, but for now, we'll talk about them like they are to facilitate understanding. We'll discuss more about the differences in the [how do URLs translate into file paths?](#how_do_urls_translate_into_file_paths) section.
 
 Let's look at an example server directory structure:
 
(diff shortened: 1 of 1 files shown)


files/en-us/learn_web_development/core/structuring_content/webpage_metadata/index.md
```diff
@@ -201,8 +201,8 @@ The humble favicon has been around for many years. It is the first icon of this
 
 A favicon can be added to your page by:
 
-1. Saving it in the same directory as the site's index page, saved in `.ico` format (most also support favicons in more common formats like `.gif` or `.png`)
-2. Adding the following line into your HTML's {{HTMLElement("head")}} block to reference it:
+1. Saving it in a supported format such as `.ico`, `.gif`, or `.png` somewhere inside your website directory (in this example, it would be saved in the same directory as the HTML file).
```
Josh-Cena: I think the main confusion is that we are mixing "directory" and "URL". If you use any framework the "directory" of the site (if there's such a concept at all) is not going to be the same thing as the root URL (such as the OP which has `templates` as the site directory, but not every file under `templates` will become the root of the site). Rather, you want to serve the favicon at `/favicon.ico` via whatever capability your tool provides, such as a `static` or `public` folder. Although you can technically place it anywhere, it's canonically at `/favicon.ico`, because if you do so you don't have to add a `<link>` at all: all browsers can automatically get this icon. Some external tools might even not read the HTML document at all and assume favicons at this location.

chrisdavidmills: Seems to me like you're overthinking this. Here we are talking about basic, static HTML, and by "directory", I'm talking about the physical folder on your computer. The reader of this is unlikely to be using a framework. Maybe I should just say "folder", update the URL to "/favicon.ico", and mention in a note that if it is placed here, most browsers will pick it up automatically?

Odysseus-GH24: Hello. I have set up the framework 'Flask' some time ago, and this was indeed what confused me. As time passed, and my experience with web development being quite new, I somehow considered the Flask directories architecture as a basic HTML one. I only heard about the addition of favicons being a thing in Flask a day or two later in its documentation. I guess a small note talking about this would be nice? This was a head scratcher for me. I hope you won't curse me and my family for 10 generations, @chrisdavidmills.

chrisdavidmills: @Odysseus-GH24 I stand corrected! I think the curses are all mine ;-)

Seriously, this is really interesting direct reader feedback, and also highlights the danger of making assumptions.

I'll update it tomorrow.

chrisdavidmills: Ok, I've made some more updates. I also took this opportunity to update the paths and URLs primer, to add some more details, distinguish between paths and URLs, and show how the two relate. That stuff was originally written about 10 years ago...

files/en-us/learn_web_development/core/structuring_content/creating_links/index.md
```diff
@@ -264,42 +294,88 @@ To do this you first have to assign an [`id`](/en-US/docs/Web/HTML/Reference/Glo
...
+```
 
-A relative URL will point to different places depending on the actual location of the file you refer from — for example if we moved our `index.html` file out of the `projects` directory and into the **root** of the website (the top level, not in any directories), the `pdfs/project-brief.pdf` relative URL link inside it would now point to a file located at `https://www.example.com/pdfs/project-brief.pdf`, not a file located at `https://www.example.com/projects/pdfs/project-brief.pdf`.
+However, there are a couple of good reasons why you shouldn't use full URLs for internal links:
 
-Of course, the location of the `project-brief.pdf` file and `pdfs` folder won't suddenly change because you moved the `index.html` file — this would make your link point to the wrong place, so it wouldn't work if clicked on. You need to be careful!
+- Efficiency: It takes more resources to look up a full URL. If you specify `https://www.example.com/projects/`, when a user clicks on the link the browser will first look up the server location, then look for the linked page. If you specify `projects`, the browser will immediately look for the linked page on the same server it is already on.
```
Josh-Cena: This is false AFAIK. Browsers are smart enough to know that this link is effectively internal and won't do a DNS lookup. I remember we had a complaint years back regarding similar content elsewhere, so we removed such claims.

chrisdavidmills:
> Hi @chrisdavidmills I have major concerns about the "creating links" guide regarding its conflation of URLs with file paths almost until the very end, so I took a stab at rewriting it slightly. You can check [ad82b72](https://github.com/mdn/content/commit/ad82b729691a3c8b76cd34a3041fff4b5707bbcf) and feel free to merge if you agree with the change.

@Josh-Cena I appreciate your efforts to try to make this more precise, and I think you mostly succeeded. Thanks.

I made a few edits to improve clarity, as some of the bits were a little confusing. This is the thing — being precise is important, but sometimes it makes sense to gloss over things in the name of aiding understanding for complete beginners. This is what guided my original wording in this article. When you know this stuff well, you think about it implicitly and automatically, but it is challenging to explain it precisely to complete beginners.

PR #41840, opened 2025-11-07
Editorial review: Add note about bgra8unorm read-only storage textures bring deprecated

chrisdavidmills:
### Description

Use of the `bgra8unorm` format for read-only storage textures is deprecated. The specification explicitly disallows this, as this format is intended for write-only access and is not portable. Any browser support for this combination is considered a bug.

Chrome previously allowed this, but it is deprecated from version 140 onwards. See https://developer.chrome.com/blog/new-in-webgpu-140#deprecate_bgra8unorm_read-only_storage_texture_usage.

This PR adds a note to make the deprecation clear. Note that I'm not intending to add a browser compat data point for this change, as:

- It is expected behavior anyway.
- We don't really want to encourage people to use it in browsers that wrongly allow it.


chrisdavidmills:
@beaufortfrancois looked at this and approved over chat, so moving to editorial review.

PR #41851, opened 2025-11-09
Fix formatting of Forms->Option description

soloiaros:
In the Descriptions column of the table, for the <option> tag, a part of the description was missing suitable formatting that  other elements of that kind had. The fix helps users navigate  the page and better understand the emphases in it.


pepelsbey (approved):
Oh, good catch! Thank you :)

Issue #41865, opened 2025-11-10
"appropriate" width tested but not discussed previously


gurglovt:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Test_your_skills/Images

### What specific section or headline is this issue about?

#Task 1

### What information was incorrect, unhelpful, or incomplete?

Minor mismatch between the test and previous article -- 
The articles talk about the importance of including `height=` and `width=` attributes, but the test asks for only width. 

Additionally, the exact size of the image is given, but the solution sets the width to `400` rather than the intrinsic `615`.

 There's some mention of using an "appropriate" width for the aspect ratio, but my understanding is that the aspect ratio will automatically match the source image if only one of the `height=` or `width= `attributes are used.

### What did you expect to see?

Some more talk about how to set an appropriate width for the aspect ratio in the previous section, or a solution with a `width=` and `height=` matching the intrinsic size of the image.

### Do you have anything more you want to share?

Wonderful course! :)

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/learn_web_development/core/structuring_content/test_your_skills/images`
* MDN URL: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Test_your_skills/Images
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/learn_web_development/core/structuring_content/test_your_skills/images/index.md
* Last commit: https://github.com/mdn/content/commit/7615562a3689a3e23a2b6b623597f4391740a53e
* Document last modified: 2025-10-13T01:34:32.000Z

</details>

chrisdavidmills:
Thanks for reporting, @gurglovt. I've submitted a fix for it.

PR #41885, opened 2025-11-10
Clarify learn HTML image width and height info

chrisdavidmills:
### Description

As noted in https://github.com/mdn/content/issues/41865, the wording of one HTML image skill test is a little confusing, and doesn't seem to match up with the guidance given in the previous article.

This PR aims to resolve that issue. 

Fixes https://github.com/mdn/content/issues/41865.

### DIFF
--- a/files/en-us/learn_web_development/core/structuring_content/html_images/index.md
+++ b/files/en-us/learn_web_development/core/structuring_content/html_images/index.md
@@ -185,7 +185,7 @@ This means that when the image has been downloaded, the browser doesn't have to 
 
 For an excellent article on the history of this feature, see [Setting height and width on images is important again](https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/).
 
-Note that, if there is no content below the image, so re-rendering is not a problem, you can just set the `width` of the image — it you set a `width` but don't set a `height`, the `height` defaults to `auto` which means it is set to a value that maintains the image's [aspect ratio](/en-US/docs/Glossary/Aspect_ratio).
+Note that if there's no content below the image, re-rendering isn't a problem because resizing the image won't cause other elements to shift. In that case, you can set the image's `width` only. If you set a `width` but don't set a `height`, the `height` defaults to `auto` which means it is set to a value that maintains the image's [aspect ratio](/en-US/docs/Glossary/Aspect_ratio).
 
 #### Resizing images
 


files/en-us/learn_web_development/core/structuring_content/html_images/index.md
```diff
@@ -175,22 +175,25 @@ Once the image is loaded, the browser adds the image to the page. Because the im
 
 ![Comparison of page layout while the browser is loading a page and when it has finished, when no size is specified for the image.](no-size.png)
 
-Moving the text like this is extremely distracting to users, especially if they have already started to read it.
+Moving the text like this is extremely distracting to users, especially if they have already started to read it, and it also causes the browser to re-render the content, which is bad for performance.
```
bsmth:
```suggestion
Moving the text like this is extremely distracting to users, especially if they have already started to read it, and it also causes the browser to re-render the page, which is bad for performance.
```

files/en-us/learn_web_development/core/structuring_content/html_images/index.md
```diff
@@ -175,22 +175,25 @@ Once the image is loaded, the browser adds the image to the page. Because the im
...
 
-> [!NOTE]
-> Although, as we have said, it is good practice to specify the _actual_ size of your images using HTML attributes, you should not use them to _resize_ images.
->
-> If you set the image size too big, you'll end up with images that look grainy, fuzzy, or too small, and wasting bandwidth downloading an image that is not fitting the user's needs. The image may also end up looking distorted, if you don't maintain the correct {{glossary("aspect ratio")}}. You should use an image editor to put your image at the correct size before putting it on your webpage.
->
-> If you do need to alter an image's size, you should use [CSS](/en-US/docs/Learn_web_development/Core/Styling_basics) instead.
+Note that, if there is no content below the image, so re-rendering is not a problem, you can just set the `width` of the image — it you set a `width` but don't set a `height`, the `height` defaults to `auto` which means it is set to a value that maintains the image's [aspect ratio](/en-US/docs/Glossary/Aspect_ratio).
```
bsmth: I'm not sure I follow, is this accurate?

```suggestion
Note that if there's no content below the image, re-rendering isn't a problem because resizing the image won't cause other elements to shift. In that case, you can set the image's `width` only. If you set a `width` but don't set a `height`, the `height` defaults to `auto` which means it is set to a value that maintains the image's [aspect ratio](/en-US/docs/Glossary/Aspect_ratio).
```

files/en-us/learn_web_development/core/structuring_content/html_images/index.md
```diff
@@ -175,22 +175,25 @@ Once the image is loaded, the browser adds the image to the page. Because the im
...
+
+#### Resizing images
+
+Although, as we have said, it is good practice to specify the _actual_ size of your images using HTML attributes, you should not use them to _resize_ images.
+
+If you set the image size too big, you'll end up with images that look grainy, fuzzy, or too small, and wasting bandwidth downloading an image that is not fitting the user's needs. The image may also end up looking distorted, if you don't maintain the correct {{glossary("aspect ratio")}}. You should use an image editor to put your image at the correct size before putting it on your webpage.
```
bsmth, line 194: Do we mention https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/aspect-ratio here?

files/en-us/learn_web_development/core/structuring_content/html_images/index.md
chrisdavidmills: Yes, that's what I meant — much better, thanks!

files/en-us/learn_web_development/core/structuring_content/html_images/index.md
chrisdavidmills: We don't, but I don't think it is needed here. I might add using `aspect-ratio` as a potential solution in the test your skills article.

PR #41899, opened 2025-11-11
Remove HDR color spec URL from colors guide

estelle:
Removed a spec URL related to HDR color..
not part of the color modules. this is a different specification


estelle:
It was erroneously added to the page, so reverting it.
Module landing pages reflect a spec. Multiple versions  of the same spec are included, but this is a "version 1" of a different spec, so should not be on the page.

PR #41901, opened 2025-11-11
Add docs for the CSS text-decoration-inset property

chrisdavidmills:
### Description

Firefox 146 onwards has the `text-decoration-inset` property enabled by default (see https://bugzilla.mozilla.org/show_bug.cgi?id=1993043; it was called `text-decoration-trim`, but it got renamed). I've tested it, and it doesn't work in Chrome/Safari yet.

This PR adds a reference page for the new property, and links to it from the text decoration module page.

### DIFF
--- a/files/en-us/web/css/reference/properties/text-decoration-inset/index.md
+++ b/files/en-us/web/css/reference/properties/text-decoration-inset/index.md
@@ -73,7 +73,7 @@ One or two {{cssxref("&lt;length>")}} values, or the keyword `auto`.
 - {{cssxref("&lt;length>")}}
   - : Specifies the amount to adjust the text decoration position by. Positive values inset the text decoration (make it shorter) while negative values outset the text decoration (make it longer). If one value is specified, it applies to both the text decoration start and end points. If two values are specified, the first one applies to the text decoration start point and the second one applies to the text decoration end point.
 - `auto`
-  - : The browser chooses a start and end inset amount to ensure that, if two decorated text boxes appear side-by-side, a gap is created between the two so they do not appear to have a single text decoration.
+  - : The browser chooses a start and end inset amount to ensure that, if two decorated text boxes appear side-by-side, the appearence of a gap is created between their text decorations so they do not appear to have a single text decoration.
 
 ## Description
 
@@ -83,7 +83,7 @@ The `text-decoration-inset` property allows you to adjust the start and/or end p
 
 A single `<length>` value will set the inset (if positive) or outset (if negative) on the start and end positions of the text decoration. To set the start and end positions separately, you can use two `<length>` values — the first one applies to the start position of the text decoration and the second one applies to the end.
 
-The `text-decoration-inset` property can also take the `auto` keyword. This causes the browser to inset the start and end positions to ensure that, if two decorated text boxes appear side-by-side, a gap is created between them so they do not appear to have a single text decoration. The `auto` value is particularly important when rendering Chinese text, where underlining is used to [punctuate proper nouns](https://www.w3.org/TR/clreq/#id88), and adjacent proper nouns should have separate underlines. See [Effect of the `auto` value](/en-US/docs/Web/CSS/Reference/Properties/text-decoration-inset#effect_of_the_auto_value) for an example.
+The `text-decoration-inset` property can also take the `auto` keyword. This causes the browser to inset the text decoration start and end points to ensure that, if two decorated text boxes appear side-by-side, they do not appear to have a single text decoration. The `auto` value is particularly important when rendering Chinese text, where underlining is used to [punctuate proper nouns](https://www.w3.org/TR/clreq/#id88), and adjacent proper nouns should have separate underlines. See [Effect of the `auto` value](/en-US/docs/Web/CSS/Reference/Properties/text-decoration-inset#effect_of_the_auto_value) for an example.
 
 The `auto` value does not have the same effect as the initial value `0`. Setting `text-decoration-inset` to `0` causes there to be no space between decorations.
 


files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+A single `<length>` value will set the inset (if positive) or outset (if negative) on the start and end positions of the text decoration. To set the start and end positions separately, you can use two `<length>` values — the first one applies to the start position of the text decoration and the second one applies to the end.
+
+The `text-decoration-inset` property can also take the `auto` keyword, which causes the browser to choose a value to ensure that, if two decorated text boxes appear side-by-side, a gap is created between them so they do not appear to have a single text decoration. The `auto` value is particularly important in languages such as Chinese, where underlining is a form of punctuation. See [Effect of the `auto` value](/en-US/docs/Web/CSS/Reference/Properties/text-decoration-inset#effect_of_the_auto_value) for en example.
```
yisibl: > The `auto` value is particularly important in languages such as Chinese, where underlining is a form of punctuation.

I believe it can be clearly stated that this is: `Proper noun marks`, and add spec links: https://www.w3.org/TR/clreq/#id88

cc @xfq

chrisdavidmills: That's really interesting; thanks for the additional information. I've updated the text to add this in https://github.com/mdn/content/pull/41901/commits/64b3ea28a8a1cc9b9e8b6649b6093a4dcb0ea1d1.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+text-decoration-inset: revert;
+text-decoration-inset: revert-layer;
+text-decoration-inset: unset;
+```
+
+### Values
+
+One or two {{cssxref("&lt;length>")}} values, or the keyword `auto`.
+
+- {{cssxref("&lt;length>")}} values
```
estelle:
```suggestion
- {{cssxref("&lt;length>")}}
```

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+  - : Specifies the amount to adjust the text decoration position by. Positive values inset the text decoration (make it shorter) while negative values outset the text decoration (make it longer). If one value is specified, it applies to both the text decoration start and end points. If two values are specified, the first one applies to the text decoration start point and the second one applies to the text decoration end point.
+- `auto`
+  - : The browser chooses an inset amount to ensure that, if two decorated text boxes appear side-by-side, a gap is created between the two so they do not appear to have a single text decoration.
+
+## Description
+
+The `text-decoration-inset` property allows you to adjust the start and/or end points of a text container's text decoration, as set by the {{cssxref("text-decoration")}} shorthand and associated longhand properties. This is useful for creating effects where you want the text decoration to be inset or outset from the text itself, or shifted in position. See [Basic use cases](/en-US/docs/Web/CSS/Reference/Properties/text-decoration-inset#basic_use_cases) for an example of each.
```
estelle:
```suggestion
By default, an element's text decoration is the same size as the rendered text. The `text-decoration-inset` property allows you to adjust the start and/or end points of a text container's text decoration, as set by the {{cssxref("text-decoration")}} shorthand and associated longhand properties. This is useful for creating effects where you want the text decoration to be inset or outset from the text itself, or shifted in position. See [Basic use cases](/en-US/docs/Web/CSS/Reference/Properties/text-decoration-inset#basic_use_cases) for an example of each.
```

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+A single `<length>` value will set the inset (if positive) or outset (if negative) on the start and end positions of the text decoration. To set the start and end positions separately, you can use two `<length>` values — the first one applies to the start position of the text decoration and the second one applies to the end.
+
+The `text-decoration-inset` property can also take the `auto` keyword, which causes the browser to choose a value to ensure that, if two decorated text boxes appear side-by-side, a gap is created between them so they do not appear to have a single text decoration. The `auto` value is particularly important when rendering Chinese text, as underlining is used to [punctuate proper nouns](https://www.w3.org/TR/clreq/#id88), and adjacent proper nouns should have separate underlines. See [Effect of the `auto` value](/en-US/docs/Web/CSS/Reference/Properties/text-decoration-inset#effect_of_the_auto_value) for an example.
+
+The `auto` value does not have the same effect as a `text-decoration-inset` value of `0` (which is the initial value) — `0` causes there to be no space between decorations.
```
estelle:
```suggestion
The `auto` value does not have the same effect as a `text-decoration-inset` value of `0`. The initial value `0` causes there to be no space between decorations.
```

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Formal syntax
+
+{{csssyntax}}
+
+## Examples
+
+### Basic use cases
+
+In this example we demonstrate an outset, inset, and "shifted" use case.
```
estelle:
```suggestion
In this example, we demonstrate the outset, inset, and "shifted" use cases.
```

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+  text-decoration: line-through 5px white;
+  text-decoration-inset: 0.5em;
+}
+
+#three {
+  text-decoration: underline wavy 2px blue;
+  text-decoration-inset: 1em -1em;
+}
+```
+
```
estelle, line 161: Let's add something like the following to both examples as "css hidden":

```css hidden
@supports not (text-decoration-inset: auto) {
  body::before {
    content: "Your browser doesn't support the `text-decoration-inset` property.";
    background-color: wheat;
    display: block;
    width: 100%;
    text-align: center;
  }
}
```

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+```html live-sample___auto-example
+<p id="one"><u>石井</u><u>艾俐俐</u></p>
+
+<p id="two"><u>石井</u><u>艾俐俐</u></p>
+```
+
+#### CSS
+
+Each `<u>` element has a `red` color and `3px` thickness set on its underline. The first group of `<u>` elements has a `text-decoration-inset` value of `auto` set on them, while the second set has a `text-decoration-inset` value of `0` set on them, for comparison purposes:
```
estelle: @copilot 
```suggestion
Each `<u>` element has a `red` color and `3px` thickness set on its underline. The first group of `<u>` elements has a `text-decoration-inset` value of `auto` set on them, while the second set has the initial `text-decoration-inset` value of `0` explicitly set on them, for comparison purposes:
```
just to iterate 0 is the default.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+u {
+  font-family: sans-serif;
+  font-size: 2em;
+}
+```
+
+```css live-sample___auto-example
+u {
+  text-decoration-color: red;
+  text-decoration-thickness: 3px;
```
estelle, line 207: i would go with `text-decoration: red 3px underline` , but not required edit.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+### Effect of the `auto` value
+
+This example demonstrates the effect of the `text-decoration-inset: auto` value.
+
+#### HTML
+
+We define two groups of side-by-side {{htmlelement("u")}} elements:
+
+```html live-sample___auto-example
+<p id="one"><u>石井</u><u>艾俐俐</u></p>
```
estelle: should we add the `lang` attribute?

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+}
+```
+
+#### Result
+
+This renders like so:
+
+{{embedlivesample("auto-example", "100%", "200")}}
+
+Note how the `auto` value creates a gap in the underlines, whereas the `0` value results in no gap.
```
estelle:
```suggestion
Note how the `auto` value insets the text decoration, creating a gap in the underline between the two elements (but adding no space between the two elements), whereas the `0` value results in no gap.
```
it insets on both sides, which is subtle, so may not be noticed if not pointed out, so suggest explicitly pointing it out. Also, the "gap" is in the underlines but not in the text, so want to point that out too. My suggestion may not be the best verbiage, but i think it is important to point out.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+A single `<length>` value will set the inset (if positive) or outset (if negative) on the start and end positions of the text decoration. To set the start and end positions separately, you can use two `<length>` values — the first one applies to the start position of the text decoration and the second one applies to the end.
+
+The `text-decoration-inset` property can also take the `auto` keyword, which causes the browser to choose a value to ensure that, if two decorated text boxes appear side-by-side, a gap is created between them so they do not appear to have a single text decoration. The `auto` value is particularly important when rendering Chinese text, as underlining is used to [punctuate proper nouns](https://www.w3.org/TR/clreq/#id88), and adjacent proper nouns should have separate underlines. See [Effect of the `auto` value](/en-US/docs/Web/CSS/Reference/Properties/text-decoration-inset#effect_of_the_auto_value) for an example.
```
estelle:
```suggestion
The `text-decoration-inset` property can also take the `auto` keyword, which causes the browser to inset the start and end positions ensuring that, if two decorated text boxes appear side-by-side, a gap is created between them so they do not appear to have a single text decoration. The `auto` value is particularly important when rendering Chinese text, as underlining is used to [punctuate proper nouns](https://www.w3.org/TR/clreq/#id88), and adjacent proper nouns should have separate underlines. See [Effect of the `auto` value](/en-US/docs/Web/CSS/Reference/Properties/text-decoration-inset#effect_of_the_auto_value) for an example.
```

estelle (commented):
Looks good. Just a few suggestions.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
chrisdavidmills: I'd rather not do this; I was kind of going for a subtle hint that the underline created on `<u>` is a text decoration. But at the same time, I didn't want to waffle on about that explicitly because it seemed out of scope for the article.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
yisibl: It's not between two elements, but rather the underscore shortens at both ends, creating the appearance of a gap.

I believe it's important to explicitly clarify the rendering method here, as it's otherwise prone to misunderstanding. Before this property existed, Chinese developers would create gaps using `u+u { margin-left: .125em;}`. but it's not the same as `text-decoration-inset`.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
chrisdavidmills: Saying this is a good idea, but I've shifted the order of the text around a bit, so that the first sentence says:

> By default, an element's text decoration, as set by the {{cssxref("text-decoration")}} shorthand and associated longhand properties, is the same size as the rendered text.

I think it reads better this way.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
chrisdavidmills: Makes sense, but I've reordered this to:

> The `auto` value does not have the same effect as the initial value `0`. Setting `text-decoration-inset` to `0` causes there to be no space between decorations. I thought it better to mention it is the initial value on the first mention of `0`, not the second.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
yisibl: Additionally, the approach in this article that uses `border-image` to simulate `text-decoration-inset` can be considered a polyfill. Could you look into adding it to this documentation?

https://chaoli.club/index.php/7000

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
chrisdavidmills: Great idea. I've tweaked the styles a little bit and added the block to both examples. Works nicely.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
chrisdavidmills: Yeah, I think that's probably a good idea. Added.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
chrisdavidmills: Good call. I played with the text a bit and ended up updating it to:

> Note how the `auto` value insets the text decoration subtly on both sides, creating a gap in between the underlines of the two elements (no space is added between the two elements themselves). The `0` value results in no gap.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
chrisdavidmills: Again, good to make this clear. In my next commit, I've added a variation of this, but also made a few other tweaks to the paragraph.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
chrisdavidmills: > It's not between two elements, but rather the underscore shortens at both ends, creating the appearance of a gap.

@yisibl OK, thanks for the clarification. In light of your comment, I've improved the `auto` value description on the page further.

In terms of adding polyfill/previous technique information, I'm not sure about this — this is not something we generally do on MDN, at least not these days. Polyfills change and go out of date, and adding previous technique information would also create a huge maintenance burden. So, I'm going to leave this information out.

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+The `text-decoration-inset` property can also take the `auto` keyword. This causes the browser to inset the text decoration start and end points to ensure that, if two decorated text boxes appear side-by-side, they do not appear to have a single text decoration. The `auto` value is particularly important when rendering Chinese text, where underlining is used to [punctuate proper nouns](https://www.w3.org/TR/clreq/#id88), and adjacent proper nouns should have separate underlines. See [Effect of the `auto` value](/en-US/docs/Web/CSS/Reference/Properties/text-decoration-inset#effect_of_the_auto_value) for an example.
+
+The `auto` value does not have the same effect as the initial value `0`. Setting `text-decoration-inset` to `0` causes there to be no space between decorations.
+
+The `text-decoration-inset` property is not inherited, and it is not a constituent property of the {{cssxref("text-decoration")}} shorthand.
```
pepelsbey, line 90: I would suggest saying it a little simpler.

```suggestion
The `text-decoration-inset` property is not inherited, and it is not a part of the {{cssxref("text-decoration")}} shorthand.
```

pepelsbey (approved):
Looks good! Just a small nit below

files/en-us/web/css/reference/properties/text-decoration-inset/index.md
chrisdavidmills: I'm not sure I agree with this. My text is precise, and we do use this term commonly around the CSS section of CSS.

PR #41903, opened 2025-11-11
Add Fx146 rel note for text-decoration-inset support

chrisdavidmills:
### Description

Firefox 146 onwards has the `text-decoration-inset` property enabled by default (see https://bugzilla.mozilla.org/show_bug.cgi?id=1993043; it was called `text-decoration-trim`, but it got renamed).

This PR adds a rel note for it in the Firefox 146 rel notes.

### DIFF
--- a/files/en-us/mozilla/firefox/releases/146/index.md
+++ b/files/en-us/mozilla/firefox/releases/146/index.md
@@ -1,13 +1,13 @@
 ---
-title: Firefox 146 release notes for developers (Nightly)
-short-title: Firefox 146 (Nightly)
+title: Firefox 146 release notes for developers (Beta)
+short-title: Firefox 146 (Beta)
 slug: Mozilla/Firefox/Releases/146
 page-type: firefox-release-notes-active
 sidebar: firefox
 ---
 
 This article provides information about the changes in Firefox 146 that affect developers.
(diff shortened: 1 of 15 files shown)


files/en-us/mozilla/firefox/releases/146/index.md
```diff
@@ -34,6 +34,8 @@ Firefox 146 is the current [Nightly version of Firefox](https://www.firefox.com/
 
 ### CSS
 
+- The {{cssxref("text-decoration-inset")}} property is now supported, which enables adjusting the start and end points of an element's {{cssxref("text-decoration")}} so it can be shortened, lengthened, or have its position shifted with respect to the rendered text.
```
bsmth, line 37: Small suggestion if you like it:

```suggestion
- The {{cssxref("text-decoration-inset")}} property is now supported, which enables adjusting the start and end points of an element's {{cssxref("text-decoration")}} so decorative lines on the text can be shortened, lengthened, or have their position shifted with respect to the rendered text.
```

chrisdavidmills: I don't think this is needed; saying `text-decoration` should be enough for people to know what we mean. Those who don't know what text decoration is can click through to the property pages, look at demos, and read more.

Issue #39084, opened 2025-04-11
Mention Firefox Anti-fingerprinting on `getImageData()` page


caugner:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData

### What information was incorrect, unhelpful, or incomplete?

The page for `CanvasRenderingContext2D.getImageData()` does not mention that the result may be influenced by anti-fingerprinting, e.g. [Firefox's protection against fingerprinting](https://support.mozilla.org/en-US/kb/firefox-protection-against-fingerprinting):

> Random data is introduced to background images when the image is read back by the website. If a website merely renders data to the background, it will render without alteration. Although typically this does not happen, if the website reads the image data in the background (and potentially displays it to you again), it will have subtle noise that may affect how the image is displayed.

### What did you expect to see?

It would be useful for web developers that use this feature, to mention this potential impact of anti-fingerprinting on the page.

### Do you have any supporting links, references, or citations?

This was originally reported in [this BCD issue](https://github.com/mdn/browser-compat-data/issues/26256#issuecomment-2797565202).

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/api/canvasrenderingcontext2d/getimagedata`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/canvasrenderingcontext2d/getimagedata/index.md
* Last commit: https://github.com/mdn/content/commit/e9b6cd1b7fa8612257b72b2a85a96dd7d45c0200
* Document last modified: 2025-04-10T14:56:07.000Z

</details>

MakenRosa:
A focused docs change can stay on the `getImageData()` page and avoid turning this into a broad privacy article.

Implementation slice:

1. Add a short note in the usage/return-value area explaining that browsers may alter read-back pixel data when anti-fingerprinting protections are enabled.
2. Phrase it as an implementation/privacy behavior rather than a spec guarantee: the returned `ImageData` can contain subtle noise or other privacy-preserving changes.
3. Mention the practical impact: code that compares exact pixel values, performs visual regression checks, or round-trips canvas pixels may see differences even when drawing commands are identical.
4. Link to the existing anti-fingerprinting/support reference from the issue, and keep any browser-specific wording limited to Firefox as the cited example.
[...]

PR #41906, opened 2025-11-11
Warn that `CanvasRenderingContext2D#getImageData()`, `toDataURL`, and `toBlob` may slightly randomize its return value

valadaptive:
### Description

[Firefox 145 enables additional tracking protections by default](https://blog.mozilla.org/en/firefox/fingerprinting-protections/), one of which is randomizing the data returned by `getImageData` (and possibly `toDataURL` and `toBlob`). Previously, I believe Brave did this as well, and Firefox did it with `privacy.resistFingerprinting` enabled, but this is the first time a major browser has enabled it by default.

### Motivation

Developers should be aware that they cannot rely on the canvas API to render consistently across browsers or browsing sessions. More specifically, while most developers probably expect things like antialiasing to look different across browser engines, they probably *don't* expect something like a solid color fill to have random noise added to it.

There seem to be a lot of tutorials that expect `getImageData()`, `toDataURL()`, and `toBlob()` to return the actual pixel data displayed to the end user. In this PR, I've only focused on adding warnings about the surprising behavior. I'm not yet sure what to do about the tutorials. One suggestion for "advanced" users requiring deterministic rendering would be to ship a 2D graphics library in WebAssembly, but I'm not sure about the rest.

### Additional details

Brave described their randomization approach in https://github.com/whatwg/html/issues/5373. I'm not sure if Firefox has done so anywhere.

### DIFF
--- a/files/en-us/web/api/canvasrenderingcontext2d/getimagedata/index.md
+++ b/files/en-us/web/api/canvasrenderingcontext2d/getimagedata/index.md
@@ -28,7 +28,7 @@ transparent black in the returned `ImageData` object.
 >
 > This means that the returned pixel data might not always be what you expect. For
 > instance, if you set the fill to `rgba(255, 127, 0, 100%)`, draw a rectangle, and call
-> `getImageData()`, the raw RGBA pixel data may *not* be `[255, 127, 0, 255, ...]`!
+> `getImageData()`, the raw RGBA pixel data may _not_ be `[255, 127, 0, 255, ...]`!
 
 You can find more information about `getImageData()` and general
 manipulation of canvas contents in [Pixel manipulation with canvas](/en-US/docs/Web/API/Canvas_API/Tutorial/Pixel_manipulation_with_canvas).


Josh-Cena:
This seems to fix https://github.com/mdn/content/issues/39084?

Josh-Cena (commented):
1. Could you remove the extra wrapping? We either wrap at each sentence or not wrap lines at all (I do the latter).
2. I think we ought to say more about this. Perhaps a central paragraph somewhere (such as `getImageData`) that says (1) why is this a fingerprinting vector (2) to what degree can it be randomized (e.g. how many pixels? how much perturbance?) (3) does it always randomize or only when reading cross-origin etc. Then we can rephrase the other warnings to link to this paragraph instead.

valadaptive:
I'm not sure I can bring myself to write any paragraphs justifying this behavior. As someone who's written a lot of multimedia code on the web (even something as simple as resizing an image requires canvas readback), I'm very unhappy about this change and wanted to add these warnings to dissuade unwitting developers from using the canvas API going forward.

PR #41908, opened 2025-11-12
Fix jQuery content escaping description in documentation

ajayojha:
### Description
 jQuery .attr() does not sanitize values.

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
@@ -43,7 +43,7 @@ Here, the contents of `data.className` or `data.color` could contain HTML that c
 
 ### jQuery
 
-When using jQuery, the `text()` function escapes content as it's added to a DOM. So, the "favorite color" example from above, implemented in jQuery, would look like this:
+When using jQuery, functions such as `attr()` and `text()` avoid treating any character as HTML syntax. So, the "favorite color" example from above, implemented in jQuery, would look like this:
 
 ```js example-good
 let node = $("<div>");


files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
```diff
@@ -43,7 +43,7 @@ Here, the contents of `data.className` or `data.color` could contain HTML that c
 
 ### jQuery
 
-When using jQuery, functions such as `attr()` and `text()` escape content as it's added to a DOM. So, the "favorite color" example from above, implemented in jQuery, would look like this:
+When using jQuery, the `text()` function escapes content as it's added to a DOM. So, the "favorite color" example from above, implemented in jQuery, would look like this:
 
 ```js example-good
 let node = $("</div>");
```
Josh-Cena:
```suggestion
const node = $("<div>");
```

files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
```diff
@@ -43,10 +43,10 @@ Here, the contents of `data.className` or `data.color` could contain HTML that c
 
 ### jQuery
 
-When using jQuery, functions such as `attr()` and `text()` escape content as it's added to a DOM. So, the "favorite color" example from above, implemented in jQuery, would look like this:
```
Josh-Cena, line 46: What do you mean by `attr()` not escaping content? It's escaping content perfectly for me:

```js
const node = $("<div>");
const node2 = $("<span>");
node2.attr("a", '""\'\'');
node.append(node2);
console.log(node.html());
// <span a="&quot;&quot;''"></span>
```

Josh-Cena (changes requested):

files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
ajayojha: The html() method escapes certain characters, it's not attr() method.

Josh-Cena: What the text here means by "escape" is not that when you run `node.getAttribute("a")` you will get a string that literally contains the substring `&quot;`. It's saying that your provided text is guaranteed to be inserted as text, and never interpreted as HTML. For example, with the naïve example of `` `<div class='${data.className}'>Your favorite color is now ${data.color}</div>` `` above, if `className` contains `'`, then this causes HTML injection. This is impossible if you use `attr()`. The "escape" here is a sloppy equivalent of saying "avoid HTML injection".

ajayojha: Thanks, I agree `attr()` prevents HTML parsing-based injection in the attribute context. My concern is that the current phrase “**escape content**” can be read as sanitizing the value, which `attr()` does not do. A value can be safely inserted as an attribute yet still be semantically dangerous.   ex. 
```js
const element = $("<a>").text("Click Me");
element.attr("href", "javascript:alert('XSS')"); // remains javascript: in the DOM
$("body").append(element);
console.log(element[0].getAttribute("href")); // "javascript:alert('XSS')"
```

Josh-Cena: Sanitization is not escaping though. Those are definitely unrelated terms. Perhaps:

> When using jQuery, functions such as `attr()` and `text()` avoids treating any character as HTML syntax.

ajayojha: Thanks, your suggested wording is good. I have made small fix "**avoids** >> **avoid**":

> When using jQuery, functions such as attr() and text() avoid treating any character as HTML syntax.

files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
```diff
@@ -43,7 +43,7 @@ Here, the contents of `data.className` or `data.color` could contain HTML that c
...
+When using jQuery, the `text()` function escapes content as it's added to a DOM. So, the "favorite color" example from above, implemented in jQuery, would look like this:
```
Josh-Cena:
```suggestion
When using jQuery, functions such as `attr()` and `text()` avoid treating any character as HTML syntax. So, the "favorite color" example from above, implemented in jQuery, would look like this:
```

PR #41910, opened 2025-11-12
Update gradientTransform translation value

DioF:
Fixing a bug that prevents the second (right-most) square to show the applied gradient transformation.

### Description

Adjusted the x-value of the `translate` transformation to put the gradient into the right position to become visible.

### Motivation

The second (right) blue square which is supposed to show the gradient transform is flat blue. The issue is that it uses the `gradientUnits="userSpaceOnUse"` attribute but the values for cx, cy, etc. are still refering to the position of the first (left) square.

### Additional details

Fixing example at https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/gradientTransform#example

### DIFF
--- a/files/en-us/web/svg/reference/attribute/gradienttransform/index.md
+++ b/files/en-us/web/svg/reference/attribute/gradienttransform/index.md
@@ -47,7 +47,7 @@ svg {
     r="100"
     fx="100"
     fy="100"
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
LGTM; thanks for the fix, @DioF!

PR #41920, opened 2025-11-13
Fix copy-pasta in break-inside example

igorsantos07:
### Description

Pretty much a fix to a typo from a possible copy-paste of an example for break-before.

### Motivation

Fix a typo that was confusing to me.

### Additional details

I checked the `break-after` example as well, that's already correct.

_None_


estelle (approved):
Thank you! 🎉

PR #41922, opened 2025-11-13
CSS guides: change css links to macros

estelle:


chrisdavidmills:
Removing my review; feel free to add me back when this is out of draft/ready for review.

hamishwillee:
@estelle Previous discussions of using macro vs links have resulted in "it literally does not matter which you use, even for consistency". Just FMI, is there no some sort of guideline that changes that? ( I don't "care", other than I might have missed a way of working rule).

estelle:
> @estelle Previous discussions of using macro vs links have resulted in "it literally does not matter which you use, even for consistency". Just FMI, is there no some sort of guideline that changes that? ( I don't "care", other than I might have missed a way of working rule).

Cognitively macros are much easier, esp in the CSS section, so limiting my edits to this section (and a few glossary pages)

PR #41928, opened 2025-11-13
relnote(Fx145): BTP runs in 'stateless' mode by default now

bsmth:
### Description

Relnote for BTP which runs in 'stateless' mode by default in Fx145

- https://bugzilla.mozilla.org/show_bug.cgi?id=1990831
- Explainer: https://github.com/privacycg/nav-tracking-mitigations/blob/b28a3a7479f6f5c68ebfe344f53fb924cc093065/explainers/bounce-tracking-mitigations.md

- [x] parent: https://github.com/mdn/content/issues/41819

### DIFF
--- a/files/en-us/mozilla/firefox/releases/145/index.md
+++ b/files/en-us/mozilla/firefox/releases/145/index.md
@@ -40,7 +40,7 @@ No notable changes.
 ### Security
 
 - When Bounce Tracking Protection (BTP) is enabled, it now runs in "stateless" mode by default.
-  In "stateless" mode, the browser no longer flags only sites that are part of a "bounce" that set state information (such as a cookie); it flags _all_ sites that are part of a "bounce". See [Bounce tracking mitigations](/en-US/docs/Web/Privacy/Guides/Bounce_tracking_mitigations) for more information on how BTP works.
+  In "stateless" mode, the browser no longer only flags sites that are part of a "bounce" that set state information (such as a cookie); it flags _all_ sites that are part of a "bounce". See [Bounce tracking mitigations](/en-US/docs/Web/Privacy/Guides/Bounce_tracking_mitigations) for more information on how BTP works.
   ([Firefox bug 1990831](https://bugzil.la/1990831)).
 
 ### APIs


files/en-us/mozilla/firefox/releases/145/index.md
```diff
@@ -37,6 +37,13 @@ No notable changes.
   Note that the [`endpoints`](/en-US/docs/Web/HTTP/Reference/Headers/Integrity-Policy#endpoints) key is not yet supported (violations are logged to console).
   ([Firefox bug 1984973](https://bugzil.la/1984973)).
 
+### Security
+
+- When Bounce Tracking Protection (BTP) is enabled, it now runs in "stateless" mode by default.
+  In "stateless" mode, the browser no longer stores which sites acted as bounce trackers and instead detects tracking behavior during navigation for simpler protection that works consistently across browsers.
```
chrisdavidmills: This doesn't sound quite right to me. Assuming bounce tracking protection works similarly across Fx and Chrome, I think this should be changed to talk more about the fact that the browser detects _all_ sites that are part of a "bounce" that haven't been interacted with in any way for a certain timeframe (see [this bit of my article](https://developer.mozilla.org/en-US/docs/Web/Privacy/Guides/Bounce_tracking_mitigations#how_bounce_tracking_mitigations_work)) and deletes their state, and not just just sites that are part of a "stateful bounce", i.e. a bounce where state was set by the redirect site.

In stateless mode (which Chrome also does by default now), the browser still stores which sites are part of a bounce and therefore may need to have their state deleted. However, it now stores all sites, whether they set some kind of state or not, not just ones that set state.

I asked the Chrome engineers about this, and they said the reason for monitoring all bounces and not just stateful bounces is as follows:

> "State" is a bit overloaded of a term, because to some it means cookies, localStorage, and other manually set state; and to others it also includes network state—which is automatically set but can still be manipulated to track users.  Once you consider network state, which we now do, every bounce becomes effectively stateful and it is no longer useful to discuss "stateful bounce" as a separate concept.

chrisdavidmills (changes requested):
@bsmth a comment for you, based on my recent documenting of this feature.

files/en-us/mozilla/firefox/releases/145/index.md
chrisdavidmills: @bsmth I'll update this PR as well, and you can review.

chrisdavidmills: Updated in https://github.com/mdn/content/pull/41928/commits/085658288155b31aaa7d4815bb6df1b733d14210; Let me know what you think.

files/en-us/mozilla/firefox/releases/145/index.md
```diff
@@ -37,6 +37,12 @@ No notable changes.
   Note that the [`endpoints`](/en-US/docs/Web/HTTP/Reference/Headers/Integrity-Policy#endpoints) key is not yet supported (violations are logged to console).
   ([Firefox bug 1984973](https://bugzil.la/1984973)).
 
+### Security
+
+- When Bounce Tracking Protection (BTP) is enabled, it now runs in "stateless" mode by default.
+  In "stateless" mode, the browser no longer flags only sites that are part of a "bounce" that set state information (such as a cookie); it flags _all_ sites that are part of a "bounce". See [Bounce tracking mitigations](/en-US/docs/Web/Privacy/Guides/Bounce_tracking_mitigations) for more information on how BTP works.
```
bsmth:
```suggestion
  In "stateless" mode, the browser no longer only flags sites that are part of a "bounce" that set state information (such as a cookie); it flags _all_ sites that are part of a "bounce". See [Bounce tracking mitigations](/en-US/docs/Web/Privacy/Guides/Bounce_tracking_mitigations) for more information on how BTP works.
```

files/en-us/mozilla/firefox/releases/145/index.md
bsmth: Nice thanks! I can't approve seeing as I opened the PR, so I'll let you add a +1 if you're happy 👍🏻

bsmth:
> LGTM. Sorted!

Thanks a lot!

PR #41939, opened 2025-11-14
Added example showing normal text characters can also be percent-encoded

Worldwidebrine:
### Description

Added an example showing that normal text characters can also be percent-encoded, even if they don't have to.

### Motivation

```js
const hex = [];
hex.length = 256;
for (let i = 0; i < 256; i++) {
  hex[i] = i.toString(16).padStart(2, "0");
}
console.log(hex);
fetch(`data:application/octet-stream;,%${hex.join("%")}`)
  .then((resp) => resp.arrayBuffer())
  .then((buf) => {
    const bytes = new Uint8Array(buf);
    console.log([...bytes]);
    for (let i = 0; i < 256; i++) {
      if (bytes[i] !== i) {
        console.log(`buf[${i}] !== ${i}`);
      }
    }
    console.log("finished");
  });
```

### DIFF
--- a/files/en-us/web/uri/reference/schemes/data/index.md
+++ b/files/en-us/web/uri/reference/schemes/data/index.md
@@ -32,7 +32,7 @@ A few examples:
 - `data:,Hello%2C%20World%21`
   - : The text/plain data `Hello, World!`. Note how the comma is {{Glossary("Percent-encoding", "percent-encoded")}} as `%2C`, and the space character as `%20`.
 - `data:text/plain,Hello%2C%20%57%6F%72%6C%64%21`
-  - : The text/plain data `Hello, World!` that also has `World` escaped. Similar to {{jsxref("decodeURIComponent()")}}, it decodes all {{Glossary("Percent-encoding", "percent-encoded")}} characters, even if they don't have to be escaped.
+  - : The text/plain data `Hello, World!` that also has `World` escaped. Similar to {{jsxref("decodeURIComponent()")}}, it decodes all {{Glossary("Percent-encoding", "percent-encoded")}} characters, even if they don't have to be escaped.
 - `data:text/plain;base64,SGVsbG8sIFdvcmxkIQ==`
   - : base64-encoded version of the above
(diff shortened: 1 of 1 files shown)


files/en-us/web/uri/reference/schemes/data/index.md
```diff
@@ -31,6 +31,8 @@ A few examples:
 
 - `data:,Hello%2C%20World%21`
   - : The text/plain data `Hello, World!`. Note how the comma is {{Glossary("Percent-encoding", "percent-encoded")}} as `%2C`, and the space character as `%20`.
+- `data:text/plain,Hello%2C%20%57%6F%72%6C%64%21`
+  - : The text/plain data `Hello, World!` that also has `World` escaped. Similar to {{jsxref("decodeURIComponent()")}}, it decodes all {{Glossary("Percent-encoding", "percent-encoded")}} characters, even if they don't have to be escaped.
```
chrisdavidmills:
```suggestion
  - : The text/plain data `Hello, World!`, with the `World` characters percent-encoded as well as the comma and space characters: you can percent-encode any characters, if they don't need to be encoded. Note that {{jsxref("decodeURIComponent()")}} can be used to decode all encoded characters.
```

Would something like the above be more appropriate? Take my suggestion with a pinch of salt, as I have a relatively limited understanding of these kinds of JS built-ins.

But looking at this from a relative beginner's point of view, your point is basically that you can percent-encode any characters, even if they don't need to be percent-encoded, and you are also trying to say that `decodeURIComponent()` can decode any of these characters?

Your original text sounded like it was saying that the data URI representation was decoding characters, just the same as `decodeURIComponent()` can, which doesn't sound quite right — the data URI is surely just a representation; it is not decoding anything?

chrisdavidmills (changes requested):
Thanks, @Worldwidebrine.

As you'll see from my comments below, I did find the addition a little confusing. If I've totally misunderstood this, maybe you can explain your motivation a bit more, to help me understand.

chrisdavidmills (approved):
LGTM

PR #41945, opened 2025-11-14
SanitizerConfig - define valid config

### LINKED ISSUE #41634: setHTML() - non-normalized SanitizerConfig explanation is incomplete
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTML

Exceptions

> non-normalized [SanitizerConfig](https://developer.mozilla.org/en-US/docs/Web/API/SanitizerConfig) (one that includes both "allowed" and "removed" configuration settings).

This makes it sound like to me that a non-normalized config is just that specific case, but there are many other non-valid configurations, such as those with duplicate attributes/elements etc.

Probably better to put this information onto some other page. https://wicg.github.io/sanitizer-api/#sanitizerconfig-valid is complicated and probably deserves it own section somewhere like the Sanitizer constructor.

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/api/element/sethtml`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTML
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/element/sethtml/index.md
* Last commit: https://github.com/mdn/content/commit/3e543cdfe8dddfb4774a64bf3decdcbab42a4111
* Document last modified: 2025-06-23T16:41:39.000Z

</details>

hamishwillee:
Fixes #41634

Defines what a valid SanitizerConfig is more precisely and links to this from the places where an invalid config can throw a TypeError.

Related docs work can be tracked in #41649

### DIFF
--- a/files/en-us/web/api/sanitizer/removeelement/index.md
+++ b/files/en-us/web/api/sanitizer/removeelement/index.md
@@ -14,7 +14,7 @@ The **`removeElement()`** method of the {{domxref("Sanitizer")}} interface sets 
 
 The method can be used with either an [allow configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#allow_configurations) or a [remove configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#remove_configurations).
 If used with a remove configuration, the specified element is added to the `removeElements` array.
-If used with an allow configuration, the attribute is removed from the `elements` array (if present).
+If used with an allow configuration, the element is removed from the `elements` array (if present).
 
 ## Syntax
 
@@ -38,7 +38,7 @@ removeElement(element)
 
 Note that `false` might be returned if the internal configuration:
 
-- defines a [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) array array and the element is already omitted (it does not need to be removed)
+- defines an [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) array and the element is already omitted (it does not need to be removed)
 - instead defines the [`removeElements`](/en-US/docs/Web/API/SanitizerConfig#removeelements) array and the specified element is already present (and is hence already filtered)
 
 ## Examples


files/en-us/web/api/sanitizerconfig/index.md
```diff
@@ -94,7 +86,36 @@ Note that normally a {{domxref("Sanitizer")}} instance would be passed as the op
...
+- [`Document.parseHTML()`](/en-US/docs/Web/API/Document/parseHTML_static) or [`Document.parseHTMLUnsafe()`](/en-US/docs/Web/API/Document/parseHTMLUnsafe_static) static methods.
+
+### Valid configuration
+
+The configuration object structure allows for the declaration of filter options that are contradictory or redundant, such as specifying an element in both allow and remove lists, or listing an attribute in a list multiple times.
+In order to avoid any ambiguity, methods that take a `SanitizerConfig` instance require that a _valid_ configuration object be passed, and will throw a {{jsxref("TypeError")}} if an invalid configuration is used.
+
+In a valid sanitizer configuration:
+
+- Either the `elements` or `removeElements` array may be defined, but not both
```
hamishwillee: @evilpie This section attempts to fix the issue you raised in https://github.com/mdn/content/issues/41634 by providing more detail on what "counts" as a valid configuration.

1. I have put this in SanitizerConfig rather than Sanitizer constructor, because an invalid config can be passed in a number of places.
2. This particular point differs from the spec in that the spec says _one of the arrays must be defined_ but not both, and I've said either may be define but not both.
   Testing shows that you don't have to pass a config that has `elements` or `removeElements`. My perspective is developers wanting to know what they can pass - and the spec looks like it is written from after some "work" has been done by the browser.
   
Can you please sanity check that this makes sense?

evilpie: You are right that in the specification, the "valid" check happens after canonicalization, which always adds e.g. `removeElements` when both are missing. I think from a documentation perspective your approach makes sense.

(Maybe we could add some note about valid being after canonicalization, to also make the documentation technically correct?)

files/en-us/web/api/sanitizerconfig/index.md
```diff
@@ -94,7 +86,36 @@ Note that normally a {{domxref("Sanitizer")}} instance would be passed as the op
...
+- Either the `attributes` or `removeAttributes` array may be defined, but not both
+- Within an element, either the `attributes` or `removeAttributes` array may be defined, but not both
```
evilpie: An element can have both `attributes` and `removeAttributes`.

files/en-us/web/api/sanitizerconfig/index.md
```diff
@@ -94,7 +86,36 @@ Note that normally a {{domxref("Sanitizer")}} instance would be passed as the op
...
+- No array may contain duplicate elements or attributes
+- The `replaceWithChildrenElements` array, if defined, may not have any elements in common with `elements` or `removeElements`
+- A global attribute, defined in `attributes`, may not also be defined in an element's `attribute` or `removeAttribute` list.
+- Custom `data-*` attributes may only be specified within element attribute arrays: not in the global `attributes` array, and only if `dataAttributes` is `true`.
```
evilpie: This is wrong. If `dataAttributes` is true (which can only happens if you have a global `attributes`), neither global nor per-element `attributes` can have data-* attributes. The per-element `removeAttributes` array can have data-* attributes (so that you can disallow specific data attributes again).

files/en-us/web/api/sanitizerconfig/index.md
evilpie: Oh, but not when you have a global `removeAttributes`, then only per-element `attributes` or `removeAttributes` is allowed. https://github.com/WICG/sanitizer-api/issues/305

files/en-us/web/api/sanitizerconfig/index.md
```diff
@@ -94,7 +86,36 @@ Note that normally a {{domxref("Sanitizer")}} instance would be passed as the op
...
+In order to avoid any ambiguity, methods that take a `SanitizerConfig` instance require that a _valid_ configuration object be passed, and will throw a {{jsxref("TypeError")}} if an invalid configuration is used.
+
+In a valid sanitizer configuration:
+
+- Either the `elements` or `removeElements` array may be defined, but not both
+- Either the `attributes` or `removeAttributes` array may be defined, but not both
+- Within an element, either the `attributes` or `removeAttributes` array may be defined, but not both
+- No array may contain duplicate elements or attributes
+- The `replaceWithChildrenElements` array, if defined, may not have any elements in common with `elements` or `removeElements`
+- A global attribute, defined in `attributes`, may not also be defined in an element's `attribute` or `removeAttribute` list.
```
evilpie: This isn't entirely correct, the condition is more complicated. 

One mildly interesting use case to demonstrate this:
```{
  attributes: ["class"],
  elements: [{ name: "div", removeAttributes: ["class"] }, "span"]
}
```
Which allows the `class` attribute for all elements, but not `<div>`.


- The intersection of `config["attributes"]` and `element["attributes"]` is empty.
- `element["removeAttributes"]` is a subset of `config["attributes"]`.
- The intersection of `config["removeAttributes"]` and `element["attributes"]` is empty.
- The intersection of `config["removeAttributes"]` and `element["removeAttributes"]` empty.


(Copying from https://wicg.github.io/sanitizer-api/#sanitizerconfig-valid)

files/en-us/web/api/sanitizerconfig/index.md
evilpie: Actually I am not sure what "Custom `data-*` attributes may only be specified within element attribute arrays" means either.

You can totally write: `new Sanitizer({ attributes: ["data-foo"], dataAttributes: false })`.

evilpie: If find the best way to think of `dataAttributes: true` is to imagine there being an implicit `attributes: ["data-*"]`.

hamishwillee: Ah, I get it now. Thank you. `dataAttributes: true` means that data attributes are accepted without having to specify them in the `attributes` list, while `false` means that if you want to allow the attributes you'll have to explicitly list them.

files/en-us/web/api/sanitizerconfig/index.md
hamishwillee: Done below in 

> [!NOTE]
> The conditions above are from the perspective of a web developer.
> The [validity check defined in the specification](https://wicg.github.io/sanitizer-api/#sanitizerconfig-valid) is slightly different because it is executed after canonicalization of the configuration, such as adding `removeElements` when both are missing, and adding default namespaces.

files/en-us/web/api/sanitizerconfig/index.md
```diff
@@ -94,7 +86,46 @@ Note that normally a {{domxref("Sanitizer")}} instance would be passed as the op
...
+- If the global `removeAttributes` array is defined:
+  - An element may specify either `attributes` or `removeAttributes`, but not both
+  - An element's `attributes` or `removeAttributes` array, depending on which (if either) is defined, must not share any values in common with the global `removeAttributes` array.
+  - The global `dataAttributes` array must not be defined.
+
+The empty object `{}` is a valid configuration.
+
+> [!NOTE]
+> The conditions above are from the perspective of a web developer.
+> The [validity check defined in the specification](https://wicg.github.io/sanitizer-api/#sanitizerconfig-valid) is slightly different because it is executed after canonicalization of the configuration, such as adding `removeElements` when both are missing, and adding default namespaces.
```
hamishwillee, line 132: @evilpie Thanks very much - extremely helpful. This is now updated following the spec pattern of splitting the element attribute behaviour based on whether the global `attributes` or `removeAttributes` are defined.
There is a slight reduction in repetition, and of course `{}` is valid in this context.

Following reading this, it seems there have been other changes I haven't track, such as most of the sanitizer methods now returning booleans. I've updated those too. You can ignore those changes if you want - this one is the complicated bit.

files/en-us/web/api/sanitizer/allowelement/index.md
```diff
@@ -11,13 +11,9 @@ browser-compat: api.Sanitizer.allowElement
 {{APIRef("HTML Sanitizer API")}}{{SeeCompatTable}}
 
 The **`allowElement()`** method of the {{domxref("Sanitizer")}} interface sets that the specified element is allowed in the output when the sanitizer is used.
-The element can be specified with lists of attributes that are allowed or disallowed on elements of that type.
 
-The specified element is added to the [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) list in this sanitizer's configuration.
-If the element is already present in the list, then the existing entry is first removed and the new definition is appended to the end of the list.
```
hamishwillee, line 17: Note, for all these methods I have removed comments about what happens under the hood.

A spec author needs to know at this level that the internal configuration might specify either an allow or remove list, and what the update process is different in each case.
A website developer does not - all they need to know at this level is that specifying an element will ensure that it is allowed.

files/en-us/web/api/sanitizer/allowelement/index.md
```diff
@@ -54,7 +50,12 @@ allowElement(element)
 
 ### Return value
 
-None (`undefined`).
+`true` if the operation changed the configuration to allow the element, and `false` if the configuration already allowed the element.
+
+Note that `false` might be returned if the internal configuration:
+
+- defines an [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) array array and the element is already present (it does not need to be added again)
+- instead defines the [`removeElements`](/en-US/docs/Web/API/SanitizerConfig#removeelements) array and the specified element is not present (and is hence already not filtered)
```
hamishwillee: Note that since I last updated this the return type became a boolean. Note that this indicates whether the configuration was changed by the operation - not whether the change worked - it always works.

files/en-us/web/api/sanitizer/allowelement/index.md
```diff
@@ -11,13 +11,9 @@ browser-compat: api.Sanitizer.allowElement
...
 
 The **`allowElement()`** method of the {{domxref("Sanitizer")}} interface sets that the specified element is allowed in the output when the sanitizer is used.
-The element can be specified with lists of attributes that are allowed or disallowed on elements of that type.
 
-The specified element is added to the [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) list in this sanitizer's configuration.
-If the element is already present in the list, then the existing entry is first removed and the new definition is appended to the end of the list.
-Note that if you need both per-element add-attribute and remove-attribute lists, they must be added in a single call to this method (since if done in two calls, the second call will replace the element definition added in the first call).
-
-The specified element is removed from the sanitizer configuration [`removeElements`](/en-US/docs/Web/API/SanitizerConfig#removeelements) or [`replaceWithChildrenElements`](/en-US/docs/Web/API/SanitizerConfig#replacewithchildrenelements) lists if present.
+The element can be specified with arrays of attributes that are allowed or disallowed on elements of that type.
```
evilpie: While I think I agree with your general idea of not leaking too many internals. There is one specific case here that should be worth calling out.

The following does nothing and should produce a warning:
```
s = new Sanitizer({ removeElements: ["div"] })
s.allowElement({name: "div", attributes: ["class"] })
```

https://wicg.github.io/sanitizer-api/#dom-sanitizer-allowelement:~:text=If%20element%5B%22attributes%22%5D%20exists%20or%20element%5B%22removeAttributes%22%5D%20with%20default%20%C2%AB%20%C2%BB%20is%20not%20empty%3A

This is because you can't have `removeElements` combined with per-element `attributes`/`removeAttributes`.

evilpie:
With the current examples, I think it's not clear that you can have an allow list of elements and a remove list of attributes (and vice versa). It might not really be an interesting use case to call out, though.

files/en-us/web/api/sanitizer/allowelement/index.md
hamishwillee: Thanks @evilpie - this is quite a tricky API. Addressed as part of https://github.com/mdn/content/pull/41945/commits/5aa40c77640d33160198b65a432bd0c06881eee7

I think I missed this "point" because it isn't an invalid configuration (so not in algorithm) - it's an impossible configuration inferred from the fact that if you have `removeElements` you can't have `elements`. 

To address this I added a note to the relevant point in [Valid configuration](https://pr41945.review.mdn.allizom.net/en-US/docs/Web/API/SanitizerConfig#valid_configuration) and in [`allowElement()`](https://pr41945.review.mdn.allizom.net/en-US/docs/Web/API/Sanitizer/allowElement) covered this in the return value (the warning) and expanded out the description.

I have also tried to capture what the [allow and remove configuration means](https://pr41945.review.mdn.allizom.net/en-US/docs/Web/API/HTML_Sanitizer_API#allow_and_remove_configurations) in the overview and that the type doesn't change after construction - see [`Sanitizer()` constructor](https://pr41945.review.mdn.allizom.net/en-US/docs/Web/API/Sanitizer/Sanitizer#allow_and_remove_configuration_is_persistent). 

Anyway, I think we're getting there.

1. One thing, is there a recommendation on whether you _should_ use an allow or remove configuration. My thinking would be that an `allow` configuration would almost always be better, in particular if you start from the default Sanitizer.

   My reasoning is that it is generally more restrictive and easier to make restrictive. With remove lists if other elements are later identified as risks, you don't need to update the sanitizer.
   In addition, it allows this case of per-element attributes, which I think is likely be to useful. 

   Not sure where remove configurations would be better. Perhaps where you  are worried about a very particular subset of elements/attributes?


> With the current examples, I think it's not clear that you can have an allow list of elements and a remove list of attributes (and vice versa). It might not really be an interesting use case to call out, though.

W.r.t. https://github.com/mdn/content/pull/41945#issuecomment-3540924623 This is mentioned now. I don't know how interesting it is.

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -58,25 +58,40 @@ someElement.setHTMLUnsafe(untrustedString, { sanitizer });
...
-You can build up a configuration in two ways: allow configurations and remove configurations.
+You can build up a configuration in two ways: specifying the set of elements and attributes that you will allow in the output, or specifying the set that must not be present in the output.
 
-In "allow configurations" you specify the elements and attributes you wish to _allow_ (or replace with child elements): all other elements/attributes in the input will be dropped.
-For example, the following configuration allows only {{htmlelement("p")}} and {{htmlelement("div")}} elements, and `cite` and `onclick` attributes on any element.
-It will also replace {{htmlelement("b")}} elements with their child nodes, effectively stripping the style their nested content.
+Note that it is not valid to specify both allowed and removed elements in a configuration, or both allowed and removed attributes.
+You can. however. have a configuration that allows elements and removes attributes, or vice versa.
+
+##### Allow configurations
```
hamishwillee, line 110: FYI Expanded this out to 
- provide better allow/remove links and to be able to provide more information, as these provide different opportunities. 
- Not sure if we can provide "which should I use" advice other than what I put in the first paragraph below.
- Wanted to show some cases of using `Sanitizer` as this was quite SantizerConfig centric - and most should prefer `Sanitizer`

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -86,50 +101,72 @@ const sanitizer = new Sanitizer({
...
+The safe HTML sanitization methods don't use trusted types.
 Because they always filter all XSS-unsafe entities before input HTML is injected, there is no need to sanitize the input string, or audit the methods.
 
 However the unsafe HTML sanitization methods may inject untrusted HTML, depending on the sanitizer, and so will work with trusted types.
 The methods can take either a string or a `TrustedType` as input.
 If a sanitizer is also supplied, the transformation function will be run first, and then the sanitizer.
 
-Note that the transformation function doesn't have to sanitize the input string in this case (although it can), because you can use the sanitizer API for that.
-What trusted types provide in this case is information about where potentially unsafe strings are injected, making it easier to locate them and check that the sanitizer is appropriately configured.
+Note that the behavior of the transformation function in this case will depend on the website policy (which might be to reject all use of the unsafe methods).
```
hamishwillee, line 210: Note, trusted type feedback on how trusted types and unsafe methods was not very "precise". What I took from it was that you should always use Trusted Types and that the policy for trusted types is whatever the website says it is. In other words, we shouldn't really offer guidance on to use the transformation, or a sanitizer or whatever.

files/en-us/web/api/html_sanitizer_api/using_the_html_sanitizer_api/index.md
```diff
@@ -1,7 +1,8 @@
----
+gin---
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
---
```

files/en-us/web/api/html_sanitizer_api/using_the_html_sanitizer_api/index.md
```diff
@@ -192,7 +192,8 @@ const sanitizer = new Sanitizer({
 In both cases you can also specify each attribute as an object with `name` and `namespace` properties.
 You can also use set the attribute properties using the same element object passed to {{domxref("Sanitizer.allowElement()")}}.
 
-Note however that you can't specify both element `attributes` and `removeAttributes` in one call. Attempting to do so will raise an exception.
+Note that it is impossible to define per-element attribute behavior on a Sanitizer with a remove configuration, as the (needed) `elements` array is not present.
```
hamishwillee, line 200: FYI another note hat you can't define per-element attributes with a remove configuraiton

files/en-us/web/api/sanitizerconfig/index.md
```diff
@@ -94,13 +86,64 @@ Note that normally a {{domxref("Sanitizer")}} instance would be passed as the op
...
+
+### Allow and remove configurations
+
+One of the main implications of the previous section is that a valid configuration can specify either `elements` or `removeElements` arrays (but not both) and either the `attributes` or `removeAttributes` arrays (but not both).
+
+A configuration that has the `elements` and/or `attributes` arrays is referred to as an [allow configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#allow_configurations), as it defines the sanitization behavior in terms of the values that are allowed to be present in the output.
+A [remove configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#remove_configurations) is one that has either of `removeElements` and/or `removeAttributes`, and defines the behavior in terms of the values that will be removed from the output.
```
hamishwillee, line 139: I thought worth having a "thin" section here on the difference, since this is where you will create the configurations.

files/en-us/web/api/sanitizerconfig/index.md
hamishwillee: Addressed. This section updated to match the structure of the spec.

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -58,25 +58,40 @@ someElement.setHTMLUnsafe(untrustedString, { sanitizer });
...
 
-- {{domxref('SanitizerConfig')}} is a dictionary object that defines arrays of elements or attributes that are allowed or disallowed when the configuration is used, properties that indicate whether comments and data attributes will be allowed or omitted, and so on.
-- {{domxref('Sanitizer')}} is essentially a wrapper around a {{domxref('SanitizerConfig')}} that provides methods to ergonomically and consistently add and remove entities from the various lists in configuration.
-  For example, you can use a method to add an allowed attribute, and it will also remove the attribute from the disallowed array (if present).
+- {{domxref('SanitizerConfig')}} is a dictionary object that defines arrays for the allowed/disallowed elements and attributes and boolean properties that indicate whether comments and data attributes will be allowed or omitted, and so on.
+
+  Only a subset of possible configuration options may be specified in a particular configuration in order to reduce redundancy and ambiguity.
+  The allowed subset is described in [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration).
```
wbamberg:
```suggestion
  The allowed subset is described in [Valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration).
```

(grammatically this is quoting the section title. If you'd said "in our section on valid configurations" it would be lowercase)

files/en-us/web/api/html_sanitizer_api/index.md
wbamberg: It's probably worth saying that allow is safer than remove, because with remove you've got to be sure you have listed everything that might be dangerous.

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -58,25 +58,40 @@ someElement.setHTMLUnsafe(untrustedString, { sanitizer });
...
 
-You can build up a configuration in two ways: allow configurations and remove configurations.
+You can build up a configuration in two ways: specifying the set of elements and attributes that you will allow in the output, or specifying the set that must not be present in the output.
 
-In "allow configurations" you specify the elements and attributes you wish to _allow_ (or replace with child elements): all other elements/attributes in the input will be dropped.
-For example, the following configuration allows only {{htmlelement("p")}} and {{htmlelement("div")}} elements, and `cite` and `onclick` attributes on any element.
-It will also replace {{htmlelement("b")}} elements with their child nodes, effectively stripping the style their nested content.
+Note that it is not valid to specify both allowed and removed elements in a configuration, or both allowed and removed attributes.
+You can. however. have a configuration that allows elements and removes attributes, or vice versa.
+
```
wbamberg: What if you have an `elements` array, and one or more of the items lists [allowed attributes](https://pr41945.review.mdn.allizom.net/en-US/docs/Web/API/SanitizerConfig#attributes), and you also have `removeAttributes`? Is this allowed and if so who wins in case of conflict?

In general this section doesn't mention the `attributes` and `removeAttributes` properties of `elements`, maybe it should?

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -58,25 +58,40 @@ someElement.setHTMLUnsafe(untrustedString, { sanitizer });
...
+##### Allow configurations
+
+With "allow configurations" you specify the elements and attributes you wish to _allow_ (or replace with child elements) — all other elements/attributes in the input will be dropped.
+This makes it easy to understand what elements will be allowed in the DOM when the HTML is parsed.
+They are useful when you know exactly what HTML entities you want to be able to inject in a particular context.
+
+Allow configurations are created by defining a `Sanitizer` that wraps a {{domxref("SanitizerConfig")}} that includes the [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) and/or [`attributes`](/en-US/docs/Web/API/SanitizerConfig#attributes) arrays (and not the `removeElements` or `removeAttributes` arrays).
+
+For example, the following configuration is created by passing a {{domxref('SanitizerConfig')}} that allows {{htmlelement("p")}} and {{htmlelement("div")}} elements, and `cite` and `onclick` attributes on any element.
```
wbamberg:
```suggestion
For example, the following configuration is created by passing a {{domxref('SanitizerConfig')}} that allows {{htmlelement("p")}} and {{htmlelement("div")}} elements, and `cite` and `onclick` attributes on any allowed element.
```

(I guess?)

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -58,25 +58,40 @@ someElement.setHTMLUnsafe(untrustedString, { sanitizer });
...
+It will also replace {{htmlelement("b")}} elements with their child nodes, stripping the style from the element's content.
```
wbamberg: HTML purists will hate you for calling `<b>` a style. I think you could just omit the second clause here.

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -58,25 +58,40 @@ someElement.setHTMLUnsafe(untrustedString, { sanitizer });
...
 #### Allow and remove configurations
 
-You can build up a configuration in two ways: allow configurations and remove configurations.
+You can build up a configuration in two ways: specifying the set of elements and attributes that you will allow in the output, or specifying the set that must not be present in the output.
 
-In "allow configurations" you specify the elements and attributes you wish to _allow_ (or replace with child elements): all other elements/attributes in the input will be dropped.
-For example, the following configuration allows only {{htmlelement("p")}} and {{htmlelement("div")}} elements, and `cite` and `onclick` attributes on any element.
-It will also replace {{htmlelement("b")}} elements with their child nodes, effectively stripping the style their nested content.
+Note that it is not valid to specify both allowed and removed elements in a configuration, or both allowed and removed attributes.
+You can. however. have a configuration that allows elements and removes attributes, or vice versa.
```
wbamberg:
```suggestion
You can, however, have a configuration that allows elements and removes attributes, or vice versa.
```

wbamberg: It's awkward to express this precisely. Maybe consider a table?

| Allow | Remove | Valid? |
| --------- | --------- | --------- |
| `elements`    | -    | ✓    |
| `elements`    | `removeAttributes`    | ✓    |
| `elements`    | `removeElements`    | ❌    |
| `attributes`    | -    | ✓    |
| `attributes`    | `removeElements`    | ✓    |
| `attributes`    | `removeAttributes`    | ❌    |
| -    | `removeElements`    | ✓    |
| -    | `removeAttributes`    | ✓    |

files/en-us/web/api/html_sanitizer_api/using_the_html_sanitizer_api/index.md
```diff
@@ -28,7 +28,7 @@ console.log(someElement.innerHTML); // abc def
 
 The other XSS-safe methods, {{domxref('ShadowRoot.setHTML()')}} and {{domxref('Document/parseHTML_static','Document.parseHTML()')}}, are used in the same way.
 
-### Safe methods further restrict allowed entities
+### Safe methods can further restrict allowed entities
```
wbamberg: I don't exactly know why I think this, but it's grating to me when headings tell me things.

```suggestion
### Using safe methods with a sanitizer
```

I also think, if this "Safe methods can further restrict allowed entities" is the main point of this section, that it should be introduced at the start of it, not at the end. As it is, I read the heading text, then I'm puzzled that the content doesn't seem to be about "further restriction" until I get to the last paragraph. i.e. something like:

"If you pass a sanitizer to a safe method, then any restrictions imposed by the sanitizer are applied _on top of_ the restrictions already applied by default by the safe method. That is, passing a sanitizer to a safe method can further restrict the allowed entities, but can never loosen restrictions."

I also think the first sentence here:

> You can specify the HTML entities that you want to allow or remove by passing a [Sanitizer](https://pr41945.review.mdn.allizom.net/en-US/docs/Web/API/Sanitizer) in the second argument of all the sanitizer methods.

...is out of place in a section about how safe methods further restrict allowed entries. It would be better to have a section before this, that introduces the concept of a sanitizer.

files/en-us/web/api/html_sanitizer_api/using_the_html_sanitizer_api/index.md
```diff
@@ -192,7 +192,8 @@ const sanitizer = new Sanitizer({
 In both cases you can also specify each attribute as an object with `name` and `namespace` properties.
 You can also use set the attribute properties using the same element object passed to {{domxref("Sanitizer.allowElement()")}}.
 
-Note however that you can't specify both element `attributes` and `removeAttributes` in one call. Attempting to do so will raise an exception.
+Note that it is impossible to define per-element attribute behavior on a Sanitizer with a remove configuration, as the (needed) `elements` array is not present.
+Other restrictions on per-element attributes are covered in [valid configurations](/en-US/docs/Web/API/SanitizerConfig#valid_configuration)
```
wbamberg:
```suggestion
Other restrictions on per-element attributes are covered in [Valid configurations](/en-US/docs/Web/API/SanitizerConfig#valid_configuration)
```

files/en-us/web/api/html_sanitizer_api/using_the_html_sanitizer_api/index.md
```diff
@@ -296,9 +297,9 @@ sanitizer.removeAttribute("lang");
 
 ## Comments and data attributes
 
-The {{domxref("SanitizerConfig")}} can also be used to specify whether comments and `data-` attributes will be filtered from injected content, using the [comments](/en-US/docs/Web/API/SanitizerConfig#comments) and [dataAttributes](/en-US/docs/Web/API/SanitizerConfig#dataattributes) boolean properties, respectively.
+The {{domxref("SanitizerConfig")}} can also be used to specify whether comments will be filtered from injected content, using the [comments](/en-US/docs/Web/API/SanitizerConfig#comments) property, and whether `data-` attributes are allowed without having to add them to the `attributes` array using the [dataAttributes](/en-US/docs/Web/API/SanitizerConfig#dataattributes) boolean property.
```
wbamberg:
```suggestion
The {{domxref("SanitizerConfig")}} can also be used to specify whether comments will be filtered from injected content, using the [`comments`](/en-US/docs/Web/API/SanitizerConfig#comments) property, and whether `data-` attributes are allowed without having to add them to the `attributes` array using the [`dataAttributes`](/en-US/docs/Web/API/SanitizerConfig#dataattributes) boolean property.
```

files/en-us/web/api/html_sanitizer_api/using_the_html_sanitizer_api/index.md
```diff
@@ -307,7 +308,7 @@ const sanitizer = new Sanitizer({
 });
 ```
 
-You can similarly enable or disable the comments or data-attributes on an existing sanitizer using {{domxref("Sanitizer.setComments()")}} and {{domxref("Sanitizer.setDataAttributes()")}} methods:
+You can similarly set the comments or data-attributes properties on an existing sanitizer using {{domxref("Sanitizer.setComments()")}} and {{domxref("Sanitizer.setDataAttributes()")}} methods:
```
wbamberg:
```suggestion
You can similarly set the comments or data-attributes properties on an existing sanitizer using {{domxref("Sanitizer.setComments()")}} and {{domxref("Sanitizer.setDataAttributes()")}}:
```

files/en-us/web/api/html_sanitizer_api/using_the_html_sanitizer_api/index.md
```diff
@@ -319,15 +320,19 @@ sanitizer.setDataAttributes(true);
 
 All the sanitization methods can be passed a sanitizer configuration that is either a {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} instance.
 
+The {{domxref("SanitizerConfig")}} provides a compact method for specifying multiple elements or attributes that should be allowed or removed at the same time.
+Some care may be required when modifying this object to ensure that it remains a [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration).
```
wbamberg, line 329: not used as a title here

files/en-us/web/api/sanitizer/allowelement/index.md
```diff
@@ -54,7 +49,58 @@ allowElement(element)
...
+```
+
+If you need both per-element add-attribute and remove-attribute arrays as shown above, they must be added in a single call to this method.
+If you were to do this in two calls the the second call would replace the element definition added in the first call.
+
+When using a `Sanitizer` with a [remove configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#remove_configurations), similar code to add per-element attribute allow or remove arrays will be ignored and generate a console warning.
+This is because internally the sanitizer doesn't have the `elements` array required to specify per-element attributes.
+For example, the following code won't allow the `<div>` element (at all):
+
+```js
```
wbamberg:
```suggestion
```js example-bad
```

files/en-us/web/api/sanitizer/allowelement/index.md
```diff
@@ -54,7 +49,58 @@ allowElement(element)
...
 
-None (`undefined`).
+`true` if the operation changed the configuration to allow the element, and `false` if the configuration was not changed (usually because the element was already allowed, but potentially because the change could not be made).
+
+Note that `false` might be returned if the internal configuration:
+
+- defines the [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) array and the element is already present (it does not need to be added again)
+- defines the [`removeElements`](/en-US/docs/Web/API/SanitizerConfig#removeelements) array and the specified element is not present (and is hence already not filtered).
+- defines the [`removeElements`](/en-US/docs/Web/API/SanitizerConfig#removeelements) array and attempts to allow an element with per-element attributes.
+  This operation is not supported because in a [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration) you can't have both `removeElements` and `elements` arrays, and per-element attributes are added in the `elements` array.
```
wbamberg, line 59: Is it not simpler than this?

```suggestion
- defines the [`removeElements`](/en-US/docs/Web/API/SanitizerConfig#removeelements) array.
  This operation is not supported because in a [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration) you can't have both `removeElements` and `elements` arrays, and allowed elements are added in the `elements` array.
```

i.e. your `allowElement()` doesn't need to try to change attributes for it to fail in this case. The description below also gives the impression that it is setting attributes in the call that is not allowed, when ISTM that just making the call at all is not allowed.

files/en-us/web/api/sanitizer/sanitizer/index.md
```diff
@@ -39,7 +31,27 @@ An instance of the {{domxref("Sanitizer")}} object.
 ### Exceptions
 
 - {{jsxref("TypeError")}}
-  - : Thrown if a non-normalized {{domxref("SanitizerConfig")}} is passed (one that includes both "allowed" and "removed" configuration settings), or if a string is passed that isn't `"default"`.
+  - : The `configuration` parameter is passed a:
```
wbamberg: super nit, but it will read better with the article "a" duplicated in the bullets, and maybe even better with "passed one of the following"

files/en-us/web/api/sanitizer/sanitizer/index.md
```diff
@@ -39,7 +31,27 @@ An instance of the {{domxref("Sanitizer")}} object.
...
+
+The default `Sanitizer` allows only XSS-safe input by default, omitting elements such as {{HTMLElement("script")}}, {{HTMLElement("frame")}}, {{HTMLElement("iframe")}}, {{HTMLElement("object")}}, `<use>`, and event handler attributes from their respective allow lists, and disallowing data attributes, and comments.
+It is created if `"default"` or no object is passed to the constructor.
+
+The constructor can be passed a {{domxref("SanitizerConfig")}} with a [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration) to customize the sanitizer behavior.
+
+### Allow and remove configuration is persistent
+
+A valid configuration can specify either `elements` or `removeElements` arrays (but not both) and either the `attributes` or `removeAttributes` arrays (but not both).
+The arrays that are defined when the object is constructed persist for its lifetime, and their contents may be modified by the `Sanitizer` methods.
```
wbamberg: "persist for its lifetime" is weird. What else might happen to them?

files/en-us/web/api/sanitizer/setdataattributes/index.md
```diff
@@ -10,9 +10,9 @@ browser-compat: api.Sanitizer.setDataAttributes
 
 {{APIRef("HTML Sanitizer API")}}{{SeeCompatTable}}
 
-The **`setDataAttributes()`** method of the {{domxref("Sanitizer")}} interface sets whether [data attributes](/en-US/docs/Web/HTML/Reference/Global_attributes/data-*) will be allowed or removed by the sanitizer.
+The **`setDataAttributes()`** method of the {{domxref("Sanitizer")}} interface sets whether all [`data-*` attributes](/en-US/docs/Web/HTML/Reference/Global_attributes/data-*) will be allowed by the sanitizer, or if they must be individually specified.
 
-The method sets the [`dataAttributes`](/en-US/docs/Web/API/SanitizerConfig#dataattributes) property in this sanitizer's configuration.
+If this is set `true`, then data attributes are automatically allowed without needing to add them individually using {{domxref('Sanitizer.allowAttribute()')}} (or {{domxref('Sanitizer.allowElement()')}} for local attributes).
```
wbamberg: What if it's true but you add some data attributes to `removeAttributes`?

files/en-us/web/api/html_sanitizer_api/index.md
hamishwillee:
```suggestion
It will also replace {{htmlelement("b")}} elements with their child nodes.
```

files/en-us/web/api/sanitizer/sanitizer/index.md
```diff
@@ -39,7 +31,27 @@ An instance of the {{domxref("Sanitizer")}} object.
 ### Exceptions
 
 - {{jsxref("TypeError")}}
-  - : Thrown if a non-normalized {{domxref("SanitizerConfig")}} is passed (one that includes both "allowed" and "removed" configuration settings), or if a string is passed that isn't `"default"`.
+  - : The `configuration` parameter is passed a:
+    - {{domxref("SanitizerConfig")}} that isn't a valid configuration.
+      For example, a configuration that includes both "allowed" and "removed" configuration settings.
+    - String that does not have the value `"default"`.
```
hamishwillee:
```suggestion
  - : The `configuration` parameter is passed one of the following:
    - a {{domxref("SanitizerConfig")}} that isn't a valid configuration.
      For example, a configuration that includes both "allowed" and "removed" configuration settings.
    - a string that does not have the value `"default"`.
```

files/en-us/web/api/sanitizer/sanitizer/index.md
hamishwillee: Done in https://github.com/mdn/content/pull/41945/files#r2558426713

files/en-us/web/api/sanitizer/sanitizer/index.md
```diff
@@ -39,7 +31,27 @@ An instance of the {{domxref("Sanitizer")}} object.
...
+
+The constructor can be passed a {{domxref("SanitizerConfig")}} with a [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration) to customize the sanitizer behavior.
+
+### Allow and remove configuration is persistent
+
+A valid configuration can specify either `elements` or `removeElements` arrays (but not both) and either the `attributes` or `removeAttributes` arrays (but not both).
+The arrays that are defined when the object is constructed persist for its lifetime, and their contents may be modified by the `Sanitizer` methods.
+In most cases which arrays are defined does not particularly matter because, for example, the {{domxref("Sanitizer/allowAttribute","allowAttribute()")}} method can implement the same behavior by adding the attribute to the `attributes` array or by removing it from the `removeAttributes` array.
+
+The main case to note is that if you have a remove configuration then you cannot have per-element attributes, as these must be defined on the `elements` array.
```
hamishwillee:
```suggestion
The constructor can be passed a {{domxref("SanitizerConfig")}} with a [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration) to customize the sanitizer behavior.

A valid configuration can specify either `elements` or `removeElements` arrays (but not both) and either the `attributes` or `removeAttributes` arrays (but not both).
In most cases it does not matter which of these arrays you use because, for example, the {{domxref("Sanitizer/allowAttribute","allowAttribute()")}} method can implement the same behavior by adding the attribute to the `attributes` array or by removing it from the `removeAttributes` array.
The main thing to note is that if you have a configuration with `removeElements` then you cannot have per-element attributes, as these must be defined on the `elements` array.
```

files/en-us/web/api/sanitizer/sanitizer/index.md
hamishwillee: They could in theory change type over their lifetime because almost any allow configuration can be specified as a remove configuration.

I thought it reasonable to mention because that is what I thought the `Sanitizer` object did when I originally wrote the docs, and that lead me to a number of false impressions.
However given no one else is likely to have my preconceptions, I am replacing this whole section with https://github.com/mdn/content/pull/41945/files#r2558439002

files/en-us/web/api/html_sanitizer_api/index.md
hamishwillee: Yes. Was waiting on Tom for confirmation, but perhaps this is obvious.

Fixed by last sentence in this update to previous section https://github.com/mdn/content/pull/41945/commits/8912664aba29d846b725a38a0ada10ead27002c7

files/en-us/web/api/html_sanitizer_api/index.md
hamishwillee: Fixed in update https://github.com/mdn/content/pull/41945/commits/8912664aba29d846b725a38a0ada10ead27002c7

I used a similar table and a few words.

files/en-us/web/api/html_sanitizer_api/index.md
hamishwillee: It depends on the whether the global attributes is an allow or remove configuration - if you see the valid configuration it is one of the two nested bullet paths.

Yes I should note. Addressed as a comment in https://github.com/mdn/content/pull/41945/commits/8912664aba29d846b725a38a0ada10ead27002c7 - mostly saying "see the config"

files/en-us/web/api/html_sanitizer_api/using_the_html_sanitizer_api/index.md
hamishwillee: OK, I have tidied this structure in line with your suggestion here https://github.com/mdn/content/pull/41945/commits/c3bd19c4aeb612c117dd5ffad64ca1a575f6782c

files/en-us/web/api/sanitizer/allowelement/index.md
```diff
@@ -54,7 +49,58 @@ allowElement(element)
 
 ### Return value
 
-None (`undefined`).
+`true` if the operation changed the configuration to allow the element, and `false` if the configuration was not changed (usually because the element was already allowed, but potentially because the change could not be made).
+
+Note that `false` might be returned if the internal configuration:
+
+- defines the [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) array and the element is already present (it does not need to be added again)
```
hamishwillee:
```suggestion
- defines the [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) array and the element is already present (it does not need to be added again).
```

files/en-us/web/api/sanitizer/allowelement/index.md
hamishwillee: Note True or false does not indicate failure - it's a measure of whether the configuration was modified.   Don't ask me why that is useful - you do know that if the result is `true` it worked, but `false` does not mean that the configuration does not now allow the change.

I think you are saying "isn't defining `removeElements` "enough to return false when allowing an element?

No. It will return true if it changes the configuration - so if you allow an element and `removeElements` doesn't exists
- if the element is already in `removeElements` that element will be removed, and the method will return true. 
- if the element is not in  `removeElements`  it will return false
- if the element is in [replaceWithChildrenElements](https://wicg.github.io/sanitizer-api/#dom-sanitizerconfig-replacewithchildrenelements) it will be removed from there, which would return also return `true` indicating a change.
- if  you try to add an element with attributes this is "impossible" - so you get the error - this is true whether or not the element is already present in `removeElements`. So you might get false for a different reason. 


Fixed up the description to be more clear in https://github.com/mdn/content/pull/41945/commits/1a42cf6f25298dad3a904d82ea0fc0106b6b40bb

files/en-us/web/api/sanitizer/setdataattributes/index.md
hamishwillee: If you already have a `removeAttributes` then this will return `false`. It's invalid to have both `removeAttributes` and dataAttributes so this can't happen.

files/en-us/web/api/sanitizer/setdataattributes/index.md
```diff
@@ -23,26 +23,27 @@ setDataAttributes(allow);
 ### Parameters
 
 - `allow`
-  - : `true` if data attributes are allowed, and `false` if they are to be removed.
+  - : `true` if all `data-*` attributes are allowed, and `false` if they must be explicitly specified.
 
 ### Return value
 
-None (`undefined`).
+`true` if the operation changed the configuration, and `false` if the configuration already set [`dataAttributes`](/en-US/docs/Web/API/SanitizerConfig#dataattributes) to the specified value.
```
hamishwillee:
```suggestion
`true` if the operation changed the configuration, and `false` if the configuration already set [`dataAttributes`](/en-US/docs/Web/API/SanitizerConfig#dataattributes) to the specified value or `dataAttributes` cannot be set `true`.
```

files/en-us/web/api/sanitizer/setdataattributes/index.md
hamishwillee: And you can't actually add them either way, because it is invalid.

I did update https://github.com/mdn/content/pull/41945/files#r2558821461

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -58,25 +58,62 @@ someElement.setHTMLUnsafe(untrustedString, { sanitizer });
...
+| `elements`                        | -                  | ✔️     |
+| `elements`                        | `attributes`       | ✔️     |
+| `elements`                        | `removeAttributes` | ✔️     |
+| `removeElements`                  | -                  | ✔️     |
+| `removeElements`                  | `attributes`       | ✔️     |
+| `removeElements`                  | `removeAttributes` | ✔️     |
+| -                                 | `attributes`       | ✔️     |
+| -                                 | `removeAttributes` | ✔️     |
+| `elements` + `removeElements`     | (anything)         | ❌     |
+| `attributes` + `removeAttributes` | (anything)         | ❌     |
```
wbamberg:
```suggestion
| (anything) | `attributes` + `removeAttributes`         | ❌     |
```

(I hope you think this table is helpful!)

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -58,25 +58,62 @@ someElement.setHTMLUnsafe(untrustedString, { sanitizer });
...
+  For example, you can use a method to add an allowed element, and it will also remove the element from the `replaceWithChildrenElements` array (if present).
   The interface also provides methods to return a copy of the underlying {{domxref('SanitizerConfig')}} and also to update the sanitizer so that it is XSS-safe.
   It may provide normalizations of the sanitizer configuration used to construct it, making it easier to understand and reuse.
 
 While you can use either interface in any of the sanitizing methods, `Sanitizer` is likely to be more efficient to share and reuse than `SanitizerConfig`.
 
 #### Allow and remove configurations
 
-You can build up a configuration in two ways: allow configurations and remove configurations.
+You can build up a configuration in two ways: specifying the set of elements and/or attributes that you will allow in the output, or specifying the set that must not be present in the output.
```
wbamberg:
```suggestion
You can build up a configuration in two ways:

- As an _allow configuration_: specifying the set of elements and/or attributes that you will allow in the output.
- As a _remove configuration_: specifying the set that must not be present in the output.
```

Just a suggestion. But I think if we are going to use specific terminology it's always good to very explicitly define it.

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -58,25 +58,62 @@ someElement.setHTMLUnsafe(untrustedString, { sanitizer });
...
+| `removeElements`                  | `removeAttributes` | ✔️     |
+| -                                 | `attributes`       | ✔️     |
+| -                                 | `removeAttributes` | ✔️     |
+| `elements` + `removeElements`     | (anything)         | ❌     |
+| `attributes` + `removeAttributes` | (anything)         | ❌     |
+| -                                 | -                  | ✔️     |
+
+An allow configuration can optionally specify whether per-element attributes should be allowed and/or removed in its `elements` array.
+The allowed configuration for these local attributes depends on whether or not global `attributes` or `removedAttributes` is defined.
+The [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration) outlines the restrictions.
```
wbamberg:
```suggestion
The [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration) section outlines the restrictions.
```

files/en-us/web/api/sanitizer/allowelement/index.md
wbamberg: I get it now - I had thought that `allowElement()` was only for allow configurations, i.e. it added the element to `elements`. Probably because it's called "allow".

Is it worth being explicit and saying something like "In an allow configuration, this will add the element to the configuration's `elements` array. In a remove configuration, it will remove the element from the configuration's `removeElements` array."

Or is this just me?

wbamberg (commented):
Thanks for the updates!

files/en-us/web/api/sanitizer/allowelement/index.md
hamishwillee: Probably. We don't know how people are going to reach this page. 

When I get back.

files/en-us/web/api/html_sanitizer_api/index.md
hamishwillee: I do!

files/en-us/web/api/html_sanitizer_api/index.md
hamishwillee: Also added a section that makes it clear that `Sanitizer` allows calling both add and remove methods (and how that works) and what's going on with the errors returned by its methods (i.e. don't rely on them as a means to understand whether the sanitizer will work as intended). See https://github.com/mdn/content/pull/41945/commits/e1c2823b5ffe298f034deb3cbd9a877a293801a8

files/en-us/web/api/sanitizer/allowelement/index.md
hamishwillee: I've started fixing this by adding a concepts section that makes it clear that you CAN call all the methods irrespective of the sanitizer type: https://github.com/mdn/content/pull/41945/commits/e1c2823b5ffe298f034deb3cbd9a877a293801a8

hamishwillee: > Or is this just me?

I have added notes to the methods https://github.com/mdn/content/pull/41945/commits/f5465ee1862fc54c5e4ed5252c6f7f9b34648d11 for this case 

I was trying to avoid this because the different types of configuration are "kind of supposed to be" an implementation detail at the point you're using a method. But then they aren't because the internal behaviour leaks into usage in a number of cases.

Anyway, I don't think this is excessive.

files/en-us/web/api/sanitizer/setdataattributes/index.md
```diff
@@ -10,9 +10,12 @@ browser-compat: api.Sanitizer.setDataAttributes
...
 {{APIRef("HTML Sanitizer API")}}{{SeeCompatTable}}
 
-The **`setDataAttributes()`** method of the {{domxref("Sanitizer")}} interface sets whether [data attributes](/en-US/docs/Web/HTML/Reference/Global_attributes/data-*) will be allowed or removed by the sanitizer.
+The **`setDataAttributes()`** method of the {{domxref("Sanitizer")}} interface sets whether all [`data-*` attributes](/en-US/docs/Web/HTML/Reference/Global_attributes/data-*) will be allowed by the sanitizer, or if they must be individually specified.
 
-The method sets the [`dataAttributes`](/en-US/docs/Web/API/SanitizerConfig#dataattributes) property in this sanitizer's configuration.
+If this is set `true`, then data attributes are automatically allowed and you should not add them individually using {{domxref('Sanitizer.allowAttribute()')}} (or {{domxref('Sanitizer.allowElement()')}} for local attributes).
+
+Note that this method is only useful for [allow configurations](/en-US/docs/Web/API/HTML_Sanitizer_API#allow_configurations), as a remove configuration that doesn't specify any `data-*` attributes already allows all `data-*` attributes.
+If will return `false` for remove configurations.
```
hamishwillee:
```suggestion
Note that this method is useful for [allow configurations](/en-US/docs/Web/API/HTML_Sanitizer_API#allow_configurations) which have a lot of `data-*` attributes that you want to allow.
The method returns `false` when used with remove configurations, which can allow all `data-*` attributes simply by omitting them.
```

hamishwillee:
Thanks for the review @wbamberg - I've tried to address your concerns/confusions, in particular about how the `Sanitizer` methods abstract the underlying allow/remove configuration, so you can use all of its methods. Hope it makes more sense.

Ready for another look.

files/en-us/web/api/sanitizer/removeelement/index.md
```diff
@@ -10,10 +10,11 @@ browser-compat: api.Sanitizer.removeElement
...
 {{APIRef("HTML Sanitizer API")}}{{SeeCompatTable}}
 
-The **`removeElement()`** method of the {{domxref("Sanitizer")}} interface sets an element to be disallowed — removed from input when the sanitizer is used.
+The **`removeElement()`** method of the {{domxref("Sanitizer")}} interface sets the specified element be removed from the output when the sanitizer is used.
 
-The specified element is added to the list of [`removeElements`](/en-US/docs/Web/API/SanitizerConfig#removeelements) in this sanitizer's configuration.
-The element is removed from the [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) or [`replaceWithChildrenElements`](/en-US/docs/Web/API/SanitizerConfig#replacewithchildrenelements) lists if present.
+The method can be used with either an [allow configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#allow_configurations) or a [remove configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#remove_configurations).
+If used with a remove configuration, the specified element is added to the `removeElements` array.
+If used with an allow configuration, the attribute is removed from the `elements` array (if present).
```
wbamberg:
```suggestion
If used with an allow configuration, the element is removed from the `elements` array (if present).
```

files/en-us/web/api/sanitizer/removeelement/index.md
```diff
@@ -33,7 +34,12 @@ removeElement(element)
 
 ### Return value
 
-None (`undefined`).
+`true` if the operation changed the configuration to disallow the element, and `false` if the element was already disallowed.
+
+Note that `false` might be returned if the internal configuration:
+
+- defines a [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) array array and the element is already omitted (it does not need to be removed)
```
wbamberg:
```suggestion
- defines an [`elements`](/en-US/docs/Web/API/SanitizerConfig#elements) array and the element is already omitted (it does not need to be removed)
```

hamishwillee:
> Thanks Hamish! I liked the extra explanations here. I fixed a couple of obvious typos.

Thanks for the reviews and patience - I'd like to say that I injected those typos on purpose, but it seems my brain and keyboard are not connected. Definitely better for the feedback.

PR #41938, opened 2025-11-13
permissions-policy is not forbidden request or response header

hamishwillee:
Fixes #41843

Permissions-Policy is not a forbidden request header.

PR #41946, opened 2025-11-14
HTTP response headers - replace forbidden request

hamishwillee:
As discussed in https://github.com/mdn/content/pull/41938#issuecomment-3530152718 , response headers should not have the section on "forbidden request headers". I have modified this so that response headers listed in https://github.com/mdn/content/pull/41938#issuecomment-3530227774 now have "Forbidden response header" instead.

I also updated the request and response headers to include the forbidden response information.

### DIFF
--- a/files/en-us/web/http/reference/headers/accept-ch/index.md
+++ b/files/en-us/web/http/reference/headers/accept-ch/index.md
@@ -19,10 +19,6 @@ To ensure client hints are sent reliably, the `Accept-CH` header should be persi
       <td>{{Glossary("Response header")}}</td>
     </tr>
     <tr>
-      <th scope="row">{{Glossary("Forbidden response header name","Forbidden response header")}}</th>
-      <td>No</td>
-    </tr>
-    <tr>
       <th scope="row">
         {{Glossary("CORS-safelisted response header")}}
       </th>
--- a/files/en-us/web/http/reference/headers/accept-encoding/index.md
+++ b/files/en-us/web/http/reference/headers/accept-encoding/index.md
@@ -35,10 +35,6 @@ As long as the `identity;q=0` or `*;q=0` directives do not explicitly forbid the
       <th scope="row">{{Glossary("Forbidden request header")}}</th>
       <td>Yes</td>
     </tr>
-    <tr>
-      <th scope="row">{{Glossary("Forbidden response header name","Forbidden response header")}}</th>
-      <td>No</td>
-    </tr>
   </tbody>
 </table>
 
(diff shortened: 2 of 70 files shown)


files/en-us/web/http/reference/headers/set-cookie/index.md
```diff
@@ -28,8 +28,8 @@ For more information, see the guide on [Using HTTP cookies](/en-US/docs/Web/HTTP
       <td>No</td>
     </tr>
     <tr>
-      <th scope="row">{{Glossary("Forbidden response header name")}}</th>
-      <td>Yes</td>
+      <th scope="row">{{Glossary("Forbidden response header name","Forbidden response header")}}</th>
+      <td>No</td>
```
hamishwillee:
```suggestion
      <td>Yes</td>
```

files/en-us/web/http/reference/headers/set-cookie/index.md
```diff
@@ -28,7 +28,7 @@ For more information, see the guide on [Using HTTP cookies](/en-US/docs/Web/HTTP
       <td>No</td>
     </tr>
     <tr>
-      <th scope="row">{{Glossary("Forbidden response header name")}}</th>
+      <th scope="row">{{Glossary("Forbidden response header name","Forbidden response header")}}</th>
```
bsmth, line 31: This is the only one, so it might be good to think about changing this to:

```md
<th scope="row">Header type</th>
<td>{{Glossary("Response header")}}, {{Glossary("Forbidden response header name", "Forbidden response header")}}</td>
```

And remove all of the `No` `<tr>` entries seeing as they don't add much additional info.

bsmth (approved):
Thanks a lot for the cleanup. One suggestion to think about for UX purposes / removing redundant info, but already a good fix, tnx 👍🏻

files/en-us/web/http/reference/headers/set-cookie/index.md
bsmth: See also: https://github.com/orgs/mdn/discussions/807

hamishwillee: @bsmth I did think about this. I accept it would be "better" to remove the row on the basis that given there are so few of these cases people will get use to see "no" and ignore it.

My reason for not doing this was a bit "weak" - to me it looks ugly to have a single header table. Below is what this looks like 3 ways: HTML table as it is now, HTML table with the row removed, and using a markdown table. 

Do you have a preference? If not, I propose to simple remove the rows leaving us with the middle option.

image

bsmth: >it looks ugly to have a single header table.

That's a good point, there's not much left in the table.

>Do you have a preference? If not, I propose to simple remove the rows leaving us with the middle option.

I think that's cleaner for now. I guess eventually this is front-matter material, but removing the redundant ones is a good cleanup, IMO 👍🏻

hamishwillee:
Thanks @bsmth  - as discussed, I have stripped out all the redundant "Forbidden response header" rows from my last change, so now this looks like I have removed all the Forbidden request header rows from the response header cases.

bsmth:
Thanks, Hamish 👍🏻

PR #41949, opened 2025-11-14
Update note on colspan and colgroup usage

nusdub:
### Description

Update note on colspan and colgroup usage


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #41957, opened 2025-11-15
Update index.md

SYudee:
### Description

Fixed typo, its BCE in text, not BC.


hamishwillee:
It's not a typo - BC is "Before Christ", and BCE is the secular alternative "Before Common Era". Normally I'd argue for the secular alternative, but the vast majority of native English speakers use BC, and most would not even know what BCE means.

chrisdavidmills:
> It's not a typo - BC is "Before Christ", and BCE is the secular alternative "Before Common Era". Normally I'd argue for the secular alternative, but the vast majority of native English speakers use BC, and most would not even know what BCE means.

@hamishwillee I'd be inclined to agree with this assessment, but:

1. BCE is used everywhere else in the assessment; this one instance makes it inconsistent.
2. If a reader doesn't know what something is, they can look it up. Research is a key skill for aspiring developers to become comfortable with.

For these two reasons, I'm reopening and accepting this PR. Thanks for the fix, @SYudee.

PR #41958, opened 2025-11-16
Clarify that the universal selector does not match pseudo-elements

codersjj:
### Description

Add note clarifying that the universal selector does not match pseudo-elements.

This PR adds a clarification note to the Universal selector (*) documentation to explain that it matches elements only, not pseudo-elements. This helps avoid the common misunderstanding that the universal selector can directly select pseudo-elements.

### Motivation

The current MDN page states that:

> “The CSS universal selector (`*`) matches elements of any type.”

However:

- It does not explicitly distinguish “elements” from “pseudo-elements”.
- Developers often confuse `*::before` with “universal selector matching pseudo-elements”.
- In reality, `*::before` means “match all elements, then apply the `::before` pseudo-element”.

This PR adds a concise NOTE block clarifying this behavior.

#### CSS Selectors Level 4

1. Universal selector definition:
  - [“The universal selector represents an element of any element type.”](https://www.w3.org/TR/selectors-4/#the-universal-selector:~:text=The%20universal%20selector%20is%20a%20special%20type%20selector%2C%20that%20represents%20an%20element%20of%20any%20element%20type.)

2. Pseudo-elements:

- ["a pseudo-element represents an element not directly present in the document tree. They are used to create abstractions about the document tree beyond those provided by the document tree."](https://drafts.csswg.org/selectors/#pseudo-elements:~:text=a%20pseudo%2Delement%20represents%20an%20element%20not%20directly%20present%20in%20the%20document%20tree.%20They%20are%20used%20to%20create%20abstractions%20about%20the%20document%20tree%20beyond%20those%20provided%20by%20the%20document%20tree.)
- [“Pseudo-elements do not exist independently in the tree: they are always bound to another element on the page, called their originating element.”](https://drafts.csswg.org/selectors/#pseudo-elements:~:text=Pseudo%2Delements%20do%20not%20exist%20independently%20in%20the%20tree%3A%20they%20are%20always%20bound%20to%20another%20element%20on%20the%20page%2C%20called%20their%20originating%20element.)

These specifications collectively establish that pseudo-elements are abstract representations bound to originating elements rather than actual elements in the document tree, therefore the universal selector—which exclusively targets elements—cannot match pseudo-elements directly.


files/en-us/web/css/reference/selectors/universal_selectors/index.md
```diff
@@ -21,6 +21,12 @@ The universal selector is a special [type selector](/en-US/docs/Web/CSS/Referenc
 - `*|*` - matches all elements
 - `|*` - matches all elements without any declared namespace
 
+> [!NOTE]
+> The universal selector (`*`) matches **elements only**.
+> It does **not** match pseudo-elements by itself.
```
chrisdavidmills:
```suggestion
> It does **not** directly match pseudo-elements by itself.
```

files/en-us/web/css/reference/selectors/universal_selectors/index.md
```diff
@@ -21,6 +21,12 @@ The universal selector is a special [type selector](/en-US/docs/Web/CSS/Referenc
...
+>
+> Selectors like `*::before` work because the `*` matches all **elements**, and the pseudo-element selector (`::before`) applies to those elements. The universal selector does not directly select pseudo-elements.
```
chrisdavidmills:
```suggestion
> To match all {{cssxref("::before")}} pseudo-elements on a page, for example, you would have to use a selector like `*::before`. This works because the `*` matches all elements, and the `::before` pseudo-element is available on all elements.
```

chrisdavidmills (changes requested):
@codersjj You are correct, and I think this is a good addition to the page. However, I feel that your second and last sentences are a bit repetitive. I've provided some suggestions for fixing this problem and making the provided example a bit clearer/more instructive.

Let me know what you think.

codersjj:
> @codersjj You are correct, and I think this is a good addition to the page. However, I feel that your second and last sentences are a bit repetitive. I've provided some suggestions for fixing this problem and making the provided example a bit clearer/more instructive.
> 
> Let me know what you think.

@chrisdavidmills  LGTM! 👍

PR #41960, opened 2025-11-16
At-rule pages: change css links to macros

estelle:


files/en-us/web/css/reference/at-rules/@import/index.md
```diff
@@ -30,7 +30,7 @@ where:
 - _list-of-media-queries_
   - : Is a comma-separated list of [media queries](/en-US/docs/Web/CSS/Guides/Media_queries/Using), which specify the media-dependent conditions for applying the CSS rules defined in the linked URL. If the browser does not support any of these queries, it does not load the linked resource.
 - _layer-name_
-  - : Is the name of a [cascade layer](/en-US/docs/Web/CSS/Reference/At-rules/@layer) into which the contents of the linked resource are imported. See [`layer()`](/en-US/docs/Web/CSS/Reference/At-rules/@import/layer_function) for more information.
+  - : Is the name of a {{cssxref("@layer", "cascade layer")}} into which the contents of the linked resource are imported. See [`layer()`](/en-US/docs/Web/CSS/Reference/At-rules/@import/layer_function) for more information.
```
Josh-Cena, line 33: @estelle @chrisdavidmills Was there a motivation for this change? I think most of the writers are indifferent about the two ways and we'd treat the change from one to another as unnecessary. In fact, at this particular moment, doing so introduces _more_ flaws because the macro produces a redirect. For this particular case—and a lot others involving `@layer`—this change is actually a regression because the original link is not code-formatted but cssxref produces code formatting, and this xref macro, unlike others, does not allow disabling code formatting. I will have to revert some of these regressive changes.

Josh-Cena:
Just caught another one: https://github.com/mdn/content/pull/41973#discussion_r2535344327. I suspect there are a lot of other cases like this one, where the eventual markup changed inadvertently, but my checker can only catch the cases where it introduced dubious content (spaces in code, self-link etc.) but not others (such as changing `Reference/Values/position-area_value` to `Reference/Properties/position-area` on pages that's not the `position-area` property itself, or inadvertent code formatting for link texts without spaces). I would suggest that all merged PRs be carefully reviewed again to make sure these regressions get corrected; and again, I highly, highly agree with https://github.com/mdn/content/pull/41922#issuecomment-3539417038 that this is not a well-motivated change; if it establishes a writing convention, then it needs to be discussed first, even if not documented. I'm generally in favor of using macros but I don't think a PR that migrates links to macros is productive.

bsmth:
My 2c: it may be better to revert this series of PRs rather than review each again to find and correct regressions as the benefits don't seem to justify the time cost at the moment.

chrisdavidmills:
Q: Would the time be better spent fixing the macro, or is it going to be really complicated? I mean, surely you could query the page type and generate the URL string from there?

Would it be worth spending an hour evaluating the complexity of that before reverting all this stuff?

bsmth:
I think @argl has 'fixing the macro' on his TODO list for a while, so he can weigh in about how long it will take or how complex it will be. Certainly my personal preference would be to revert instead of spend time on it while it's having reader impact on prod, but happy to hear what others think.

chrisdavidmills:
I think I'm just gonna revert them for now. Once it is fixed, we can automate such updates if it is something we feel is worth spending the time on.

PR #41961, opened 2025-11-16
Glossary page: css macros instead of links

estelle:


pepelsbey (approved):
Looks good! Thank you :)

PR #41973, opened 2025-11-17
CSS properties: convert links to macros (part 5, the end)

estelle:


chrisdavidmills (approved):
LGTM

files/en-us/web/css/reference/properties/position-area/index.md
```diff
@@ -59,7 +59,7 @@ position-area: unset;
 
 The property value is two `<position-area>` keyterms, or the keyword `none`. If only one `<position-area>` keyterm is provided, the second keyterm is implied.
 
-- [`<position-area>`](/en-US/docs/Web/CSS/Reference/Values/position-area_value)
+- {{cssxref("&lt;position-area&gt;")}}
```
Josh-Cena, line 62: This is also a regression. Unfortunately the cssxref macro has never been smart enough to know that `&lt;&gt;` is supposed to link to the `*_value` page. It just prepends `/en-US/docs/Web/CSS`. So this inadvertently changed the link to `CSS/position-area`, which is this page itself.

PR #41977, opened 2025-11-17
CSS guides: macros missed in the first pass

estelle:
found some stragglers

### DIFF
--- a/files/en-us/web/css/guides/colors/index.md
+++ b/files/en-us/web/css/guides/colors/index.md
@@ -37,19 +37,19 @@ The CSS colors module also introduces the {{cssxref("@color-profile")}} at-rule,
 ### Functions
 
 - Color functions:
-  - [`rgb()`](/en-US/docs/Web/CSS/Reference/Values/color_value/rgb)
-  - [`hsl()`](/en-US/docs/Web/CSS/Reference/Values/color_value/hsl)
-  - [`hwb()`](/en-US/docs/Web/CSS/Reference/Values/color_value/hwb)
-  - [`lab()`](/en-US/docs/Web/CSS/Reference/Values/color_value/lab)
-  - [`lch()`](/en-US/docs/Web/CSS/Reference/Values/color_value/lch)
-  - [`oklab()`](/en-US/docs/Web/CSS/Reference/Values/color_value/oklab)
-  - [`oklch()`](/en-US/docs/Web/CSS/Reference/Values/color_value/oklch)
-  - [`color()`](/en-US/docs/Web/CSS/Reference/Values/color_value/color)
-- [`color-mix()`](/en-US/docs/Web/CSS/Reference/Values/color_value/color-mix)
-- [`contrast-color()`](/en-US/docs/Web/CSS/Reference/Values/color_value/contrast-color)
-- {{CSSXref("color_value/light-dark", "light-dark()")}}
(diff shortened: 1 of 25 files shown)


files/en-us/web/css/guides/colors/index.md
```diff
@@ -37,19 +37,19 @@ The CSS colors module also introduces the {{cssxref("@color-profile")}} at-rule,
...
+  - {{cssxref("rgb()")}}
+  - {{cssxref("hsl()")}}
+  - {{cssxref("hwb()")}}
+  - {{cssxref("lab()")}}
+  - {{cssxref("lch()")}}
+  - {{cssxref("oklab()")}}
+  - {{cssxref("oklch()")}}
+  - {{cssxref("color()")}}
+- {{cssxref("color-mix()")}}
+- {{cssxref("contrast-color()")}}
```
bsmth, line 49: As far as I understand, these don't work yet since the CSS move. There's an issue tracking this here:

- [ ] https://github.com/mdn/rari/issues/361

Is it okay to wait to migrate these until the issue is resolved?

chrisdavidmills: Ah. I approved and merged several such PRs from @estelle earlier today, as it looked like the updated links in the preview pages were working (I tested several of them to make sure, and it seemed to me like some redirecting was happening to make them work for now).

Was I wrong to do this? Sorry if I've caused any problems.

Josh-Cena:
There isn't an issue with this because the macro produces `CSS/xxx` and the redirects will cause all of them to be replaced by `CSS/Reference/Blah/xxx`. This is only a problem for new pages to be introduced after the move.

bsmth:
> There isn't an issue with this because the macro produces `CSS/xxx` and the redirects will cause all of them to be replaced by `CSS/Reference/Blah/xxx`. This is only a problem for new pages to be introduced after the move.

I don't think so, because `hsl()` for instance:

https://github.com/mdn/content/blame/main/files/en-us/web/css/reference/values/calc/index.md#L224

Worked before as:

```markdown
[`hsl()`](/en-US/docs/Web/CSS/Reference/Values/color_value/hsl)
```

But this is a red link on https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/calc#adjusting_color_channels_in_relative_colors

estelle:
> ```
> [`hsl()`](/en-US/docs/Web/CSS/Reference/Values/color_value/hsl)
> ```
> 
> But this is a red link on https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/calc#adjusting_color_channels_in_relative_colors


the rgb() right before it does work.

Josh-Cena:
Oh I see. Yes for nested pages this is a problem because the `cssxref` macro just prepends `/en-US/docs/Web/CSS/` to the first argument. `rgb` works because `/en-US/docs/CSS/rgb` exists as a redirect. It has been necessary to write `{{cssxref("color_value/hsl()")}}` for a long time. (Luckily the cssxref macro is at least smart enough to remove the prefix).

files/en-us/web/css/guides/colors/index.md
bsmth: Thanks! It might be worth reverting the macros that are broken until the linked issue is fixed. Are you able to narrow down the list to broken ones, only?

PR #41984, opened 2025-11-17
fix(docs): Ensure "See also" in CSS guides contains link to the parent module page [part1]

dipikabh:
### Description

This PR updates "See also" in CSS guides to ensure:
- There's a link to the underlying module page
- The link to the module page is the first or higher among the list of the links where applicable

### Motivation

Ensure discoverability and relational linking

### DIFF
--- a/files/en-us/web/css/guides/colors/index.md
+++ b/files/en-us/web/css/guides/colors/index.md
@@ -131,10 +131,11 @@ The CSS color module also introduces the `CSSColorProfileRule` interface. Curren
 
 ## See also
 
-- [CSS color adjustment](/en-US/docs/Web/CSS/Guides/Color_adjustment) module and the {{cssxref("print-color-adjust")}} property.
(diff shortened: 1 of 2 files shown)


estelle:
> The link to the module page is the first or higher among the list of the links where applicable

I think the module landing page should come after in the see also, as it's a less specific / more generic link than links to features withn the spec

dipikabh:
> I think the module landing page should come after in the see also, as it's a less specific / more generic link than links to features withn the spec

The idea is that every guide is linked to (and is a child page in the directory structure of) a module landing page, at least at the moment. Prioritizing that link near the top of the "See also" helps clarify that connection. The module landing page is the root that connects to other content related to the guide.

For example, in https://pr41984.review.mdn.allizom.net/en-US/docs/Web/CSS/Guides/Cascade/Inheritance#see_also, the "CSS cascading and inheritance module" list item is intentionally widely separated from the "CSS nesting module" list item; the first is the module that defines the inheritance concept and should appear higher up in the list, whereas the second one is connected less directly.

estelle (approved):

PR #42001, opened 2025-11-18
Add CSS modules to CSS reference landing page

estelle:


dipikabh:
Hi @estelle, the module page listing is available at https://developer.mozilla.org/en-US/docs/Web/CSS/Guides.
Close this one?

estelle:
> Hi @estelle, the module page listing is available at https://developer.mozilla.org/en-US/docs/Web/CSS/Guides. Close this one?

The drop down menu reads "guides", but under that is a list of module landing pages, not guides.
We should add modules to the reference pages, as they are reference pages,, or css-modules, not guides.  The guides page (https://developer.mozilla.org/en-US/docs/Web/CSS/Guides) could then to link to all the guides (the css-guide pages, rather than the module landing pages). 

Alternatively, we could add the guides to where the cookbook links are (We should not be highlighting the cookbook. Two years ago I updated all the cookbook pages to make them less archaic, bug we still only include fixes to things that are no longer issues)

dipikabh:
> We should add modules to the reference pages, as they are reference pages,, or css-modules, not guides. 

For all practical purposes, we're regarding the module pages as guides. They're guides to different parts of the spec related to a feature.

The current sidebar and breadcrumbs now correctly reflect how modules and guides are organized in the directory structure, that is, they're not under `css/reference` but under `css/guides`. So in the sidebar and in the breadcrumbs, they're both under one umbrella, "Guides".

On a related note, linking to the parent (module landing page) in the "Reference" section in the sidebar (as it was before the move) and the children (related guides) in the "Guides" section in the sidebar didn't make the relationship between the two very clear in the navigation, and imo, wasn't very intuitive. It also created a visual shift to different areas of the sidebar. 

The current sidebar isnt't ideal either. Two immediate issues:
- We don't yet have a listing page for the "Modules" entry in the sidebar.
- Each parent guide node is unlinked. Ideally, each node should link to its module landing page. For example, "Anchor positioning" in the sidebar does not lead to a page at the moment. It can very easily be linked to its module page at https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Anchor_positioning.

Both of these are a deviation from the rest of the items in the CSS sidebar, where every node is linked to a listing or a landing page.
**Proposal**: I'd like to merge the individual guides and the "Modules" menu in the sidebar (removing "Modules") and then link each node under "Guides" to the respective module landing page.
WDYT about this change?

> The guides page (https://developer.mozilla.org/en-US/docs/Web/CSS/Guides) could then to link to all the guides (the css-guide pages, rather than the module landing pages).

I agree with this, though I think we should list the link to all guides as a subset under the module landing page they correspond to. Perhaps we can request an enhancement to the [`{{subpageswithsummaries}}` macro](https://github.com/mdn/rari/blob/main/crates/rari-doc/src/templ/templs/subpages_with_summaries.rs#L11) to support this.

dipikabh:
Hi @estelle, do you have any further thoughts about the proposal in https://github.com/mdn/content/pull/42001#issuecomment-3553525255?

PR #42014, opened 2025-11-19
Enhance documentation for 'object-fit' CSS property

youssef-el-atmani:
Include a note on the importance of setting `width` and `height` for effective `object-fit` application.

### Description

I added a note about how to use the `object-fit` property effectively
### Motivation

 - Because I lost a lot of time trying to understand why the `object-fit` property doesn't work on my personal examples, I did what the docs instruct, but that property fails to apply. And just later I discovered that I should set a **height** and **width** to the image in order for that `object-fit` to be applied effectively. I think that works the same for all the replaced-elements, but I only tested it for `img` elements.
 - Readers will be able to use that `object-fit` effectively.
### Additional details

There is no clue or note about that even in the [object-fit mdn-reference](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/object-fit)

### DIFF
--- a/files/en-us/learn_web_development/core/styling_basics/images_media_forms/index.md
+++ b/files/en-us/learn_web_development/core/styling_basics/images_media_forms/index.md
@@ -154,10 +154,6 @@ img {
 
 You could also try the value of `fill`, which will fill the box but not maintain the aspect ratio.
 
-> [!NOTE]
-> For the {{cssxref("object-fit")}} property to be applied successfully to an image, `width` and `height` properties should typically be set to `100%`.
-> In the example above, you can comment out the `height: 100%;` or `width: 100%;` declarations —or both— from the `img` selctor, to observe that the {{cssxref("object-fit")}} property is no longer effective.
-
 ## Replaced elements in layout
 
 When using various CSS layout techniques on replaced elements, you may well find that they behave slightly differently from other elements. For example, in a grid layout, elements are stretched by default to fill their entire [grid areas](/en-US/docs/Glossary/Grid_Areas). Images do not stretch; instead, they are aligned to the start of their grid areas.


youssef-el-atmani:
Hi @chrisdavidmills, would you please review this PR.
By saying **"I lost a lot of time"** in the **Motivation** for this PR; I am not complaining, I just want to emphasis the importance of adding that note.

chrisdavidmills (changes requested):
Hi @youssef-el-atmani! Yup, I'm happy to provide a review. In fact, I am the main overseer of the Learn area, so feel free to ask me for reviews on any PRs that touch Learn area content.

And don't worry, you didn't sound like you were complaining. It is a fact that when you get stuck on something and the documentation doesn't provide enough information, it can be frustrating. You've implied that the current docs caused you frustration, and then gone on to provide a fix rather than just moaning about it, which is the best of all worlds. 

On to my review:

Adding a comment is valid, but I suggest phrasing it differently. We already taught the reader that they need to size images for them to be constrained inside their containers earlier on in the article, so this should be more of a reminder, which makes it clear that `object-fit` only specifies how an image should fit inside its container, and doesn't actually do any sizing on its own.

So maybe something along the lines of

"As we saw previously, images that are larger than their containers need to have `width` and/or `height` explicitly set to constrain them inside the container. This is still true when using `object-fit` — it only sets how an image should fit inside a box, and doesn't provide any sizing on its own. For example, if you try commenting out the `img` ruleset in the previous example, you'll see that `object-fit` no longer has any effect."

What do you think?

youssef-el-atmani:
Hi @chrisdavidmills, sorry for the late respond, I read your review earlier, but I didn't have the time to process it.

You're right, my note is not accurate, and actually I was totally wrong about that property, I have just realized that.

I did a lot of playing with that `property`, I guess I have a solid idea about it now.

I think it is better if we explain first how `object-fit` property works under the hood (since it is simple and straight forward), and just after that we can link that functionality to handling image-overflow within containers.

I am suggesting something similar to the following:

> [!note]
> Please note that my suggestion is just a sketch, or an idea, it could be much better if it have some code snippets and visual examples; which unfortunately doesn't have.


---
### [Handling image overflow with object-fit](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Images_media_forms#handling_image_overflow_with_object-fit)

> *I suggest replacing all that section with something similar to the following*

For handling image overflaw, we have a special property named `object-fit`, to understand how that property work,
Let us assume that we have a large image **1500px * 750px**, that we want to scale down to a specific `width` and `height`, let's say **700px * 300px**.
If we  set:
```css
width: 700px;
height: 300px;
```
The image loose its aspect-ratio.

to scale down the image without loosing its aspect-ratio, we can use the `object-fit` property,
what it does, is first: it creates an invisible box (let's name it the imaginary-box) that has those dimension we specified on the image, for our case `width: 700px;` and `height: 300px;`, and displays the image in that box based on the **value** you choose for the `object-fit` property:
First assume that we set a `background-color` and `border` to the image:
```css
img {
width: 700px;
height: 300px;
background-color: green;
border: 2px black solid;
}
```
- `contain`: the image will be scaled down to fit inside the invisible-box, the whole image will be displayed, you will see that the free space for that box will be filled with that green background color.
- `cover`: the image will cover all that box but it will be cropped, and no background-color will be seen
- `fill`: the whole image will be displayed, and it will be stretched to fill all the space available in the box, similar to the initial state before introducing the `object-fit` property, the image will loose it aspect-ratio.

If you have a container that has an image that overflow, you can use that `object-fit` by giving its *imaginary-box* identical width and height to the container, thus the imagined-box that has the image will set top of the container, and then you can use the `object-fit` power to manage the overflow behavior. You're controlling the image within the *imagined-box*, but since it has identical size to the container, it fits perfectly within the container.
A shortcut to make that imagined-box size identical to the container is using `width: 100%` and `height: 100%`.

---
Please tell me what do you think?

I really think that that section need a slightly more work, which has nothing to do with the original Note I did suggest, and that `object-fit` property seems to me very important, because I think controlling image display within containers or just setting specific dimension to them without loosing their aspect-ratio is very common.

youssef-el-atmani:
@chrisdavidmills,
Here is a further explenation of the cause of the struggling with that property,
The docs say:
> You can make other choices about images inside containers. For example, you may want to size an image so it completely covers a box.

> The [object-fit](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/object-fit) property can help you here. When using object-fit the replaced element _can be sized to fit a box in a variety of ways_.

What I get from that **can be sized to fit a box in a variety of ways.**, is that when we have an image that is a child of a box-container, the `object-fit` property could be used to position the image within that box-container (somehow similar to background-size for background-image: that's what I thought in the beginning), but actually that is not true, the `object-fit` position the image within the image dimension (width & height). And that `object-fit` position the image within the box-container only when the image has the same size as the box container.


I don't know if that suggested-sketch I made (for replacing the current section) in the previous comment fit within the MDN-style, if it doesn't, please see how you can improve the original section, or rewrite it in different way.
If you think I can provide any help, I would be happy to collaborate.

youssef-el-atmani:
> As we saw previously, images that are larger than their containers need to have `width` and/or `height` explicitly set to constrain them inside the container. This is still true when using `object-fit` — it only sets how an image should fit inside a box, and doesn't provide any sizing on its own. For example, if you try commenting out the `img` ruleset in the previous example, you'll see that `object-fit` no longer has any effect.

I had to respond to this first before providing any further discussion, sorry for that @chrisdavidmills.



> it only sets how an image should fit inside a box, and doesn't provide any sizing on its own.

The first half seems to me inaccurate, because the `object-fit` doesn't control how an image _fit inside box_ by default, it only achieve that when tweaks it, when making the image's size same as the box's size.
And that only tell part of the truth, because it doesn't  show how actually object-fit work.

I know that what I said here is redundant to what I said earlier, but I felt that it is inappropriate to not respond directly to your review.

chrisdavidmills:
Hi there, @youssef-el-atmani! 

Thank you for the last few comments; it is useful to help me understand the trouble you've had with understanding this property. I think I am going to close this PR and create my own to try to fix this. Then I'll invite you to have a look and see if you think it fixes the problem.

I say this because I want to try to be as concise here as possible, and not turn it into too big a discussion — this article is for beginners, after all, and a more detailed explanation should go on the object-fit reference page, if anywhere. There are also a few other things I want to clear up, for example:

- object-fit changes the way the image content (i.e., the image itself) is resized to fit its container, which is the `` element. Setting its `width` and `height` to `100%` makes the `` element stretch to fit its container element in both dimensions, but this does create a bit of confusion over what container we are talking about.
- I don't think including `background-color` helps much, and possibly makes it sound like the behavior you describe above is special to `` elements and `object-fit`. In reality, this is always the case when box content doesn't fill its container.

Anyway, I'll create a PR now and you can let me know what you think.

youssef-el-atmani:
Thanks @chrisdavidmills,
I am looking forward to see your edit

PR #42020, opened 2025-11-19
Scroll-driven animations: animation-range and <timeline-range-name> data type

estelle:
* create new page for `<timeline-range-name>`
* rewrite `animation-range` shorthand property.
* make example detail how the property works and interacts with afm

### DIFF
--- a/.vscode/dictionaries/code-entities.txt
+++ b/.vscode/dictionaries/code-entities.txt
@@ -256,6 +256,7 @@ Gloc
 glossarysidebar
 gonm
 GPOS
+grba
 Grek
 greklow
 GSUB
@@ -553,6 +554,7 @@ orya
 oscpu
 osfile.jsm
 osma
+otpauth
 otpcredential
 outbound-rtpp
 overlayinfo
--- a/.vscode/dictionaries/ignore-list.txt
+++ b/.vscode/dictionaries/ignore-list.txt
@@ -216,6 +216,7 @@ mdnsidebar.yaml
 megastore
 meinheld
 morethanwarm
+MQCHJLS6FJXT2BGQJ6QMG3WCAVUC2HJZ
 MYAGE
 mybooks
 myclass
--- a/.vscode/dictionaries/proper-names.txt
+++ b/.vscode/dictionaries/proper-names.txt
@@ -303,9 +303,11 @@ Kahlo
 Kaku
 Kang
 Kaply
+Karmadrome
 Kaspersky
 Katari
 Kazam
+Keegan
 kenji-yamasaki
 Kerckhoffs
 Kersee
--- a/files/en-us/learn_web_development/core/styling_basics/sizing/index.md
+++ b/files/en-us/learn_web_development/core/styling_basics/sizing/index.md
@@ -8,7 +8,7 @@ sidebar: learnsidebar
 
 {{PreviousMenuNext("Learn_web_development/Core/Styling_basics/Test_your_skills/Values", "Learn_web_development/Core/Styling_basics/Test_your_skills/Sizing", "Learn_web_development/Core/Styling_basics")}}
 
-In the various lessons so far, you have come across several ways to size items on a web page using CSS. Understanding how big the different features in your design will be is important. So, in this lesson we will summarize the various ways elements get a size via CSS and define a few terms about sizing that will help you in the future.
+In the various lessons so far, you have come across several ways to size items on a web page using CSS. Understanding how big the different features in your design will be is important. So, in this lesson, we will summarize the various ways elements get a size via CSS and define a few terms about sizing that will help you in the future.
 
 <table>
   <tbody>
@@ -114,7 +114,7 @@ Due to this problem of overflow, fixing the height of elements with lengths or p
 
 ### Using percentages
 
-In many ways, percentages act like length units, and as we [discussed in the lesson on values and units](/en-US/docs/Learn_web_development/Core/Styling_basics/Values_and_units#percentages), they can often be used interchangeably with lengths. When using a percentage you need to be aware what it is a percentage _of_. In the case of a box inside another container, if you give the child box a percentage width it will be a percentage of the width of the parent container.
+In many ways, percentages act like length units, and as we [discussed in the lesson on values and units](/en-US/docs/Learn_web_development/Core/Styling_basics/Values_and_units#percentages), they can often be used interchangeably with lengths. When using a percentage, you need to be aware of what it is a percentage _of_. In the case of a box inside another container, if you give the child box a percentage width, it will be a percentage of the width of the parent container.
 
 ```html live-sample___percent-width
 <div class="container">
@@ -147,7 +147,7 @@ Try editing the above example:
 
 If you set `margins` and `padding` as a percentage, you may notice some strange behavior.
 
-In the below example we have a box, on which we've set a {{cssxref("margin")}} of 10% and a {{cssxref("padding")}} of `10%`. The padding and margin on the top and bottom of the box are the same size as the padding and margin on the left and right.
+In the below example, we have a box, on which we've set a {{cssxref("margin")}} of 10% and a {{cssxref("padding")}} of `10%`. The padding and margin on the top and bottom of the box are the same size as the padding and margin on the left and right.
 
 ```html live-sample___percent-mp
 <div class="box">I have margin and padding set to 10% on all sides.</div>
@@ -169,7 +169,7 @@ body {
 
 You might expect the percentage top and bottom margins to be a percentage of the element's height, and the percentage left and right margins to be a percentage of the element's width. However, this is not the case!
 
-When you use margin and padding set in percentages, the value is calculated from the **inline size** of the containing block — therefore the width when working in a horizontal language. In our example, all of the margins and padding are `10%` of the width. This means you can have equal-sized margins and padding all around the box. This is a fact worth remembering if you use percentages in this way.
+When you use margin and padding set in percentages, the value is calculated from the **inline size** of the containing block — therefore, the width when working in a horizontal language. In our example, all of the margins and padding are `10%` of the width. This means you can have equal-sized margins and padding all around the box. This is a fact worth remembering if you use percentages in this way.
 
 ## min- and max- sizes
 
@@ -181,7 +181,7 @@ In the next example, you can see two boxes, both with a defined `min-height` of 
 <div class="wrapper">
   <div class="box"></div>
   <div class="box">
-    These boxes both have a min-height set, this box has content in it which
+    These boxes both have a min-height set. This box has content in it, which
     will need more space than the assigned height, and so it grows from the
     minimum.
   </div>
@@ -309,13 +309,13 @@ body {
 
 {{EmbedLiveSample("vw-vh")}}
 
-If you change the `vh` and `vw` values this will change the size of the box and font; changing the viewport size will also change their sizes because they are sized relative to the viewport. To see the example change when you change the viewport size you will need to load the example in a new browser window that you can resize (as the embedded `<iframe>` that contains the example shown above is its viewport). Open the example, resize the browser window, and observe what happens to the size of the box and text.
+Changing the `vh` and `vw` values will change the size of the box and font, respectively; changing the viewport size will also change the box and font sizes because they are sized relative to the viewport. To see the box and text size change when you adjust the viewport size, {{LiveSampleLink("vw-vh", "load the example in a new tab")}} and resize the browser window.
 
-Sizing things according to the viewport can be useful in your designs. For example, if you want a full-page hero section to show before the rest of your content, making that part of your page `100vh` high will push the rest of the content below the viewport, meaning that it will only appear once the document is scrolled.
+Sizing things according to the viewport can be useful in your designs. For example, if you want a full-page banner to show before the rest of your content, making that part of your page `100vh` high will push the rest of the content below the viewport, meaning that it will only appear once the document is scrolled.
 
 ## Summary
 
-This lesson has given you a rundown of some key issues that you might run into when sizing things on the web. When you move onto [CSS Layout](/en-US/docs/Learn_web_development/Core/CSS_layout), sizing will become very important in mastering the different layout methods, so it is worth understanding the concepts here before moving on.
+This lesson has given you a rundown of some key issues that you might run into when sizing things on the web. When you move on to [CSS Layout](/en-US/docs/Learn_web_development/Core/CSS_layout), sizing will become very important in mastering the different layout methods, so it is worth understanding the concepts here before moving on.
 
 In the next article, we'll give you some tests that you can use to check how well you've understood and retained the information we've provided on CSS sizing.
 
--- a/files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
@@ -43,7 +43,7 @@ Here, the contents of `data.className` or `data.color` could contain HTML that c
 
 ### jQuery
 
-When using jQuery, functions such as `attr()` and `text()` escape content as it's added to a DOM. So, the "favorite color" example from above, implemented in jQuery, would look like this:
+When using jQuery, functions such as `attr()` and `text()` avoid treating any character as HTML syntax. So, the "favorite color" example from above, implemented in jQuery, would look like this:
 
 ```js example-good
 let node = $("<div>");
--- a/files/en-us/web/api/credentialscontainer/create/index.md
+++ b/files/en-us/web/api/credentialscontainer/create/index.md
@@ -74,7 +74,7 @@ This example creates a password credential from a {{domxref("PasswordCredentialI
 
 ```js
 const credInit = {
-  id: "serp1234", // "username" in a typical username/password pair
+  id: "serpent1234", // "username" in a typical username/password pair
   name: "Serpentina", // display name for credential
   origin: "https://example.org",
   password: "the last visible dog",
@@ -89,7 +89,7 @@ makeCredential.addEventListener("click", async () => {
   console.log(cred.name);
   // Serpentina
   console.log(cred.id);
-  // serp1234
+  // serpent1234
   console.log(cred.password);
   // the last visible dog
 });
--- a/files/en-us/web/api/headers/get/index.md
+++ b/files/en-us/web/api/headers/get/index.md
@@ -27,8 +27,10 @@ get(name)
 
 - `name`
   - : The name of the HTTP header whose values you want to retrieve from the
-    `Headers` object. If the given name is not the name of an HTTP header, this
-    method throws a {{jsxref("TypeError")}}. The name is case-insensitive.
+    `Headers` object. If the given name doesn't match the
+    [field-name](https://httpwg.org/specs/rfc9110.html#fields.names)
+    production in the HTTP specification, this method throws a {{jsxref("TypeError")}}.
+    The name is case-insensitive.
 
 ### Return value
 
--- a/files/en-us/web/api/passwordcredentialinit/index.md
+++ b/files/en-us/web/api/passwordcredentialinit/index.md
@@ -45,7 +45,7 @@ This example constructs an object literal to initialize a password credential.
 
 ```js
 const credInit = {
-  id: "serp1234", // "username" in a typical username/password pair
+  id: "serpent1234", // "username" in a typical username/password pair
   name: "Serpentina", // display name for credential
   origin: "https://example.org",
   password: "the last visible dog",
@@ -60,7 +60,7 @@ makeCredential.addEventListener("click", async () => {
   console.log(cred.name);
   // Serpentina
   console.log(cred.id);
-  // serp1234
+  // serpent1234
   console.log(cred.password);
   // the last visible dog
 });
--- a/files/en-us/web/css/guides/cascade/specificity/index.md
+++ b/files/en-us/web/css/guides/cascade/specificity/index.md
@@ -487,6 +487,6 @@ A few things to remember about specificity:
 - Values: [initial](/en-US/docs/Web/CSS/Guides/Cascade/Property_value_processing#initial_value), [computed](/en-US/docs/Web/CSS/Guides/Cascade/Property_value_processing#computed_value), [used](/en-US/docs/Web/CSS/Guides/Cascade/Property_value_processing#used_value), and [actual](/en-US/docs/Web/CSS/Guides/Cascade/Property_value_processing#actual_value)
 - [Value definition syntax](/en-US/docs/Web/CSS/Guides/Values_and_units/Value_definition_syntax)
 - [CSS nesting](/en-US/docs/Web/CSS/Guides/Nesting) module
-- [Specificity Calculator](https://specificity.keegan.st/) on specificity.keegan.st: An interactive website to test and understand your own CSS rules
+- [Specificity Calculator](https://specificity.keegan.st/) by Keegan Street: An interactive website to test and understand your own CSS rules
 - [SpeciFISHity](https://specifishity.com/) on specifishity.com: A fun way to learn about CSS specificity
 - [_ID-CLASS-TYPE_ exercise](https://estelle.github.io/CSS/selectors/exercises/specificity.html): A specificity quiz by Estelle Weyl
--- a/files/en-us/web/css/reference/properties/math-shift/index.md
+++ b/files/en-us/web/css/reference/properties/math-shift/index.md
@@ -2,13 +2,9 @@
 title: math-shift
 slug: Web/CSS/Reference/Properties/math-shift
 page-type: css-property
-status:
-  - experimental
 browser-compat: css.properties.math-shift
 sidebar: cssref
 ---
-
-{{SeeCompatTable}}
 
 The `math-shift` property indicates whether superscripts inside MathML formulas should be raised by a normal or compact shift.
 
--- a/files/en-us/web/css/reference/properties/text-decoration-inset/index.md
+++ b/files/en-us/web/css/reference/properties/text-decoration-inset/index.md
@@ -77,7 +77,7 @@ One or two {{cssxref("&lt;length>")}} values, or the keyword `auto`.
 - {{cssxref("&lt;length>")}}
   - : Specifies the amount to adjust the text decoration position by. Positive values inset the text decoration (make it shorter) while negative values outset the text decoration (make it longer). If one value is specified, it applies to both the text decoration start and end points. If two values are specified, the first one applies to the text decoration start point and the second one applies to the text decoration end point.
 - `auto`
-  - : The browser chooses a start and end inset amount to ensure that, if two decorated text boxes appear side-by-side, the appearence of a gap is created between their text decorations so they do not appear to have a single text decoration.
+  - : The browser chooses a start and end inset amount to ensure that, if two decorated text boxes appear side-by-side, the appearance of a gap is created between their text decorations so they do not appear to have a single text decoration.
 
 ## Description
 
--- a/files/en-us/web/html/guides/microdata/index.md
+++ b/files/en-us/web/html/guides/microdata/index.md
@@ -147,7 +147,7 @@ In some cases, search engines covering specific regions may provide locally-spec
 {{ EmbedLiveSample('HTML', '', '100') }}
 
 > [!NOTE]
-> A handy tool for extracting microdata structures from HTML is Google's [Structured Data Testing Tool](https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data). Try it on the HTML shown above.
+> A handy tool for extracting and verifying microdata structures from HTML is the [Schema Markup Validator](https://validator.schema.org/). Try it on the HTML shown above.
 
 ## See also
 
--- a/files/en-us/web/javascript/reference/global_objects/asyncdisposablestack/adopt/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/asyncdisposablestack/adopt/index.md
@@ -45,7 +45,7 @@ This function creates a file handle (as a Node.js [`FileHandle`](https://nodejs.
 async function readFile(path) {
   await using disposer = new AsyncDisposableStack();
   const handle = disposer.adopt(
-    fs.open(path),
+    await fs.open(path),
     async (handle) => await handle.close(),
   );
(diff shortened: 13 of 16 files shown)


files/en-us/web/css/reference/values/timeline-range-name/index.md
```diff
@@ -0,0 +1,61 @@
...
+    - If the subject element is smaller than the scrollport, it ranges from the point where the subject element is first completely contained by the scroll port (`0%` progress), to the point where it is no longer completely contained by the scroll port (`100%` progress).
+    - If the subject element is larger than the scrollport, it ranges from the point where the subject element first completely covers the scroll port (`0%` progress), to the point where it no longer completely covers the scroll port (`100%` progress).
+
+- `entry`
+  - : Represents the range of a view progress timeline from the point where the subject element first starts to enter the scroll port, to the point where it has completely entered the scroll port. `0%` is equivalent to `0%` of the `cover` range. `100%` is equivalent to `0%` of the `contain` range.
+
+- `exit`
+  - : Represents the range of a view progress timeline from the point where the subject element first starts to exit the scroll port, to the point where it has completely exited the scroll port. `0%` is equivalent to `0%` of the `contain` range. `100%` is equivalent to `0%` of the `cover` range.
+
+- `entry-crossing`
```
chrisdavidmills, line 36: The `entry-crossing` and `exit-crossing` descriptions are very dense and super hard to understand. I've read them several times and am not sure what they mean, or how these values differ from `entry` and `exit`. Can you try to clarify?

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,57 @@ This property is a shorthand for the following CSS properties:
...
+animation-range: 100px;
+animation-range: cover;
+animation-range: contain;
+
+/* Two value syntax */
+/* Named timeline range value */
+animation-range: cover 20%;
+animation-range: contain 100px;
+/* Range start and end */
 animation-range: normal 25%;
```
chrisdavidmills, line 34: Shouldn't these first two be in the "Named timeline range value" section? And shouldn't "Named timeline range value" be something like "Named timeline range offset"? You're basically specifying one timeline range name, but specifying an offer from the start of that range, right?

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,57 @@ This property is a shorthand for the following CSS properties:
...
 
-- If the value is a {{cssxref("length-percentage")}} or `normal`, `<animation-range-start>` will take that value, and `<animation-range-end>` will equal `normal`.
-- If the value is a named timeline range without a `<length-percentage>` following it, the range will be between that named timeline range at 0% and 100%.
-- If the value is a named timeline range with a `<length-percentage>` following it, the range will start at that named timeline range and percentage, and end at that named timeline range and 100%.
+## Description
 
-### Values
+The `animation-range` shorthand property sets the `animation-range-start` and `animation-range-end` values, defining where along the animation timeline the animation will start and end. The `0%` keyframe effect will not be applied to the element until the range start is reached if the {{cssxref("animation-fill-mode")}} is not set to `backwards` or `both`.
```
chrisdavidmills:
```suggestion
The `animation-range` shorthand property sets the `animation-range-start` and `animation-range-end` values, defining where along the animation timeline the animation will start and end. The `0%` keyframe effect will not be applied to the element until the range start is reached unless the {{cssxref("animation-fill-mode")}} is set to `backwards` or `both`.
```

Does that work? The double negative was confusing.

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,57 @@ This property is a shorthand for the following CSS properties:
...
+/* Four value syntax */
+animation-range: cover 0% cover 200px;
+animation-range: entry 10% exit 100%;
 ```
 
-The `animation-range` shorthand property can be applied to a container element as a combination of the `<animation-range-start>` and `<animation-range-end>` values. If both the values are specified, they will be interpreted in the order `<animation-range-start>` then `<animation-range-end>`.
+### Values
 
-As shown by the comments in the syntax block above, if only a single value is provided there are a couple of possible interpretations:
+- `<animation-range-start>`
```
chrisdavidmills, line 58: Mention that one or two values can be specified?

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,57 @@ This property is a shorthand for the following CSS properties:
...
 
-- If the value is a {{cssxref("length-percentage")}} or `normal`, `<animation-range-start>` will take that value, and `<animation-range-end>` will equal `normal`.
-- If the value is a named timeline range without a `<length-percentage>` following it, the range will be between that named timeline range at 0% and 100%.
-- If the value is a named timeline range with a `<length-percentage>` following it, the range will start at that named timeline range and percentage, and end at that named timeline range and 100%.
+## Description
 
-### Values
+The `animation-range` shorthand property sets the `animation-range-start` and `animation-range-end` values, defining where along the animation timeline the animation will start and end. The `0%` keyframe effect will not be applied to the element until the range start is reached if the {{cssxref("animation-fill-mode")}} is not set to `backwards` or `both`.
+
+The property can be applied to a container element as a combination of the `<animation-range-start>` and `<animation-range-end>` values. If both the values are specified, they will be interpreted in the order `<animation-range-start>` then `<animation-range-end>`.
```
chrisdavidmills:
```suggestion
If two values are specified as components of the `<animation-range>` property, they will be interpreted in the order `<animation-range-start>` then `<animation-range-end>`.
```

chrisdavidmills: That first sentence was confusing, I thought.

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,57 @@ This property is a shorthand for the following CSS properties:
...
+
+The value of each component is either the keyword `normal`, a {{cssxref("length-percentage")}}, or a [`<timeline-range-name>`](/en-US/docs/Web/CSS/Reference/Values/timeline-range-name), optionally defined with a `<length-percentage>`
+
```
chrisdavidmills, line 70: Now you are talking about what happens if only a single component is specified for animation-range. Make this more obvious with a lead-up sentence:

"if a single component is specified in the value of animation-range, it will be interpreted as follows:"

And then include the different possibilities below it as bullet points.

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,57 @@ This property is a shorthand for the following CSS properties:
...
+If the value is a {{cssxref("length-percentage")}} or `normal`, this is the value for the `<animation-range-start>` and `<animation-range-end>` will equal `normal`. For example, `animation-range: 20%;` is the equivalent to `animation-range: 20% normal;` and `animation-range: normal;` is the equivalent to `animation-range: normal normal;`.
```
chrisdavidmills:
```suggestion
If the value is a {{cssxref("length-percentage")}} or `normal`, this is interpreted as the value for `<animation-range-start>`. The `<animation-range-end>` value will be implicitly set to `normal`. For example, `animation-range: 20%;` is equivalent to `animation-range: 20% normal;` and `animation-range: normal;` is equivalent to `animation-range: normal normal;`.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,57 @@ This property is a shorthand for the following CSS properties:
...
+
+If the value is a [`<timeline-range-name>`](/en-US/docs/Web/CSS/Reference/Values/timeline-range-name) without a `<length-percentage>` following it, that timeline range name is applied to both the start and end animation range components at `0%` and `100%`. For example, `animation-range: contain;` is equivalent to `animation-range: contain 0% contain 100%;`.
```
chrisdavidmills:
```suggestion
If the value is a [`<timeline-range-name>`](/en-US/docs/Web/CSS/Reference/Values/timeline-range-name) without a `<length-percentage>` following it, that timeline range name is applied to the `<animation-range-start>` and `<animation-range-end>` components at `0%` and `100%`, respectively. For example, `animation-range: contain;` is equivalent to `animation-range: contain 0% contain 100%;`.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,57 @@ This property is a shorthand for the following CSS properties:
...
-
-- `<timeline-range-name> <length-percentage>`
-  - : A combination value that equals the specified percentage or distance through the specified named timeline range, measured from the start of that timeline range.
-
-> [!NOTE]
-> The scroll port (see {{glossary("Scroll container")}} for more details) area known as the view progress visibility range is the area inside which the subject element of a _named view progress timeline_ animation is deemed to be visible. By default, this is the full range of the scrollport, but it can be adjusted using the {{cssxref("view-timeline-inset")}} property.
+If the value is a single `<timeline-range-name>` with a `<length-percentage>` following it, in that order, the pair defines the range start and the `<timeline-range-name>` is also applied to the range end set to `100%`. For example, `animation-range: cover 20%;` is equivalent to `animation-range: cover 20% cover 100%` and `animation-range: contain 100px;` is equivalent to `animation-range: contain 100px contain 100%`.
```
chrisdavidmills:
```suggestion
If the value is a single `<timeline-range-name>` with a `<length-percentage>` following it, in that order, the pair defines the `<animation-range-start>`. The `<animation-range-end>` is set to the same `<timeline-range-name>` at `100%`. For example, `animation-range: cover 20%;` is equivalent to `animation-range: cover 20% cover 100%` and `animation-range: contain 100px;` is equivalent to `animation-range: contain 100px contain 100%`.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,57 @@ This property is a shorthand for the following CSS properties:
...
+
+Otherwise, both `<animation-range-start>` and `<animation-range-end>` are defined. If the first value is the keyword `normal` or a `<length-percentage>`, that defines the `<animation-range-start>`, and the rest of the value defines the `<animation-range-end>`. For example, `animation-range: normal 25%;` is equivalent to `animation-range-start: normal; animation-range-end: 25%`, `animation-range: 25% 50%;` is equivalent to `animation-range-start: 25%; animation-range-end: 50%`, and `animation-range: 25% contain;` is equivalent to `animation-range-start: 25%; animation-range-end: contain`.
```
chrisdavidmills: These examples should be moved up to where we discuss setting two values. It feels a bit weird down here.

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -94,22 +80,15 @@ One or two values representing the [`animation-range-start`](/en-US/docs/Web/CSS
...
-See the [View Timeline Ranges Visualizer](https://scroll-driven-animations.style/tools/view-timeline/ranges/) for a visual explanation of what all the value types mean.
-
-### Creating a named view progress timeline with range
+### Basic usage of the `animation-range` property
 
-A view progress timeline named `--subject-reveal` is defined using the `view-timeline` property on a subject element with a `class` of `animation`.
-This is then set as the timeline for the same element using `animation-timeline: --subject-reveal;`. The result is that the subject element animates as it moves upwards through the document as it is scrolled.
-
-An `animation-range` declaration is also set to make the animation begin later than expected, and finish earlier.
+In this example, we reduce the duration of the view progress scroll animation by using the `animation-range` property to offset the start and end of the animation.
```
chrisdavidmills:
```suggestion
In this example, we reduce the duration of a view progress scroll animation by using the `animation-range` property to offset the start and end of the animation.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -141,21 +125,61 @@ The HTML for the example is shown below.
...
+  </p>
 </div>
 ```
 
 #### CSS
 
-The `subject` element and its containing `content` element are styled minimally, and the text content is given some basic font settings:
+We define an animation that animates an element's opacity, scale, and color, causing it to fade in, size up, and change colors as the animation progresses. We apply this animation to the `animatedElement` with the {{cssxref("animation")}} shorthand.
+
+A view progress timeline is created by setting the [`view()`]() function as the value of the {{cssxref("animation-timeline")}} property on our `animatedElement`. The result is that the element animates as it moves upwards through the document as it is scrolled. We declared the shorthand before the `animation-timeline`, as the shorthand resets this property.
```
chrisdavidmills:
```suggestion
A view progress timeline is created by setting the [`view()`]() function as the value of the {{cssxref("animation-timeline")}} property on our `animatedElement`. This results in the element animating as it moves upwards through the document when scrolled. We declared the shorthand before the `animation-timeline`, as the shorthand resets this property.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -141,21 +125,61 @@ The HTML for the example is shown below.
...
+
+Lastly, an `animation-range` declaration is also set to make the animation begin later than expected, and finish earlier.
+
+it moves up the scroller.
```
chrisdavidmills: Delete? This doesn't make any sense.

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -141,21 +125,61 @@ The HTML for the example is shown below.
...
+
+  to {
+    background-color: darkturquoise;
+    opacity: 0.75;
+    transform: scaleX(0.75);
+  }
+}
+```
+
+The `animatedElement` and its container's other styles were hidden for the sake of brevity.
```
chrisdavidmills:
```suggestion
The other styles are hidden for the sake of brevity.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -177,39 +201,15 @@ p {
...
-  }
-}
-```
+Scroll to see the element being animated.
 
-#### Result
+{{EmbedLiveSample("Examples", "100%", "480px")}}
 
-Scroll to see the subject element being animated.
+Note how the `from`, or `0%`, keyframe property values are not applied to the animated element until the top block border edge is 10% past the container's bottom edge; it is full size, fully opaque, and magenta, then jumps to the values defined by the `0%` keyframe selector when the animation keyframes are applied, and jumps back to the original values when the `animation-range-end` is reached, which is 25% from the top of the scrollport. The jumping to the default state is because we did not set the `animation-fill-mode` property on the element, which can be used to apply an animation's styles to an element before and after the animation's execution.
```
chrisdavidmills:
```suggestion
Note how the `from`, or `0%`, keyframe property values are not applied to the animated element until the top block border edge is `10%` past the container's bottom edge; it is full size, fully opaque, and magenta. At that point, the animation is applied and it is styled with the values defined by the `0%` keyframe selector. When the `animation-range-end` is reached, 25% from the top of the scrollport, it jumps back to its original styling.

The jumping to the default state occurs because we did not set the `animation-fill-mode` property on the element, which can be used to apply an animation's styles to an element before and after the animation's execution.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -177,39 +201,15 @@ p {
...
 
-{{EmbedLiveSample("Creating a named view progress timeline with range", "100%", "480px")}}
+Scroll to the end of the content and check the checkbox to add ``animation-fill-mode: both` to the element, which applies the `0%` keyframe before the animation starts and the `100%` keyframe property values after the animation ends (when the range-end is reached).
```
chrisdavidmills:
```suggestion
Scroll to the end of the content and check the checkbox to add `animation-fill-mode: both` to the element, which applies the `0%` keyframe before the animation starts and the `100%` keyframe property values after the animation ends (when the range-end is reached).
```

files/en-us/web/css/reference/properties/animation-range/index.md
chrisdavidmills: Also, the animation behaves like this in Chrome, but it behaves very differently in Firefox. Is this due to a bug, and can you work around it? It should probably be mentioned in some capacity, as the demo just looks broken in Fx

files/en-us/web/css/reference/properties/animation-range/index.md
chrisdavidmills: Maybe include a comment to say that this is most likely the behavior you want, so it is important to set `animation-fill-mode` appropriately.

files/en-us/web/css/reference/properties/animation-range/index.md
estelle: i didn't think it was supported at all in FF, so didn't even look. I don't think any of this is supported in FF

files/en-us/web/css/reference/values/timeline-range-name/index.md
estelle: i literally just copied and pasted from what was already on MDN. LOL.

files/en-us/web/css/reference/properties/animation-range/index.md
chrisdavidmills: I tested it, and it did seem supported to some degree at least.

files/en-us/web/css/reference/properties/animation-range/index.md
estelle: updated

files/en-us/web/css/reference/properties/animation-range/index.md
estelle: added under syntax

files/en-us/web/css/reference/properties/animation-range/index.md
estelle: reorganized the whole section and added headers

files/en-us/web/css/reference/properties/animation-range/index.md
estelle: reorganized with headers

files/en-us/web/css/reference/properties/animation-range/index.md
estelle: FF doesn't support animation-range, it's always animating when in view.

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,87 @@ This property is a shorthand for the following CSS properties:
...
+- `<animation-range-end>`
+  - : The keyword `normal`, a `<length-percentage>`, a `<timeline-range-name>`, or a `<timeline-range-name> <length-percentage>` pair, representing the [`animation-range-end`](/en-US/docs/Web/CSS/animation-range-end). If a `<timeline-range-name>` is set without a `<length-percentage>`, the `<length-percentage>` defaults to `100%`.
 
-### Values
+## Description
+
+The `animation-range` shorthand property sets the `animation-range-start` and `animation-range-end` values, defining where along the animation timeline the animation will start and end. By default, animation keyframe effects are only applied to the elements when the animation is applied to the element based on the `animation-timeline`, meaning the animation is only applied between the range start and range end. To apply the animation outside of this range, set the {{cssxref("animation-fill-mode")}} to `backwards`, `forwards`, or `both`.
```
chrisdavidmills: "By default, animation keyframe effects are only applied to the elements when the animation is applied to the element based on the `animation-timeline`..." - this is really confusing, and I'm not sure what you're trying to say.

Maybe something like "By default, animation styles are only applied to animated elements during the animation..."

files/en-us/web/css/reference/properties/animation-range/index.md
chrisdavidmills: You still don't still what the subsections are covering. After I read "If two values are specified as components of the `<animation-range>` property, they will be interpreted in the order `<animation-range-start>` then `<animation-range-end>`.", I thought "what happens if it's just one value"?

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,87 @@ This property is a shorthand for the following CSS properties:
...
+
+The {{glossary("Scroll container", "scroll port")}} area known as the view progress visibility range is the area inside which the subject element of a [named view progress timeline](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timelines#view_progress_timelines) animation is deemed to be visible. By default, this is the full range of the scrollport, but it can be adjusted using the {{cssxref("view-timeline-inset")}} property.
+
+If two values are specified as components of the `<animation-range>` property, they will be interpreted in the order `<animation-range-start>` then `<animation-range-end>`. The value of each component is either the keyword `normal` or a {{cssxref("length-percentage")}} or a [`<timeline-range-name>`](/en-US/docs/Web/CSS/Reference/Values/timeline-range-name), optionally defined with a `<length-percentage>`. These values are space-separated. Normal is equal to `0%` for start and `100%` for end. Setting `normal` with a `<length-percentage>` for either the start or end range is invalid.
```
chrisdavidmills: "The value of each component is either the keyword `normal` or a {{cssxref("length-percentage")}} or a [`<timeline-range-name>`](/en-US/docs/Web/CSS/Reference/Values/timeline-range-name)" - it can be a `<length-percentage>` _and_ a `<timeline-range-name>`, no? Did you mean and/or?

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -18,71 +18,87 @@ This property is a shorthand for the following CSS properties:
...
+
+### Defining range start and defaulting range end
+
+If the value is a {{cssxref("length-percentage")}} or `normal`, this is interpreted as the value for `<animation-range-start>`. The `<animation-range-end>` value will be implicitly set to `normal`. For example:
+
+- `animation-range: 20%;` is the equivalent to `animation-range-start: 20%; animation-range-end: normal;`
+- `animation-range: normal;` is the equivalent to `animation-range-start: 0%; animation-range-end: 100%;`
+
+If the value is a [`<timeline-range-name>`](/en-US/docs/Web/CSS/Reference/Values/timeline-range-name) without a `<length-percentage>` following it, that timeline range name is applied to the `<animation-range-start>` and `<animation-range-end>` components at `0%` and `100%`, respectively. For example:
```
chrisdavidmills: This sounds like Defining the range start and end. You are setting an explicit start and end `<timeline-range-name>`.

chrisdavidmills: Same for the next one? Maybe put those under a different heading?

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -175,41 +236,26 @@ p {
...
 
-Scroll to see the subject element being animated.
+Scroll to see the element being animated.
+
+{{EmbedLiveSample("Examples", "100%", "480px")}}
+
+Note how the `from`, or `0%`, keyframe property values are not applied to the animated element until the top block border edge is `10%` past the container's bottom edge; it is full size, fully opaque, and magenta. At that point, the animation is applied and it is styled with the values defined by the `0%` keyframe selector. When the `animation-range-end` is reached, 25% from the top of the scrollport, it jumps back to its original styling.
 
-{{EmbedLiveSample("Creating a named view progress timeline with range", "100%", "480px")}}
+The jumping to the default state occurs because we did not set the {{cssxref("animation-fill-mode")}} property on the element, which can be used to apply an animation's styles to an element before and after the animation's execution. We've omitted the property in this example to better enable visualizing the effects of `animation-range`.But, generally, you will want to set `animation-fill-mode: both` when creating [scroll-driven animations](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations).
```
chrisdavidmills: This explanation doesn't make sense. You've mentioned that you'll probably want to set `animation-fill-mode: both`, which is good, but you say that we've omitted it from this example, which is incorrect — you've provided a checkbox to allow readers to compare the results with and without. But you no longer talk about it.

estelle: rearranged it to clarify

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -94,22 +112,16 @@ One or two values representing the [`animation-range-start`](/en-US/docs/Web/CSS
...
-A view progress timeline named `--subject-reveal` is defined using the `view-timeline` property on a subject element with a `class` of `animation`.
-This is then set as the timeline for the same element using `animation-timeline: --subject-reveal;`. The result is that the subject element animates as it moves upwards through the document as it is scrolled.
-
-An `animation-range` declaration is also set to make the animation begin later than expected, and finish earlier.
+In this example, we reduce the duration of the view progress scroll animation by using the `animation-range` property to offset the start and end of the animation, and demonstrate the effect of the {{cssxref("animation-fill-mode")}} property on shortened animation timelines.
 
 #### HTML
 
-The HTML for the example is shown below.
+In the middle of a wall of text, we include an element that we will animate. We added a lot of text in to ensure that our content overflows its container, but this was hidden for the sake of brevity.
```
chrisdavidmills:
```suggestion
In the middle of a wall of text, we include an element that we will animate. We include a lot of text to ensure that our content overflows its container, but this is hidden for the sake of brevity.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -94,22 +112,16 @@ One or two values representing the [`animation-range-start`](/en-US/docs/Web/CSS
...
+The checkbox included to toggle the demonstration of the effect of the {{cssxref("animation-fill-mode")}} property on shortened animation timelines is also hidden.
```
chrisdavidmills:
```suggestion
We also include a checkbox to toggle the {{cssxref("animation-fill-mode")}} property on and off, to demonstrate its effect on shortened animation timelines. This is also hidden.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -128,34 +140,85 @@ The HTML for the example is shown below.
...
+  <input type="checkbox" /> Add <code>animation-fill-mode: both;</code>
+</label>
   </p>
 </div>
 ```
 
 #### CSS
 
-The `subject` element and its containing `content` element are styled minimally, and the text content is given some basic font settings:
+We define an animation that animates an element's opacity, scale, and color, causing it to fade in, size up, and change colors as the animation progresses. We apply this animation to the `animatedElement` with the {{cssxref("animation")}} shorthand.
```
chrisdavidmills:
```suggestion
We define an animation that animates an element's opacity, scale, and background color, causing it to fade in, scale up, and change color as the animation progresses. We apply this animation to the `animatedElement` with the {{cssxref("animation")}} shorthand.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -128,34 +140,85 @@ The HTML for the example is shown below.
...
+
+A view progress timeline is created by setting the [`view()`](/en-US/docs/Web/CSS/Reference/Properties/animation-timeline/view) function as the value of the {{cssxref("animation-timeline")}} property on our `animatedElement`. The result is that the element animates as it moves upwards through the document as it is scrolled. We declared the shorthand before the `animation-timeline`, as the shorthand resets this property.
```
chrisdavidmills:
```suggestion
A view progress timeline is created by setting the [`view()`](/en-US/docs/Web/CSS/Reference/Properties/animation-timeline/view) function as the value of the {{cssxref("animation-timeline")}} property on our `animatedElement`. The result is that the element animates as it moves upwards through the document as it is scrolled. We declare `animation-timeline` property after the shorthand, as the shorthand resets this property.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -128,34 +140,85 @@ The HTML for the example is shown below.
...
+
+Lastly, an `animation-range` declaration is also set to make the animation begin later than expected, and finish earlier.
```
chrisdavidmills:
```suggestion
Lastly, an `animation-range` declaration is set to make the animation begin later than expected and finish earlier.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -128,34 +140,85 @@ The HTML for the example is shown below.
...
+
+We also include conditional styling: when the checkbox is checked, the `animation-fill-mode` property gets applied to the animated element:
+
+```css
+:has(:checked) .animatedElement {
+  animation-fill-mode: both;
+}
+```
+
+The other styles were hidden for the sake of brevity.
```
chrisdavidmills:
```suggestion
The other styles are hidden for the sake of brevity.
```

files/en-us/web/css/reference/properties/animation-range/index.md
```diff
@@ -175,41 +238,26 @@ p {
...
+
+Note how the `from`, or `0%`, keyframe property values are not applied to the animated element until the top block border edge is `10%` past the container's bottom edge; it is full size, fully opaque, and magenta. At that point, the animation is applied and it is styled with the values defined by the `0%` keyframe selector. When the `animation-range-end` is reached, 25% from the top of the scrollport, it jumps back to its original styling.
 
-{{EmbedLiveSample("Creating a named view progress timeline with range", "100%", "480px")}}
+Generally, you will want to set `animation-fill-mode: both` when creating [scroll-driven animations](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations). The jumping to the default state occurs because we did not set the {{cssxref("animation-fill-mode")}} property on the element, which can be used to apply an animation's styles to an element before and after the animation's execution. We initially omitted the property in this example to better enable visualizing the effects of `animation-range`. Check the checkbox to add the `animation-fill-mode` in, then re-scroll to note how the animation animation styles are continuously applied.
```
chrisdavidmills:
```suggestion
Generally, you will want to set `animation-fill-mode: both` when creating [scroll-driven animations](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations). The jump to the default state occurs because we did not set the {{cssxref("animation-fill-mode")}} property on the element, which can be used to apply an animation's styles to an element before and after the animation's execution. We initially omitted the property in this example to better enable visualizing the effects of `animation-range`.

Check the checkbox to apply the `animation-fill-mode` property to the animated element in then re-scroll: the animation styles should now be continuously applied.
```

PR #42034, opened 2025-11-20
Fix typo in description of CSSStyleRule.style

RavuAlHemio:
### Description

Typo fix: `stye` -> `style`

### Motivation

Typos halt the flow of reading (brains often perform a double-take), which is counterproductive when consuming more complex documentation.

### Additional details

n/a

n/a


bsmth:
That's a good one, nice catch! Do you want to get the others:

```
grep -ri "stye" ./files/**/*.md
./files/en-us/web/api/cssstyleproperties/index.md:- {{DOMxRef("CSSStyleRule.style")}}: Used to get and set the styles of a stye rule ({{DOMxRef("CSSStyleRule")}}).
./files/en-us/web/api/htmlelement/style/index.md:However it is not an inline stye for the nested element, so its value is set to the empty string (`""`).
./files/en-us/web/css/guides/anchor_positioning/using/index.md:Let's demonstrate some of these values; this example uses the same HTML and base CSS styes as the previous example, except that we've included a {{htmlelement("select")}} element to enable changing the positioned element's `position-area` value.
```

RavuAlHemio:
Sure thing, here y'go!

PR #42038, opened 2025-11-20
fix typo and improve articulation

prakhar-pal:
### Description

This PR fixes a typo in the img documentation (the changed sentence contained duplicate `more`). I've also improved articulation of that sentence to read more naturally 

### Motivation

Mainly to fix the typo and additionally to improve concept's articulation.


files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -175,7 +175,7 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
 
-    While explicit [`width`](#width) and [`height`](#height) attributes are recommended for all images to avoid layout shift, they are especially important for lazy-loaded ones. Lazy-loaded images will never be loaded if they do not intersect a visible part of an element, even if loading them would change that, because unloaded images have a `width` and `height` of `0`. It is also more a more disturbing user experience if the page reflows in the middle of reading.
+    While explicit [`width`](#width) and [`height`](#height) attributes are recommended for all images to avoid layout shift, they are especially important for lazy-loaded ones. Lazy-loaded images will never be loaded if they do not intersect a visible part of an element, even if loading them would change that, because unloaded images have a `width` and `height` of `0`. It also creates a more disruptive user experience when the page reflows in the middle of reading.
```
```suggestion
    While explicit [`width`](#width) and [`height`](#height) attributes are recommended for all images to avoid layout shift, they are especially important for lazy-loaded ones. Lazy-loaded images will never be loaded if they do not intersect a visible part of an element, even if loading them would change that, because unloaded images have a `width` and `height` of `0`. It creates an even more disruptive user experience when the content visible in the viewport reflows in the middle of reading it.
```

chrisdavidmills: What you've done so far is good, but I wonder if the sentence should be modified further to something like this. The page already mentions setting `width`/`height` to avoid layout shift, and that this is important, in a note inside the `height` description. So the existing text here seems a bit repetitive. I think it is supposed to say that the layout shift is particularly bad when you don't set `width`/`height` on lazy-loaded images because the shift can happen to the currently-visible content.

prakhar-pal: Yes, agreed, this makes it clearer, will update the text.

prakhar-pal: pushed this change @chrisdavidmills

chrisdavidmills (approved):
Cool, let's get this merged. Thanks again, @prakhar-pal!

Issue #41357, opened 2025-10-01
WebSocket guide uses `pageshow` event handler


bsmth:
> I think that this demo is using the `pageshow` event handler to both initialize the WebSocket connection at initial load as well as restoring the same logic for BFCache restore.
> 
> I believe most developers will not use the pageshow event handler this way.
> 
> +1 to change this to initialize WebSocket normally on pageload and run the same logic on pageshow event.persisted=true to make it work again for BFCache. 

 _Originally posted by @gilbertococchi in [#40931](https://github.com/mdn/content/pull/40931/files/21592f9756a9e38bafa3e11201e391136ceba34d..f15529db65e4abd362907808fe703e08f22ced13#r2393793650)_

wbamberg:
In case it's missed: https://github.com/mdn/content/pull/40931#discussion_r2395029953 and https://github.com/mdn/content/pull/40931#discussion_r2395197143.

And also https://github.com/mdn/content/pull/40931#discussion_r2316763023 where this was discussed in the original review.

fergald:
To address the question of why it's not realistic to open the `WebSocket` in `pageshow`...

[`pageshow`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event) occurs after [`load`](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event).

> The load event is fired when the whole page has loaded, including all dependent resources such as stylesheets, scripts (including async, deferred, and module scripts), iframes, and images, except those that are [loaded lazily](https://developer.mozilla.org/en-US/docs/Web/Performance/Guides/Lazy_loading#images_and_iframes). This is in contrast to [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event), which is fired as soon as the page DOM has been loaded, without waiting for resources to finish loading.

I would imagine anyone installed e.g. a chat widget that should be active from the start of the page, wants that widget to be active without having to wait for all of the above.

More fundamantally, if I was writing a library to do something with `WebSocket`s it would create the `WebSocket` when someone calls `library.StartDoingStuff()`. It wouldn't install a `pageshow` handler to create the `WebSocket` later on. The library has no idea if `pageshow` is even going to be fired. Maybe the page loaded 5 minutes ago and the user has just now clicked on "chat with customer service" or whatever.

gilbertococchi:
+1 to what Fergal said.

The change is relatively trivial:

[...]

PR #42044, opened 2025-11-21
chore(ws): Improve reconnection logic for bfcache example

bsmth:
### Description

This changes the example to use a more robust initializer and only use `pageshow` for reconnection.

Note that this would require updates to https://github.com/mdn/dom-examples/tree/main/websockets if accepted.

### Motivation

> this demo is using the `pageshow` event handler to both initialize the WebSocket connection at initial load as well as restoring the same logic for BFCache restore. 
> I believe most developers will not use the pageshow event handler this way.
> +1 to change this to initialize WebSocket normally on pageload and run the same logic on pageshow event.persisted=true to make it work again for BFCache.

See https://github.com/mdn/content/issues/41357#issuecomment-3562205750

### Additional details

Fixes https://github.com/mdn/content/issues/41357

- [x] https://github.com/mdn/content/pull/40931

CC @gilbertococchi

### DIFF
--- a/files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
+++ b/files/en-us/web/api/websockets_api/writing_websocket_client_applications/index.md
@@ -148,38 +148,45 @@ window.addEventListener("pagehide", () => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

gilbertococchi:
Thank you @bsmth!

bsmth:
Thank you!

PR #42045, opened 2025-11-21
Reorder CSS Grid guides in the sidebar

laurentbardin:
### Description

This commit properly reorders two out-of-order CSS Grid guides in the sidebar according to the layout presented in the CSS Grid guides index.

### Motivation

While following the CSS Grid guides, I realised that they probably weren't listed in the correct order in the sidebar because, as I reached the [Grid template areas](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Grid_layout/Grid_template_areas) section, the `grid-area` property was described as "already encountered" when it wasn't. After taking a look at the [guides index](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Grid_layout#guides), there was inded a discrepancy between the two lists, and I determined that the guides index was the correct one, hence this fix and PR.

As a bonus, the [Grid template areas](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Grid_layout/Grid_template_areas) guide contained a an extraneous link to the next guide at the end, which I took the opportunity of removing.

### Additional details

No additional details I can think of.

I did not find any related issue or PR.

### DIFF
--- a/files/sidebars/cssref.yaml
+++ b/files/sidebars/cssref.yaml
@@ -196,9 +196,9 @@ sidebar:
       - /Web/CSS/Guides/Grid_layout/Auto-placement
       - /Web/CSS/Guides/Grid_layout/Box_alignment
       - /Web/CSS/Guides/Grid_layout/Logical_values_and_writing_modes
+      - /Web/CSS/Guides/Grid_layout/Accessibility
       - /Web/CSS/Guides/Grid_layout/Common_grid_layouts
       - /Web/CSS/Guides/Grid_layout/Subgrid
-      - /Web/CSS/Guides/Grid_layout/Accessibility
       - /Web/CSS/Guides/Grid_layout/Masonry_layout
   - title: Images
     details: closed


files/sidebars/cssref.yaml
```diff
@@ -190,8 +190,8 @@ sidebar:
     children:
       - /Web/CSS/Guides/Grid_layout/Basic_concepts
       - /Web/CSS/Guides/Grid_layout/Relationship_with_other_layout_methods
-      - /Web/CSS/Guides/Grid_layout/Grid_template_areas
       - /Web/CSS/Guides/Grid_layout/Line-based_placement
+      - /Web/CSS/Guides/Grid_layout/Grid_template_areas
       - /Web/CSS/Guides/Grid_layout/Named_grid_lines
       - /Web/CSS/Guides/Grid_layout/Auto-placement
       - /Web/CSS/Guides/Grid_layout/Box_alignment
```
chrisdavidmills, line 197: If we are following the order set out in the guides index,

```
- /Web/CSS/Guides/Grid_layout/Accessibility
```

Should be underneath

```
- /Web/CSS/Guides/Grid_layout/Logical_values_and_writing_modes
```

In the sidebar.

files/en-us/web/css/guides/grid_layout/grid_template_areas/index.md
```diff
@@ -484,4 +484,4 @@ We will revisit the other functionality offered by this shorthand when we take a
 
 ## Next steps
 
-If you've been working along the [grid guides](/en-US/docs/Web/CSS/Guides/Grid_layout#guides), you should be in a position to create grid layouts using [line-based placement](/en-US/docs/Web/CSS/Guides/Grid_layout/Line-based_placement) or [named areas](/en-US/docs/Web/CSS/Guides/Grid_layout/Named_grid_lines). Now let's take a look at creating [grid layouts using named grid lines](/en-US/docs/Web/CSS/Guides/Grid_layout/Named_grid_lines).
+If you've been working along the [grid guides](/en-US/docs/Web/CSS/Guides/Grid_layout#guides), you should be in a position to create grid layouts using [line-based placement](/en-US/docs/Web/CSS/Guides/Grid_layout/Line-based_placement) or named areas. Now let's take a look at creating [grid layouts using named grid lines](/en-US/docs/Web/CSS/Guides/Grid_layout/Named_grid_lines).
```
chrisdavidmills:
```suggestion
If you've been working along the [grid guides](/en-US/docs/Web/CSS/Guides/Grid_layout#guides), you should be in a position to create grid layouts using [line-based placement](/en-US/docs/Web/CSS/Guides/Grid_layout/Line-based_placement) or named template areas. Now let's take a look at creating [grid layouts using named grid lines](/en-US/docs/Web/CSS/Guides/Grid_layout/Named_grid_lines).
```

chrisdavidmills: for consistency with how it's referred to in the article.

chrisdavidmills (changes requested):
Good spot, thanks for the fixes, @laurentbardin!

I think what you have done is fine, but I've got a couple of suggestions for further small fixes we could do while we are here.

files/sidebars/cssref.yaml
laurentbardin: Indeed, I didn't realise there was a difference toward the end of the list as well (because I haven't reached it yet, probably). I'll fix that. 👍

PR #42053, opened 2025-11-22
Add close  button in htmldialogelement/show/index.md example

leevigraham:
### Description

Example was missing cancel button. Added a close button matching the /en-US/docs/Web/API/HTMLDialogElement/close example

### Motivation

Fixing example.

### Additional details

N/A

N/A

### DIFF
--- a/.github/workflows/auto-cleanup-bot.yml
+++ b/.github/workflows/auto-cleanup-bot.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           persist-credentials: false
 
@@ -45,7 +45,7 @@ jobs:
 
       - name: Create PR with only fixable issues
         if: success()
-        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        uses: peter-evans/create-pull-request@84ae59a2cdc2258d6fa0732dd66352dddae2a412 # v7.0.9
         with:
           commit-message: "chore: auto-fix Markdownlint, Prettier, and front-matter issues"
           branch: markdownlint-auto-cleanup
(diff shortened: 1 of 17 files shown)


files/en-us/web/api/htmldialogelement/show/index.md
```diff
@@ -43,6 +43,7 @@ button.
 <!-- Simple pop-up dialog box, containing a form -->
 <dialog id="favDialog">
   <form method="dialog">
+    <button type="button" id="close" aria-label="close">X</button>
```
hamishwillee: Thank you. Can you change this to be "cancel" and then revert the other changes.

The main reason being that the text before this refers to it as a cancel button, and this is a minimal change to meet the original intent.

leevigraham: Done. I explicitly set the button label to "Cancel" rather than "X" to show clearer intent.

leevigraham: Was the original intent to have the Cancel button be the reset button?

```html
<button type="reset">Reset</button>
```

hamishwillee: @leevigraham Thanks! To make this easier to use and test the result, I made this a live example. As part of that I reverted the "X" button and updated the text so that it is clear that "Cancel" and X are the same thing - it just looks better. The extra heading levels are how API examples are supposed to be written according to the current web templates. I also modified the layout a little to kill the arbitrary line breaks.

> Was the original intent to have the Cancel button be the reset button?

No, the reset button clears the selection in the dialog. We could talk about this but not sure it is relevant to the specific example. Do you think it would be better to delete that button?

Otherwise I am happy to merge this - just wanted to give you a chance to comment.

leevigraham: Given this was my first PR I didn't want to overstep with the live example…

Now I've seen a live example and reviewed the other Instance Method docs I've found some more inconsistencies:

* https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/close - Looks good
* https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/requestClose - Could add notes that the `dialog.requestClose` method triggers the `cancel` event
* https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/show - needs live example (this PR)
* https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal - Has cancel button… but it's the reset button.

I'm happy to attempt to create a PR that update all the Instance Method docs so the live examples all follow the same layout with javascript implementations reflecting the instance method

hamishwillee: That would be welcome. I'll merge this one though.

Note that you might want to link back to this PR to demonstrate the consistency you are trying to achieve. Reason being that I might not be the reviewer.

leevigraham: I'll give it a shot :)

leevigraham:
@hamishwillee I've created a new PR for consolidating the dialog examples if you're interested: https://github.com/mdn/content/pull/42127

PR #42057, opened 2025-11-22
Refined punctuation and wording in the "Inheriting properties" section to remove ambiguity

DanBoicea:
### Description

Updated the "Inheriting properties" section in the prototype chain explanation. Added the word "also" and refined punctuation to make the description clearer and reduce ambiguity.

### Motivation

The original text could be misread as implying that property lookup on the object and its prototype chain are mutually exclusive. This change clarifies that the search is cumulative for readers learning about JavaScript inheritance.

### DIFF
--- a/files/en-us/web/javascript/guide/inheritance_and_the_prototype_chain/index.md
+++ b/files/en-us/web/javascript/guide/inheritance_and_the_prototype_chain/index.md
@@ -15,7 +15,7 @@ Although classes are now widely adopted and have become a new paradigm in JavaSc
 
 ### Inheriting properties
 
-JavaScript objects are dynamic "bags" of properties (referred to as **own properties**). JavaScript objects have a link to a prototype object. When trying to access a property of an object, the property will not only be sought on the object, but also on the prototype of the object, the prototype of the prototype and so on, until either a property with a matching name is found or the end of the prototype chain is reached.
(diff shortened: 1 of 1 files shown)


files/en-us/web/javascript/guide/inheritance_and_the_prototype_chain/index.md
```diff
@@ -15,7 +15,7 @@ Although classes are now widely adopted and have become a new paradigm in JavaSc
 
 ### Inheriting properties
 
-JavaScript objects are dynamic "bags" of properties (referred to as **own properties**). JavaScript objects have a link to a prototype object. When trying to access a property of an object, the property will not only be sought on the object but on the prototype of the object, the prototype of the prototype, and so on until either a property with a matching name is found or the end of the prototype chain is reached.
+JavaScript objects are dynamic "bags" of properties (referred to as **own properties**). JavaScript objects have a link to a prototype object. When trying to access a property of an object, the property will not only be sought on the object, but also on the prototype of the object, the prototype of the prototype and so on, until either a property with a matching name is found or the end of the prototype chain is reached.
```
Josh-Cena:
```suggestion
JavaScript objects are dynamic "bags" of properties (referred to as **own properties**). JavaScript objects have a link to a prototype object. When trying to access a property of an object, the property will not only be sought on the object, but also on the prototype of the object, the prototype of the prototype, and so on, until either a property with a matching name is found or the end of the prototype chain is reached.
```

Josh-Cena (approved):
Nice, this reads better, thank you

Issue #41683, opened 2025-10-26
The directives in multipart's Content-Disposition are not required to be quoted in RFCs


Menci:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#as_a_header_for_a_multipart_body

### What specific section or headline is this issue about?

As a header for a multipart body

### What information was incorrect, unhelpful, or incomplete?

According to [MDN's documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#as_a_header_for_a_multipart_body), the values must be quoted. But I can't find the corresponding statements in RFC (both newer [RFC 7578](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) and older [RFC 2388](https://www.rfc-editor.org/rfc/rfc2388#section-3)).

Since [RFC 7578](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) doesn't have a declaration for the Content-Disposition header, [RFC 6266](https://httpwg.org/specs/rfc6266.html#n-grammar) applies:

[4.1.](https://httpwg.org/specs/rfc6266.html#rfc.section.4.1) [Grammar](https://httpwg.org/specs/rfc6266.html#n-grammar)

<pre>
  content-disposition = "Content-Disposition" ":"
                         disposition-type *( ";" disposition-parm )

  disposition-type    = "inline" | "attachment" | disp-ext-type
                      ; case-insensitive
  disp-ext-type       = token

  disposition-parm    = filename-parm | disp-ext-parm

  filename-parm       = "filename" "=" value
                      | "filename*" "=" ext-value
  
  disp-ext-parm       = token "=" value
                      | ext-token "=" ext-value
  ext-token           = &lt;the characters in token, followed by "*">
</pre>

Defined in [[RFC2616]](https://httpwg.org/specs/rfc6266.html#RFC2616):

<pre>
  token         = &lt;token, defined in [[RFC2616]](https://httpwg.org/specs/rfc6266.html#RFC2616), [Section 2.2](https://www.rfc-editor.org/rfc/rfc2616.html#section-2.2)>
  quoted-string = &lt;quoted-string, defined in [[RFC2616]](https://httpwg.org/specs/rfc6266.html#RFC2616), [Section 2.2](https://www.rfc-editor.org/rfc/rfc2616.html#section-2.2)>
[...]

PR #42065, opened 2025-11-23
docs(headers): clarify quotes requirement in Content-Disposition of multipart/form-data

Menci:
### Description

Clarify that quotes in Content-Disposition's directives are not required in the RFC standard but are encouraged for compatibility.

### Motivation

The current content says the quotes in directives of multipart/form-data's Content-Disposition are required, but it's NOT required in the RFC standard. See the issue linked below for more details.

### Additional details

https://www.rfc-editor.org/rfc/rfc7578#section-4.2
https://httpwg.org/specs/rfc6266.html#n-grammar

https://github.com/mdn/content/issues/41683

### DIFF
--- a/files/en-us/web/http/reference/headers/content-disposition/index.md
+++ b/files/en-us/web/http/reference/headers/content-disposition/index.md
@@ -57,7 +57,9 @@ Browsers may apply transformations to conform to the file system requirements, s
 
 A `multipart/form-data` body requires a `Content-Disposition` header to provide information about each subpart of the form (e.g., for every form field and any files that are part of field data).
 The first directive is always `form-data`, and the header must also include a `name` parameter to identify the relevant field. Additional directives are case-insensitive.
-Their arguments usually use quoted-string syntax after the `=` sign; although the standard allows unquoted tokens here, many server implementations expect the values to be quoted.
+The value of any arguments (after the `=` sign) may be a either token or a quoted string.
+Quoted strings are recommended, and many server implementations require the values to be quoted.
+This is because a token must be US-ASCII for MIME type headers like `Content-Disposition`, and US-ASCII does not allow some characters that are common in filenames and other values.
 Multiple parameters are separated by a semicolon (`;`).
 
 ```http


files/en-us/web/http/reference/headers/content-disposition/index.md
```diff
@@ -55,7 +55,7 @@ Browsers may apply transformations to conform to the file system requirements, s
 
 ### As a header for a multipart body
 
-A `multipart/form-data` body requires a `Content-Disposition` header to provide information about each subpart of the form (e.g., for every form field and any files that are part of field data). The first directive is always `form-data`, and the header must also include a `name` parameter to identify the relevant field. Additional directives are case-insensitive and have arguments that use quoted-string syntax after the `=` sign. Multiple parameters are separated by a semicolon (`;`).
+A `multipart/form-data` body requires a `Content-Disposition` header to provide information about each subpart of the form (e.g., for every form field and any files that are part of field data). The first directive is always `form-data`, and the header must also include a `name` parameter to identify the relevant field. Additional directives are case-insensitive. Their arguments usually use quoted-string syntax after the `=` sign; although the standard allows unquoted tokens here, many server implementations expect the values to be quoted. Multiple parameters are separated by a semicolon (`;`).
```
hamishwillee: Splitting into line "to my taste" and to make the comment I am going to make more obvious.
```suggestion
A `multipart/form-data` body requires a `Content-Disposition` header to provide information about each subpart of the form (e.g., for every form field and any files that are part of field data).
The first directive is always `form-data`, and the header must also include a `name` parameter to identify the relevant field. Additional directives are case-insensitive.
Their arguments usually use quoted-string syntax after the `=` sign; although the standard allows unquoted tokens here, many server implementations expect the values to be quoted.
Multiple parameters are separated by a semicolon (`;`).
```

files/en-us/web/http/reference/headers/content-disposition/index.md
```diff
@@ -55,7 +55,10 @@ Browsers may apply transformations to conform to the file system requirements, s
 
 ### As a header for a multipart body
 
-A `multipart/form-data` body requires a `Content-Disposition` header to provide information about each subpart of the form (e.g., for every form field and any files that are part of field data). The first directive is always `form-data`, and the header must also include a `name` parameter to identify the relevant field. Additional directives are case-insensitive and have arguments that use quoted-string syntax after the `=` sign. Multiple parameters are separated by a semicolon (`;`).
+A `multipart/form-data` body requires a `Content-Disposition` header to provide information about each subpart of the form (e.g., for every form field and any files that are part of field data).
+The first directive is always `form-data`, and the header must also include a `name` parameter to identify the relevant field. Additional directives are case-insensitive.
+Their arguments usually use quoted-string syntax after the `=` sign; although the standard allows unquoted tokens here, many server implementations expect the values to be quoted.
```
hamishwillee: I want to make it clear what set is allowed, mostly so people can debug why their string might not work, and so this does not come up again. Make sense?

```suggestion
The value of any arguments (after the `=` sign) may be a either token or a quoted string.
Quoted strings are recommended, and many server implementations require the values to be quoted.
This is because a token must be US-ASCII for MIME type headers like `Content-Disposition`, and US-ASCII does not allow some characters that are common in filenames and other values.
```

hamishwillee (approved):

hamishwillee:
Thanks @Menci

PR #42069, opened 2025-11-24
Ff146 css math shift

hamishwillee:
FF146 adds support for the MathML CSS [`math-shift`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/math-shift) property.

This adds adds a release note and also fixes up the example which wasn't working properly.

Related docs work can be tracked in #41877


files/en-us/web/css/reference/properties/math-shift/index.md
```diff
@@ -44,29 +44,29 @@ math-shift: unset;
 
 ## Examples
 
-### CSS
+### Normal and compact shift
```
hamishwillee, line 43: The example wasn't bundled together under headings correctly. It is now. I also decided to make the normal shift apply at the top level and remove the specific selector.

bsmth (approved):
Thanks, Hamish, looking good 👍🏻

hamishwillee:
THanks for the review @bsmth

PR #42042, opened 2025-11-21
WorkerGlobalScope.setInterval() - update for trusted types

hamishwillee:
This updates `WorkerGlobalScope.setInterval()` for trusted types.

This is virtually the same as  `Window.setInterval()` which has a lot of description. To reduce duplication I link to those docs for the description and examples (which this also did previously).

I added the Trusted type header but in it also link to `Window.setInterval()` for the security considerations.
Upshot, this highlights the main issue using the header but mostly users will need to go to the Window method for detailed information. I think this is the right level of docs.

Related work tracked in #41507

PR #42072, opened 2025-11-24
setTimeout() on Window/Worker plus consistency fixes

hamishwillee:
This updates `Window.setTimeout()` and `WorkerGlobalScope.setTimeout()` for trusted types, following the same approach as https://github.com/mdn/content/pull/42042
- More or less full docs in the windows method. There is a page header warning in the top and a security considerations section in the body.
- worker method has same top section, but description and examples are cross linked to window method docs. The header links to the security considerations in the window method topic.
- Examples do not cover the use of TT. We have pretty strong warnings up the top about using this feature at all, so I don't see we need to show TT use. I have linked to the `eval()` example on this added in #42462

This also did some consistency fixes across both this and the `setInterval` methods so that they have the same name for the function param (`func`) and the arguments are referred to as `param` rather than `args` - which is consistent with the template for web apis.

There were a few layout change and improvements to the docs - these map the similar changes done previously for `setInterval()`

This is part of project being tracked in #41507

### DIFF
--- a/files/en-us/web/api/window/settimeout/index.md
+++ b/files/en-us/web/api/window/settimeout/index.md
@@ -128,12 +128,18 @@ myObject.log(); // myProperty: 12
 setTimeout(myObject.log, 1000); // myProperty: undefined
 ```
 
-You can use [arrow functions](/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) to adopt the `this` of the function in which `setTimeout()` is called (arrow functions have a lexical `this`).
+You can use a wrapper function, such as an [arrow function](/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), to adopt the `this` of the function in which `setTimeout()` is called (arrow functions have a lexical `this`).
 
-You can test this with the following code.
+You can test this with the following code:
 
 ```js
+// Arrow function callback
 setTimeout(() => myObject.log(), 2000); // myProperty: 12 after 2 seconds
+
+// Anonymous function callback
+setTimeout(function () {
+  myObject.log();
+}, 3000); // myProperty: 12 after 3 seconds
 ```
 
 You might also use the [`Function.prototype.bind()`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) method, which lets you specify the value that should be used as `this` for all calls to a given function.


files/en-us/web/api/window/settimeout/index.md
```diff
@@ -16,76 +23,63 @@ The **`setTimeout()`** method of the {{domxref("Window")}} interface sets a time
...
-  console.log("Delayed for 1 second.");
-}, 1000);
-```
+If you wish to call a function repeatedly (e.g., every _N_ milliseconds), you can use {{domxref("Window.setInterval", "setInterval()")}}.
 
 ### Working with asynchronous functions
 
-`setTimeout()` is an asynchronous function, meaning that the timer function will not pause execution of other functions in the functions stack.
-In other words, you cannot use `setTimeout()` to create a "pause" before the next function in the function stack fires.
+`setTimeout()` is an asynchronous function, meaning that the timer function will not pause execution of other functions in the function's stack.
```
Josh-Cena: Original is more accurate; there's only one stack, which contains "functions". Perhaps "call stack" would be less confusing, but both are technically wrong, because the example doesn't demonstrate a "stack" at all; I just see a sequence of function calls. Really, this just means "will not pause execution of subsequent code".

files/en-us/web/api/window/settimeout/index.md
```diff
@@ -16,76 +23,63 @@ The **`setTimeout()`** method of the {{domxref("Window")}} interface sets a time
...
+In other words, you cannot use `setTimeout()` to create a "pause" before the next function in the stack fires.
```
hamishwillee:
```suggestion
`setTimeout()` is an asynchronous function, meaning that it completes immediately, and does not block the execution of the following code.
In other words, you cannot use `setTimeout()` to create a "pause" before the next line is executed.
```

files/en-us/web/api/window/settimeout/index.md
hamishwillee: Thanks @Josh-Cena . How about https://github.com/mdn/content/pull/42072/files#r2557944117

files/en-us/web/api/window/settimeout/index.md
Josh-Cena: I wouldn't say "next function in the stack is executed", because `setTimeout` enters a stack frame and immediately exits it, meaning you are still in the same function. Perhaps "before the next line is executed".

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -76,8 +76,7 @@ Browsers may enforce even more stringent minimum values for the interval under s
 Note also that the actual amount of time that elapses between calls to the callback may be longer than the given `delay`; see [Reasons for delays longer than specified](/en-US/docs/Web/API/Window/setTimeout#reasons_for_delays_longer_than_specified) for examples.
```
hamishwillee:
```suggestion
Note also that the actual amount of time that elapses between calls to the callback may be longer than the given `delay`; see [Reasons for delays longer than specified](/en-US/docs/Web/API/Window/setTimeout#reasons_for_longer_delays_than_specified) for examples.
```

hamishwillee:
@pepelsbey Can you please review?

hamishwillee:
@wbamberg Would you also/instead consider reviewing this. I ask because it is for TrustedTypes, and we've both been looking at all of those.

wbamberg:
I'm happy to look at this, yes, but it will take me a few days until I can find the time.

files/en-us/web/api/workerglobalscope/settimeout/index.md
```diff
@@ -16,39 +23,51 @@ The **`setTimeout()`** method of the {{domxref("WorkerGlobalScope")}} interface
...
-    is used, meaning execute "immediately", or more accurately, the next event cycle.
-
-    Note that in either case, the actual delay may be longer than intended; see [Reasons for delays longer than specified](/en-US/docs/Web/API/Window/setTimeout#reasons_for_delays_longer_than_specified).
-
-    Also note that if the value isn't a number, implicit {{glossary("type coercion")}} is silently done on the value to convert it to a number — which can lead to unexpected and surprising results; see [Non-number delay values are silently coerced into numbers](/en-US/docs/Web/API/Window/setTimeout#non-number_delay_values_are_silently_coerced_into_numbers) for an example.
+  - : The time that the timer should wait before the specified function or code is executed, in milliseconds.
+    Defaults to 0 if not specified.
+
+    Note:
+    - The delay has a maximum value of 2147483647 ms — specifying larger values may result in overflow or a 0 value being used.
```
pepelsbey: I wonder if we could be a bit more useful here. Is it defined in the spec? Does it depend on browser implementation? It sounds a bit like “who knows what’s going to happen?”

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -150,24 +149,38 @@ That lets you bypass problems where it's unclear what `this` will be, depending
...
+
+For example, the following code shows how `setInterval()` might execute `untrustedCode` provided by a user:
 
 ```js example-bad
 const untrustedCode = "alert('Potentially evil code!');";
 const id = setInterval(untrustedCode, 1000);
 ```
 
-Websites with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) will prevent such code running by default; if you need to use the method with `code` then you will first need to allow the [`unsafe-eval`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#unsafe-eval) in your CSP [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src).
+Websites with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that specifies [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src) will prevent such code running by default.
```
wbamberg:
```suggestion
Websites with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that specifies [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src) or [`default-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src) will prevent such code running by default.
```

files/en-us/web/api/window/setinterval/index.md
```diff
@@ -150,24 +149,38 @@ That lets you bypass problems where it's unclear what `this` will be, depending
...
+You can specify [`unsafe-eval`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#unsafe-eval) in your CSP to allow `setInterval()` to execute, but this is unsafe as it disables one of the main protections of CSP.
```
wbamberg, line 161: if you like:

```suggestion
You can specify [`unsafe-eval`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#unsafe-eval) in your CSP to allow `setInterval()` to execute, but this is unsafe as it disables one of the main protections of CSP.

See [Inline JavaScript](/en-US/docs/Web/HTTP/Guides/CSP#inline_javascript) in the CSP guide.
```

files/en-us/web/api/window/settimeout/index.md
```diff
@@ -16,76 +23,63 @@ The **`setTimeout()`** method of the {{domxref("Window")}} interface sets a time
...
-  console.log("Delayed for 1 second.");
-}, 1000);
-```
+If you wish to call a function repeatedly (e.g., every _N_ milliseconds), you can use {{domxref("Window.setInterval", "setInterval()")}}.
 
 ### Working with asynchronous functions
 
-`setTimeout()` is an asynchronous function, meaning that the timer function will not pause execution of other functions in the functions stack.
-In other words, you cannot use `setTimeout()` to create a "pause" before the next function in the function stack fires.
+`setTimeout()` is an asynchronous function, meaning that it completes immediately, and does not block the execution of the following code.
```
wbamberg:
```suggestion
`setTimeout()` is an asynchronous function, meaning that it completes immediately, and does not block the execution of the code that follows it.
```

maybe? "the following code" makes it seem like this text is going to be followed by an example showing some code that isn't blocked.

files/en-us/web/api/window/settimeout/index.md
```diff
@@ -16,76 +23,63 @@ The **`setTimeout()`** method of the {{domxref("Window")}} interface sets a time
...
-But in many cases, the implicit type coercion can lead to unexpected and surprising results. For example, when the following code runs, the string `"1 second"` ultimately gets coerced into the number `0` — and so, the code executes immediately, with zero delay.
+## Description
 
-```js example-bad
-setTimeout(() => {
-  console.log("Delayed for 1 second.");
-}, "1 second");
-```
+The `setTimeout()` function is commonly used to call a function that is executed just once, after a delay.
+The timeout can be cancelled before it completes using {{domxref("Window.clearTimeout()")}}.
```
wbamberg:
```suggestion
You can call {{domxref("Window.clearTimeout()")}} to cancel the timeout before it completes.
```
"The timeout can be cancelled before it completes using {{domxref("Window.clearTimeout()")}}." is ambiguous - do I cancel it using clearTimeout, or does it complete using clearTimeout?

files/en-us/web/api/window/settimeout/index.md
```diff
@@ -107,90 +101,48 @@ setTimeout(() => {
...
-Notice that the first function does not create a 5-second "pause" before calling the second function. Instead, the first function is called, but waits 5 seconds to
-execute. While the first function is waiting to execute, the second function is called, and a 3-second wait is applied to the second function before it executes. Since neither
-the first nor the second function's timers have completed, the third function is called and completes its execution first. Then the second follows. Then finally the first function
-is executed after its timer finally completes.
+Notice that the first function does not create a 5-second "pause" before calling the second function.
+Instead, the first function is called, but waits 5 seconds to execute.
+While the first function is waiting to execute, the second function is called, and a 3-second wait is applied to the second function before it executes.
+Since neither the first nor the second function's timers have completed, the third function is called and completes its execution first.
+Then the second follows.
+Then finally the first function is executed after its timer finally completes.
```
wbamberg: Is it me or is this explanation very confusing? It doesn't seem to distinguish the setTimeout function and the callback argument, when it talks about "functions". Like:

> Instead, the first function is called, but waits 5 seconds to execute.

Surely the first function (i.e. the first setTimeout` call) is called, and executes, and returns. It's not waiting to execute (itself). It's scheduled the callback to run after 5 seconds.

> While the first function is waiting to execute, the second function is called, and a 3-second wait is applied to the second function before it executes.

What is "it" here? The second function? Like, the second `setTimeout`? But that has executed, it's the callback that hasn't.

And:

> Since neither the first nor the second function's timers have completed, the third function is called and completes its execution first.

...implies that the third function is only called if the first two haven't completed ("since...the third function is called"), but that's completely unrelated.

files/en-us/web/api/window/settimeout/index.md
```diff
@@ -107,90 +101,48 @@ setTimeout(() => {
...
-set `this` in the call or with `bind`, it will default to
-the `window` (or `global`) object, even in [strict mode](/en-US/docs/Web/JavaScript/Reference/Strict_mode). It will not be the same as the
-`this` value for the function that called `setTimeout`.
-
-See the following example:
+The following example demonstrates how this can cause unexpected behavior:
 
 ```js
-const myArray = ["zero", "one", "two"];
+myArray = ["zero", "one", "two"];
```
wbamberg:
```suggestion
const myArray = ["zero", "one", "two"];
```

files/en-us/web/api/window/settimeout/index.md
```diff
@@ -107,90 +101,48 @@ setTimeout(() => {
...
+
 myArray.myMethod = function (sProperty) {
-  console.log(arguments.length > 0 ? this[sProperty] : this);
+  alert(arguments.length > 0 ? this[sProperty] : this);
 };
 
 myArray.myMethod(); // prints "zero,one,two"
 myArray.myMethod(1); // prints "one"
+setTimeout(myArray.myMethod, 1000); // Alerts "[object Window]" after 1 second
+setTimeout(myArray.myMethod, 1500, "1"); // Alerts "undefined" after 1.5 seconds
```
wbamberg: So maybe I've just eaten too much tofurkey, but this seems like a very complex example to illustrate a relatively simple thing. Why can't we demonstrate using something like:

```
const myObject = {
	log() {
    	console.log(`myProperty: ${this.myProperty}`)
	},
	myProperty: 12,
}

myObject.log();
// myProperty: 12

setTimeout(myObject.log, 1000);
// myProperty: undefined
```

files/en-us/web/api/window/settimeout/index.md
```diff
@@ -107,90 +101,48 @@ setTimeout(() => {
...
-
-```js
-setTimeout(myArray.myMethod, 1.0 * 1000); // prints "[object Window]" after 1 second
-setTimeout(myArray.myMethod, 1.5 * 1000, "1"); // prints "undefined" after 1.5 seconds
-```
-
-The `myArray.myMethod` function is passed to `setTimeout`, then
-when it's called, its `this` is not set, so it defaults to the
-`window` object.
+You can use [arrow functions](/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) to adopt the `this` of the function in which `setTimeout()` is called (arrow functions have a lexical `this`).
```
wbamberg: So I might be confused here again, but AFAIK this is about using a wrapper, in general, not using an arrow function in particular. That is, you can do the same thing with this:

```js
setTimeout(function() {myArray.myMethod()}, 1000);
```

(that's also what the original says: https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#use_a_wrapper_function)

files/en-us/web/api/window/settimeout/index.md
```diff
@@ -368,53 +344,58 @@ After setTimeout
...
+For example, the following code shows how `setTimeout()` might execute `untrustedCode` provided by a user:
+
+```js example-bad
+const untrustedCode = "alert('Potentially evil code!');";
+const id = setTimeout(untrustedCode, 1000);
 ```
 
-…results in the timeout being executed immediately (since `2**32 - 5000` overflows to a negative number), while the following code:
+Websites with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that specifies [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src) will prevent such code running by default.
+You can specify [`unsafe-eval`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#unsafe-eval) in your CSP to allow `setTimeout()` to execute, but this is unsafe as it disables one of the main protections of CSP.
```
wbamberg: same comments as above

wbamberg (changes requested):

files/en-us/web/api/window/setinterval/index.md
hamishwillee: Yes I do :-)

files/en-us/web/api/window/settimeout/index.md
hamishwillee:
```suggestion
Websites with a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that specifies [`script-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src) or [`default-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src) will prevent such code running by default.
You can specify [`unsafe-eval`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#unsafe-eval) in your CSP to allow `setTimeout()` to execute, but this is unsafe as it disables one of the main protections of CSP.
See [Inline JavaScript](/en-US/docs/Web/HTTP/Guides/CSP#inline_javascript) in the CSP guide.
```

files/en-us/web/api/window/settimeout/index.md
hamishwillee: Yes. I didn't think it worth the effort to tidy it. Arguably in reference, and at this point in JavaScript you shouldn't even need such an example and comment on what asynchronous means.

Fixed in https://github.com/mdn/content/pull/42072/commits/5008d2036a6efaf490c23d4c52ba7f749ec3fe74

files/en-us/web/api/window/settimeout/index.md
hamishwillee: Yes it is. Again, did not really even look at this as orthogonal to "the point".

Anyhow, I have mostly fixed in https://github.com/mdn/content/pull/42072/commits/40a026c4335b95259006cb031265d9c9b4678319

I say mostly, because I used this approach to highlight the problem, and the arrow solution. However I don't understand how to integrate it with `bind()` (and I don't want to at this point). So the example is still complicated in the final part.

files/en-us/web/api/window/settimeout/index.md
hamishwillee: Arrgg, missed this. Added both examples in https://github.com/mdn/content/pull/42072/commits/dd7fe0b22a73d2688f7e5b9957d8b1519b41e8f3

files/en-us/web/api/window/settimeout/index.md
wbamberg: Thanks Hamish. It is true that these are unrelated to your PR, and I would not have commented if it had not been in the diffs. But if it's in the diffs, I have to read it, and if I read it I have to comment :). And I appreciate your willingness to leave things better than you found them.

hamishwillee: If you're willing to add a comment with a fix suggestion I will always try to integrate it. I'm just explaining why I didn't in the first place. Excuses, excuses!

wbamberg (approved):
👍 thank you!

hamishwillee:
Thank you @wbamberg - wish you the very happiest new year.

PR #42073, opened 2025-11-24
TrustedTypes: Worker()/SharedWorker() constructors

hamishwillee:
The `Worker()` and `SharedWorker()` constructors takes a `url` parameter that that can be a `TrustedScriptURL`. This attempts to follow the pattern of our other trusted types docs and add the header warning and security considerations section.

`SharedWorker` appears to be the same as Worker but of course constructs a different object. To reduce duplication I have mostly linked from `SharedWorker()` rather than copy-pasting. 

Related project: #41507

### DIFF
--- a/files/en-us/web/api/worker/worker/index.md
+++ b/files/en-us/web/api/worker/worker/index.md
@@ -27,7 +27,6 @@ new Worker(url, options)
 ### Parameters
 
 - `url`
-
   - : A {{domxref("TrustedScriptURL")}} object or a string representing the URL of the script or module that the worker will execute.
 
     This must be same-origin with the caller's document, or a `blob:` or `data:` URL.


files/en-us/web/api/worker/worker/index.md
```diff
@@ -8,10 +8,16 @@ browser-compat: api.Worker.Worker
...
-The **`Worker()`** constructor creates a {{domxref("Worker")}} object that executes the script at the specified URL. This script must obey the [same-origin policy](/en-US/docs/Web/Security/Same-origin_policy).
+> [!WARNING]
+> This script passed to the `url` element is executed.
+> APIs like this are known as [injection sinks](/en-US/docs/Web/API/Trusted_Types_API#concepts_and_usage), and are potentially a vector for [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
+>
+> You can mitigate this risk by having a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that restricts the locations from which scripts can be loaded, and by always assigning {{domxref("TrustedScriptURL")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](#security_considerations) for more information.
 
-> [!NOTE]
-> There is a disagreement among browser manufacturers about whether a data URL is of the same origin or not. Though Firefox 10 and later accept data URLs, that's not the case in all other browsers.
```
hamishwillee, line 14: FYI, I have removed this because it is no longer true. This is covered below in section "`data:` and `blob:` URLs" - that is my interpretation of the spec after some discussion with a few very confused AIs.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -8,10 +8,16 @@ browser-compat: api.Worker.Worker
...
+The **`Worker()`** constructor creates a {{domxref("Worker")}} object that executes the script at the specified URL.
+
+This script must obey the [same-origin policy](/en-US/docs/Web/Security/Same-origin_policy).
```
hamishwillee: This is very confusing, because it implies that the worker must be same-origin with its owner. Perhaps that was once true. Or more likely this means more generally that you can only load same-origin docs by default but you are restricted by CORS rules for cross origin resources.

I infer this because the options you can pass only make sense if you can pass cross origin urls.

@wbamberg Do you think this is reasonable?

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,35 +29,72 @@ new Worker(url, options)
...
-    > Bundlers, including [webpack](https://webpack.js.org/guides/web-workers/), [Vite](https://vite.dev/guide/features.html#web-workers), and [Parcel](https://parceljs.org/languages/javascript/#web-workers), recommend passing URLs that are relative to [`import.meta.url`](/en-US/docs/Web/JavaScript/Reference/Operators/import.meta#url) to the `Worker()` constructor. For example:
-    >
-    > ```js
-    > const myWorker = new Worker(new URL("worker.js", import.meta.url));
-    > ```
-    >
-    > This way, the path is relative to the current script instead of the current HTML page, which allows the bundler to safely do optimizations like renaming (because otherwise the `worker.js` URL may point to a file not controlled by the bundler, so it cannot make any assumptions).
-
+  - : A {{domxref("TrustedScriptURL")}} object or a string representing the URL of the script the worker will execute.
+    It must obey the same-origin policy.
```
hamishwillee: Again with the same-origin thing. Depending on answer above I think I will remove this and have a section on the URL policy below. If it was only same origin TT wouldn't make much sense IMO. Though I guess you could still use it to block URLs with URL parameters from the same site.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,35 +29,72 @@ new Worker(url, options)
...
   - : Thrown if the MIME type of the worker script is incorrect. It _should_ always be `text/javascript`
     (for historical reasons [other JavaScript MIME types](/en-US/docs/Web/HTTP/Guides/MIME_types#legacy_javascript_mime_types) may be accepted).
+- `SecurityError` {{domxref("DOMException")}}
+  - : Thrown if the document is not allowed to start workers, e.g., if the URL has an invalid syntax or if the same-origin policy is violated.
 - `SyntaxError` {{domxref("DOMException")}}
-  - : Thrown if _aURL_ cannot be parsed.
+  - : Thrown if `url` cannot be parsed.
+- `TypeError`
+  - : Thrown if the `url` parameter is set with a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
+    <!-- This is also thrown if the fetched URL cannot be successfully parsed as its indicated type. -->
```
hamishwillee: This was true in other case I copied the error from. Not sure yet if true here.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,35 +29,72 @@ new Worker(url, options)
...
+
+It is extremely risky to accept and execute arbitrary URLs from untrusted origins.
+A website should control what scripts that are allowed to run using a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) with the [`worker-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/worker-src) directive (or a fallback defined in [`default-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src)).
+This can restrict scripts to those from the current origin, or a specific set of origins, or even particular files.
+
+If you're using this property and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) (using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive), you will need to always assign {{domxref("TrustedScriptURL")}} objects instead of strings.
+This ensures that the input is passed through a transformation function, which has the chance to reject or modify the URL before it is injected.
+
+<!-- Need to check COORs/COEP rules.-->
```
hamishwillee: So as above I think this is subject to CORS, which means that a non-same-origin URL is blocked by default. I see this as lower risk, though I guess a site might host "evil" files by accident (i.e. if it were a wiki) and then these could be loaded in the same-site context.

But more generally a cross-origin resource would have to allow "evil js" explicitly for it to be accessible. So what do I say here - just that?

files/en-us/web/api/worker/worker/index.md
hamishwillee: OK, so this appears to be true. The initial script or module must be same-origin with the document. Note though that the imported scripts from that module need not be.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,35 +27,114 @@ new Worker(url, options)
...
+
+The specified script must be [same-origin](/en-US/docs/Web/Security/Same-origin_policy) with the assocated document, but may itself import scripts or modules that are cross-origin (if permitted by CORS and other restrictions).
+
+### Module and classic workers
+
+Workers can be classic workers or module workers.
+The type of worker affects the worker constructor options, how the worker script is fetched, and how it is executed.
+
+The code below shows two ways you can construct a classic worker, and also how you specify the `type` of `"module"` to create a module worker.
+In both cases the script must be same-origin with the loading document and is resolved relative to the location of the launching document.
```
hamishwillee, line 77: Sometimes AIs say this is correct. Others say that the modules are loaded from any origin provided the CORS headers are set. This is the result of my testing - only the same origin works.

In support the WPT tests ONLY show same-origin imports (and data and blob). No cross origin.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,35 +27,114 @@ new Worker(url, options)
...
+```
+
+Module workers and their dependencies are loaded and executed using ECMAScript module semantics:
+
+- Fetched using [CORS](/en-US/docs/Web/HTTP/Guides/CORS)
+- All modules are resolved before any code is executed
+- Must be served with the media type `Content-Type: text/javascript`
+- Executed in {{glossary("Strict mode")}}
+- Dependencies are loaded via static [`import` statements](../en-US/docs/Web/JavaScript/Reference/Statements/import)
+- Dependencies are restricted by the [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP), in particular the `script-src` and `worker-src` directives
```
hamishwillee: It might be that `worker-src` should apply to just the worker, and `script-src` to their imports (and ChatGPT told me this when I asked why worker-src matters). My testing seems to indicate  `worker-src` controls the imported modules.

But CSP doesn't seem to affect the classic worker imports. Would appreciate independent testing. I haven't commented on this below.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,35 +27,114 @@ new Worker(url, options)
...
+
+It is extremely risky to accept and execute arbitrary URLs from untrusted origins.
+A website should control what scripts that are allowed to run using a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) with the [`worker-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/worker-src) directive (or a fallback defined in [`default-src`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/default-src)).
+This can restrict scripts to those from the current origin, or a specific set of origins, or even particular files.
+
+If you're using this property and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) (using the [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) CSP directive), you will need to always assign {{domxref("TrustedScriptURL")}} objects instead of strings.
+This ensures that the input is passed through a transformation function, which has the chance to reject or modify the URL before it is injected.
 
 ## Examples
```
hamishwillee, line 156: I still plan to add TT example

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,35 +27,114 @@ new Worker(url, options)
...
+});
+```
+
+Module workers and their dependencies are loaded and executed using ECMAScript module semantics:
+
+- Fetched using [CORS](/en-US/docs/Web/HTTP/Guides/CORS)
+- All modules are resolved before any code is executed
+- Must be served with the media type `Content-Type: text/javascript`
+- Executed in {{glossary("Strict mode")}}
+- Dependencies are loaded via static [`import` statements](../en-US/docs/Web/JavaScript/Reference/Statements/import)
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
- Dependencies are loaded via static [`import` statements](/en-US/docs/Web/JavaScript/Reference/Statements/import)
```

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
+- Executed in {{glossary("Strict mode")}}
+- Dependencies are loaded via static [`import` statements](/en-US/docs/Web/JavaScript/Reference/Statements/import)
+
+All same-origin workers can load same-origin scripts by default.
+However because module scripts are fetched using CORS, cross-origin scripts must be served with the {{httpheader("Access-Control-Allow-Origin")}} header.
+In addition, the if the document has a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP), it must allow the origins of imported worker scripts in `worker-src` (with fallback to `script-src` and `default-src` directives).
+
+Classic workers are fetched in `no-cors` mode, executed as scripts, and import other scripts using {{domxref("WorkerGlobalScope.importScripts()")}} (not `import` statements).
+Because they are fetched in `no-cors` mode, cross-origin requests are not blocked by CORS, and the response does not have to include the `Access-Control-Allow-Origin` header.
+Unlike for modules, the document [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) is not affected by the `worker-src` directive, but it is still affected by sources listed in the `script-src` directive.
```
hamishwillee: My testing shows that 
-  **module** workers respect `worker-src`. So if you set a `CSP: worker-src 'self'` you can load a module (which is always same origin), then it can `import` further same origin modules, but NOT cross-origin modules.
- *classic/script** workers do not respect `worker-src`.  So if you set a `CSP: worker-src 'self'`, a wokrer script can import a cross origin script using  `importScript` .
- *classic/script** workers do respect `script-src`,  So if you set a `CSP: script-src 'self'` this will block a worker script loading a cross origin script.

files/en-us/web/api/sharedworker/sharedworker/index.md
```diff
@@ -29,51 +29,66 @@ new SharedWorker(url, options)
 ### Parameters
 
 - `url`
-  - : A string representing the URL of the script the worker will
-    execute. It must obey the same-origin policy.
+  - : A {{domxref("TrustedScriptURL")}} object or a string representing the URL of the script or module that the worker will execute.
+    This must be same-origin with the associated document (`blob:` URLs are same-origin), or a `data:` URL.
+    The URL is resolved relative to the current HTML page's location.
```
hamishwillee, line 33: This is the only real change in this section - the rest is layout.

files/en-us/web/api/sharedworker/sharedworker/index.md
```diff
@@ -8,15 +8,15 @@ browser-compat: api.SharedWorker.SharedWorker
...
+> You can mitigate this risk by having a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that restricts the locations from which scripts can be loaded, and by always assigning {{domxref("TrustedScriptURL")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](/en-US/docs/Web/API/Worker/Worker#security_considerations) in the `Worker()` constructor for more information.
 
-> [!NOTE]
-> There is disagreement among browser manufacturers about
-> whether a data URL is of the same origin or not. Although Firefox 10.0
-> and later accept data URLs, that's not the case in all other
-> browsers.
+The **`SharedWorker()`** constructor creates a {{domxref("SharedWorker")}} object that executes the script at the specified URL.
+This script must obey the [same-origin policy](/en-US/docs/Web/Security/Defenses/Same-origin_policy).
```
wbamberg: Does this mean something like "the script must be same-origin with the caller's document"? If so. it should say that. I think that's more precise than "This script must obey the same-origin policy" which honestly doesn't really make sense - everything has to "obey the same-origin policy", it's a browser policy.

I mean, if you _could_ load a cross-origin script, you wouldn't explain this by saying "The script doesn't have to obey the SOP" - that wouldn't make sense. So nor does the inverse.

Or, better probably, it could say nothing here since the info is given in the docs for the `url` argument, which is a better place.

files/en-us/web/api/sharedworker/sharedworker/index.md
```diff
@@ -29,51 +29,66 @@ new SharedWorker(url, options)
 ### Parameters
 
 - `url`
-  - : A string representing the URL of the script the worker will
-    execute. It must obey the same-origin policy.
+  - : A {{domxref("TrustedScriptURL")}} object or a string representing the URL of the script or module that the worker will execute.
+    This must be same-origin with the associated document (`blob:` URLs are same-origin), or a `data:` URL.
```
wbamberg: I think "associated" is a bit slippery. Could we say as I did above "the caller's document"? Is that correct?

files/en-us/web/api/sharedworker/sharedworker/index.md
```diff
@@ -29,51 +29,66 @@ new SharedWorker(url, options)
...
 - `SecurityError` {{domxref("DOMException")}}
   - : Thrown if the document is not allowed to start workers, for example if the URL has an invalid syntax or if the same-origin policy is violated, or if the `sameSiteCookies` value is not supported in the given context.
 - `NetworkError` {{domxref("DOMException")}}
-  - : Thrown if the MIME type of the worker script is incorrect. It should _always_ be `text/javascript` (for historical reasons [other JavaScript MIME types](/en-US/docs/Web/HTTP/Guides/MIME_types#textjavascript) may be accepted).
+  - : Thrown if the MIME type of the worker script is incorrect.
+    It should _always_ be `text/javascript` (for historical reasons [other JavaScript MIME types](/en-US/docs/Web/HTTP/Guides/MIME_types#textjavascript) may be accepted).
 - `SyntaxError` {{domxref("DOMException")}}
   - : Thrown if `url` cannot be parsed.
+- `TypeError`
+  - : Thrown if the `url` parameter is set with a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
```
wbamberg:
```suggestion
  - : Thrown if the `url` parameter is a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
```

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
-    > Bundlers, including [webpack](https://webpack.js.org/guides/web-workers/), [Vite](https://vite.dev/guide/features.html#web-workers), and [Parcel](https://parceljs.org/languages/javascript/#web-workers), recommend passing URLs that are relative to [`import.meta.url`](/en-US/docs/Web/JavaScript/Reference/Operators/import.meta#url) to the `Worker()` constructor. For example:
-    >
-    > ```js
-    > const myWorker = new Worker(new URL("worker.js", import.meta.url));
-    > ```
-    >
-    > This way, the path is relative to the current script instead of the current HTML page, which allows the bundler to safely do optimizations like renaming (because otherwise the `worker.js` URL may point to a file not controlled by the bundler, so it cannot make any assumptions).
-
+  - : A {{domxref("TrustedScriptURL")}} object or a string representing the URL of the script or module that the worker will execute.
+    This must be same-origin with the associated document (`blob:` URLs are same-origin), or a `data:` URL.
```
wbamberg: Same comment as before.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
+    The URL is resolved relative to the current HTML page's location.
 - `options` {{optional_inline}}
-  - : An object containing option properties that can be set when creating the object instance. Available properties are as follows:
-    - `type`
-      - : A string specifying the type of worker to create. The value can be `classic` or `module`. If not specified, the default used is `classic`.
+  - : An object containing option properties that can be set when creating the object instance.
+    Available properties are as follows:
     - `credentials`
```
wbamberg, line 38: I guess this has the same meaning (and constraints?) as https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#credentials, where it is all described more fully (including saying what credentials are). Maybe link there and say this?

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
-      - : A string specifying the type of credentials to use for the worker. The value can be `omit`, `same-origin`, or _`include`. If not specified, or if type is `classic`, the default used is `same-origin` (only include credentials for same-origin requests)._
+      - : A string specifying the type of credentials to use for the worker.
+        The value can be `omit`, `same-origin`, or _`include`.
```
wbamberg: I don't understand, if the script must be same-origin, what's the difference here between `same-origin` and `include`?

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
+        If not specified, or if type is `classic`, the default used is `same-origin` (only include credentials for same-origin requests)._
```
wbamberg: I don't understand this. I read it as two separate statements really:

1. "if not specified, the default is same-origin"
2. "if type is `classic`, the default is same-origin" 

But "default" just means "the value used if the option isn't specified", so doesn't (2) just collapse into (1)? Put another way, what's the default if type is `module`?

Does it want to say something like "if type is `classic`, then this option is ignored, and `credentials`are only sent for same-origin requests"?

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
 - `SecurityError` {{domxref("DOMException")}}
   - : Thrown if the document is not allowed to start workers, e.g., if the URL has an invalid syntax or if the same-origin policy is violated.
-- `NetworkError` {{domxref("DOMException")}}
-  - : Thrown if the MIME type of the worker script is incorrect. It _should_ always be `text/javascript`
-    (for historical reasons [other JavaScript MIME types](/en-US/docs/Web/HTTP/Guides/MIME_types#legacy_javascript_mime_types) may be accepted).
 - `SyntaxError` {{domxref("DOMException")}}
-  - : Thrown if _aURL_ cannot be parsed.
+  - : Thrown if `url` cannot be parsed.
+- `TypeError`
+  - : Thrown if the `url` parameter is set with a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
```
wbamberg:
```suggestion
  - : Thrown if the `url` parameter is a string when [Trusted Types](/en-US/docs/Web/API/Trusted_Types_API) are [enforced by a CSP](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types) and no default policy is defined.
```

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
+- Executed in {{glossary("Strict mode")}}
+- Dependencies are loaded via static [`import` statements](/en-US/docs/Web/JavaScript/Reference/Statements/import)
+
+All same-origin workers can load same-origin scripts by default.
+However because module scripts are fetched using CORS, cross-origin scripts must be served with the {{httpheader("Access-Control-Allow-Origin")}} header.
+In addition, the if the document has a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP), it must allow the origins of imported worker scripts in `worker-src` (with fallback to `script-src` and `default-src` directives).
+
+Classic workers are fetched in `no-cors` mode, executed as scripts, and import other scripts using {{domxref("WorkerGlobalScope.importScripts()")}} (not `import` statements).
+Because they are fetched in `no-cors` mode, cross-origin requests are not blocked by CORS, and the response does not have to include the `Access-Control-Allow-Origin` header.
+Unlike for modules, the document [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) is not affected by the `worker-src` directive, but it is still affected by sources listed in the `script-src` directive.
```
wbamberg: Am I right to understand that these 2 paras are exclusively about scripts that the worker script itself loads? i.e. they're _not_ about loading the worker script itself (since that's always same-origin)?

If so then I think you ought to have a heading like "Loading scripts from a worker" or something, to make this very clear.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
+- Must be served with the media type `Content-Type: text/javascript`
+- Executed in {{glossary("Strict mode")}}
+- Dependencies are loaded via static [`import` statements](/en-US/docs/Web/JavaScript/Reference/Statements/import)
+
+All same-origin workers can load same-origin scripts by default.
+However because module scripts are fetched using CORS, cross-origin scripts must be served with the {{httpheader("Access-Control-Allow-Origin")}} header.
+In addition, the if the document has a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP), it must allow the origins of imported worker scripts in `worker-src` (with fallback to `script-src` and `default-src` directives).
+
+Classic workers are fetched in `no-cors` mode, executed as scripts, and import other scripts using {{domxref("WorkerGlobalScope.importScripts()")}} (not `import` statements).
+Because they are fetched in `no-cors` mode, cross-origin requests are not blocked by CORS, and the response does not have to include the `Access-Control-Allow-Origin` header.
```
wbamberg: I don't really understand this. It seems to be saying: because classic workers are fetched in `no-cors`, then they can make cross-origin requests without CORS headers (btw "not blocked by CORS" is not a good way to put it, CORS doesn't block anything, it only unblocks things).

But is that right? Why does the mode a script was fetched in affect the mode of the requests it makes? I don't think this is true. If a resource is fetched in no-cors then _the fetching of it_ has certain attributes, but I don't think that "carries over" to the requests that it makes, does it?

files/en-us/web/api/worker/worker/index.md
wbamberg: It sounds like you're saying that `worker-src` controls the scripts that a worker can load? Not the source of the worker itself? In that case isn't the example here wrong: https://w3c.github.io/webappsec-csp/#directive-worker-src ?

files/en-us/web/api/sharedworker/sharedworker/index.md
hamishwillee: Yes it is

files/en-us/web/api/worker/worker/index.md
hamishwillee: @wbamberg I have updated my note above to state more clearly what testing shows.

But to summarize, I'm saying that
1. `worker-src` affects the scripts that can be imported into a module worker, but not the scripts that can be imported into a classic worker.
2. `worker-src` seems to be pointless, since if I specify `Content-Security-Policy: worker-src https://example.com/ ` neither a module or classic script at that address will load. ONLY same-origin modules ever load.

Clearly the intent is that the `worker-src` allows control over where module scripts can be loaded from, and "perhaps" also where they may load. But that's not what my testing is showing. Of course my testing might be wrong.

files/en-us/web/api/worker/worker/index.md
hamishwillee: OK

files/en-us/web/api/sharedworker/sharedworker/index.md
hamishwillee: Yes. I'll update as you suggest. - I'm so exhausted by the testing.

files/en-us/web/api/worker/worker/index.md
hamishwillee: So the these options are what make me think that maybe it should be possible to import a worker cross origin.

Or to put it another way, "nothing". I don't understand the point of these options. They make me question all my tests.

They might affect the requests made for imports from the loaded same origin script.

files/en-us/web/api/worker/worker/index.md
hamishwillee: This predates me - I will need to check.

files/en-us/web/api/worker/worker/index.md
hamishwillee: @wbamberg So yes, I am saying that classic workers are fetched in `no-cors` mode. My understanding was that this mean't that any requirements of CORS are ignored - the request will succeed as if CORS headers were not set.

I did not mean to make any statement about what mode a classic script would import its scripts using - I just tried to make clear that a classic script uses `importScripts()` to load further scripts, and not `import`.

You're right that we can't draw any assumptions about this. That said, that is what my tests showed. I'll try check again on Friday.

wbamberg: > @wbamberg So yes, I am saying that classic workers are fetched in no-cors mode. My understanding was that this mean't that any requirements of CORS are ignored - the request will succeed as if CORS headers were not set.

But that's only relevant for cross-origin fetches, right? So if fetching a worker has to be same-origin, it's irrelevant. Put another way, the doc says:

> Because they are fetched in `no-cors` mode, cross-origin requests are not blocked by CORS,

...what are these "cross-origin requests", if fetching a worker has to be same-origin?

files/en-us/web/api/worker/worker/index.md
hamishwillee: @jakearchibald I wonder if you might offer some advice here on Worker/SharedWorker construction?

The original docs and my testing indicate that a document can only create a module or classic worker by loading a script/module that is **same-origin** with it. 

However this does not seem logical given that `CSP: worker-src` is supposed to control the locations of where worker scripts can be loaded from. So for example   `Content-Security-Policy: worker-src https://example.com/ ` should allow a module or classic script at that address to load, but testing seems to show that only same-origin modules ever load - any cross origin value you put is ignored.
We also have these options like "include" on worker construction that imply you can load a script cross origin - so trying to work out when they might apply if I am correct.

My testing could be crap, but I'd like to "know" so that I don't waste time trying to test something that can't work.

If there are examples showing cross-origin loading of module scripts can you point me to them, and/or confirm the rules?

And if you are not sure, is there anyone you could suggest we ask?

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
+## Description
+
+The **`Worker()`** constructor creates a {{domxref("Worker")}} object that executes the classic script or module at the specified URL.
+
+The script must be [same-origin](/en-US/docs/Web/Security/Same-origin_policy) with the associated document, but may itself import scripts or modules that are cross-origin (if permitted by CORS and other restrictions).
+If a cross-origin worker is required, users must load it from an intermediate same-origin worker or a blob.
+
+### Module and classic workers
+
+Workers can be classic workers or module workers.
```
hamishwillee:
```suggestion
A classic worker is one that hat is constructed from a classic script, while a module worker is constructed from an [ECMASCript module](/en-US/docs/Web/JavaScript/Guide/Modules).
```

files/en-us/web/api/worker/worker/index.md
hamishwillee: Cross origin import requests made by the worker.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,170 @@ new Worker(url, options)
...
+- Must be served with the media type `Content-Type: text/javascript`
+- Executed in {{glossary("Strict mode")}}
+- Dependencies are loaded via static [`import` statements](/en-US/docs/Web/JavaScript/Reference/Statements/import)
+
+All same-origin workers can load same-origin scripts by default.
+However because module scripts are fetched using CORS, cross-origin scripts must be served with the {{httpheader("Access-Control-Allow-Origin")}} header.
+In addition, the if the document has a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP), it must allow the origins of imported worker scripts in `worker-src` (with fallback to `script-src` and `default-src` directives).
+
+Classic workers are fetched in `no-cors` mode, executed as scripts, and import other scripts using {{domxref("WorkerGlobalScope.importScripts()")}} (not `import` statements).
+Because they are fetched in `no-cors` mode, cross-origin requests are not blocked by CORS, and the response does not have to include the `Access-Control-Allow-Origin` header.
```
hamishwillee: Attempts to clarify https://github.com/mdn/content/pull/42073/files#r2663618688

Essentially my [reading of the spec](https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-classic-worker-imported-script) is that all imports on a classic worker are made `no-cors`. It doesn't say that directly, but unlike module requests they are syncronous and don't report errors in the same way as a cors request does. So this part of the spec reads like the `<script>` tag with `no-cors`. Certainly this matches testing.

```suggestion
Classic workers are fetched in `no-cors` mode and are executed as scripts rather than modules.
They can themselves import other scripts using {{domxref("WorkerGlobalScope.importScripts()")}} (not `import` statements), and these are also fetched in `no-cors` mode.
Because imports are fetched in `no-cors` mode,  they can be requested cross-origin even if the server does not set the appropriate CORS headers.
```

files/en-us/web/api/worker/worker/index.md
hamishwillee: I have addressed in https://github.com/mdn/content/pull/42073/files#r2674625053

files/en-us/web/api/worker/worker/index.md
hamishwillee: Hoping to get some clarity from Jake in https://github.com/mdn/content/pull/42073/files#r2666249206

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,171 @@ new Worker(url, options)
...
+- Executed in {{glossary("Strict mode")}}
+- Dependencies are loaded via static [`import` statements](/en-US/docs/Web/JavaScript/Reference/Statements/import)
+
+All same-origin workers can load same-origin scripts by default.
+However because module scripts are fetched using CORS, cross-origin scripts must be served with the {{httpheader("Access-Control-Allow-Origin")}} header.
+In addition, the if the document has a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP), it must allow the origins of imported worker scripts in `worker-src` (with fallback to `script-src` and `default-src` directives).
+
+Classic workers are fetched in `no-cors` mode and are executed as scripts rather than modules.
+They can themselves import other scripts using {{domxref("WorkerGlobalScope.importScripts()")}} (not `import` statements), and these are also fetched in `no-cors` mode.
+Because imports are fetched in `no-cors` mode,  they can be requested cross-origin even if the server does not set the appropriate CORS headers.
```
hamishwillee:
```suggestion
Because imports are fetched in `no-cors` mode, scripts can be requested cross-origin even if the server does not set the appropriate CORS headers.
```

files/en-us/web/api/sharedworker/sharedworker/index.md
hamishwillee: Keeping some text. I find this so surpising it is worth mentioning here.

```suggestion
This script must be [same-origin](/en-US/docs/Web/Security/Defenses/Same-origin_policy) with the caller's document.
```

files/en-us/web/api/sharedworker/sharedworker/index.md
hamishwillee:
```suggestion
    This must be same-origin with the caller's document, a `blob:`, or a `data:` URL.
```

files/en-us/web/api/worker/worker/index.md
hamishwillee:
```suggestion
    This must be same-origin with the caller's document, or a `blob:` or `data:` URL.
```

files/en-us/web/api/sharedworker/sharedworker/index.md
```diff
@@ -8,15 +8,15 @@ browser-compat: api.SharedWorker.SharedWorker
...
+> You can mitigate this risk by having a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) that restricts the locations from which scripts can be loaded, and by always assigning {{domxref("TrustedScriptURL")}} objects instead of strings and [enforcing trusted types](/en-US/docs/Web/API/Trusted_Types_API#using_a_csp_to_enforce_trusted_types).
+> See [Security considerations](/en-US/docs/Web/API/Worker/Worker#security_considerations) in the `Worker()` constructor for more information.
 
-> [!NOTE]
-> There is disagreement among browser manufacturers about
-> whether a data URL is of the same origin or not. Although Firefox 10.0
-> and later accept data URLs, that's not the case in all other
-> browsers.
+The **`SharedWorker()`** constructor creates a {{domxref("SharedWorker")}} object that executes the script at the specified URL.
+This script must be [same-origin](/en-US/docs/Web/Security/Defenses/Same-origin_policy) with the caller's document.
```
hamishwillee:
```suggestion
```

files/en-us/web/api/sharedworker/sharedworker/index.md
```diff
@@ -29,51 +29,66 @@ new SharedWorker(url, options)
 ### Parameters
 
 - `url`
-  - : A string representing the URL of the script the worker will
-    execute. It must obey the same-origin policy.
+  - : A {{domxref("TrustedScriptURL")}} object or a string representing the URL of the script or module that the worker will execute.
+    This must be same-origin with the caller's document, a `blob:`, or a `data:` URL.
```
hamishwillee:
```suggestion
    This must be same-origin with the caller's document, or a `blob:` or `data:` URL.
```

files/en-us/web/api/worker/worker/index.md
hamishwillee: Yes. Done in https://github.com/mdn/content/pull/42073/commits/1837809b5b1bfa92d4d7f2f64d19e8fb7fd910d4

This controls whether a module worker sends credentials when it imports a cross-origin module.

This is ignored for classic modules. They send credentials for same-origin imports. They generally won't send them for cross-origin imports - it isn't under the control of the interface. They might if the remote server allows it.

files/en-us/web/api/worker/worker/index.md
hamishwillee: Addressed. See https://github.com/mdn/content/pull/42073/commits/aa1c134cdad1ed45b97024d9b954c714dbf46b08#r2674765716

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,189 @@ new Worker(url, options)
...
+- Dependencies are imported via static [`import` statements](/en-US/docs/Web/JavaScript/Reference/Statements/import)
+- Fetched asynchronously using [CORS](/en-US/docs/Web/HTTP/Guides/CORS).
+- All modules are resolved before any code is executed
+- Must be served with the media type `Content-Type: text/javascript`
+- Executed in {{glossary("Strict mode")}}
+
+Classic workers are fetched and executed as scripts:
+
+- Dependencies are imported using the {{domxref("WorkerGlobalScope.importScripts()")}} method
+- Fetched synchonously in `no-cors` mode
```
hamishwillee, line 103: We would expand this with other differences if we knew them.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,189 @@ new Worker(url, options)
...
+Module workers can import [ECMASCript modules](/en-US/docs/Web/JavaScript/Guide/Modules) using [`import` statements](/en-US/docs/Web/JavaScript/Reference/Statements/import).
+Modules are fetched using CORS, so cross-origin modules must be served with the {{httpheader("Access-Control-Allow-Origin")}} header in order to be loaded.
+Developers can specify whether or not credentials should be sent in cross-origin imports.
+
+Classic workers can import scripts (but not modules) using the {{domxref("WorkerGlobalScope.importScripts()")}} method.
+Unlike modules, scripts are fetched in `no-cors` mode, and can be requested cross-origin even if the server does not set the appropriate CORS headers.
+Credentials are sent for same-origin imports, but are usully not sent for cross-origin requests.
+
+In addition, if the document has a [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP), it must allow the origins of imported scripts or modules.
+For modules the allowed sources are specified in `worker-src` (with fallback to `script-src` and `default-src` directives), while for classic scripts the sources are specified in `script-src` (with fallback to the `default-src` directives).
```
hamishwillee, line 116: @wbamberg I think this is ready for another look.

Unfortunately I ate the commits while trying to create a version where it would be easy to be determine the changes.

Since last you looked I have addressed your comments, including adding this section on importing scripts and the credentials.

Essentially I think my testing provides a consistent story (or or less) - workers must be constructed with a same origin script/module. The credentials apply to the modules that are imported by a module worker (but not a script worker, over which you have no control via the API).

The remaining inconsitency is this section - why would you use `worker-src` if you can't import the main script cross-origin, and why is it ignored for scripts. However this is what my testing shows about CSP. If we question this too much, we could drop this line. Or just put it in, and hope to tidy later.

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,189 @@ new Worker(url, options)
...
-  - : An object containing option properties that can be set when creating the object instance. Available properties are as follows:
-    - `type`
-      - : A string specifying the type of worker to create. The value can be `classic` or `module`. If not specified, the default used is `classic`.
+
+  - : An object containing option properties that can be set when creating the object instance.
+    Available properties are as follows:
+
     - `credentials`
-      - : A string specifying the type of credentials to use for the worker. The value can be `omit`, `same-origin`, or _`include`. If not specified, or if type is `classic`, the default used is `same-origin` (only include credentials for same-origin requests)._
+
```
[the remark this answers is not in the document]
hamishwillee: I ran prettier, so I think these are invaid.

files/en-us/web/api/worker/worker/index.md
jakearchibald: As in me? Which part? I think GitHub is being unhelpful and hiding something

wbamberg: Hamish is out for the next two weeks but I promised to try to get this PR merged if I could :).

The question we have is, when a page loads a worker script, does the worker have to be same origin with the loading document? The [page for the `Worker()` constructor](https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker) says:

> This script must obey the [same-origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Same-origin_policy).

...which seems to me a curious thing to say (I mean, everything has to obey the SOP, it's a browser policy, so what is the substantive force of this statement?). We weren't sure whether this means "the script must be same origin with the loading document". Hamish's tests indicate that it _does_ have to be same origin.

But that raises further questions, such as, what is the `worker-src` CSP directive for, then? The [spec for worker-src](https://w3c.github.io/webappsec-csp/#directive-worker-src) even has an example of how you would use the directive to block workers from a different source, and why would you need that if the worker needs to be same origin?

So we thought we could ask you since you probably know.

jakearchibald: Yeah the root worker script needs to be same origin. However, scripts from other origins can be loaded via ES imports and `importScripts`. I guess that's what worker-src is for

wbamberg: OK, thank you! Then it looks like Hamish's tests are all right.

>  I guess that's what worker-src is for

That is still odd, because it's not what the example from the spec is showing: https://w3c.github.io/webappsec-csp/#example-fd6c5849. But I'm going to let it go now :).

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,189 @@ new Worker(url, options)
...
+### Importing scripts or modules
+
+Module workers can import [ECMASCript modules](/en-US/docs/Web/JavaScript/Guide/Modules) using [`import` statements](/en-US/docs/Web/JavaScript/Reference/Statements/import).
+Modules are fetched using CORS, so cross-origin modules must be served with the {{httpheader("Access-Control-Allow-Origin")}} header in order to be loaded.
+Developers can specify whether or not credentials should be sent in cross-origin imports.
+
+Classic workers can import scripts (but not modules) using the {{domxref("WorkerGlobalScope.importScripts()")}} method.
+Unlike modules, scripts are fetched in `no-cors` mode, and can be requested cross-origin even if the server does not set the appropriate CORS headers.
+Credentials are sent for same-origin imports, but are usully not sent for cross-origin requests.
```
wbamberg:
```suggestion
Credentials are sent for same-origin imports, but are usually not sent for cross-origin requests.
```

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,189 @@ new Worker(url, options)
...
+// Create a TrustedScriptURL instance using the policy
+const trustedScriptURL = policy.createScriptURL(untrustedScriptURL);
+
+// Construct the worker with the trusted URL
+const myWorker = new Worker(trustedScriptURL);
+```
+
+### Creating a classic worker
+
+The following code snippet shows creation of a classic {{domxref("Worker")}} object using the `Worker()` constructo, and subsequent usage of the object:
```
wbamberg:
```suggestion
The following code snippet shows creation of a classic {{domxref("Worker")}} object using the `Worker()` constructor, and subsequent usage of the object:
```

files/en-us/web/api/worker/worker/index.md
```diff
@@ -23,39 +27,186 @@ new Worker(url, options)
 ### Parameters
 
 - `url`
-  - : A string representing the URL of the script the worker will execute. It must obey the same-origin policy. The URL is resolved relative to the current HTML page's location.
```
wbamberg:
```suggestion
```

files/en-us/web/api/worker/worker/index.md
hamishwillee: Thanks @jakearchibald and @wbamberg . It is a relief to know that all the testing was not just "wrong". I'm OK to let this all go for now - as you say it is a bit odd but it is what I saw.

PR #42079, opened 2025-11-24
Fix rotationRate property descriptions order

antonegas:
### Description

Updated the order for the rotationRate property fields the match the definition in the standard.

### Motivation

The current order does not follow what is defined in the W3S standard of the `DeviceMotionEvent` properties or the current documentation of the `acceleration` and `accelerationIncludingGravity` property.

### Additional details

Matches what is defined in the W3S standard: https://w3c.github.io/deviceorientation/#device-motion-event-rotation-rate-api
Aligns with the documentation of the acceleration property: https://developer.mozilla.org/en-US/docs/Web/API/DeviceMotionEvent/acceleration
and accelerationIncludingGravity property: https://developer.mozilla.org/en-US/docs/Web/API/DeviceMotionEvent/accelerationIncludingGravity

None that I can find

### DIFF
--- a/files/en-us/web/api/devicemotionevent/rotationrate/index.md
+++ b/files/en-us/web/api/devicemotionevent/rotationrate/index.md
@@ -21,12 +21,12 @@ The `rotationRate` property is a read only object describing the rotation
 rates of the device around each of its axes:
 
 - `alpha`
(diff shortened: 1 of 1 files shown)


sideshowbarker:
Anton, congrats on landing your first docs change here — welcome aboard 🎉

PR #42086, opened 2025-11-24
Fix getRed/setRed example to use HSL→RGB conversion

ShilpaG813:
Fixes mdn#42059

### Description
Updates the Color class example so that getRed() and setRed() correctly operate on the red
channel after the internal representation was changed from RGB to HSL. Previously, the example
incorrectly accessed values[0], which is the hue component in HSL.

### Motivation
The documentation states that the internal storage is changed to HSL using rgbToHSL().
Therefore, getRed() and setRed() must convert HSL → RGB when reading or modifying the red channel.
This PR aligns the example with the encapsulation explanation and corrects the logic.

### Additional details
- getRed() now returns the red component via hslToRGB(this.values)[0]
- setRed() now converts HSL → RGB, updates the red value, and converts back to HSL
- Example output updated to show 255

### DIFF
--- a/files/en-us/web/javascript/guide/using_classes/index.md
+++ b/files/en-us/web/javascript/guide/using_classes/index.md
@@ -313,14 +313,10 @@ class Color {
     // values is now an HSL array!
     this.values = rgbToHSL([r, g, b]);
   }
-
   getRed() {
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
Nice, thank you

PR #42095, opened 2025-11-25
chore: replace the HTML link with a Markdown link

yin1999:
### Description

replace the HTML link with a Markdown link


estelle (approved):
Thanks! 🎉

PR #42096, opened 2025-11-25
FF146 MathML operator mirroring and stretching

hamishwillee:
FF146 adds support for operator mirroring by default in https://bugzilla.mozilla.org/show_bug.cgi?id=1994172

More correctly, it has supported operator mirroring for some time in normal text for some time but this hasn't worked properly with operators and stretching. Also previously the behaviour was non standard, and now it is part of the spec.

This adds a release note. 
I also update the global `dir` attribute to have a brief section on this topic and an example that shows how it works.
Lastly, I linked to that from the authoring page. This is important because there is only one known font that can be used with this feature.

@fred-wang Could you please sanity check these changes?

Related docs work can be tracked #41878

### DIFF
--- a/files/en-us/web/mathml/reference/global_attributes/dir/index.md
+++ b/files/en-us/web/mathml/reference/global_attributes/dir/index.md
@@ -20,7 +20,7 @@ The **`dir`** [global attribute](/en-US/docs/Web/MathML/Reference/Global_attribu
 - `ltr`
   - : Left to the right rendering of mathematical expressions (e.g., in English or Moroccan).
 - `rtl`
-  - : Right to the left rendering of mathematical expressions (e.g., in Maghreb or Machrek).
+  - : Right to the left rendering of mathematical expressions (e.g., in Arabic, Hebrew, Thaana, Maghreb, Machrek, and so on).
 
 ## Description
 
@@ -42,7 +42,7 @@ Note that effective mirroring and stretching requires a font that includes both 
 
 ### Prefer `dir` to the CSS `direction` property
 
-This attribute can be overridden by the CSS property {{ cssxref("direction") }}, if a CSS page is active and the element supports these properties.
+This attribute can be overridden by the {{ cssxref("direction") }} CSS property, if a CSS page is active and the element supports these properties.
 As the directionality of mathematics is semantically related to its content and not to its presentation, it is recommended that web developers use this attribute instead of the related CSS properties when possible.
 That way, the formulas will display correctly even on a browser that doesn't support CSS or has the CSS deactivated.
 


files/en-us/web/mathml/reference/global_attributes/dir/index.md
```diff
@@ -49,25 +94,161 @@ body {
 </math>
 ```
 
-{{ EmbedLiveSample("Example", "", 150) }}
+{{ EmbedLiveSample("Basic usage", "", 150) }}
 
-## Syntax
+### Mirroring and stretching
```
hamishwillee, line 99: Note, this pulls the example from https://groups.google.com/a/chromium.org/g/blink-dev/c/IpmWeJSPR0g/m/usfX0V9hAQAJ , as suggested by @fred-wang . This compares ltr and rtl of the equation without stretching, and rtl of the equation with stretching.

files/en-us/web/mathml/reference/global_attributes/dir/index.md
```diff
@@ -8,7 +8,52 @@ sidebar: mathmlref
...
+This attribute can be overridden by the CSS property {{ cssxref("direction") }}, if a CSS page is active and the element supports these properties.
+As the directionality of mathematics is semantically related to its content and not to its presentation, it is recommended that web developers use this attribute instead of the related CSS properties when possible.
+That way, the formulas will display correctly even on a browser that doesn't support CSS or has the CSS deactivated.
+
+> [!NOTE]
+> The `dir` attribute is often set to `rtl` in the Arabic-speaking world.
+> However, languages written from right to left often embed mathematical content written from left to right.
+> Consequently, the `auto` keyword from the HTML `dir` attribute is not recognized and by default the [user agent stylesheet](/en-US/docs/Web/CSS/Guides/Cascade/Introduction#user-agent_stylesheets) resets the direction property on the [`math`](/en-US/docs/Web/MathML/Reference/Element/math) element.
+
+## Examples
```
hamishwillee, line 54: FYI Examples are commonly the second item in many/most MathML docs, after the description line.

That is poor design except for trivial features where you don't have to explain much. I have pulled it down to match the WebAPI layout so that I can have multiple examples.

I am fine with that - because there is no MathML template https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types

fred-wang:
cc @eerii

eerii:
Thank you for writing this, the explanation is very complete and accurate. For examples, I was going to suggest the clockwise contour integral (∲), very representative of the `rtlm` feature, which after this patch mirrors correctly. However, Firefox still has a preexisting bug where mirrored characters are not correctly clipped/positioned and it doesn't look quite right yet, so I wouldn't use it in an example for now. So I think the examples are good as is, until we can sort this out.

| Before | Now | Correct |
| --- | --- | --- |
| image | image | image |

files/en-us/web/mathml/reference/global_attributes/dir/index.md
```diff
@@ -8,7 +8,52 @@ sidebar: mathmlref
...
+- _Glyph-level mirroring_: replacing a character with a glyph that can represent it in a mirrored content (if no direct replacement exists).
+  This requires that the font used supports the [Right to left mirror forms (`rtlm`) font feature](https://en.wikipedia.org/wiki/List_of_typographic_features#Features_depending_on_writing_direction).
+
+The base glyph may then be used as a key in the OpenType MathVariant table to get either a larger glyph or a glyph assembly.
+
+Note that effective mirroring and stretching requires a font that includes both a OpenType MathVariant table and support for the `rtlm` font feature, such as XITS.
+
+### Prefer `dir` to the CSS `direction` property
+
+This attribute can be overridden by the CSS property {{ cssxref("direction") }}, if a CSS page is active and the element supports these properties.
```
bsmth: nit: 

```suggestion
This attribute can be overridden by the {{ cssxref("direction") }} CSS property, if a CSS page is active and the element supports these properties.
```

files/en-us/web/mathml/reference/global_attributes/dir/index.md
```diff
@@ -8,7 +8,52 @@ sidebar: mathmlref
...
+As the directionality of mathematics is semantically related to its content and not to its presentation, it is recommended that web developers use this attribute instead of the related CSS properties when possible.
+That way, the formulas will display correctly even on a browser that doesn't support CSS or has the CSS deactivated.
+
+> [!NOTE]
+> The `dir` attribute is often set to `rtl` in the Arabic-speaking world.
```
bsmth, line 50: I guess like so?


```suggestion
> The `dir` attribute is often set to `rtl` by web developers in the Arabic-speaking world.
```

bsmth: And w3 mentions:

>the usual candidates for right-to-left text direction (Arabic, Hebrew, N'Ko, Syriac, Thaana, etc.)

to cover a few more?

https://www.w3.org/International/questions/qa-scripts.en.html#historical

files/en-us/web/mathml/reference/global_attributes/dir/index.md
```diff
@@ -8,7 +8,52 @@ sidebar: mathmlref
...
+<math dir="ltr">
+<math dir="rtl">
+```
+
+### Values
+
+- `ltr`
+  - : Left to the right rendering of mathematical expressions (e.g., in English or Moroccan).
+- `rtl`
+  - : Right to the left rendering of mathematical expressions (e.g., in Maghreb or Machrek).
```
hamishwillee:
```suggestion
  - : Right to the left rendering of mathematical expressions (e.g., in Arabic, Hebrew, Thaana, Maghreb, Machrek, and so on).
```

files/en-us/web/mathml/reference/global_attributes/dir/index.md
hamishwillee: Thanks. I was deliberately not stating any languages since that's not the point of this paragraph. I've added a few more options above in https://github.com/mdn/content/pull/42096/files#r2579140776 . Not adding web developers because no one else is going to be setting this attribute.

hamishwillee:
Thanks @bsmth @eerii for the review, and @fred-wang for all your help.

@eerii Agree it isn't the right time for the ∲ example. he example in place may not show all the flaws in what FF/Chrome/ etc do, but it does show quite a lot of what they do correctly.

fred-wang:
> @eerii Agree it isn't the right time for the ∲ example. he example in place may not show all the flaws in what FF/Chrome/ etc do, but it does show quite a lot of what they do correctly.

@hamishwillee FYI, a patch for the ∲ has landed in https://bugzilla.mozilla.org/show_bug.cgi?id=2008690 ; so Firefox Nightly should render it correctly with the fonts like XITS supporting rtlm (with other fonts, probably we have https://bugzilla.mozilla.org/show_bug.cgi?id=2008688).

hamishwillee:
Thanks @fred-wang - I've added `dev-doc-needed` to that issue - that will ensure that it gets looked at for the next release.

PR #42113, opened 2025-11-25
Update upgrade-insecure-requests note to clarify block-all-mixed-content deprecation.

### LINKED ISSUE #42094: `upgrade-insecure-requests` talks about `block-all-mixed-content` as if it's still a thing
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/upgrade-insecure-requests

_No response_

> Note: The upgrade-insecure-requests directive is evaluated before [block-all-mixed-content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/block-all-mixed-content) and if it is set, the latter is effectively a no-op. It is recommended to set either directive, but not both, unless you want to force HTTPS on older browsers that do not force it after a redirect to HTTP.

As https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/block-all-mixed-content says:
> 🗑️ Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/block-all-mixed-content#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

> ⚠️  Warning: This directive is marked as obsolete in the specification. This directive was previously used to prevent "optionally blockable" mixed content from being fetched insecurely and displayed. Content that isn't blocked is now always upgraded to a secure connection, so this directive is not needed.

I'd appreciate it if the linking page didn't imply that the thing was still alive.

_No response_

_No response_

badwriter123:
Fixes #42094

Hello,.

For the File --> en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/upgrade-insecure-requests

1. I have updated the " NOTE " on the upgrade-insecure-requests page to clarify that block-all-mixed-content is deprecated and should not be used in most of the cases.
2. I rewrote the note to explicitly state that block-all-mixed-content is deprecated.

Thank You... 😊

### DIFF
--- a/files/en-us/web/http/reference/headers/content-security-policy/upgrade-insecure-requests/index.md
+++ b/files/en-us/web/http/reference/headers/content-security-policy/upgrade-insecure-requests/index.md
@@ -14,14 +14,6 @@ have been replaced with secure URLs (those served over HTTPS). This directive is
 intended for websites with large numbers of insecure legacy URLs that need to be
 rewritten.
 
-> [!NOTE]
-> The `upgrade-insecure-requests` directive is evaluated before
-> {{CSP("block-all-mixed-content")}}. If `upgrade-insecure-requests` is set,
-> {{CSP("block-all-mixed-content")}} effectively becomes a no-op. As {{CSP("block-all-mixed-content")}}
-> is now depricated, it is recommended to use `upgrade-insecure-requests` exclusively in most scenarios.
-> Both directives should only be set together if enforcement of HTTPS is required on legacy browsers
-> that do not enforce it following a redirect to HTTP.
-
 The `upgrade-insecure-requests` directive will not ensure that users
 visiting your site via links on third-party sites will be upgraded to HTTPS for the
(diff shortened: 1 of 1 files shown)


wbamberg:
I think it would be better to remove the whole note, and the link in "See also". As far as I can tell, `block-all-mixed-content` is not useful unless `upgrade-insecure-requests` is not supported, and the latter seems to have been supported in all main browsers for almost 10 years.

badwriter123:
Hey @wbamberg, the update makes sense. I'll remove the note and the link and push the updates shortly.

PR #42123, opened 2025-11-25
Rename header Idempotency-Key → Deduplication-Key

ctapobep:
Deduplication doesn't lead to Idempotency. This is a common confusion, popularized by Kafka.

Deduplication is needed when the operation is NOT idempotent. Here's a difference:

1. Client sends Request1 -> the result is State1
2. Client sends Request2 -> the result is State2
3. Client re-sends Request1

Idempotent operation results in State1, while the API w/ Deduplication results in State2.

---

### Additional details

It's easiest to understand Idempotency from the Math perspective. E.g. multiplying by 0 is an idempotent operation. It's not that the 2nd multiplication should be deduplicated, it's that the operation itself has such a property that it can be repeated many times w/o the change of the original result.

---

In this PR I deleted the previous page, but I guess it's better to keep it and just link it to the new one?

### DIFF
--- a/files/en-us/web/http/reference/headers/deduplication-key/index.md
+++ b/files/en-us/web/http/reference/headers/deduplication-key/index.md
@@ -1,18 +1,19 @@
 ---
-title: Idempotency-Key header
-short-title: Idempotency-Key
-slug: Web/HTTP/Reference/Headers/Idempotency-Key
+title: Deduplication-Key header
+short-title: Deduplication-Key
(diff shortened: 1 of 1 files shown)


Josh-Cena:
This is a standards-track feature: https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/. If you want to see it renamed, please propose to IETF or to the proposal champions. This is not the right venue to propose normative changes.

PR #42127, opened 2025-11-26
Revise: Web APIs / HTMLDialogElement Docs

leevigraham:
### Description

* Rewrote html / javascript examples for consistency, added consistent log output.
* Added examples of `returnValue` in close and requestClose
* Added additional context around close requests (linking out to spec) and cancel events

The examples are now much simpler focussing on the api methods. I've removed a lot of form related cruft which may have been confusing to developers new to the APIs.

I'm pretty sure the examples were originally copied from https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#examples

If this PR is successful I'll start on consolidating the examples in the https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog docs.

### Motivation

After finding a broken example in the docs (https://github.com/mdn/content/pull/42053) I found more inconsistencies in the examples across all the Web APIs / HTMLDialogElement docs.

### Additional details

Questions for reviewers… this is my first major PR.

Is there a style guide for when to use links vs code blocks for api methods in related docs? I've seen examples of both `returnValue`, `{{domxref("HTMLDialogElement.returnValue", "returnValue")}}`, and {{domxref("HTMLDialogElement.returnValue")}}.

Is there a style guide for writing code examples? In this case I've stuck with id selectors from the first example I updated, this also makes selection a bit clearer for new developers IMO, but it also may set a bad precedent.

Should the details page for a property include more information that the property on the element page. Example: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby contained more information about how the property worked than the instance property: https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/closedBy. Should these docs be rewritten to reflect each other?

https://github.com/mdn/content/pull/42053

// ping @hamishwillee

### DIFF
--- a/files/en-us/web/api/htmldialogelement/open/index.md
+++ b/files/en-us/web/api/htmldialogelement/open/index.md
@@ -18,7 +18,7 @@ available for interaction.
 A boolean value representing the state of the [`open`](/en-US/docs/Web/HTML/Reference/Elements/dialog#open) HTML attribute. A value of `true` means that the dialog is showing, while `false` means it's not showing.
 
 > [!WARNING]
-> While the `open` property is technically not read-only and can be set directly, doing so is strongly discouraged by [the HTML specification](https://html.spec.whatwg.org/multipage/interactive-elements.html#attr-dialog-closedby), as it can break normal dialog interactions in unexpected ways.
+> While the `open` property is technically not read-only and can be set directly, doing so is strongly discouraged by [the HTML specification](https://html.spec.whatwg.org/multipage/interactive-elements.html#note-dialog-remove-open-attribute), as it can break normal dialog interactions in unexpected ways.
 > For example, the [`close`](/en-US/docs/Web/API/HTMLDialogElement/close_event) event won't fire when programmatically setting `open` to `false`, and subsequent calls to the [`close()`](/en-US/docs/Web/API/HTMLDialogElement/close) and [`requestClose()`](/en-US/docs/Web/API/HTMLDialogElement/requestClose) methods will have no effect.
 > Instead, it's better to use methods such as [`show()`](/en-US/docs/Web/API/HTMLDialogElement/show), [`showModal()`](/en-US/docs/Web/API/HTMLDialogElement/showModal), `close()`, and `requestClose()` to change the value of the `open` attribute.
 


wbamberg:
@hamishwillee , since you looked at the other one, you have better context for this one than me. But LMK if you don't have time and I will have a look.

wbamberg:
Oh, but since you had some questions:

> Is there a style guide for when to use links vs code blocks for api methods in related docs? I've seen examples of both returnValue, {{domxref("HTMLDialogElement.returnValue", "returnValue")}}, and {{domxref("HTMLDialogElement.returnValue")}}.

I'm not sure if there are strict rules here. I would link at least the first mention of the thing in the page, and probably the first mention of the thing in that section (i.e. in that example) but not any subsequent mentions. Whether I would include the interface in the link text or not would depend on how obvious it is from the context.

> Is there a style guide for writing code examples? In this case I've stuck with id selectors from the first example I updated, this also makes selection a bit clearer for new developers IMO, but it also may set a bad precedent.

Yes! https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide. And it does actually [cover this case](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide/CSS#use_class_selectors):

> Use classes for styling, and reserve IDs for non-CSS purposes, **such as for use in JavaScript** ...

...so I would say you are OK to use ID attributes here.

> Should the details page for a property include more information that the property on the element page. Example: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby contained more information about how the property worked than the instance property: https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/closedBy. Should these docs be rewritten to reflect each other?

This came up quite recently. IMO we should have the detailed documentation on the HTML element page, because the DOM API property is just defined as a reflection of the HTML element. If we document it to the same level in both places we will end up with a LOT of duplication (there are hundreds of these pages).

leevigraham:
> I'm not sure if there are strict rules here. I would link at least the first mention of the thing in the page, and probably the first mention of the thing in that section (i.e. in that example) but not any subsequent mentions. Whether I would include the interface in the link text or not would depend on how obvious it is from the context.

I tired to stick to something like that.

If I'm referencing the same method on it's page it was  `code` markup, If I'm referencing a property or method that relates to the element / interface I was using a link without the interface. If it was a property or method that didn't relate to the element / interface I was using a link with the interface. I didn't really see it as a problem if the same property / method was linked multiple times as it made it easier to navigate without going back up to the original reference.

> so I would say you are OK to use ID attributes here.

Perfect. I see that the HTML guide says use kebab-case so I'll update the examples.

> This came up quite recently. IMO we should have the detailed documentation on the HTML element page, because the DOM API property is just defined as a reflection of the HTML element.

Was there a GitHub discussion / ticket I can take a look at?

I can see it from both sides. DOM API properties are set as attributes. I think it's the case that all attributes have a matching DOM property, but not all DOM properties have a matching attribute?

There's always a page for the Element which lists the attributes in a section. This list could be quite long depending on the Element. 

There's always a page for the DOM Property which is self contained.

My expectation is:

* The element page lists the attribute, gives a short, high-level description, and links to the DOM property.
* The DOM property page is the canonical place for the full behaviour, including defaults, state transitions, edge cases, and examples (this might not make sense for more declarative attributes like `popover` `commandfor` and `command`).

Therefore it makes more sense to me that the Element page is the overview of all the attributes with an overview of each attribute, and a link to the DOM Property to dig deeper. This would make the Element page more scannable / readable and allow for more examples on the DOM Property page.

>  If we document it to the same level in both places we will end up with a LOT of duplication (there are hundreds of these pages).

That I agree with :)

leevigraham:
Updated all examples to use consistent `kebab-case` ids.

I've also realised to keep everything in this section consistent I'll also need to revise: https://github.com/mdn/content/blob/main/files/en-us/web/api/htmldialogelement/index.md

wbamberg:
> Was there a GitHub discussion / ticket I can take a look at?

It came up here: https://github.com/mdn/content/pull/40260/files/2d031a2511f2a2a2f2fa9cf770dc7df049925836#r2196957148 .

leevigraham:
@hamishwillee @wbamberg This PR is ready for review.

I've rewritten all the examples for consistency. This mainly means:

* simplification, removing unused elements like forms when were specifically documenting js apis
* Adding a form example for the cancel related documentation
* clear naming for buttons, ids
* hidden logging javascript for output in docs

I've expanded the cancel event and return value docs with references to a the spec "close request"

wbamberg:
@leevigraham , you don't need to merge main into this PR whenever main changes.

hamishwillee:
Thanks. I'll look at this today. 

> IMO we should have the detailed documentation on the HTML element page, because the DOM API property is just defined as a reflection of the HTML element. 

FWIW I don't completely agree with this, mostly because I find the HTML docs less easy to digest. Doesn't matter though - this is the agreed thing and trying to change it is not justifiable.
It isn't entirely true either - there are exceptions and oddities - such as the aria reflected attributes that take references.

I'm not really sure where the level of detail should be in Web API given the above. The doc still needs to be self-contained enough to be useful.

I treat the templates as a good guide for broad structure https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types

files/en-us/web/api/htmldialogelement/cancel_event/index.md
```diff
@@ -8,11 +8,15 @@ browser-compat: api.HTMLDialogElement.cancel_event
 
 {{APIRef("HTML DOM")}}
 
-The **`cancel`** event fires on a {{HTMLElement("dialog")}} element when the user instructs the browser that they wish to dismiss the current open dialog. The browser fires this event when the user presses the <kbd>Esc</kbd> key.
+The **`cancel`** event fires on a {{HTMLElement("dialog")}} element when the user sends a [close request](https://html.spec.whatwg.org/multipage/interaction.html#close-request) to the user agent.
```
hamishwillee: 1. Linking to the spec is frowned upon, unless absolutely necessary. MDN is supposed to be a human readable version of the spec. Defining the close request below does the job nicely.
2. The user doesn't really send anything - the browser does. 
3. Terminology is mixed, but use browser by preference. I tend to use browser agent for web APIs that also exist in node/deno.

Propose

```suggestion
The **`cancel`** event fires on a {{HTMLElement("dialog")}} element when the user triggers a close request.
```

files/en-us/web/api/htmldialogelement/cancel_event/index.md
```diff
@@ -8,11 +8,15 @@ browser-compat: api.HTMLDialogElement.cancel_event
...
 
-This event is cancelable but can not bubble.
+Examples of close requests are:
```
hamishwillee:
```suggestion
Close requests might be triggered by:
```

files/en-us/web/api/htmldialogelement/cancel_event/index.md
```diff
@@ -32,55 +36,80 @@ A generic {{domxref("Event")}}.
 
 ### Canceling a dialog
 
+The following example shows a button that, when clicked, opens a {{htmlelement("dialog")}} via the {{domxref("HTMLDialogElement.showModal()", "showModal()")}} method.
+
+From there you can will trigger the `cancel` event by either clicking _Request Close_ button to close the dialog (via the {{domxref("HTMLDialogElement.requestClose()", "requestClose()")}} method) or press the <kbd>Esc</kbd> key.
```
hamishwillee:
```suggestion
From there you can trigger the `cancel` event by either clicking _Request Close_ button to close the dialog (via the {{domxref("HTMLDialogElement.requestClose()", "requestClose()")}} method), or press the <kbd>Esc</kbd> key.
```

files/en-us/web/api/htmldialogelement/cancel_event/index.md
```diff
@@ -32,55 +36,80 @@ A generic {{domxref("Event")}}.
...
+```html hidden
+<pre id="status-text"></pre>
 ```
 
 ```css hidden
-button,
-div {
-  margin: 0.5rem;
+#status-text {
+  height: 120px;
```
hamishwillee: Consider doubling this as well

files/en-us/web/api/htmldialogelement/cancel_event/index.md
```diff
@@ -32,55 +36,80 @@ A generic {{domxref("Event")}}.
 
 ### Canceling a dialog
 
+The following example shows a button that, when clicked, opens a {{htmlelement("dialog")}} via the {{domxref("HTMLDialogElement.showModal()", "showModal()")}} method.
+
+From there you can trigger the `cancel` event by either clicking _Request Close_ button to close the dialog (via the {{domxref("HTMLDialogElement.requestClose()", "requestClose()")}} method), or press the <kbd>Esc</kbd> key.
+
+Preventing the dialog from closing is demonstrated with a checkbox.
```
hamishwillee: This is obviously a lot better than no explanation!

If I am doing a  method-specific document, I tend to structure like "
This example shows how to cancel a dialog, and how you can also prevent a dialog from closing by ...
Then either have detailed "how" or interleave in the text.

However in this case we have a number of examples that seem to be the same - such as requestClose. In that case I might keep your current structure, but in `requestClose()` remove the example, and just have "See the [`cancel` event](LINK) example for blah blah).

Thoughts?

files/en-us/web/api/htmldialogelement/cancel_event/index.md
```diff
@@ -32,55 +36,80 @@ A generic {{domxref("Event")}}.
...
-  dialog.close();
+dialog.addEventListener("close", (event) => {
+  log("Dialog closed");
 });
 ```
 
 #### Result
 
-{{ EmbedLiveSample('Canceling a dialog', '100%', '100px') }}
+{{ EmbedLiveSample('Canceling a dialog', '100%', '250px') }}
```
hamishwillee, line 121: Interestingly, I'm showing Chrome sending a `close` event for the Escape key, while FF follows the spec and sends a `cancel` events. We should update the compatibility data to reflect that.

files/en-us/web/api/htmldialogelement/close/index.md
```diff
@@ -29,68 +29,74 @@ None ({{jsxref("undefined")}}).
 
 ## Examples
 
-The following example shows a simple button that, when clicked, opens a {{htmlelement("dialog")}} containing a form via the `showModal()` method.
-From there you can click the _X_ button to close the dialog (via the `HTMLDialogElement.close()` method), or submit the form via the submit button.
+The following example shows a button that, when clicked, opens a {{htmlelement("dialog")}} via the {{domxref("HTMLDialogElement.showModal()", "showModal()")}} method.
+From there you can click the either _Close_ button to close the dialog (via the `close()` method).
+
+The _Close_ button closes the dialog without a {{domxref("HTMLDialogElement.returnValue", "returnValue")}}, while the _Close w/ return value_ button closes the dialog with a {{domxref("HTMLDialogElement.returnValue", "returnValue")}}.
```
hamishwillee, line 40: The template likes a level 3 heading for Web API examples, or it should :-)

```suggestion
### Closing a dialog

The following example demonstrates how you can close a dialog, optionally passing a return value.

It displays a button that can be clicked to open a {{htmlelement("dialog")}} by calling the {{domxref("HTMLDialogElement.showModal()", "showModal()")}} method.

The dialog that is displayed can be closed by clicking either _Close_ button.
The _Close_ button closes the dialog without a {{domxref("HTMLDialogElement.returnValue", "returnValue")}}, while the _Close w/ return value_ button closes the dialog with a {{domxref("HTMLDialogElement.returnValue", "returnValue")}}.
```

files/en-us/web/api/htmldialogelement/close/index.md
```diff
@@ -9,7 +9,7 @@ browser-compat: api.HTMLDialogElement.close
 {{ APIRef("HTML DOM") }}
 
 The **`close()`** method of the {{domxref("HTMLDialogElement")}} interface closes the {{htmlelement("dialog")}}.
-An optional string may be passed as an argument, updating the `returnValue` of the dialog.
+An optional string may be passed as an argument, updating the {{domxref("HTMLDialogElement.returnValue", "returnValue")}} of the dialog.
```
hamishwillee, line 13: Worth noting what event(s) is/are fired? Can closing be prevented?

hamishwillee: ... and adding those relevant events as See also section at end.

files/en-us/web/api/htmldialogelement/close_event/index.md
```diff
@@ -33,46 +33,57 @@ A generic {{domxref("Event")}}.
 #### HTML
```
hamishwillee, line 40: Let's change `### Live example` above to a better name - up to you - but perhaps "Handling close events". Note that the example macro needs a corresponding name

hamishwillee: Did you want to explain it? In particular perhaps link to information on ` <form method="dialog">` since it isn't clear why closing via method dialog is interesting, other that we have a button for it.

files/en-us/web/api/htmldialogelement/closedby/index.md
```diff
@@ -16,11 +16,18 @@ The **`closedBy`** property of the
 A string; possible values are:
 
 - `any`
-  - : The dialog can be dismissed with a light dismiss user action, a platform-specific user action, or a developer-specified mechanism.
+  - : The dialog can be dismissed by a [close request](https://html.spec.whatwg.org/multipage/interaction.html#close-request) or clicks outside the dialog.
```
hamishwillee: No spec links

hamishwillee: Also, I would not purge the term "light dismiss" - its in the spec and other docs. You could define it if you want. 

I'd be tempted to revert this. If you want the pattern then maybe define earlier what the different close options are, such as close request and light dismiss and platform-specific user action.

hamishwillee: Ditto below

files/en-us/web/api/htmldialogelement/closedby/index.md
```diff
@@ -16,11 +16,18 @@ The **`closedBy`** property of the
...
 - `closerequest`
-  - : The dialog can be dismissed with a platform-specific user action or a developer-specified mechanism.
+  - : The dialog can only be dismissed by a [close request](https://html.spec.whatwg.org/multipage/interaction.html#close-request).
 - `none`
-  - : The dialog can only be dismissed with a developer-specified mechanism.
+  - : No user actions automatically close the dialog. The dialog can only be dismissed with a developer-specified mechanism.
```
hamishwillee: An improvement, but perhaps be even more specific - such as

```suggestion
  - : The dialog can only be dismissed with a developer-specified mechanism. User actions such as clicking outside the dialog have no effect.
```

files/en-us/web/api/htmldialogelement/closedby/index.md
```diff
@@ -30,14 +37,14 @@ A string; possible values are:
 <dialog open closedby="any">
   <h2>My dialog</h2>
   <p>
-    Closable using the Esc key, or by clicking outside the dialog. "Light
-    dismiss" behavior.
+    Closable using the <kbd>Esc</kbd> key, or by clicking outside the dialog.
+    "Light dismiss" behavior.
```
hamishwillee:
```suggestion
    Closable using the <kbd>Esc</kbd> key, or by clicking outside the dialog ("light dismiss").
```

files/en-us/web/api/htmldialogelement/index.md
```diff
@@ -19,7 +19,7 @@ The **`HTMLDialogElement`** interface provides methods to manipulate {{HTMLEleme
 _Also inherits properties from its parent interface, {{domxref("HTMLElement")}}._
 
 - {{domxref("HTMLDialogElement.closedBy")}}
-  - : A string that sets or returns the [`closedby`](/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby) attribute value of the `<dialog>` element, which indicates the types of user actions that can be used to close the dialog.
+  - : A string that sets or returns the [`closedby`](/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby) HTML attribute, indicating the types of user actions that can be used to close the dialog.
```
hamishwillee, line 22: Perhaps remove "user", since some are not user actions.

```suggestion
  - : A string that sets or returns the [`closedby`](/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby) HTML attribute, indicating the types of actions that can be used to close the dialog.
```

files/en-us/web/api/htmldialogelement/index.md
```diff
@@ -30,13 +30,13 @@ _Also inherits properties from its parent interface, {{domxref("HTMLElement")}}.
 _Also inherits methods from its parent interface, {{domxref("HTMLElement")}}._
 
 - {{domxref("HTMLDialogElement.close()")}}
-  - : Closes the dialog. An optional string may be passed as an argument, updating the `returnValue` of the dialog.
+  - : Closes the dialog. An optional string may be passed as an argument, updating the {{domxref("HTMLDialogElement.returnValue", "returnValue")}} of the dialog.
```
hamishwillee, line 33: I might instead link to the anchor `#htmldialogelement.returnvalue` of the item above.  There is no standard on this, but I have a slight preference to stay on the same page if I can get the information just above. Entirely up to you.

files/en-us/web/api/htmldialogelement/index.md
```diff
@@ -30,13 +30,13 @@ _Also inherits properties from its parent interface, {{domxref("HTMLElement")}}.
...
 - {{domxref("HTMLDialogElement.requestClose()")}}
-  - : Requests to close the dialog. An optional string may be passed as an argument, updating the `returnValue` of the dialog.
+  - : Requests to close the dialog. An optional string may be passed as an argument, updating the {{domxref("HTMLDialogElement.returnValue", "returnValue")}} of the dialog.
 - {{domxref("HTMLDialogElement.show()")}}
   - : Displays the dialog modelessly, i.e., still allowing interaction with content outside of the dialog.
 - {{domxref("HTMLDialogElement.showModal()")}}
-  - : Displays the dialog as a modal, over the top of any other dialogs that might be present. Everything outside the dialog are [inert](/en-US/docs/Web/API/HTMLElement/inert) with interactions outside the dialog being blocked.
+  - : Displays the dialog as a modal, over the top of any other dialogs that might be present. Everything outside the dialog is {{DOMxRef("HTMLElement.inert", "inert")}} with interactions outside the dialog being blocked.
```
hamishwillee, line 39: FYI, Web API does not care if you use the link form or the macro. So you could have just done `inert` as the change.

files/en-us/web/api/htmldialogelement/index.md
```diff
@@ -45,124 +45,139 @@ _Also inherits events from its parent interface, {{DOMxRef("HTMLElement")}}._
 Listen to these events using {{DOMxRef("EventTarget.addEventListener", "addEventListener()")}} or by assigning an event listener to the `oneventname` property of this interface.
 
 - {{domxref("HTMLDialogElement/cancel_event", "cancel")}}
-  - : Fired when the dialog is requested to close, whether with the escape key, or via the `HTMLDialogElement.requestClose()` method.
+  - : Fired when the dialog is requested to close, whether with the escape key, or via the {{domxref("HTMLDialogElement.close()", "requestClose()")}} method.
```
hamishwillee: Do you mean?
```suggestion
  - : Fired when the dialog is requested to close, whether with the escape key, or via the {{domxref("HTMLDialogElement.requestClose()", "requestClose()")}} method.
```

files/en-us/web/api/htmldialogelement/index.md
```diff
@@ -45,124 +45,139 @@ _Also inherits events from its parent interface, {{DOMxRef("HTMLElement")}}._
...
 - {{domxref("HTMLDialogElement/close_event", "close")}}
-  - : Fired when the dialog is closed, whether with the escape key, the `HTMLDialogElement.close()` method, or via submitting a form within the dialog with [`method="dialog"`](/en-US/docs/Web/HTML/Reference/Elements/form#method).
+  - : Fired when the dialog is closed, whether with the escape key, the {{domxref("HTMLDialogElement.close()", "close()")}} method, or via submitting a form within the dialog with [`method="dialog"`](/en-US/docs/Web/HTML/Reference/Elements/form#method).
```
hamishwillee: Is it fired for `requestClose()`? I "think" that close fires the `close` event, whereas the requestClose fires the cancellable `cancel` event, which then results in the `close` event if not prevented. That should be captured in all this.

files/en-us/web/api/htmldialogelement/index.md
```diff
@@ -45,124 +45,139 @@ _Also inherits events from its parent interface, {{DOMxRef("HTMLElement")}}._
...
 
 ```html hidden
-<pre id="log"></pre>
+<pre id="status-text"></pre>
 ```
 
 ```css hidden
-#log {
-  height: 150px;
+#status-text {
```
hamishwillee: Suggest you revert the change to `status-text` everywhere. The live example guide suggests `#log` and it is all over the docs. Changing it therefore makes things less readable for people who are working in the source - and it isn't visible to users anyway.

files/en-us/web/api/htmldialogelement/show/index.md
```diff
@@ -33,67 +33,34 @@ None ({{jsxref("undefined")}}).
 
 ### Basic usage
 
-The following example shows a simple button that, when clicked, opens a {{htmlelement("dialog")}} containing a form via the `show()` method.
-From there you can click the _Cancel_ button ("X") to close the dialog (via the {{domxref("HTMLDialogElement.close()")}} method), or submit the form via the submit button.
+The following example shows a simple button that, when clicked, opens a {{htmlelement("dialog")}} via the `show()` method.
```
hamishwillee: For the show and showmodal examples I think it is an improvement to show an actual "basic usage" without the selection. 

What I think is problematic, and it predates your changes, is that neither example highlights the difference between modal and non-modal dialogs.

leevigraham:
Thanks for the detailed feedback @hamishwillee. It will better help me understand writing docs in a consistent MDN style. I'll go through it all in the next couple of days.

files/en-us/web/api/htmldialogelement/closedby/index.md
```diff
@@ -30,14 +37,13 @@ A string; possible values are:
 <dialog open closedby="any">
   <h2>My dialog</h2>
   <p>
-    Closable using the Esc key, or by clicking outside the dialog. "Light
-    dismiss" behavior.
+    Closable using the <kbd>Esc</kbd> key, or by clicking outside the dialog ("light dismiss").
```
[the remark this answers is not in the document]
hamishwillee: Cursed HTML linter.

files/en-us/web/api/htmldialogelement/cancel_event/index.md
leevigraham: @hamishwillee The original `requestClose()` `cancel` event documentation "inspired" me to work on refactoring the docs :) 

I think an example in both `requestClose()` `cancel` event documentation is worth having as they demonstrate two different things (method, event). 

To simplify the `cancel` docs I'll remove the checkbox and just log the cancel event. That will simplify the demo UI and still show the event.

files/en-us/web/api/htmldialogelement/cancel_event/index.md
leevigraham: I'm seeing both `cancel` and `close` events  in Chrome 143 (Mac)

image

image

files/en-us/web/api/htmldialogelement/close/index.md
leevigraham: > Worth noting what event(s) is/are fired? 

Added. 

> Can closing be prevented?

No, I don't think it can. `requestClose()` can be prevented in the `cancel` event

> and adding those relevant events as See also section at end.

Added.

I referred to http://localhost:5042/en-US/docs/Web/API/HTMLFormElement/requestSubmit and http://localhost:5042/en-US/docs/Web/API/HTMLFormElement/submit for patterns. Neither of these docs pages have events listed in "See Also".

files/en-us/web/api/htmldialogelement/close_event/index.md
leevigraham: This one was a tricky one. 

On one hand I think the `close` docs should just demonstrate the event… not all the ways it's triggered. The existing documentation patterns suggest that the usage examples should be on https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog

On the other hand I always refer to the technical API docs first.

I'll tweak the docs and go from there.

files/en-us/web/api/htmldialogelement/index.md
leevigraham: I didn't change this from the original docs. The spec mentions "user actions" for the `closedby` attribute: https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element

image

files/en-us/web/api/htmldialogelement/index.md
leevigraham: I think this one is fine. If you're clicking to find more information viewing the expanded docs page makes sense, rather than clicking up to something you've just read to then click through.

files/en-us/web/api/htmldialogelement/index.md
leevigraham: Is there a preference for links vs macros? I would have thought the macro is more flexible in the future if the link html needs to change? Would you prefer a smaller commit when it's just spelling grammar?

files/en-us/web/api/htmldialogelement/index.md
leevigraham: Yep :)

files/en-us/web/api/htmldialogelement/show/index.md
leevigraham: I've added a secondary "Trigger alert" button and documentation on when the "Trigger alert" button can be clicked.

leevigraham:
@hamishwillee I've pushed up changes related to your feedback.

<del> It looks like updating the branch introduced additional changes. Any idea how to resolve this?  I'm tempted to branch of main and copy my changes back over… then force push to my branch?</del>

Force pushed a single commit of my changes.

leevigraham:
@hamishwillee One thing I think can be improved and made more consistent is the format of what gets printed to the `#log`, Is there a guide on this?

files/en-us/web/api/htmldialogelement/index.md
hamishwillee: MDN is a human readable version of the spec - if closedby can be triggered by some non user action then we aren't bound by those words.

hamishwillee:
> Force pushed a single commit of my changes.

Good. FWIW I never ever "update with merge". I always rebase, which ensures that the updates are clean and can be tidied up with out clashes later. Sometimes I will squash commits after rebasing - for example if an image is deleted and subsequently restored. 

> One thing I think can be improved and made more consistent is the format of what gets printed to the #log, Is there a guide on this?

There are two examples of how to create example logs in https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Live_samples - one for an append log and one for a growing log.

FWIW We don't talk about the format. I'd be very careful about formalizing this too far. Even though you can of course add more rules, there is a lot to be said for giving authors the flexibility to write examples how they want and let reviewers decide if the result is clear.

We're all on holiday on and off. I'll try have another look in the next week, but with holidays and other priorities there are no guarantees.

Thanks for working on this, and sorry for the delay!

files/en-us/web/api/htmldialogelement/cancel_event/index.md
```diff
@@ -8,11 +8,15 @@ browser-compat: api.HTMLDialogElement.cancel_event
...
 
-This event is cancelable but can not bubble.
+Close requests might be triggered by:
 
-When a `<dialog>` is dismissed with the <kbd>Esc</kbd> key, both the `cancel` and {{domxref("HTMLDialogElement/close_event", "close")}} events are fired.
+- Pressing the <kbd>Esc</kbd> key on desktop platforms
+- Calling the {{domxref("HTMLDialogElement.requestClose()", "requestClose()")}} method.
+- The back button on mobile platforms
+
+This event is cancelable but cannot bubble.
```
hamishwillee:
```suggestion
This event is cancelable and does not bubble.
```

files/en-us/web/api/htmldialogelement/cancel_event/index.md
hamishwillee: Me too now. Thanks.

files/en-us/web/api/htmldialogelement/index.md
hamishwillee: In Web API there is no preference (CSS section of the docs use the macro, for reasons I consider spurious). The site already does processing of both link forms, and will do so in future if needed.

My take is that since it doesn't matter, you don't change what is there because it adds to the cost of reviewing. That said, 
- in this case I see you did have to make some change, because `inert` was not code styled.
- I also wouldn't undo a change already made.

files/en-us/web/api/htmldialogelement/cancel_event/index.md
```diff
@@ -93,3 +123,4 @@ closeButton.addEventListener("click", () => {
 ## See also
 
 - HTML {{HTMLElement("dialog")}} element
+- The [`Event`](/en-US/docs/Web/API/Event) interface
```
hamishwillee: I'm going to remove this everywhere. It isn't a see also - it is on every page up in the Event type.

files/en-us/web/api/htmldialogelement/cancel_event/index.md
```diff
@@ -8,11 +8,18 @@ browser-compat: api.HTMLDialogElement.cancel_event
 
 {{APIRef("HTML DOM")}}
 
-The **`cancel`** event fires on a {{HTMLElement("dialog")}} element when the user instructs the browser that they wish to dismiss the current open dialog. The browser fires this event when the user presses the <kbd>Esc</kbd> key.
+The **`cancel`** event fires on a {{HTMLElement("dialog")}} element when the user triggers a close request.
 
-This event is cancelable but can not bubble.
+The `cancel` event handler can be used to override the default behavior on receiving a close request, and prevent the dialog from closing.
+If the default behavior is not prevented, the dialog will close and fire a {{domxref("HTMLDialogElement/close_event", "close")}} event.
```
hamishwillee, line 14: FYI I think this is important to note.

files/en-us/web/api/htmldialogelement/close/index.md
hamishwillee: The question is not what other areas do, but what would be useful for readers.

To me it is important that users understand that when you close a dialog this event is fired, giving you a chance to override the default close behaviour, and then the close event is fired. Doesn't have to be in the see also though - I put it in the intro.

files/en-us/web/api/htmldialogelement/closedby/index.md
```diff
@@ -16,33 +16,54 @@ The **`closedBy`** property of the
 A string; possible values are:
 
 - `any`
-  - : The dialog can be dismissed with a light dismiss user action, a platform-specific user action, or a developer-specified mechanism.
+  - : The dialog can be dismissed by a user initiated close request or clicks outside the dialog.
 - `closerequest`
-  - : The dialog can be dismissed with a platform-specific user action or a developer-specified mechanism.
+  - : The dialog can only be dismissed by a user initiated close request.
```
hamishwillee: I don't think we should change any mention of `light dismiss" user action, a platform-specific user action, or a developer-specified mechanism.` They have specific meanings that are not captured by the summary. If possible though, we should link to the definitions.

At least please undo these changes and any other mentions.

hamishwillee (commented):
@leevigraham Thanks for all your fixes. I did a subedit (see last commit). I think this is good enough to go, bar my comment in https://github.com/mdn/content/pull/42127/files#r2652116737

I'm aware some of my changes might be slightly inconsistent. I'd still like to get this in sooner rather than later because I'm away on holiday soon and I don't want this to sit for a month.

Edit, PS, if you are wondering "why" I made particular changes, you are welcome to ask. They might just be to my taste :-0

files/en-us/web/api/htmldialogelement/cancel_event/index.md
leevigraham: Agreed. Understanding close requests and the cancel event was the reason I started on the updating the docs.

leevigraham:
> bar my comment in https://github.com/mdn/content/pull/42127/files#r2652116737

@hamishwillee I'm not sure if Github is linking this up correctly.

Are you referring to the unresolved comment here:

image

image

hamishwillee:
> > bar my comment in https://github.com/mdn/content/pull/42127/files#r2652116737
> 
> @hamishwillee I'm not sure if Github is linking this up correctly.
> 
> Are you referring to the unresolved comment here:

Yes

files/en-us/web/api/htmldialogelement/open/index.md
```diff
@@ -18,7 +18,7 @@ available for interaction.
 A boolean value representing the state of the [`open`](/en-US/docs/Web/HTML/Reference/Elements/dialog#open) HTML attribute. A value of `true` means that the dialog is showing, while `false` means it's not showing.
 
 > [!WARNING]
-> While the `open` property is technically not read-only and can be set directly, doing so is strongly discouraged by [the HTML specification](https://html.spec.whatwg.org/multipage/interactive-elements.html#attr-dialog-closedby), as it can break normal dialog interactions in unexpected ways.
+> While the `open` property is technically not read-only and can be set directly, doing so is strongly discouraged by [the HTML specification](https://html.spec.whatwg.org/multipage/interactive-elements.html#note-dialog-remove-open-attribute), as it can break normal dialog interactions in unexpected ways.
```
leevigraham, line 21: @hamishwillee I've changed this link to point directly to the warning.

files/en-us/web/api/htmldialogelement/closedby/index.md
leevigraham: @hamishwillee These changes have been reverted. I searched the docs for other references to "light dismiss" and couldn't find any links.

There is a description of those terms here: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby but I don't like the idea of linking out to the element page when you're reading the api docs.

hamishwillee: Thanks. Merged for now. We should define the term somewhere and cross link.

leevigraham:
@hamishwillee Thanks for your guidance and patience on my first large PR! Glad we got it over the line.

PR #42128, opened 2025-11-26
Use a better equivalent function to an arrow-function example

youssef-el-atmani:
The original provided example is correct, but I think the `anonymous function` fits better in that context.

### DIFF
--- a/files/en-us/learn_web_development/core/scripting/functions/index.md
+++ b/files/en-us/learn_web_development/core/scripting/functions/index.md
@@ -36,13 +36,13 @@ Another essential concept in coding is **functions**, which allow you to store a
 
 ## Where do I find functions?
 
-In JavaScript, you'll find functions everywhere. In fact, we've been using functions all the way through the course so far; we've just not been talking about them very much. Now is the time, however, for us to start talking about functions explicitly, and really exploring their syntax.
+In JavaScript, you'll find functions everywhere. In fact, we've been using functions throughout the course so far; we've just not been talking about them very much. Now is the time, however, for us to start talking about functions explicitly and exploring their syntax.
 
 Pretty much anytime you make use of a JavaScript structure that features a pair of parentheses — `()` — and you're **not** using a common built-in language structure like a [for loop](/en-US/docs/Learn_web_development/Core/Scripting/Loops#the_standard_for_loop), [while or do...while loop](/en-US/docs/Learn_web_development/Core/Scripting/Loops#while_and_do...while), or [if...else statement](/en-US/docs/Learn_web_development/Core/Scripting/Conditionals#if...else_statements), you are making use of a function.
 
 ## Built-in browser functions
 
-We've used functions built into the browser a lot in this course.
+We've made extensive use of built-in browser functions in this course.
 
 Every time we manipulated a text string, for example:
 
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Thanks, you need the surrounding `map(...)` otherwise it's a syntax error by itself. However @chrisdavidmills may have ideas about rewording it, because currently this still reads a bit weird to me.

files/en-us/learn_web_development/core/scripting/functions/index.md
```diff
@@ -269,9 +269,9 @@ The `map()` method takes each item in the array in turn, passing it into the giv
 So in the example above, `item => item * 2` is the arrow function equivalent of:
 
 ```js
-function doubleItem(item) {
+map((function (item) {
   return item * 2;
-}
+});
```
chrisdavidmills, line 274: There is an extra opening paren in your code, and you also need to include the `originals` object instance for it to work. I'd probably write the whole thing out again, but with the regular anonymous function:

```js
const doubled = originals.map(function (item) {
  return item * 2;
});
```

files/en-us/learn_web_development/core/scripting/functions/index.md
```diff
@@ -269,9 +269,9 @@ The `map()` method takes each item in the array in turn, passing it into the giv
 So in the example above, `item => item * 2` is the arrow function equivalent of:
```
chrisdavidmills:
```suggestion
The arrow function is very concise; rewriting our `map()` code to use a regular anonymous callback function would look like this:
```

files/en-us/learn_web_development/core/scripting/functions/index.md
```diff
@@ -269,9 +269,9 @@ The `map()` method takes each item in the array in turn, passing it into the giv
...
 
 ```js
-function doubleItem(item) {
+map((function (item) {
   return item * 2;
-}
+});
 ```
 
 You can use the same concise syntax to rewrite the `addEventListener` example.
```
chrisdavidmills:
```suggestion
You can use the same concise arrow function syntax to rewrite the `addEventListener()` example.
```

chrisdavidmills (changes requested):
Thanks for this, @youssef-el-atmani! I can see what you are trying to do here — writing the whole thing out and not using a named function is more consistent and less confusing.

This is going in the right direction; I've got a few suggestions for improvement.

files/en-us/learn_web_development/core/scripting/functions/index.md
youssef-el-atmani: Hi @charlesroper, Isn't that tailing coma a syntax error?

chrisdavidmills: It should be a semi-colon.

youssef-el-atmani: There is no curly braces,  I guess a semi-colon will leads to an error.
I tested it in VSCode and I got an error.

I think it should be with **no** `colon` or `semi-colon`, like the following:
```js
textBox.addEventListener("keydown", (event) =>
  console.log(`You pressed "${event.key}".`)
);
```

chrisdavidmills: Ah yes, you are right, good work ;-)

files/en-us/learn_web_development/core/scripting/functions/index.md
```diff
@@ -266,19 +266,19 @@ console.log(doubled); // [2, 4, 6]
...
+});
 ```
 
-You can use the same concise syntax to rewrite the `addEventListener` example.
+You can use the same concise arrow function syntax to rewrite the `addEventListener()` example:
 
 ```js
 textBox.addEventListener("keydown", (event) =>
-  console.log(`You pressed "${event.key}".`),
+  console.log(`You pressed "${event.key}".`)
```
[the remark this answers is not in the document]
youssef-el-atmani: sorry @chrisdavidmills, I think this is taking too much time than needed.
But as you see, I removed the `colon`, and that bot suggests that it should be returned again.

If you think that that bot-suggestion could be safely ignored, and there is nothing to be added for this PR, just give it a happy merging ending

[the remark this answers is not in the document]
chrisdavidmills: No worries at all, @youssef-el-atmani!

I've added `-nolint` to the opening code fence of the code block to stop the linter complaining and trying to add the comma in. It is valid syntax, but it is confusing in this context.

I've also committed a few grammar fixes, as I thought I might as well do while I'm here.

chrisdavidmills (approved):
LGTM

PR #42131, opened 2025-11-26
CSS properties only: Links to macros

estelle:
in the css directory, changed all the CSS links to CSS properties only to macros


bsmth:
Thanks! Going to remove myself from review on this one for now.

PR #42137, opened 2025-11-26
Await `fs.open` call

JesseTG:
### Description

Node.js's documentation [says that](https://nodejs.org/api/fs.html#fspromisesopenpath-flags-mode) `fs.open` returns a `Promise<FileHandle>`, so it needs to be `await`ed here.

### Motivation

To spare readers the pain of wrong documentation.

### Additional details

- https://nodejs.org/api/fs.html#fspromisesopenpath-flags-mode


Josh-Cena (approved):
Nice catch, thank you

PR #42139, opened 2025-11-27
Update Headers.get docs

texonidas:
### Description

Updating the description to be more correct; the existing definition implies that if the header is not present, or if `name` is not a header from the HTTP spec that a `TypeError` will be thrown.

### Motivation

Clarity


sideshowbarker:
Tex, congrats on landing your first docs change here — welcome aboard 🎉

PR #42140, opened 2025-11-27
Correct terminology for contributing sources in metadata

fuzzy-coats:
A single SSRC can have multiple participating users each with their own contributing sources, CSRCs.

### DIFF
--- a/files/en-us/web/api/rtcencodedvideoframe/getmetadata/index.md
+++ b/files/en-us/web/api/rtcencodedvideoframe/getmetadata/index.md
@@ -27,9 +27,9 @@ None.
 An object with the following properties:
 
 - `contributingSources`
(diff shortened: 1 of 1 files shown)


hamishwillee (commented):
IMO is not correct /more ambiguous.

Each source has an SSRC which is an identifier for every packet in the particular RTSP stream, used for syncing timing etc.
A stream might have been created from multiple other streams which are "contributing sources". The `contributingSources` contains that list of these ids - which are SSRCs.

This change makes it look like there is come kind of "other CSRC identifer value" - you can't tell that the ids in this are actually SSRC values, which is the main point.

I've left it open because perhaps we might actually be more specific 

> An {{jsxref("Array")}} of sources (SSRC **ids**) that have contributed to the frame.

hamishwillee:
Closing as no response from poster, and I think this is not an improvement.

PR #42147, opened 2025-11-27
Fix array find grammar

NA-V10:
### Description

This PR fixes a small grammar issue in the `Array.prototype.find()` documentation.  
The sentence “If no values satisfy the testing function…” has been corrected to  
“If no value satisfies the testing function…”, which is accurate because `find()`  
returns only a single element.

### Motivation

Improves correctness and clarity in the documentation by fixing a pluralization error.  
This makes the behavior easier to understand for readers.

### Additional details

Documentation-only change, no examples or code behavior were modified.

None.


Josh-Cena:
See https://github.com/mdn/content/pull/42148

PR #42153, opened 2025-11-27
background-repeat x and y

estelle:
Updated background-repeat references to include background-repeat-x and background-repeat-y as supported, because Safari does support these two properties. Reference pages do need to be written.

https://drafts.csswg.org/css-backgrounds-4/#background-repeat-longhands
https://openwebdocs.github.io/web-docs-backlog/webkit/


bsmth (approved):
LGTM, thanks 👍🏻

PR #42159, opened 2025-11-28
Avoid using the same id as tag name without visible HTML

### LINKED ISSUE #42155: Poor choice of canvas id
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Path2D/Path2D

https://developer.mozilla.org/en-US/docs/Web/API/Path2D/Path2D#examples

Using the id "canvas" (as in `<canvas id="canvas"></canvas>`) might cause some people to create something like 
`<div id="canvas">` and then getting the error `getContext is not a function` when running 
```
const canvas = document.getElementById("canvas");
const ctx = canvas.getContext("2d");
```

I guess I was not the only one running into this error since this answer on StackOverflow [getContext is not a function](https://stackoverflow.com/a/47602558) also got 103 upvotes

`<canvas id="myCanvas">`

or 

`<canvas id="canvasArea">` (so that it's different from [w3schools](https://www.w3schools.com/html/html5_canvas.asp))

_No response_

_No response_

satyam22i:
### Description
This PR updates the `<canvas>` element ID used in the Path2D examples.  
The previous examples used `id="canvas"`, which can lead beginners to accidentally use a non-canvas element with the same ID, resulting in the common error:

`getContext is not a function`

The examples now consistently use `id="myCanvas"`.

### Motivation
Using a more specific and less generic ID helps prevent confusion for new developers.  
It avoids accidental misuse of `<div id="canvas">` and improves the clarity and reliability of the example code.

### Additional details
- Updated the HTML `<canvas>` ID in both examples.  
- Updated corresponding JavaScript `getElementById()` calls.  

Fixes #42155


wbamberg:
There are hundreds of these... should we try to fix them all? There doesn't seem to be a whole lot of point in just fixing one, if we think this is a big enough problem to be worth fixing at all.

Josh-Cena:
Personally +1 that this is a problem. It's not immediately obvious where `getContext` comes from since we are not documenting the canvas element itself. Even just unhiding the HTML code goes a long way.

groda (commented):
- unhide HTML ✔
- use `my-canvas` as id ✔
- use `const` instead of `let`  for variables that aren't going to be changed ✔
Looks good to me

groda:
There's 556 such cases  (excluding archived repositories) across the mdn organization: 
https://github.com/search?q=%22%3Ccanvas+id%3D%5C%22canvas%5C%22%22+org%3Amdn+++NOT+is%3Aarchived&type=code&ref=advsearch

I also searched for `<xx id="xx"` to see if this occurs for other tags across all repos and found that out of a total of 1654 results `canvas` is the major culprit:
```
 % grep -ohrI "<\([a-z]\+\)\s\+id=\"\\1\"" . |sort|uniq -c|sort -rn -k1
1163 <canvas id="canvas"
  97 <div id="div"
  62 <svg id="svg"
  54 <video id="video"
  47 <form id="form"
  35 <output id="output"
  27 <select id="select"
  20 <button id="button"
  19 <dialog id="dialog"
  16 <p id="p"
  16 <b id="b"
  16 <a id="a"
  15 <progress id="progress"
  13 <input id="input"
  12 <rect id="rect"
  12 <mask id="mask"
   7 <ul id="ul"
   6 <body id="body"
   4 <template id="template"
   4 <circle id="circle"
   3 <audio id="audio"
   2 <filter id="filter"
   1 <section id="section"
   1 <meter  id="meter"
   1 <hatch id="hatch"
   1 <g id="g"
   ```

To be really beginner-friendly one could also recommend not to use the same id as the tag name in https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id

Josh-Cena:
I will look at it soon.

Josh-Cena:
I have fixed all cases. I only fixed those where the HTML source is hidden but the reference to the ID in CSS or JS is visible. If the HTML is also visible, it would take less than 3 seconds for readers to figure out what the code means. It is... just basic web programming.

Josh-Cena (approved):

PR #42160, opened 2025-11-28
Add 'json' as a value for Sec-Fetch-Dest

bakkot:
### Description

The [docs for the Sec-Fetch-Dest header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Dest) were missing the valid value `json`. This fixes adds it.

I wasn't totally sure how best to link to the docs for JSON imports. I did something which works but if there's a better way feel free to push it up.

### Motivation

Match spec.

### Additional details

See [fetch spec](https://fetch.spec.whatwg.org/#concept-request-destination) for request destinations, and [fetch metadata spec](https://w3c.github.io/webappsec-fetch-metadata/#sec-fetch-dest-header) for the bit which says that this is reflected in the Sec-Fetch-Dest header.

None that I could find.


wbamberg (approved):
👍 thank you @bakkot !

PR #42164, opened 2025-11-29
 Replace short-note by a detailed section to clarify some vars scope inconsistency

youssef-el-atmani:
The previous note was stating that loops and conditional block scope behaves differently than the function scope, which I think is only true when declaring variables with `var`.

### Motivation

The removed note was accurate when there was only `var`, but not today .. when we almost never use `var`.

### Additional details

The content in this PR is written by @chrisdavidmills in the following comment: https://github.com/mdn/content/pull/42164#pullrequestreview-3524303530

### DIFF
--- a/files/en-us/learn_web_development/core/scripting/functions/index.md
+++ b/files/en-us/learn_web_development/core/scripting/functions/index.md
@@ -467,7 +467,7 @@ Let's look at a real example to demonstrate scoping.
 > [!NOTE]
 > The [ReferenceError: "x" is not defined](/en-US/docs/Web/JavaScript/Reference/Errors/Not_defined) error is one of the most common you'll encounter. If you get this error and you are sure that you have defined the variable in question, check what scope it is in.
 
-#### an aside on loop and conditional scope
+#### An aside on loop and conditional scope
 
 It is worth noting that the scope of values declared inside [conditionals](/en-US/docs/Learn_web_development/Core/Scripting/Conditionals) and [loops](/en-US/docs/Learn_web_development/Core/Scripting/Loops) works the same as function scope when declaring values with `let` and `const`. For example, if you added the following blocks to the above example:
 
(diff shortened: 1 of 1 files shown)


youssef-el-atmani:
Hi @chrisdavidmills,
I think that updated note was a part of an old version of that doc. And when everything was updated,  that note was missed.
What do you think? is that right, or that note is still relevant today?

youssef-el-atmani:
Since this PR isn't reviewed yet, I want to mention that the first thing I was attempting to do for this PR is to remove that note permanently, because it related to `var` which is not used in recent code.
And then I said maybe it worth keeping it for whoever is going to deal with code that is still use the `var`.

I am mentioning that because to me REMOVING is a solid option, although I personally prefer the updated version, that's why I went with it.

youssef-el-atmani:
Hi @chrisdavidmills,
That looks great to me.

I have a little suggestion:

---
#### an aside on loop and conditional scope

It is worth noting that the scope of values declared inside [conditionals](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Conditionals) and [loops](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Loops) works the same as function scope when declaring values with `let` and `const`. For example, if you added the following blocks to the above example:

```js
if(x === 1) {
  const c = 4;
  let d = 5;
}
    
for(let i = 0; i <= 1; i++ ) {
  const e = 6;
  let f = 7;
}
```

Calling `output(c)`, `output(d)`, `output(e)`, and `output(f)` would result in the same **"ReferenceError: [variable-name] is not defined"** error seen earlier. The `output()` function cannot access those variables _c, d, e, and f_ because they are locked inside their own scope.

The legacy `var` keyword works differently. If `c` and `e` were declared using `var`:
```js
if(x === 1) {
  var c = 4;
}
    
for(let i = 0; i <= 1; i++ ) {
  var e = 5;
}
```
they would be hoisted to the global scope; therefore, `output(c)` and `output(e)` would result in their values being output to the console. Variables declared with `var` inside functions, however, still have their scope limited to those functions.

This inconsistency can cause confusion and errors, and is another reason why you should use `let` and `const` instead of `var`.

---

I added a `let` example inside the `for` and `if` code blocks.
and created a code snippet for `var` variables.

I am not sure if the sentence --**The `output()` function cannot access those variables _c, d, e, and f_ because they are locked inside their own scope.**--  is correct, I had to rephrase it to add the `e` and `f` variables

What do you think, are those additions worth to keep, or should I apply the original one you have created?

chrisdavidmills:
> What do you think, are those additions worth to keep, or should I apply the original one you have created?

@youssef-el-atmani I think you should keep the new sentence suggestion simple — "The output() function cannot access these variables because they are locked inside their own scope."

I like the suggestions overall.

PR #42185, opened 2025-12-01
Make a single line of code-example easier to read and understand

youssef-el-atmani:
For learners that never learned DOM, I guess `body.removeChild(panel)` is clearer than `panel.parentNode.removeChild(panel)`.
And I think it fits better with the previous line of code: `body.appendChild(panel)`.

### DIFF
--- a/files/en-us/learn_web_development/core/scripting/build_your_own_function/index.md
+++ b/files/en-us/learn_web_development/core/scripting/build_your_own_function/index.md
@@ -115,7 +115,7 @@ panel.appendChild(closeBtn);
 
 Finally, we call {{domxref("EventTarget/addEventListener", "addEventListener()")}} to add a function that will be called when the user clicks the "close" button. The code will delete the whole panel from the page — to close the message box.
 
-Briefly, the `addEventListener()` method is provided by the button (or in fact, any element on the page) that can be passed a function and the name of an event. In this case, the name of the event is 'click', meaning that when the user clicks the button, the function will run. You'll learn a lot more about events in our [events article](/en-US/docs/Learn_web_development/Core/Scripting/Events). The line inside the function uses the {{domxref("Node.removeChild()")}} DOM API function to specify that we want to remove a specific child element of the HTML element — in this case, the panel `<div>`.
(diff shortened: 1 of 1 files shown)


files/en-us/learn_web_development/core/scripting/build_your_own_function/index.md
```diff
@@ -120,7 +118,7 @@ Finally, we call {{domxref("EventTarget/addEventListener", "addEventListener()")
 Briefly, the `addEventListener()` method is provided by the button (or in fact, any element on the page) that can be passed a function and the name of an event. In this case, the name of the event is 'click', meaning that when the user clicks the button, the function will run. You'll learn a lot more about events in our [events article](/en-US/docs/Learn_web_development/Core/Scripting/Events). The line inside the function uses the {{domxref("Node.removeChild()")}} DOM API function to specify that we want to remove a specific child element of the HTML element — in this case, the panel `<div>`.
```
youssef-el-atmani: Hi @chrisdavidmills, I hope I am not over-calling you.

That sentence:

---
Briefly, the `addEventListener()` method is provided by the button (or in fact, any element on the page) that can be passed a function and the name of an event. In this case, the name of the event is 'click', meaning that when the user clicks the button, the function will run.

---
Looks a bit awkward to me.
Rephrasing it is outside of my scope. And if you think it can use some enhancement, please give it a shot.

chrisdavidmills: OK, how about this:

```suggestion
Briefly, the `addEventListener()` method can be called on any element on the page, and is usually passed two arguments: the name of an event and a function to run when the event occurs. In this case, the event name is `click`, meaning that when the user clicks the button, the function will run. You'll learn a lot more about events in our [events article](/en-US/docs/Learn_web_development/Core/Scripting/Events). The line inside the function uses the {{domxref("Node.removeChild()", "removeChild()")}} method to specify that we want to remove a specific child element of the `<body>` element: in this case, the panel `<div>`.
```

chrisdavidmills: Your new code works fine, but I just wanted to draw your attention to another option — `Element.remove()`. This would be called on the panel `<div>` to make it remove itself.

I don't think either approach is necessarily better, but I wanted to mention it just in case.

youssef-el-atmani: Thanks @chrisdavidmills, the new updated section looks way better.

And thanks for telling me about the `Element.remove()`, I think it's easier to use in practice than the other one `parentElement.removeChild(childElement)` since you only need to know the element you want to remove.

PR #42207, opened 2025-12-03
Improve `typeof undefined` docs

blue2cat:
### Description
Clarify how lexical declarations interact with `typeof undefined` checks when not initialized. 

### Motivation
The [current docs on `undefined`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#typeof_operator_and_undefined) aren't clear about the fact that variables created with `let` and `const` will throw a `ReferenceError` if they're compared using `typeof undefined` before initialization.


Josh-Cena:
I agree with the idea, but I can only review or merge after you make it ready for review.

blue2cat:
@Josh-Cena thanks, I've marked as ready for review.

PR #42208, opened 2025-12-03
Improve the readability of some CSS selectors

youssef-el-atmani:
### Description
Little change.

### Motivation

Improve the **readability** of some `pseudo-classes` by adding to them a preceding `*`.
### Additional details

We learned that in: [Basic selectors](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Basic_selectors#using_the_universal_selector_to_make_your_selectors_easier_to_read) lesson.

### DIFF
--- a/files/en-us/learn_web_development/core/styling_basics/tables/index.md
+++ b/files/en-us/learn_web_development/core/styling_basics/tables/index.md
@@ -284,23 +284,23 @@ Save your code and refresh your browser to see the results.
 Next up, we'll deal with alignment of the different types of data inside their cells. Best practice dictates that you should align text to the left and numbers to the right; the following CSS will achieve that, so add it to the bottom of your CSS file now.
 
 ```css
-tr :nth-child(2),
-tr :nth-child(3) {
+tr *:nth-child(2),
+tr *:nth-child(3) {
   text-align: right;
   width: 15%;
 }
 
-tr :nth-child(1),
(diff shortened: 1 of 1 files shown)


youssef-el-atmani:
Hi @chrisdavidmills, I thought it is a followed practice.

That section in the `Basic Selectors` ends with:
> Although both do the same thing, the readability is significantly improved.

I guess that is true, and specially for quick eye scanning.

I think it would be helpful to add a note that that `syntax` not widely adopted in real life code, but I think that usage is worth keeping. For example for me, I'm happy that I know it, and I will use it for my personal code if possible,  _although its just a personal preference_.

And since it is not adopted by other MDN docs, I agree that this PR should not be merged for consistency.

chrisdavidmills:
> I think it would be helpful to add a note that that `syntax` not widely adopted in real life code, but I think that usage is worth keeping. For example for me, I'm happy that I know it, and I will use it for my personal code if possible, _although its just a personal preference.

Yeah, I think that would be OK. Are you interested in opening a PR to do that?
 
> And since it is not adopted by other MDN docs, I agree that this PR should not be merged for consistency.

Cool, this was my concern.

Since you agree, I'll close it.

youssef-el-atmani:
Yes @chrisdavidmills, I would be happy to open that PR.

Issue #39295, opened 2025-04-27
CSS Anchor Positioning: anchor-scope documentation missing


schalkneethling:
### MDN URL

https://developer.mozilla.org/en-US/search?q=anchor-scope

### What specific section or headline is this issue about?

Entire page

### What information was incorrect, unhelpful, or incomplete?

When looking up `anchor-scope` (part of CSS Anchor Positioning), I found that the documentation is missing from MDN Web Docs. I also searched here on GitHub and could not find an existing or previous issue concerning documentation for this property.

### What did you expect to see?

I was expecting the property to be documented on MDN Web Docs.

### Do you have any supporting links, references, or citations?

While slightly newer than say `anchor-name`, support information for this property is available on CanIUse - https://caniuse.com/?search=anchor-scope

Documentation for the property can be found as part of the CSS Anchor Positioning specification here https://drafts.csswg.org/css-anchor-position-1/#anchor-scope

### Do you have anything more you want to share?

I would be more than happy to assist and getting this one documented but, would need some support as it has been a while :) Thank you!

Josh-Cena:
cc @chrisdavidmills @estelle

LcsGa:
Hey! Is there any news on this?

chrisdavidmills:
Ooops, we forgot about this one. I will put it at the top of my list and try to get it done before the end of the year.

chrisdavidmills:
Thanks for the reminder, @LcsGa!

PR #42227, opened 2025-12-04
Editorial review: Document anchor-scope property

chrisdavidmills:
### Description

Chrome 131 supports the [`anchor-scope`](https://drafts.csswg.org/css-anchor-position-1/#anchor-scope) property; see https://chromestatus.com/feature/5094192052436992. It is also supported in Safari 26, and behind a pref in Fx Nightly.

This PR documents `anchor-scope`.

Fixes https://github.com/mdn/content/issues/39295

### DIFF
--- a/files/en-us/web/css/reference/properties/anchor-scope/index.md
+++ b/files/en-us/web/css/reference/properties/anchor-scope/index.md
@@ -44,11 +44,19 @@ This can be a problem in certain situations. For example, if a document contains
 
 The `anchor-scope` property can fix this problem by limiting the visibility, or "scope", of an `anchor-name` value to a specific subtree. The result is that each positioned element can only be anchored to an element within the same subtree of the element that has the scope set on it.
 
-- `anchor-scope: all` sets the scope so that _any_ `anchor-name` values set in the subtree can only be bound to by positioned elements in the same subtree.
-- `anchor-scope: <dashed-ident>#` sets the scope so that the specified `anchor-name` values, when set in the subtree, can only be bound to by positioned elements in the same subtree.
-- `anchor-scope: none` is the default value; it specifies that no anchor scoping is set.
+- `anchor-scope: all` sets the scope so that _any_ `anchor-name` values set in the subtree can only be bound to by positioned elements in the same subtree. Let's say we include multiple anchors in a document, all with `anchor-name: --my-anchor` set on them, and place them in separate containers. We will then set `anchor-scope: all` on each container. If we then include a positioned element inside one of the containers and give it `--my-anchor` as the value of its `position-anchor` property, it will be positioned relative to the anchor inside the same container.
 
-If you have, for example, three `anchor-name` values set inside a subtree (say, <code>&#8209;&#8209;anchor1</code>, <code>&#8209;&#8209;anchor2</code>, and <code>&#8209;&#8209;anchor3</code>), setting <code>anchor-scope: &#8209;&#8209;anchor1, &#8209;&#8209;anchor2, &#8209;&#8209;anchor3</code> on the top-level element of the subtree would be equivalent to setting `anchor-scope: all`.
+  Furthermore, if we create another positioned element outside the containers and give it the same anchor name, or a different anchor name, it will not be anchor-positioned relative to any of the anchors, regardless of whether the anchors have those anchor names included inside their `anchor-name` values. `anchor-scope: all` limits the anchor scope for containers it is set on for _any_ anchors, regardless of `anchor-name`, to only positioned elements inside the same containers.
+
+- `anchor-scope: <dashed-ident>#` sets the scope so that the specified `anchor-name` values, when set in the subtree, can only be bound to by positioned elements in the same subtree. If we return to the same example described in the previous bullet, but change the `anchor-scope` value set on the containers to `--my-anchor`:
+  - Positioned elements with `position-anchor: --my-anchor` set on them will be limited to the scope imposed by the `anchor-scope` setting. They will only be positioned relative to the anchors if they are placed inside the containers.
+  - However, positioned elements with different `position-anchor` names — for example, `--another-anchor` — _can_ be positioned relative to one of the anchors, whether it is placed inside or outside the containers, provided you add the `--another-anchor` anchor name to the anchor's `anchor-name` property. The `anchor-scope` property only limits the scope for the `--my-anchor` anchor name, so it will have no effect on other anchor names.
+
+    If multiple anchors are given the `--another-anchor` anchor name, the positioned elements with that `position-anchor` value will be positioned relative to the last anchor in the source order with that name.
+
+- `anchor-scope: none` is the default value, which specifies that no anchor scope is set. If multiple anchors exist in a document with the same `anchor-name`, and a positioned element is given this name as the value of its `position-anchor` property, it will be positioned relative to the last anchor element in the source order, regardless of where it is placed in the DOM hierarchy.
+
+If you have, for example, three `anchor-name` values set inside a subtree (say, `--anchor1`, `--anchor2`, and `--anchor3`), setting `anchor-scope: --anchor1, --anchor2, --anchor3` on the top-level element of the subtree would be equivalent to setting `anchor-scope: all`.
 
 Anchor scopes only affect [explicit anchor associations](/en-US/docs/Web/CSS/Guides/Anchor_positioning/Using#explicit_css_anchor_association), that is, those made between an anchor element with an `anchor-name` set on it, and a positioned element referencing the anchor element's anchor name in its `position-anchor` value. Anchor scopes do not affect [implicit anchor associations](/en-US/docs/Web/CSS/Guides/Anchor_positioning/Using#implicit_anchor_association).
 
@@ -183,7 +191,7 @@ Finally, we include a {{htmlelement("form")}} containing three different [`<inpu
 
 #### CSS
 
-We start by specifying our anchor elements as anchors by giving them two {{cssxref("anchor-name")}} values: <code>&#8209;&#8209;my-anchor</code> and `--another-anchor`.
+We start by specifying our anchor elements as anchors by giving them two {{cssxref("anchor-name")}} values: `--my-anchor` and `--another-anchor`.
 
 ```css hidden live-sample___comparing-values
 body {
@@ -213,11 +221,11 @@ body {
 
 .positioned,
 .positioned2 {
-  background: orange;
-  border: 2px solid #ddd;
+  border: 1px solid black;
   border-radius: 3px;
   width: fit-content;
-  padding: 3px;
+  padding: 3px 6px;
+  box-shadow: 3px 3px 3px rgba(0 0 0 / 0.2);
 }
 
 form {
@@ -278,7 +286,7 @@ The example renders as follows:
 Check out the initial positioning effect applied to the positioned elements with `anchor-scope: all` set on the `<section>` elements, and then try selecting the other available `anchor-scope` values to see what their effect is. You should observe the following:
 
 - `all`: The scope for positioning elements relative to anchor elements that are descendants of the `<section>` elements is limited to positioned elements that are themselves descendants of the `<section>` elements, regardless of the `anchor-name` value used to associate them. As a result, the positioned elements inside the `<section>` elements ("Positioned 1–3") are anchor-positioned as expected, but the positioned element outside the `<section>` elements ("Positioned 4") is not anchor-positioned. It is out of scope.
-- `--my-anchor`: The scope for positioning elements relative to anchor elements that are descendants of the `<section>` elements is limited to positioned elements that are themselves descendants of the `<section>` elements, only if the `--my-anchor` `anchor-name` is used to associate them. As a result, the positioned elements inside the `<section>` elements ("Positioned 1–3") are anchor-positioned as expected, and the positioned element outside the `<section>` elements ("Positioned 4") is also anchor-positioned as expected. In the former case, the positioned elements are inside the set scope, and in the latter case, the positioned element is not affected by the set scope, as it is using an anchor name outside the scope (<code>&#8209;&#8209;another-anchor</code>). The "Positioned 4" element is positioned relative to the last anchor element in the source that has the matching anchor name.
+- `--my-anchor`: The scope for positioning elements relative to anchor elements that are descendants of the `<section>` elements is limited to positioned elements that are themselves descendants of the `<section>` elements, only if the `--my-anchor` `anchor-name` is used to associate them. As a result, the positioned elements inside the `<section>` elements ("Positioned 1–3") are anchor-positioned as expected, and the positioned element outside the `<section>` elements ("Positioned 4") is also anchor-positioned as expected. In the former case, the positioned elements are inside the set scope, and in the latter case, the positioned element is not affected by the set scope, as it is using an anchor name outside the scope (`--another-anchor`). The "Positioned 4" element is positioned relative to the last anchor element in the source that has the matching anchor name.
 - `none`: As no anchor scope is set on the `<section>` elements, all of the positioned elements are positioned relative to the last anchor element in the source order.
 
 ## Specifications


files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,76 @@ For example, to stop a customizable `<select>` element's picker from being ancho
...
+  background: #eee;
+}
+
+.anchor {
+  font-size: 1.8rem;
+  color: white;
+  text-shadow: 1px 1px 1px black;
+  background-color: blue;
+  width: fit-content;
+  padding: 3px;
```
estelle, line 178: suggest hiding css that isn't lesson-related.

files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,76 @@ For example, to stop a customizable `<select>` element's picker from being ancho
...
+}
+```
+
+This results in the following positioning behavior:
+
+{{ EmbedLiveSample("anchor-scope", "100%", "225") }}
+
+The first positioned element is positioned relative to the anchor. It is in scope for positioning relative to the `--my-anchor` anchor, as it is inside the `<section>` element where `anchor-scope: --my-anchor` is set. The second positioned element is not positioned relative to the anchor. It is not a descendant of the `<section>` element, so it is outside the anchor scope.
+
+See the {{cssxref("anchor-scope")}} reference page for additional information and examples.
```
estelle:
```suggestion
```

files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,76 @@ For example, to stop a customizable `<select>` element's picker from being ancho
 }
 ```
 
+## Anchor scoping
+
+When multiple anchor elements on a page are given the same {{cssxref("anchor-name")}} value and a positioned element is associated with that anchor name (by specifying the name as its {{cssxref("position-anchor")}} property value), the positioned element will be associated with the _last_ anchor element in the source order with that anchor name.
+
+This can be a problem in certain situations. For example, when you have multiple HTML components repeated on a page, and each one features a positioned element anchored to the component, all of the positioned elements will be anchored to the last component on the page, unless each component uses a different anchor name.
```
estelle:
```suggestion
When multiple anchor elements are given the same {{cssxref("anchor-name")}} value and a positioned element is associated with that anchor name (by specifying the name as its {{cssxref("position-anchor")}} property value), the positioned element will be associated with the _last_ anchor element in the source order with that anchor name.

For example, if a document contains multiple repeated components, each with a positioned element tethered to an anchor, all the positioned elements will be anchored to the last anchor on the page unless each component uses a different anchor name. This is likely not the desired behavior.
```
suggest reordering, and associated the issue example (2nd p) with the issue summary  (1st p) but making it a single paragraph

files/en-us/web/css/guides/anchor_positioning/using/index.md
chrisdavidmills: This bit _is_ all hidden.

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+.positioned {
+  position: absolute;
+  position-anchor: --my-anchor;
+  position-area: right;
+}
+```
+
+#### Result
+
+The examples renders like so:
```
andruud: "example renders"?

or

"examples render"

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+});
+
+sections.forEach((section) => (section.style.anchorScope = "all"));
+```
+
+#### Result
+
+The examples renders like so:
+
+{{ EmbedLiveSample("basic_usage", "100%", "225") }}
```
andruud: I don't know how MDN works, but naively I'd assume this embeds the exact same example as on line 133.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: Oops. Fixed.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: Yeah, this should be "comparing-values". Fixed.

chrisdavidmills:
Thanks for the review, @andruud; I'll move this one on to the editorial review stage now.

files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,74 @@ For example, to stop a customizable `<select>` element's picker from being ancho
 }
 ```
 
+## Anchor scoping
+
+When multiple anchor elements are given the same {{cssxref("anchor-name")}} value and a positioned element is associated with that anchor name (by specifying the name as its {{cssxref("position-anchor")}} property value), the positioned element will be associated with the _last_ anchor element in the source order with that anchor name.
```
estelle:
```suggestion
When multiple anchor elements are given the same {{cssxref("anchor-name")}} value and a positioned element has that name as its {{cssxref("position-anchor")}} property value, the positioned element will be associated with the _last_ anchor element in the source order with that `anchor-name` value.
```
the element is associated with another element, not with the name.
not 100% sure if all the elements have that name, or if really only the last one has a name. the last edit makes finding the answer unnecessary.

files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,74 @@ For example, to stop a customizable `<select>` element's picker from being ancho
...
+
+For example, if a document contains multiple repeated components, each with a positioned element tethered to an anchor, all the positioned elements will be anchored to the last anchor on the page unless each component uses a different anchor name. This is likely not the desired behavior.
+
+The {{cssxref("anchor-scope")}} property can fix this problem by making it so that each positioned element can only be anchored to an element within the same subtree of the element that has the scope set on it:
```
estelle: how about something like:

```md
With the {{cssxref("anchor-scope")}} property, you can limit the visibility, or "scope', of an `anchor-name` value to a specific subtree, which enables the association of multiple same-named anchors with their own positioned elements within a single document:
```

files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,74 @@ For example, to stop a customizable `<select>` element's picker from being ancho
...
+
+- `anchor-scope: all` sets the scope so that _any_ `anchor-name` values set in the subtree can only be bound to by positioned elements in the same subtree.
+- `anchor-scope: --my-anchor` sets the scope so that the specified `anchor-name` value, when set in the subtree, can only be bound to by positioned elements in the same subtree.
```
estelle: suggest mentioning `none` is the default, and somehow alluding to the fact that it can be a comma separated list.

files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,74 @@ For example, to stop a customizable `<select>` element's picker from being ancho
...
+
+For example, let's say you have an anchor and a anchor-positioned element inside a scoped container, and another positioned element outside the container:
```
estelle: the intro says this is useful when we have multiple elements with the same anchor name. Let show an example with multiple anchors and positioned elements. 

maybe use an ordered list with 3 sets, and do something like 

```html
<li>
  <div class="anchor">⚓︎⚓︎⚓︎</div>`
<div class="positioned">Positioned 3</div>
</li>
``` 
as the last in the list.

files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,74 @@ For example, to stop a customizable `<select>` element's picker from being ancho
...
+  background-color: blue;
+  width: fit-content;
+  padding: 3px;
+}
+
+.positioned {
+  background: orange;
+  width: fit-content;
+  padding: 3px;
+  position: absolute;
```
estelle: end the hidden part here (i can't add closing tick marks in this editor)

```suggestion
```css live-sample___anchor-scope
li {
  anchor-scope: --my-anchor;
}
.positioned {
  position: absolute;
```

files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,74 @@ For example, to stop a customizable `<select>` element's picker from being ancho
...
+.anchor {
+  anchor-name: --my-anchor;
+}
+```
+
+```css live-sample___anchor-scope
+.scoped {
+  anchor-scope: --my-anchor;
+}
+```
```
estelle, line 202:
```suggestion
```
moved up

files/en-us/web/css/guides/anchor_positioning/using/index.md
```diff
@@ -119,9 +119,74 @@ For example, to stop a customizable `<select>` element's picker from being ancho
...
+
+This results in the following positioning behavior:
+
+{{ EmbedLiveSample("anchor-scope", "100%", "225") }}
+
+The first positioned element is positioned relative to the anchor. It is in scope for positioning relative to the `--my-anchor` anchor, as it is inside the `<section>` element where `anchor-scope: --my-anchor` is set. The second positioned element is not positioned relative to the anchor. It is not a descendant of the `<section>` element, so it is outside the anchor scope.
```
estelle: update to go with the multiple being associated with their own anchors

files/en-us/web/css/reference/properties/anchor-name/index.md
```diff
@@ -381,6 +381,7 @@ Scroll the page to see how both of the infoboxes are tethered to the anchor.
 
 ## See also
 
+- {{cssxref("anchor-scope")}}
```
estelle: move this down under position-anchor. i think position-anchor will be used more than anchor scope

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
+---
+title: anchor-scope
+slug: Web/CSS/Reference/Properties/anchor-scope
+page-type: css-property
+browser-compat: css.properties.anchor-scope
+sidebar: cssref
+---
+
+The **`anchor-scope`** [CSS](/en-US/docs/Web/CSS) property can be used to limit the scope of the DOM within which positioned elements can be associated with anchor elements to a particular subtree.
```
estelle:
```suggestion
The **`anchor-scope`** [CSS](/en-US/docs/Web/CSS) property can be used to limit the scope in which positioned elements can be associated with anchor elements to one or more subtrees.
```

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+anchor-scope: initial;
+anchor-scope: revert;
+anchor-scope: revert-layer;
+anchor-scope: unset;
+```
+
+### Values
+
+- `none`
+  - : The default value. No anchor scope limiting happens on an element when this value is set.
```
estelle:
```suggestion
  - : No anchor scope limiting happens on an element. This is the default value.
```

if you have `<main><section><artilce><ol><li><div anchor>` and `none` is set on the ol, but `--foo` is set on the li and --bar on the article, which scoping if any is removed? how does nested scoping and removal work?
- let's remove "is set" language from the list. It's a given.

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+- `all`
+  - : Any {{cssxref("anchor-name")}}s declared on an element with this value set and its descendants can only be bound to by positioned elements that are descendants of the same element.
```
estelle: - what does "bound to" mean? what does the "same element" mean here? this is hard to grok.
- remove the "set" verbiage

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+- {{cssxref("dashed-ident")}}#
+  - : One or more comma-separated {{cssxref("dashed-ident")}}s representing anchor names. Any of the specified `anchor-name`s declared on an element with this value set and its descendants can only be bound to by positioned elements that are descendants of the same element.
```
estelle: - define a single ident here. see `animation-timeline` as an example.
- what happens if there no matching anchor-names?

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Description
+
+When multiple anchor elements on a page are given the same {{cssxref("anchor-name")}} value and a positioned element is associated with that anchor name (by specifying the name as its {{cssxref("position-anchor")}} property value), the positioned element will be associated with the _last_ anchor element in the source order with that anchor name.
```
estelle: link "anchor elements" to the guide on defining anchor elements.
same comments as for the guide

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+
+This can be a problem in certain situations. For example, when you have multiple HTML components repeated on a page, and each one features a positioned element anchored to the component, all of the positioned elements will be anchored to the last component on the page, unless each component uses a different anchor name.
+
+The `anchor-scope` property can fix this problem by making it so that each positioned element can only be anchored to an element within the same subtree of the element that has the scope set on it:
```
estelle:
```suggestion
The `anchor-scope` property can be used to limit the anchoring of positioned elements to anchor elements within specific DOM subtrees:
```

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+
+- `anchor-scope: all` sets the scope so that _any_ `anchor-name` values set in the subtree can only be bound to by positioned elements in the same subtree.
```
estelle: move most of the explanation from the example at the end of the page here. 

also, add `none`, mentioning that it's the default, and explain how / what it overrides (see question above)

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+- `anchor-scope: <dashed-ident>#` sets the scope so that the specified `anchor-name` values, when set in the subtree, can only be bound to by positioned elements in the same subtree.
+
+If you have, for example, three `anchor-name` values set inside a subtree (say, <code>&#8209;&#8209;anchor1</code>, <code>&#8209;&#8209;anchor2</code>, and <code>&#8209;&#8209;anchor3</code>), setting <code>anchor-scope: &#8209;&#8209;anchor1, &#8209;&#8209;anchor2, &#8209;&#8209;anchor3</code> on the top-level element of the subtree would be equivalent to setting `anchor-scope: all`.
```
estelle: use ticks and dashes
not sure if you're saying 3 same-named and no other anchors?

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+
+Anchor scopes only affect **explicit anchor associations**, that is, those made between an anchor element with an `anchor-name` set on it, and a positioned element referencing the anchor element's anchor name in its `position-anchor` value. Anchor scopes do not affect [implicit anchor associations](/en-US/docs/Web/CSS/Guides/Anchor_positioning/Using#implicit_anchor_association).
```
estelle:
```suggestion
Anchor scopes only affect **[explicit anchor associations](/en-US/docs/Web/CSS/Guides/Anchor_positioning/Using#explicit_css_anchor_association)**, made between an anchor element with an `anchor-name` set on it, and a positioned element referencing the anchor element's anchor name in its `position-anchor` value. Anchor scopes do not affect  unnamed [implicit anchor associations](/en-US/docs/Web/CSS/Guides/Anchor_positioning/Using#implicit_anchor_association).
```
not 100% sure if "unnamed" is correct, but i think it is

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+
+For more information on anchor features and usage, see the [CSS anchor positioning](/en-US/docs/Web/CSS/Guides/Anchor_positioning) module landing page and the [Using CSS anchor positioning](/en-US/docs/Web/CSS/Guides/Anchor_positioning/Using) guide.
```
estelle:
```suggestion
For more information on anchor features and usage, see the [CSS anchor positioning](/en-US/docs/Web/CSS/Guides/Anchor_positioning) module and the [Using CSS anchor positioning](/en-US/docs/Web/CSS/Guides/Anchor_positioning/Using) guide.
```

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Formal syntax
+
+{{csssyntax}}
+
+## Examples
+
+### Basic usage
+
+This example shows that, when an `anchor-scope` is set on an element, anchor element descendants of that element with `anchor-name`s that match the names specified in the `anchor-scope`, can only be bound to by positioned elements inside that same subtree.
```
estelle: - either "`anchor-scope`" or "an anchor scope"
- i would break this into two or three sentences. it's a lot to parse.

I think a better demonstration would be something similar to what i suggested in the guide-- with several anchors and several anchor positioned elements, all with the same anchor name.

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+We handle setting `anchor-scope` on the `<section>` elements using JavaScript. First of all, we grab references to the `<section>` and `<form>` elements. We then add an `input` event listener to the form so that when any of its descendant radio inputs are selected, the `anchor-scope` of all the `<section>` elements is set to the radio input value.
+
+Finally, we set the `anchor-scope` of all the `<section>` elements to `all` on page load as an initial value (that is also the initial selected radio button value).
+
+```js live-sample___comparing-values
+const sections = document.querySelectorAll("section");
+const form = document.querySelector("form");
+
+form.addEventListener("input", (e) => {
+  sections.forEach((section) => (section.style.anchorScope = e.target.value));
```
estelle: I think we can hide this whole thing for brevity, but this is not what i was expecting to see. i would have thought you would update the `anchor-scope` value of the `.scope` class. That said, if we do keep this js, let's make it DRY; make 265 a function call: "`updateScoped(e.target.value)`, make 268 a function call `updateScoped("all")`, and add the updateScoped function 

```js
function updateScope(val) {
   sections.forEach((section) => (section.style.anchorScope = val));
   }
```

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+We specify three {{htmlelement("section")}} elements that we will set an anchor scope on. Each `<section>` contains two {{htmlelement("div")}} elements, one to be turned into an anchor, and one to be an anchor-positioned element.
+
+We also include an additional `<div>` outside the `<section>` elements, which we will also turn into an anchor-positioned element. This will have different anchor positioning settings from the others.
+
+Finally, we include a {{htmlelement("form")}} containing three different [`<input type="radio">`](/en-US/docs/Web/HTML/Reference/Elements/input/radio) elements to allow different `anchor-scope` values to be set on the `<section>` elements.
+
+```html live-sample___comparing-values
+<section class="scoped">
+  <div class="anchor">⚓︎</div>
+  <div class="positioned">Positioned 1</div>
```
estelle, line 163: change to 1, 2, 3, 4

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+<div class="positioned2">Positioned 2</div>
+
+<form>
+  <fieldset>
+    <legend>Select <code>anchor-scope</code></legend>
+
+    <input type="radio" id="all" name="scope" value="all" checked />
+    <label for="all"><code>all</code></label>
+    <input type="radio" id="--my-anchor" name="scope" value="--my-anchor" />
+    <label for="all"><code>--my-anchor</code></label>
```
estelle:
```suggestion
    <label for="--my-anchor"><code>--my-anchor</code></label>
```
what happens if we do `--my-anchor, --another-anchor` too?

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+<section class="scoped">
+  <div class="anchor">⚓︎</div>
+  <div class="positioned">Positioned 1</div>
+</section>
+
+<div class="positioned2">Positioned 2</div>
+
+<form>
+  <fieldset>
+    <legend>Select <code>anchor-scope</code></legend>
```
estelle:
```suggestion
    <legend>Select an <code>anchor-scope</code> value</legend>
```

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+    <label for="all"><code>--my-anchor</code></label>
+    <input type="radio" id="none" name="scope" value="none" />
+    <label for="none"><code>none</code></label>
+  </fieldset>
+</form>
+```
+
+#### CSS
+
+We start by specifying our anchor elements as anchors by giving them two {{cssxref("anchor-name")}} values: <code>&#8209;&#8209;my-anchor</code> and `--another-anchor`.
```
estelle: same (ticks and dashes)

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+  color: white;
+  text-shadow: 1px 1px 1px black;
+  background-color: blue;
+  width: fit-content;
+  padding: 3px;
+}
+
+.positioned,
+.positioned2 {
+  background: orange;
```
estelle: make this 50% opaque so when they're on top of each other, the numbers (that we changed per the recommended edits in the HTML) show thru and you can really see they're on top of each other.

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+
+```css live-sample___comparing-values
+.anchor {
+  anchor-name: --my-anchor, --another-anchor;
+}
+```
+
+Next, we position our `.positioned` elements relative to an anchor element. We absolutely position them, give them a {{cssxref("position-anchor")}} value of `--my-anchor` to associate them with an anchor, and position them relative to the anchor with a {{cssxref("position-area")}} value of `right`.
+
+The `.positioned2` element is positioned in a similar way, except that it is given the other available anchor name as its `position-anchor` value — `--another-anchor` — and it is positioned to the `bottom` of the anchor instead. We also give it a {{cssxref("bottom")}} value of `5px`, so that if the anchor positioning is not working, it will be positioned to the bottom of the `<body>`.
```
estelle: explain why you're including this second positioned element and how you are expecting it to behave and why.

files/en-us/web/css/reference/properties/anchor-scope/index.md
estelle: alternatively, you could 

1. remove the `checked` in the HTML and force the event by adding some js that checks the all value.
2. remove the `checked`, allowing it to default to `none`, and remove line 68 and the language that says it is set to all on line 277

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+- `--my-anchor`: The scope for positioning elements relative to anchor elements that are descendants of the `<section>` elements is limited to positioned elements that are themselves descendants of the `<section>` elements, only if the `--my-anchor` `anchor-name` is used to associate them. As a result, the positioned elements inside the `<section>` elements ("Positioned 1") are anchor-positioned as expected, and the positioned element outside the `<section>` elements ("Positioned 2") is also anchor-positioned as expected. In the former case, the positioned elements are inside the set scope, and in the latter case, the positioned element is not affected by the set scope, as it is using an anchor name outside the scope (<code>&#8209;&#8209;another-anchor</code>). The "Positioned 2" element is positioned relative to the last anchor element in the source that has the matching anchor name.
```
estelle: same (code/dash)

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+- `none`: No anchor scope is set on the `<section>` elements. As a result, all of the positioned elements are positioned relative to the last anchor element in the source order.
```
estelle:
```suggestion
- `none`: As no anchor scope is set on the `<section>` elements, all of the positioned elements are positioned relative to the last anchor element in the source order.
```

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+
+{{ EmbedLiveSample("basic_usage", "100%", "225") }}
+
+Note how the first positioned element is positioned to the right of the anchor. It is in scope for positioning relative to the `--my-anchor` anchor, as it is inside the `<section>` element where the `anchor-scope: --my-anchor` is set.
+
+However, the second positioned element is not positioned relative to the anchor. It is not a descendant of the `<section>` element, so it is outside the anchor scope.
+
+### Comparing different `anchor-scope` values
+
+This example allows you to apply different `anchor-scope` values to multiple containers, to demonstrate the effects of each.
```
estelle: maybe something like:
```
This example demonstrates the different `anchor-scope` values by allowing you to apply different `anchor-scope` values to multiple containers, all containing anchors with the same `anchor-name` value.
```

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,296 @@
...
+
+#### Result
+
+The example renders as follows:
+
+{{ EmbedLiveSample("comparing-values", "100%", "225") }}
+
+Check out the initial positioning effect applied to the positioned elements with `anchor-scope: all` set on the `<section>` elements, and then try selecting the other available `anchor-scope` values to see what their effect is. You should observe the following:
+
+- `all`: The scope for positioning elements relative to anchor elements that are descendants of the `<section>` elements is limited to positioned elements that are themselves descendants of the `<section>` elements, regardless of the `anchor-name` value used to associate them. As a result, the positioned elements inside the `<section>` elements ("Positioned 1") are anchor-positioned as expected, but the positioned element outside the `<section>` elements ("Positioned 2") is not anchor-positioned. It is out of scope.
```
estelle: move most of these explanations, made more generic, into the description area values section.

files/en-us/web/css/reference/properties/position-anchor/index.md
```diff
@@ -297,6 +297,7 @@ Select different values from the drop-down menus to change the anchors that the
 
 ## See also
 
+- {{cssxref("anchor-scope")}}
 - {{cssxref("anchor-name")}}
```
estelle: invert these two

files/en-us/web/css/reference/properties/position-anchor/index.md
```diff
@@ -297,6 +297,7 @@ Select different values from the drop-down menus to change the anchors that the
 
 ## See also
 
+- {{cssxref("anchor-scope")}}
```
estelle: make this one second

estelle (commented):
sorry, didn't realize this was ready for review.

files/en-us/web/css/guides/anchor_positioning/using/index.md
chrisdavidmills: I like the first half of the suggestion, but I felt like the second half was a bit confusing. In the next commit, I've updated it to:

> The {{cssxref("anchor-scope")}} property can fix this problem by limiting the visibility, or "scope", of an `anchor-name` value to a specific subtree. The result is that each positioned element can only be anchored to an element within the same subtree of the element that has the scope set on it:

files/en-us/web/css/guides/anchor_positioning/using/index.md
chrisdavidmills: Good call. I've updated the second bullet to indicate that multiple comma-separated values can be set, and added a third bullet to explain the `none` value.

files/en-us/web/css/guides/anchor_positioning/using/index.md
chrisdavidmills: OK, I've rewritten the example to include three sets.

files/en-us/web/css/guides/anchor_positioning/using/index.md
chrisdavidmills: done

files/en-us/web/css/guides/anchor_positioning/using/index.md
chrisdavidmills: done

files/en-us/web/css/guides/anchor_positioning/using/index.md
chrisdavidmills: updated

files/en-us/web/css/reference/properties/anchor-name/index.md
chrisdavidmills: done

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I don't think "one or more substrees" is correct — each positioned element inside a scoped container can be allowed to position itself relative to one or more anchors inside that same scoped container, but you can't allow a positioned element to be positionable relative to anchors in multiple separate scoped containers.

I've updated it to hopefully clarify:

> The **`anchor-scope`** [CSS](/en-US/docs/Web/CSS) property can be used to limit the scope in which a positioned element can be associated with anchor elements to a particular subtree.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I've made the language suggestion; that was a good call.

In terms of your question, the `anchor-scope` set at each hierarchy level would prevail for that level, so the article would keep its scope, but then anything inside its child ol would have its scoping removed (but a different child element of article, say, a p element, would follow the scope of its parent.) Then when you get to the li, it has a separate scope applied, which would apply to it and its child.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: Bah, I've confused myself here. The explanations of these values in the usage guide are much better, so I've duplicated them here as well.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: But there can be more than one.

If there are no matching anchor names, the anchors are not affected by the scoping.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: Updates made as per guide; link added.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I've updated this as per the guide.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I've added an explanation of none.

I'm not particularly sure how much override explanation is needed; let me know what you think once you've read my answer.

I don't think moving the explanation from the example to up here really works; that explanation is written in the context of the example. I think both are useful.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: It is more obvious when you look at the rendered page — it is 3 differently-named anchors.

I've used non-breaking dash entity references here so that the dashed-ident dashes always appear on the same line. Otherwise, it can be really confusing for the reader. I need to use `<code></code>` elements for them to be rendered properly. If you use ticks, the literal characters get rendered out.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I've added the link. I don't think "unnamed" is very helpful — my understanding is that implicit anchor associations are named, but they are internal names that are not accessible to the developer.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I don't think this change is needed.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I've made the first suggested change and simplified the sentence (complete rewrite).

The second example in this article has multiple anchors and positioned elements, as suggested. This is supposed to be a really simple demonstration.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: OK, Improvements made, plus I've hidden the JS for brevity.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: done

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: Good spot, thanks.

If we did `--my-anchor, --another-anchor`, it would have the same effect as `all`.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: updated

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: See previous comment.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I've tried it, and TBH, it doesn't really allow you to clearly see that. It just makes the positioned elements harder to read. So I removed it again.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: Explanation added.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: Again, see previous comment.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: updated

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: updated

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I don't want to do that. I think they are useful here, explained in the context of the example. It is tricky to understand the effects these values have, so this serves as a useful recap.

files/en-us/web/css/reference/properties/position-anchor/index.md
chrisdavidmills: done

files/en-us/web/css/reference/properties/anchor-scope/index.md
estelle: They aren't "landing pages". A landing page serves as a menu, of sorts, for its subpages. #42644 removes the misnomer in the css section

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,297 @@
...
+
+The `anchor-scope` property can fix this problem by limiting the visibility, or "scope", of an `anchor-name` value to a specific subtree. The result is that each positioned element can only be anchored to an element within the same subtree of the element that has the scope set on it.
+
+- `anchor-scope: all` sets the scope so that _any_ `anchor-name` values set in the subtree can only be bound to by positioned elements in the same subtree.
+- `anchor-scope: <dashed-ident>#` sets the scope so that the specified `anchor-name` values, when set in the subtree, can only be bound to by positioned elements in the same subtree.
+- `anchor-scope: none` is the default value; it specifies that no anchor scoping is set.
+
+If you have, for example, three `anchor-name` values set inside a subtree (say, <code>&#8209;&#8209;anchor1</code>, <code>&#8209;&#8209;anchor2</code>, and <code>&#8209;&#8209;anchor3</code>), setting <code>anchor-scope: &#8209;&#8209;anchor1, &#8209;&#8209;anchor2, &#8209;&#8209;anchor3</code> on the top-level element of the subtree would be equivalent to setting `anchor-scope: all`.
```
estelle:
```suggestion
If you have, for example, three `anchor-name` values set inside a subtree (say, `‑‑anchor1`, `‑‑anchor2`, and `‑‑anchor3`), setting `anchor-scope: ‑‑anchor1, ‑‑anchor2, ‑‑anchor3` on the top-level element of the subtree would be equivalent to setting `anchor-scope: all`.
```
I used non-breaking hyphens.

files/en-us/web/css/reference/properties/anchor-scope/index.md
estelle: I was thinking we might want to globally replace all to non-breaking anchor characters,  but i don't think a breaking double dash is an issue (we don't do this anywhere else), so i recommend changing all to ndashes and getting rid of the `<code>`s.

files/en-us/web/css/reference/properties/anchor-scope/index.md
```diff
@@ -0,0 +1,297 @@
...
+- `--my-anchor`: The scope for positioning elements relative to anchor elements that are descendants of the `<section>` elements is limited to positioned elements that are themselves descendants of the `<section>` elements, only if the `--my-anchor` `anchor-name` is used to associate them. As a result, the positioned elements inside the `<section>` elements ("Positioned 1–3") are anchor-positioned as expected, and the positioned element outside the `<section>` elements ("Positioned 4") is also anchor-positioned as expected. In the former case, the positioned elements are inside the set scope, and in the latter case, the positioned element is not affected by the set scope, as it is using an anchor name outside the scope (<code>&#8209;&#8209;another-anchor</code>). The "Positioned 4" element is positioned relative to the last anchor element in the source that has the matching anchor name.
```
estelle: Add the explanation provided here to the description. Currently, the description does not expand at all on the value section content. Examples should add to what is covered in the description, illustrating what is explained there. The examples should illustrate what was already covered. The content here is good, but will be missed by many readers.

files/en-us/web/css/reference/properties/anchor-scope/index.md
estelle: `    box-shadow: 3px 3px 3px rgba(0 0 0 / .2)` does the trick - when 3 are on top of each other it's apparent.

estelle (commented):
left some comments. 

I've been thinking about the non-breaking dashes. I think we should stick with regular dashes. breaking double dashes hasn't been an issue thus far on MDN, and they are not valid. https://codepen.io/estelle/pen/WbxwRPa

files/en-us/web/css/reference/properties/anchor-scope/index.md
estelle: we can keep this here, but add the guts to the description, and make this a summary of what is happening, allowing the explanation to be in the description with a "select the \[example of the \`all\` value\]\(#anchor)" lingo with a link to this example

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: OK, fine, committing.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: OK, I've gone through and replaced them all with regular dashes and backticks. I think your reasoning about the non-breaking dashes being invalid (and therefore causing problems if people decide to copy and paste them) is the main argument that has convinced me.

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: I've added this, and I've also removed the background color from the positioned elements, so that when they are overlapping, you get a darker box shadow and overlapping text,

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: done

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: OK, I've had a go at representing the extra information from the example in a more generic fashion inside the description section. It has got quite wordy, and I'm not convinced it is that great. This stuff really is quite complicated to explain. Let me know what you think ;-)

files/en-us/web/css/reference/properties/anchor-scope/index.md
chrisdavidmills: done

PR #42238, opened 2025-12-05
TrustedTypes: Document.execCommand() with command insertHTML

hamishwillee:
[`Document.execCommand()`](http://localhost:5042/en-US/docs/Web/API/Document/execCommand#commandname) with command `insertHTML` is an injection sink.

Given that this is just a small part of what the method does, and that the method is deprecated, I have added the disclaimer to the point where `insertHTML` is documented. The disclaimer does not link to a security considerations section, but instead points to the TT API overview. I think this is the right level.

Project tracking in #41507


wbamberg (approved):
👍 thanks, looks great.

PR #42241, opened 2025-12-05
feat(jsondata): add translations for CSS formal definitions

argl:
### Description

Adds translations for the mdn-data-free version of rari. Existing translations on the formal definition values cannot be carried over, so there is a new translations file with only en-US values pre-filled and the rest auto-translated for a baseline.

### Motivation

Remove mdn-data dependency from the publishing pipeline.

### Additional details

This is compatible with the old and the new approach to formal definitions. Before merging the rari-part of this, nothing will change on the output.

See related rari PR: https://github.com/mdn/rari/pull/390


files/jsondata/L10n-CSSFormalDefinitions.json
bsmth, line 1: The spellchecker doesn't like this:

Image

We may want to temporarily ignore in `.vscode/cspell.json`:

```diff
  "ignorePaths": [
    ".vscode/cspell.json",
    ".vscode/extensions.json",
    ".markdownlint-cli2.jsonc",
+   "files/jsondata/L10n-CSSFormalDefinitions.json"
  ],
```

caugner: How about:

```diff
  "ignorePaths": [
    ".vscode/cspell.json",
    ".vscode/extensions.json",
    ".markdownlint-cli2.jsonc",
+   "files/jsondata/L10n-*.json"
  ],
```

(`ignorePaths` supports wildcards, see: https://cspell.org/docs/getting-started#2-add-words-to-the-project-dictionary)

caugner: @bsmth Can you (spell-)check again?

PR #42264, opened 2025-12-05
The note makes more sense as a warning

adriancuadrado:
Notes are easy to miss because they are meant to show information that is good to know. Warnings on the other hand are pieces of information people pay more attention to because they are more important, and this detail is very important to keep in mind.

https://developer.mozilla.org/en-US/docs/Web/API/Selection/addRange#examples

### DIFF
--- a/files/en-us/web/api/selection/addrange/index.md
+++ b/files/en-us/web/api/selection/addrange/index.md
@@ -28,7 +28,7 @@ None ({{jsxref("undefined")}}).
 
 ## Examples
 
-Only Firefox supports multiple selection ranges. Other browsers will not add new ranges to the selection if it already contains one.
+Note that only Firefox supports multiple selection ranges. In this example, other browsers will not add new ranges to the selection if it already contains one.
 
 ### HTML
 


files/en-us/web/api/selection/addrange/index.md
```diff
@@ -28,7 +28,7 @@ None ({{jsxref("undefined")}}).
 
 ## Examples
 
-> [!NOTE]
+> [!WARNING]
 > Currently only Firefox supports multiple selection ranges, other browsers will not
 > add new ranges to the selection if it already contains one.
```
wbamberg:
```suggestion
Only Firefox supports multiple selection ranges. Other browsers will not add new ranges to the selection if it already contains one.
```

wbamberg (approved):
Thank you for the PR! I don't agree though that this should be a warning - in fact I don't even think it should be a note. However I do think there should be a note in the compat data indicating this, so I filed https://github.com/mdn/browser-compat-data/issues/28742.

files/en-us/web/api/selection/addrange/index.md
```diff
@@ -28,9 +28,7 @@ None ({{jsxref("undefined")}}).
 
 ## Examples
 
-> [!NOTE]
-> Currently only Firefox supports multiple selection ranges, other browsers will not
-> add new ranges to the selection if it already contains one.
+Only Firefox supports multiple selection ranges. Other browsers will not add new ranges to the selection if it already contains one.
```
wbamberg:
```suggestion
Note that only Firefox supports multiple selection ranges. In this example, other browsers will not add new ranges to the selection if it already contains one.
```

PR #42271, opened 2025-12-06
Clarify that `x{n}?` behaves the same as `x{n}`

### LINKED ISSUE #42270: Syntactically valid, but pointless example for non-greedy modification of an exact-number quantifier
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Cheatsheet#quantifiers

Quantifiers

The regex documentation generally only lists _useful_ expressions, rather than expressions that compile, but ultimately serve no practical purpose.

On this page a list is given to illustrate non-greedy quantifiers:
```
x*?
x+?
x??
x{n}?
x{n,}?
x{n,m}?
```
I believe the `x{n}?` example may be superfluous - since I think it is always functionally equivalent to `x{n}`. I assume it may affect the order in which the regex engine tries alternatives during backtracking, but since `{n}` has only one possible match length, this makes no practical difference, or have any specific utility.

Checking the [`RepeatMatcher` algorithm](https://tc39.es/ecma262/#sec-runtime-semantics-repeatmatcher-abstract-operation) (see below) confirms it can have no effect.

```
x*?
x+?
x??
x{n,}?
x{n,m}?
```

https://tc39.es/ecma262/#sec-runtime-semantics-repeatmatcher-abstract-operation

The `greedy` flag only affects behaviour in steps 9-12 of the `RepeatMatcher`
Steps 9-12 are only executed when `min = 0`
When `min = max = n` (for `{n}`), we can only reach `min = 0` when `max = 0` as well
At that point (step 1), we return immediately: `If max = 0, return c(x)`

Hence, the `greedy is false` (step 9) can never be reached with `x{n}?` form.

_No response_

sroiis:
Fixes #42270
Fixes https://github.com/mdn/content/issues/42273

Added a note explaining that `x{n}?` behaves the same as `x{n}` because `{n}` is fixed-length and the lazy modifier has no effect.

### Description
Clarifies the behavior of `x{n}?` by adding a note explaining that it behaves the same as `x{n}`.

### Motivation
This avoids confusion by clearly stating that the `?` lazy modifier has no effect on fixed-length quantifiers like `{n}`.

### Additional details
Documentation-only change. No code or behavior is modified.

Fixes #42270

### DIFF
--- a/files/en-us/web/javascript/guide/regular_expressions/character_classes/index.md
+++ b/files/en-us/web/javascript/guide/regular_expressions/character_classes/index.md
@@ -216,7 +216,7 @@ console.log(moods.match(regexpEmoticons));
           Matches a control character using
           <a href="https://en.wikipedia.org/wiki/Caret_notation"
             >caret notation</a
-          >, where "X" is a letter from A–Z (corresponding to code points
+          >, where "X" is a letter from A–Z or a–z (corresponding to code points
           <code>U+0001</code><em>–</em><code>U+001A</code>). For example,
           <code>/\cM\cJ/</code> matches "\r\n".
         </p>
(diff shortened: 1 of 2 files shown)


TomStrepsil:
As per [this comment](https://github.com/mdn/content/issues/42270#issuecomment-3621822931), worth noting that [the main quantifiers page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Quantifier) also includes an example of this form.

PR #42282, opened 2025-12-08
Fix ABSN.start()'s 'duration' parameter description

Kaiido:
### Description

Fixes the description of the AudioBufferSourceNode `start()`'s `duration` parameter so it's not said to be similar to calling `stop(duration)`.

### Motivation

Since this specs change [1] (in 2018) the `duration` argument is specced [2] as being the duration of the buffer's content to be played, and not necessarily the one of the AudioContext's clock. So if the source node has a `playbackRate` that is not `1`, using this parameter will not be the same as using `stop(duration)`.  

### Additional details

I based the wordings on the specs text, but I don't mind another formulation altogether.

Might also be noted that there is an interop issue here, where only Firefox does follow the specs. I'll open issues to get this sorted out though.

[1]: https://github.com/WebAudio/web-audio-api/pull/1681
[2]: https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-start-when-offset-duration-duration
- https://github.com/WebAudio/web-audio-api/pull/1681
- https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-start-when-offset-duration-duration

N/A

### DIFF
--- a/files/en-us/web/api/audiobuffersourcenode/start/index.md
+++ b/files/en-us/web/api/audiobuffersourcenode/start/index.md
@@ -43,7 +43,7 @@ start(when, offset, duration)
   - : The duration of the audio data to be played, specified as seconds of total buffer content.
     If this parameter isn't specified, the sound plays until it reaches its natural conclusion or
     is stopped using the {{domxref("AudioScheduledSourceNode.stop", "stop()")}} method. The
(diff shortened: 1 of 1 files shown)


Kaiido:
Browsers issues:
 - https://issues.chromium.org/issues/467103139
 - https://bugs.webkit.org/show_bug.cgi?id=303790

PR #42295, opened 2025-12-08
Update definition of CSS pseudo-classes

estelle:
Clarified the definition of CSS pseudo-classes to include selection based on external information; original seemed to be when we were limited to links and form control states

### DIFF
--- a/files/en-us/web/css/reference/selectors/pseudo-classes/index.md
+++ b/files/en-us/web/css/reference/selectors/pseudo-classes/index.md
@@ -10,7 +10,7 @@ spec-urls:
 sidebar: cssref
 ---
 
-A [CSS](/en-US/docs/Web/CSS) **_pseudo-class_** is a keyword added to a selector that lets you select elements based on information that lies outside of the document tree, such as by a specific state of the selected element(s). For example, the pseudo-class {{CSSxRef(":hover")}} can be used to select a button when a user's pointer hovers over the button and this selected button can then be styled.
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/reference/selectors/pseudo-classes/index.md
```diff
@@ -10,7 +10,7 @@ spec-urls:
 sidebar: cssref
 ---
 
-A [CSS](/en-US/docs/Web/CSS) **_pseudo-class_** is a keyword added to a selector that lets you style a specific state of the selected element(s). For example, the pseudo-class {{CSSxRef(":hover")}} can be used to select a button when a user's pointer hovers over the button and this selected button can then be styled.
+A [CSS](/en-US/docs/Web/CSS) **_pseudo-class_** is a keyword added to a selector that lets you select elements based on information that lies outside of the document tree, such as by a specific state of the selected element(s). For example, the pseudo-class {{CSSxRef(":hover")}} can be used to select a button when a user's pointer hovers over the button and this selected button can then be styled.
```
chrisdavidmills:
```suggestion
A [CSS](/en-US/docs/Web/CSS) **_pseudo-class_** is a keyword added to a selector that lets you select elements based on information that lies outside of the document tree, such as a specific state of the selected element(s). For example, the pseudo-class {{CSSxRef(":hover")}} can be used to style a button when a user's pointer hovers over it.
```

That last bit I've suggested removing was not your fault, but it was wordy and awkward.

PR #42306, opened 2025-12-09
FF147 Relnote: CSS Module scripts

hamishwillee:
FF147 adds support for module scripts in https://bugzilla.mozilla.org/show_bug.cgi?id=1986681. This allows you to do stuff like below to import a stylesheet and adopt it into your document.

```js
import sheet from './styles.css' with { type: 'css' };
document.adoptedStyleSheets = [sheet];
```

This adds a release note.

Related docs work can be tracked in #42255

### DIFF
--- a/files/en-us/mozilla/firefox/experimental_features/index.md
+++ b/files/en-us/mozilla/firefox/experimental_features/index.md
@@ -384,21 +384,7 @@ The [`@custom-media`](/en-US/docs/Web/CSS/Reference/At-rules/@custom-media) CSS 
 
 ## JavaScript
 
-### CSS module scripts
-
-CSS module scripts are now supported, allowing a stylesheet to be loaded into a script as a {{domxref("CSSStyleSheet")}} instance using the [`import`](/en-US/docs/Web/JavaScript/Reference/Statements/import) statement.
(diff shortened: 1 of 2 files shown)


hamishwillee:
@Josh-Cena [import attributes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with) covers CSS module scripts, but doesn't really have a great place to link to - i.e. you have to read the whole document really to understand them.

Would you be amenable to updating the description in [Import attributes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with#description)  to add a section "Module `type` attribute. Otherwise much the same information?

Josh-Cena:
Yes, I think that's reasonable. Currently this information is entangled within the main description and I'm not sure how to pull that out into a separate subsection, but happy with the direction.

hamishwillee:
Thanks very much. I've accepted the suggstions. I'm leaving this open as I will change the link if #42309 is accepted.

pepelsbey (approved):
Looks good! Thank you :)

Just a nit:

```
No experimental features in this release cycle.
```

Some of our lower level headings are bold and text sized. I would keep it like a normal text. We also use non-bold text for "no notable changes" in the release notes.

hamishwillee:
> Looks good! Thank you :)
> 
> Just a nit:
> 
> ```
> No experimental features in this release cycle.
> ```
> 
> Some of our lower level headings are bold and text sized. I would keep it like a normal text. We also use non-bold text for "no notable changes" in the release notes.

Thanks @pepelsbey 

This text was copied from another section (it appears in 2 other places). No issue with it being changed to anything you like, but I am being consistent. Either way, should be done outside of this PR.

PR #42307, opened 2025-12-09
Clarify JavaScript MIME type parameter handling in script type

lionel-rowe:
### Description

Updated the explanation regarding the use of parameters in the JavaScript MIME type for `<script type=...>`

### Motivation

Current wording is outdated, because `module` and `importmap` now exist.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type#value

### DIFF
--- a/files/en-us/web/http/guides/mime_types/index.md
+++ b/files/en-us/web/http/guides/mime_types/index.md
@@ -149,9 +149,9 @@ All HTML content should be served with this type. Alternative MIME types for XHT
 JavaScript content should always be served with the MIME type `text/javascript`.
 For historic reasons, browsers may support some [legacy JavaScript types](#legacy_javascript_mime_types) listed below, but you should not assume scripts served with any MIME type other than `text/javascript` will always load or run.
 
-Note that in HTML the [`type`](/en-US/docs/Web/HTML/Reference/Elements/script/type) attribute for {{htmlelement("script")}} elements may only contain the **JavaScript MIME type essence**: `text/javascript`.
-Including any parameter, such as `charset=utf-8`, is the same as setting the `type` to [an unrecognized value](/en-US/docs/Web/HTML/Reference/Elements/script/type#any_other_value): the script content is treated as a data block and is not executed as JavaScript.
-(Note that setting the `type` to a JavaScript MIME type is a deprecated feature itself: you should omit the `type` in this case.)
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/guides/mime_types/index.md
```diff
@@ -150,7 +150,7 @@ JavaScript content should always be served with the MIME type `text/javascript`.
 For historic reasons, browsers may support some [legacy JavaScript types](#legacy_javascript_mime_types) listed below, but you should not assume scripts served with any MIME type other than `text/javascript` will always load or run.
 
 Note that in HTML the [`type`](/en-US/docs/Web/HTML/Reference/Elements/script/type) attribute for {{htmlelement("script")}} elements may only contain the **JavaScript MIME type essence**: `text/javascript`.
```
bsmth:
```suggestion
Note that in HTML the [`type`](/en-US/docs/Web/HTML/Reference/Elements/script/type) attribute for {{htmlelement("script")}} elements may only contain the **JavaScript MIME type essence**: `text/javascript` or one of the keywords `module` (for ES modules) or `importmap`.
```

files/en-us/web/http/guides/mime_types/index.md
```diff
@@ -150,7 +150,7 @@ JavaScript content should always be served with the MIME type `text/javascript`.
...
-Including any parameter, such as `charset=utf-8`, is the same as setting the `type` to any other MIME type: the script content is treated as a data block and is not executed as JavaScript.
+Including any parameter, such as `charset=utf-8`, is the same as setting the `type` to [an unrecognized value](/en-US/docs/Web/HTML/Reference/Elements/script/type#any_other_value): the script content is treated as a data block and is not executed as JavaScript.
```
bsmth:
```suggestion
Including any parameter in the `type` attribute, such as `charset=utf-8`, is the same as setting the `type` to [an unrecognized value](/en-US/docs/Web/HTML/Reference/Elements/script/type#any_other_value): the script content is treated as a data block and is not executed as JavaScript.
```

files/en-us/web/http/guides/mime_types/index.md
```diff
@@ -150,7 +150,7 @@ JavaScript content should always be served with the MIME type `text/javascript`.
...
 (Note that setting the `type` to a JavaScript MIME type is a deprecated feature itself: you should omit the `type` in this case.)
```
bsmth: Another one to clean up, I think:

```suggestion
Note that setting `type="text/javascript"` is no longer necessary; this is the default for `<script>` elements, so you may omit the `type` attribute entirely in this case.
```

Issue #42283, opened 2025-12-08
atan2() calculation always returns an <angle> in degrees, not radians


Jollus174:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/atan2

### What specific section or headline is this issue about?

The introduction (and possibly other sections of page).

### What information was incorrect, unhelpful, or incomplete?

I might be misunderstanding here, but it mentions.

> The function accepts two arguments and returns the number of radians representing an [<angle>](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/angle) between -180deg and 180deg.

I've tested this, and wouldn't this actually be the number of **degrees** between -180deg and 180deg?
For example, applying a radians-to-degress conversion via:
```
  --angle-rad: atan2(3, 2);
  --angle-deg: calc(var(--angle-rad) * 180 / pi);
```
would end up with a wildly incorrect result for `--angle-deg`.
In this example, I'd used CSS trigonmetry calculations to offset a gradient based off the position of the mouse relative to the centre of the screen: https://codepen.io/Jollus/pen/WbweZoX
We can see in `JS angle rad` in the top-left what the radians is expected to be, calculated with JavaScript. However, this is not what ends up being applied to the `--angle` (the CSS calculation).

<img width="1861" height="892" alt="Image" src="https://github.com/user-attachments/assets/50dbef9e-1a20-47b0-9922-1e330a118791" />

To me, it seems clear that atan2() is returning the result in degrees, and not radians. This may be so for other trigonometry calculations as well, although I've not tested those.

### What did you expect to see?

I would think that any reference to atan2() returning **radians** would need to be updated to show that it returns **degrees** instead, unless there's something I'm missing here.

### Do you have any supporting links, references, or citations?

https://codepen.io/Jollus/pen/WbweZoX

<img width="1861" height="892" alt="Image" src="https://github.com/user-attachments/assets/64065463-823d-43b3-b3d6-39f3c4c9a3d5" />

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/css/reference/values/atan2`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/atan2
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/reference/values/atan2/index.md
* Last commit: https://github.com/mdn/content/commit/85fccefc8066bd49af4ddafc12c77f35265c7e2d
[...]

PR #42308, opened 2025-12-09
fix(css): clarify atan2() return value documentation

abhinav12222363:
This PR fixes incorrect wording in the CSS atan2() documentation.

The original text incorrectly stated that the function returns “the number of radians representing an angle between -180deg and 180deg”, which mixes units and caused confusion.

Updated the sentence to:

- Remove the incorrect reference to radians  
- Clarify that the function returns an angle between -180deg and 180deg, as suggested by the MDN maintainer in issue #42283.


estelle (approved):
Thank you! 🎉

PR #42309, opened 2025-12-09
Restructure import assertions description to ease maintainability

hamishwillee:
As discussed in https://github.com/mdn/content/pull/42306#issuecomment-3629944371 the description of import assertions does not provide an easy way to link to particular parts of the concepts, such as a definition of CSS Modules imported using `with { type: "css"}`.

This restructures the existing material to provide some headings. Should allow clean linking to the doc, easier parsing for readers, and maintenance if we later add other module types.

@Josh-Cena Obviously "for discussion and iteration". It is almost all a restructure, with only a tiny amount of new content, which I will highlight.

Related docs work being done in #42255

### DIFF
--- a/files/en-us/web/javascript/reference/statements/import/with/index.md
+++ b/files/en-us/web/javascript/reference/statements/import/with/index.md
@@ -41,7 +41,8 @@ export { names } from "module-name" with { key: "data", key2: "data2", /* …, *
 
 - {{jsxref("TypeError")}}
   - : An unsupported `key` was specified in a _dynamic import_.
-Note that the behaviour for unsupported keys is undefined, but might result in an exception on some browsers.
+
+Note that specifying an unsupported value for a supported key may also result in an exception in some cases, depending on the key.
 
 ## Description
 


files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -34,6 +34,14 @@ export { names } from "module-name" with { key: "data", key2: "data2", /* …, *
 - `"dataN"`
   - : An attribute value. Must be a string literal.
 
+### Exceptions
+
+- {{jsxref("SyntaxError")}}
+  - : An unsupported `key` was specified.
+
+- {{jsxref("TypeError")}}
+  - : An unsupported value was used with a supported `key`.
```
hamishwillee: This is new. Tested on firefox and chrome using types 'css' and type "unlikely" and also a key of 'unlikelykey'.

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
-> The actual `type` attribute value does not correspond directly to the MIME type. It's separately specified by the [HTML specification](https://html.spec.whatwg.org/multipage/webappapis.html#module-type-allowed).
+### Standard attributes
+
+The available attributes depend on the language and runtime environment (the ECMAScript standard does not define any attributes).
 
-Therefore, the code above should be re-written as:
+The HTML specification [defines the `type` attribute](https://html.spec.whatwg.org/multipage/webappapis.html#module-type-allowed), which is supported in browser environments.
+This has allowed values `"json"` and `"css"`.
+
+##### JSON Modules (`{ type: "json" }`)
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
#### JSON Modules (`{ type: "json" }`)
```

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
+Browsers, on the other hand, implicitly assume that the module is JavaScript if the `type` is not specified, and fail if the module is not JavaScript (for example, JSON). This ensures that module types are always strictly validated and prevents any security risks.
+Non-browser runtimes such as Node and Deno align with browser semantics and enforce `type` for JSON modules.
+
+In other words, if you omit the `type` and attempt to import a file as `"application/json"` you will usually get an error like the following:
+
+```plain
+Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "application/json". Strict MIME type checking is enforced for module scripts per HTML spec.
+```
+
+##### CSS Modules (`{ type: "css" }`)
```
hamishwillee:
```suggestion
#### CSS Modules (`{ type: "css" }`)
```

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
 
-Instead, you must provide an attribute to tell the runtime that this file must contain JSON. To validate the module's type (via MIME type), you use the attribute key called `type`. To validate that the module is a JSON module, the value is `"json"`.
+A runtime can read and process any attributes that it understands.
+However, you cannot use unknown attributes — a runtime will throw a syntax error if it encounters an unknown attribute.
 
-> [!NOTE]
-> The actual `type` attribute value does not correspond directly to the MIME type. It's separately specified by the [HTML specification](https://html.spec.whatwg.org/multipage/webappapis.html#module-type-allowed).
+### Standard attributes
+
+The available attributes depend on the language and runtime environment (the ECMAScript standard does not define any attributes).
```
Josh-Cena: This isn't actually true. ECMAScript defines `type: "json"`; see https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#sec-HostLoadImportedModule

hamishwillee: Thanks! Fixed in https://github.com/mdn/content/pull/42309/commits/600f8b6fd13fe2ddfafa9523fde717acbe385f2b

files/en-us/web/javascript/reference/statements/import/with/index.md
Josh-Cena: Sooo, this is complicated. `import foo from "foo" with { unlikely: "bah" }` is a SyntaxError, but `import("foo", { with: { unlikely: "bah" } })` is a TypeError. As for unsupported values, the host is not _required_ to throw an error, or rather it can choose to do anything _including_ throwing an error, and the behavior is wholly dependent on the `key` itself. This was chiefly why I left it off because basically nothing is certain, but okay if you want to add it; just need to point out these two things.

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
+Modules are identified and parsed only according to their served [media type (MIME type)](/en-US/docs/Web/HTTP/Guides/MIME_types) — the file extension in the URL cannot be used to identify a file's type. In this case, the MIME type is `application/json`, which tells the browser that the file is JSON and must be parsed as JSON. If, for some reason (e.g., the server is hijacked or bogus), the MIME type in the server response is set to `text/javascript` (for JavaScript source), then the file would be parsed and executed as code. If the "JSON" file actually contains malicious code, the `import` declaration would unintentionally execute external code, posing a serious security threat.
```
Josh-Cena: For my education, because Brian has made the same comment: why is "MIME type" not in fashion anymore?

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
+When using import attributes the code above would be written as shown below, and would fail if the file was served with any media type other than `application/json`:
 
-```plain
-Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "application/json". Strict MIME type checking is enforced for module scripts per HTML spec.
+```js
+import data from "https://example.com/data.json" with { type: "json" };
 ```
 
-Instead, you must provide an attribute to tell the runtime that this file must contain JSON. To validate the module's type (via MIME type), you use the attribute key called `type`. To validate that the module is a JSON module, the value is `"json"`.
+A runtime can read and process any attributes that it understands.
```
Josh-Cena: This sounds tautological. Perhaps just "all attributes you specify must be understood by the runtime, otherwise a syntax error is thrown"?

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
+Modules are identified and parsed only according to their served [media type (MIME type)](/en-US/docs/Web/HTTP/Guides/MIME_types) — the file extension in the URL cannot be used to identify a file's type. In this case, the MIME type is `application/json`, which tells the browser that the file is JSON and must be parsed as JSON. If, for some reason (e.g., the server is hijacked or bogus), the MIME type in the server response is set to `text/javascript` (for JavaScript source), then the file would be parsed and executed as code. If the "JSON" file actually contains malicious code, the `import` declaration would unintentionally execute external code, posing a serious security threat.
 
-Import attributes fix this problem by allowing the author to explicitly specify how a module should be validated. For example, the import statement above, which lacks an attribute, would actually fail:
+Import attributes fix this problem by allowing the author to explicitly specify how a module should be validated.
+When using import attributes the code above would be written as shown below, and would fail if the file was served with any media type other than `application/json`:
```
Josh-Cena: This flows a bit weirder than before IMO. The intended logic was: browsers always have strict MIME validation, and the import above failed validation > MIME validation is done via the `type` key (implying that other keys can exist for other purposes) > JSON modules correspond to `type: "json"` > then the example.

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
+
+The HTML spec defines the `css` type, which imports a stylesheet into a script as an {{domxref("CSSStyleSheet")}} object.
+
+The code below shows how you might import a style and add it to your document.
+The import will throw an exception if `example_styles.css` is served with any media type other than `"text/css"`.
 
 ```js
-import styles from "https://example.com/styles.css" with { type: "css" };
+import exampleStyles from "https://example.com/example_styles.css" with { type: "css" };
+document.adoptedStyleSheets = [exampleStyles];
```
Josh-Cena:
```suggestion

document.adoptedStyleSheets.push(exampleStyles);
```

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
 ```
 
-The attributes syntax is designed to be extensible — although only `type` is specified by the language, the runtime can read and process other attributes. An attribute can change the runtime's behavior at every stage of the module loading process:
+Note that importing CSS modules into workers is not usually supported.
+It is allowed by the HTML specification if `CSSStyleSheet` is exposed to workers, but the CSSOM specification only exposes `CSSStyleSheet` in the main window thread.
```
Josh-Cena: This reads slightly awkward. Not sure how HTML spec can "allow" but not "specify" something; is it a blanket term somewhere that CSSOM decides to override?

```suggestion
The HTML specification allows `CSSStyleSheet` to be exposed to workers, but the CSSOM specification only exposes `CSSStyleSheet` in the window context.
```

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
+
+### Attributes modify runtime behavior
```
Josh-Cena:
```suggestion
### Intended semantics for import attributes
```

files/en-us/web/javascript/reference/statements/import/with/index.md
hamishwillee: Thanks - that is messy. Isn't the reason for the different cases because import foo is a static compile/parsing check while import() is a runtime value check? 
If so, I guess the problem is really that we're using this topic for both static and dynamic import.

How about this - stating what we know to be true.

```suggestion
### Exceptions

- {{jsxref("SyntaxError")}}
  - : An unsupported `key` was specified in a _static import_.

- {{jsxref("TypeError")}}
  - : An unsupported `key` was specified in a _dynamic import_.
  
Note that the behaviour for unsupported keys is undefined, but might result in an exception on some browsers.
```

files/en-us/web/javascript/reference/statements/import/with/index.md
hamishwillee: The IANA decided they didn't like MIME, AIUI because the "E" is for email - and media types are used much more widely.  They own the list of media types, and decided that henceforth we'll use this term.

FWIW I don't like it because media type is a much more general term that could mean anything, not just the registered types formerly known as MIME.  Hence the clarification on first use.

files/en-us/web/javascript/reference/statements/import/with/index.md
hamishwillee: Yes, that's better

```suggestion
All attributes you specify must be understood by the runtime, otherwise a syntax error is thrown.
```

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -53,36 +61,69 @@ Content-Type: application/json; charset=utf-8
...
 
 ```js
-import styles from "https://example.com/styles.css" with { type: "css" };
+import exampleStyles from "https://example.com/example_styles.css" with { type: "css" };
+document.adoptedStyleSheets = [exampleStyles];
 ```
 
-The attributes syntax is designed to be extensible — although only `type` is specified by the language, the runtime can read and process other attributes. An attribute can change the runtime's behavior at every stage of the module loading process:
+Note that importing CSS modules into workers is not usually supported.
+It is allowed by the HTML specification if `CSSStyleSheet` is exposed to workers, but the CSSOM specification only exposes `CSSStyleSheet` in the main window thread.
```
hamishwillee:
```suggestion
Note that importing CSS modules into workers is usually not supported, because  the CSSOM specification only exposes `CSSStyleSheet` in the window context.
```

files/en-us/web/javascript/reference/statements/import/with/index.md
hamishwillee: The relevant test is https://html.spec.whatwg.org/multipage/webappapis.html#module-type-allowed step 2:

> If moduleType is "css" and the [CSSStyleSheet](https://drafts.csswg.org/cssom/#the-cssstylesheet-interface) interface is not [exposed](https://webidl.spec.whatwg.org/#dfn-exposed) in settings's [realm](https://html.spec.whatwg.org/multipage/webappapis.html#environment-settings-object's-realm), then return false.

So the HTML spec is saying you can't fetch a CSS module at all if the CSSStyleSheet interface is not exposed in the environment, such as a worker. The CSSDP< spec says  CSSStyleSheet spec IDL says that it is only exposed in window.

I'm overcomplicating this though. Fix in https://github.com/mdn/content/pull/42309/files#r2612966421

files/en-us/web/javascript/reference/statements/import/with/index.md
hamishwillee: Thanks. I've attempted to more closely match your intent in https://github.com/mdn/content/pull/42309/changes/3300ed07ea22a33ab969e3da1794f066a3204cd7

hamishwillee:
@Josh-Cena Did the updates address your concerns?

Josh-Cena:
Sorry, was extremely busy. I'll look later this week.

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -34,6 +34,15 @@ export { names } from "module-name" with { key: "data", key2: "data2", /* …, *
...
   - : An attribute value. Must be a string literal.
 
+### Exceptions
+
+- {{jsxref("SyntaxError")}}
+  - : An unsupported `key` was specified in a _static import_.
+
+- {{jsxref("TypeError")}}
+  - : An unsupported `key` was specified in a _dynamic import_.
+Note that the behaviour for unsupported keys is undefined, but might result in an exception on some browsers.
```
hamishwillee:
```suggestion

Note that specifying an unsupported value for a supported key may also result in an exception in some cases, depending on the key.
```

PR #42310, opened 2025-12-09
fix a custom property name

mfuji09:
The property name is `--boxSize` in the description, but `--box-size` is used in the code.

### Description

Fix a custom property name `--boxSize` to `--box-size`.

### Motivation

The actual CSS and JavaScript codes use `--box-size`.


chrisdavidmills (approved):
Makes sense to me. Thanks for the fix, @mfuji09!

PR #42314, opened 2025-12-09
fix: bugzilla ID mismatch

bsmth:
### Description

Link text doesn't match the href


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #42316, opened 2025-12-09
Fx147 devtools edit pseudo element selector

chrisdavidmills:
### Description

In Firefox 147, when pseudo-element nodes are selected in the HTML pane of the Inspector — for example, those representing generated content created by `::before` and `::after` rules — you can now edit the selectors of the corresponding rules in the Rule view. See https://bugzilla.mozilla.org/show_bug.cgi?id=1998704.

### DIFF
--- a/files/en-us/mozilla/firefox/releases/147/index.md
+++ b/files/en-us/mozilla/firefox/releases/147/index.md
@@ -18,7 +18,7 @@ Firefox 147 is the current [Nightly version of Firefox](https://www.firefox.com/
 
 ### Developer Tools
 
-- When [pseudo-element](/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-elements) nodes are selected in the HTML pane of the Inspector — for example, those representing generated content created by {{cssxref("::before")}} and {{cssxref("::after")}} rules — you can now edit the selectors of the corresponding rules in the Rule view.
+- When you select a [pseudo-element](/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-elements) (such as {{cssxref("::before")}} or {{cssxref("::after")}}) in the HTML pane of the Inspector, you can now edit the selector of the corresponding rule in the CSS pane.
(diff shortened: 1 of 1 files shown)


files/en-us/mozilla/firefox/releases/147/index.md
```diff
@@ -16,7 +16,10 @@ Firefox 147 is the current [Nightly version of Firefox](https://www.firefox.com/
 
 ## Changes for web developers
 
-<!-- ### Developer Tools -->
+### Developer Tools
+
+- When [pseudo-element](/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-elements) nodes are selected in the HTML pane of the Inspector — for example, those representing generated content created by {{cssxref("::before")}} and {{cssxref("::after")}} rules — you can now edit the selectors of the corresponding rules in the Rule view.
```
dipikabh: - I referred to https://firefox-source-docs.mozilla.org/devtools-user/page_inspector/ui_tour/index.html for the UI terminology and suggesting CSS pane instead of Rule view.
- I guess the pseudo-element itself can be selected in the HTML pane but if using the Inspector picker on the web page, then in that case you'd be selecting the content generated by the pseudo element.
```suggestion
- When you select a [pseudo-element](/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-elements) (such as {{cssxref("::before")}} or {{cssxref("::after")}}) in the HTML pane of the Inspector, you can now edit the selector of the corresponding rule in the CSS pane.
```
- Alternatively, we could skip "- or select its generated content on the web page -"

dipikabh: From https://bugzilla.mozilla.org/show_bug.cgi?id=1998704#c8 and the codepen, it seems we can select the pseudo-element only in the HTML pane. And I also tested it again. So we should remove the generated content part.

chrisdavidmills: Yeah, that makes sense to me. I've edited your suggestion to remove that part, and will now commit it.

PR #41897, opened 2025-11-11
Add Firefox release automation script

pepelsbey:
### Description

It’s an automation script that helps to create Firefox release notes.

I created the [Firefox 145 release notes PR](https://github.com/mdn/content/pull/41896) using this script.

### Motivation

To simplify release notes workflow.

### Additional details

How it works, for example, with the newly released Firefox 145:

```sh
node scripts/content/release-firefox.js 145
```

What it does:

Takes the newly released version as a parameter (N), and:

- Removes active from N-1
- Updates N to Stable
  - Uncomments HTML, CSS, JavaScript “no notable changes” sections
  - Removes all other unused commented sections
- Updates N+1 to Beta
- Adds Nightly N+2 page

For release dates, it uses official JSON API — https://whattrainisitnow.com/api/firefox/calendar/future/

### Related issues and pull requests

- https://github.com/mdn/content/pull/41896

PR #42317, opened 2025-12-09
Use both current and future APIs for release notes script

pepelsbey:
### Description

Uses both current and future release note APIs so that the script can be used at any moment.

### Motivation

Otherwise, script doesn’t work after the release date.

### Additional details

Follow up for https://github.com/mdn/content/pull/41897


dipikabh (approved):
Thank you! 🙌

PR #42323, opened 2025-12-09
Add shadow root reference target information

meyerweb:
### Description

Adds brief summaries of `referenceTarget` on `attachShadow()` and `ShadowRoot`, and `shadowrootreferencetarget` on `<template>`.

### Motivation

Implemented experimentally in Chrome and positive signals from Mozilla and funding to implement in WebKit as well as progress in the HTML and WHATNOT arenas.

### Additional details

* Explainer: https://github.com/WICG/webcomponents/blob/gh-pages/proposals/reference-target-explainer.md
* HTML progress: https://github.com/whatwg/html/issues/10707
* WHATNOT progress: https://github.com/whatwg/html/issues/11648#issuecomment-3293932780
* Chrome implementation status: https://chromestatus.com/feature/5188237101891584
* Positive signal from Mozilla: https://mozilla.github.io/standards-positions/#1035
* Plans for WebKit implementation: https://blogs.igalia.com/mrego/solving-cross-root-aria-issues-in-shadow-dom/


files/en-us/web/html/reference/elements/template/index.md
```diff
@@ -38,6 +38,9 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
     If this is set and a non-focusable element in the shadow tree is selected, then focus is delegated to the first focusable element in the tree.
     The value defaults to `false`.
 
+- `shadowrootreferencetarget` {{Experimental_Inline}}
+  - : Sets the value of the `referenceTarget` property of a [`ShadowRoot`](/en-US/docs/Web/API/ShadowRoot) created using this element. The value should be the ID of an element inside the shadow DOM. If set, target references to the host element from outside the shadow DOM will cause the referenced target element to become the effective target of the reference to the host element.
```
alice, line 42: > The value should be the ID of an element inside the shadow DOM. If set, target references to the host element from outside the shadow DOM will cause the referenced target element to become the effective target of the reference to the host element.

I think this wording is great, how would you feel about including it for each of these brief descriptions?

files/en-us/web/api/shadowroot/index.md
```diff
@@ -38,6 +38,8 @@ You can retrieve a reference to an element's shadow root using its {{domxref("El
 - {{DOMxRef("ShadowRoot.pointerLockElement")}} {{ReadOnlyInline}}
   - : Returns the {{DOMxRef('Element')}} set as the target for mouse events while the pointer is locked.
     `null` if lock is pending, pointer is unlocked, or if the target is in another tree.
+- `ShadowRoot.referenceTarget` {{ReadOnlyInline}} {{Experimental_Inline}}
+  - : A string value that indicates the effective target of any element reference made against the shadow host from outside the host element.
```
alice:
```suggestion
  - : A string value that indicates the effective target of any element reference made against the shadow host from outside the host element. The value should be the ID of an element inside the shadow DOM. If set, target references to the host element from outside the shadow DOM will cause the referenced target element to become the effective target of the reference to the host element.
```

files/en-us/web/api/element/attachshadow/index.md
```diff
@@ -83,6 +83,9 @@ attachShadow(options)
       - : A boolean that, when set to `true`, specifies behavior that mitigates custom element issues around focusability.
         When a non-focusable part of the shadow DOM is clicked, the first focusable part is given focus, and the shadow host is given any available `:focus` styling. Its default value is `false`.
 
+    - `referenceTarget` {{Optional_Inline}} {{Experimental_Inline}}
+      - : A string specifying the _reference target_ of any element references made against the host element of the shadow DOM tree. The reference target must be an element held within the shadow DOM.
```
alice:
```suggestion
      - : A string value that indicates the effective target of any element reference made against the shadow host from outside the host element. The value should be the ID of an element inside the shadow DOM. If set, target references to the host element from outside the shadow DOM will cause the referenced target element to become the effective target of the reference to the host element.
```

files/en-us/web/html/reference/elements/template/index.md
estelle, line 42: what does "effective target" mean?

meyerweb: It means the thing that becomes the target of a reference, when the thing that passed the reference is targeting the host instead of the thing that ends up being the actual reference.  Would “actual target” be better?

meyerweb: https://meyerweb.com/eric/thoughts/2025/12/19/targeting-by-reference-in-the-shadow-dom/ makes it a little more clear, I hope?

estelle: This is my interpretation (which may be more grokable or may be wrong)
```suggestion
  - : Sets the value of the `referenceTarget` property of a [`ShadowRoot`](/en-US/docs/Web/API/ShadowRoot) created using this element. The value should be the ID of an element inside the shadow DOM. If set, references to that id from from outside the shadow DOM will cause the referenced target element within the host element to become the effective target of the reference, rather than the host element itself.
```

alice: Yeah, this is wrong, unfortunately.

I think this is one of those things that makes a lot of sense once it clicks, but getting it to click with a succinct explanation like this is a bit of a struggle.

There's also the problem (which was also a problem for naming the API as well as for writing the spec/implementations) that a lot of relevant terms refer to multiple related things. The way we're using the term "reference" here is intended to capture both ID references:

```html
<!-- refer to section1-heading as the section's (only) aria-labelledby element -->
<section aria-labelledby="section1-heading">
  <custom-heading id="section1-heading">Section 1</custom-heading>
</section>
```

and element references:

```html
<button id="btn">Open popover</btn>
<custom-popover>Pop</custom-popover>
<script>
let btn = document.getElementById("btn");
let popover = document.querySelector("custom-popover");

// refer to popover as btn's popover target element
btn.popoverTargetElement = popover;
</script>
```

So in each of those cases, if those `custom-` elements had a shadow root with a reference target set, it might look something like:

```html
<!-- refer to section1-heading as the section's (only) aria-labelledby element -->
<section aria-labelledby="section1-heading">
  <custom-heading id="section1-heading">
    <template shadowRootMode="open" shadowRootReferenceTarget="inner-heading">
      <div role="heading" id="inner-heading"><slot>Section 1</slot></div>
      <div>
        Extra text which isn't part of the label
      </div>
    </template>
  </custom-heading>
</section>
```

```html
<button id="btn">Open popover</btn>
<custom-popover>
  <template shadowRootMode="open" shadowRootReferenceTarget="inner-popover">
    <div id="inner-popover" popover><slot>Pop</slot></div>
    <div>Some content which isn't part of the popover</div>
  </template>
</custom-popover>
<script>
let btn = document.getElementById("btn");
let popover = document.querySelector("custom-popover");

// refer to popover as btn's popover target element
btn.popoverTargetElement = popover;
</script>
```

Hopefully that helps explain the concepts?

So your wording would suggest ID references to `"inner-heading"` and `"inner-popover"` from outside the shadow root would be forwarded, which they wouldn't. Rather, it would be ID references to `"section1-heading"` which would be forwarded to the `"inner-heading"` element inside the shadow root, and element references to the `<custom-popover>` element which would be forwarded to the `"inner-popover"` element.

PR #42326, opened 2025-12-09
Marker pseudo-element counter and quote property support

chrisdavidmills:
### Description

Firefox 147 adds support for the [`counter-*`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/counter-increment) and [`quotes`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/quotes) properties on the [`::marker`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::marker) pseudo-element. See https://bugzilla.mozilla.org/show_bug.cgi?id=2000404.

This PR adds this information to the `::marker` page, and also adds a Fx147 rel note to cover it.

#### Test results and supporting details

I have created a test case at https://codepen.io/Chris-Mills/pen/yyOGdGJ?editors=1100. In testing, I found that the features:

- Work in Firefox 147, but not 146
- Don't work in Chrome Canary or Safari technology preview.


bsmth (approved):
LGTM, thanks 👍🏻

Issue #35484, opened 2024-08-16
Django tutorial - needs a signup page


hamishwillee:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Authentication

### What specific section or headline is this issue about?

Adding users

### What information was incorrect, unhelpful, or incomplete?

The tutorial covers the standard views for login and user management, but uses the admin interface for signing users up in the first place.

### What did you expect to see?

A section explaining that user signup has no default views and setup, so you have to do it yourself ... and how.

### Do you have any supporting links, references, or citations?

Hints on how to do this: https://www.pythontutorial.net/django-tutorial/django-registration/

PR #42327, opened 2025-12-09
Django tutorial - needs a signup page #35484

Abhi-Sharma:
This pull request adds missing documentation for user registration (signup) in the Django authentication tutorial. The current MDN page explains login, logout, and permissions, but it does not mention that Django does not include a built-in signup view, which leads to confusion for learners.
This PR introduces a clear, modern example of how to implement a registration route, form, view, and template using Django’s standard tools.

### DIFF
--- a/files/en-us/learn_web_development/extensions/server-side/django/authentication/index.md
+++ b/files/en-us/learn_web_development/extensions/server-side/django/authentication/index.md
@@ -64,19 +64,6 @@ MIDDLEWARE = [
     # …
 ```
 
-## For Django Registration
-
-> [!NOTE]
-> Django does not include a built-in signup (user registration) view.
-> Some web frameworks provide a "create account" page automatically, but Django leaves
-> this for developers to implement if needed.
->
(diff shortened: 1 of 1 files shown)


hamishwillee (changes requested):
Thank you. It is worth saying that Django does not support a signup page out of the box - since that is something that users might expect to be able to do.

However the lack of a registration page in this app is deliberate. Usually when you join a library you don't self-register. Instead you are registered by a librarian who performs various checks that you are who you say you are. In the design of this library that is done through the administration interface, which is available to librarians.

So what is required here is a note explaining those two things. Would you mind ammeding appropriately - remove the code changes and add a note?

---

Note that for code changes to the tutorial we require a parallel PR to https://github.com/mdn/django-locallibrary-tutorial - as that makes it easy to verify and test changes.

Abhi-Sharma:
Hi, I have removed the signup code and added the note explaining the registration design choice.
Please review the updated changes. Thank you!

Issue #38324, opened 2023-09-13
Article element in a wrong implementation


Amer-Adel:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article

### What specific section or headline is this issue about?

Try it

### What information was incorrect, unhelpful, or incomplete?

each day’s forecast (`<article class="day-forecast">`) doesn’t make sense on its own and is too dependent on the main forecast (`<article class="forecast">`), then it might not be appropriate to use the `<article>` tag for the individual day forecasts.

Instead, you can put the name of the city in each sub article but this lead to lots of repetition, so you could consider using the `<section>` tag for each day’s forecast, as it is used to group related content. If the individual forecasts are not intended to stand alone and are part of a larger whole (the main forecast), then the `<section>` tag would be more appropriate.

### What did you expect to see?

```html
<article class="forecast">
  <h1>Weather forecast for Seattle</h1>
  <section class="day-forecast">
    <h2>03 March 2018</h2>
    <p>Rain.</p>
  </section>
  <section class="day-forecast">
    <h2>04 March 2018</h2>
    <p>Periods of rain.</p>
  </section>
  <section class="day-forecast">
    <h2>05 March 2018</h2>
    <p>Heavy rain.</p>
  </section>
</article>
```

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/html/element/article`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/element/article/index.md
* Last commit: https://github.com/mdn/content/commit/942a529383ee7ee3996fb234187641c08935f3ff
* Document last modified: 2023-04-13T06:13:15.000Z

</details>

PR #42328, opened 2025-12-09
Article element in a wrong implementation #38324

Abhi-Sharma:
This PR updates the interactive example for the <article> element.
The original code used nested <article> elements for each daily weather forecast, which is semantically incorrect because these items are not independently meaningful or distributable.

Each day’s forecast depends on the main forecast, so they should be grouped using <section> elements instead.

image


Abhi-Sharma:
Hi,
I’ve removed the unintended first commit so that this PR only includes the article content changes.
Please review the updated version. Thank you!

hamishwillee:
Thanks @Abhi-Sharma

I disagree with this change. You might reasonably take just one weather report article out of the set and display it on another page  that is nothing to do with the main weather report. For example, a  calendar app might pull the expected weather for each date. So the assertion that this is not independent is incorrect.

Closing. We can reopen if I am missing something.

PR #42338, opened 2025-12-10
Add passkeys guide

wbamberg:
Here's a page on passkeys.

The questions I had in https://github.com/mdn/content/pull/42338#issuecomment-3651865285 are still kind of valid, but I'm reasonably happy with what I have here.

**Note that the diagrams are broken in the preview page because of https://github.com/mdn/fred/issues/904.**

### DIFF
--- a/files/en-us/web/security/authentication/passkeys/index.md
+++ b/files/en-us/web/security/authentication/passkeys/index.md
@@ -5,7 +5,9 @@ page-type: guide
 sidebar: security
 ---
 
-Passkeys enable websites to authenticate users without the user having to enter any passwords or other secret codes on the site itself. They're considered [the most secure authentication method available to websites](#security_properties_of_passkeys), and we recommend that sites should adopt passkeys as their preferred authentication method, and [phase out the use of passwords](#migrating_from_passwords).
+Passkeys enable websites to authenticate users without the user having to enter any passwords or other secret codes on the site itself. They address [many of the most serious weaknesses of other authentication methods](#security_properties_of_passkeys) such as passwords.
+
+They're considered [the most secure authentication method available to websites](#security_properties_of_passkeys), and we recommend that sites should adopt passkeys as their preferred authentication method, and [phase out the use of passwords](#migrating_from_passwords).
 
 Instead of a shared secret, passkeys rely on public-key cryptography. A passkey is a {{glossary("Public-key cryptography", "public/private key pair")}} bound to a specific user's account on a particular website.
 


wbamberg:
Some questions.

## Attestation

Spec says (https://w3c.github.io/webauthn/#sctn-attestation-limitations):

> However, it is important to note that an attestation statement, on its own, provides no means for a Relying Party to verify that an attestation object was generated by the authenticator the user intended, and not by a man-in-the-middle attacker. For example, such an attacker could use malicious code injected into Relying Party script. The Relying Party must therefore rely on other means, e.g., TLS and related technologies, to protect the attestation object from man-in-the-middle attacks.
>
> ...In all cases it is possible for a man-in-the-middle attacker to replace the PublicKeyCredential object, including the attestation statement and the credential public key to be registered, and subsequently tamper with future authentication assertions scoped for the same Relying Party and passing through the same attacker.

I don't understand this. If the attestation is signed by a key rooted in "ACME Authenticator Inc", how can an attacker replace the public key credential in the attestation without breaking the signature (or at least replacing iut with a new signature whose key is _not_ rooted in "ACME Authenticator Inc")? 

Also, what should our recommendation be about attestation? Some docs I've read basically un-recommend it for websites. Is that what we should do?

## Language around "user verification"

I struggle not to call this "authentication", like "the user authenticates themselves to the authenticator". Is this OK?

## Passkey portability/transferring passkeys

I haven't really looked into this yet but I'd be happy to get any pointers about what to say here.

timcappalli:
Hi! Typically MDN covers the web platform components, not the entire ecosystem. Passkeys are just one type of credential that can be created via the Web Authentication API.

We would love for you to contribute some of this content to passkeys.dev, which is the ecosystem level developer resource for passkeys. This site is maintained by members of W3C and FIDO Alliance, who work within the wider passkeys ecosystem.

torgo:
Hi @timcappalli one of the ethoses (ethii?) of [Open Web Docs](https://openwebdocs.org) (who are doing this work) is to bring the web documentation to the places where the web developers are present - and right now that's MDN. (And also - through maintenance of the browser compatibility data - places like CanIUse.com.) I'd encourage you to consider the advantages of having high quality, cross-platform documentation where developers are actually looking for it.

wbamberg:
> Hi! Typically MDN covers the web platform components, not the entire ecosystem. Passkeys are just one type of credential that can be created via the Web Authentication API.
>
> We would love for you to contribute some of this content to passkeys.dev, which is the ecosystem level developer resource for passkeys. This site is maintained by members of W3C and FIDO Alliance, who work within the wider passkeys ecosystem.

I'm not sure whether this comment is suggesting that we _should not_ document passkeys on MDN, or that we _should also_ document passkeys on passkeys.dev.

MDN does have [inclusion guidelines](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Criteria_for_inclusion), and among other things they say:

> Our mission statement is "to provide developers with the information they need to easily build projects on the open web".

Helping web developers understand how to work with passkeys seems straightforwardly included in that. I do agree that there are aspects of passkeys that are not in scope for MDN (in particular, documentation for other audiences such as  authenticator/OS/browser vendors), but that's not the subject of this article.

I'm not against contributing to [passkeys.dev](http://passkeys.dev/), but we're a small team with limited resources and have to decide where we can maximise the effect of our work. The advantage of contributing docs to MDN is that it is a very popular and trusted docs site, and we are likely to reach a lot of web developers by contributing to it. Please also note that MDN content is [CC-BY-SA licensed](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Attrib_copyright_license), so it's fine to include MDN content elsewhere, subject to the terms of that license.

I hope we will get input from subject matter experts, to help ensure our docs are giving web developers the best possible guidance. We've benefited greatly from such involvement in our other security docs work such as the guides on XSS, XS-leaks, prototype pollution, etc.

nadalin:
Appreciate the comments but this out of the scope of the WebAuthn WG or specification, I also encourage people withing the passkey eco system to contribute

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+### User verification
+
+When a website calls {{domxref("CredentialsContainer.create()")}} to create a new passkey, or calls {{domxref("CredentialsContainer.get()")}} to create an assertion, the authenticator will always ask the user to consent to the operation.
+
+The RP can also ask the authenticator to perform _user verification_, which means the user will be asked to authorize the use of their credential, for example by entering a PIN or a biometric such as a fingerprint.
+
+When this happens, it's considered to be a form of {{glossary("multi-factor authentication")}}: the authenticator itself is "something the user has" while the PIN or biometric are respectively "something they know" or "something they are".
+
+Note that not all authenticators support user verification.
```
wbamberg, line 160: I believe (although I'm not sure) that passkeys should always require user verification, and if so, we should say this here.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+The WebAuthn specification distinguishes between _discoverable_ and _non-discoverable_ credentials.
+
+- Discoverable credentials are those that can be used without the RP first needing to identify the user who is being authenticated: that is, the "allowed credentials" array passed into {{domxref("CredentialsContainer.get()")}} can be empty.
+
+- Non-discoverable credentials are those for which the RP must first identify the user who is being authenticated (for example, by having them enter their username), and then pass the associated credential ID into {{domxref("CredentialsContainer.get()")}}, in the "allowed credentials" array.
+
+To create a discoverable credential, the RP should set the `residentKey` option to `"required"` and the `requireResidentKey` option to `true` when it creates a new credential in the {{domxref("CredentialsContainer.create()")}} call.
+
+_Passkeys must always be discoverable credentials, so RPs implementing passkey-based authentication should always set these options_.
```
wbamberg: We could perhaps say here that one advantage of discoverable (aka resident) keys is that they can be used with techniques like autofill UI (explained further on in the doc) and that's why passkeys must be discoverable.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Creating passkeys alongside passwords
+
+The first step here is to offer users the opportunity to create a passkey when they successfully sign into your site with a password.
+
+#### Conditional create
+
+An additional step is a feature called _conditional create_. This allows an RP to create a new passkey for a user's account without requiring any user interaction, when certain conditions are met.
+
+To enable conditional create, the RP calls {{domxref("CredentialsContainer.create()")}}, passing the [`mediation`](/en-US/docs/Web/API/CredentialsContainer/create#mediation) option set to `"conditional"`:
```
wbamberg, line 303: This link doesn't work, yet. See https://github.com/mdn/content/issues/42516.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+The RP's front-end sends the assertion to the server, which verifies the signature using the public key it stored. If verification is successful, then the user can be signed in.
+
+## Features of WebAuthn
+
+In this section we'll go into some more detail about various aspects of the WebAuthn API.
+
+### Platform and roaming authenticators
+
+The WebAuthn API distinguishes two main types of authenticator:
```
wbamberg:
```suggestion
The WebAuthn API distinguishes two types of authenticator:
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+}
+```
+
+With this option:
+
+- If the user has just signed in with a password, using a password manager that also supports passkeys (that is, a _credentials manager_ that can also function as an authenticator), then the browser will ask that credentials manager to create a new passkey for the user, without asking the user.
+
+- Otherwise, the `create()` call will fail.
+
+From the user's point of view, if the create failed, they don't know it was made, and if it succeeds, the RP can inform them that they havce a passkey that they can use to sign in next time.
```
wbamberg:
```suggestion
From the user's point of view, if the create failed, they don't know it was made, and if it succeeds, the RP can inform them that they have a passkey that they can use to sign in next time.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+- Summarise the [security properties of passkeys](#security_properties_of_passkeys).
+- Explore some good practices to help prevent users from being locked out if they [lose their passkeys](#handling_lost_passkeys), and to help users [migrate from passwords](#migrating_from_passwords).
+
+## The WebAuthn API
+
+To interact with an authenticator, a website uses the [Web Authentication API (WebAuthn)](/en-US/docs/Web/API/Web_Authentication_API). In the WebAuthn specification, a website that uses passkeys to authenticate users is called a _Relying Party_ (RP), and we'll use that term in this guide.
+
+WebAuthn is an extension of the [Credential Management API](/en-US/docs/Web/API/Credential_Management_API), which is a framework for managing {{glossary("credential", "credentials")}} for various authentication methods, including [passwords](/en-US/docs/Web/Security/Authentication/Passwords) and [federated identity](/en-US/docs/Web/Security/Authentication/Federated_identity), as well as passkeys.
+
+The two main APIs are:
```
hamishwillee: Maybe

```suggestion
The two main methods are:
```

hamishwillee: Though what I'm really thinking is main "for who" -so maybe  "two main methods for RPs"

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+- **Authenticator preferences**: What [type of authenticator](#platform_and_roaming_authenticators) to use, whether the authenticator should perform [user verification](#user_verification) before creating the passkey.
+
+- **Challenge**: The [challenge](#challenges) generated by the RP's server.
+
+- **Website information**: A human readable name and ID for the RP that will be associated with the new passkey. The ID determines the [scope](#passkey_scope) of the resulting passkey.
+
+- **User information**: Information about the user that will be associated with the new passkey, including a human-readable display name, an account identifier, and a human-readable account identifier such as an email address or username.
+
+Depending on the authenticator capabilities and the preferences of the RP, the authenticator may ask the user to authorize passkey creation via some [user verification](#user_verification) method: for example, using a {{glossary("biometric")}} such as a fingerprint.
```
hamishwillee: I'm getting a redlink on biometric

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+![Overview of user registration with passkeys.](passkeys-register.svg)
+
+When the user asks to register on a site, the RP's front-end code first asks its server for a [_challenge_](#challenges): this is a random value generated on the server, that the server will later use to ensure that the resulting passkey was generated in response to this request.
+
+Next, the RP's front-end code calls {{domxref("CredentialsContainer.create()")}}. It can specify various options, including:
+
+- **Attestation preferences**: Whether the RP is interested in authenticator [attestation](#attestation), and if so, what form the attestation should take.
+
+- **Authenticator preferences**: What [type of authenticator](#platform_and_roaming_authenticators) to use, whether the authenticator should perform [user verification](#user_verification) before creating the passkey.
```
hamishwillee: Assuming this is a list of two this might be better.

```suggestion
- **Authenticator preferences**: What [type of authenticator](#platform_and_roaming_authenticators) to use, and whether the authenticator should perform [user verification](#user_verification) before creating the passkey.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+page-type: guide
+sidebar: security
+---
+
+Passkeys enable websites to authenticate users without the user having to enter any passwords or other secret codes on the site itself. They're considered [the most secure authentication method available to websites](#security_properties_of_passkeys), and we recommend that sites should adopt passkeys as their preferred authentication method, and [phase out the use of passwords](#migrating_from_passwords).
+
+Instead of a shared secret, passkeys depend on public key cryptography. A passkey is a {{glossary("Public-key cryptography", "public/private key pair")}} that's specific to a particular user's account on a particular website.
+
+The private key is stored in a module called an _authenticator_, that's [in, or attached to, the user's device](#platform_and_roaming_authenticators). The public key is stored in the website's server. When the user signs in, the authenticator uses the private key to {{glossary("digital signature", "digitally sign")}} a statement about the user's identity, which is called an _assertion_. The website's server can use the public key to verify the assertion's signature, and sign the user in.
```
hamishwillee: You might want to give examples such as below to outline what a typical authenticator is "at this level". Of course many readers will draw the conclusion anyway.

```suggestion
The private key is stored in a module called an _authenticator_, that's [in, or attached to, the user's device](#platform_and_roaming_authenticators), such as a credential manager or some kind of hardware key. The public key is stored in the website's server. When the user signs in, the authenticator uses the private key to {{glossary("digital signature", "digitally sign")}} a statement about the user's identity, which is called an _assertion_. The website's server can use the public key to verify the assertion's signature, and sign the user in.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+- **Challenge**: The [challenge](#challenges) generated by the RP's server.
+
+- **Website information**: A human readable name and ID for the RP that will be associated with the new passkey. The ID determines the [scope](#passkey_scope) of the resulting passkey.
+
+- **User information**: Information about the user that will be associated with the new passkey, including a human-readable display name, an account identifier, and a human-readable account identifier such as an email address or username.
+
+Depending on the authenticator capabilities and the preferences of the RP, the authenticator may ask the user to authorize passkey creation via some [user verification](#user_verification) method: for example, using a {{glossary("biometric")}} such as a fingerprint.
+
+The authenticator then creates a passkey for the account. It stores the private key locally and returns an object containing the public key, challenge, and some additional information. If the authenticator is performing attestation, then this is all {{glossary("digital signature", "digitally signed")}} with either the private key or an [attestation](#attestation) key belonging to the authenticator.
```
hamishwillee, line 59: I started this review by reading passwords topic, which explains that you store a hash to avoid the password leaking via Database attack. I guess with passkeys you store a public key on the site, so there is nothing to leak that isn't public anyway.  In fact that is pretty much true of all the attacks since your device/manager holds the stuff you want to protect.

Or to put it another way, later on I'm hoping for a section on attacks that perhaps explains why passkeys are more resistant to many attacks.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+In this section we'll walk through the flow used to create a new passkey and use it to set up a new user account.
+
+![Overview of user registration with passkeys.](passkeys-register.svg)
+
+When the user asks to register on a site, the RP's front-end code first asks its server for a [_challenge_](#challenges): this is a random value generated on the server, that the server will later use to ensure that the resulting passkey was generated in response to this request.
+
+Next, the RP's front-end code calls {{domxref("CredentialsContainer.create()")}}. It can specify various options, including:
+
+- **Attestation preferences**: Whether the RP is interested in authenticator [attestation](#attestation), and if so, what form the attestation should take.
```
hamishwillee: If you can fit a succinct explanation so readers don't have to read forward then IMO worth it for new terms. Something like

```suggestion
- **Attestation preferences**: Whether the RP is interested in authenticator [attestation](#attestation) (a mechanism for the RP to decide if it trusts the authenticator), and if so, what form the attestation should take.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+When an RP creates a new passkey it can ask which type of authenticator it wants to use, as part of the [`authenticatorSelection`](/en-US/docs/Web/API/PublicKeyCredentialCreationOptions#authenticatorselection) option that it passes to {{domxref("CredentialsContainer.create()")}}.
+
+The main advantage of a platform authenticator is that it's convenient for the user: they don't have to keep track of a separate piece of hardware. The main disadvantage is that it can only be used with its host device.
+
+Platform authenticators can sometimes function as roaming authenticators: for example, a platform authenticator on a mobile device might be available to a laptop as a roaming authenticator, via a Bluetooth connection.
+
+Although platform authenticators can't be removed from their device, they might be able to share their passkeys with other authenticators via cloud sync or import/export functions. For example, a platform provider might enable users to share their passkeys across all the devices belonging to their product family.
```
hamishwillee: Just because it is so common, maybe

```suggestion
Although platform authenticators can't be removed from their device, they can often share their passkeys with other authenticators via cloud sync or import/export functions. For example, a platform provider might enable users to share their passkeys across all the devices belonging to their product family.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+The RP's front-end sends the assertion to the server, which verifies the signature using the public key it stored. If verification is successful, then the user can be signed in.
+
+## Features of WebAuthn
+
+In this section we'll go into some more detail about various aspects of the WebAuthn API.
+
+### Platform and roaming authenticators
+
+The WebAuthn API distinguishes two types of authenticator:
```
hamishwillee: If we strongly suggest credential managers, should we talk about where they fit into this categorization?

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: You could. The first thing I thought here was "so why both options, and which is better"? Which should I use? 
EDIT, that is partially covered by next section. 

You might forward reference a bit at the top

> Discoverable credentials ("resident keys") are those that

Either way I'd be tempted to "also" cover benefits like this in a section on "why passkeys are so resistant to attacks.".

Also, did you consider a `> [!NOTE]` instead of italic? It isn't really an "aside" so I get it.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+When an RP asks an authenticator to create a new passkey or to use an existing passkey, it must provide a _challenge_. This is a random value, specific to the request, that would not be predictable by an attacker. The challenge must be generated in a trusted environment (which generally means, on the server, not the front end).
+
+The RP's front-end code passes the challenge into the `create()` or `get()` call, and the browser includes the same value in the object returned by these methods. In the case of `get()`, the challenge value is also part of the input to the digital signature calculated by the authenticator.
+
+When the web server verifies the response from the authenticator, the web server needs to check that the challenge is the same value it originally provided.
+
+The web server should also invalidate the challenge value after about 10 minutes, and reject any responses containing the challenge that have arrived after this time.
+
+The challenge represents evidence that the authenticator's response was a response to _this_ request, and not an old response to some previous request that an attacker has managed to steal. This kind of attack is known as a _replay attack_.
```
hamishwillee:
```suggestion
The challenge represents evidence that the authenticator's response was a response to _this_ request, and not an old response to some previous request that an attacker has managed to steal. This kind of attack is known as a {{glossary("replay attack")}}.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+This means that by default, a passkey can only be used by a page from the same origin (excluding the port) as the page that originally created it.
+
+Websites are allowed to relax these rules, within some constraints.
+
+- When a website creates a passkey, it can pass an ID into {{domxref("CredentialsContainer.create()")}}, and the authenticator will use this as the RP ID.
+
+- Similarly, when a website tries to use a passkey, it can pass an ID into {{domxref("CredentialsContainer.get()")}}, and the authenticator will compare this ID with the stored RP ID.
+
+For both `create()` and `get()`, the passed value must be a {{glossary("registrable domain")}} that is a _domain suffix_ of the domain of the caller's origin.
```
hamishwillee, line 180: Redlink on the glossary

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+By default:
+
+- When a page creates a passkey by calling {{domxref("CredentialsContainer.create()")}}, the browser sets the passkey's _RP ID_ to the domain component of the caller's {{glossary("origin")}}, and the authenticator stores this value alongside the passkey.
+
+- When a page uses a passkey by calling {{domxref("CredentialsContainer.get()")}}, the browser passes the domain component of the caller's {{glossary("origin")}} to the authenticator, and the authenticator will only allow the passkey to be used if this value matches the stored RP ID.
+
+This means that by default, a passkey can only be used by a page from the same origin (excluding the port) as the page that originally created it.
+
+Websites are allowed to relax these rules, within some constraints.
```
hamishwillee: Maybe:

```suggestion
Websites are allowed to relax these rules, within some constraints:
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+- If the caller was embedded as an {{htmlelement("iframe")}}, whether the caller had the same origin as the top-level document.
+- The origin of the top-level document, if the caller was embedded as an {{htmlelement("iframe")}} and was not same-origin with the caller.
+
+When the RP server verifies the assertion, it must check that these values are what it expects to see.
+
+This provides a layer of protection against [phishing](/en-US/docs/Web/Security/Attacks/Phishing) attacks, in addition to that provided by [passkey scope](#passkey_scope).
+
+## Security properties of passkeys
+
+We can note some features of this design that make it inherently more secure than passwords:
```
hamishwillee:
```suggestion
Passkeys are inherently more secure than passwords:
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+- The {{glossary("origin")}} of the document that called {{domxref("CredentialsContainer.get()")}}.
+- If the caller was embedded as an {{htmlelement("iframe")}}, whether the caller had the same origin as the top-level document.
+- The origin of the top-level document, if the caller was embedded as an {{htmlelement("iframe")}} and was not same-origin with the caller.
+
+When the RP server verifies the assertion, it must check that these values are what it expects to see.
+
+This provides a layer of protection against [phishing](/en-US/docs/Web/Security/Attacks/Phishing) attacks, in addition to that provided by [passkey scope](#passkey_scope).
+
+## Security properties of passkeys
```
hamishwillee, line 200: Ah, I guess this is the answer to my questions about "why are passkeys so great, comparatively". 

Great. YOu might cover this more exhaustively, including things like size outside of their control and resistance to password lists etc - all the stuff in the passwords that counts as attacks.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+- [Enabling users to use their passkeys, instead of their passwords](#using_passkeys_alongside_passwords)
+- [Enabling users to delete their passwords](#retiring_passwords)
+
+### Creating passkeys alongside passwords
+
+The first step here is to offer users the opportunity to create a passkey when they successfully sign into your site with a password.
+
+#### Conditional create
+
+An additional step is a feature called _conditional create_. This allows an RP to create a new passkey for a user's account without requiring any user interaction, when certain conditions are met.
```
hamishwillee: I'm not sure about this first sentence - step in what flow? Perhaps "Another feature you can use is"

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,330 @@
...
+
+With this option:
+
+- If the user has just signed in with a password, using a password manager that also supports passkeys (that is, a _credentials manager_ that can also function as an authenticator), then the browser will ask that credentials manager to create a new passkey for the user, without asking the user.
+
+- Otherwise, the `create()` call will fail.
+
+From the user's point of view, if the create failed, they don't know it was made, and if it succeeds, the RP can inform them that they have a passkey that they can use to sign in next time.
+
+The theory here is that if the user is relying on a credentials manager for sign-in already, then they implicitly trust it to look after their sign-in credentials _in general_, so they can trust it to create a new credential form for them.
```
hamishwillee:
```suggestion
The theory here is that if the user is relying on a credentials manager for sign-in already, then they implicitly trust it to look after their sign-in credentials _in general_, so they can trust it to create a new form of credential for them.
```

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: Ah "ish" https://github.com/mdn/content/pull/42338/files#r2650138701

hamishwillee (commented):
@wbamberg Great job. There is nothing that I do not think I broadly understand, or could not understand with minimal effort. There are some English nits, which are perhaps just matters of taste.

The one thing you might consider is https://github.com/mdn/content/pull/42338/files#r2650138701 -. By contrast with passwords there is only a very high level summary of some of the benefits somewhere near the end. 
Perhaps worth moving this to the end and making a little more exhaustive and comparative.
That would better support the initial recommendation at the top of the page that people should use passkeys and why.

Do we need to talk about retiring passkeys?

wbamberg:
> Do we need to talk about retiring passkeys?

You mean, for example, if a user loses their authenticator, they'll want to de-activate the passkeys it contains? I think we perhaps do and I was wondering about it. I thought we could have a section "Managing passkeys" that covers:

- how a user needs to be able to see the passkeys for the site and delete their entries on the site.
- how the site can make sure that the credentials it recognises are kept in sync with the passkeys living in various authenticators - the WebAuthn API provides a set of methods for this (the "signal API") that we could summarise here

Does that sound like it would cover your comment?

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: Well AIUI all authenticators are "credential managers" - to the extent that a credential manager is an entity that looks after your credentials. But not all credential managers are authenticators (some might only be capable of password storage).

So I'm not sure we do "strongly suggest credential managers". I **think** that what we strongly suggest is (1) that RPs use passkeys and (2) that they encourage users to create passkeys in multiple authenticators, especially in a mix of platform and roaming authenticators, to make users more resilient against authenticator loss.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: See https://github.com/mdn/content/pull/42338/files#r2650360176 - AFAICT a hardware key is also a credential manager. We could perhaps try some alternative wording here, but I'm worried about trying to pack too much into this intro, which I do want to keep pretty quick. There's lots of space to get into the details later on. But I'll think about it.

(I did originally have a separate "authenticators" H2 at the start, here, but thought it slowed things down, so collapsed it into the intro. Maybe I took out too much.)

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: yes, this isn't written yet. I should at least file an issue for it though.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: yes, also not written yet. Most of the relevant content is in the entry for "site", but IMO it should be split out of there.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: > "so why both options, and which is better"? Which should I use?

Yeah, it's a weird bit of the spec. Basically, for passkeys, you have no choice, they are by definition resident AKA discoverable. So none of this is really actionable. But I think it is still worth explaining what the concept is, otherwise it is very mysterious.

It took me a while to understand exactly why non-resident keys are a thing, it isn't well explained. https://fy.blackhats.net.au/blog/2023-02-02-how-hype-will-turn-your-security-key-into-junk/ is interesting here.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I guess, another step towards getting users to adopt passkeys? Maybe I should just say that :)

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: -> https://github.com/mdn/content/pull/42338/commits/589aa1e9d298dbb455b5df246a6b12f9c4b90370

hamishwillee:
> Does that sound like it would cover your comment?

Yes and yes. Covering key management/lifecycle would be great.

My thinking was mostly about comparison to password loss/management - where you do need to retire old and compromised keys. Somewhere in this you talk about backups, and my mind was going towards what you do if you create too many backups and need to remove some keys so you can manage them better.

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: I guess I'm still hanging onto the idea that people know what a credential manager is, but they don't necessarily grok what an authenticator is. Which is why I keep trying to use the comfortable terminology. 

Fine not to add it though, you have just defined what they do.

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: Just the right level IMO.

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: That link is interesting and useful thanks.

IMO this and the next section are not as strong as some the others. Mostly because we understand from this that non-resident is possibly good, but not then why passkeys must be resident and what implications this has.

Unfortunately I don't know how to "fix" it. If it were me I'd probably try combine the sections in one and make an explicit comment that explains that the implication.

Or you could just leave it. "Not as strong" is not "rubbish".

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: From a WebAuth point of view my understand is "no" - user verification is treated as a second level of authentication, after posession of the device that has the key.

FIDO https://fidoalliance.org/passkeys/#faq FAQ describes it as though it is required, but in https://fidoalliance.org/specifications/ you see "Biometric information, **if used**, never leaves the user’s device."

I leant towards "strongly encouraged".

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I agree that this section is a bit weird, but I think it is worth keeping. The reason I went into this detail is that I read [MDN on discoverable credentials](https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API#discoverable_credentials_and_conditional_mediation) and found it incredibly confusing. It goes on about how you have to make your keys discoverable, and why this is good, and how you do it, but never explains why some keys are not discoverable, or, critically, what the point is in having any non-discoverable keys.

So as a reader you think, well, why even have non-discoverable keys? Why is there this flag I have to set, to enable this "better key"? Why aren't they all better?

But actually non-discoverable, aka non-resident, keys are a very neat idea, because you can store (many) more keys in a relatively small USB device.

I don't want to combine "autofill UI" into this section, because this section is background explanation, and autofill is more of a how-to (part of how to manage a userbase who have a mixture of passwords and passkeys).

But I think I should _mention_ autofill UI here, and basically say:

- non-resident keys have this space-saving advantage over resident keys, and that's why they are supported in the WebAuthn spec
- but you can't do autofill UI with them
- and for this reason (and maybe some others), passkeys always have to be resident

hamishwillee: This may well work. It certainly gives the section more context and I think it is worth trying.

As I said though, I don't know the "optimal" way to fix this without trying some alternatives. One other idea that popped up reading ^^^ was to rename the resident/non resident section "Why aren't passkeys non-discoverable keys".
That would allow you to explain all of this more naturally.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I'm going to let this issue sit until we get a confirmation one way or another. FWIW I'm OK to merge this PR without resolving the issue, but I don't want us to guess.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: Had a go at redoing this section a bit -> https://github.com/mdn/content/pull/42338/commits/9ecf9e89ab37c49c8cf1b42222afc1933d3a2607.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: -> https://github.com/mdn/content/pull/42338/commits/69d19f2321ad11801ec9481c255b26465fd9fce6 . I don't want to use "method" because we use the same word in the previous para to talk about "authentication methods".

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I don't want to do this because none of the other linked terms do it, and this is really overview, deferring all the details to the next section. I mean if we do this why not also gloss "challenge" ("a mechanism for preventing replay attacks") and "scope" etc, and before you know it this section is quite long and complex and bristling with asides.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: -> https://github.com/mdn/content/pull/42588

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: -> https://github.com/mdn/content/pull/42338/commits/fb5c6f4d2614d4c03a478c5212d59076da2f4190 to be honest I don't like conditional create, at all. I don't think browsers should be able to create passkeys for users without the user consenting.

wbamberg:
> > Does that sound like it would cover your comment?
> 
> Yes and yes. Covering key management/lifecycle would be great.
> 
> My thinking was mostly about comparison to password loss/management - where you do need to retire old and compromised keys. Somewhere in this you talk about backups, and my mind was going towards what you do if you create too many backups and need to remove some keys so you can manage them better.

-> https://github.com/mdn/content/pull/42338/commits/b6eb7d4962c112f60256b09c2ca4b7b8fe349160.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: -> https://github.com/mdn/content/pull/42338/commits/126c5d62ecf258a6d0bbb1cf1debb57b5b355031 

I didn't move it to the end because I think it makes the flow weird:

1. here's what passkeys are
2. here's how to migrate from passwords
3. here's why passkeys are better

versus:

1. here's what passkeys are
2. here's why passkeys are better
3. here's how to migrate from passwords

wbamberg:
Thanks for the review! I made some updates. I think I've tried to address all your comments except https://github.com/mdn/content/pull/42338#discussion_r2650098144.

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: This works for me - it's dense, but it does answer the question.

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: Fair enough. The rewrite is really good.

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: I hadn't thought of that, but you may have a point. Are you any less secure IFF biometrics/pin are required?
I guess in theory you are, because you're creating a bigger attack surface. Also if the key is automatically stored to a hardware key you'd be eating limited resources.

hamishwillee (approved):
Thanks @wbamberg ! The changes all make it much clearer to me. 
I hope to read it with fresh eyes on Monday - just checked the changes this time. I'm approving because I think it makes sense and this is just a final sanity check. Merge if you need to!

wbamberg:
Thanks Hamish, it's much better for your comments, as usual. There's no great urgency to merge it - I still hope to get a technical review.

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: Having re-read this section again I think we should make it clear this is associated with the authenticator, not the process - see also https://github.com/mdn/content/pull/42338/files#r2660175485

```suggestion
- **Attestation preferences**: Whether the RP is interested in [authenticator attestation](#attestation), and if so, what form the attestation should take.
```

hamishwillee: Or maybe better ...

```suggestion
- **Attestation preferences**: Whether the RP is interested in validating that the authenticator can be trusted and if so, what form the [authenticator attestation](#attestation), should take.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+The WebAuthn API distinguishes two types of authenticator:
+
+- **Platform authenticators**
+  - : These authenticators are not removable from the device. For example, authenticators built into the device's operating system, like the [Touch ID](https://en.wikipedia.org/wiki/Touch_ID) system in Apple devices or the [Windows Hello](https://en.wikipedia.org/wiki/Windows_10#System_security) system.
+- **Roaming authenticators**
+  - : These authenticators can be removed from the device and attached to a different device. The classic example of this is an authenticator implemented in a USB key, like a [YubiKey](https://en.wikipedia.org/wiki/YubiKey).
+
+When an RP creates a new passkey it can ask which type of authenticator it wants to use, as part of the [`authenticatorSelection`](/en-US/docs/Web/API/PublicKeyCredentialCreationOptions#authenticatorselection) option that it passes to {{domxref("CredentialsContainer.create()")}}.
+
+The main advantage of a platform authenticator is that it's convenient for the user: they don't have to keep track of a separate piece of hardware. The main disadvantage is that it can only be used with its host device.
```
hamishwillee, line 106: Indeed. Another main benefit, is the fact that you're not limited to the number of resident keys. Perhaps not worth saying, but if so, maybe as a problem with many dedicated-hardware oaming authenticators and passkeys.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: > The reason being that we use attestation all over the place - the term is overloaded.

Where else in this guide do we talk about attestation in any other sense? I can't really think of anywhere in MDN that we use it.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+The RP's front-end code passes the challenge into the `create()` or `get()` call, and the browser includes the same value in the object returned by these methods. In the case of `get()`, the challenge value is also part of the input to the digital signature calculated by the authenticator.
+
+When the web server verifies the response from the authenticator, the web server needs to check that the challenge is the same value it originally provided.
+
+The web server should also invalidate the challenge value after about 10 minutes, and reject any responses containing the challenge that have arrived after this time.
+
+The challenge represents evidence that the authenticator's response was a response to _this_ request, and not an old response to some previous request that an attacker has managed to steal. This kind of attack is known as a {{glossary("replay attack")}}.
+
+### Attestation
```
hamishwillee, line 142: So as per my other comment, here we are talking abou Authenticator attestation and not "the authenticator sending back an attestation". Perhaps 

```suggestion
### Autenticator attestation
```

and corresponding link fixes.

hamishwillee: Actually, I've confused myself - intermixed the terms attestation and assertion in my brain.

So the point is invalid. That said, I still think this change is a good idea, because it makes it more clear it is about the authenticator itself, not "the process"

hamishwillee:
> Thanks Hamish, it's much better for your comments, as usual. There's no great urgency to merge it - I still hope to get a technical review.

I read again. It still feels really solid. A few comments that you might ignore. I like reviewing things that teach me new stuff. In particular when I thought I already understood it.

files/en-us/web/security/authentication/passkeys/index.md
hamishwillee: @wbamberg We don't - I was confusing with assertion as per https://github.com/mdn/content/pull/42338/files#r2660175485 - but either way I'd still consider this change.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+The public key is stored in the website's server. When the user signs in, the authenticator uses the private key to {{glossary("digital signature", "digitally sign")}} a statement about the user's identity, which is called an _assertion_. The website's server can use the public key to verify the assertion's signature, and sign the user in.
+
+In this guide we'll:
+
+- Introduce the [Web Authentication API (WebAuthn)](/en-US/docs/Web/API/Web_Authentication_API), which enables web apps to use passkeys.
+- Go through the two main flows supported by WebAuthn: [registration](#registration) and [sign-in](#sign_in).
+- Explore some of the main [features of the WebAuthn API](#features_of_webauthn).
+- Summarise the [security properties of passkeys](#security_properties_of_passkeys).
```
Elchi3:
```suggestion
- Summarize the [security properties of passkeys](#security_properties_of_passkeys).
```

files/en-us/web/security/authentication/passkeys/index.md
Elchi3:
```suggestion
The WebAuthn API distinguishes two types of authenticators:
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+- {{domxref("PublicKeyCredential.signalAllAcceptedCredentials_static", "PublicKeyCredential.signalAllAcceptedCredentials()")}} gives the browser the identifiers of all the passkeys that the RP currently accepts as valid, to enable all attached authenticators to update their stored keys. It could be called every time the user successfully authenticates. This API must only be called for authenticated users, because it exposes the user's credential IDs.
+
+- {{domxref("PublicKeyCredential.signalCurrentUserDetails_static", "PublicKeyCredential.signalCurrentUserDetails()")}} tells the browser the user's current username and display name, and should be called when an authenticated user changes these values. This API must only be called for authenticated users, because it exposes uses data.
```
Elchi3:
```suggestion
- {{domxref("PublicKeyCredential.signalCurrentUserDetails_static", "PublicKeyCredential.signalCurrentUserDetails()")}} tells the browser the user's current username and display name, and should be called when an authenticated user changes these values. This API must only be called for authenticated users, because it exposes user data.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+This provides a layer of protection against [phishing](/en-US/docs/Web/Security/Attacks/Phishing) attacks, in addition to that provided by [passkey scope](#passkey_scope).
+
+## Security properties of passkeys
+
+Passkeys are more secure than passwords, and we can see how their design addresses the most serious [weaknesses of passwords](/en-US/docs/Web/Security/Authentication/Passwords#weaknesses_of_password-based_authentication):
+
+- Unlike a password, the user never invents a passkey value or needs to remember it. This means users can't choose weak passkey values, so they are not vulnerable to [guessing](/en-US/docs/Web/Security/Authentication/Passwords#guessing) attacks.
+
+- Passkeys are never reused across sites, so that they are not vulnerable to [credential stuffing](/en-US/docs/Web/Security/Authentication/Passwords#credential_stuffing) attacks. If an attacker does get access to a passkey, they can only use it for the site that originally created it.
```
wbamberg:
```suggestion
- Passkeys are never reused across sites, so they are not vulnerable to [credential stuffing](/en-US/docs/Web/Security/Authentication/Passwords#credential_stuffing) attacks. If an attacker does get access to a passkey, they can only use it for the site that originally created it.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
+---
+title: Passkeys
+slug: Web/Security/Authentication/Passkeys
+page-type: guide
+sidebar: security
+---
+
+Passkeys enable websites to authenticate users without the user having to enter any passwords or other secret codes on the site itself. They're considered [the most secure authentication method available to websites](#security_properties_of_passkeys), and we recommend that sites should adopt passkeys as their preferred authentication method, and [phase out the use of passwords](#migrating_from_passwords).
```
simoneonofri:
```suggestion
Passkeys enable websites to authenticate users without requiring users to enter any passwords or other secret codes on the site itself. When properly implemented, passkeys provide stronger resistance to
phishing, credential reuse, and replay attacks compared to a classical password-based authentication.

For these reasons, websites are adopting passkeys as a preferred authentication method and [phasing out password use](#migrating_from_passwords). However, the security improvements depend on correct implementation and appropriate recovery mechanisms.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+Instead of a shared secret, passkeys depend on public key cryptography. A passkey is a {{glossary("Public-key cryptography", "public/private key pair")}} that's specific to a particular user's account on a particular website.
```
simoneonofri:
```suggestion
Instead of a shared secret, passkeys rely on public-key cryptography. A passkey is a {{glossary("Public-key cryptography", "public/private key pair")}} bound to a specific user's account on a particular website (known as the Relying Party).
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+The private key is stored in a module called an _authenticator_, that's [in, or attached to, the user's device](#platform_and_roaming_authenticators). An authenticator might be built into the platform, like the [Touch ID](https://en.wikipedia.org/wiki/Touch_ID) system in Apple devices, or a separate hardware key like a [YubiKey](https://en.wikipedia.org/wiki/YubiKey), or a credential manager app like [KeePassXC](https://keepassxc.org/).
```
simoneonofri:
```suggestion
The private key is stored in a module called an _authenticator_, that's [in, or attached to, the user's device](#platform_and_roaming_authenticators). An authenticator might be built into the platform, and accessed using biometric (like the [Touch ID](https://en.wikipedia.org/wiki/Touch_ID) or [Face ID](https://en.wikipedia.org/wiki/Face_ID) system in Apple devices), or a separate hardware key like a [YubiKey](https://en.wikipedia.org/wiki/YubiKey), or a credential manager app like [KeePassXC](https://keepassxc.org/).
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+The public key is stored in the website's server. When the user signs in, the authenticator uses the private key to {{glossary("digital signature", "digitally sign")}} a statement about the user's identity, which is called an _assertion_. The website's server can use the public key to verify the assertion's signature, and sign the user in.
```
simoneonofri:
```suggestion
The public key is stored on the website’s server.

When the user signs in, the server sends a {{glossary("digital signature", "digitally signed")}}  strong, unpredictable challenge to the authenticator. The authenticator signs this challenge, together with contextual information such as the requesting origin, using the private key. This is called an _assertion_.

The website's server uses the public key to verify the assertion's signature and sign the user in.
```

files/en-us/web/security/authentication/passkeys/index.md
simoneonofri: +1 on defining attestation in the specific context. There are a lot of definitions, and it's good to specify them here.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+## Sign in
+
+In this section we'll walk through the flow used to sign a user in with a passkey.
+
+![Overview of user sign-in with passkeys.](passkeys-sign-in.svg)
+
+When the user tries to sign in, the RP's front-end code again asks the server for a [challenge](#challenges) value.
+
+Next, the RP's front-end code calls {{domxref("CredentialsContainer.get()")}}. It can specify various options, including:
```
simoneonofri, line 75:
```suggestion
Next, the RP's front-end code calls {{domxref("CredentialsContainer.get()")}}, providing parameters
that constrain which credentials the browser and authenticator may use for authentication. These
parameters can include:
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+- **Allowed credentials**: An array of identifiers for the passkeys that the RP will accept. This array may be empty or omitted, in which case any suitable passkeys may be used.
+
+- **Challenge**: The [challenge](#challenges) generated by the RP's server.
```
simoneonofri, line 79:
```suggestion
- **Challenge**: The [challenge](#challenges) generated by the RP's server. The challenge helps to protect from replay attacks.
```

files/en-us/web/security/authentication/passkeys/index.md
simoneonofri: Whether the passkey can be backed up depends on how it is generated at the RP's request and on the authenticator's capabilities. Using Apple's example again, it is possible to synchronize passkeys across different devices. https://support.apple.com/en-gb/guide/ipad/ipada39a7fa0/ipados, and this is stated some paragraphs later

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+The main advantage of a platform authenticator is that it's convenient for the user: they don't have to keep track of a separate piece of hardware. The main disadvantage is that it can only be used with its host device.
+
+Platform authenticators can sometimes function as roaming authenticators: for example, a platform authenticator on a mobile device might be available to a laptop as a roaming authenticator, via a Bluetooth connection.
+
+Although platform authenticators can't be removed from their device, they can often share their passkeys with other authenticators via cloud sync or import/export functions. For example, a platform provider might enable users to share their passkeys across all the devices belonging to their product family.
+
+### Discoverable and non-discoverable credentials
+
+The WebAuthn specification distinguishes between _discoverable_ and _non-discoverable_ credentials.
```
simoneonofri, line 114:
```suggestion
The WebAuthn specification distinguishes between _discoverable_ and _non-discoverable_ credentials, which differ in how credentials are selected during authentication and have important usability and
security implications.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+- **Discoverable credentials**, also known as _resident keys_, are those that can be used without the RP first needing to identify the user who is being authenticated: that is, the "allowed credentials" array passed into {{domxref("CredentialsContainer.get()")}} can be empty. With a discoverable credential, all the signing key material is stored in the authenticator, so the authenticator is able to generate signatures without needing any input from the RP.
```
simoneonofri, line 116:
```suggestion
- **Discoverable credentials**, also known as _resident keys_, are those that can be used without the RP first needing to identify the user who is being authenticated: that is, the "allowed credentials" array passed into {{domxref("CredentialsContainer.get()")}} can be empty. With the discoverable credential, all signing key material is stored in the authenticator, allowing it to generate signatures without any input from the RP. With discoverable credentials, the authenticator retains sufficient credential information to locate and use the appropriate private key without additional input from the RP. This enables account selection and autofill-style user experiences.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+  Non-discoverable credentials need the credential ID because they do not store the signing key itself in the authenticator, but instead generate the signing key every time it is needed, from an internal seed and the credential ID value. That is, the account key is not resident in the authenticator.
+
+The advantage of using non-discoverable credentials is that an authenticator with limited storage can support a potentially unlimited number of accounts, because the key material for each account is not stored in the authenticator.
+
+The advantage of using discoverable credentials is that they enable a browser to implement [autofill](/en-US/docs/Web/Security/Authentication/Passkeys#autofill_ui) with public key credentials, which makes it much easier for users to sign in, especially when they might have both public key credentials and passwords for a given site.
```
simoneonofri, line 124:
```suggestion
One of the advantages of using discoverable credentials is that they enable a browser to implement [autofill](/en-US/docs/Web/Security/Authentication/Passkeys#autofill_ui) with public key credentials, which makes it much easier for users to sign in, especially when they might have both public key credentials and passwords for a given site, and generally reduces friction during sign-in.

File: files/en-us/web/security/authentication/passkeys/index.md [hunk 35]
```diff
@@ -0,0 +1,372 @@
...
+
+**For this reason, passkeys must always be discoverable credentials, so RPs implementing passkey-based authentication should always make them discoverable**.
```
simoneonofri, line 126:
```suggestion
**For this reason, passkeys can be defined as discoverable credentials, so RPs implementing passkey-based authentication should always make them discoverable**.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+When the web server verifies the response from the authenticator, the web server needs to check that the challenge is the same value it originally provided.
+
+The web server should also invalidate the challenge value after about 10 minutes, and reject any responses containing the challenge that have arrived after this time.
+
+The challenge represents evidence that the authenticator's response was a response to _this_ request, and not an old response to some previous request that an attacker has managed to steal. This kind of attack is known as a {{glossary("replay attack")}}.
+
+### Attestation
+
+The security of a passkey depends on the reliability of the authenticator used. For example, if an authenticator does not protect the private keys it stores, then an attacker could steal the keys and impersonate users. In _attestation_, an authenticator provides verifiable evidence to the RP about its origin, and about the data it produces (such as key pairs or signed assertions). This can help the RP decide whether it wants to rely on the authenticator to authenticate its users.
```
simoneonofri:
```suggestion
The security of a passkey depends in part on the reliability of the authenticator used. For example, if an authenticator does not protect the private keys it stores, then an attacker could steal the keys and impersonate users. WebAuthn defines an optional mechanism called _attestation_, in which an authenticator can provide verifiable evidence to the RP about the authenticator and the data it produces (such as key pairs or signed assertions). This can help the RP decide whether it wants to rely on the authenticator to authenticate its users.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+To implement attestation, the authenticator contains a key pair called an _attestation key_, which was built into the device at manufacturing time, and which is {{glossary("digital certificate", "certified")}} as belonging to the organization that made this authenticator. For example, the certificate could state that this authenticator was produced by "Acme Authenticator Incorporated".
```
simoneonofri, line 146:
```suggestion
To implement attestation, the authenticator may use a key pair called an _attestation key_, which was built into the device at manufacturing time, a {{glossary("digital certificate", "certified")}} belonging to the authenticator manufacturer. For example, the certificate could state that this authenticator was produced by "Acme Authenticator Incorporated".
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+When the authenticator creates a new passkey, it signs the resulting object with its attestation key. The RP verifies the signature and the associated certificate, and then has evidence that the passkey was made by an authenticator produced by "Acme Authenticator Incorporated".
```
simoneonofri, line 148:
```suggestion
When the authenticator creates a new passkey, it may include an attestation statement signed with its attestation key. The RP can verify the signature and the associated certificate, and then has evidence that the passkey was made by an authenticator produced by "Acme Authenticator Incorporated".
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+Not all authenticators support attestation, and RPs can indicate that they are not interested in attestation. In these situations, the object returned by a call to {{domxref("CredentialsContainer.create()")}} may not be signed at all, or it may be signed using the passkey itself (this is referred to as _self attestation_). In these situations, the RP has no reliable evidence of the authenticator's origin or capabilities.
```
simoneonofri:
```suggestion
Not all authenticators support attestation, and RPs may indicate that they are not interested in attestation. In these situations, the object returned by a call to {{domxref("CredentialsContainer.create()")}} may not be signed at all, or it may be signed using the passkey itself (this is referred to as _self attestation_). In these situations, the RP has no reliable evidence of the authenticator's origin or capabilities.

In this case, even if the RPs do not receive verified information about the authenticator's origin, they can still use secure credentials. Also, this can improve user privacy by preventing the disclosure of the authenticator's information.  
```

files/en-us/web/security/authentication/passkeys/index.md
simoneonofri: it is an option

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+The signed [assertion](#assertions) returned by an authenticator includes information about the context of the caller:
+
+- The {{glossary("origin")}} of the document that called {{domxref("CredentialsContainer.get()")}}.
+- If the caller was embedded as an {{htmlelement("iframe")}}, whether the caller had the same origin as the top-level document.
+- The origin of the top-level document, if the caller was embedded as an {{htmlelement("iframe")}} and was not same-origin with the caller.
+
+When the RP server verifies the assertion, it must check that these values are what it expects to see.
+
+This provides a layer of protection against [phishing](/en-US/docs/Web/Security/Attacks/Phishing) attacks, in addition to that provided by [passkey scope](#passkey_scope).
+
```
simoneonofri, line 199: Maybe later, it should be possible to add the privacy-discussed Related Origin Feature https://passkeys.dev/docs/advanced/related-origins/

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+## Security properties of passkeys
+
+Passkeys are more secure than passwords, and we can see how their design addresses the most serious [weaknesses of passwords](/en-US/docs/Web/Security/Authentication/Passwords#weaknesses_of_password-based_authentication):
+
+- Unlike a password, the user never invents a passkey value or needs to remember it. This means users can't choose weak passkey values, so they are not vulnerable to [guessing](/en-US/docs/Web/Security/Authentication/Passwords#guessing) attacks.
```
simoneonofri:
```suggestion
- Unlike a password, the user never invents a passkey value or needs to remember it. This means users can't choose weak passkey values, so they are not vulnerable to [guessing](/en-US/docs/Web/Security/Authentication/Passwords#guessing) attacks. Passkey generation is transferred from the user to the authenticator.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+- Passkeys are never reused across sites, so they are not vulnerable to [credential stuffing](/en-US/docs/Web/Security/Authentication/Passwords#credential_stuffing) attacks. If an attacker does get access to a passkey, they can only use it for the site that originally created it.
```
simoneonofri, line 206: Yes and no, see the concept of RORs.

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+- With passkeys, the server never has to store any secrets, so if an attacker [breaks into the server's database](/en-US/docs/Web/Security/Authentication/Passwords#database_compromise), they can't compromise accounts using the stored data. Note, however, that they can compromise user accounts if they can _write_ fake credentials into the server's database.
```
simoneonofri:
```suggestion
- With passkeys, the server never has to store any secrets: it just stores the public key. So if an attacker [breaks into the server's database](/en-US/docs/Web/Security/Authentication/Passwords#database_compromise), they can't compromise the private key, which is stored in the authenticator. Note, however, that they can compromise user accounts if they can _write_ fake credentials into the server's database.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+- When the user tries to sign in, the browser will only look for passkeys whose scope matches the requesting site, and the RP's server can verify that the origin of the requester was what they expected. This makes passkeys resistant to [phishing](/en-US/docs/Web/Security/Attacks/Phishing) attacks, because front-end code served from a phishing site like `https://examp1e.com` is not able to use the passkey associated with `https://example.com`.
+
```
simoneonofri, line 211:
```suggestion
These properties significantly reduce common web authentication attacks, but they do not eliminate
all the threats, such as device compromise or insecure account recovery mechanisms.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+Since widespread deployment of passkeys is relatively new, there isn't yet a mature understanding of the attacks that passkeys may face. It's likely that attacks would focus on the user's devices: for example, convincing them to install a malicious authenticator.
+
+## Handling lost passkeys
+
+If a user loses an authenticator, whether it's a separate module or integrated into their phone, they lose all the passkeys it contains.
```
simoneonofri, line 216:
```suggestion
If a user loses an authenticator, they may lose access to the passkeys stored on that authenticator.
Whether the passkeys are permanently lost depends on the type of authenticator and whether the
passkeys are backed up or synced across devices. In general, even with passkeys, account recovery procedures are important on the RP side.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+An RP can use this information to help a user manage their credentials. For example:
+
+- If the passkey is not backup eligible, then the RP might respond by inviting the user to create another passkey in another authenticator that could be used as a backup.
+
+- If the RP is migrating users away from passwords, and the user has an old password as well as a passkey, and the assertion indicates that the passkey has been backed up, then the RP might invite the user to delete their old password, since they don't need it as a backup any more.
+
+## Managing passkeys
+
+We've seen that a user may have multiple passkeys for a single account, spread across multiple authenticators and multiple devices. Each passkey is represented as a private key stored in the authenticator, and a corresponding public key stored in the RP's server as part of the user's account information.
```
simoneonofri:
```suggestion
We've seen that a user may have multiple passkeys for a single account, distributed across multiple authenticators and multiple devices. Each passkey corresponds to a WebAuthn credential, with private key material protected by the authenticator and a corresponding public key stored by the RP as part of the user's account information.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+Sometimes the user might need to delete a passkey for their account: this essentially means, deleting the public key stored in the RP's server, so that the corresponding private key can't be used to sign the user in any more. This is generally needed when the user doesn't have control of the authenticator any more, for example because they have lost the device containing it.
```
simoneonofri:
```suggestion
Sometimes the user might need to delete a passkey for their RP account: this essentially means deleting the public key stored in the RP's server, so that the corresponding private key can't be used to sign the user in anymore. This is generally needed when the user doesn't have control of the authenticator anymore, for example, because they have lost the device containing it.
```

files/en-us/web/security/authentication/passkeys/index.md
```diff
@@ -0,0 +1,372 @@
...
+
+This means that an RP must implement a means for an authenticated user to view the registered passkeys for their account and delete specific public keys. For each key, the RP should display information to help a user understand which key it is and which authenticator it is associated with. This can include:
```
simoneonofri:
```suggestion
This means that an RP should implement a means for an authenticated user to view the registered passkeys for their account and delete specific public keys. For each key, the RP should display information to help a user understand which key it is and which authenticator it is associated with. This can include:
```

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I introduce the term RP explicitly at the start of [the next section](https://pr42338.review.mdn.allizom.net/en-US/docs/Web/Security/Authentication/Passkeys#the_webauthn_api), so I don't think the definition is needed. But the rest of this comment is fine :).

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: It is my understanding that passkeys must be discoverable:

> The high level, end-user centric term for a FIDO2/WebAuthn [Discoverable Credential](https://passkeys.dev/docs/reference/terms/#discoverable-credential).

(https://passkeys.dev/docs/reference/terms/#passkey)

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: Aren't "makes it much easier for users to sign in" and "reduces friction during sign-in" the same thing? If not, what's the difference?

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I don't want to accept this suggestion because not all the listed options are given to "constrain which credentials the browser and authenticator may use for authentication", in fact AFAICS only the first one is.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: This edit doesn't quite make sense, it will read:

> the authenticator may use a key pair called an _attestation key_, which was built into the device at manufacturing time, a certified belonging to the authenticator manufacturer.

Is it intended as:

> the authenticator may use a key pair called an _attestation key_, which was built into the device at manufacturing time, and a digital certificate belonging to the authenticator manufacturer.

? But if so I am not sure it is an improvement, since (a) istm that the certificate isn't necessarily in the authenticator, and (b) it doesn't capture what istm is the important bit, that the attestation signing key is _certified as_ belonging to the manufacturer.  That is, the edit says there is a key and a certificate, but doesn't link them.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: This is describing what happens when you do attestation, so I'm not sure why we change it it to "may" and "can". You don't have to do attestation, but if you are, then you _do_ include an attestation, and the RP _does_ verify it.

caugner:
> * Macro glossary produces link /en-US/docs/Glossary/biometric which doesn't resolve

Do you intend to add a biometric glossary page? Or is it common to add glossary links to not-yet-written pages?

https://github.com/mdn/content/blob/d6622470b762474e9958a1b811d843a93f933ce8/files/en-us/web/security/authentication/passkeys/index.md?plain=1#L55

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I guess the reasoning behind this edit is that Touch ID isn't an authenticator (in itself)? Which is fair. But this edit  implies that platform authenticators always use biometrics (and by implication, roaming authenticators do not) which is misleading. I think I'd prefer to remove the Apple example:

```suggestion
The private key is stored in a module called an _authenticator_, that's [in, or attached to, the user's device](#platform_and_roaming_authenticators). An authenticator might be built into the platform, or a separate hardware key like a [YubiKey](https://en.wikipedia.org/wiki/YubiKey), or a credential manager app like [KeePassXC](https://keepassxc.org/).
```

files/en-us/web/security/authentication/passkeys/index.md
wbamberg:
```suggestion
Instead of a shared secret, passkeys rely on public-key cryptography. A passkey is a {{glossary("Public-key cryptography", "public/private key pair")}} bound to a specific user's account on a particular website.
```

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: "websites are adopting passkeys as a preferred authentication method and phasing out password use" feels like an observation, but I wanted to make a clear recommendation here. Also, while we could list some of the specific attacks where passkeys do better, I think it is more useful, in the first sentence, to link to somewhere we explain this in more detail.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I don't think the challenge is signed, is it?

```suggestion
The public key is stored on the website's server. When the user signs in, the authenticator uses the private key to {{glossary("digital signature", "digitally sign")}} a [_challenge_](#challenges) value from the server, together with contextual information such as the requesting {{glossary("origin")}}. The resulting object is called an _assertion_. The website's server can use the public key to verify the assertion's signature, and sign the user in.
```

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: -> https://github.com/mdn/content/pull/42338/commits/8fd50819aa5fd866aec0773d86bac440aad92045

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: -> https://github.com/mdn/content/pull/42338/commits/8fd50819aa5fd866aec0773d86bac440aad92045

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I don't think this change really adds anything.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: This seems quite repetitive:

> With the discoverable credential, all signing key material is stored in the authenticator, allowing it to generate signatures without any input from the RP.

> With discoverable credentials, the authenticator retains sufficient credential information to locate and use the appropriate private key without additional input from the RP.

...and also repeats stuff that is in the next para-but-one.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I've not accepted the extra para here because I'm not that sure what "can still use secure credentials" means. It's fairly obvious, isn't it, that you can still use passkeys without attestation? And:

> this can improve user privacy by preventing the disclosure of the authenticator's information.  

...the referent of "this" is unclear, and it is my understanding that privacy-preserving attestation methods also exist. That said, attestation is complex and really getting into it is outside the scope of this article.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: sorry, I accidentally edited your suggestion. My reasoning: the private key is not necessarily in the user's **device** (depending on what "device" means here) and the bit about reuse belong to the previous paragraph.

wbamberg:
> > * Macro glossary produces link /en-US/docs/Glossary/biometric which doesn't resolve
> 
> Do you intend to add a biometric glossary page? Or is it common to add glossary links to not-yet-written pages?
> 
> https://github.com/mdn/content/blob/d6622470b762474e9958a1b811d843a93f933ce8/files/en-us/web/security/authentication/passkeys/index.md?plain=1#L55

Sometimes I add links when I intend to add a glossary page either in this PR or in a quick follow-up. But for now I just removed the link, in the interests of merging this.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: Yeah, I probably should talk about this. But I would like to defer it in the interests of getting this PR merged.

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I think this edit breaks the flow of the section: the next part "in this section" doesn't really make sense after it. With this edit, and looking at the following bit we would end up saying something like:

files/en-us/web/security/authentication/passkeys/index.md
wbamberg: I've tried to integrate this comment into the next paragraph -> https://github.com/mdn/content/pull/42338/commits/6db846ee947f0664409968600ff1d26c50b35c5f.

files/en-us/web/security/authentication/passkeys/index.md
Elchi3: I like this para because we have 3 clear messages:

> [passkeys are] the most secure authentication method available to websites
> we recommend that sites should adopt passkeys as their preferred authentication method, 
> and phase out the use of passwords

Would we put these statements in https://github.com/w3c-cg/swag/blob/main/docs/security_guidelines.md?
I think if we do, then we want these key messages here, too.

(Maybe something for the SWAG call on Monday to decide)

Elchi3:
Thanks everyone, let's merge this. Follow-up clarifications and refinements are welcome but I think this is quite solid, well researched and well written. Thank you so much @wbamberg!

caugner:
> **Note that the diagrams are broken in the preview page because of [mdn/fred#904](https://github.com/mdn/fred/issues/904).**

This is fixed now via https://github.com/mdn/dex/pull/199:

image

Issue #42331, opened 2025-12-09
Improperly identified scope in HTML solution for Challenge: Planet data table


ssolunna:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Planet_data_table

### What specific section or headline is this issue about?

Example (HTML solution)

### What information was incorrect, unhelpful, or incomplete?

`  <th colspan="2" scope="row">Dwarf planets</th>`

### What did you expect to see?

`  <th colspan="2" scope="rowgroup">Dwarf planets</th>`

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/learn_web_development/core/structuring_content/planet_data_table`
* MDN URL: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Planet_data_table
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/learn_web_development/core/structuring_content/planet_data_table/index.md
* Last commit: https://github.com/mdn/content/commit/4693bfe3aadfe74869eacb5602d147abfd22756b
* Document last modified: 2025-11-25T08:09:32.000Z

</details>

chrisdavidmills:
Thanks for reporting this, @ssolunna. I've created a PR to fix it.

ssolunna:
You're welcome.

PR #42340, opened 2025-12-10
Correct scope for "Dwarf planets" table header

chrisdavidmills:
Also makes a few grammatical fixes.

Fixes https://github.com/mdn/content/issues/42331.


pepelsbey (approved):
Good catch! Thank you :)

PR #42345, opened 2025-12-10
Clarify that multipart Content-Disposition parameters need not be quoted

ky-p23:
Fixes mdn/content#41683
Clarify quoting rules for multipart Content-Disposition parameters


bsmth:
Thanks! This has already been fixed in https://github.com/mdn/content/pull/42065, so I'm closing for now

PR #42346, opened 2025-12-10
chore(cssxref): Update markdown links with `cssxref` in `/web/css`

dipikabh:
### Description

1. This PR replaces the markdown links in `/files/en-us/web/css/` with `{{cssxref}}` macro (the Search and Replace patterns are in the commit messages):

    - Property links: `/en-US/docs/Web/CSS/Reference/Properties/` → `{{cssxref("property")}}`
    - Selector links: `/en-US/docs/Web/CSS/Reference/Selectors/` → `{{cssxref(":selector")}}` and `{{cssxref("::selector")}}`
    - At-rule links: `/en-US/docs/Web/CSS/Reference/At-rules/` → `{{cssxref("@at-rule")}}`
    - Value links: `/en-US/docs/Web/CSS/Reference/Values/` → `{{cssxref("keyword")}}`, `{{cssxref("function()")}}`, and `{{cssxref("data-type")}}`

2. For the already existing macro calls to functions, functional pseudo-classes, and functional pseudo-elements, I've simplified the macro to remove the first parameter because the macro can recognize functions and format them appropriately.

    For example, updated `{{cssxref("calc", "calc()")}}` to `{{cssxref("calc()")}}`.

### Additional details

The updated [cssxref macro](https://github.com/mdn/rari/blob/main/crates/rari-doc/src/templ/templs/links/cssxref.rs) now automatically handles several cases.

- URL generation based on prefix and suffix, then properties, then values:
 
   (from cssxref code:)
   ```comment
    /// - Data types (starting with `<` or `&lt;`): `/Web/CSS/Reference/Values/{slug}`
    /// - Pseudo-classes/elements (starting with `:`): `/Web/CSS/Reference/Selectors/{slug}`
    /// - At-rules (starting with `@`): `/Web/CSS/Reference/At-rules/{slug}`
    /// - Functions (ending with `()`): `/Web/CSS/Reference/Values/{slug}`
    /// - Properties: `/Web/CSS/Reference/Properties/{slug}` (checked first)
    /// - Other values: `/Web/CSS/Reference/Values/{slug}` (fallback)
    /// - If page not found in new structure: fallback to `/Web/CSS/{slug}`
    ```

    So, `{{cssxref(":hover")}}`, `{{cssxref("@media")}}`, `{{cssxref("calc()")}}`, `{{cssxref("unset")}}` and `{{cssxref("image")}}` route to the respective directory within `Web/CSS/Reference`.

- Special mappings:
    
    (from cssxref code:)
    - `"&lt;color&gt;"` | `"<color>"` => `"color_value"`
    - `"&lt;flex&gt;"` | `"<flex>"` => `"flex_value"`
    - `"&lt;overflow&gt;"` | `"<overflow>"` => `"overflow_value"`
    - `"&lt;position&gt;"` | `"<position>"` => `"position_value"`
    - `":host()"` => `":host_function"`
    - `"fit-content()"` => `"fit-content_function"`

    So while the macro call for the `<image>` data type is `{{cssxref("image")}}`, for `<color>` data type it is `{{cssxref("&lt;color&gt;")}}` to disambiguate from the `color` property.

   **Note**: `url_value` and `url_function` are [not yet](https://github.com/mdn/rari/pull/431#issuecomment-3638641088) part of this treatment.

- Auto-formatting based on page type

    The macro takes `page-type` into account for formatting the display text. So, data types like `{{cssxref("length")}}` and keywords like `{{cssxref("unset")}}` get correctly formatted as `<length>` and `unset` based on their page types, even though the location of both resolves to `Web/CSS/Reference/Values`. So additional html encodings (`&lt;` and `&gt;`) for data types is probably not needed. A lot of `cssxref` macro instances for data types do have these encodings, some even with only `&lt;` and `>` (instead of `&gt;`). I've removed the html encodings from some but left the remaining as-is for now; the macro works with or without them.

### Motivation

Our `cssxref`macro was not equipped to handle the directory structure changes we made to the CSS directory [during the reorg](https://github.com/mdn/mdn/issues/662). And so the macro got replaced with markdown links. With macro code now fixed and updated in https://github.com/mdn/rari/pull/382, the links in content can again be replaced with the macro.

- https://github.com/mdn/rari/pull/431

### DIFF
--- a/files/en-us/web/css/guides/grid_layout/index.md
+++ b/files/en-us/web/css/guides/grid_layout/index.md
@@ -101,7 +101,7 @@ This sample animation uses {{cssxref("display")}}, {{cssxref("grid-template-colu
 
 - {{cssxref("repeat()")}}
 - {{cssxref("minmax()")}}
-- {{CSSxRef("fit-content_function", "fit-content()")}}
+- {{cssxref("fit-content()")}}
 
 ### Data types and values
 
@@ -197,7 +197,7 @@ This sample animation uses {{cssxref("display")}}, {{cssxref("grid-template-colu
 - {{cssxref("min-content")}} value
 - {{cssxref("max-content")}} value
 - {{cssxref("fit-content")}} value
-- {{cssxref("fit-content_function", "fit-content()")}} function
+- {{cssxref("fit-content()")}} function
 
 ## Specifications
 
--- a/files/en-us/web/css/reference/values/fit-content/index.md
+++ b/files/en-us/web/css/reference/values/fit-content/index.md
@@ -10,7 +10,7 @@ The `fit-content` sizing keyword represents an element size that adapts to its c
 The keyword ensures that the element is never smaller than its minimum intrinsic size ({{cssxref("min-content")}}) or larger than its maximum intrinsic size ({{cssxref("max-content")}}).
 
 > [!NOTE]
-> This keyword is different from the {{cssxref("fit-content_function", "fit-content()")}} function. The function is used for grid track sizing (for example in {{cssxref("grid-template-columns")}} and {{cssxref("grid-auto-rows")}}) and for laid-out box sizing for properties such as {{cssxref("width")}}, {{cssxref("height")}}, {{cssxref("min-width")}}, and {{cssxref("max-height")}}.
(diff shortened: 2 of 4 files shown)


dipikabh:
Thanks for the review!

PR #42347, opened 2025-12-11
Update warnings about event handler attributes

addsoupbase:
Added the implicit `with` statement wrapping that event handler attributes do, and the issues that come with it

### Description

Mentioned  a (lesser known?) feature of event handler attributes


pepelsbey (approved):
It sounds like good advice! Thank you :)

PR #42348, opened 2025-12-11
Firefox 147 supports media fragments in SVG URLs

chrisdavidmills:
### Description

Firefox 147 adds support for [media fragment](https://www.w3.org/TR/media-frags/) syntax when using SVG files as [image sources](/en-US/docs/Web/SVG/Guides/SVG_as_an_image). See https://bugzilla.mozilla.org/show_bug.cgi?id=1999989.

This PR adds a rel note to cover this addition.

I've not added any other content covering this functionality, because it is a very general feature that otherwise has no support (except for temporal syntax on `<video>`/`<audio>` sources), and to add information about it just for one browser in one specific context seems weird. You could argue that it's the kind of information that belongs in BCD, but we don't really have anywhere to record BCD for a feature like this.

Let me know if you disagree with this. I'd appreciate a second opinion.


pepelsbey:
I would say if we have non-standard Firefox-only [`element()` CSS function documented](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/element), then we should definitely have docs for standard features like these.

chrisdavidmills:
> I would say if we have non-standard Firefox-only [`element()` CSS function documented](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/element), then we should definitely have docs for standard features like these.

@pepelsbey OK, I guess so. But, where to put it? In terms of both content and BCD? It is a type of URL fragment, which leads me to think maybe I could put it next to the text fragment docs, here: https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment.

In terms of BCD, I'm really not sure. It can be used anywhere where an SVG can be used as an image, like ``, `background-image`, `HTMLImageElement.src`/`<canvas>`, various SVG elements, etc. Potentially we could put a subfeature inside every single one, which seems a bit odd to me. WDYT?

chrisdavidmills:
@pepelsbey OK, I've added a page covering media fragment syntax, and linked to it from the rel notes instead of the spec.

Let me know what you think. As you'll see, I've included a hand-written browser compat section, with a few useful notes on what I found when I did some (relatively) extensive testing. I'm not convinced it is worth my while trying to create comprehensive browser compatibility data for this feature, as it would be long-winded and painful ;-/

pepelsbey:
Filed an issue https://github.com/mdn/browser-compat-data/issues/28679

PR #42319, opened 2025-12-09
docs(firefox-release): Finalize release note for FF146 release

dipikabh:
### Description

Ran `node scripts/content/release-firefox.js 146`:

```
🦊 Firefox Release Notes Updater

📋 Planned updates:
   Firefox 145 → Remove active status
   Firefox 146 → Stable
   Firefox 147 → Beta
   Firefox 148 → Nightly (new page)

📡 Fetching release dates...
   Stable release date: December 8, 2025
   Beta release date: January 12, 2026
   Nightly release date: February 23, 2026

🔄 Updating release notes...

✅ Removed active status from Firefox 145
✅ Updated Firefox 146 to (Stable)
✅ Updated Firefox 147 to (Beta)
✅ Created new Nightly page for Firefox 148
```

Updated the dates manually for this run (off by a day) to match the release calendar.
(Updated script coming up in https://github.com/mdn/content/pull/42318)

PR #42372, opened 2025-12-12
chore: Cleanup older Firefox release notes

dipikabh:
### Description

This PR updates the Firefox release note template (follow-up to https://github.com/mdn/content/pull/42319#discussion_r2603150935) and a few previous release note files.

1. In the template file:
   ~~- Added instruction to delete "Removals" if the section is empty.~~
   ~~- Added section-specific deletion instruction, "Delete entire [section] section if no updates", because we remove the parent section for HTTP, SVG, MathML, Security, etc if they're empty. (It'll be good to have this info in the template rather than having to check our convention every time.)~~
   ~~- Added "No notable changes." to "Changes for add-on developers" as well since we always keep this section. (We use this label for the core sections for HTML, CSS, and JavaScript that we don't delete.)~~
   - Reordered sections to follow this sequence:
      - Developer Tools
      - HTML
      - CSS
      - JavaScript
      - HTTP
      - SVG
      - MathML
      - Security
      - APIs
      - WebAssembly
      - WebDriver

~~2. In the latest released (146), beta (147), and nightly (148) files, made updates to reflect changes in the template.~~
3. In the remaining files:
    - Updated casing in "## Changes for Web developers" to "## Changes for web developers"
    - Updated the placement of the "Developer Tools" section to the top (when present)
    - Removed "Developer Tools" section where it was empty (123)

### Motivation

Ordering the technologies/areas in the logical progression, starting with Dev Tools, then core technologies, followed by others

/cc @pepelsbey

### DIFF
--- a/files/en-us/mozilla/firefox/releases/148/index.md
+++ b/files/en-us/mozilla/firefox/releases/148/index.md
@@ -34,8 +34,6 @@ Firefox 148 is the current [Nightly version of Firefox](https://www.firefox.com/
 
 <!-- ### CSS -->
 
-<!-- No notable changes. -->
-
 <!-- #### Removals -->
 
 <!-- ### JavaScript -->
--- a/scripts/content/release-firefox.md
+++ b/scripts/content/release-firefox.md
@@ -34,8 +34,6 @@ Firefox {{VERSION}} is the current [Nightly version of Firefox](https://www.fire
 
 <!-- ### CSS -->
 
-<!-- No notable changes. -->
-
 <!-- #### Removals -->
 
 <!-- ### JavaScript -->


dipikabh:
@pepelsbey I've removed the comments as discussed (they're not needed because the script automatically handles empty sections).

pepelsbey (commented):
I see you reordered MathML and SVG, but it was on purpose to have them right after HTML, since they're also markup languages.

Also, there are still many "remove X if empty" comments left that we don't need, as the release script handles removals automatically.

So, the idea is:

1. All section headings are commented and will be removed automatically if empty, apart from HTML, CSS, and JavaScript.
2. **Only** HTML, CSS, and JavaScript sections have "no notable changes" comments that will be uncommented.

We don't need any other comments or changes to the template, apart from adding the DevTools section.

dipikabh:
> there are still many "remove X if empty" comments

Oops! I removed them from the template file but missed removing them from 147 and 148 relnote files. All gone now 👍 

> I see you reordered MathML and SVG, but it was on purpose to have them right after HTML, since they're also markup languages.

Yes, that makes logical sense and I gathered that would have been the reason.
The sequence HTML → CSS → JavaScript → HTTP → SVG/MathML → APIs makes more sense though because it prioritizes the core technologies more prominently at the top.

In fact, I applied the reordering because I assumed that was the sequence we've used historically.
So I checked what we've done in the past. Of all the relnotes that contain MathML or SVG sections (excluding 146 and 147):
- 7 place them immediately after HTML
- 54 place them after JavaScript or HTTP

(I just resolved some merge conflicts in [147](https://pr42372.review.mdn.allizom.net/en-US/docs/Mozilla/Firefox/Releases/147) and I haven't updated the order in it yet, pending what we decide to do with the sequence.)

pepelsbey:
> 7 place them immediately after HTML
> 54 place them after JavaScript or HTTP

There was a moment when I started doing this for logical grouping. That's why it doesn't go all the way back and there are only 7 in that order.

I would suggest keeping changes to a minimum, given the plans for this section, and focus on improving the template for the future releases.

dipikabh:
> I would suggest keeping changes to a minimum, given the plans for this section, and focus on improving the template for the future releases.

Makes sense. I'll push an update shortly to undo the template changes

PR #42375, opened 2025-12-12
Number federated identity overview diagram

wbamberg:
The overview diagram in https://developer.mozilla.org/en-US/docs/Web/Security/Authentication/Federated_identity would be easier to follow if the steps were numbered, so people could see the order in which the steps happen. This PR adds the numbers.


hamishwillee (approved):
Much better

PR #42379, opened 2025-12-13
Enforce naming convention in `rel` attributes

ramiy:
### Description

Use consistent headings for `rel` attributes.

**Before:**

before

**After:**

after


dipikabh (approved):
Thank you for spotting and fixing!

PR #42412, opened 2025-12-15
Correct calculation

Abhi-Sharma:
Summary

This pull request corrects an incorrect numerical estimation in the Window.requestAnimationFrame() documentation related to request ID overflow timing.

Problem

The warning section under Return value stated that request ID overflow would occur after approximately 500 days when rendering at 60 Hz with 100 requestAnimationFrame() calls per frame.
This value is mathematically incorrect.

Solution

The calculation has been corrected to reflect the actual overflow timeframe of approximately 8 days, based on a 32-bit counter and a call rate of 6,000 requests per second (60 Hz × 100 calls per frame).
An additional clarification was added to emphasize that this scenario represents an extreme and uncommon usage pattern.

Rationale

Accurate numerical information is important for developer understanding, especially for long-running or performance-critical applications such as games, simulations, and dashboards. This change ensures the documentation remains technically correct and avoids misleading assumptions.

Scope of Change

Updated the overflow time estimation in the warning section

Added clarification about the edge-case nature of this scenario

No behavioral or API changes

References

Calculation:

2^32 ÷ (60 × 100) ÷ 60 ÷ 60 ÷ 24 ≈ 8.27 days


hamishwillee:
Closing, because this is already being addressed in #42412

hamishwillee:
But thanks!

Issue #42025, opened 2025-11-19
FedCM API's description of the returned IdentityCredential is inaccurate


wbamberg:
In the FedCM API, the RP calls [`navigator.credentials.get()`](https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/get), and if all goes well gets back an [`IdentityCredential`](https://developer.mozilla.org/en-US/docs/Web/API/IdentityCredential) object.

MDN's documentation for this operation and the returned object indicates that the `IdentityCredential` contains a token signed with the IdP's key, and that the RP needs to verify the signature, and if it's correct, can sign the user in. See e.g. https://developer.mozilla.org/en-US/docs/Web/API/FedCM_API/RP_sign-in:

> The method returns a promise that fulfills with an [IdentityCredential](https://developer.mozilla.org/en-US/docs/Web/API/IdentityCredential) object if the user identity is successfully validated by the chosen IdP. This object contains a token that includes user identity information that has been signed with the IdP's [digital certificate](https://developer.mozilla.org/en-US/docs/Glossary/Digital_certificate).
>
> The RP sends the token to its server to validate the certificate, and on success can use the (now trusted) identity information in the token to sign them into their service (starting a new session), sign them up to their service if they are a new user, etc.

But the FedCM API makes no promises at all about what the returned token contains or what the RP needs to do with it:

> The content of the [token](https://w3c-fedid.github.io/FedCM/#dom-identityassertionresponse-token) is opaque to the user agent and can contain anything that the [IDP](https://w3c-fedid.github.io/FedCM/#idp) would like to pass to the [RP](https://w3c-fedid.github.io/FedCM/#rp) to facilitate the login. For this reason the [RP](https://w3c-fedid.github.io/FedCM/#rp) is expected to be the party responsible for validating the [token](https://w3c-fedid.github.io/FedCM/#dom-identityassertionresponse-token) passed along from the [IDP](https://w3c-fedid.github.io/FedCM/#idp) using the appropriate token validation algorithms defined.

(https://w3c-fedid.github.io/FedCM/#idp-api-id-assertion-endpoint)

In particular, in the OAuth/OIDC profile for FedCM (https://github.com/w3c-fedid/FedCM/issues/599), the token in the `IdentityCredential` is an OAuth "authorization code", which the RP has to send back to the IdP to get the actual signed identity/access token. That is, FedCM only gets you through the first step of the two-step OAuth "authorization code flow" that's recommended for OAuth/OIDC clients.

[...]

Issue #42136, opened 2025-11-26
FedCM docs don't document active mode properly


wbamberg:
So I might be confused here, but.

In FedCM, if an RP calls `navigator.credentials.get()`, and the user is not logged into the IdP, is the user given the opportunity to sign into the IdP, or does `get()` just fail?

It looks to me like the user does get the chance to log in, if the `navigator.credentials.get()` call has `mode` set to `"active"` (and certain other constraints are met). From the spec (https://w3c-fedid.github.io/FedCM/#create-identity-credential, step 7.6):

> If loginStatus is logged-out:
> - If mode is active and mediation is not "silent":
>     - Let result be the result of running fetch the config file and show an IDP login dialog with provider and globalObject.
>     - If result is failure, return (failure, true).
>     - Otherwise, set loginStatus to logged-in.
>  - Otherwise, set providerMap[providerOrigin] to "logged-out" and continue.

This also seems to be addressing e.g. https://github.com/w3c-fedid/active-mode/issues/2, which sounds like my use case. But the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/FedCM_API/RP_sign-in#calling_the_get_method), although they describe mode, don't discuss this possibility:

> If the user has never signed into an IdP or is logged out, the get() method rejects with an error and the RP can direct the user to an IdP page to sign in or create an account.

In fact from the [documentation for active mode](https://developer.mozilla.org/en-US/docs/Web/API/FedCM_API/RP_sign-in#active_versus_passive_mode), it's not obvious what the point in setting `mode` to `"active"` is.

hamishwillee:
@chrisdavidmills Do you know? I think I found this confusing originally and understood at some point - but you might know off the top of your head.

chrisdavidmills:
> [@chrisdavidmills](https://github.com/chrisdavidmills) Do you know? I think I found this confusing originally and understood at some point - but you might know off the top of your head.

From the page Will linked above:

> The default value for mode is passive. If mode is not set, or is set explicitly to passive, the browser can initiate the sign-in flow via a get() call without direct user interaction. For example, you might want to initiate the sign-in flow as soon as the user navigates to the sign-in page, provided they have IdP accounts to sign in with. In this mode, browsers typically present the user with a sign-in dialog window containing all the different sign-in options specified in the providers object, and they can choose whichever one suits them best and then enter the appropriate credentials.
>
[...]

PR #42419, opened 2025-12-16
token structure and usage is not defined by FedCM

wbamberg:
Fixes https://github.com/mdn/content/issues/42025.

### DIFF
--- a/files/en-us/web/api/fedcm_api/rp_sign-in/index.md
+++ b/files/en-us/web/api/fedcm_api/rp_sign-in/index.md
@@ -24,7 +24,7 @@ The FedCM API does not define the structure of the `token` object or what the RP
 
 For example, in the [FedCM for OAuth](https://github.com/aaronpk/oauth-fedcm-profile) profile, which describes how the [OpenID Connect (OIDC)](/en-US/docs/Web/Security/Authentication/Federated_identity#openid_connect) protocol could be implemented using FedCM, the token returned by `CredentialsContainer.get()` is an OAuth authorization code. The RP uses this code to retrieve the identity token from the IdP's token endpoint.
 
-When an RP chooses to work with a particular IdP, they will provide instructions for how to use the returned `token` value.
+When an RP chooses to work with a particular IdP, the IdP will provide instructions for how to use the returned `token` value.
 
 ### Example request
 


files/en-us/web/api/fedcm_api/rp_sign-in/index.md
```diff
@@ -10,16 +10,23 @@ This article describes the process by which a {{glossary("Relying party", "relyi
 
 ## Calling the `get()` method
 
-RPs can call {{domxref("CredentialsContainer.get", "navigator.credentials.get()")}} with an `identity` option to request that a user be given the option to sign in to the RP with a choice of existing IdP accounts (that they are already signed in with on the browser). The IdPs identify the RP by its `clientId`, which was issued by each IdP to the RP in a separate IdP-specific process. The chosen IdP identifies the specific user who is attempting to sign-in with the credentials (cookies) provided to the browser during the [sign-in flow](#fedcm_sign-in_flow).
+RPs can call {{domxref("CredentialsContainer.get", "navigator.credentials.get()")}} with an `identity` option to request that a user be given the option to sign in to the RP with a choice of existing IdP accounts. The IdPs identify the RP by its `clientId`, which was issued by each IdP to the RP in a separate IdP-specific process. The chosen IdP identifies the specific user who is attempting to sign-in with the credentials (cookies) provided to the browser during the [sign-in flow](#fedcm_sign-in_flow).
```
wbamberg, line 13: "(that they are already signed in with on the browser)" is not correct (see https://github.com/mdn/content/issues/42136). I haven't addressed this issue completely in this PR though.

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
```diff
@@ -10,16 +10,23 @@ This article describes the process by which a {{glossary("Relying party", "relyi
...
 
-The method returns a promise that fulfills with an {{domxref("IdentityCredential")}} object if the user identity is successfully validated by the chosen IdP. This object contains a token that includes user identity information that has been signed with the IdP's {{glossary("digital certificate")}}.
+If the user has never signed into an IdP or is logged out, `CredentialsContainer.get()` rejects with an error and the RP can direct the user to an IdP page to sign in or create an account.
```
wbamberg, line 15: So e.g. this bit is still incorrect. I'll address https://github.com/mdn/content/issues/42136 in a separate PR.

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
```diff
@@ -10,16 +10,23 @@ This article describes the process by which a {{glossary("Relying party", "relyi
...
+The `IdentityCredential` includes a `token` property which the RP can use to sign the user in.
+
+The FedCM API does not define the structure of the `token` object or what the RP should do with it: this depends entirely on the federated identity protocol that the IdP implements.
+
+For example, in the [FedCM for OAuth](https://github.com/aaronpk/oauth-fedcm-profile) profile, which describes how the [OpenID Connect (OIDC)](/en-US/docs/Web/Security/Authentication/Federated_identity#openid_connect) protocol could be implemented using FedCM, the token returned by `CredentialsContainer.get()` is an OAuth authorization code. The RP uses this code to retrieve the identity token from the IdP's token endpoint.
+
+When an RP chooses to work with a particular IdP, they will provide instructions for how to use the returned `token` value.
```
wbamberg:
```suggestion
When an RP chooses to work with a particular IdP, the IdP will provide instructions for how to use the returned `token` value.
```

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
hamishwillee: My understanding (remembered, not checked) is that 
- in passive mode the login happens on page load without user interaction. If the user is logged in (checked via Login Status API maybe?) the `get()` request will succeed. If not, it fails and presumably get() returns nothing useful.
- In active mode you explicitly login on button press. If you're not logged in you get a prompt to enter credentials. 

So I "think" the already signed in thing is not quite true - the difference is whether you're auto prompted or not  to sign in. 

I'm OK with this though.

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
hamishwillee: Yes, as per my "rememberence" in https://github.com/mdn/content/pull/42419/files#r2633150183 thi sis corred, but what happens depends on the mode.

hamishwillee (approved):
Added a comment. Could be wrong. IMO this is definitely "more correct" even if not 100% ight, so approving.

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
wbamberg: > So I "think" the already signed in thing is not quite true - the difference is whether you're auto prompted or not to sign in.

Yes, I deleted the parenthetical here because it's not true _as a rule_ (it is only true in passive mode) which I think is what you are saying too.

I think explaining active/passive mode is a bigger thing than I want to deal with in this PR though (I'll file a follow-up for that).

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
wbamberg: Again, AIUI it's incorrect that:

> If the user has never signed into an IdP or is logged out, `CredentialsContainer.get()` rejects with an error and the RP can direct the user to an IdP page to sign in or create an account.

It's correct that:

> If:
> - the user has never signed into an IdP or
> - the user is is logged out **and the request is made in passive mode**
>
> `CredentialsContainer.get()` rejects with an error and the RP can direct the user to an IdP page to sign in or create an account.

...but we need to explain this all properly in a follow-up PR.

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
hamishwillee: You did. I added https://github.com/mdn/content/pull/42419#discussion_r2633150183 mostly because I wanted to clarify in my own mind what I thought about https://github.com/mdn/content/issues/42136

wbamberg: Fair enough. I think we are in agreement :).

files/en-us/web/api/fedcm_api/rp_sign-in/index.md
hamishwillee: For the first case, I thought the _browser_ would pop up a dialog in active mode even if the user is logged out (and possibly if never signed up). 

Sounds like you know I'm wrong. You've done the research and I approved because I trust you more than me on this.

PR #42430, opened 2025-12-16
Add Query to the list of URI components

rivea0:
### Description

Added Query to the list of URI components on the URI index page under the Reference section.

### Motivation

The URI reference page (`/en-US/docs/Web/URI/Reference`) has the list of links to all the URI components using the macro `{{SubpagesWithSummaries}}`.
However, the index page for URI (`/en-US/docs/Web/URI`) is missing Query in the list under the Reference section.


hamishwillee (approved):
Thank you very much.

PR #42432, opened 2025-12-16
Technical review: storage-access activation per origin not site

chrisdavidmills:
### Description

While using the [Storage Access API](https://developer.mozilla.org/en-US/docs/Web/API/Storage_Access_API), when embedded content activates a previously-granted `storage-access` permission via the `requestStorageAccess()` method or the `Activate-Storage-Access` header, 3rd party cookies are now only sent with requests to the calling embed's exact origin.

Previously, 3rd party cookies were sent with requests to the calling embed's site.

This behavioral change has occurred in:

- Chrome 141: See https://chromestatus.com/feature/5169937372676096.
- Firefox 140: See https://wpt.fyi/results/storage-access-api/requestStorageAccess-cross-origin-fetch.sub.https.window.html?run_id=5176618970775552
- Safari is the only browser left that still supports the old model.

This PR adds content to document this change.

@cfredric / @hamishwillee.

### DIFF
--- a/files/en-us/web/api/document/requeststorageaccess/index.md
+++ b/files/en-us/web/api/document/requeststorageaccess/index.md
@@ -22,7 +22,7 @@ The [storage access headers](/en-US/docs/Web/API/Storage_Access_API#storage_acce
 Note that the headers can activate a granted permission for any embedded resource, such as credentialed images, not just code embedded in an {{htmlelement("iframe")}}.
 
 It is also possible to activate a granted permission for a cross-origin, same-site endpoint by calling `requestStorageAccess()` (this time without the requirement for transient activation).
-However this only works to activate permission for embedded code.
+However, this only works to activate permission for embedded code.
 It is also less efficient than using the headers, because the resource needs to be loaded in order to activate the permission.
 
 > [!NOTE]
--- a/files/en-us/web/http/reference/headers/activate-storage-access/index.md
+++ b/files/en-us/web/http/reference/headers/activate-storage-access/index.md
@@ -46,21 +46,21 @@ Activate-Storage-Access: load
 - `retry`
   - : The server uses this token to indicate that it needs its third party cookies in order to properly respond to this request.
 
-    The server should check for `Sec-Fetch-Storage-Access: inactive` in the request before responding with this token, in order to check that the permission has already been granted (but is inactive).
+    The server should check for `Sec-Fetch-Storage-Access: inactive` in the request before responding with this token to check that the permission has already been granted (but is inactive).
     The `allowed-origin` parameter must be specified to allow the specific origin (specify `*` to allow any origin).
 
     The browser should respond by activating an _already-granted_ storage-access permission, and retrying the request with unpartitioned cookies included.
 
 - `load`
-  - : The server uses this token to indicate that it is sending the browser a HTML document that needs to activate a pre-existing `storage-access` permission grant — in order to access unpartiioned cookies while it loads.
+  - : The server uses this token to indicate that it is sending the browser an HTML document that needs to activate a pre-existing `storage-access` permission grant — in order to access unpartitioned cookies while it loads.
 
-    The server should check for `Sec-Fetch-Storage-Access: inactive` or `Sec-Fetch-Storage-Access: active` in the request before responding with `load`, in order to confirm that the permission has already been granted.
+    The server should check for `Sec-Fetch-Storage-Access: inactive` or `Sec-Fetch-Storage-Access: active` in the request before responding with `load` to confirm that the permission has already been granted.
 
     The browser should respond by loading the resource and granting it access to its unpartitioned cookies.
 
 ## Description
 
-The [Storage Access API](/en-US/docs/Web/API/Storage_Access_API#) provides a JavaScript mechanism to allow an embedded resource to request `storage-access` permission.
+The [Storage Access API](/en-US/docs/Web/API/Storage_Access_API) provides a JavaScript mechanism to allow an embedded resource to request `storage-access` permission.
 This enables sending third-party cookies in requests, which would otherwise be blocked by default in most browsers.
 The resource must first be requested without cookies, so the server returns an uncredentialed version of the resource that will not have access to its own cookies.
 After loading, this resource can call {{domxref("Document.requestStorageAccess()")}} with transient activation to request the storage-access permission.
@@ -73,7 +73,6 @@ Similarly, if you load another origin in the same site, the permission will be g
 
 The resource has to be loaded at least once to be granted the storage-access permission.
 However, once granted, a server can use `Activate-Storage-Access` to activate the permission for other origins and contexts.
-Note that it is also possible (but less efficient), to activate a permission by loading a resource and calling `Document.requestStorageAccess()`.
 
 The way this works is that:
 
@@ -84,6 +83,9 @@ The way this works is that:
    Once loaded by the browser, this resource has access to its cookies as though it were a first-party resource.
 
 Responses must also include the {{httpheader("Vary")}} header with `Sec-Fetch-Storage-Access`.
+
+> [!NOTE]
+> It is also possible (but less efficient) to activate a permission by loading a resource and calling `Document.requestStorageAccess()`.
 
 ## Examples
 


files/en-us/web/http/reference/headers/activate-storage-access/index.md
```diff
@@ -59,8 +59,9 @@ If granted by the user, the permission is stored by the browser in a key associa
...
 
 The permission is granted for a particular embedder/embedded site, but only activated for a particular context, such as an `<iframe>` or browser tab.
+
 This means that if you load the same page in a new tab or `<iframe>`, the permission state of that context will be `inactive`; it won't become `active` until the permission is activated.
-The normal storage access flow is to again request the resource without cookies, call `Document.requestStorageAccess()` to activate the existing permission, then reload the resource with cookies.
+The normal storage access flow is to again request the resource without cookies, call `Document.requestStorageAccess()` to activate the existing permission, then reload the resource with cookies. The resource's cross-origin requests follow the [same-origin policy](/en-US/docs/Web/Security/Defenses/Same-origin_policy), therefore third-party cookies are sent only with requests to the resource's exact origin. Other origins within the same site wishing to access third-party cookies will need to activate the storage-access permission separately.
```
cfredric: This is a pre-existing issue, but I wonder if it would be better to mention `Activate-Storage-Access: load` here, instead of recommending to load without cookies, call `document.requestStorageAccess()`, and then refresh the embed? The `Activate-Storage-Access: load` header was designed to solve [this scenario](https://github.com/privacycg/storage-access-headers?tab=readme-ov-file#revisiting-a-previously-allowed-authenticated-embed) with only one network round-trip.

Now that [Firefox](https://groups.google.com/a/mozilla.org/g/dev-platform/c/OPmJiLLZdak/m/bcmz_8g9AgAJ) and Chrome (and Edge, I assume?) have announced support for SAH, the `Activate-Storage-Access: load` flow should work in every browser except Safari.

chrisdavidmills: If you read to the end of the section, it later mentions what the flow looks like using `Activate-Storage-Access` instead of reloading. I think the idea is to talk through what the solution looks like using `requestStorageAccess()`, then talk about the problem associated with this, and then talk about how `Activate-Storage-Access` solves that problem.

However, it could be phrased a bit more clearly, and perhaps we could also make the flow more obvious with some subheadings to introduce each phase of the story.

Also, it would be good if the description clearly and succinctly stated the difference between the `retry` and `load` values, and when they would be used. I must admit that this is not clear to me right now.

cfredric: Oh, I see - I should have kept reading. Yeah, maybe it could be rephrased to talk about the "old" way vs the "new" way or something like that. (The fact that the "`document.requestStorageAccess()` + refresh" flow was described as the "normal" flow threw me off.)

> Also, it would be good if the description clearly and succinctly stated the difference between the retry and load values, and when they would be used. I must admit that this is not clear to me right now.

I realize this means the explainer needs some improvement, but let me try to explain here anyway.

* `retry` is the server's way of indicating "the browser didn't send unpartitioned cookies, but I need them in order to properly respond to this request. The browser should 'retry' the request, but also include unpartitioned cookies this time."
  * Since the browser can only include unpartitioned cookies if the `storage-access` permission has been granted (but not activated yet), the server should check for `Sec-Fetch-Storage-Access: inactive` in the request before responding with `retry`.
* `load` is the server's way of saying "I'm serving the browser some HTML document, and that document needs to access unpartitioned cookies while it loads (e.g. when fetching other subresources). The browser should act as if that document calls `document.requestStorageAccess()` before running any of its scripts or fetching any of its subresources."
  * Since the browser won't show a user prompt while the iframe is loading, the `load` token is only supported if the `storage-access` permission has been granted already. I.e., the server should check for `Sec-Fetch-Storage-Access: inactive` or `Sec-Fetch-Storage-Access: active` in the request before responding with `load`.

Hopefully that's clearer? LMK. (I'm still fine with addressing this separately, since the same-origin change is more straightforward.)

files/en-us/web/api/document/requeststorageaccess/index.md
```diff
@@ -12,6 +12,8 @@ The **`requestStorageAccess()`** method of the {{domxref("Document")}} interface
 
 To check whether permission to access third-party cookies has already been granted, you can call {{domxref("Permissions.query()")}}, specifying the feature name `"storage-access"`.
 
+After an embed has activated `storage-access` permission via `requestStorageAccess()`, it should reload itself. The browser will re-request the resource with third-party cookies included, and make them available to the embedded resource once it has loaded. The embed's cross-origin requests follow the [same-origin policy](/en-US/docs/Web/Security/Defenses/Same-origin_policy), therefore third-party cookies are sent only with requests to the embedded resource's exact origin. Other origins within the same site wishing to access third-party cookies will need to activate the storage-access permission separately.
```
hamishwillee: To my taste.
1. Youl'll note I use line breaks on sentences. MDN does care, but I find it makes reviews much easier in github since the flow of the author thinking is like a list. It also makes the logic of the text easier to follow, and extending text much easier in future.
2. I want to capture what this "activate separately" actually means, and point to the headers too.



```suggestion
After an embed has activated `storage-access` permission via `requestStorageAccess()`, it should reload itself.
The browser will re-request the resource with [same-origin](/en-US/docs/Web/Security/Defenses/Same-origin_policy) third-party cookies included, and make them available to the embedded resource once it has loaded.

Third-party cookies are sent only with requests to the embedded resource's exact origin.
Other origins within the same site wishing to access third-party cookies will need to activate the storage-access permission separately.
While it is possible to do so by loading a page in each origin and calling this method, it is more efficient to use the [storage access headers](/en-US/docs/Web/API/Storage_Access_API#storage_access_headers) to activate an already granted permission.
```

files/en-us/web/api/storage_access_api/index.md
```diff
@@ -97,7 +97,7 @@ Third party content embedded in an {{htmlelement("iframe")}} that needs to acces
...
 
 5. After an embed has activated the storage-access permission it should reload itself.
-   The browser will re-request the resource with third-party cookies included, and make them available to the embedded resource once it has loaded.
+   The browser will re-request the resource with third-party cookies included, and make them available to the embedded resource once it has loaded. The embed's cross-origin requests follow the [same-origin policy](/en-US/docs/Web/Security/Defenses/Same-origin_policy), therefore third-party cookies are sent only with requests to the embedded resource's exact origin. Other origins within the same site wishing to access third-party cookies will need to activate the storage-access permission separately.
```
hamishwillee, line 100: Something like https://github.com/mdn/content/pull/42432/files#r2632895541

files/en-us/web/api/document/requeststorageaccess/index.md
hamishwillee: Actually, I'm going to make a few changes to this myself and you can fix it up if you don't like it :-)

files/en-us/web/http/reference/headers/activate-storage-access/index.md
hamishwillee: @chrisdavidmills @cfredric  Yes, when I added the header docs to the API  (not that long ago) they were new, and I wrote it like "this is normal and this is the new and improved". It was based off the explainer, and I was learning as I went.

For the requestStorageAccess and Activate-... header I have updated the retry/load using info provided by @cfredric and also removed terminology such as "normal" from those docs. @cfredric Can you check those?

@chrisdavidmills If you are happy with those changes we can merge this - or wait until you are back. 

@chrisdavidmills  I have not updated the storage API overview. I think we should rewrite and invert the docs to "This is how you do it using the headers ... and for interest, if you need to do it in older browsers you can do it this way via the pure API." 

I don't have bandwidth for that now - will you in the new year? If NOT add a note here and I'll create an issue and assign to me.

files/en-us/web/api/document/requeststorageaccess/index.md
```diff
@@ -8,11 +8,17 @@ browser-compat: api.Document.requestStorageAccess
...
 
 To check whether permission to access third-party cookies has already been granted, you can call {{domxref("Permissions.query()")}}, specifying the feature name `"storage-access"`.
 
-After an embed has activated `storage-access` permission via `requestStorageAccess()`, it should reload itself. The browser will re-request the resource with third-party cookies included, and make them available to the embedded resource once it has loaded. The embed's cross-origin requests follow the [same-origin policy](/en-US/docs/Web/Security/Defenses/Same-origin_policy), therefore third-party cookies are sent only with requests to the embedded resource's exact origin. Other origins within the same site wishing to access third-party cookies will need to activate the storage-access permission separately.
+After an embed has activated `storage-access` permission via `requestStorageAccess()`, it should reload itself.
+The browser will re-request the resource with [same-origin](/en-US/docs/Web/Security/Defenses/Same-origin_policy) third-party cookies included, and make them available to the embedded resource once it has loaded.
```
cfredric: "same-origin third-party" cookies might be confusing to readers, IMO. Since the navigation is a reload, the destination is same-origin to the source, by assumption, so the "same-origin" part isn't adding anything.

I think it would be clearer to keep the original phrasing ("... with third-party cookies included") or s/third-party/unpartitioned/.

files/en-us/web/api/document/requeststorageaccess/index.md
```diff
@@ -8,11 +8,17 @@ browser-compat: api.Document.requestStorageAccess
...
+
+Third-party cookies are sent only with requests to the embedded resource's exact origin.
+Other origins within the same site that wish to access their third-party cookies will need to _activate_ the granted storage-access permission.
+The [storage access headers](/en-US/docs/Web/API/Storage_Access_API#storage_access_headers) can be used to activate an already granted storage-access permission (you can also activate a granted permission by calling this method, but this is less efficient as the page has to be loaded).
```
cfredric: > (you can also activate a granted permission by calling this method, but this is less efficient as the page has to be loaded)

This is only true if the origin that needs to receive cookies already has an iframe on the page. This is what makes Storage Access Headers valuable: they allow an origin's server to receive requests with cookies, _without_ requiring that origin to have an iframe on the page (with the ability to execute JS).

I would say something like "If the embed needs to send a credentialed request to a cross-origin, same-site endpoint, that endpoint can use Storage Access Headers to activate the `storage-access` permission for that request."

files/en-us/web/http/reference/headers/activate-storage-access/index.md
```diff
@@ -44,10 +44,20 @@ Activate-Storage-Access: load
...
+  - : The server uses this token to indicate that it needs its third party cookies in order to properly respond to this request.
+
+    The server should check for `Sec-Fetch-Storage-Access: inactive` in the request before responding with this token, in order to check that the permission has already been granted (but is inactive).
     The `allowed-origin` parameter must be specified to allow the specific origin (specify `*` to allow any origin).
+
+    The browser should respond by activating an _already-granted_ storage-access permission, and retrying the request with unpartitioned cookies included.
+
 - `load`
-  - : Request that the browser activate the store-access permission for the context and then load the resource.
+  - : The server uses this token to indicate that it is serving the browser a HTML document that has an already granted storage-access permission, and the document needs to access its third party unpartitioned cookies while it loads.
```
cfredric: > that has an already granted

Nit: this phrasing is a little confusing to me -- it's not really that the document has the permission grant already, since the permission grant is managed by the browser itself. It's that the document wants to _use_ the permission grant, which the browser just confirmed has already been granted (via the `Sec-Fetch-Storage-Access: inactive` or `Sec-Fetch-Storage-Access: active` header). Does that make sense?

I'd rephrase as something like "... to indicate that it is serving the browser an HTML document that needs to activate a pre-existing `storage-access` permission grant in order to access unpartitioned cookies while it loads."

files/en-us/web/http/reference/headers/activate-storage-access/index.md
```diff
@@ -44,10 +44,20 @@ Activate-Storage-Access: load
...
+
+    The server should check for `Sec-Fetch-Storage-Access: inactive` or `Sec-Fetch-Storage-Access: active` in the request before responding with `load`, in order to confirm that the permission has already been granted.
+
+    The browser should respond by loading the resource and granting it access to its unpartitioned cookies.
+    If the storage access permission is granted but not activated, it should also first be activated.
```
cfredric: I'd remove this line -- I think it's a little ambiguous whether it's documenting a requirement that the web developer should satisfy, or documenting what the browser's expected behavior is when the `load` header is used. In the former case, the web dev already did that by using the `load` header; and in the latter case, I think the activation part is implied by the "granting it access to its unpartitioned cookies" part of the previous line.

files/en-us/web/api/document/requeststorageaccess/index.md
hamishwillee: Fair enough!
```suggestion
The browser will re-request the resource with third-party unpartitioned cookies included, and make them available to the embedded resource once it has loaded.
```

files/en-us/web/http/reference/headers/activate-storage-access/index.md
hamishwillee: I'm happy to remove this. FWIW my reason for the statement was  that I was thinking that the server might send `load` out of sequence and this would cause the permission to be activated. But thinking about this, if retry was not sent, the server won't have received the request with the cookies it needs to respond properly, so this wouldn't have any benefit.

```suggestion
```

files/en-us/web/http/reference/headers/activate-storage-access/index.md
hamishwillee: Thanks

```suggestion
  - : The server uses this token to indicate that it is sending the browser a HTML document that needs to activate a pre-existing `storage-access` permission grant — in order to access unpartiioned cookies while it loads.
```

files/en-us/web/api/document/requeststorageaccess/index.md
hamishwillee: @cfredric Good point, but I think this is a bit of a leap for most readers - i.e. they won't understand from that suggested text the nuance that the headers allow the permission to be activated for any credentialed resources, not just those in iframes. The explainer covered this pretty well FWIW.

I think we need to spell it out. What about this?:

```suggestion
The [storage access headers](/en-US/docs/Web/API/Storage_Access_API#storage_access_headers) should be used for activating a granted `storage-access` permission.
Note that the headers can activate a granted permission for any embedded resource, such as credentialed images, not just code embedded in an {{htmlelement("iframe")}}.

It is also possible to activate a granted permission for a cross-origin, same-site endpoint by calling `requestStorageAccess()` (this time without the requirement for transient activation).
However this only works to activate permission for embedded code.
It is also less efficient than using the headers, because the resource needs to be loaded in order to activate the permission.
```

cfredric: That suggestion LGTM!

files/en-us/web/http/reference/headers/activate-storage-access/index.md
hamishwillee: @cfredric Thanks for your review and advice.

@chrisdavidmills I'm approving so you can merge if you like, or do needed corrections. 

See my note above - if you don't want to look at corrections to the the API overview now then you can merge this, but please create an issue for the updates and assign to either/both of us. I'm happy to do it, just not within the next month.

chrisdavidmills: Thanks for all the work in my absence, @hamishwillee and @cfredric. I will merge this now, and then create an issue for the other updates.

PR #42433, opened 2025-12-16
removes word "annoying", explains example clearly

blunket:
### Description

While reading this document, I noticed the word "annoying" and an overly-simple explanation of the example, so I improved the text.

### Motivation

The word "annoying" is opinionated and unnecessary. The example text also did not explain the example clearly.

### DIFF
--- a/files/en-us/web/html/how_to/use_data_attributes/index.md
+++ b/files/en-us/web/html/how_to/use_data_attributes/index.md
@@ -73,7 +73,7 @@ threeColumnArticles.forEach((article) => {
 
 ## CSS access
 
-Note that, as data attributes are plain HTML attributes, you can even access them from [CSS](/en-US/docs/Web/CSS). For example to show the parent data on the article you can use [generated content](/en-US/docs/Web/CSS/Reference/Properties/content) in CSS with the [`attr()`](/en-US/docs/Web/CSS/Reference/Values/attr) function:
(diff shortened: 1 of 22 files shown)


files/en-us/web/css/reference/at-rules/@media/prefers-contrast/index.md
```diff
@@ -25,7 +25,7 @@ Various operating systems do support such preferences and user agents are likely
 
 ## Examples
 
-This example has an annoying low contrast by default.
+This example has a box that uses a low-contrast dashed outline by default. When the **`prefers-contrast`** media feature is set to `more`, the outline is solid instead.
```
chrisdavidmills:
```suggestion
This example includes a box with a dashed {{cssxref("outline")}} set on it by default. When the `prefers-contrast: more` media query matches, the applied outline is given a higher contrast `solid` style instead.
```

chrisdavidmills: I would also suggest adding a small `margin` on the box too, so that it is not flush to the edge of the body and therefore the outline is easier to see.

blunket:
Thanks @chrisdavidmills , I like your phrasing better! I've added a small margin to the CSS block as well. I assumed that's enough to update the live example; I've never worked on this repo before so I wasn't sure. Glad to contribute!

chrisdavidmills:
> Thanks @chrisdavidmills , I like your phrasing better! I've added a small margin to the CSS block as well. I assumed that's enough to update the live example; I've never worked on this repo before so I wasn't sure. Glad to contribute!

@blunket yes, that's perfect. Thanks again for the contribution, and I wish you a peaceful holiday season.

PR #42437, opened 2025-12-17
Fix `display` toggle in playground not changing value

Kiwow:
### Description

Changes a `<select>` value in a playground to be read after each change of the `<select>`, correctly reacting to the change.

### Motivation

The [`display: flow-root` playground](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Display/Multi-keyword_syntax#display_block_flow-root_and_display_inline_flow-root) has a `<select>` toggle, which is supposed to change the `display` value of some elements to demonstrate the effects of the `flow-root` value.

Currently, its value is read only once when the playground JS is first executed, which results in `display: block` always getting set on the elements.


chrisdavidmills (approved):
Thanks for the fix, @Kiwow! This certainly seems more like what the demo was supposed to be doing.

However, I also noticed that there seemed to be some additional markup in the rendered output that I don't think is supposed to be there. I have taken the liberty of editing your PR to add a live sample ID to each related code block, and then referencing that in the `{{EmbedLiveSample}}` macro, so that only the exact code for the demo is rendered in that block.

Let me know if you think I've got this correct.

Kiwow:
@chrisdavidmills It took me a bit to understand what's going on there, since I don't have any experience with the macro system, but I think I understand it at least a little now, and the changes look good to me!

There's one more thing I'd like to bring up. When reading through the section again, I got a bit stuck on [this sentence](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Display/Multi-keyword_syntax#display_block_flow-root_and_display_inline_flow-root:~:text=The%20first%20%3Cdiv%3E%20element%20with%20the%20demo%20controls%20is%20hidden%20so%20we%20can%20focus%20on%20the%20elements%20that%20follow%20instead.):

> The first `<div>` element with the demo controls is hidden so we can focus on the elements that follow instead.

I couldn't figure out what this meant, because I took "is hidden" as it being visually hidden in the rendered demo. I wonder if it might be less confusing to

- reword that sentence, or
- omitting it altogether (I feel like this is the better option)

Kiwow:
After checking out the conflicts, I see that the original problem I opened this PR for was fixed in #42590. I've fixed the conflicts to use that PR's version.

Since this PR still contains your changes to the sample ID, I think it's still worth merging @chrisdavidmills

chrisdavidmills:
> Since this PR still contains your changes to the sample ID, I think it's still worth merging @chrisdavidmills

Agreed, thanks for alterting me to this, @Kiwow. And thanks for all the original work you did on this.

Kiwow:
> Agreed, thanks for alterting me to this, @Kiwow. And thanks for all the original work you did on this.

Happy to help!

PR #42438, opened 2025-12-17
chore: improve professional tone in Undefined type description

ZaidQtaish:
### Description

Updated the description of the null type to remove informal phrasing ("make up an excuse for") and replaced it with "aligns with".

### Motivation

The previous wording was colloquial and unnecessarily personified the language. This change improves the professional tone of the documentation.


Josh-Cena:
Hello and thanks for the contribution! I don't really agree that "make an excuse for" is colloquial; it's as formal as it can be. And indeed `typeof null` is a bug, so there's no real "reason" for it, and any explanation we give is an excuse rather than a reason.

PR #42439, opened 2025-12-17
content: update undeclared variable error description to include let and const

ZaidQtaish:
### Description

Updated the explanation of undeclared variable assignments to include let and const alongside var.

### Motivation

The current text only mentions var, implying that only var is used to declare variables. Adding let and const aligns the explanation with modern JavaScript practices and the code examples provided later in the document.


ZaidQtaish:
@Josh-Cena Could you take a look at this please?

Josh-Cena (approved):
Thanks! Because this description is now unbiased towards `var`, I shrank it so it doesn't implicitly try to compare `var x = ...` with `x = ...`. Good catch :+1:

PR #42460, opened 2025-12-19
Fix CSS, HTML, and HTTP landing pages to match sidebar content and order

dipikabh:
### Description

@class pointed out that our [CSS landing page](https://developer.mozilla.org/en-US/docs/Web/CSS#reference), specifically the Reference section, is missing a link to [Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties).

I noticed a few other issues:

- CSS landing page
  - "Guides" and "How to" sections are missing.
  - The [Tools](https://developer.mozilla.org/en-US/docs/Web/CSS#tools_for_css_development) section does not list all the tools we have in the sidebar.
  - The [meta bug](https://developer.mozilla.org/en-US/docs/Web/CSS#meta_bugs) link is not needed on our landing page.

- [CSS guides landing page](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides)
  - The intro states that it lists guides when it actually lists modules.
     I've updated the text and also added how to get to the individual guides.

- [HTML landing page](https://developer.mozilla.org/en-US/docs/Web/HTML)
  - Not all [guides](https://developer.mozilla.org/en-US/docs/Web/HTML#guides) are listed. Some of the guides appear in the "Reference" section.
  - The "How to" section is missing.

- [HTTP landing page](https://developer.mozilla.org/en-US/docs/Web/HTTP)
  - "Reference" section is before "Guides", which doesn't match the order in the sidebar.

### Motivation

To match the content and order of sections on the landing pages with the sidebar for consistency

- https://github.com/mdn/content/pull/42458

### DIFF
--- a/files/en-us/web/css/guides/index.md
+++ b/files/en-us/web/css/guides/index.md
@@ -8,7 +8,7 @@ sidebar: cssref
 
 This page lists CSS modules, which reflect how CSS specifications are structured. Each module landing page provides an overview of the specification, lists the features it provides (such as properties, at-rules, and data types), and shows how they relate to other parts of CSS and the web platform. You can use the module pages as a starting point to navigate through the related guides and dive deeper into reference documentation for specific features.
 
-To access the individual guides within each module, click on a module name below to visit its landing page, or navigate directly to guides on the topic of your interest.
+To access the individual guides within each module, click on a module name from the following list to visit its landing page, or use the sidebar to navigate directly to a topic and discover all related guides.
 
 {{subpageswithsummaries}}
 


files/en-us/web/css/index.md
```diff
@@ -25,36 +25,61 @@ Our [learn web development core modules](/en-US/docs/Learn_web_development/Core)
...
+  - : Add {{cssxref("box-shadow")}} effects to your CSS objects.
+- [Color format converter](/en-US/docs/Web/CSS/Guides/Colors/Color_format_converter)
+  - : Enter or pick a color and copy its corresponding value in any CSS {{cssxref("color_value", "color format")}}.
+- [Color mixer](/en-US/docs/Web/CSS/Guides/Colors/Color_mixer)
+  - : Mix two colors in any color space using the {{cssxref("color_value/color-mix", "color-mix()")}} function and copy the resulting color in any CSS color format.
+- [Shape generator](/en-US/docs/Web/CSS/Guides/Shapes/Shape_generator)
+  - : Define coordinates and syntax for {{cssxref("basic-shape")}} features.
 
-## Tools for CSS development
+You can also use the following resources:
```
hamishwillee, line 65: Maybe  - just because what these share in common is that they aren't part of MDN

```suggestion
You can also use the following external resources:
```

hamishwillee: Meh, not necessary.

dipikabh:
Thanks for reviewing!

PR #42474, opened 2025-12-19
Drop 'read-only' from role

mhochk:
### Description

Removing the (incorrect) claim that the ElementInternals.role property is read-only.

### Motivation

This change avoids confusing readers who may think this property is read-only. It aligns with the spec (already referenced from this page) and the implementation of (at least several) browsers.


sideshowbarker:
Hoch, congrats on landing your first docs change here — welcome aboard 🎉

PR #42479, opened 2025-12-20
docs(learn): remove invalid type attribute from video example

kashish2710:
Removes the invalid `type` attribute from the \<video> element in the
Creepy-crawly splash page solution and updates the instructions to avoid
misleading learners. The `type` attribute should only be used with
\<source> elements, not directly on \<video>.

### DIFF
--- a/files/en-us/learn_web_development/core/structuring_content/splash_page/index.md
+++ b/files/en-us/learn_web_development/core/structuring_content/splash_page/index.md
@@ -362,7 +362,6 @@ In this assessment we are presenting you with a mostly-finished splash page abou
 
 Just below the `<h1>`, add a `<video>` element that embeds our header video into the page. We'd like it to do the following:
 
-- If you provide multiple video sources, specify the media type using <source> elements.
 - Autoplay the video on load (for this to work in at least some browsers, you'll also need to specify that the video should be muted).
 - Loop endlessly rather than playing once.
 - Preload the video content.


kashish2710:
I noticed the MD033/no-inline-html warning from reviewdog regarding the use of <source> in Markdown. To follow the linting rules and avoid inline HTML, I removed the bullet about specifying <source> elements and the media type.
The rest of the instructions remain the same. This ensures the Markdown stays valid while keeping the guidance clear for learners.

wbamberg:
> I noticed the MD033/no-inline-html warning from reviewdog regarding the use of in Markdown. To follow the linting rules and avoid inline HTML, I removed the bullet about specifying elements and the media type.

The reason for this is that you referred to "\<video>" unescaped in Markdown, and because you can embed HTML in Markdown, that gets interpreted as HTML. Options for fixing it, from worse to better:

- escape the opening "\<", like "\\<video"
- put it in code formatting, like "\`\<video>\`"
- use the htmlelement macro, like '{{htmlelement("video")}}'

However, I think in this case it is better to omit the bullet entirely, as you have done, because there's no expectation that the reader will use multiple files (only one is provided).

@tristantheb , thank you for filing this issue. Does the fix look good to you?

files/en-us/learn_web_development/core/structuring_content/splash_page/index.md
```diff
@@ -362,7 +362,7 @@ In this assessment we are presenting you with a mostly-finished splash page abou
 
 Just below the `<h1>`, add a `<video>` element that embeds our header video into the page. We'd like it to do the following:
 
-- Specify the [media type](/en-US/docs/Web/HTTP/Guides/MIME_types) of the video.
+- If you provide multiple video sources, specify the media type using <source> elements.
```
tristantheb, line 365:
```diff
-- If you provide multiple video sources, specify the media type using <source> elements.
+- If you provide multiple video sources, specify the media type using {{HTMLElement("source")}} elements.
```

When you want to add an html element, it's with this macro or with backticks like this

```diff
-- If you provide multiple video sources, specify the media type using <source> elements.
+- If you provide multiple video sources, specify the media type using `<source>` elements.
```

;)

tristantheb:
> @tristantheb , thank you for filing this issue. Does the fix look good to you?

Looks good for me, when it's sent, I'll let the person who reported the issue on Discord know that it's being fixed :D

Thanks 🙏🏻

PR #42481, opened 2025-12-20
Add security examples and new CRC glossary entry

muhammadfahaddev:


wbamberg:
Thanks for your PR!

I don't think these examples are particularly helpful, and don't think we should be adding them to glossary pages. Glossary pages should be short: they should explain what the thing stands for (if the thing is an acronym), briefly describe what it means, and link to more information.

Glossary pages do not document the feature: that's what the extra links do. In these cases the examples are so short that they don't really tell anyone anything, and in some cases they seem to be incorrect (for example, I can't think why you would have a CSP with the same `script-src` and `default-src`).

I think checksum might be worth adding, but why document CRC and not checksum?

muhammadfahaddev:
Thanks for the feedback! I've removed the examples from the glossary pages to keep them concise. I also updated the CRC page to be a simple definition and added a new Checksum glossary page as suggested. Cross-references have been added to relevant security and database pages. I've updated the PR with these changes

wbamberg:
>  I've removed the examples from the glossary pages to keep them concise. 

You haven't, though.

I also may not have been clear here:

> I think checksum might be worth adding, but why document CRC and not checksum?

I really meant "we might want an entry for checksum, but we don't need a separate one for CRC" (IMO).

wbamberg:
Closing due to lack of response.

PR #42487, opened 2025-12-21
Add note on CSP to Compression Dictionary Transport

ryantownsend:
### Description

Adds a note that implementing a standalone compression dictionary on websites that have a Content Security Policy requires that the `connect-src` directive permits the dictionary location.

### Motivation

It's unclear which CSP directive applies to compression dictionaries, I only got an answer by asking @pmeenan so it's worth documenting for others.

### Additional details

N/A

N/A

### DIFF
--- a/files/en-us/web/http/guides/compression_dictionary_transport/index.md
+++ b/files/en-us/web/http/guides/compression_dictionary_transport/index.md
@@ -191,7 +191,7 @@ Compression algorithms are at risk of security attacks, so there are a number of
 Additionally, dictionaries could themselves become tracking vectors so browsers may restrict this feature when cookies are disabled or when other extra privacy protections are enabled.
 
 As with other resources, if a website uses the {{HTTPHeader("Content-Security-Policy")}} header, the compression dictionary must be an allowed source in order for it to be loaded.
-In particular, when loading a [separate dictionary](#separate_dictionary) using [`<link rel="compression-dictionary">`](/en-US/docs/Web/HTML/Reference/Attributes/rel/compression-dictionary), the `connect-src` directive (or  `default-src`, if `connect-src` is not set) must allow the dictionary location.
+In particular, when loading a [separate dictionary](#separate_dictionary) using [`<link rel="compression-dictionary">`](/en-US/docs/Web/HTML/Reference/Attributes/rel/compression-dictionary), the `connect-src` directive (or `default-src`, if `connect-src` is not set) must allow the dictionary location.
 
 ## Specifications
 


files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -155,6 +155,8 @@ Use-As-Dictionary: match="/js/app.*.js"
 
 From here the process is similar to the previous example when a matching resources is requested.
 
+Note: if the website has a [Content-Security-Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP), the `connect-src` directive must permit the location of your dictionary to avoid the request being blocked.
```
hamishwillee: Thanks very much @ryantownsend - definitely a "gotcha".

The default restriction is same-origin by default for dictionary resources, so anyone using a CSP of `self` as a `default-src` is likely to never notice this potential problem.

To my mind CSP broadly fits in the "Security considerations" category of information, so can we move this down and fit it appropriately into the `## Restrictions` section?

I had a brief look into this and I think you are correct, but perhaps not complete. CSP generally applies to controlling what resources can be loaded and depends on the retrieval mechanism. If it's a stand alone fetch request then `connect-src` would be right, with fallback to `default-src`. However you can use any file as a dictionary, and it might be fetched in some other way, such that the initial request is govered by `script-src`.  Perhaps worth capturing some of that?

ryantownsend: Thanks @hamishwillee will update.

As far as I know `<link rel="compression-dictionary" />` won't use anything other than `connect-src`, even if it's a `.js` or `.css` file being loaded in the `href`. Will validate this though... unless @pmeenan already knows the answer to this?

pmeenan: Correct for the link fetch. Not necessarily for ALL dictionary use cases though since the source of a `<srcipt>`, stylesheet, preload, fetch, etc can all be valid dictionaries as well. It's in the stand-alone dictionary paragraph so it should be implied, but it might not hurt to be specific that "in order to fetch stand-alone dictionaries using a link tag, ..."

hamishwillee: Cheers. Note that I'm in and out around Xmas so reviews are likely to be slow.

ryantownsend:
@hamishwillee @pmeenan I've updated the notice and added a shorter copy to the `<link rel=compression-dictionary>` page too. Let me know if you want any further changes or the commits squashing before merging.

files/en-us/web/http/guides/compression_dictionary_transport/index.md
```diff
@@ -192,6 +190,8 @@ Compression algorithms are at risk of security attacks, so there are a number of
 
 Additionally, dictionaries could themselves become tracking vectors so browsers may restrict this feature when cookies are disabled or when other extra privacy protections are enabled.
 
+As with other resources, when loading a [separate (standalone) dictionary](#separate_dictionary) via a [`<link rel="compression-dictionary">`](/en-US/docs/Web/HTML/Reference/Attributes/rel/compression-dictionary) tag, if the website has a {{HTTPHeader("Content-Security-Policy")}} header, the `connect-src` directive (or `default-src`, if `connect-src` is not set) must permit the location of the dictionary resource to avoid the request being blocked.
```
hamishwillee: Thanks @ryantownsend 

This is strictly accurate, but as per https://github.com/mdn/content/pull/42487#discussion_r2641313470 , while it covers a main way of getting a compression dictionary, it doesn't cover the other ways you might get a dictionary. We should at least make it clear that there are other ways - even if we don't want to list them all here.

How about something like:

```suggestion
As with other resources, if the website has a {{HTTPHeader("Content-Security-Policy")}} header, the compression dictionary must be an allowed source in order for it to be loaded.
The directive used to specify allowed sources depends on how the dictionary is fetched.
When loading a [separate (standalone) dictionary](#separate_dictionary) using [`<link rel="compression-dictionary">`](/en-US/docs/Web/HTML/Reference/Attributes/rel/compression-dictionary) (the most common way), the `connect-src` directive (or `default-src`, if `connect-src` is not set) would be used to specify the location.
```

ryantownsend: Personally, this seems needlessly verbose.  

The other ways are not _specifically_ ‘fetch’ing a dictionary, they are just loading a resource that _happens_ to also be a dictionary due to its `Use-As-Dictionary` header.

This should be far more obvious these resources require their appropriate CSP directive sources, otherwise the stylesheet/script/font/etc wouldn’t load. Those situations would typically be more obvious visually/functionally and they typically have their own more obviously-named directives.

My main concern with your suggestion is the middle sentence:

> The directive used to specify allowed sources depends on how the dictionary is fetched. 

Other than my original comment, it doesn’t go on to explain ‘how else’ they might be fetched (and I think it would be even more verbose to do so), so I feel this just opens more questions rather than answering any.

hamishwillee: OK, let's remove that sentence. I don't mind being more verbose to cover the various cases, but you make a good point that this is the most important one.

```suggestion
As with other resources, if a website uses the {{HTTPHeader("Content-Security-Policy")}} header, the compression dictionary must be an allowed source in order for it to be loaded.
In particular, when loading a [separate dictionary](#separate_dictionary) using [`<link rel="compression-dictionary">`](/en-US/docs/Web/HTML/Reference/Attributes/rel/compression-dictionary), the `connect-src` directive (or  `default-src`, if `connect-src` is not set) must allow the dictionary location.
```

hamishwillee (approved):
Thanks very much @ryantownsend

PR #42497, opened 2025-12-22
Clarify fetch() error handling behavior for Response

Akarsh-Kiran-Gowda:
This PR adds a small clarification note explaining that fetch() only
rejects on network errors and that HTTP error responses should be
handled using Response.ok or Response.status.

### DIFF
--- a/files/en-us/web/api/response/index.md
+++ b/files/en-us/web/api/response/index.md
@@ -70,9 +70,6 @@ You can create a new `Response` object using the {{domxref("Response.Response", 
 In our [basic fetch example](https://github.com/mdn/dom-examples/tree/main/fetch/basic-fetch) ([run example live](https://mdn.github.io/dom-examples/fetch/basic-fetch/)) we use a simple `fetch()` call to grab an image and display it in an {{htmlelement("img")}} element.
 The `fetch()` call returns a promise, which resolves to the `Response` object associated with the resource fetch operation.
 
-> [!NOTE]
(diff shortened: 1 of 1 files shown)


Akarsh-Kiran-Gowda:
Thanks for the suggestion! I’ve updated the example to explicitly handle HTTP error cases using `response.ok`, following the pattern used in the fetch() and Using the Fetch API pages. Please let me know if this looks good.

files/en-us/web/api/response/index.md
```diff
@@ -70,15 +70,26 @@ You can create a new `Response` object using the {{domxref("Response.Response",
 In our [basic fetch example](https://github.com/mdn/dom-examples/tree/main/fetch/basic-fetch) ([run example live](https://mdn.github.io/dom-examples/fetch/basic-fetch/)) we use a simple `fetch()` call to grab an image and display it in an {{htmlelement("img")}} element.
 The `fetch()` call returns a promise, which resolves to the `Response` object associated with the resource fetch operation.
 
+> [!NOTE]
+> The `fetch()` promise only rejects on network errors (such as connection failures). HTTP error responses (like 404 or 500) still resolve successfully. You should check {{domxref("Response.ok")}} or {{domxref("Response.status")}} to handle HTTP errors.
+
```
wbamberg: I don't think we need this. It's not quite correct (`fetch()` can reject for [lots of reasons](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#exceptions)), and explaining it properly would be documentation for `fetch()`, not `Response` (which is the subject of this page).

Akarsh-Kiran-Gowda: Thanks for the clarification! I’ve removed the note, since the example now demonstrates the correct handling and the detailed behavior belongs on the fetch() page.

wbamberg (approved):
👍 thank you

PR #42519, opened 2025-12-24
Remove superfluous argument in requestAnimationFrame example.

advay168:
### Description

Removed the timestamp being passed to `animate` in the `performance.now()` example as it is not used.

### Motivation

Makes the example clearer (and possibly more correct).


sideshowbarker (approved):

sideshowbarker:
@advay168 , congrats on landing your first docs change here — welcome aboard 🎉

Issue #42514, opened 2025-12-23
Wrong statement: "A value which is not specified is set to its initial value."


dbplunkett:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Shorthand_properties

### What specific section or headline is this issue about?

#omitting_properties

### What information was incorrect, unhelpful, or incomplete?

It says, "A value which is not specified is set to its initial value." However, there is at least one counter-example: The one-value syntax for `flex` that omits `flex-basis` (e.g. `flex: 1;`) sets `flex-basis` to `0`, but the initial value for `flex-basis` is `auto`.

### What did you expect to see?

It should say something like "A value which is not specified is set to a default value, which may or may not be the property's initial value."

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/css/guides/cascade/shorthand_properties`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Shorthand_properties
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/guides/cascade/shorthand_properties/index.md
* Last commit: https://github.com/mdn/content/commit/ca5d9f9e63b460fc0c9e15ac57d9739e10e4ea0d
* Document last modified: 2025-11-20T07:05:04.000Z

</details>

MUSKANNISHAD:
Hi, I’d like to work on this and propose a small documentation fix aligning the wording with the actual shorthand behavior. I’ll open a PR shortly.

chrisdavidmills:
Fixed by https://github.com/mdn/content/pull/42533.

PR #42533, opened 2025-12-26
Fix misleading statement about omitted values in CSS shorthand properties

MUSKANNISHAD:
This updates a misleading statement about omitted values in CSS shorthand properties.

Omitted values are set to defaults defined by the shorthand, which may differ from the property's initial value.

Refs #42514


MUSKANNISHAD:
Hi maintainers 👋
I’ve addressed the misleading statement about omitted values in CSS shorthand properties and fixed all formatting issues.
CI checks are now passing.
Thanks for taking a look!

chrisdavidmills (approved):
LGTM; thanks, @MUSKANNISHAD.

PR #42534, opened 2025-12-26
Update and clarify Permissions-Policy usage

bban160:
### Description + Motivation

Regarding the [Permissions-Policy: fullscreen directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy/fullscreen) page:

- The "General example" is incomplete. Cross-origin iframes **must** have `allow="fullscreen"` in addition to their origin being allowlisted in the Permissions-Policy HTTP response header.

- The "With an \<iframe> element" example is invalid. The `Permissions-Policy: fullscreen=(self)` HTTP response header prevents other origins from using fullscreen. The iframe `allow` attribute cannot grant fullscreen access because that would violate the [_most restrictive subset_ inheritance strategy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Permissions_Policy#inheritance_of_policies_for_embedded_content).

After correcting the above, it seemed appropriate to recategorize which one is the "general example" (more widely applicable) and which is the secondary example (more niche).

### Additional details

I tested both Chromium and Firefox to confirm the updated examples are correct (aside from Firefox ignoring the `Permissions-Policy` header due to lack of support).

### DIFF
--- a/files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
+++ b/files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
@@ -49,11 +49,11 @@ SecureCorp Inc. must also include an {{HTMLElement('iframe','allow','#Attributes
 ```
 
 > [!NOTE]
-> Specifying the `Permissions-Policy` header in this manner disallows `bluetooth` for other origins, even if they were allowed by the `<iframe>` `allow` attribute.
+> Specifying the `Permissions-Policy` header in this manner disallows `bluetooth` for other origins, even if they are allowed by the `<iframe>` `allow` attribute.
 
 ### Using the default policy
 
-Without delivering an HTTP response header defining a Permissions Policy for `bluetooth`, user agents will apply the default allowlist `self`. In this mode, `bluetooth` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
+If an allowlist for `bluetooth` is not defined by a `Permissions-Policy` response header, user agents will apply the default allowlist `self`. In this mode, `bluetooth` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
 
 To allow `bluetooth` in a cross-origin iframe, include an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the `<iframe>` element:
 
--- a/files/en-us/web/http/reference/headers/permissions-policy/fullscreen/index.md
+++ b/files/en-us/web/http/reference/headers/permissions-policy/fullscreen/index.md
@@ -48,11 +48,11 @@ SecureCorp Inc. must also include an {{HTMLElement('iframe','allow','#Attributes
 ```
 
 > [!NOTE]
-> Specifying the `Permissions-Policy` header in this manner disallows `fullscreen` for other origins, even if they were allowed by the `<iframe>` `allow` attribute.
+> Specifying the `Permissions-Policy` header in this manner disallows `fullscreen` for other origins, even if they are allowed by the `<iframe>` `allow` attribute.
 
 ### Using the default policy
 
-Without delivering an HTTP response header defining a Permissions Policy for `fullscreen`, user agents will apply the default allowlist `self`. In this mode, `fullscreen` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
+If an allowlist for `fullscreen` is not defined by a `Permissions-Policy` response header, user agents will apply the default allowlist `self`. In this mode, `fullscreen` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
 
 To allow `fullscreen` in a cross-origin iframe, include an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the `<iframe>` element:
 
--- a/files/en-us/web/http/reference/headers/permissions-policy/geolocation/index.md
+++ b/files/en-us/web/http/reference/headers/permissions-policy/geolocation/index.md
@@ -51,11 +51,11 @@ SecureCorp Inc. must also include an {{HTMLElement('iframe','allow','#Attributes
 ```
 
 > [!NOTE]
-> Specifying the `Permissions-Policy` header in this manner disallows `geolocation` for other origins, even if they were allowed by the `<iframe>` `allow` attribute.
+> Specifying the `Permissions-Policy` header in this manner disallows `geolocation` for other origins, even if they are allowed by the `<iframe>` `allow` attribute.
 
 ### Using the default policy
 
-Without delivering an HTTP response header defining a Permissions Policy for `geolocation`, user agents will apply the default allowlist `self`. In this mode, `geolocation` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
+If an allowlist for `geolocation` is not defined by a `Permissions-Policy` response header, user agents will apply the default allowlist `self`. In this mode, `geolocation` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
 
 To allow `geolocation` in a cross-origin iframe, include an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the `<iframe>` element:
 


hamishwillee:
@bban160 You're right in all your main points:
- You need to set the allowlist on a cross origin frame to allow it to use the feature, whether or not you have sent the header
- If the header is not sent the default allowlist is `*`, which means that you can just set the allow list on the frame if the header is not sent.

Two points thought:
- This document is about the header, and it grates to see the "general example" that does not show the use of the header.
- I'm not sure that this captures both those points above.

So a suggestion. 
- Make your heading "With a Permissions-Policy header" into "Basic usage" - that explains the main use.
- Move "general example" below the first example and rename to something like "Cross-origin access without the header" and then explain that second point above.

Does that make sense?

bban160:
> If the header is not sent the default allowlist is `*`, which means that you can just set the allow list on the frame if the header is not sent.

This is not actually correct. I was initially confused about how Permissions Policy propagates as well, but I now believe I understand it correctly after reading numerous specs over the past week and testing in Chromium and Firefox:

- If the default allowlist is `*`, cross-origin frames automatically have access to the feature (unless it is explicitly restricted by the header or `allow` attribute).
- If the default allowlist is `self`, cross-origin frames must be _explicitly opted in_ to use the feature by the `allow` attribute.
  - If a `Permissions-Policy` header sent with the top-level document includes the feature in question, the allowlist in that header must _also_ permit the separate origin to use the feature.

This is very unintuitive, and I do not believe it is documented in plain English on MDN or any other site at the moment. I would like to contribute this to MDN in the future, though I do not mind if someone else gets to it first.

That tangent aside, the part that relates to this PR is that `fullscreen` has a default allowlist of `self`, not `*`.

> This document is about the header, and it grates to see the "general example" that does not show the use of the header.

You are right. I should have looked more closely at the framing of the article.

This PR is not the proper venue for this discussion, but I believe the ideal solution would be to rewrite most of the pages concerning Permissions Policy to better reflect the actual standard. Which is to say, the HTTP response header and iframe `allow` attribute are both equally relevant and equally important. The pages documenting individual directives like `fullscreen` should discuss both ways of controlling Permissions Policy without prioritizing one over the other. The browser compatibility data should also reflect that Permissions Policy directives are supported in Firefox and Safari in the iframe `allow` attribute, even though the HTTP header is only supported on Chromium.

---

With all that in mind, would you like me to still make the changes you suggested?

My main goal with this PR is to be a (relatively) simple fix of the incorrect/misleading information. I noticed a few other pages have the exact same issue and plan to use this PR as a template to fix those, but not until receiving feedback on this first one.

The broader overhaul is something I would like to do eventually but do not have the bandwidth to take on right now. It will also take time to complete and probably deserves more discussion of its own.

hamishwillee:
Arrggg. After reading the spec yesterday I had come to the conclusion there was a different default for if the header wasn't specified (at all) vs the usual default allowlist that would apply if the header was sent but the specific feature was not set. Reading today _I have no idea how I came to that conclusion_.

I think your summary matches the spec here: https://w3c.github.io/webappsec-permissions-policy/#default-allowlists - putting it down below to summarise my thoughts:

- `*` allows the feature in all origins for a document and for any nested contexts. 
- `self` allows the feature in the document and same-origin nested contexts, but not cross origin contexts.
- You can set the container policy using [`allow`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#allow), but only to grant access to a feature that is already allowed by the parent node 
  - see  https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-allow
    > Neither [allow](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-allow) nor [allowfullscreen](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-allowfullscreen) can grant access to a feature in an [iframe](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element) element's [content navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#content-navigable) if the element's [node document](https://dom.spec.whatwg.org/#concept-node-document) is not already allowed to use that feature.
- What this means is that
  -  if the default for a feature is `*` and no header is set you can always just use the feature in an iframe. You can also make the access more specific, for example setting `allow = "fullscreen exmaple.com"`.
  - if the default is `self` and no header is set then for a cross-origin frame you can't add an `allow`  to grant the feature to the frame.
  - if the header **has** been set, then that defines what the parent node can access. My understanding is that `allow` can only grant nested contexts the same features the parent has - so if the parent featre was restricted to access to example.com, then you couldn't use allow to grant the feature to a nested frame that wasn't example.com.
- For `fullscreen` the default is `self`. That means you _must_ have the header set to allow a cross origin iframe to be granted `fullscreen`. 

1. So the `allow` is not equally important, because it can only permit access to a feature already granted to the parent.
2. The iframe example we have now should not work. Because the default is `self`, without the header you can't grant the crsoss origin frame access to the feature. 
3. So I think we should omit that, and just have the "Basic usage" example, with an explanation.

Hope that makes sense, and I haven't confused both of us again.

bban160:
> if the default is `self` and no header is set then for a cross-origin frame you can't add an `allow` to grant the feature to the frame.

This is not correct. If the default is `self` and no header is set, then the feature can be enabled in the cross-origin frame **if specified in the iframe `allow` attribute**. In other words, the feature is already available in the top-level document, which is allowed to "delegate" access to the nested document.

> For `fullscreen` the default is `self`. That means you _must_ have the header set to allow a cross origin iframe to be granted `fullscreen`.

See above. If it worked this way, it would be an untenable situation in terms of website breakage. Before Permissions Policy existed, you only needed the `allowfullscreen` attribute on an iframe. Now, you can use `allowfullscreen` or `allow="fullscreen *"` or `allow="fullscreen"` (and `allowfullscreen` has been redefined as `allow="fullscreen *"` [as per my other PR](https://github.com/mdn/content/pull/42542)). No Permissions-Policy header is needed to enable this behavior. Also note that Firefox and Safari do not support the header, but they do support `allow` the same way as Chromium behaves when no header is sent.

> Hope that makes sense, and I haven't confused both of us again.

No worries, I fully sympathize with the unintuitiveness. If it helps, you can run [`document.featurePolicy.allowsFeature("fullscreen")`](https://developer.mozilla.org/en-US/docs/Web/API/FeaturePolicy/allowsFeature) (returns a boolean) in both the top-level and iframe contexts to experiment (Firefox requires `dom.security.featurePolicy.webidl.enabled` to expose this interface).

hamishwillee:
Thanks very much @bban160 

I'd been misreading the spec to mean that the parent must allow the feature for the origin of the child in order for the feature to be delegated. Once I understand that the parent can delgate the feature if it has it itself, this does indeed make sense - also hanks for the test method, which I used to verify all this.

But I am still confused. Looking at the example as it is now, we have:

```http
Permissions-Policy: fullscreen=(self "https://example.com")
```

From our conversation above, if `sefl` is specified you should not need to specify "https://example.com" to delegate the feature to a child with "https://example.com" - because the parent has the feature.
Further, if `self` (or `*` is not specified you can't delegate to `https://example.com` even same-origin because the parent does not have the feature.

Also in the this case you still need to specify `allow="fullscreen"` to delegate the feature, so it isn't some matter of allowing delegation automatically.

In other words, what's the point of specifying anything other than * or self on the parent?

hamishwillee:
@miketaylr You might want to be involved in this discussion. The assertion here is that:
-  parent must have a feature in order to delegate it to a nested context such as an iframe.
- When using default allow lists (i.e. no header or feature not specified).
  - If the parent has `*` for the feature then the feature is available in all nested contexts by default. 
  - If the parent has `self` for a feature then it is available in same-origin contexts by default but is only allowed in cross-origin contexts if explicitly delegated (such as using `allow="featurename"`)
- When you explicitly specify a feature using the header, the behaviour is slightly different
  - Specifically a policy of `self` only allows the feature for the parent and same-origin resources. For cross-origin resources the policy must include the origin of the cross origin resources to be allowed AND the `allow` attribute must still be specified.
  - `*` behaves the same way as previously.

bban160:
I will circle back to this PR after finding a better way to illustrate the rules (probably a flowchart). Other MDN readers will benefit from this as well.

bban160:
> The purpose of specifying origins in the header isn't clear to me though (other than the parent origin)- because if the parent doesn't have the feature it can't delegate. So you can't just specify other origins you want to use the feature with, because the parent won't have the feature and the child can't get it. But if the parent does have the feature via `self` then what's th point of listing another origin?

So, if you have a header `fullscreen=(self)`, that _prevents_ other origins from using fullscreen. It cannot be delegated.

If you have no header at all, fullscreen can be delegated to frames from any origin.

If you have a header `fullscreen=(self "https://example.com")`, fullscreen can be delegated to frames from https://example.com but not to any other cross-origin frames.

bban160:
> In other words, what's the point of specifying anything other than * or self on the parent?

My usecase for this is probably quite underrepresented on MDN. I host a lot of software, mostly web apps, written by other people. It's very easy for me to control HTTP headers in my reverse proxy, but modifying the HTML itself can be a significant maintenance burden. So it is important in my usecase to implement security hardening via headers, while modifying HTML is only done out of necessity.

You can see specific examples of my methodology on [my website](https://v1.colloid.foo/64e8ea5f-fc49-4ce0-8a6d-9c92ddafbfa9/1c5642b9-393b-4f67-8b9d-09be9464163b/) (sorry for the shameless plug, but it really is relevant here). A notable example would be [autofill](https://v1.colloid.foo/64e8ea5f-fc49-4ce0-8a6d-9c92ddafbfa9/1c5642b9-393b-4f67-8b9d-09be9464163b/#autofill), where I do not want to risk user PII being automatically delivered to a third-party frame without _my_ explicit approval (remember: the HTML may not be written by me). I also use Content-Security-Policy to limit third-party frames, but that is beyond the scope of this discussion.

hamishwillee:
@bban160 Ah, my test case on this was broken.

1. Anyway, yes, we do need to update the main docs, and if you want to take a shot at that, it would be very welcome.

2. For this example, 
   - I still think first having heading "Basic usage"  which uses your "With a Permissions-Policy header" example would be best. 
   - For the second example I think we should consider calling it something like "Using default permissions", and again this can be mostly the content in the current "General example".  

We do probably need a few more words of explanation in each case - i.e. in the second case, note that `self` is the default, and so the feature is granted automatically to same origin but needs to be explicitly delegated for cross origin frames. Also note that this differs from when explicitly setting the policy for a a feature, where you must specify the origins.

bban160:
@hamishwillee Sure, I think we are on the same page now. I will finish this PR and some other small ones first, then tackle the main docs which will be a much larger PR.

Since a few other pages are in the exact same situation as this page and need the exact same changes, should I integrate them into this PR? Or do a separate PR for each one?

hamishwillee:
> Since a few other pages are in the exact same situation as this page and need the exact same changes, should I integrate them into this PR? Or do a separate PR for each one?

After you've done the main docs you may wish to cross link here, or reduce what you say here because it is all said elsewhere. For that reason I'd be tempted to do this first, then the main docs, then another new PR which does all the rest.

That's just a suggestion though. If you want to do all of them now in one PR I won't be complaining - but it is likely you won't get this all reviewed for a bit in that case, as I'm on holiday for 2 weeks from Friday.

bban160:
Yes, the plan is to modify quite a few pages alongside the main docs update to reduce redundancy.

Since we are on the same page now with this PR, I'll go ahead and integrate the remaining pages with the same changeset (they are so identical that reviewing one means all of them have been reviewed).

hamishwillee:
> Since we are on the same page now with this PR, I'll go ahead and integrate the remaining pages with the same changeset (they are so identical that reviewing one means all of them have been reviewed).

Well yes, but more churn since they will all have to be reviewed twice if there is modification after the main docs are updated.

bban160:
I understand it will be more churn, but revising the main docs will take a lot longer (I have some unanswered questions I still need to research myself).

Right now I am more concerned with _fixing the incorrect_ and doing so quickly. The larger changes will follow once I finish researching and think about the best way to structure the docs.

It's not ideal, but that's the best I can offer right now.

bban160:
@hamishwillee Ready for review.

bluetooth, fullscreen, and geolocation are updated as we discussed.

For compute-pressure, it's just a fix of the header syntax. It could have been a separate PR, but it's a tiny change and fits under this PR anyway.

hamishwillee:
@chrisdavidmills I'm on holiday for two weeks. If you happen to have bandwidth for this then please do review. Otherwise I'll look when I return. @bban160 has a better understanding of this than me, though finally I do get it.

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
```diff
@@ -30,36 +30,35 @@ Permissions-Policy: bluetooth=<allowlist>;
 
 ## Default policy
 
-The default allowlist for `bluetooth` is `self`.
+The default allowlist for `bluetooth` is `self`. The top-level browsing context and same-origin iframes are allowed access to the `bluetooth` feature by default.
```
chrisdavidmills, line 33:
```suggestion
The default allowlist for `bluetooth` is `self`. The top-level browsing context and same-origin `<iframe>` embeds are allowed access to the `bluetooth` feature by default.
```

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
```diff
@@ -30,36 +30,35 @@ Permissions-Policy: bluetooth=<allowlist>;
...
 
 ```http
 Permissions-Policy: bluetooth=(self "https://example.com")
 ```
 
-### With an \<iframe> element
-
-FastCorp Inc. wants to disable `bluetooth` for all cross-origin child frames, except for a specific `<iframe>`.
-It can do so by delivering the following HTTP response header to define a Permissions Policy:
+Then include an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the `<iframe>` element:
```
chrisdavidmills:
```suggestion
If we then want to embed content from `https://example.com`, we also need to include an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the embedding `<iframe>` element:
```

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
```diff
@@ -30,36 +30,35 @@ Permissions-Policy: bluetooth=<allowlist>;
...
 
-```http
-Permissions-Policy: bluetooth=(self https://other.com/blue)
+```html
+<iframe src="https://example.com/blue" allow="bluetooth"></iframe>
 ```
 
-Then include an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the `<iframe>` element:
+> [!NOTE]
+> Specifying the `Permissions-Policy` header in this manner disallows `bluetooth` for other origins, even if they would be allowed by the `<iframe>` `allow` attribute.
```
chrisdavidmills:
```suggestion
> Specifying the `Permissions-Policy` header in this manner disallows `bluetooth` for other origins, even if they were allowed by the `<iframe>` `allow` attribute.
```

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
```diff
@@ -30,36 +30,35 @@ Permissions-Policy: bluetooth=<allowlist>;
...
+
+### Using the default policy
+
+FastCorp Inc. wants to allow `bluetooth` in a specific cross-origin child frame. Since the default allowlist for `bluetooth` is `self`, it can do so by including an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the `<iframe>` element:
```
chrisdavidmills: This explanation requires some expansion, imo. When you say "specific cross-origin child frame", do you mean an `<iframe>` included inside the content embedded in the `<iframe>` from the previous example? Or is this a standalone `<iframe>` included at the same level? If this example is building on the previous one, and not a stand-alone example (which I suspect it is), you need to be clear about that. Also, in relation to this, you should expand the explanation beginning "Since the default allowlist for `bluetooth` is `self`..." to say why this allows the child iframe to be handled like this, based on the policy set in the previous example...if indeed it relates to that.

files/en-us/web/http/reference/headers/permissions-policy/fullscreen/index.md
```diff
@@ -31,34 +29,35 @@ Permissions-Policy: fullscreen=<allowlist>;
 
 ## Default policy
 
-The default allowlist for `fullscreen` is `self`.
+The default allowlist for `fullscreen` is `self`. The top-level browsing context and same-origin iframes are allowed access to the `fullscreen` feature by default.
```
chrisdavidmills, line 32:
```suggestion
The default allowlist for `fullscreen` is `self`. The top-level browsing context and same-origin `<iframe>` embeds are allowed access to the `fullscreen` feature by default.
```

files/en-us/web/http/reference/headers/permissions-policy/fullscreen/index.md
```diff
@@ -31,34 +29,35 @@ Permissions-Policy: fullscreen=<allowlist>;
...
 
 ## Examples
 
-### General example
+### Basic usage
```
chrisdavidmills, line 36: I see this has the same structure as the `bluetooth` page examples. In that case, from here onwards, the same comments apply to this page too.

files/en-us/web/http/reference/headers/permissions-policy/geolocation/index.md
```diff
@@ -36,40 +32,35 @@ Permissions-Policy: geolocation=<allowlist>;
 
 ## Default policy
 
-The default allowlist for `geolocation` is `self`.
+The default allowlist for `geolocation` is `self`. The top-level browsing context and same-origin iframes are allowed access to the `geolocation` feature by default.
```
chrisdavidmills, line 35:
```suggestion
The default allowlist for `geolocation` is `self`. The top-level browsing context and same-origin `<iframe>` embeds are allowed access to the `geolocation` feature by default.
```

files/en-us/web/http/reference/headers/permissions-policy/geolocation/index.md
```diff
@@ -36,40 +32,35 @@ Permissions-Policy: geolocation=<allowlist>;
...
 
 ## Examples
 
-### General example
+### Basic usage
```
chrisdavidmills, line 39: The same comments apply to this page too.

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
bban160: I thought about being more verbose here but ultimately decided against it, since both [existing writing on MDN](https://github.com/mdn/content/blob/26cfe9e86b714bdce65c58089f0496206b3436f3/files/en-us/web/http/reference/headers/permissions-policy/fullscreen/index.md?plain=1#L60) and [the specification itself](https://www.w3.org/TR/2025/WD-permissions-policy-1-20251006/#example-1db1f085) set a precedent for being less verbose here.

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
bban160: Both the HTTP header and iframe `allow` attribute must be used simultaneously for this example (see prior discussion).

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
bban160: It is a standalone example, not building on the previous one.

Originally I had them in the opposite order, but @hamishwillee suggested reversing the order (https://github.com/mdn/content/pull/42534#issuecomment-3708670492). Since these pages fall under the Permissions-Policy header rather than the Permissions Policy mechanism as a whole (which includes the iframe `allow` attribute), I agree with keeping this order for now, but I may propose reversing them again when I return for the larger docs update which will include moving pages around (IMO, the second example is much more valuable in general, it's just slightly out of place here given the page is about the header alone).

I'm open to suggestions on how to make it more clear that the two examples are independent. The original author reused the examples from the spec almost verbatim — where they are presented inside their own boxes: https://www.w3.org/TR/2025/WD-permissions-policy-1-20251006/#example-1db1f085

> Also, in relation to this, you should expand the explanation beginning "Since the default allowlist for `bluetooth` is `self`..." to say why this allows the child iframe to be handled like this, based on the policy set in the previous example...if indeed it relates to that.

The header indicated in the previous example replaces the default policy. The default policy is the one compiled into browsers that takes effect in the absence of configuration by HTTP header.

As for why the default policy works differently from a declared policy (HTTP header)… the reason is most likely to minimize breakage. But explaining this in detail will make the example quite long-winded and introduce a lot of unnecessary redundancy, so I am putting off the explanation of the algorithm for a future PR (see https://github.com/mdn/content/pull/42534#issuecomment-3709142444 and https://github.com/mdn/content/pull/42534#issuecomment-3708569823). This PR is merely meant to be a stepping stone so I can start fixing issues on a more manageable scale.

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
chrisdavidmills: This is more of a style/preference thing. I prefer codifying HTML element mentions, but others don't. Since the current style is used elsewhere in this part of the docs, I'm happy for you to ignore this comment.

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
chrisdavidmills: I got that; this suggestion was designed to make that fact a bit clearer.

chrisdavidmills: Having looked at the example again, the description still seems a bit terse and brief to me. I think it would also help if you updated the first paragraph to specify that the usage of the `https://example.com` origin is going to be limited to embeds (I know this is implied, but it still came as a surprise later on).

Maybe update "...within all browsing contexts except for its own origin and those whose origin is `https://example.com`." to something like "...within all browsing contexts except for its own origin and embedded pages whose origin is `https://example.com`."?

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
bban160: I understand. I also usually prefer to be more explicit like you, but in this case I think the docs will be too long-winded if I try to maintain that style consistently.

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
bban160: I see what you mean and will revise this. The feedback is appreciated :+1:

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
chrisdavidmills: Thanks for the explanation. So, is the point of this example that we are allowing `bluetooth` in a specific cross-origin `<iframe>`, _in addition_ to its own origin, or that we are allowing `bluetooth` _only_ in a specific cross-origin `<iframe>`, and not in its own origin?

This is currently not clear, and could do with clarification. I don't think it would require a lot of extra text to do so.

bban160: Apologies for the significant rambling in my previous comment. I wrote it in somewhat of a rush and didn't spend enough time rereading it before posting…

bban160: > Thanks for the explanation. So, is the point of this example that we are allowing `bluetooth` in a specific cross-origin `<iframe>`, _in addition_ to its own origin, or that we are allowing `bluetooth` _only_ in a specific cross-origin `<iframe>`, and not in its own origin?
> 
> This is currently not clear, and could do with clarification. I don't think it would require a lot of extra text to do so.

`bluetooth` does remain allowed for the top-level context. I will make sure to clarify this.

bban160:
@chrisdavidmills Should I split out the `compute-pressure` change to its own PR so it does not get held up by this one? It seems like we have a few more rounds of review before the rest of this PR is finalized.

chrisdavidmills:
> @chrisdavidmills Should I split out the `compute-pressure` change to its own PR so it does not get held up by this one? It seems like we have a few more rounds of review before the rest of this PR is finalized.

You can do it if you want, but I don't mind either way. I don't think it will take us more than a few days to resolve this review.

bban160:
@chrisdavidmills The compute-pressure change now lives in #42793 since I haven't had time to work on this yet. That PR however got assigned to hamishwillee, so I guess it might not be merged any sooner unless you want to take over that PR too.

chrisdavidmills:
> @chrisdavidmills The compute-pressure change now lives in #42793 since I haven't had time to work on this yet. That PR however got assigned to hamishwillee, so I guess it might not be merged any sooner unless you want to take over that PR too.

I've taken it over ;-)

hamishwillee:
> I've taken it over ;-)

Thanks very much! Looks like you two are having fun :-)

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
```diff
@@ -30,36 +30,37 @@ Permissions-Policy: bluetooth=<allowlist>;
...
-
-```http
-Permissions-Policy: bluetooth=(self https://other.com/blue)
+```html
+<iframe src="https://example.com/blue" allow="bluetooth"></iframe>
 ```
 
-Then include an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the `<iframe>` element:
+> [!NOTE]
+> Specifying the `Permissions-Policy` header in this manner disallows `bluetooth` for other origins, even if they were allowed by the `<iframe>` `allow` attribute.
```
chrisdavidmills:
```suggestion
> Specifying the `Permissions-Policy` header in this manner disallows `bluetooth` for other origins, even if they are allowed by the `<iframe>` `allow` attribute.
```

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
```diff
@@ -30,36 +30,37 @@ Permissions-Policy: bluetooth=<allowlist>;
...
+
+### Using the default policy
+
+Without delivering an HTTP response header defining a Permissions Policy for `bluetooth`, user agents will apply the default allowlist `self`. In this mode, `bluetooth` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
```
chrisdavidmills:
```suggestion
If a `Permissions-Policy` response header is not sent to define a Permissions Policy for `bluetooth`, user agents will apply the default allowlist `self`. In this mode, `bluetooth` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
```

files/en-us/web/http/reference/headers/permissions-policy/fullscreen/index.md
```diff
@@ -31,34 +29,37 @@ Permissions-Policy: fullscreen=<allowlist>;
...
 
-```http
-Permissions-Policy: fullscreen=(self)
+```html
+<iframe src="https://example.com/presentation" allow="fullscreen"></iframe>
 ```
 
-Then include an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the `<iframe>` element:
+> [!NOTE]
+> Specifying the `Permissions-Policy` header in this manner disallows `fullscreen` for other origins, even if they were allowed by the `<iframe>` `allow` attribute.
```
chrisdavidmills:
```suggestion
> Specifying the `Permissions-Policy` header in this manner disallows `fullscreen` for other origins, even if they are allowed by the `<iframe>` `allow` attribute.
```

files/en-us/web/http/reference/headers/permissions-policy/fullscreen/index.md
```diff
@@ -31,34 +29,37 @@ Permissions-Policy: fullscreen=<allowlist>;
...
+
+### Using the default policy
+
+Without delivering an HTTP response header defining a Permissions Policy for `fullscreen`, user agents will apply the default allowlist `self`. In this mode, `fullscreen` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
```
chrisdavidmills:
```suggestion
If a `Permissions-Policy` response header is not sent to define a Permissions Policy for `fullscreen`, user agents will apply the default allowlist `self`. In this mode, `fullscreen` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
```

files/en-us/web/http/reference/headers/permissions-policy/geolocation/index.md
```diff
@@ -36,40 +32,37 @@ Permissions-Policy: geolocation=<allowlist>;
...
-```http
-Permissions-Policy: geolocation=(self)
+```html
+<iframe src="https://example.com/map" allow="geolocation"></iframe>
 ```
 
-Then include an {{HTMLElement('iframe','allow','#Attributes')}} attribute on the
-`<iframe>` element:
+> [!NOTE]
+> Specifying the `Permissions-Policy` header in this manner disallows `geolocation` for other origins, even if they were allowed by the `<iframe>` `allow` attribute.
```
chrisdavidmills:
```suggestion
> Specifying the `Permissions-Policy` header in this manner disallows `geolocation` for other origins, even if they are allowed by the `<iframe>` `allow` attribute.
```

files/en-us/web/http/reference/headers/permissions-policy/geolocation/index.md
```diff
@@ -36,40 +32,37 @@ Permissions-Policy: geolocation=<allowlist>;
...
+
+### Using the default policy
+
+Without delivering an HTTP response header defining a Permissions Policy for `geolocation`, user agents will apply the default allowlist `self`. In this mode, `geolocation` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
```
chrisdavidmills:
```suggestion
If a `Permissions-Policy` response header is not sent to define a Permissions Policy for `geolocation`, user agents will apply the default allowlist `self`. In this mode, `geolocation` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
```

files/en-us/web/http/reference/headers/permissions-policy/bluetooth/index.md
bban160: The specific wording here makes a big difference. What matters is not whether a `Permissions-Policy` header is sent, but whether a sent `Permissions-Policy` header (if any) includes the `bluetooth` directive.

chrisdavidmills: My wording suggestion still includes "not sent to define a Permissions Policy for `bluetooth`" — IMO, this is specific enough. If the wording is not quite right, feel free to suggest an alternative. I just felt that the "Without delivering an HTTP response header defining" bit was awkwardly phrased.

bban160: How about:

```suggestion
If a Permissions Policy for `bluetooth` is not defined by a `Permissions-Policy` response header, user agents will apply the default allowlist `self`. In this mode, `bluetooth` is automatically allowed in the top-level browsing context and same-origin iframes, but not in cross-origin iframes.
```

I think this does a better job of emphasizing the important detail.

chrisdavidmills: Yes, that works for me, thanks @bban160!

bban160: I ultimately changed it to "If an allowlist for…" which is slightly more succinct.

hamishwillee:
Thanks to both of you. @bban160 - any particular ETA on the equivalent changes in the guide? If you don't have a timeline for this can you let me know - then I can create an issue and we can make sure this doesn't get forgotten.

bban160:
> any particular ETA on the equivalent changes in the guide? If you don't have a timeline for this can you let me know - then I can create an issue and we can make sure this doesn't get forgotten.

No timeline at the moment, a dedicated issue would be welcome.

Unfortunately I've been a little too busy to prioritize this work, hence the approach of slicing it into smaller chunks that I can tackle over time.

Issue #42530, opened 2025-12-25
audioData.copyTo is missing format option


dy:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/AudioData/copyTo

### What specific section or headline is this issue about?

options

### What information was incorrect, unhelpful, or incomplete?

https://developer.mozilla.org/en-US/docs/Web/API/AudioData/copyTo#options

### What did you expect to see?

Have `format` option that can copy to Float32Array. If not AI I would not be able to find that out

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/api/audiodata/copyto`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/AudioData/copyTo
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/audiodata/copyto/index.md
* Last commit: https://github.com/mdn/content/commit/06b418a190b8e4a46682ab706d14984e7db34862
* Document last modified: 2024-09-11T03:36:25.000Z

</details>

MUSKANNISHAD:
I checked the WebCodecs spec (#dom-audiodata-copyto) and AudioDataCopyToOptions explicitly includes a format option, which affects how samples are written and converted.

The MDN page doesn’t currently document this option, so I think this is a documentation gap rather than a spec issue. Happy to open a PR to add this.

PR #42536, opened 2025-12-26
docs(webapi): document format option for AudioData.copyTo

MUSKANNISHAD:
### Description
This PR documents the `format` option for the `AudioData.copyTo()` method, which was missing from the MDN documentation.

### Motivation
The WebCodecs specification defines a `format` option in `AudioDataCopyToOptions`, but the MDN page did not mention it. This could mislead readers into assuming the option does not exist.

### Related issue
#42530

### DIFF
--- a/files/en-us/web/api/audiodata/copyto/index.md
+++ b/files/en-us/web/api/audiodata/copyto/index.md
@@ -24,10 +24,16 @@ copyTo(destination, options)
   - : An object containing the following:
     - `planeIndex`
       - : The index of the plane to copy from.
+    - `format` {{optional_inline}}
+      - : The sample format to use when copying audio data.
+        Possible values include `"f32"` and `"s16"`.
+        If omitted, the format of the destination buffer is used.
     - `frameOffset` {{optional_inline}}
-      - : An integer giving an offset into the plane data indicating which plane to begin copying from. Defaults to `0`.
+      - : An integer giving an offset into the plane data indicating which plane to begin copying from.
+        Defaults to `0`.
     - `frameCount` {{optional_inline}}
-      - : An integer giving the number of frames to copy. If omitted then all frames in the plane will be copied, beginning with the frame specified in `frameOffset`.
+      - : An integer giving the number of frames to copy.
+        If omitted then all frames in the plane will be copied, beginning with the frame specified in `frameOffset`.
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/audiodata/copyto/index.md
```diff
@@ -24,10 +24,16 @@ copyTo(destination, options)
   - : An object containing the following:
     - `planeIndex`
       - : The index of the plane to copy from.
+    - `format` {{optional_inline}}
+      - : The sample format to use when copying audio data.
+        Possible values include `"f32"` and `"s16"`.
```
wbamberg, line 29: We can't just document a couple of the possible values. But we probably don't have to list them all here, since they seem to be the same as https://developer.mozilla.org/en-US/docs/Web/API/AudioData/format.

So we could probably say:

```suggestion
        For possible values, see {{domxref("AudioData.format")}}.
```

files/en-us/web/api/audiodata/copyto/index.md
```diff
@@ -24,10 +24,16 @@ copyTo(destination, options)
...
+        If omitted, the format of the destination buffer is used.
```
wbamberg, line 30:
```suggestion
        If this option is omitted, the format of the destination buffer is used.
```

files/en-us/web/api/audiodata/copyto/index.md
```diff
@@ -24,10 +24,16 @@ copyTo(destination, options)
...
     - `frameOffset` {{optional_inline}}
-      - : An integer giving an offset into the plane data indicating which plane to begin copying from. Defaults to `0`.
+      - : An integer giving an offset into the plane data indicating which plane to begin copying from.
+        Defaults to `0`.
     - `frameCount` {{optional_inline}}
-      - : An integer giving the number of frames to copy. If omitted then all frames in the plane will be copied, beginning with the frame specified in `frameOffset`.
+      - : An integer giving the number of frames to copy.
+        If omitted then all frames in the plane will be copied, beginning with the frame specified in `frameOffset`.
```
wbamberg, line 36:
```suggestion
        If this option is omitted, then all frames in the plane will be copied, beginning with the frame specified in `frameOffset`.
```

PR #42543, opened 2025-12-26
fix: correct inconsistent IIR filter coefficient variable name

limonatasuyu:


wbamberg:
Thank you @limonatasuyu !

PR #42544, opened 2025-12-26
Fix default allowlist for gamepad policy-controlled feature

bban160:
### Description

The default allowlist was changed from `self` to `*`. The spec and all 3 major browser implementations reflect this change.

### Additional details

- https://www.w3.org/TR/2025/WD-gamepad-20250710/#permission-policy
- https://github.com/w3c/gamepad/pull/156
- https://issues.chromium.org/issues/40101717#comment7
- https://phabricator.services.mozilla.com/D119471
- https://bugs.webkit.org/show_bug.cgi?id=230136#c2

Relates to https://github.com/mdn/browser-compat-data/pull/11770


hamishwillee (approved):
Thanks very much. The compatibility data is a bit confused on this one so I also added a specification link, and we might come back to this when https://github.com/mdn/browser-compat-data/pull/28744 is looked at

Issue #42547, opened 2025-12-27
Range interface documentation has incorrect inheritance description


BenTownsend21:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Range

### What specific section or headline is this issue about?

Instance Properties

### What information was incorrect, unhelpful, or incomplete?

The Instance Properties section of the Range interface says that "There are no inherited properties."

Compare this to the Static Range page which says "The properties below are inherited from its parent interface, AbstractRange"

### What did you expect to see?

I expected to see the Range interface documentation correctly indicate that it inherits its properties from its parent interface AbstractRange, as is done in the similar StaticRange interface.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/api/range`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Range
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/range/index.md
* Last commit: https://github.com/mdn/content/commit/2937558d5ed1e03d7f60b2de71dd9c17f490166e
* Document last modified: 2023-02-19T10:03:58.000Z

</details>

PR #42549, opened 2025-12-27
docs(webapi): fix inherited properties description for Range

MUSKANNISHAD:
### Description
Fixes an incorrect statement in the Range interface documentation that claimed there are no inherited properties.

The page now correctly states that Range inherits properties from `AbstractRange`.

### Motivation
The previous description contradicted both the DOM specification and related MDN pages (e.g. StaticRange), which could confuse readers about the Range inheritance model.

### Related issue
#42547

### DIFF
--- a/files/en-us/web/api/window/requestanimationframe/index.md
+++ b/files/en-us/web/api/window/requestanimationframe/index.md
@@ -146,7 +146,7 @@ function animate() {
   const value = (performance.now() - zero) / duration;
   if (value < 1) {
     element.style.opacity = value;
-    requestAnimationFrame(animate);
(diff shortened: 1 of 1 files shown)


wbamberg:
Hi! This PR contains some unrelated changes. I see you've opened quite a few PRs recently, and many of them have similar problems.

Please check when you open a PR that it has exactly the changes you expect. I think it would be better for you to open fewer PRs and to be more careful with the PRs that you open.

MUSKANNISHAD:
Thanks for pointing this out — you’re right.
I’ll update this PR to include only the intended change and will be more careful to keep future PRs focused.

PR #42566, opened 2025-12-29
fix a typo in `flex-shrink`

mfuji09:
### Description

Fix a typo (`flex-basis` to `flex-shrink`)

### Motivation

This article discusses about `flex-shrink`, but it talks about `flex-basis` incorrectly.


hamishwillee (approved):
Makes sense. Thanks @mfuji09

PR #42569, opened 2025-12-29
Clarify that HTML document prefetching is no longer recommended

jikrana-del:
### Description
This pull request updates the `rel="prefetch"` documentation to clarify that
prefetching HTML documents for navigation is no longer recommended in modern browsers.

### Motivation

The documentation previously suggested that HTML documents could be prefetched
for future navigation using `rel="prefetch"`. However, modern browser guidance
recommends using the Speculation Rules API for document prefetching instead.

This change helps prevent readers from adopting outdated or suboptimal patterns
and aligns MDN guidance with current best practices.

### Additional details

- Added clearer wording to discourage HTML document prefetching with `rel="prefetch"`.
- Emphasized the use of `rel="prefetch"` for subresources.
- Referenced the Speculation Rules API as the preferred alternative for navigation prefetching.

#42569


hamishwillee:
Thanks @jikrana-del  but this is incorrect. 
- [Speculation rules](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API) are only supported on chromium browsers. 
- Even if all browsers supported them, you might still use prefetch instead in many cases. The speculation rules are extremely heavyweight and apply to whole documents and not individual resources - they are a great choice when present in many but not all cases.

PR #42575, opened 2025-12-30
Fix(HTTP): Update Prefer header syntax to allow lists

### LINKED ISSUE #42535: `<preference>` specification for `Prefers` is misleading -- _at least_ one preference is required (`1#` for ABNF as per the RFC)
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Prefer

_No response_

The RFC says `1#` which means "at least one instance of".

Header values like `foo=1, bar=2` are permitted, with `foo=1` and `bar=2` being two distinct _preferences_, hence `<preference>` in MDN is misleading -- in the very least an improvement can be `<preference>...` although I am not sure what's the convention to communicate values in a list and how HTTP _kind of_ facilitates such lists using _multiple_ headers with the same name in a single request.

Should say `<preference> [ "," <preference> ]...` or some such (I can't tell for sure what syntax convention MDN uses).

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/http/reference/headers/prefer`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Prefer
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/reference/headers/prefer/index.md
* Last commit: https://github.com/mdn/content/commit/ad5b5e31f81795d692e66dadb7818ba8b220ad15
* Document last modified: 2025-07-04T01:10:07.000Z

</details>

Harish3005-cloud:
Updated the syntax block to show that multiple preferences are allowed as per the RFC. Added an example demonstrating multiple preferences in a single header. Fixes #42535

### Description

 I updated the "Syntax" section to reflect that the `Prefer` header accepts a comma-separated list of preferences. I also added a new "Providing multiple preferences" example to demonstrate this behavior.

### Motivation

The RFC 7240 specification defines the `Prefer` header grammar as `1#preference`, meaning multiple preferences are allowed in a comma-separated list. The previous documentation implied only a single preference was valid, which was misleading for developers trying to set multiple directives (e.g., `respond-async` and `wait`).

### Additional details

Reference: [RFC 7240 Section 2 - The Prefer Header Field](https://datatracker.ietf.org/doc/html/rfc7240#section-2)

### DIFF
--- a/files/en-us/web/http/reference/headers/prefer/index.md
+++ b/files/en-us/web/http/reference/headers/prefer/index.md
@@ -33,7 +33,7 @@ The HTTP **`Prefer`** header allows clients to indicate preferences for specific
 ## Syntax
 
 ```http
-Prefer: <preference>
+Prefer: <preference>, <preference>, ...
 ```
 
 ## Directives


hamishwillee (approved):
Thanks very much. There are a number of ways to show optional lists in HTTP section, but this is the most common one.

PR #42576, opened 2025-12-30
Fix glossary redirect warning

Harish3005-cloud:
Updated the glossary macro to point directly to "Forbidden request header" to resolve the build warning about the redirect.

### DIFF
--- a/files/en-us/web/http/reference/headers/prefer/index.md
+++ b/files/en-us/web/http/reference/headers/prefer/index.md
@@ -24,10 +24,7 @@ The HTTP **`Prefer`** header allows clients to indicate preferences for specific
       </td>
     </tr>
     <tr>
-      <tr>
-      <th scope="row">{{Glossary("Forbidden request header", "Forbidden header name")}}</th>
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/reference/headers/prefer/index.md
```diff
@@ -24,7 +24,10 @@ The HTTP **`Prefer`** header allows clients to indicate preferences for specific
       </td>
     </tr>
     <tr>
-      <th scope="row">{{Glossary("Forbidden header name")}}</th>
+      <tr>
+      <th scope="row">{{Glossary("Forbidden request header", "Forbidden header name")}}</th>
+      <td>No</td>
+    </tr>
```
hamishwillee: FYI This was wrong - multiple nested rows. But either way if you look at other examples this is how they now look. Also, if you're working on the same document in multiple PRs best to combine all the changes in one.


```suggestion
      <th scope="row">{{Glossary("Forbidden request header")}}</th>
```

PR #42577, opened 2025-12-30
Removed incorrect extra example for AES-CTR

Meterel:
### Description

Removed an incorrect extra example for AES-CTR encryption

### Motivation

It is likely a leftover draft that doesn't have an equivalent in https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/decrypt and that doesn't follow the rest by not being based upon the full example code and that uses in window.crypto.subtle.encrypt's algorithm parameter a length of 128 instead of the recommended 64 according to https://developer.mozilla.org/en-US/docs/Web/API/AesCtrParams, it also uses window.crypto.getRandomValues and window.crypto.subtle.importKey to generate the key instead of just window.crypto.subtle.generateKey

### DIFF
--- a/files/en-us/web/api/subtlecrypto/encrypt/index.md
+++ b/files/en-us/web/api/subtlecrypto/encrypt/index.md
@@ -162,34 +162,6 @@ function encryptMessage(key) {
 }
 ```
 
-```js
-let iv = window.crypto.getRandomValues(new Uint8Array(16));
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

sideshowbarker:
@Meterel, congrats on landing your first docs change here — welcome aboard 🎉

PR #42579, opened 2025-12-30
Improve maxTouchPoints example

tkrotoff:
### Description

Current maxTouchPoints example is misleading IMHO

### Motivation

I've checked the possible values of maxTouchPoints for different devices https://gist.github.com/tkrotoff/cd7ea4ec29f83111365772aaa21bb25b
and devices without touch screen ("classic" desktop computers) return 0

### DIFF
--- a/files/en-us/web/api/navigator/maxtouchpoints/index.md
+++ b/files/en-us/web/api/navigator/maxtouchpoints/index.md
@@ -16,7 +16,7 @@ contact points that are supported by the current device.
 
 A number.
 
-"Classic" desktop computers (Mac, Windows and Linux) return 0 while smartphones (Android and iOS) typically return a value of 5.
+This value is hardware-dependent. Desktop computers without a touchscreen (Mac, Windows and Linux) return 0 while smartphones (Android and iOS) typically return 5.
 
 ## Example
 


Josh-Cena:
In what way is this useful to the reader? You can't assume the value of this property and doing so doesn't really simplify any coding pattern anyway.

tkrotoff:
> In what way is this useful to the reader?

It should be `navigator.maxTouchPoints > 0`, and not `> 1` (original MDN example), to detect a device with a touch screen.

Josh-Cena:
The previous example says that it detects "multi-touch", not "touch". Usually if the device just supports a single touch point, then it isn't operationally distinct from a trackpad or mouse, so you shouldn't really do anything special. And I was talking about the extra remark under "Value".

files/en-us/web/api/navigator/maxtouchpoints/index.md
```diff
@@ -16,11 +16,13 @@ contact points that are supported by the current device.
...
+"Classic" desktop computers (Mac, Windows and Linux) return 0 while smartphones (Android and iOS) typically return a value of 5.
+
+## Example
 
 ```js
-if (navigator.maxTouchPoints > 1) {
-  // browser supports multi-touch
+if (navigator.maxTouchPoints > 0) {
+  // The device has a touch screen
 }
```
Josh-Cena: Let's make it more meaningful. Again, I don't think detecting "0 touch points" or "1 touch point" is ever meaningful because most gestures are implemented with `click`, `mousedown`, `mousemove`, etc., which don't distinguish between touch or cursor gestures.

```suggestion
if (navigator.maxTouchPoints > 1) {
  // Device supports tracking at least 2 touch points; offer complex
  // interaction gestures such as swiping with two/three fingers
} else {
  // Device only has 1 touch point or is not a touch screen.
  // Offer basic gestures such as dragging and clicking
}
```

files/en-us/web/api/navigator/maxtouchpoints/index.md
```diff
@@ -16,11 +16,17 @@ contact points that are supported by the current device.
 
 A number.
 
-## Examples
+"Classic" desktop computers (Mac, Windows and Linux) return 0 while smartphones (Android and iOS) typically return a value of 5.
```
Josh-Cena:
```suggestion
This value is hardware-dependent. Desktop computers without a touchscreen (Mac, Windows and Linux) return 0 while smartphones (Android and iOS) typically return 5.
```

PR #42580, opened 2025-12-30
Refactor `chrome_settings_overrides` page to use `<dl>` syntax

RobertCraigie:
### Description

Adds anchors for all the fields in the `chrome_settings_overrides` web extensions docs so they can be directly linked to.

### Motivation

I wanted to share a link to the `search_provider` section but it is currently awkward to share a direct link to the right position as you can only use the "Copy link to highlight" version[0], and even then it doesn't scroll as nicely as it does with an explicit anchor on the row imo.

### Additional details

Screenshot 2026-01-02 at 8 49 51 pm

[0]: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/chrome_settings_overrides#search_url:~:text=Defines%20a%20search%20provider%20to%20add%20to%20the%20browser


wbamberg:
I'm not a reviewer for the webextensions docs, so this is just informational, but: an alternative which is adopted by the other MDN docs is to use a definition list for the properties, with [MDN's `<dl>` syntax](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#definition_lists).

These get anchors auto-applied in the platform so you can link to things like [`RequestInit.mode`](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#mode) or [`icons.sizes`](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/icons#sizes).

Advantages of doing this:
- it makes the webextension docs more consistent with the rest of MDN
- you can do it in just Markdown
- the description gets the whole width of the page, rather than being squeezed into a single column

The disadvantage is that you don't have a consistent/structured way to capture the property type, you just have to put it in the description.

If @rebloor wanted to do this, it should be done for all the manifest members of course, so as to be internally consistent.

rebloor:
Thanks @wbamberg I'm already reviewing this with the team.

Rob--W:
> I'm not a reviewer for the webextensions docs, so this is just informational, but: an alternative which is adopted by the other MDN docs is to use a definition list for the properties, with [MDN's `<dl>` syntax](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#definition_lists).
> 
> These get anchors auto-applied in the platform so you can link to things like [`RequestInit.mode`](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#mode) or [`icons.sizes`](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/icons#sizes).

I am in favor of following MDN's convention here. Ideally consistently across all articles that have such a table. We have to start somewhere, I'd be okay with this PR covering `chrome_settings_overrides` only.

One potential challenge here is that properties can be nested. In this specific example even, the `search_provider` row encapsulates several other properties. I'm not sure if the result (of using definition lists) would be as readable, but I'm willing to consider it.

> The disadvantage is that you don't have a consistent/structured way to capture the property type, you just have to put it in the description.

That's fine by me. We don't need a whole column for just a type; many other WebExtension documentation have the type inline at the front (e.g. [MDN: webRequest.onBeforeRequest](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onBeforeRequest#details)).

rebloor:
> One potential challenge here is that properties can be nested. In this specific example even, the `search_provider` row encapsulates several other properties. I'm not sure if the result (of using definition lists) would be as readable, but I'm willing to consider it.

We have numerouse example fo nexted properties lists in the API\s, e.g. [browserAction.onClicked](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/onClicked)

image

Arguably that's a much cleaner, obvious presentation of nested properties.

rebloor:
@RobertCraigie would you like to go ahead and make the changes. You can find more information on the syntax in. [MDN's <dl> syntax](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#definition_lists). If you have any issues, please let me know.

RobertCraigie:
Updated! It now looks like this:
Screenshot 2026-01-02 at 8 49 51 pm

Definitely much easier to read.

files/en-us/mozilla/add-ons/webextensions/manifest.json/chrome_settings_overrides/index.md
```diff
@@ -38,141 +38,98 @@ Use the `chrome_settings_overrides` key to override the browser's home page and
...
+
+    If two or more extensions both set this value, then the setting from
+    the most recently installed one will take precedence.
+
+    To override new tabs, use [chrome_url_overrides](/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/chrome_url_overrides) instead.
+
+    This is a [localizable property](/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization#internationalizing_manifest.json).
+
+- `search_provider`
+  - : Defines a search provider to add to the browser.
```
rebloor:
```suggestion
  - : `object`. Defines a search provider to add to the browser.
```

files/en-us/mozilla/add-ons/webextensions/manifest.json/chrome_settings_overrides/index.md
```diff
@@ -38,141 +38,98 @@ Use the `chrome_settings_overrides` key to override the browser's home page and
...
+
+    The search provider has a name and a primary search URL. Alternative
+    URLs may be provided, including URLs for more specialized searches
+    like image search. In the URL you supply, use
+    `{searchTerms}` to interpolate the search term into the
+    URL, like:
+    `https://www.discogs.com/search/?q={searchTerms}`. You can
+    also provide POST parameters to be sent along with the search.
+
+    The search provider will be presented to the user alongside the
```
rebloor:
```suggestion
    The search provider is presented to the user alongside the
```

files/en-us/mozilla/add-ons/webextensions/manifest.json/chrome_settings_overrides/index.md
```diff
@@ -38,141 +38,98 @@ Use the `chrome_settings_overrides` key to override the browser's home page and
...
+    built-in providers. If you include the
+    `is_default` property and set it to `true`, the
+    new search provider will be the default option. By supplying the
+    `keyword` property, you enable the user to select your
+    search provider by typing the keyword into the search/address bar
+    before the search term.
+
+    This is an object with the properties listed below. All string
```
rebloor:
```suggestion
    This object has these properties. All string
```

rebloor (changes requested):
@RobertCraigie thanks for these changes. I've left a couple of comments to flag that the `search_provider` is an object in its opening sentence and adjust the text before the properties list, give the reader now knows that it's an object.

PR #42588, opened 2025-12-31
Add glossary page for "registrable domain" instead of "eTLD"

wbamberg:
I'd like to link to a glossary definition of "registrable domain" [in the passkeys guide](https://github.com/mdn/content/pull/42338). It turns out that although we don't have one, we do have a page for eTLD.

But I think "registrable domain" is a better glossary entry, for three reasons:
- "registrable domain" is the [spec term](https://url.spec.whatwg.org/#host-registrable-domain). The spec doesn't talk about eTLD AFAICT.
- "registrable domain" is more comprehensible
- we never actually link to "eTLD" we only link to "eTLD+1", which is the same as registrable domain. So we currently have to link into a fragment of the glossary entry, rather than directly to it.

### DIFF
--- a/files/en-us/glossary/registrable_domain/index.md
+++ b/files/en-us/glossary/registrable_domain/index.md
@@ -17,7 +17,7 @@ theguardian.com
 sussex.ac.uk
 ```
 
-This means that all domains under each of these domains belong to the same organization. For example:
+All domains under a registrable domain belong to the same organization. For example:
 
 ```plain
  film.theguardian.com


files/en-us/glossary/registrable_domain/index.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Registrable domain
+slug: Glossary/Registrable_domain
+page-type: glossary-definition
+sidebar: glossarysidebar
+---
+
+A **registrable domain** is a domain that can be (or has already been) registered by an individual or an organization through a domain name registrar. It's the highest level domain for a single {{glossary("site")}} on the web.
+
+Registrable domains are created directly underneath an _effective top-level domain_ (eTLD), such as `.org`, `.com`, or `.ac.uk`. For this reason a registrable domain is sometimes called an "eTLD+1".
```
hamishwillee, line 10: Underneath doesn't feel like the right word - its a more specific version

It might be better to word something like:

> Registrable consists of an _effective top-level domain_ (eTLD), such as `.org`, `.com`, or `.ac.uk`, with a preceding suffix chosen by the owner (such as., example.com). This structure is often referred to as eTLD+1.

files/en-us/glossary/registrable_domain/index.md
```diff
@@ -0,0 +1,42 @@
...
+
+For example, all of the following are registrable domains:
+
+```plain
+crookedtimber.org
+theguardian.com
+sussex.ac.uk
+```
+
+This means that all domains under each of these domains belong to the same organization. For example:
```
hamishwillee: I'm not sure about "this means", because this does not follow ..

Mabye something like

```suggestion
All domains under a registrable domain belong to the same organization. For example:
```

wbamberg: I think there is causality here - because an entity registers the domain, they own it and its subdomains. But I don't object to your version either.

files/en-us/glossary/registrable_domain/index.md
wbamberg: I don't know about this. Isn't "under" exactly what it is? I mean DNS is a hierarchy, that's why we call ".com" (for example) the **top-level**. Then "example.com" is created directly under ".com". Then "mail.example.com" is under "example.com".

e.g. from https://en.wikipedia.org/wiki/Domain_Name_System#Domain_name_syntax,_internationalization

> The right-most label conveys the [top-level domain](https://en.wikipedia.org/wiki/Top-level_domain); for example, the domain name www.example.com belongs to the top-level domain _com_.
>
> The hierarchy of domains **descends** from right to left; each label to the left specifies a subdivision, or [subdomain](https://en.wikipedia.org/wiki/Subdomain) of the domain to the right. For example, the label _example_ specifies a subdomain of the _com_ domain, and `www` is a subdomain of example.com.

(my emphasis)

wbamberg: We could I suppose say something like "A registrable domain is created as a subdomain of an eTLD".

hamishwillee: Hmmm. When I think about it I can't tell you why underneath annoys me, and  yet "below" does not. Perhaps because I see underneath as "literally over the top of" whereas "below is just at a lower level".
Anyway, I can't defend this, so closing.

files/en-us/glossary/registrable_domain/index.md
hamishwillee: Yes, but it reads like "Registrable domains are created directly underneath an _effective top-level domain_" is the _cause of_ "All domains under a registrable domain belonging to the same organisation". These are two facts, but they don't follow each other from what I can see. Anyway, that's why I commented.

PR #42592, opened 2026-01-01
Update meta charset tag in HTML example

airbr:
Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.

Not very critical, but good to model in our example demo of very minimal HTML.

### Description

This is a single character update I propose, because the trailing slash is not needed, and this demo example aims to be as simple as possible.

### Motivation

Motivation is to fine tune the examples around minimal working HTML examples, which I am investigating.


chrisdavidmills:
Hi @airbr! I completely agree with your suggestion; however, MDN uses Prettier to lint all of its code listings, and Prettier lints HTML according to XHTML rules. This is good for things like quoting attribute values, but bad for trailing slashes.

We need to address this issue at some point, but apparently, it's not that straightforward.

PR #42595, opened 2026-01-01
Fix macros linking to `fit-content()` and `url()`

mfuji09:
### Description

Fix macros linking to `fit-content()` and `url()`.

### Motivation

To link to `fit-content()` page,
`{{cssxref("fit-content_function", "fit-content()")}}` is needed.
To link to `url()` page,
`{{cssxref("url_function", "url()")}}` is needed or it is shown wrongly.


estelle (approved):
I think we should update the macro.... but that is for a different day.

Thanks!

PR #42608, opened 2026-01-02
HTML `<hr>` element - new examples for lists and select

ramiy:
### Description

The article about `<hr>` element is basic. We can add additional examples.

### DIFF
--- a/files/en-us/web/html/reference/elements/hr/index.md
+++ b/files/en-us/web/html/reference/elements/hr/index.md
@@ -99,7 +99,7 @@ The `<hr>` tag can be placed within a list item for visual separation, to create
 <ol>
   <li>Mix dry ingredients thoroughly.</li>
   <li>Pour in wet ingredients.</li>
-  <li><hr style="background-color: blue;" /></li>
+  <li><hr /></li>
   <li>Mix for 10 minutes.</li>
   <li>Bake for one hour at 300 degrees.</li>
 </ol>
--- a/files/en-us/web/html/reference/elements/hr/index.md
+++ b/files/en-us/web/html/reference/elements/hr/index.md
@@ -82,12 +82,6 @@ The following example adds a thematic break between paragraph-level elements.
 
 The `<hr>` tag can be placed within a list item for visual separation, to create a separator between sections of a list.
 
-There are three primary types of lists in HTML, all of which use `<li>` elements for the individual list items:
-
-* Unordered List (`<ul>`): Used for grouping items where the order does not matter. Items are typically displayed with bullet points.
-* Ordered List (`<ol>`): Used for items where the order is meaningful (e.g., steps in a recipe). Items are typically displayed with a numerical or alphabetical marker.
-* Description List (`<dl>`): Used for term/description pairs, using `<dt>` for the term and `<dd>` for the description. 
-
 #### HTML
 
 ```html
--- a/files/en-us/web/html/reference/elements/hr/index.md
+++ b/files/en-us/web/html/reference/elements/hr/index.md
@@ -89,7 +89,6 @@ The `<hr>` tag can be placed within a list item for visual separation, to create
   <li>Cut</li>
   <li>Copy</li>
   <li>Paste</li>
-  <li>Paste Style</li>
   <li><hr /></li>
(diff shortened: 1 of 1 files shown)


files/en-us/web/html/reference/elements/hr/index.md
```diff
@@ -56,25 +56,80 @@ This element's attributes include the [global attributes](/en-US/docs/Web/HTML/R
...
+
+#### HTML
+
+```html-nolint
+<h2>My To-Do List</h2>
+<ul>
+  <li>Task 1</li>
+  <li>Task 2</li>
+  <li>Task 3</li>
+  <li><hr /> <!-- Thematic break --></li>
```
estelle: Semantically, and likely a11y-wise, I don't feel comfortable with this example. An item is created, but the item has no content (HR has a role of presentation). The screen reader would read "list with 6 items", but only 5 items have content. The HR would have a bullet?

files/en-us/web/html/reference/elements/hr/index.md
```diff
@@ -56,25 +56,80 @@ This element's attributes include the [global attributes](/en-US/docs/Web/HTML/R
...
+  <li>Completed Task A</li>
+  <li>Completed Task B</li>
+</ul>
+
+<h2>Steps in a Recipe</h2>
+<ol>
+  <li>Mix dry ingredients thoroughly.</li>
+  <li>Pour in wet ingredients.</li>
+  <li><hr /> <!-- Thematic break --></li>
```
estelle: similar issue, but even more concening as step 3 has no content.

estelle (commented):
i don't htink we should be using the li examples

ramiy:
@estelle Thanks for the feedback. You’re absolutely right, it wasn’t a good example. I’ve replaced it with a more practical one. I’d love to hear your thoughts.

files/en-us/web/html/reference/elements/hr/index.md
```diff
@@ -56,25 +56,88 @@ This element's attributes include the [global attributes](/en-US/docs/Web/HTML/R
...
+The `<hr>` tag can be placed within a list item for visual separation, to create a separator between sections of a list.
+
+#### HTML
+
+```html
+<ul>
+  <li>Cut</li>
+  <li>Copy</li>
+  <li>Paste</li>
+  <li role="presentation"><hr /></li>
```
estelle, line 94: i have been marinating over this example, and have asked a few people in the a11y community their thoughts. 

Feedback was:

> Use `aria-hidden="true"` in place of `role="presentation"`  to completely remove the list item from the accessibility tree to keep it from being enumerated. The `<hr>` might have a functional purpose in that it clusters related content, but may be more confusing/annoying to a blind person than helpful, even if it is more "semantically pure."

Technically, it is a valid and accessible example, but is it a good one? Generally, if I need to create a thematic break between list items, I close out the list and open up a new list with the next "theme".

As MDN is considered gospel in many in-shop arguments over whether something is a good idea or not, I think we should omit valid content if there is a better, recommended way of doing it. By including this example, we are essentially saying "this is a good idea". So, while `<hr>` represents a thematic break between elements, all elements, not just paragraph level ones, it doesn't mean we need to demonstrate every use case. We should limit it to recommended use cases. I don't think we should be recommending this use case.

Josh-Cena: @estelle The HTML spec says:

> The [hr](https://html.spec.whatwg.org/multipage/grouping-content.html#the-hr-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) a [paragraph](https://html.spec.whatwg.org/multipage/dom.html#paragraph)-level thematic break, e.g., a scene change in a story, or a transition to another topic within a section of a reference book; alternatively, it represents a separator between a set of options of a [select](https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element) element.

So this is not a blessed use case.

estelle (approved):
not sure how this fell off my radar.

Josh-Cena:
@estelle I thought we were agreeing in https://github.com/mdn/content/pull/42608#discussion_r2668687602 that the list example should be removed?

PR #42612, opened 2026-01-03
Add RFC 7240 to the specifications list

LiviaMedeiros:
### Description

Add RFC 7240 to the known specs list.

### Motivation

This RFC is referenced as `spec-urls` in related pages: [`Prefer`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Prefer), [`Preference-Applied`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Preference-Applied), but shows up as `Unknown specification`.


wbamberg:
I've got nothing against this change, but it won't fix:

> This RFC is referenced as `spec-urls` in related pages: [`Prefer`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Prefer), [`Preference-Applied`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Preference-Applied), but shows up as `Unknown specification`.

This comes out of the {{Specifications}} macro, which IIRC at least used to use the https://www.npmjs.com/package/browser-specs package to look up the spec name from its URL. It looks like RFC 7240 is not included in https://github.com/w3c/browser-specs/blob/main/specs.json, perhaps [we could file an issue to request it](https://github.com/w3c/browser-specs/blob/main/CONTRIBUTING.md#how-to-add-a-spec)?

(that's assuming {{Specifications}} does still use browser-specs. I don't know where the code for it is any more).

wbamberg:
I filed https://github.com/w3c/browser-specs/issues/2282, let's see what happens :).

LiviaMedeiros:
Thanks!

If this relies on `browser-specs`, I'm not sure what's the best way to solve this. `Prefer` and `Preference-Applied` are not really related to *browsers* so there's a chance that https://github.com/w3c/browser-specs/issues/2282 would be wontfix'd. But let's see.

Edit: nvm, https://github.com/w3c/browser-specs/pull/2283 got merged. 🎉

wbamberg:
> Thanks!
> 
> If this relies on `browser-specs`, I'm not sure what's the best way to solve this. `Prefer` and `Preference-Applied` are not really related to _browsers_ so there's a chance that [w3c/browser-specs#2282](https://github.com/w3c/browser-specs/issues/2282) would be wontfix'd. But let's see.
> 
> Edit: nvm, [w3c/browser-specs#2283](https://github.com/w3c/browser-specs/pull/2283) got merged. 🎉

🎉  I don't know how long it will take for MDN to pick up a new version of browser-specs, but when it does we can see if that fixes this issue.

wbamberg:
This change made it into browser-specs 4.65. I just updated my local mdn/content, it includes 4.65, and the spec table shows the right title:

Screenshot 2026-01-06 at 7 47 50 AM

I guess this will make it into production in the next day or so. Thanks for reporting the issue!

I'll also merge this PR, since it seems like a sensible addition.

PR #42619, opened 2026-01-04
Fix command event example

harapeko:
Use event.command instead of event.action in the CommandEvent example.

Spec: https://html.spec.whatwg.org/multipage/interaction.html#the-commandevent-interface

### Description

Fixes the HTMLElement command event example to check event.command.

### Motivation

The example should use CommandEvent.command per the spec; event.action is for NotificationEvent.

### Additional details

Spec link above.

 Relates to: https://github.com/mdn/translated-content/pull/31782


wbamberg (approved):
👍 thank you @harapeko !

PR #42620, opened 2026-01-04
Fix a property name of `CommandEvent`

mfuji09:
### Description

Fix a property name of `CommandEvent`.

### Motivation

A `CommandEvent` object has a property `command` but doesn't have `action`.


mfuji09:
duplicated: #42619

PR #42622, opened 2026-01-04
Fix link to Tutorials in games index

davidrabinovich:


Josh-Cena:
The original link is correct.

PR #42623, opened 2026-01-05
Typo fix: Ensure consistent formatting in cell

nikelborm:
Reformatted the representation of GDK_KEY_AudioForward and Qt:Key_AudioForward for consistency with other cells' style.

### DIFF
--- a/files/en-us/web/api/ui_events/keyboard_event_key_values/index.md
+++ b/files/en-us/web/api/ui_events/keyboard_event_key_values/index.md
@@ -2417,8 +2417,7 @@ The multimedia keys are extra buttons or keys for controlling media devices, fou
       </td>
       <td><code>APPCOMMAND_MEDIA_FAST_FORWARD</code></td>
       <td></td>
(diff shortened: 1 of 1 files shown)


nikelborm:
Oh, yes, you're right! Sorry, I didn't mean to make it this way. Thanks for catching that! I was overconfident in the change, because it seemed very trivial, so I didn't validate it does what I intended, before sending it, because I made it in GitHub's UI. I meant to make 2 code tags that cover only the words individually instead of 1 code tag that spans across the whole cell. In all other cells there are 2 code tags instead of 1 spanning across the whole cell. Basically the fix of the same fashion as in https://github.com/mdn/content/pull/42624. Fixed the tag that wasn't split-apart in-between. I now fixed the PR. This is the consistent style I wanted to bring it to:

|Screenshot From 2026-01-07 02-44-40|Screenshot From 2026-01-07 02-44-31|
|-|-|

|Main before PR|Current state of the PR|
|-|-|
|Screenshot From 2026-01-07 02-48-18|Screenshot From 2026-01-07 02-47-46|

PR #42624, opened 2026-01-05
Fixed table in Document keys section

nikelborm:


wbamberg (approved):
👍 thank you @nikelborm !

PR #42630, opened 2026-01-05
Clarify timestamp reduced precision behavior across browsers/APIs

### LINKED ISSUE #35239: HTMLMediaElement.currentTime: precision seems different from Date.now
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime

Value

"A double-precision floating-point value indicating the current playback time in seconds."

I'm wondering if there is something that should be added to that definition. I've been looking a a slider that uses onChange to record the currentTime and the date from Date.now. Interestingly, sometimes the currentTime value doesn't change even though Date.now _has_ changed. Given that currentTime is shown to six decimal places this seems unlikely. I've been able to replicate this in different browsers and sliders so I suspect I'm looking at 'feature' rather than 'bug' but can't work out what it is. Is currentTime the time of the most recent 'chunk' of audio (albeit a small one)?

Here's a quick grab from the log showing currentTime on the left and Date.now() on the right formatted as h:mm:ss.000. The second and third entries have the same currentTime.
4.550919	9:38:10.768
4.558815	9:38:10.776
4.558815	9:38:10.785
4.566465	9:38:10.792

Apologies for being such a pedant!

### LINKED ISSUE #40673: Event timeStamp accuracy description is misleading
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp

_No response_

Says
> The value is a [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) accurate to 5 microseconds (0.005 ms)

It's unclear what is even trying to be communicated.  Looking at what the link says ("DOMHighResTimeStamp types are coarsened"), it appears that values should be rounded to the nearest 0.005 in isolated contexts or rounded to 0.100 in non-isolated contexts.  This simply isn't the case (see more details below).  Thus it appears that this is false:

> The value is a [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp)

It's just a number that may or may not meet the definitions of that type.  If it's going to be, one day, then it would be good to note that it is not yet.  (As it says [here](https://developer.mozilla.org/en-US/docs/Web/API/Event): "work is underway to change this to be a [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) instead.")

Says
> the value of `privacy.resistFingerprinting.reduceTimerPrecision.microseconds`

Should be 1/1000 this value.  We're talking about milliseconds.

Likewise
> or `privacy.resistFingerprinting.reduceTimerPrecision.microseconds`

1/1000 of this, because the value in milliseconds is being discussed.

Says
> will always be a multiple of 2, or a multiple of 100 (or ...

No.  With `privacy.resistFingerprinting` and `privacy.resistFingerprinting.reduceTimerPrecision.microseconds` at default of 1000, it's a number, usually with 3 decimal places, not always an integer like in the examples and as assumed in this statement.  (Because of this fact, it's confusing to speak of reduced precision.  There's precision more-or-less to the microsecond.  Perhaps the accuracy is reduced, not the apparent precision.)

That's the case until `privacy.resistFingerprinting.reduceTimerPrecision.microseconds` reaches about 16667 (or 16670 with fewer rounding errors); then the precision is a multiple of 1/1000 the value set, but all values of `privacy.resistFingerprinting.reduceTimerPrecision.microseconds` below this number do not produce any reliable multiples.

Says
> the precision will be 100ms

Also says
> a multiple of 100

Seems to assume the default of `privacy.resistFingerprinting.reduceTimerPrecision.microseconds` is 100000, but it's 1000.

Says
> the precision will be 100ms or the value of `privacy.resistFingerprinting.reduceTimerPrecision.microseconds`, whichever is larger.

No, as described above, the precision starts at values around 16667 and greater (equivalent to 16.667 ms), and at lower values, there is no apparent reduction in precision.  But according to this ("whichever is larger"), the precision would never be finer than 100 ms, regardless of the value.

Says
```
// reduced time precision with `privacy.resistFingerprinting` enabled
event.timeStamp;
// Might be:
// 53500
// 58900
// 64400
// …
```

It might be, but probably not, since 100000 is not the default.

More accurate descriptions would be nice.

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/api/event/timestamp`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/event/timestamp/index.md
* Last commit: https://github.com/mdn/content/commit/3b5a1c0dfd59257c0a51052a9efa7b0108f8ecca
* Document last modified: 2024-07-04T22:52:34.000Z

</details>

harshshah-2023:
This updates the Event.timeStamp documentation to remove misleading accuracy guarantees and clarify how time resolution may be affected by browser privacy settings.

Fixes #40673

Fixes https://github.com/mdn/content/issues/35239


Josh-Cena (approved):
Thank you. I went through the browser sources and discovered AT LOT of errors/omissions in our current content, even regarding Firefox. Basically `resistFingerprinting` hasn't worked as described for at least 3 years. Updated everything to describe the current state of things, removing copy/paste wherever appropriate.

Going to merge now. This might be subject to further change/update, but at least this reflects the best of my knowledge.

PR #42635, opened 2026-01-05
fix conic gradient various units example

mateuszJS:
### Description

Somebody probably thought rad works same as turn. Here is fixed examples that correctly uses radians to demonstrate that conic-gradient accepts various units. Example is not displayed anywhere, it's just snippet of code.


chrisdavidmills (changes requested):
Hi there @mateuszJS, and thank you for your contribution. Your code certainly works better.

I agree with the copilot suggestion that `calc(pi * 1.5 * 1rad)` should be simplified to `calc(pi * 1.5rad)`, as the example is already pretty complex. Once you've made this change, I'll get it merged.

mateuszJS:
> Hi there @mateuszJS, and thank you for your contribution. Your code certainly works better.
> 
> I agree with the copilot suggestion that `calc(pi * 1.5 * 1rad)` should be simplified to `calc(pi * 1.5rad)`, as the example is already pretty complex. Once you've made this change, I'll get it merged.

Agree, fixed: 8d1447285441cbee71dadff02e63855ccc01ed9e

chrisdavidmills (approved):
Perfect, thanks @mateuszJS.

PR #42636, opened 2026-01-05
Added interactive example

asummers1:
### Description

I've added an interactive pointer example to this page, because I think it's one of the more common usages of `cursor`.
### Motivation

Adding a `cursor: pointer` example makes it quicker for readers to learn about and use this CSS property.


chrisdavidmills (changes requested):
Hi @asummers1, and thank you for your contribution! 

I agree that this is a worthy addition — `pointer` is definitely one of the most commonly used `cursor` values.

One issue created by your addition is that, in the rendered page, the interactive example choices are now a bit squashed, and a scroll bar is now rendered (you can see this in the generated live preview link). I wondered if you would consider removing another, less commonly-used value choice, maybe `grab`?

asummers1:
@chrisdavidmills Updated, thanks!

PR #42638, opened 2026-01-06
TrustedTypes: Indirect sinks

hamishwillee:
This expands on the concept of Indirect injection sinks with a real example. Note that I haven't tried to list all the alternatives, just explained that TT does deal with these cases.

@fred-wang If you have bandwidth, would appreciate your sanity check. This fall out of our discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1928932#c22

Related docs work in #37518 and #41507

### DIFF
--- a/files/en-us/web/api/trusted_types_api/index.md
+++ b/files/en-us/web/api/trusted_types_api/index.md
@@ -209,31 +209,8 @@ When the text node is created there is no reason for the browser to assume it is
 Instead, browsers run the checks when the script element becomes executable — i.e., in this example, when `document.body.appendChild(script)` is called to add the script element to the document.
 
 The browser will first check if the string used as the script content is trusted.
-Whether the string can be trusted depends on how it was set as the {{htmlelement("script")}} source.
-
-<!-- List below derived from https://wpt.fyi/results/trusted-types/script-enforcement-001.html?label=experimental&label=master&aligned -->
-
-The source string is trusted if the script was modified using the following APIs:
-
-- Script source set via {{domxref("TrustedScript")}} sink on {{domxref("HTMLScriptElement.innerText")}}, {{domxref("HTMLScriptElement.textContent")}}, {{domxref("HTMLScriptElement.text")}}
-- Splitting script source via {{domxref("Text.splitText()")}}
-- Normalizing script source via `Element.normalize()`
-
-The source string is not trusted if modified with the following APIs:
-
-- Setting script source via {{domxref("HTMLElement.innerText")}}
-- Setting script source via {{domxref("Node.textContent")}} or {{domxref("Node.nodeValue")}}
-- Script source set via {{domxref("TrustedHTML")}} sink {{domxref("Element.innerHTML")}} or {{domxref("Element.setHTMLUnsafe()")}}
-- Script source cannot be set via {{domxref("Element.setHTML()")}}.
-- Setting script source via {{domxref("CharacterData")}} property {{domxref("CharacterData.data","data")}} and methods {{domxref("CharacterData.appendData()","appendData()")}}, {{domxref("CharacterData.insertData()","insertData()")}}, {{domxref("CharacterData.replaceData()","replaceData()")}}, {{domxref("CharacterData.deleteData()")}}, {{domxref("CharacterData.before()","before()")}}, {{domxref("CharacterData.after()","after()")}}, {{domxref("CharacterData.remove()","remove()")}}, {{domxref("CharacterData.replaceWith()","replaceWith()")}}
-- Setting script source via {{domxref("Node.appendChild()")}}, {{domxref("Node.insertBefore()")}}, {{domxref("Node.replaceChild()")}}, {{domxref("Node.removeChild()")}}
-- Setting script source via {{domxref("Element")}} methods {{domxref("Element.prepend()","prepend()")}}, {{domxref("Element.append()","append()")}}, {{domxref("Element.replaceChildren()","replaceChildren()")}}, {{domxref("Element.moveBefore()","moveBefore()")}}
-- Setting script source via {{domxref("TrustedHTML")}} sink `Node.insertAdjacentHTML()`
-- Setting script source via `Node.insertAdjacentText()`
-- Setting script source via {{domxref("Range.insertNode()")}}
-- Setting script source via {{domxref("Range.deleteContents()")}}
-- Cloning a script via {{domxref("Node.cloneNode()")}}
-- Cloning a script via {{domxref("Range.cloneContents()")}}
+Any operation that allows the text source of a {{htmlelement("script")}} to be modified without explicitly setting a {{domxref("TrustedScript")}} makes it untrusted.
+The {{domxref("Node.appendChild()")}} method used above is just one example (a number of others are listed in the WPT Live tests at <https://wpt.live/trusted-types/script-enforcement-001.htm>).
 
 If the string is not trusted and trusted types are enforced, the browser will attempt to obtain a `TrustedScript` from a [default policy](#the_default_policy) to use for source instead.
 If a default policy is not defined, or does not return a `TrustedScript`, the operation will throw an exception.


files/en-us/web/api/trusted_types_api/index.md
```diff
@@ -178,6 +178,31 @@ These cases are evaluated the untrusted script is added to the document.
...
+Indirect injection sinks are those where untrusted strings are injected into the DOM via an intermediate mechanism that doesn't trigger script execution, and then evaluated.
+These kinds of sinks bypass the trusted type checks that have been added to the direct injection sinks.
+
+The following code shows how this might work (note that there are many other mechanisms for indirect injection).
+First a text node is created using a string provided by a user, and then a {{htmlelement("script")}} element is constructed and the text node is appended as a child element.
+Next the script element is added to the DOM as a child of the {{htmlelement("body")}} element, potentially allowing any scripts defined in the original string to be run.
+
+```js
+// Create a text node
+const unstrustedString = "/* Potentially malcious JavaScript code */";
```
fred-wang: nit: malicious*

files/en-us/web/api/trusted_types_api/index.md
```diff
@@ -138,8 +138,8 @@ element.innerHTML = userInput;
 
 This section provides a list of "direct" injection sink interfaces.
 
-Note that there are cases where untrusted strings may be "indirectly injected", such as when an untrusted string is added as the child node of a script element, and then the element is added to the document.
-These cases are evaluated the untrusted script is added to the document.
+These are the API properties and methods which perform trusted type checks when they are evaluated.
+They can be passed trusted types (`TrustedHTML`, `TrustedScript`, or `TrustedScriptURL`) as well as strings, and must be passed trusted types when trusted type enforcement is enabled.
```
fred-wang: "when trusted type enforcement is enabled" you can still pass strings if there is a default policy that implicitly converts the input to trusted types.

files/en-us/web/api/trusted_types_api/index.md
```diff
@@ -178,6 +178,31 @@ These cases are evaluated the untrusted script is added to the document.
...
+const unstrustedString = "/* Potentially malcious JavaScript code */";
+const textNode = document.createTextNode(unstrustedString);
+
+// Create a script element and add the text node
+const script = document.createElement("script");
+script.appendChild(textNode);
+document.body.appendChild(script);
+```
+
+Because a text node isn't always used in a context where the text might be used as an injection sink, it does enforce trusted types.
```
fred-wang: Not sure I get this? Setting appending the text node does *not* enforce trusted types (there is no TT injection sinks at all in these APIs actually) but enforcement happens when the script is actually run (internally the script text is converted to a `TrustedScript` and then stringified again).

files/en-us/web/api/trusted_types_api/index.md
```diff
@@ -178,6 +178,31 @@ These cases are evaluated the untrusted script is added to the document.
...
+
+Instead, browsers that are enforcing trusted types will run trusted type checks when indirect injection sinks are is injected into the DOM (in the example above, when `appendChild()` is called).
```
fred-wang: I'm not sure I understand this sentence "indirect injection sinks are injected into the DOM"?

And there are two `appendChild()` calls, so it's ambiguous.

As I said, for scripts, this happens when it is about to be executed and we realize it was modified via an API that does not enforce TrustedScript.

files/en-us/web/api/trusted_types_api/index.md
```diff
@@ -178,6 +178,31 @@ These cases are evaluated the untrusted script is added to the document.
 - `url` argument to [`Worker()` constructor](/en-US/docs/Web/API/Worker/Worker#url)
 - `url` argument to [`SharedWorker()` constructor](/en-US/docs/Web/API/SharedWorker/SharedWorker#url)
 
+### Indirect injection sinks
+
+Indirect injection sinks are those where untrusted strings are injected into the DOM via an intermediate mechanism that doesn't trigger script execution, and then evaluated.
```
fred-wang: note: I don't know if "direct" or "indirect" injection sinks are names used in the spec, but ok.

files/en-us/web/api/trusted_types_api/index.md
```diff
@@ -178,6 +178,31 @@ These cases are evaluated the untrusted script is added to the document.
...
+// Create a script element and add the text node
+const script = document.createElement("script");
+script.appendChild(textNode);
+document.body.appendChild(script);
+```
+
+Because a text node isn't always used in a context where the text might be used as an injection sink, it does enforce trusted types.
+
+Instead, browsers that are enforcing trusted types will run trusted type checks when indirect injection sinks are is injected into the DOM (in the example above, when `appendChild()` is called).
+This will cause an exception if the text node was constructed with a string rather than an appropriate trusted type.
```
fred-wang: Note sure I understand this either. `document.createTextNode` always creates the text node from a string. The condition is whether the script element's text was modified via a TrustedScript injection sinks (such as `HTMLScriptElement.text`, etc) or by other APIs (like appending a text node).

Incidentally, note that `HTMLScriptElement.innerHTML` is a TrustedHTML injection sinks, so enforcement would still be needed when the script is executed.

files/en-us/web/api/trusted_types_api/index.md
hamishwillee:
```suggestion
They can be passed trusted types (`TrustedHTML`, `TrustedScript`, or `TrustedScriptURL`) as well as strings, and must be passed trusted types when trusted type enforcement is enabled and no default policy is defined.
```

files/en-us/web/api/trusted_types_api/index.md
hamishwillee: I didn't find a specific section in the spec explain all this or using terms I could steal. This seems like a reasonable term/distinction to me though. Happy to accept a better one!

files/en-us/web/api/trusted_types_api/index.md
hamishwillee:
```suggestion
const unstrustedString = "/* Potentially malicious JavaScript code */";
```

files/en-us/web/api/trusted_types_api/index.md
hamishwillee: Thanks. Typo "not". But let me make this more clear

```suggestion
Because a text node won't always used in contexts where its text is executable, trusted type checking isn't performed when the untrusted string is used to create the node.
```

files/en-us/web/api/trusted_types_api/index.md
hamishwillee: Thanks - yes, that text sucks. 

I think this is better. My understand is that you track when a string is trusted and you know when its about to be executed. In this case when the script element is appended to the DOM.

```suggestion
Instead, browsers run the trusted type checks when potentially untrustworthy code is about to be executed.
In the example above, this is when `document.body.appendChild(script)` is called to add the script element to the DOM.
```

files/en-us/web/api/trusted_types_api/index.md
hamishwillee: Thanks @fred-wang . 

I think you are saying that 
- the condition to know that the script element might contain untrustworthy text/code is when the script element text changes through a mechanism that doesn't take a TrustedScript
- The trusted type when the script is about to be executed (and would check that trustworthiness).

Hence the exception occurs when the script is added to the DOM.

With respect to this sentence, it is based on what is probably a false assumption - that the developer can _do something to prevent the exception when trusted types are enforced_. I.e. I was assuming that if you passed a TrustedScript when creating the textNode then you would not get an exception.

I think you are saying that if TT is enforced, there is no way to write code like this.

If not, what am I missing!

hamishwillee: PS Is this discussion entirely related to the script element, or are there other cases (do I need to discuss this mor generically)

I am away on holiday now for a few weeks. Appreciate your help.

sideshowbarker:
Y’all can safely ignore any notifications you received today about activity on this PR. I’ve been testing out some tooling for doing PR reviews — specifically `octo.nvim` and some `gh` cli commands — and using this PR as a sandbox/test-case, and some of it seems to indirectly cause some other actions other than the thing I’m actually trying to do (which is basically just, to review the diff and comment on it)

files/en-us/web/api/trusted_types_api/index.md
hamishwillee: @fred-wang Please see my responses above https://github.com/mdn/content/pull/42638/files#r2674942952

Specifically, I think you are saying that " if TT is enforced, there is no TT-safe way to write code like this.". In other words, the check is run when potentially unsafe code is about to be executed (added to the DOM) and will result in an exception if TT is enabled.

Presumably in this case even a default transform is not enough.

fred-wang: @hamishwillee 

I think essentially yes, but just to be sure here are the details of what's happening, so you can compare:

```
const script = document.createElement("script");
```

This creates a script element. Firefox internally uses a boolean saying that this script is trusted (initially set to "trusted"), while Chromium/WebKit internally caches the latest trusted string seen (initially set to an empty string).

```
script.textContent = input1;
```

`textContent` setter accepts `TrustedScript of DOMString` per the [WebIDL](https://w3c.github.io/trusted-types/dist/spec/#dom-htmlscriptelement-textcontent). The script source is either modified immediately if `input1` is a `TrustedScript` or goes through trusted type enforcement if `input1` is a string (either rejecting or obtaining a `TrustedScript` from a default policy). In any case, Firefox internally sets its boolean to "trusted" while Chromium/WebKit updates the latest trusted string to what the script text is finally set.

```
const textNode = document.createTextNode(input2);
```

`document.createTextNode()` accepts as argument a `DOMString` per the [WebIDL](https://dom.spec.whatwg.org/#interface-document). If `input2` is a `TrustedScript`, it will just be serialized before being passed to the method, there is no way we could distinguish it from a raw string. Anyway, at this point there is no reason to perform TrustedType enforcement, we are just creating a text node and don't know what we are going to use it for.

```
script.appendChild(textNode);
```

We append the `textNode` to the script, but again it is just a node with some raw string, there is no hint that "this is trusted script source". At this point, Firefox internally sets its boolean to "untrusted" while Chromium/WebKit don't modify the latest trusted string used for the script.

```
document.body.appendChild(script);
```

The script is inserted into the document and that will trigger its execution later. We need to make sure it is trusted script when that happens. Firefox internally checks the boolean associated to the script: if it's "trusted" it just goes ahead normally with script execution, if it's "untrusted" then it will perform trusted type enforcement on the script text first (either rejecting or obtaining a `TrustedScript` from a default policy and then serializing it again). Chromium/WebKit internally compare the script text against the latest trusted string, if these strings match they just go ahead normally with script execution, otherwise they will perform trusted type enforcement on the script text first (either rejecting or obtaining a `TrustedScript` from a default policy and then serializing it again).

In the spec `textContent`, is called a trusted type sink. Chrome/WebKit behavior is the one described in the spec, but there is a discussion to use Firefox's one instead (https://github.com/w3c/trusted-types/pull/579). The main differences are that Firefox has a stricter policy (it can run script enforcement even if the source is not modified at all e.g. if `input2` is an empty string in the example) but more memory efficient (Chromium/WebKit use extra memory to cache the latest trusted string).


https://wpt.fyi/results/trusted-types/script-enforcement-001.html contains tests for some DOM APIs on script elements, checking whether they will cause trusted type enforcement to be triggered before execution.

files/en-us/web/api/trusted_types_api/index.md
```diff
@@ -181,6 +181,63 @@ These cases are evaluated the untrusted script is added to the document.
...
+- Setting script source via {{domxref("Node.appendChild()")}}, {{domxref("Node.insertBefore()")}}, {{domxref("Node.replaceChild()")}}, {{domxref("Node.removeChild()")}}
+- Setting script source via {{domxref("Element")}} methods {{domxref("Element.prepend()","prepend()")}}, {{domxref("Element.append()","append()")}}, {{domxref("Element.replaceChildren()","replaceChildren()")}}, {{domxref("Element.moveBefore()","moveBefore()")}}
+- Setting script source via {{domxref("TrustedHTML")}} sink `Node.insertAdjacentHTML()`
+- Setting script source via `Node.insertAdjacentText()`
+- Setting script source via {{domxref("Range.insertNode()")}}
+- Setting script source via {{domxref("Range.deleteContents()")}}
+- Cloning a script via {{domxref("Node.cloneNode()")}}
+- Cloning a script via {{domxref("Range.cloneContents()")}}
```
hamishwillee: Note, this is from WPT tests. Originally I was hesitant to list these because I am not sure we can keep it complete. However without this it is a bit hard for people to know what they need to be careful of.

If there is strong disagreement we could just list a few "such as X, Y, Z" and perhaps point to tests for more information.

@sideshowbarker ?

files/en-us/web/api/trusted_types_api/index.md
```diff
@@ -181,6 +181,63 @@ These cases are evaluated the untrusted script is added to the document.
...
+// Add the script into the document, where it can run
+document.body.appendChild(script);
+```
+
+When the text node is created there is no reason for the browser to assume it is intended to be used as a trusted type source, so trusted types are serialized to string, and are not enforced.
+
+Instead, browsers run the checks when the script element becomes executable — i.e., in this example, when `document.body.appendChild(script)` is called to add the script element to the document.
+
+The browser will first check if the string used as the script content is trusted.
+Whether the string can be trusted depends on how it was set as the {{htmlelement("script")}} source.
```
hamishwillee: FYI, as per https://github.com/mdn/content/pull/42638#discussion_r2663584221 the mechanisms used to determine whether string trust depends on browsers. I have avoided that difference and just tried to say that "these are the things you can do that will trigger the tests".

files/en-us/web/api/trusted_types_api/index.md
hamishwillee: Thanks so much @fred-wang. 
I'm glad you did provide that level of detail because I was missing a couple of important details. I get it now. This has been updated and should now be correct.

Feel free to check again.

files/en-us/web/api/trusted_types_api/index.md
hamishwillee: PS @sideshowbarker This is ready for re-review.

fred-wang: @hamishwillee FWIW, I'm not sure this list is exhaustive. This is what I found while checking Firefox code, but I may have easily missed things. I think in general basically anything that can modified the text source of a script and does not explicitly accept a `TrustedScript` will make it untrusted (splitText  and normalize being exceptions since they never modify the text).

files/en-us/web/api/trusted_types_api/index.md
fred-wang: Yes, I was suggesting you should mention this difference on MDN, that was just FYI. Hopefully, the interop difference will be fixed at some point...

files/en-us/web/api/trusted_types_api/index.md
hamishwillee: Thanks very much @fred-wang . I've modified to make the statement "anything that can modified the text source of a script and does not explicitly accept a TrustedScript will make it untrusted" and link to the WPT test for more information.

This is better - in particular because people tend to assume lists are cannonical when they have this many items.

sideshowbarker (approved):

PR #42644, opened 2026-01-06
module not module landing page

estelle:
change the "module landing page" to "module", as they're not landing pages

### DIFF
--- a/files/en-us/web/css/guides/index.md
+++ b/files/en-us/web/css/guides/index.md
@@ -6,7 +6,7 @@ page-type: listing-page
 sidebar: cssref
 ---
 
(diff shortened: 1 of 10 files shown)


estelle:
> Also, if you just say " module", it makes it sound like we are linking to the spec. IMO, anyway.

The individual module pages are each a summary of the spec they represent. we're basically linking to the spec's representation on MDN

estelle:
> > Also, if you just say " module", it makes it sound like we are linking to the spec. IMO, anyway.
> 
> The individual module pages are each a summary of the spec they represent. we're basically linking to the spec's representation on MDN

The learning area uses "module landing page". the reference pages use simply "module". This PR is only for the CSS reference section.

PR #42645, opened 2026-01-06
Editorial review: Add information and example for counters in generated content alt text

chrisdavidmills:
### Description

Chrome 140 adds support for `counter()` and `counters()` functions in `content` property alt text. See https://chromestatus.com/feature/5185442420621312. Note that the initial implementation was buggy; it is fixed now, but this functionality only really works in Canary at the moment. See the associated BCD for further details: https://github.com/mdn/browser-compat-data/pull/28763.

This PR adds a bit more information and an example to the `content` property page, showing how to use counters in content alt text.

I also took this opportunity to mention the fact that you can also use `attr()` functions in content alt text. I tested it, and it seems to work, but I didn't want to go much deeper on that, as it is not the central purpose of this PR.

### DIFF
--- a/files/en-us/web/css/reference/properties/content/index.md
+++ b/files/en-us/web/css/reference/properties/content/index.md
@@ -345,7 +345,7 @@ If using a screen reader, it should speak the word "MOZILLA" when it reaches the
 
 ### Including counters in alternative text
 
-This example includes a list of links to a set of book chapters, and specifies generated content before each one that includes a book icon and a counter, with alternative text that includes the literal word "Chapter" in place of the icon. This results in an elegant chapter number announcement for screenreader users before each link text is read out.
+This example features a list of links to a set of book chapters, and shows how to use generated content to include a book icon and a counter before each one, with alternative text that includes the literal word "Chapter" in place of the icon. This results in the word "chapter" and the chapter number preceding the text in each link's {{glossary("accessible name")}}, which will be announced to screenreader users when the link receives focus.
 
 #### HTML
 


danielsakhapov:
LGTM! Thanks for adding this!

files/en-us/web/css/reference/properties/content/index.md
```diff
@@ -51,6 +51,7 @@ content: image-set("image1x.png" 1x, "image2x.png" 2x);
 
 /* speech output: alternative text after a "/"  */
 content: url("../img/test.png") / "This is the alt text";
+content: "" / "Chapter " counter(chapter);
```
estelle:
```suggestion
content: counter(chapter) / "Chapter " counter(chapter);
```
not best practices to provide alt text to empty content. The left side shouldn't be an empty string (it can be, but it shouldn't be if we're providing alt), so this was my attempt. you may have a different opinion of what that content should be.

files/en-us/web/css/reference/properties/content/index.md
```diff
@@ -131,8 +132,8 @@ The keywords and data types mentioned above are described in more detail below:
 - `attr(x)`
   - : The `attr(x)` CSS function retrieves the value of an attribute of the selected element, or the pseudo-element's originating element. The value of the element's attribute `x` is an unparsed string representing the attribute name. If there is no attribute `x`, an empty string is returned. The case sensitivity of the attribute name parameter depends on the document language.
 
-- alternative text: `/ <string> | <counter>`
-  - : Alternative text may be specified for an image or any `<content-list>` items, by appending a forward slash and then a string of text or a counter. The alternative text is intended for speech output by screen-readers, but may also be displayed in some browsers. The {{cssxref("string", "/ &lt;string>")}} or {{cssxref("counter", "/ &lt;counter>")}} data types specify the "alt text" for the element.
+- alternative text: `/ <string> | <counter> | attr()`
+  - : Alternative text may be specified for an image or any `<content-list>` items, by appending a forward slash and then a combination of strings, counters, or `attr()` functions. The alternative text is intended for speech output by screen-readers, but may also be displayed in some browsers.
```
estelle:
```suggestion
  - : Alternative text may be specified for an image or any `<content-list>` items, by appending a forward slash and then a combination of strings, counters, and `attr()` functions. The alternative text is intended for speech output by screen-readers, but may also be displayed in some browsers.
```

files/en-us/web/css/reference/properties/content/index.md
```diff
@@ -342,6 +343,59 @@ a::before {
 
 If using a screen reader, it should speak the word "MOZILLA" when it reaches the image. You can select the `::before` pseudo-element with your developer tools selection tool, and view the {{glossary("accessible name")}} in the accessibility panel.
 
+### Including counters in alternative text
+
+This example includes a list of links to a set of book chapters, and specifies generated content before each one that includes a counter. This results in an elegant chapter number announcement for screenreader users before each link text is read out.
+
+#### HTML
+
+We include a heading before creating an ordered list of chapter title links using {{htmlelement("ol")}}, {{htmlelement("li")}}, and {{htmlelement("a")}} elements.
```
estelle:
```suggestion
We include a heading followed by an ordered list of chapter title links using {{htmlelement("ol")}}, {{htmlelement("li")}}, and {{htmlelement("a")}} elements.
```

files/en-us/web/css/reference/properties/content/index.md
```diff
@@ -342,6 +343,59 @@ a::before {
...
+a::before {
+  content: "" / "Chapter " counter(chapter);
+}
+```
+
+#### Result
+
+{{EmbedLiveSample('alt-counter', '100%', 220)}}
+
+When using a screenreader, supporting browsers should speak the word "Chapter" followed by the current counter number, followed by the link text, when each list item is reached. So for example "Chapter 1 A stranger calls", "Chapter 2 Two owls", etc. This provides a better experience for screenreader user navigating the list.
```
estelle:
```suggestion
When a screenreader navigates to a link within the list, supporting browsers will announce "Chapter" followed by the current counter number, followed by the link text, for example, "Chapter 1 A stranger calls" and "Chapter 2 Two owls".
```
the text is part of the link, not the list item

files/en-us/web/css/reference/properties/content/index.md
```diff
@@ -342,6 +343,59 @@ a::before {
...
+Next, we set the `<li>` elements' {{cssxref("::marker")}} pseudo-elements to have generated `content` equal to the current `chapter` counter value (retrieved using the {{cssxref("counter()")}} function) plus a colon. We then set the `<a>` elements' {{cssxref("::before")}} pseudo-elements to have generated `content` equal to an empty string (so nothing is displayed), but with alt text equal to the current `chapter` counter value preceded by the word "Chapter".
+
+```css live-sample___alt-counter
+li::marker {
+  content: counter(chapter) ": ";
+}
+
+a::before {
+  content: "" / "Chapter " counter(chapter);
+}
```
estelle, line 387: the alt should be alt to actual content, not an empty string. The screenreader user may be sited, so should be hearing what sighted users see. I suggest getting rid of the marker (`list-style: none;`) then doing something like:

```css
a::before {
  content: url(chapter.png) counter(chapter) / "Chapter " counter(chapter);
}
```
so the alt is the alt for an image.
if you don't want to create an image, you can use characters like in https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@counter-style#examples or star rating like symbols https://codepen.io/estelle/pen/YzddjJb

estelle (changes requested):
Many screen reader aren't blind. The screen reader should be reading content that is on the page. Empty strings don't need descriptive text, but you would need alt text if the counter were using symbols, such as multiple stars, or a deck of cards, or binary. So we can either use an image that reads "chapter" then provide "chapter" as the alt, or we can use a symbolic counter that needs alt text. https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@counter-style#examples . 

I created a counter with a deck of cards once. I've been looking for it to share with you, but haven't found it yet. So submitting, and will ping you if i locate it.

files/en-us/web/css/reference/properties/content/index.md
chrisdavidmills: This is fine for a basic syntax example.

files/en-us/web/css/reference/properties/content/index.md
chrisdavidmills: Really good points here. I've figured out an update to the example that uses a book emoji to represent "Chapter". I'll make the updates and push them so you can have a look.

files/en-us/web/css/reference/properties/content/index.md
```diff
@@ -342,6 +343,56 @@ a::before {
 
 If using a screen reader, it should speak the word "MOZILLA" when it reaches the image. You can select the `::before` pseudo-element with your developer tools selection tool, and view the {{glossary("accessible name")}} in the accessibility panel.
 
+### Including counters in alternative text
+
+This example includes a list of links to a set of book chapters, and specifies generated content before each one that includes a book icon and a counter, with alternative text that includes the literal word "Chapter" in place of the icon. This results in an elegant chapter number announcement for screenreader users before each link text is read out.
```
estelle:
```suggestion
This example features a list of links to a set of book chapters, and shows how to use generated content to include a book icon and a counter before each one, with alternative text that includes the literal word "Chapter" in place of the icon. This results in the word "chapter" and the chapter number preceding the text in each link's {{glossary("accessible name")}}, which will be announced to screenreader users when the link receives focus.
```
a bit more pedantic

estelle (approved):
Just one edit, but looks good and I don't need to re-review. 

Edit: let's only merge this when the fix gets merged.

files/en-us/web/css/reference/properties/content/index.md
chrisdavidmills: I edited it a bit to improve clarity.

I don't think I need to wait before merging this, as I have updated the BCD to specify what the support situation is. I can update it again when the fix hits the Chrome release track.

PR #42649, opened 2026-01-06
Clarify parent-to-child DOM access with allow-same-origin

### LINKED ISSUE #42633: Clarify parent-side DOM access with allow-same-origin in <iframe> sandbox
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

Attributes > sandbox > allow-same-origin

The current description for the allow-same-origin token within the sandbox attribute is incomplete. It states that omitting it prevents access to "some JavaScript APIs," and the associated footnote for Chrome 70 mentions that "script execution" is blocked without allow-scripts.
This creates an ambiguity: it doesn't clarify that while internal script execution is blocked, the parent document can still interact with the iframe's DOM if allow-same-origin is present. Many developers (including myself) initially assume that allow-scripts is required for any JS-based interaction (like measuring scrollHeight), which is not the case for parent-to-child access.

I expected to see a clear distinction between internal script execution (governed by allow-scripts) and external DOM access from a same-origin parent.
I suggest adding a note to the allow-same-origin section: "Note: Even without allow-scripts, a parent document with the same origin can still access and manipulate the iframe's DOM. allow-scripts only controls scripts running inside the embedded context."

This behavior is defined in the HTML Living Standard. According to the specification for the sandbox attribute, the allow-same-origin keyword allows the content to be treated as being from its real origin.
The specification distinguishes between the "sandboxed origin browsing context flag" and the "sandboxed scripts browsing context flag". My issue highlights that lifting the former (via allow-same-origin) is enough for a same-origin parent to bypass the opaque origin restriction for DOM access, independent of the scripts flag.

Here is a minimal example to reproduce the behavior. It demonstrates that a parent can read the scrollHeight of a sandboxed iframe without allow-scripts, provided allow-same-origin is set.

> <iframe id="myIframe" sandbox="allow-same-origin" src="content.html"></iframe>

**The confusion:** The current MDN footnote regarding Chrome 70 and the general phrasing under `sandbox` often lead developers to believe that any script-related interaction (including from the parent) requires `allow-scripts`. Clarifying this would help developers implement the **Principle of Least Privilege** by not enabling `allow-scripts` when they only need to measure or style the iframe from the outside.

omidfarhangnia:
Adds a note under the allow-same-origin token explaining that a same-origin parent document can still access and manipulate the iframe's DOM even if allow-scripts is not set. The allow-scripts token only controls script execution inside the embedded document and does not affect parent-to-iframe DOM interactions.

Fixes #42633

### Description

Clarifies the distinction between internal script execution within a sandboxed <iframe> and parent-side DOM access. This helps developers understand that allow-same-origin is sufficient for same-origin parent access, and allow-scripts only governs scripts running inside the iframe itself.

### Motivation

Many developers assume that allowing scripts inside the iframe (allow-scripts) is required to access or measure DOM properties like scrollHeight from the parent. This PR clears up that misconception, supporting the Principle of Least Privilege by avoiding unnecessary enablement of allow-scripts.

### Additional details

Minimal reproducible example:
```html
<iframe id="myIframe" sandbox="allow-same-origin" src="content.html"></iframe>
```
A same-origin parent can read or manipulate the iframe's DOM without setting allow-scripts.

Fixes #42633

### DIFF
--- a/files/en-us/web/html/reference/elements/iframe/index.md
+++ b/files/en-us/web/html/reference/elements/iframe/index.md
@@ -126,7 +126,7 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
     - `allow-same-origin`
       - : If this token is not used, the resource is treated as being from a special origin that always fails the {{Glossary("same-origin policy")}} (potentially preventing access to [data storage/cookies](/en-US/docs/Web/Security/Defenses/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs).
         > [!NOTE]
-        > When `allow-same-origin` is present, a same-origin parent document can access and interact with the iframe's DOM even if `allow-scripts` is not set. The `allow-scripts` token only controls script execution within the embedded browsing context and does not affect DOM access from the parent.
+        > When `allow-same-origin` is present, a same-origin parent document can still access and interact with the iframe's DOM even if `allow-scripts` is not set. The `allow-scripts` token only controls script execution within the embedded browsing context and does not affect DOM access from the parent.
     - `allow-scripts`
       - : Allows the page to run scripts (but not create pop-up windows). If this keyword is not used, this operation is not allowed.
     - `allow-storage-access-by-user-activation` {{experimental_inline}}


files/en-us/web/html/reference/elements/iframe/index.md
```diff
@@ -125,6 +125,8 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
       - : Allows embedders to have control over whether an iframe can start a [presentation session](/en-US/docs/Web/API/PresentationRequest).
     - `allow-same-origin`
       - : If this token is not used, the resource is treated as being from a special origin that always fails the {{Glossary("same-origin policy")}} (potentially preventing access to [data storage/cookies](/en-US/docs/Web/Security/Defenses/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs).
+        > [!NOTE]
+        > When `allow-same-origin` is present, a same-origin parent document can access and interact with the iframe's DOM even if `allow-scripts` is not set. The `allow-scripts` token only controls script execution within the embedded browsing context and does not affect DOM access from the parent.
```
pepelsbey:
```suggestion
        > When `allow-same-origin` is present, a same-origin parent document can still access and interact with the iframe's DOM even if `allow-scripts` is not set. The `allow-scripts` token only controls script execution within the embedded browsing context and does not affect DOM access from the parent.
```

Just to stress it a little more. What do you think?

PR #42663, opened 2026-01-07
Make GTK reduced motion config instructions comprehensive

webknjaz:
This patch adds a `gsettings` command that is respected by some applications better than the file config [[1]].

[1]: https://wiki.archlinux.org/title/Talk:GTK#c-Arash-20240319042700-Affaisseras-20240309232500

### Description

`gsettings set org.gnome.desktop.interface enable-animations false` is needed to cover a variaty of apps and GTK versions consistently, which this change documents.

### Motivation

When I was typing out the config, some apps didn't pick it up when I added the documented setting to the configs. 

### Additional details

It's probably because of GTK 3 vs. GTK 4 difference. But configuring it in the files *and* in `gsettings` did help immediately.

N/A

### DIFF
--- a/files/en-us/web/css/reference/at-rules/@media/prefers-reduced-motion/index.md
+++ b/files/en-us/web/css/reference/at-rules/@media/prefers-reduced-motion/index.md
@@ -26,7 +26,8 @@ For Firefox, the `reduce` request is honoured if:
 
 - In GTK/GNOME: Settings > Accessibility > Seeing > Reduced animation is turned on.
   - In older versions of GNOME, GNOME Tweaks > General tab (or Appearance, depending on version) > Animations is turned off.
-  - Alternatively, add `gtk-enable-animations = false` to the `[Settings]` block of [the GTK 3 configuration file](https://wiki.archlinux.org/title/GTK#Configuration). Additionally run `gsettings set org.gnome.desktop.interface enable-animations false` to make programs pick up the settings faster.
+  - Alternatively, add `gtk-enable-animations = false` to the `[Settings]` block of [the GTK 3 configuration file](https://wiki.archlinux.org/title/GTK#Configuration).
+  - Additionally, try running `gsettings set org.gnome.desktop.interface enable-animations false` to make Firefox (and other programs relying on GTK version 4) respect the `reduce` setting.
 
 - In Plasma/KDE: System Settings > Workspace Behavior -> General Behavior > "Animation speed" is set all the way to right to "Instant".
   - Alternatively, add `AnimationDurationFactor=0` to the `[KDE]` block of `~/.config/kdeglobals`.


files/en-us/web/css/reference/at-rules/@media/prefers-reduced-motion/index.md
```diff
@@ -26,7 +26,7 @@ For Firefox, the `reduce` request is honoured if:
 
 - In GTK/GNOME: Settings > Accessibility > Seeing > Reduced animation is turned on.
   - In older versions of GNOME, GNOME Tweaks > General tab (or Appearance, depending on version) > Animations is turned off.
-  - Alternatively, add `gtk-enable-animations = false` to the `[Settings]` block of [the GTK 3 configuration file](https://wiki.archlinux.org/title/GTK#Configuration).
+  - Alternatively, add `gtk-enable-animations = false` to the `[Settings]` block of [the GTK 3 configuration file](https://wiki.archlinux.org/title/GTK#Configuration). Additionally run `gsettings set org.gnome.desktop.interface enable-animations false` to make programms pick up the settings faster.
```
chrisdavidmills:
```suggestion
  - Alternatively, add `gtk-enable-animations = false` to the `[Settings]` block of [the GTK 3 configuration file](https://wiki.archlinux.org/title/GTK#Configuration). Additionally run `gsettings set org.gnome.desktop.interface enable-animations false` to make programs pick up the settings faster.
```

chrisdavidmills: Also, when you say "to make programs pick up the settings faster":

1. By "programs", do you mean Firefox in this context? If so, it might be less vague to specifically say "...to make Firefox...", given that the whole section is about Firefox honoring reduce settings.
2. By "pick up the settings faster", are you specifically referring to the `reduce` setting?

webknjaz: Thanks for catching the typo!

---

I mean Firefox and any other apps that attempt to respect GTK settings. I've been experimenting with a number of browsers and wasn't able to make Chrome respect the setting (not sure why, maybe I need a reboot :man_shrugging:). Firefox didn't work when I changed the config but executing the `gsettings set org.gnome.desktop.interface enable-animations false` command triggered an immediate response in that the test animations on https://webkit.org/blog-files/prefers-reduced-motion/prm.htm stopped moving and it started indicating the preference, without me restarting Firefox. This *probably* indicates that my build of Firefox (a pre-compiled binary in Gentoo's portage) likely uses GTK 4. I can imagine that somebody using GTK 3/2 etc might need that config edit (plus maybe a re-login or something like that).

https://wiki.archlinux.org/title/Talk:GTK#c-Arash-20240319042700-Affaisseras-20240309232500 that I linked in the commit message talks about GTK in general linking to https://wiki.archlinux.org/title/GTK#Basic_theme_configuration without specifying what text in there supports their point. But the commenter says "it's implied" so I assume they mean the phrase "If the theme is not applied for GTK 3, use `gsettings` in addition", plus maybe that "dark theme" section example that suggests using the config for GTK 3 but the command for GTK 4.

I have confirmed that there's three versions (2, 3 and 4) installed on my machine:
```console
$ eix -I x11-libs/gtk+
[I] x11-libs/gtk+
     Available versions:  
     (2)    2.24.33-r3^t{gpkg}
     (3)    3.24.51^t
       {+X aqua broadway cloudproviders colord cups examples gtk-doc +introspection sysprof test vim-syntax wayland xinerama ABI_MIPS="n32 n64 o32" ABI_S390="32 64" ABI_X86="32 64 x32"}
     Installed versions:  2.24.33-r3(2)^t{gpkg}(01:41:40 PM 07/12/2024)(cups introspection -aqua -examples -test -vim-syntax -xinerama ABI_MIPS="-n32 -n64 -o32" ABI_S390="-32 -64" ABI_X86="64 -32 -x32") 3.24.51(3)^t(04:17:46 PM 10/14/2025)(X colord cups introspection sysprof wayland -aqua -broadway -cloudproviders -examples -gtk-doc -test -vim-syntax -xinerama ABI_MIPS="-n32 -n64 -o32" ABI_S390="-32 -64" ABI_X86="64 -32 -x32")
     Homepage:            https://www.gtk.org/
     Description:         Gimp ToolKit +


$ eix -I gui-libs/gtk
[I] gui-libs/gtk
     Available versions:  (4) 4.18.6-r1^t ~4.20.3-r2^t
       {+X aqua broadway cloudproviders colord cups examples gstreamer gtk-doc +introspection sysprof test vulkan wayland CPU_FLAGS_X86="f16c"}
     Installed versions:  4.18.6-r1(4)^t(04:18:58 PM 10/14/2025)(X colord cups gstreamer introspection sysprof vulkan wayland -aqua -broadway -cloudproviders -examples -gtk-doc -test CPU_FLAGS_X86="-f16c")
     Homepage:            https://www.gtk.org/ https://gitlab.gnome.org/GNOME/gtk/
     Description:         GTK is a multi-platform toolkit for creating graphical user interfaces
```

So technically other programs might expect the same configuration to live in the config files rather than `gsettings` if they use older GTK versions. This may be browsers or something else on the system. I wanted the phrase to be generic enough to hint that configuring this in all the places is probably the best thing to do and would cover all the possibilities across various software supporting animation preferences. It's possible that somebody might've configured it in one of the configs in the past and forgot about it. They'd probably expect things to work when they switch between browsers (like from Chrome to Firefox, for example) and will get confused that it doesn't work without realizing that the underlying reason is unrelated to the app but to the underlying graphical framwork.

---

I'm not too attached to the phrasing, this is mostly a drive-by as I tend to fix things in docs/repos that show up at the top of search results, when I find solutions to my frustrations. I'll accept the above suggestion but do feel free to commit directly to this branch as you see fit.

chrisdavidmills: Thanks for the explanation, @webknjaz. I can see why you want to tread carefully here; this could easily get very verbose if we tried to cover every eventuality. I still think the phrasing is a bit vague considering the very specific context it is being discussed in.

Would you be happy with something like this:

> Additionally, run `gsettings set org.gnome.desktop.interface enable-animations false` to make Firefox (and potentially other programs, depending on the GTK version they rely on) respect the `reduce` setting more quickly.

or 

> Additionally, run `gsettings set org.gnome.desktop.interface enable-animations false` to make Firefox (and other programs relying on GTK version 4) respect the `reduce` setting more quickly.

?

webknjaz: Sounds good. Perhaps, I'd even go for the last one but drop the "more quickly" part since I might've misjudged the effect and it's probably more of "the only way to get it to work on GTK4 in certain programs" that I confused for responsiveness while trying out a number of changes before I learned more context.

webknjaz: Sounds good. Perhaps, I'd even go for the last one but drop the "more quickly" part since I might've misjudged the effect and it's probably more of "the only way to get it to work on GTK4 in certain programs" that I confused for responsiveness while trying out a number of changes before I learned more context.

PR #42695, opened 2026-01-08
Add description to background shorthand property

### LINKED ISSUE #42643: Add explanation that CSS background propagate from `body` to `html`
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/background

(Add a new section)

The [CSS specification](https://www.w3.org/TR/2024/CRD-css-backgrounds-3-20240311/#body-background) describes the following behavior when setting a background for the canvas.

> For documents whose root element is an HTML `HTML` element or an XHTML `html` element: if the computed value of background-image on the root element is none and its background-color is transparent, user agents must instead propagate the computed values of the background properties from that element’s first HTML `BODY` or XHTML `body` child element. The used values of that `BODY` element’s background properties are their initial values, and the propagated values are treated as if they were specified on the root element. It is recommended that authors of HTML documents specify the canvas background using the `BODY` element rather than the `HTML` element.

I thought this was an outdated description meant for Netscape Navigator 4.x. However, I recently encountered a strange bug with the “Force Dark Theme” in iOS Vivaldi.👉[Accent colors and dark mode](https://forum.vivaldi.net/topic/111578/accent-colors-and-dark-mode)

This is clearly a bug in Vivaldi, but in recent years, more browsers have added options to force the application of dark themes. Consequently, web developers are once again being forced to pay attention to background propagate.

Therefore, regardless of the Vivaldi bug, please consider adding an explanation of background propagate as defined in the CSS specification to MDN as well.

We will add the same text described in the CSS specification to the MDN page.

If possible, it would be better to include a mention of why such caution is necessary. Is there anyone who knows the Netscape Navigator 4.x era well?

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/css/reference/properties/background`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/background
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/reference/properties/background/index.md

</details>

estelle:
Creates the property description section.
move stuff that should be in description from other sections into the new section.
Add information of layer order.
add information about body bg being added to HTML bg
add containment exception
add links.

fixes #42643

### DIFF
--- a/files/en-us/web/css/reference/properties/background/index.md
+++ b/files/en-us/web/css/reference/properties/background/index.md
@@ -119,7 +119,7 @@ The `<bg-size>` value may only be included immediately after `<bg-position>`, se
 
 Each background layer can include zero, one, or two [`<visual-box>`](/en-US/docs/Web/CSS/Reference/Values/box-edge#visual-box) values. If only one value is included, it sets both {{cssxref("background-origin")}} and {{cssxref("background-clip")}}. If two values are present, the first occurrence specifies the `background-origin` and the second specifies the `background-clip` value. If no `<visual-box>` values are present, the `background-origin` defaults to `border-box` and the `background-clip` defaults to `padding-box`.
 
-While there is no order requirement for the other background properties, the following order is recommended for consistency and legibility, while remembering that none of the values are required:
+While there is no order requirement for the other background properties, the following order is recommended for consistency and legibility; remember that none of the values are required:
 
 `<bg-image> <bg-position> / <bg-size> <repeat-style> <attachment> <bg-clip> <bg-origin> <'background-color'>`
 


files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -6,7 +6,7 @@ browser-compat: css.properties.background
 sidebar: cssref
 ---
 
-The **`background`** [shorthand](/en-US/docs/Web/CSS/Guides/Cascade/Shorthand_properties) [CSS](/en-US/docs/Web/CSS) property sets all background style properties at once, such as color, image, origin and size, or repeat method. Component properties not set in the `background` shorthand property value declaration are set to their default values.
+The **`background`** [shorthand](/en-US/docs/Web/CSS/Guides/Cascade/Shorthand_properties) [CSS](/en-US/docs/Web/CSS) property sets all background style properties at once, such as color, image, origin and size, or repeat method.
```
chrisdavidmills:
```suggestion
The **`background`** [shorthand](/en-US/docs/Web/CSS/Guides/Cascade/Shorthand_properties) [CSS](/en-US/docs/Web/CSS) property sets all background style properties at once, such as color, image, origin, size, and repeat method.
```

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -110,12 +95,14 @@ The syntax of each layer is as follows:
 - `<bg-image>`
   - : See {{Cssxref("background-image")}}. Default: `none`.
 - `<bg-position>`
-  - : See {{cssxref("background-position")}}. Default: 0% 0%.
+  - : See {{cssxref("background-position")}}. Default: `0% 0%`.
```
chrisdavidmills, line 98: Just above here, `<'background-color'>` should be changed to `<background-color>` for consistency and correctness, surely? There are other instances further down too.

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
 background: repeat scroll 0% 0% / auto padding-box border-box none transparent;
 ```
 
+### Property components
+
+The `background` shorthand property enables you to declare all CSS background properties in a single declaration. The background lies underneath the content of an element; the content of an element is painted on top of it's background. When you have multiple, comma separated background values, each is background layer that is painted on top of the previous layers.
```
chrisdavidmills: The Description content seems correct, but weirdly structured. It should start with this bit, which explains what the shorthand does at a high level, and the bit you've got at the top should be put further down.

chrisdavidmills:
```suggestion
The `background` shorthand property enables you to declare all CSS background properties in a single declaration. The background lies underneath the content of an element. When you have multiple, comma-separated background values, each is a background layer that is painted on top of the previous layers.
```

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
...
+
+The `background` property is specified as one or more background layers, separated by commas. Each layer may include zero, one, or two `<visual-box>` and and zero or one `<attachment>`, `<bg-image>`, `<bg-position>`, `<bg-size>`, and `<repeat-style>`. If the `<bg-position>`, `<bg-size>`, or `<repeat-style>` components contain two values, the first value is the horizontal value and the second value is the vertical behavior. If only a single value is set, that value is applied the both vertices.
```
chrisdavidmills:
```suggestion
The `background` property is specified as one or more background layers, separated by commas. Each layer may include zero, one, or two `<visual-box>` components and zero or one `<attachment>`, `<bg-image>`, `<bg-position>`, `<bg-size>`, and `<repeat-style>` components. If two `<bg-position>`, `<bg-size>`, or `<repeat-style>` components are specified, the first value is the horizontal value and the second value is the vertical value. If only a single value is set, that value is applied to both dimensions.
```

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
...
+
+Component properties not set in the `background` shorthand property value declaration are set to their default values.
+
+The `<'background-color'>` value may only be included in the last layer specified.
```
chrisdavidmills:
```suggestion
The `<'background-color'>` component may only be included in the last background layer specified.
```

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
 background: repeat scroll 0% 0% / auto padding-box border-box none transparent;
 ```
 
+### Property components
+
+The `background` shorthand property enables you to declare all CSS background properties in a single declaration. The background lies underneath the content of an element; the content of an element is painted on top of it's background. When you have multiple, comma separated background values, each is background layer that is painted on top of the previous layers.
+
+The `background` property is specified as one or more background layers, separated by commas. Each layer may include zero, one, or two `<visual-box>` and and zero or one `<attachment>`, `<bg-image>`, `<bg-position>`, `<bg-size>`, and `<repeat-style>`. If the `<bg-position>`, `<bg-size>`, or `<repeat-style>` components contain two values, the first value is the horizontal value and the second value is the vertical behavior. If only a single value is set, that value is applied the both vertices.
+
+Component properties not set in the `background` shorthand property value declaration are set to their default values.
```
chrisdavidmills, line 112: Feels like this sentence should be moved below the one that follows it, to the bottom of the section.

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
...
+
+The `<'background-color'>` value may only be included in the last layer specified.
+
+### Component property order
+
+Because some of the component properties share value types, the order of those component properties within the shorthand is important.
+
+The `<bg-size>` value may only be included immediately after `<bg-position>`, separated with the '/' character. For example: `10px 10px / 80% 80%` means the background image that is 80% as tall and 80% as wide as the element will be position 10px from the top and 10px from the left of the top-left corner of that element.
```
chrisdavidmills:
```suggestion
The `<bg-size>` value may only be included immediately after `<bg-position>`, separated with the `/` character. For example: `10px 10px / 80% 80%` means the background image is `80%` as tall and as wide as the element, and will be positioned `10px` from the top and `10px` from the left of the element's top-left corner.
```

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
...
+
+Each background layer can include zero, one, or two [`<visual-box>`](/en-US/docs/Web/CSS/Reference/Values/box-edge#visual-box) values. If only one value is included, it sets both {{cssxref("background-origin")}} and {{cssxref("background-clip")}}. If two values are present, the first occurrence specifies the `background-origin` and the second specifies the `background-clip` value. If no `<visual-box>` values are present, the `background-origin` defaults to `border-box` and the `background-clip` defaults to `padding-box`.
+
```
chrisdavidmills, line 121: What about the other component types? Does their order not matter? What about where `<bg-size>` goes relative to the `<visual-box>` values? It would be worth clarifying.

Also, whenever I want to use the `background` property, I also end up Googling for an example to copy and paste, as it is so hard to remember how to write correctly, especially when it comes to where to include the background color. It might be worth including a couple of definitive reference examples that show the correct order for all the components, and which people can copy and paste when needed.

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
...
+### Image painting order
+
+If multiple comma-separate backgrounds are included, each creates a background layer that are laid on top of each other. The first value in the list is the top layer. If the layer provides no transparent areas, this is the only layer that will be visible.
```
chrisdavidmills:
```suggestion
If multiple comma-separated backgrounds are included, they create multiple background layers on top of one another. The first background in the list creates the top layer. If the top layer contains no transparent areas, this is the only layer that will be visible.
```

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
...
+
+The last layer is the bottom layer. The background color is always this last layer.
```
chrisdavidmills:
```suggestion
The last layer is the bottom layer. The background color is always included in this layer.
```

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
...
+
+### Body background applied to the entire document
+
+If the document root is the {{htmlelement("html")}} element and the computed value of `background-image` on the root element is `none` and its `background-color` is `transparent`, the browser will promote the `background` style of the first {{htmlelement("body")}} element onto the `:root`, and treat the `<body>` as if `background: initial` were set. In other words, the `:root` gets all the `background` styles that were set on `body`, and `<body>` element's background properties are their initial values.
```
chrisdavidmills:
```suggestion
If the document {{htmlelement("html")}} `:root` element's computed `background-image` value is `none` and its `background-color` is `transparent`, the browser will transfer the `background` styles set on the {{htmlelement("body")}} element onto the `:root` and treat the `<body>` as if `background: initial` were set. In other words, the `<html>` element gets all the `background` styles set on the `<body>` element, and the `<body>` element's background properties are set to their initial values.
```

I simplified this a bit. I don't think you need to include the bits saying _if_ the root element is `<html>`, and "first `<body>` element", as realistically we are talking about HTML here, and there should only be a single `<body>` element, and single `<html>` element that is the root. Yes, you can add multiple `<body>`s in and do weird things, but you really shouldn't.

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -124,6 +111,36 @@ background: transparent;
...
+
+Because of this behavior, the specification authors recommend setting your document's background styles in your `body` style block rather than your `html` style block. However, it's important to note that using containment disables this `body` to `html` element background promotion. When the used value of the {{cssxref("contain")}} property on either the `<html>` or `<body>` element is anything other than `none`, the `background` property and any longhand components, do not propagate from the `<body>` element to the root `<html>` element.
```
chrisdavidmills:
```suggestion
Because of this behavior, the specification authors recommend setting your document's background styles in your `body` style block rather than your `html` style block. However, it's important to note that using containment disables this behavior. When the {{cssxref("contain")}} property is set to anything other than `none` on either the `<html>` or `<body>` element, the `background` property and any longhand components do not propagate from the `<body>` element to the root `<html>` element.
```

files/en-us/web/css/reference/properties/background/index.md
estelle: I originally thought the same thing, but it's the only value that is an actual property, and the formal syntax has it this way.
formal syntax showing only the color with a single quote

chrisdavidmills: Ah, OK. Weird, but fair enough ;-)

files/en-us/web/css/reference/properties/background/index.md
```diff
@@ -110,20 +95,61 @@ The syntax of each layer is as follows:
...
+
+Each background layer can include zero, one, or two [`<visual-box>`](/en-US/docs/Web/CSS/Reference/Values/box-edge#visual-box) values. If only one value is included, it sets both {{cssxref("background-origin")}} and {{cssxref("background-clip")}}. If two values are present, the first occurrence specifies the `background-origin` and the second specifies the `background-clip` value. If no `<visual-box>` values are present, the `background-origin` defaults to `border-box` and the `background-clip` defaults to `padding-box`.
+
+While there is no order requirement for the other background properties, the following order is recommended for consistency and legibility, while remembering that none of the values are required:
```
chrisdavidmills:
```suggestion
While there is no order requirement for the other background properties, the following order is recommended for consistency and legibility; remember that none of the values are required:
```

files/en-us/web/css/reference/properties/background/index.md
chrisdavidmills: The update is superb; nice work.

Issue #38238, opened 2025-02-19
fedcm request cookies require SameSite=None


jespertheend:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/FedCM_API/IDP_integration

### What specific section or headline is this issue about?

The accounts list endpoint, The ID assertion endpoint

### What information was incorrect, unhelpful, or incomplete?

'The accounts list endpoint' does mention that requests are sent with cookies, but it doesn't mention that these cookies have to be `SameSite=None`

'The ID assertion endpoint' doesn't mention cookies at all, but the same applies here.

### What did you expect to see?

More info about which cookies are sent.

### Do you have any supporting links, references, or citations?

https://github.com/w3c-fedid/FedCM/issues/587
https://w3c-fedid.github.io/FedCM/#browser-api

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/api/fedcm_api/idp_integration`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/FedCM_API/IDP_integration
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/fedcm_api/idp_integration/index.md
* Last commit: https://github.com/mdn/content/commit/2b6f99e45534ce662f842d8b4d2f7845492e353c
* Document last modified: 2024-08-17T01:09:59.000Z

</details>

PR #42696, opened 2026-01-08
Fix issue 38238: IdP cookies must have SameSite=None

wbamberg:
Fixes https://github.com/mdn/content/issues/38238.

It looks like `SameSite=Lax` will eventually work as well, but not yet: https://github.com/httpwg/http-extensions/issues/3323.

### DIFF
--- a/files/en-us/web/api/fedcm_api/idp_integration/index.md
+++ b/files/en-us/web/api/fedcm_api/idp_integration/index.md
@@ -117,7 +117,9 @@ Cookie: 0x23223
 Sec-Fetch-Dest: webidentity
 ```
 
-The request is credentialed: that is, it includes cookies for the IdP's site, which the IdP can use to identify which IdP accounts the user is signed into. Note that because the browser's request to this endpoint is a cross-site request, cookies will only be included if they have a [`SameSite`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute value of `None`.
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/fedcm_api/idp_integration/index.md
```diff
@@ -117,7 +117,9 @@ Cookie: 0x23223
 Sec-Fetch-Dest: webidentity
 ```
 
-The response to a successful request returns a list of all the IdP accounts that the user is currently signed in with (not specific to any particular RP), with a JSON structure that matches the following:
+The request is credentialed: that is, it includes cookies for the IdP's site, which the IdP can use to identify which IdP accounts the user is signed into. Note that because the browser's request to this endpoint is a cross-site request, cookies will only be included if they have a [`SameSite`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute value of `"None"`.
```
wbamberg:
```suggestion
The request is credentialed: that is, it includes cookies for the IdP's site, which the IdP can use to identify which IdP accounts the user is signed into. Note that because the browser's request to this endpoint is a cross-site request, cookies will only be included if they have a [`SameSite`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute value of `None`.
```

wbamberg:
I wonder if we should mention that because of this, IdPs should take precautions against CSRF, like checking that `Sec-Fetch-Dest==="webidentity"`.

PR #42701, opened 2026-01-09
Use consistent lowercase wording in if...else example output

amitkumarpatra99:


amitkumarpatra99:
Thanks for the review! I’ve applied the suggested formatting fix and all checks should now pass.

PR #42719, opened 2026-01-10
Clarify CPU-bound limitations of promises in long-running functions

### LINKED ISSUE #42715: CPU bound promise operation with generatePrimes
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Async_JS/Introducing

The trouble with long-running synchronous functions

The section "[The trouble with long-running synchronous functions](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Async_JS/Introducing#the_trouble_with_long-running_synchronous_functions)" is concluded by suggesting that async programming will solve this type of specific problem. 

The issue is that it suggests that a CPU bound operation like generating primes will run concurrently if put into a promise, misleading programmers. CPU bound operations like this will still run synchronously even if promises are used. 

I've actually made this mistake before thinking it would make calculations faster, only to see it run the same or slower with promises. 

The difference between CPU bound and non bound tasks should described and why promises/async programming cannot solve CPU bound problems. It should note that workers should be the solution for that and using promises for CPU bound tasks will still run them synchronously. 

It should have a section about when to use workers vs async and how I/O based operations work better with async functions.

[**https://stackoverflow.com/questions/868568/what-do-the-terms-cpu-bound-and-i-o-bound-mean**](https://www.baeldung.com/cs/cpu-io-bound)

_No response_

### MDN metadata

<details>

* Folder: `en-us/learn_web_development/extensions/async_js/introducing`
* MDN URL: https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Async_JS/Introducing
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/learn_web_development/extensions/async_js/introducing/index.md
* Last commit: https://github.com/mdn/content/commit/0cc63ce1d7f43eb98746a908a9aba68ef6a36f7b
* Document last modified: 2025-08-26T12:44:08.000Z

</details>

kashish2710:
~~Fixes~~ #42715

This PR clarifies that wrapping CPU-bound work (such as generating prime
numbers) in a Promise or using async/await does not make it run in parallel
or prevent main-thread blocking.

It explains the distinction between CPU-bound and I/O-bound tasks and points
readers to Web Workers as the appropriate solution for heavy computation,
helping avoid a common misconception for beginners.


chrisdavidmills (changes requested):
Hi @kashish2710, and thank you for attempting to fix this. However, I think the fix is going to be a bit more complicated than this.

Some comments:

1. The content you've added is put in a weird place — it breaks the flow of the section it is added to, and doesn't really help the reader understand the issue we are trying to solve here. I think it would be better added at the end of the https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Async_JS/Promises article, in its own separate section. Mention the two different types (CPU-bound and I/O bound, say that promises are better for solving one (which we covered in this article), and workers are better for solving the other, and link to where workers are covered).
2. The `generatePrimes()` example is problematic as written — it alludes to an issue with sync code, but then the reader is not shown how to solve that issue, until the https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Async_JS/Introducing_workers article, much later. The issue poster is correct that this problem cannot be solved with promises, and workers are needed. I would state this at the end of the https://pr42719.review.mdn.allizom.net/en-US/docs/Learn_web_development/Extensions/Async_JS/Introducing#the_trouble_with_long-running_synchronous_functions section — say that if you are looking for a solution to this problem, see the workers article, and link to where you can find a solution.

kashish2710:
Thanks @chrisdavidmills for the detailed feedback — this is very helpful.
I’ll work on updating the PR to address these points.

Josh-Cena:
@kashish2710?

kashish2710:
Thanks for checking in. I’m no longer planning to continue this PR, so I’ll close it. Thank you for the review and feedback!

Josh-Cena:
No worries, thanks for following up!

PR #42729, opened 2026-01-11
Fix positioning example in using anchor positioning guide

myakura:
### Description

Corrected the description of the positioning rule to specify the left edge instead of the right edge.

### Motivation

That wrong example made me so confused while learning CSS anchor positioning

### Additional details

https://github.com/mdn/content/blob/2b9e1f24f660e3cf52539b7b8a7fc94085e8814d/files/en-us/web/css/guides/anchor_positioning/using/index.md?plain=1#L256-L263


chrisdavidmills (approved):
Thanks for submitting a fix for this, @myakura. The problem was that the margin was incorrectly set (in the original, `margin-left` should have been `margin-right`), but you've instead fixed this by changing the side we've anchored the positioned element to match the `margin-left`, which is equally valid.

(just including this info for the benefit of other folks who were curious).

PR #42730, opened 2026-01-11
Fix incorrect inheritance for Range interface

Pratik-Ugile28:
This PR fixes the incorrect inheritance shown for the Range interface
by declaring AbstractRange in the page frontmatter.

Fixes: https://github.com/mdn/content/issues/42547


wbamberg:
This front matter item doesn't exist.

PR #42757, opened 2026-01-12
Fix typo in intersection_observer_api

cblanken:
Change "memorizing"  to  "memoizing". "Memoising", with an "s" is another alternative.

### Description

Fix a typo. "Memorizing" is used incorrectly here in place of "memoizing" when referring to storing results of a computation.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API


Josh-Cena:
Sorry, "memorizing" is the correct term here, which is synonymous to "remembering" or "storing". "Memoizing" is referring to a specific computing technique of remembering previous _calculation results_ to avoid recomputation, but here we are just remembering the previous states for diffing, not optimizing. (Proof that a glossary page is in order!)

I understand that "memorizing" may be anthropomorphizing the computer a bit, which I've learned that some people don't like, so if you'd want another word I would suggest "storing".

wbamberg:
Sorry, I should have checked the context. I just assumed "memoizing" was meant because why would someone write "memorize" instead of "store".

> I understand that "memorizing" may be anthropomorphizing the computer a bit, which I've learned that some people don't like, so if you'd want another word I would suggest "storing".

Yes please!

PR #42763, opened 2026-01-12
Document that an attribute selector can also be used for this

lkmorlan:
### Description

Document that the attribute selector `[open]` can be used in place of `:open`.

### Motivation

`:open` has limited availability. This makes the difference between the two more clear suggests this alternative for those who do not think of it.

### DIFF
--- a/files/en-us/web/css/reference/selectors/_colon_open/index.md
+++ b/files/en-us/web/css/reference/selectors/_colon_open/index.md
@@ -6,7 +6,7 @@ browser-compat: css.selectors.open
 sidebar: cssref
 ---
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/reference/selectors/_colon_open/index.md
```diff
@@ -6,7 +6,7 @@ browser-compat: css.selectors.open
 sidebar: cssref
 ---
 
-The **`:open`** [CSS](/en-US/docs/Web/CSS) [pseudo-class](/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-classes) represents an element that has open and closed states, only when it is currently in the open state.
+The **`:open`** [CSS](/en-US/docs/Web/CSS) [pseudo-class](/en-US/docs/Web/CSS/Reference/Selectors/Pseudo-classes) represents an element that has open and closed states, only when it is currently in the open state. For {{htmlelement("details")}} elements, this selection can also be done using an attribute selector: `details[open]`.
```
estelle: This content is more appropriate for the description, such as line 23

lkmorlan: Done. Thanks

PR #42765, opened 2026-01-13
Soften the language around search engines and SEO impact of the HTTP …

AVGP:
On the treatment of 308 response code in search engines.

### Description

At least for Google I can say that we don't treat it as the original sentence made it sound like. I would be surprised if other search engines did, but of course I can only say this for sure for Google.
I think this sentence should be removed or at least, as I did in this PR, softened.

### Motivation

Currently, it's misleading as search engine behavior (that doesn't exist) is given as a motivation to implement this HTTP code. People might waste time and effort on making their implementation use 308 based on this sentence.


pepelsbey (approved):
Good catch! Thank you :) 

I wouldn't mind removing it, too. It seems to me like a vague SEO advice that's hard to prove by any publicly available search engine documentation.

AVGP:
Excellent, I hoped so and have now removed this sentence entirely.

Also, nice to see you again, Vadim. It's been a _long_ time since we met at...some conference :)

PR #42772, opened 2026-01-13
Technical review: Add documentation for the geolocation element

chrisdavidmills:
### Description

Chrome 144 adds support for the `<geolocation>` element; see https://chromestatus.com/feature/5125006551416832.

This PR documents the new element and it's DOM API interface, `HTMLGeolocationElement`

### DIFF
--- a/files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
+++ b/files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
@@ -16,7 +16,7 @@ You can query the {{domxref("HTMLGeolocationElement.isValid")}} property to chec
 
 ## Value
 
-The empty string (`""`) if the element dos not have an active blocker, or one of the following values (in priority order):
+The empty string (`""`) if the element does not have an active blocker, or one of the following values (in priority order):
 
 - `illegal_subframe`
   - : The `<geolocation>` element is nested inside a {{htmlelement("fencedframe")}} element.


files/en-us/web/api/geolocation_api/index.md
```diff
@@ -30,6 +30,9 @@ In both cases, the method call takes up to three arguments:
 
 For further information on Geolocation usage, read [Using the Geolocation API](/en-US/docs/Web/API/Geolocation_API/Using_the_Geolocation_API).
 
+> [!NOTE]
+> The {{htmlelement("geolocation")}} element provides an alternative mechanism for accessing and handling geolocation data that solves some of the shortcomings of the Geolocation API, such as providing a consistent UI, and a more intuitive permission gathering process.
```
hamishwillee: Because the big problem is granting a denied permission.

```suggestion
> The {{htmlelement("geolocation")}} element provides an alternative mechanism for accessing and handling geolocation data that solves some of the shortcomings of the Geolocation API, such as providing a consistent UI, and a more intuitive permission management process.
```

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+
+The `<geolocation>` element aims to fix these problems by:
+
+- Providing a declarative browser-defined control for sharing location data. In Chrome, the button features a "map pin" icon and intuitive text ("Use location" in English content).
+- Reinventing the permissions process to be more intuitive. For example, in Chrome, if the user previously denied permission to access location data, or dismissed the permission dialog without making a choice, they are free to press the button again to update their choice. In cases where they previously denied permission, subsequent dialogs will inform them that they previously didn't allow location data to be shared, and ask them whether they want to continue not allowing it, or to allow it.
```
hamishwillee: Just changing first sentence
```suggestion
- Making the permissions process more intuitive.
  For example, in Chrome, if the user previously denied permission to access location data, or dismissed the permission dialog without making a choice, they are free to press the button again to update their choice.
  In cases where they previously denied permission, subsequent dialogs will inform them that they previously didn't allow location data to be shared, and ask them whether they want to continue not allowing it, or to allow it.
```

files/en-us/web/api/geolocation_api/index.md
hamishwillee: I would also consider moving this to the end of the initial introduction - this is effectively a replacement.

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+- More obviously reflecting the user's conscious choice, and blocking possible usage that might trick the user into providing their location data unwittingly (see [`<geolocation> blocking`](#geolocation_blocking) for more information).
```
hamishwillee: I'm not all that comfortable with presenting a html feature in terms of what it replaces - this kind of why it exists is more for your explainer.  My reasoning is that once it hits MDN it doesn't need a justification. Once that justification exists it poisons how you construct the rest of the docs and later extend them. That is one reason the storage access API ended up being messy after they were extended with HTTP headers.

If it were me I would start with the stuff under 

> The `<geolocation>` element aims to fix these problems by:

Present what it does/is. It's mostly a minor sentence restructure. "The element allows an intuitive management of user permissions ..."

Then maybe a short note about the Geolocation API being the older alternative, with a number of problems

hamishwillee: PS if you really love this "as it is" then push it into its own section with a heading like "Historical context".

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+
+The `<geolocation>` element uses features of the Geolocation API in the background. By default, the browser requests location data once, as if the {{domxref("Geolocation.getCurrentPosition()")}} method was called. However, if the `watch` attribute is set to `true`, the browser constantly requests the data, as if {{domxref("Geolocation.watchPosition()")}} was called.
+
+If data is successfully retrieved, it is available in the {{domxref("HTMLGeolocationElement.position")}} property, which contains a {{domxref("GeolocationPosition")}} object. If data retrieval is unsuccessful, error information is available in the {{domxref("HTMLGeolocationElement.error")}} property, which contains a {{domxref("GeolocationPositionError")}} object.
+
+> [!NOTE]
+> For performance reasons, a maximum of three `<geolocation>` elements are allowed on any one page. If this quota is exceeded, the element functionality is disabled.
```
hamishwillee: Altogether or for subsequent elements of this type?

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Setting the button language
+
+The global [`lang`](/en-US/docs/Web/HTML/Reference/Global_attributes/lang) attribute is observed by the `<geolocation>` element to select a language for its rendered text. This means that you can set a `lang` attribute directly on the `<geolocation>` element or on one of its ancestors to tell the browser what language to use for the button label.
```
hamishwillee: Any particular reason for observed?

```suggestion
The global [`lang`](/en-US/docs/Web/HTML/Reference/Global_attributes/lang) attribute is used by the `<geolocation>` element to select a language for its rendered text. This means that you can set a `lang` attribute directly on the `<geolocation>` element or on one of its ancestors to tell the browser what language to use for the button label.
```

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+
+```html
+<geolocation>
+  <button id="fallback">Use location</button>
+</geolocation>
+```
+
+### `<geolocation>` blocking
+
+One key idea behind the design of the `<geolocation>` element is that it should reflect a user's conscious choice, and prevent bad actors from tricking users into activating it, for example via [clickjacking](/en-US/docs/Web/Security/Attacks/Clickjacking). Because of this, the browser keeps a record of so-called **blocker reasons** for each rendered element.
```
hamishwillee:
```suggestion
One key idea behind the design of the `<geolocation>` element is that it should reflect a user's conscious choice to expose position information, and prevent bad actors from tricking users into activating it, for example via [clickjacking](/en-US/docs/Web/Security/Attacks/Clickjacking). Because of this, the browser keeps a record of so-called **blocker reasons** for each rendered element.
```

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Styling restrictions
+
+The `<geolocation>` element has several constraints on the CSS styles that can be applied to it. Some of these constraints are designed to enforce fundamental accessibility, and will result in the button being deactivated if they are not adhered to. Some enforce certain values or value ranges for various properties.
+
+Any properties that are not listed in the following sub-sections, or logically equivalent to a physical property listed in the following sub-sections, are ignored when set on the `<geolocation>` element.
+
+#### Accessibility restrictions
+
+The rendered `<geolocation>` button is deactivated (meaning that pressing it will have no effect) if the following constaints are not adhered to:
```
hamishwillee: Maybe run a spell checker.

```suggestion
The rendered `<geolocation>` button is deactivated (meaning that pressing it will have no effect) if the following constraints are not adhered to:
```

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+
+- The [color contrast](/en-US/docs/Web/Accessibility/Guides/Understanding_WCAG/Perceivable/Color_contrast) ratio between {{cssxref("color")}} and {{cssxref("background-color")}} must be at least 3:1.
+- The {{cssxref("font-size")}} must not be smaller than the `small` value (in the case of keyword values), or its computed value (in the case of other value types).
+
+#### Value constraints
+
+The following CSS property value constaints are applied to the `<geolocation>` element. If an attempt is made to set these properties to values outside the listed constraints on the `<geolocation>` element, the value is adjusted to equal the constraint (in the case of an exact value constraint) or to equal to nearest computed value upper or lower bound (in the case of a range constraint).
```
hamishwillee:
```suggestion
The following CSS property value constraints are applied to the `<geolocation>` element. If an attempt is made to set these properties to values outside the listed constraints on the `<geolocation>` element, the value is adjusted to equal the constraint (in the case of an exact value constraint) or to equal to nearest computed value upper or lower bound (in the case of a range constraint).
```

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+- {{cssxref("will-change")}}
+- {{cssxref("z-index")}}
+
+## Accessibility
+
+The `<geolocation>` element has an accessible name written in the [language it is set to](#setting_the_button_language). It also has a [`role`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles) of [`button`](/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role) so that it is recognized as a button by screen readers.
+
+In addition, the `<geolocation>` element has a default [`tabindex`](/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex) value of `0`, so it behaves like a real `<button>` with respect to keyboard focus.
+
+Finally, refer back to the [Accessibility restrictions](#accessibility_restrictions) section for information on styling constaints applied to the `<geolocation>` element to enforce fundamental accessibility requirements.
```
hamishwillee:
```suggestion
Finally, refer back to the [Accessibility restrictions](#accessibility_restrictions) section for information on styling constraints applied to the `<geolocation>` element to enforce fundamental accessibility requirements.
```

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+
+## Examples
+
+### Basic usage example
+
+This example uses the `<geolocation>` element to retrieve your current location, which is printed out below the button in a {{htmlelement("p")}} element. The example also uses a regular `<button>` fallback to retrive the location data in non-supporting browsers.
```
hamishwillee:
```suggestion
This example uses the `<geolocation>` element to retrieve your current location, which is printed out below the button in a {{htmlelement("p")}} element. The example also uses a regular `<button>` fallback to retrieve the location data in non-supporting browsers.
```

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
...
+  <button id="fallback">Use location</button>
+</geolocation>
+<p id="output"></p>
+```
+
+#### JavaScript
+
+In our script, we start off by grabbing a reference to the output `<p>` element. We then detect whether the `<geolocation>` element is supported by testing `typeof HTMLGeolocationElement === "function"`:
+
+- If it is supported, we grab a reference to the `<geolocation>` element then add a {{domxref("HTMLGeolocationElement.location_event", "location")}} event listener so that, when the button is pressed and the data is retrieved, we print the (lat, long) coordinates to the output `<p>` (retrieved via the {{domxref("HTMLGeolocationElement.position", "position")}} property), or an error message if the data retrieval was unsuccessful (retrieved via the {{domxref("HTMLGeolocationElement.error", "error")}} property).
```
hamishwillee: Sentence too long. Maybe

```suggestion
- If it is supported, we first grab a reference to the `<geolocation>` element and then add a {{domxref("HTMLGeolocationElement.location_event", "location")}} event listener.
  When the button is pressed and the data is retrieved, the listener prints the (lat, long) coordinates to the output `<p>` (retrieved via the {{domxref("HTMLGeolocationElement.position", "position")}} property), or an error message if the data retrieval was unsuccessful (retrieved via the {{domxref("HTMLGeolocationElement.error", "error")}} property).
```

files/en-us/web/api/htmlgeolocationelement/index.md
```diff
@@ -0,0 +1,232 @@
...
+## Instance properties
+
+_Also inherits properties from its parent interface, {{domxref("HTMLElement")}}._
+
+- {{domxref("HTMLGeolocationElement.autolocate", "autolocate")}} {{experimental_inline}}
+  - : A boolean value indicating whether the browser should immediately request location data when the `<geolocation>` element is rendered, provided permission was previously granted. Mirrors the value of the `<geolocation>` [`autolocate`](/en-US/docs/Web/HTML/Reference/Elements/geolocation#autolocate) attribute.
+- {{domxref("HTMLGeolocationElement.error", "error")}} {{readonlyinline}} {{experimental_inline}}
+  - : A {{domxref("GeolocationPositionError")}} object representing error information, in the event of a failure to retrieve data.
+- {{domxref("HTMLGeolocationElement.initialPermissionStatus", "initialPermissionStatus")}} {{readonlyinline}} {{experimental_inline}}
+  - : An enumerated value representing the initial permission status for the `geolocation` feature, when the page first loads.
```
hamishwillee:
```suggestion
  - : An enumerated value representing the permission status for the `geolocation` feature when the page first loads.
```

files/en-us/web/api/htmlgeolocationelement/index.md
```diff
@@ -0,0 +1,232 @@
...
+## Instance methods
+
+_Inherits properties from its parent interface, {{domxref("HTMLElement")}}._
+
+## Events
+
+_Also inherits events from its parent interface, {{domxref("HTMLElement")}}._
+
+- {{domxref("HTMLGeolocationElement.location_event", "location")}} {{experimental_inline}}
+  - : Fired whenever the browser receives location data, or error information when the location data request was unsuccessful.
```
hamishwillee, line 54: I think
```suggestion
  - : Fired whenever the browser receives location data, or error information when a location data request was unsuccessful.
```

files/en-us/web/api/htmlgeolocationelement/index.md
```diff
@@ -0,0 +1,232 @@
+---
+title: HTMLGeolocationElement
+slug: Web/API/HTMLGeolocationElement
+page-type: web-api-interface
+browser-compat: api.HTMLGeolocationElement
+---
+
+{{APIRef("HTML DOM")}}
+
+The **`HTMLGeolocationElement`** interface represents the {{htmlelement("geolocation")}} element, and provides access to its properties and events.
```
hamishwillee: The accepted pattern is 

```suggestion
The **`HTMLGeolocationElement`** interface of the [Navigation API](/en-US/docs/Web/API/Navigation_API) represents the {{htmlelement("geolocation")}} element, and provides access to its properties and events.
```

files/en-us/web/api/htmlgeolocationelement/index.md
```diff
@@ -0,0 +1,232 @@
...
+- {{domxref("HTMLGeolocationElement.promptaction_event", "promptaction")}} {{experimental_inline}}
+  - : Fired whenever the user activates the `<geolocation>` element and selects an option from the resulting dialog, either to grant or deny `geolocation` permission.
+- {{domxref("HTMLGeolocationElement.promptdismiss_event", "promptdismiss")}} {{experimental_inline}}
+  - : Fired whenever the user activates the `<geolocation>` element and dismisses the resulting dialog, by pressing the "close" button or the <kbd>Esc</kbd> key.
+- {{domxref("HTMLGeolocationElement.validationstatuschange_event", "validationstatuschange")}} {{experimental_inline}}
+  - : Fired whenever the `<geolocation>` element's {{domxref("HTMLGeolocationElement.isValid", "isValid")}} value changes.
+
+## Description
+
+The `HTMLGeolocationElement` interface represents the {{htmlelement("geolocation")}} element, which creates an interactive control to allow the user to share their location data with the page in a more consistent and intuitive way than rolling your own solution via the [Geolocation API](/en-US/docs/Web/API/Geolocation_API).
```
hamishwillee: Yeah, I still find positioning an API purely by its historical context a bit odd. Maybe
```suggestion
The `HTMLGeolocationElement` interface represents the {{htmlelement("geolocation")}} element, which creates an interactive control to allow the user to share their location data with the page.
```

And you could mention " in a more consistent and intuitive way than rolling your own solution via the [Geolocation API](/en-US/docs/Web/API/Geolocation_API)." in a note below the top intro.

files/en-us/web/api/htmlgeolocationelement/index.md
```diff
@@ -0,0 +1,232 @@
...
+
+When the user activates the control, they are presented with a dialog box that asks them for permission to share their location data. If they grant permission, the browser will attempt to retrieve the user's location data using the Geolocation API in the background.
+
+By default, the browser requests location data once, as if the {{domxref("Geolocation.getCurrentPosition()")}} method was called. However, if the [`watch`](/en-US/docs/Web/HTML/Reference/Elements/geolocation#watch) attribute is set to `true`, the browser constantly requests the data, as if {{domxref("Geolocation.watchPosition()")}} was called.
```
hamishwillee: Constantly requests the data makes me wonder how often etc. Perhaps better to do like this

```suggestion
By default, the browser requests location data once, as if the {{domxref("Geolocation.getCurrentPosition()")}} method was called. However, if the [`watch`](/en-US/docs/Web/HTML/Reference/Elements/geolocation#watch) attribute is set to `true`, the browser updates the element whenever the device moves, as if {{domxref("Geolocation.watchPosition()")}} was called.
```

files/en-us/web/api/htmlgeolocationelement/index.md
```diff
@@ -0,0 +1,232 @@
...
+
+See the [`<geolocation>`](/en-US/docs/Web/HTML/Reference/Elements/geolocation#basic_usage_example) reference page for a walkthrough.
+
+### Embedded map example
+
+This example uses the `<geolocation>` element to retrieve your current location, which is plotted on a map rendered using [Leaflet JS](https://leafletjs.com/). The example also uses a regular `<button>` fallback to retrive the location data in non-supporting browsers.
+
+#### HTML
+
+We include a `<geolocation>` element with an `autolocate` attribute so that the browser will attempt to retrieve location data automatically, provided `geolocation` permission was previously granted. Inside the `<geolocation>` element we nest a {{htmlelement("button")}} fallback, which will be rendered in browsers that do not support `<geolocation>` to enable location data to be requested.
```
hamishwillee, line 93:
```suggestion
We include a `<geolocation>` element with an `autolocate` attribute so that the browser will attempt to retrieve location data automatically, provided `geolocation` permission was previously granted. Inside the `<geolocation>` element we nest a {{htmlelement("button")}} fallback, which will be rendered in browsers that do not support the `<geolocation>` element.
```

files/en-us/web/html/reference/elements/geolocation/index.md
```diff
@@ -0,0 +1,331 @@
+---
+title: "<geolocation>: The geolocation element"
+slug: Web/HTML/Reference/Elements/geolocation
+page-type: html-element
+browser-compat: html.elements.geolocation
+sidebar: htmlsidebar
+---
+
```
hamishwillee, line 8: I am not entirely sure where, but here and/or in the Web API we need at least see also on https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy/geolocation - if that is disallowed the user will never be prompted.

files/en-us/web/api/htmlgeolocationelement/htmlgeolocationelement/index.md
```diff
@@ -0,0 +1,55 @@
...
+title: "HTMLGeolocationElement: HTMLGeolocationElement() constructor"
+short-title: HTMLGeolocationElement()
+slug: Web/API/HTMLGeolocationElement/HTMLGeolocationElement
+page-type: web-api-constructor
+browser-compat: api.HTMLGeolocationElement.HTMLGeolocationElement
+---
+
+{{APIRef("HTML DOM")}}
+
+The **`HTMLGeolocationElement()`** constructor creates a new {{domxref("HTMLGeolocationElement")}} object instance.
```
hamishwillee, line 11: There is nothing wrong with this, except that most other HTMLElements don't document the constructor even though it is defined. I'm not sure if there is any kind of official convention. 

Thoughts? I'd be tempted to add a warning below this point that the constructor is not called directly.

files/en-us/web/api/htmlgeolocationelement/autolocate/index.md
```diff
@@ -0,0 +1,53 @@
...
+slug: Web/API/HTMLGeolocationElement/autolocate
+page-type: web-api-instance-property
+browser-compat: api.HTMLGeolocationElement.autolocate
+---
+
+{{APIRef("Navigation API")}}
+
+The **`autolocate`** property of the {{domxref("HTMLGeolocationElement")}} interface gets and sets a boolean value indicating whether the browser should immediately request location data when the {{htmlelement("geolocation")}} element is rendered, provided permission to use the `geolocation` feature was previously granted.
+
+It mirrors the value of the `<geolocation>` [`autolocate`](/en-US/docs/Web/HTML/Reference/Elements/geolocation#autolocate) attribute.
```
hamishwillee: Absolutely everywhere, the term used is "reflects". Don't ask me why.

```suggestion
It reflects the value of the `<geolocation>` [`autolocate`](/en-US/docs/Web/HTML/Reference/Elements/geolocation#autolocate) attribute.
```

files/en-us/web/api/htmlgeolocationelement/autolocate/index.md
```diff
@@ -0,0 +1,53 @@
...
+title: "HTMLGeolocationElement: autolocate property"
+short-title: autolocate
+slug: Web/API/HTMLGeolocationElement/autolocate
+page-type: web-api-instance-property
+browser-compat: api.HTMLGeolocationElement.autolocate
+---
+
+{{APIRef("Navigation API")}}
+
+The **`autolocate`** property of the {{domxref("HTMLGeolocationElement")}} interface gets and sets a boolean value indicating whether the browser should immediately request location data when the {{htmlelement("geolocation")}} element is rendered, provided permission to use the `geolocation` feature was previously granted.
```
hamishwillee: Very nit, but this covers the case where the feature was previously granted and then revoked. If changed here, should be changed in other places this is mentioned.

```suggestion
The **`autolocate`** property of the {{domxref("HTMLGeolocationElement")}} interface gets and sets a boolean value indicating whether the browser should immediately request location data when the {{htmlelement("geolocation")}} element is rendered, provided permission to use the `geolocation` feature is already granted.
```

files/en-us/web/api/htmlgeolocationelement/autolocate/index.md
```diff
@@ -0,0 +1,53 @@
...
+
+```html
+<geolocation autolocate></geolocation>
+```
+
+```js
+const geo = document.querySelector("geolocation");
+console.log(geo.autolocate); // true
+geo.autolocate = false;
+console.log(geo.autolocate); // false
```
hamishwillee: To show it mirrors/reflects you'd need to log the attribute, not the mirrored property that you just set.  But of course depends on what you wanted to show here.

files/en-us/web/api/htmlgeolocationelement/initialpermissionstatus/index.md
```diff
@@ -0,0 +1,111 @@
...
+title: "HTMLGeolocationElement: initialPermissionStatus property"
+short-title: initialPermissionStatus
+slug: Web/API/HTMLGeolocationElement/initialPermissionStatus
+page-type: web-api-instance-property
+browser-compat: api.HTMLGeolocationElement.initialPermissionStatus
+---
+
+{{APIRef("Navigation API")}}
+
+The **`initialPermissionStatus`** read-only property of the {{domxref("HTMLGeolocationElement")}} interface returns an enumerated value representing the initial permission status for the `geolocation` feature, when the page first loads.
```
hamishwillee: Up to you, but either initial position, or "when the page first loads" - both feels redundant.

```suggestion
The **`initialPermissionStatus`** read-only property of the {{domxref("HTMLGeolocationElement")}} interface returns an enumerated value representing the permission status for the `geolocation` feature when the page first loads.
```

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
```diff
@@ -0,0 +1,211 @@
...
+slug: Web/API/HTMLGeolocationElement/invalidReason
+page-type: web-api-instance-property
+browser-compat: api.HTMLGeolocationElement.invalidReason
+---
+
+{{APIRef("Navigation API")}}
+
+The **`invalidReason`** read-only property of the {{domxref("HTMLGeolocationElement")}} interface returns an enumerated value representing the reason why the associated {{htmlelement("geolocation")}} element is invalid (blocked), if that is the case.
+
+When a [blocker](/en-US/docs/Web/HTML/Reference/Elements/geolocation#geolocation_blocking) is active on a `<geolocation>` element, it is prevented from functioning (invalid), either temporarily or permanently, depending on the reason.
```
hamishwillee: Maybe.
```suggestion
When a [blocker](/en-US/docs/Web/HTML/Reference/Elements/geolocation#geolocation_blocking) is active on a `<geolocation>` element it is invalid: prevented from functioning, either temporarily or permanently, depending on the reason.
```

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
```diff
@@ -0,0 +1,211 @@
...
+
+If the `<geolocation>` element has one or more blockers active, it is invalid, and the `invalidReason` value is one of the following:
+
+1. `illegal_subframe` (Permanent): The `<geolocation>` element is nested inside a {{htmlelement("fencedframe")}} element.
+2. `unsuccesful_registration` (Temporary): More than three `<geolocation>` elements have been inserted into the same document.
+3. `recently_attached` (Expiring): The `<geolocation>` element has only recently been attached to the DOM.
+4. `intersection_changed` (Expiring): The `<geolocation>` element is being moved.
+5. `intersection_out_of_viewport_or_clipped` (Temporary): The `<geolocation>` element is rendered outside or partially inside the viewport.
+6. `intersection_occluded_or_distorted` (Temporary): The `<geolocation>` element is rendered completely inside the viewport, but it is obscured in some way, for example rendered behind other content.
+7. `style_invalid` (Temporary): The `<geolocation>` element has some restricted styles applied to it (see [Styling restrictions](/en-US/docs/Web/HTML/Reference/Elements/geolocation#styling_restrictions)).
```
hamishwillee: This is clear, but the "standard" is to do a definition list, and I would do that here too - including `""` for the active state.

Reason is that these then become linkable.

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
```diff
@@ -0,0 +1,211 @@
...
+
+Note that these invalid reasons are in priority order, from higest to lowest. If multiple blockers are active, the `invalidReason` value returned will be the highest-priority value in the list.
+
+Also note that each invalid reason has a **blocker type** listed, which is one of the following:
```
hamishwillee: You mean that the string you get out would be:

> illegal_subframe (Permanent)

And not

> illegal_subframe 

I would include the string in the definition list definition and make this "Note that the reason includes a blocker type"

If something else is happening here we should explain how you get this blocker type.

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
```diff
@@ -0,0 +1,211 @@
...
+    reasonElem.textContent = `<geolocation> is valid`;
+  } else {
+    reasonElem.textContent = `Invalid reason: ${geo.invalidReason}`;
+  }
+});
+```
+
+#### Result
+
+See this code [running live](https://mdn.github.io/dom-examples/geolocation-element/exploring-invalid-reasons/) (also see the full [source code](https://github.com/mdn/dom-examples/tree/main/geolocation-element/exploring-invalid-reasons)). Try choosing the different invalidation options to see which invalidation reasons are reported in each case.
```
hamishwillee, line 223: FMI Why are all the examples run on github and not in-page?

files/en-us/web/api/htmlgeolocationelement/permissionstatus/index.md
```diff
@@ -0,0 +1,73 @@
...
+slug: Web/API/HTMLGeolocationElement/permissionStatus
+page-type: web-api-instance-property
+browser-compat: api.HTMLGeolocationElement.permissionStatus
+---
+
+{{APIRef("Navigation API")}}
+
+The **`permissionStatus`** read-only property of the {{domxref("HTMLGeolocationElement")}} interface returns an enumerated value representing the current permission status for the `geolocation` feature.
+
+If you want to access the initial permission status for the `geolocation` feature as it was when the page first loaded, use the {{domxref("HTMLGeolocationElement.initialPermissionStatus")}} property.
```
hamishwillee:
```suggestion
If you want to access the initial permission status for the `geolocation` feature as it was when the page first loaded, use the {{domxref("HTMLGeolocationElement.initialPermissionStatus", "initialPermissionStatus")}} property.
```

files/en-us/web/api/htmlgeolocationelement/position/index.md
```diff
@@ -0,0 +1,50 @@
...
+title: "HTMLGeolocationElement: position property"
+short-title: position
+slug: Web/API/HTMLGeolocationElement/position
+page-type: web-api-instance-property
+browser-compat: api.HTMLGeolocationElement.position
+---
+
+{{APIRef("Navigation API")}}
+
+The **`position`** read-only property of the {{domxref("HTMLGeolocationElement")}} interface returns a {{domxref("GeolocationPosition")}} object representing the user's current position, in the event of successful location data retrieval.
```
hamishwillee: AFAIK it represents the element's position and or error, which may or may not be up to date. 

Whether it is the users position depends on when you read it, or if you have read it at all. It could have failed, or you could have read it once, or whatever. 

I think you should say that, and note that to read the users current position you need to read this after the `location` event has triggered

files/en-us/web/api/htmlgeolocationelement/position/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Value
+
+A {{domxref("GeolocationPosition")}} object, or `null` if location data retrieval failed.
```
hamishwillee: Or `null` if data has not yet been fetched (or fetching failed).

files/en-us/web/api/htmlgeolocationelement/watch/index.md
```diff
@@ -0,0 +1,92 @@
...
+slug: Web/API/HTMLGeolocationElement/watch
+page-type: web-api-instance-property
+browser-compat: api.HTMLGeolocationElement.watch
+---
+
+{{APIRef("Navigation API")}}
+
+The **`watch`** property of the {{domxref("HTMLGeolocationElement")}} interface gets and sets a boolean value indicating whether the browser should request location data continuously, whenever the position of the user's device changes, or only once.
+
+It mirrors the value of the `<geolocation>` [`watch`](/en-US/docs/Web/HTML/Reference/Elements/geolocation#watch) attribute.
```
hamishwillee: mirrors > reflects everywhere.

files/en-us/web/api/htmlgeolocationelement/watch/index.md
```diff
@@ -0,0 +1,92 @@
...
+title: "HTMLGeolocationElement: watch property"
+short-title: watch
+slug: Web/API/HTMLGeolocationElement/watch
+page-type: web-api-instance-property
+browser-compat: api.HTMLGeolocationElement.watch
+---
+
+{{APIRef("Navigation API")}}
+
+The **`watch`** property of the {{domxref("HTMLGeolocationElement")}} interface gets and sets a boolean value indicating whether the browser should request location data continuously, whenever the position of the user's device changes, or only once.
```
hamishwillee: Maybe
```suggestion
The **`watch`** property of the {{domxref("HTMLGeolocationElement")}} interface gets and sets a boolean value indicating whether the browser should update request location data continuously, in order to update when the position of the user's device changes, or only once.
```

files/en-us/web/api/htmlgeolocationelement/watch/index.md
```diff
@@ -0,0 +1,92 @@
...
+
+```html
+<geolocation watch></geolocation>
+```
+
+```js
+const geo = document.querySelector("geolocation");
+console.log(geo.watch); // true
+geo.watch = false;
+console.log(geo.watch); // false
```
hamishwillee: As before - perhaps read the attribute if the goal is to show this reflects the attribute.

files/en-us/web/api/htmlgeolocationelement/location_event/index.md
```diff
@@ -0,0 +1,60 @@
...
+title: "HTMLGeolocationElement: location event"
+short-title: location
+slug: Web/API/HTMLGeolocationElement/location_event
+page-type: web-api-event
+browser-compat: api.HTMLGeolocationElement.location_event
+---
+
+{{APIRef("HTML DOM")}}
+
+The **`location`** event of the {{domxref("HTMLGeolocationElement")}} interface is fired whenever the browser receives location data, or error information when the location data request was unsuccessful.
```
hamishwillee: Again, I _think_ "a" because at this point there is no obvious association with a specific location data request.

```suggestion
The **`location`** event of the {{domxref("HTMLGeolocationElement")}} interface is fired whenever the browser receives location data, or error information when a location data request was unsuccessful.
```

files/en-us/web/api/geolocation_api/index.md
chrisdavidmills: Sure. done, and done.

files/en-us/web/html/reference/elements/geolocation/index.md
chrisdavidmills: Good points here. I've restructured it as suggested. I've contained the stuff about the Geolocation API in a specific "Relationship with the Geolocation API" section, which says a little bit about it being an older option with shortcomings, but then also says what the relationship is. 

I needed to do a bit more reorganization to avoid continuity issues.

files/en-us/web/html/reference/elements/geolocation/index.md
chrisdavidmills: Altogether? I don't think I fully understand the question; I am not sure what "subsequent elements of this type" means. Is a change needed here?

files/en-us/web/api/htmlgeolocationelement/index.md
chrisdavidmills: Most of the HTML element interfaces don't follow that pattern, but I guess we should, where possible. I've added in:

> of the [HTML DOM API](/en-US/docs/Web/API/HTML_DOM_API)

files/en-us/web/api/htmlgeolocationelement/index.md
chrisdavidmills: done, and done.

files/en-us/web/api/htmlgeolocationelement/index.md
chrisdavidmills: I ended up updating it to:

> However, if the [`watch`](/en-US/docs/Web/HTML/Reference/Elements/geolocation#watch) attribute is set to `true`, the browser updates the data whenever the device position changes, as if {{domxref("Geolocation.watchPosition()")}} was called.

I also made the equivalent change on the `<geolocation>` element page.

files/en-us/web/html/reference/elements/geolocation/index.md
chrisdavidmills: I have added a link to it from the "See also" sections of both the `<geolocation>` and `HTMLGeolocationElement` pages.

files/en-us/web/api/htmlgeolocationelement/htmlgeolocationelement/index.md
chrisdavidmills: I've added a line at the end of the constructor description as follows:

> Note that this constructor is not called directly, but via a DOM method such as {{domxref("Document.createElement()")}}.

files/en-us/web/api/htmlgeolocationelement/autolocate/index.md
chrisdavidmills: Hrm, I was trying to show the setting as well as the getting, but you are right — this doesn't really prove anything WRT setting.

I have just removed the last two lines of the example, here and on the `watch` property page.

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
chrisdavidmills: Reasonable, but I felt like it needed expanding a bit. I ended up with:

> When a [blocker](/en-US/docs/Web/HTML/Reference/Elements/geolocation#geolocation_blocking) is active on a `<geolocation>` element, it is invalid: This means that it is prevented from functioning, either temporarily or permanently, depending on the reason.

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
chrisdavidmills: I did consider this, but I ended up with a numbered list because they are in strict priority order. It is also harder to read them in DL format, next to the previous two sentences.

I tried a few different things, and ended up putting them as a DL, with numbers included. Let me know what you think ;-)

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
chrisdavidmills: Dammnit, I was hoping that this wouldn't be confusing, but it obviously is. I've tried stating the blocker type in the DD rather than as part of the DT. Any better?

If not, maybe I should structure this as a table instead.

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
chrisdavidmills: Because there was something weird going on: Whenever I tried to run one of these examples as a live sample, they didn't work, and the browser console gave me an error telling me I was falling foul of the "only three `<geolocation>` elements allowed on any one page" restriction. Even when there was only one live sample on the page, and it featured one `<geolocation>` element.

So I gave up and put them on dom-examples.

files/en-us/web/api/htmlgeolocationelement/position/index.md
chrisdavidmills: I've added some information. This is mostly correct, except that `position` only represents position data. Error data is represented by the `error` property. My additions to both of these pages should hopefully make this clear.

files/en-us/web/api/htmlgeolocationelement/position/index.md
chrisdavidmills: I ened up with

> A {{domxref("GeolocationPosition")}} object, or `null` if location data retrieval failed or data has not yet been retrieved.

files/en-us/web/api/htmlgeolocationelement/watch/index.md
chrisdavidmills: updated

files/en-us/web/api/htmlgeolocationelement/watch/index.md
chrisdavidmills: I played with this a bit and ended up with

> ...a boolean value indicating whether the browser should continuously update the user's location data whenever the position of their device changes, or only retrieve it once.

files/en-us/web/api/htmlgeolocationelement/watch/index.md
chrisdavidmills: As before, I just removed the last two lines. I don't think they are very useful anyway.

files/en-us/web/api/htmlgeolocationelement/index.md
hamishwillee: Thanks.

files/en-us/web/html/reference/elements/geolocation/index.md
hamishwillee: I mean if you have three geolocation elements and you add one more, will the first three still work, or will all four not work. The way I read it now all four will not work in this case.

files/en-us/web/html/reference/elements/geolocation/index.md
hamishwillee: Really like what you did here.

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
hamishwillee: I would have said you did not need to explicitly add a number to order them - you just need a very clear note at the end, which you have. I'll propose some text.

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
hamishwillee: It is only confusing because of this sentence. Otherwise I would probably have made the correct assumption. Probably better as 

> Also note that the descriptions above includes a "blocker type" for each invalid reason, which is one of the following:"

I have proposed here https://github.com/mdn/content/pull/42772/files#r2756284510

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
```diff
@@ -0,0 +1,217 @@
...
+  - 5: `intersection_out_of_viewport_or_clipped`
+    - : The `<geolocation>` element is rendered outside or partially inside the viewport. Temporary blocker.
+  - 6: `intersection_occluded_or_distorted`
+    - : The `<geolocation>` element is rendered completely inside the viewport, but it is obscured in some way, for example rendered behind other content. Temporary blocker.
+  - 7: `style_invalid`
+    - : The `<geolocation>` element has some restricted styles applied to it (see [Styling restrictions](/en-US/docs/Web/HTML/Reference/Elements/geolocation#styling_restrictions)). Temporary blocker.
+
+Note that these invalid reasons are in priority order, from higest to lowest. If multiple blockers are active, the `invalidReason` value returned will be the highest-priority value in the list.
+
+Also note that each invalid reason has a **blocker type** listed, which is one of the following:
```
hamishwillee: This is my proposal. Note that i

```suggestion
The empty string (`""`) if the element dos not have an active blocker, or one of the following values (in priority order):

  - `illegal_subframe`
    - : The `<geolocation>` element is nested inside a {{htmlelement("fencedframe")}} element.

      Permanent blocker.
  - `unsuccesful_registration`
    - : More than three `<geolocation>` elements have been inserted into the same document.

      Temporary blocker.
  - `recently_attached`
    - : The `<geolocation>` element has only recently been attached to the DOM.

      Expiring blocker.
  - `intersection_changed`
    - : The `<geolocation>` element is being moved.

      Expiring blocker.
  - `intersection_out_of_viewport_or_clipped`
    - : The `<geolocation>` element is rendered outside or partially inside the viewport.
    
      Temporary blocker.
  - `intersection_occluded_or_distorted`
    - : The `<geolocation>` element is rendered completely inside the viewport, but it is obscured in some way, for example rendered behind other content.

      Temporary blocker.
  - `style_invalid`
    - : The `<geolocation>` element has some restricted styles applied to it (see [Styling restrictions](/en-US/docs/Web/HTML/Reference/Elements/geolocation#styling_restrictions)).

      Temporary blocker.

These invalid reasons are listed in priority order, from higest to lowest.
If multiple blockers are active, the `invalidReason` value returned will be the highest-priority value in the list.

Also note that the descriptions above includes a "blocker type" for each invalid reason, which is one of the following:
```

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
hamishwillee: Proposal https://github.com/mdn/content/pull/42772/files#r2756284510

hamishwillee (commented):
Nicely done.  See last couple of comments.

files/en-us/web/html/reference/elements/geolocation/index.md
chrisdavidmills: OK, I understand what you mean now. Just to make this completely clear, I've updated it to:

> For performance reasons, a maximum of three `<geolocation>` elements are allowed on any one page. If this quota is exceeded, all `<geolocation>` elements have their functionality disabled.

As an FYI, "have their functionality disabled" might seem like you overkill, and you might think "are disabled" would be more suitable. However, they are not disabled in the conventional sense (i.e. the effect of the `disabled` attribute): the buttons can still be pressed, but the location data is not retrieved.

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
chrisdavidmills: Cool, this works. I've mostly used your suggestion, with some tweaks.

I wanted to put the listed blocker for each reason on a separate line, but couldn't figure out how to do that. Now I know ;-)

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
chrisdavidmills: Perfect. Updated.

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
chrisdavidmills: Updated

files/en-us/web/api/htmlgeolocationelement/invalidreason/index.md
```diff
@@ -16,32 +16,50 @@ You can query the {{domxref("HTMLGeolocationElement.isValid")}} property to chec
...
-    - : The `<geolocation>` element is rendered outside or partially inside the viewport. Temporary blocker.
-  - 6: `intersection_occluded_or_distorted`
-    - : The `<geolocation>` element is rendered completely inside the viewport, but it is obscured in some way, for example rendered behind other content. Temporary blocker.
-  - 7: `style_invalid`
-    - : The `<geolocation>` element has some restricted styles applied to it (see [Styling restrictions](/en-US/docs/Web/HTML/Reference/Elements/geolocation#styling_restrictions)). Temporary blocker.
-
-Note that these invalid reasons are in priority order, from higest to lowest. If multiple blockers are active, the `invalidReason` value returned will be the highest-priority value in the list.
-
-Also note that each invalid reason has a **blocker type** listed, which is one of the following:
+The empty string (`""`) if the element dos not have an active blocker, or one of the following values (in priority order):
```
hamishwillee:
```suggestion
The empty string (`""`) if the element does not have an active blocker, or one of the following values (in priority order):
```

PR #42783, opened 2026-01-15
Align DELETE request body semantics with GET per RFC 9110

### LINKED ISSUE #42511: Body and semantic information differ between GET and DELETE requests
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/DELETE

Introductory information, request body

RFC 9110 provides the exact same semantic warning for both GET and DELETE methods. However, I found large differences between how the [page for the GET method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET) and the [page for the DELETE method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/DELETE) document the undefined semantics.

When compared to the page for the GET method, the page for the DELETE method:
- Is missing a highlighted warning about the semantics of the method being undefined. 
- Initially states requests should not have a body (`The DELETE method has no defined semantics for the message body, so this should be empty`) but then specifies "May" for "Request has body", in the table right below.

I expect the information for the GET and DELETE methods to be the exact same (or at least very similar):
- A highlighted warning that DELETE method requests should not have a body.
- "No" for the information "Request has body" in the table at the top of the page.

The RFC 9110 semantic specifications for the GET and DELETE methods are the exact same.

Section [9.3.1. GET](https://www.rfc-editor.org/rfc/rfc9110.html#name-get), 6th paragraph:
> Although request message framing is independent of the method used, content received in a GET request has no generally defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack ([Section 11.2](https://www.rfc-editor.org/rfc/rfc9112#section-11.2) of [[HTTP/1.1](https://www.rfc-editor.org/rfc/rfc9110.html#HTTP11)]). A client SHOULD NOT generate content in a GET request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported. An origin server SHOULD NOT rely on private agreements to receive content, since participants in HTTP communication are often unaware of intermediaries along the request chain.

Section [9.3.5. DELETE](https://www.rfc-editor.org/rfc/rfc9110.html#name-delete), 5th paragraph:
> Although request message framing is independent of the method used, content received in a DELETE request has no generally defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack ([Section 11.2](https://www.rfc-editor.org/rfc/rfc9112#section-11.2) of [[HTTP/1.1](https://www.rfc-editor.org/rfc/rfc9110.html#HTTP11)]). A client SHOULD NOT generate content in a DELETE request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported. An origin server SHOULD NOT rely on private agreements to receive content, since participants in HTTP communication are often unaware of intermediaries along the request chain.

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/http/reference/methods/delete`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/DELETE
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/reference/methods/delete/index.md
* Last commit: https://github.com/mdn/content/commit/ad5b5e31f81795d692e66dadb7818ba8b220ad15
* Document last modified: 2025-07-04T01:10:07.000Z

</details>

MUSKANNISHAD:
Fixes #42511

Aligns DELETE request body documentation with GET semantics per RFC 9110.
Adds a warning callout and corrects the "Request has body" table entry.

### DIFF
--- a/files/en-us/web/http/reference/methods/delete/index.md
+++ b/files/en-us/web/http/reference/methods/delete/index.md
@@ -9,7 +9,7 @@ sidebar: http
 
 The **`DELETE`** HTTP method asks the server to delete a specified resource.
 
-Requests using `DELETE` may include a body, but it has no defined semantics.
+Requests using `DELETE` should only be used to delete data and shouldn't contain a body.
 
 > [!NOTE]
 > The semantics of sending a message body in `DELETE` requests are undefined.


pepelsbey:
Thank you! Looks good to me, but just to double-check:

@fdx-vitor-santos does it solve the issue you filed? https://github.com/mdn/content/issues/42511

files/en-us/web/http/reference/methods/delete/index.md
```diff
@@ -9,13 +9,16 @@ sidebar: http
 
 The **`DELETE`** HTTP method asks the server to delete a specified resource.
 
-The `DELETE` method has no defined semantics for the message body, so this should be empty.
+Although a request body is technically allowed, its semantics are undefined for DELETE.
+
+> [!WARNING]
+> Although request message framing is independent of the method used, content received in a DELETE request has no generally defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request.
```
wbamberg: Please don't just copy text out of the spec. It looks like the report (https://github.com/mdn/content/issues/42511) expects the page for DELETE to use the same wording as the page for GET, and that seems like a reasonable approach.

wbamberg:
Oops, sorry @pepelsbey , I missed your review.

files/en-us/web/http/reference/methods/delete/index.md
MUSKANNISHAD: Thanks for the feedback!
I’ve updated the DELETE page to use the same NOTE wording and links as the GET page, rather than copying text directly from the spec.
Please let me know if this now addresses the concern.

fdx-vitor-santos:
Many thanks for taking a look at this!

@pepelsbey yes, this PR solves the issue I had raised.

files/en-us/web/http/reference/methods/delete/index.md
```diff
@@ -9,13 +9,17 @@ sidebar: http
 
 The **`DELETE`** HTTP method asks the server to delete a specified resource.
 
-The `DELETE` method has no defined semantics for the message body, so this should be empty.
+Requests using `DELETE` may include a body, but it has no defined semantics.
```
wbamberg:
```suggestion
Requests using `DELETE` should only be used to delete data and shouldn't contain a body.
```

("it" in the original is ungrammatical and unclear, and "no defined semantics" makes the subsequent note repetitive. This suggestion uses the same wording as for GET, which seems to make sense since as the issue note the spec is identical too)

wbamberg (approved):
👍 thank you @MUSKANNISHAD !

PR #42791, opened 2026-01-16
Revise: `autofocus` attribute & property Docs

sb3nder:
### Description

- Added case where a popover is displayed.
- Added a note about URI fragment identifier, disabling autofocus for page load.
- Fix attribute link.

### DIFF
--- a/files/en-us/web/api/htmlelement/autofocus/index.md
+++ b/files/en-us/web/api/htmlelement/autofocus/index.md
@@ -10,7 +10,7 @@ browser-compat: api.HTMLElement.autofocus
 
 The **`autofocus`** property of the {{domxref("HTMLElement")}} interface represents a boolean value reflecting the [`autofocus`](/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) HTML global attribute. This indicates whether the element should be focused on page load or, if it is nested inside a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element, when the `<dialog>` or popover is shown.
 
-Only one element inside a document, or part of a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element, may have this attribute specified. If applied to multiple elements the first focusable one will receive focus.
+Only one element inside a document, `<dialog>` element, or popover may have this attribute specified. If applied to multiple elements, the first focusable one will receive focus.
 
 > [!NOTE]
 > Setting this property doesn't set the focus to the associated element: it merely tells the browser to focus to it when _the element is inserted_ in the document. Setting it after the insertion, that is most of the time after the document load, has no visible effect.
--- a/files/en-us/web/api/htmlelement/autofocus/index.md
+++ b/files/en-us/web/api/htmlelement/autofocus/index.md
@@ -8,9 +8,9 @@ browser-compat: api.HTMLElement.autofocus
 
 {{APIRef("HTML DOM")}}
 
-The **`autofocus`** property of the {{domxref("HTMLElement")}} interface represents a boolean value reflecting the [`autofocus`](/en-US/docs/Web/HTML/Reference/Elements/select#autofocus) HTML global attribute, which indicates whether the control should be focused when the page loads, or when dialog or popover become shown if specified in an element inside {{htmlelement("dialog")}} elements or elements whose popover attribute is set.
+The **`autofocus`** property of the {{domxref("HTMLElement")}} interface represents a boolean value reflecting the [`autofocus`](/en-US/docs/Web/HTML/Reference/Elements/select#autofocus) HTML global attribute, which indicates whether the element should be focused on page load, or when it is part of a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element that becomes shown.
 
-Only one form-associated element inside a document, or a {{htmlelement("dialog")}} element, or an element whose `popover` attribute is set, can have this attribute specified. If there are several, the first element with the attribute set inserted, usually the first such element on the page, gets the initial focus.
+Only one element inside a document, or part of a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element, may have this attribute specified. If applied to multiple elements the first focusable one will receive focus.
 
 > [!NOTE]
 > Setting this property doesn't set the focus to the associated element: it merely tells the browser to focus to it when _the element is inserted_ in the document. Setting it after the insertion, that is most of the time after the document load, has no visible effect.
--- a/files/en-us/web/api/htmlelement/autofocus/index.md
+++ b/files/en-us/web/api/htmlelement/autofocus/index.md
@@ -8,7 +8,7 @@ browser-compat: api.HTMLElement.autofocus
 
 {{APIRef("HTML DOM")}}
 
-The **`autofocus`** property of the {{domxref("HTMLElement")}} interface represents a boolean value reflecting the [`autofocus`](/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) HTML global attribute, which indicates whether the element should be focused on page load, or when it is part of a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element that becomes shown.
+The **`autofocus`** property of the {{domxref("HTMLElement")}} interface represents a boolean value reflecting the [`autofocus`](/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) HTML global attribute. This indicates whether the element should be focused on page load or, if it is nested inside a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element, when the `<dialog>` or popover is shown.
 
 Only one element inside a document, or part of a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element, may have this attribute specified. If applied to multiple elements the first focusable one will receive focus.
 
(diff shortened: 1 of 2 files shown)


files/en-us/web/api/htmlelement/autofocus/index.md
```diff
@@ -8,9 +8,9 @@ browser-compat: api.HTMLElement.autofocus
 
 {{APIRef("HTML DOM")}}
 
-The **`autofocus`** property of the {{domxref("HTMLElement")}} interface represents a boolean value reflecting the [`autofocus`](/en-US/docs/Web/HTML/Reference/Elements/select#autofocus) HTML global attribute, which indicates whether the control should be focused when the page loads, or when dialog or popover become shown if specified in an element inside {{htmlelement("dialog")}} elements or elements whose popover attribute is set.
+The **`autofocus`** property of the {{domxref("HTMLElement")}} interface represents a boolean value reflecting the [`autofocus`](/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) HTML global attribute, which indicates whether the element should be focused on page load, or when it is part of a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element that becomes shown.
```
chrisdavidmills:
```suggestion
The **`autofocus`** property of the {{domxref("HTMLElement")}} interface represents a boolean value reflecting the [`autofocus`](/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) HTML global attribute. This indicates whether the element should be focused on page load or, if it is nested inside a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element, when the `<dialog>` or popover is shown.
```

files/en-us/web/api/htmlelement/autofocus/index.md
```diff
@@ -8,9 +8,9 @@ browser-compat: api.HTMLElement.autofocus
...
 
-Only one form-associated element inside a document, or a {{htmlelement("dialog")}} element, or an element whose `popover` attribute is set, can have this attribute specified. If there are several, the first element with the attribute set inserted, usually the first such element on the page, gets the initial focus.
+Only one element inside a document, or part of a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element, may have this attribute specified. If applied to multiple elements the first focusable one will receive focus.
```
chrisdavidmills:
```suggestion
Only one element inside a document, `<dialog>` element, or popover may have this attribute specified. If applied to multiple elements, the first focusable one will receive focus.
```

files/en-us/web/html/reference/global_attributes/autofocus/index.md
```diff
@@ -7,17 +7,20 @@ browser-compat: html.global_attributes.autofocus
...
 
-The **`autofocus`** [global attribute](/en-US/docs/Web/HTML/Reference/Global_attributes) is a Boolean attribute indicating that an element should be focused on page load, or when the {{HTMLElement("dialog")}} that it is part of is displayed.
+The **`autofocus`** [global attribute](/en-US/docs/Web/HTML/Reference/Global_attributes) is a Boolean attribute indicating that an element should be focused on page load, or when the {{HTMLElement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element that it is part of is displayed.
 
 ```html
 <input name="q" autofocus />
 ```
 
-No more than one element in the document or dialog may have the autofocus attribute. If applied to multiple elements the first one will receive focus.
+Only one element in the document, or part of a dialog or popover element, may have this attribute specified. If applied to multiple elements the first focusable one will receive focus.
```
chrisdavidmills:
```suggestion
Only one element inside a document, `<dialog>` element, or popover may have this attribute specified. If applied to multiple elements, the first focusable one will receive focus.
```

files/en-us/web/html/reference/global_attributes/autofocus/index.md
```diff
@@ -7,17 +7,20 @@ browser-compat: html.global_attributes.autofocus
 sidebar: htmlsidebar
 ---
 
-The **`autofocus`** [global attribute](/en-US/docs/Web/HTML/Reference/Global_attributes) is a Boolean attribute indicating that an element should be focused on page load, or when the {{HTMLElement("dialog")}} that it is part of is displayed.
+The **`autofocus`** [global attribute](/en-US/docs/Web/HTML/Reference/Global_attributes) is a Boolean attribute indicating that an element should be focused on page load, or when the {{HTMLElement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element that it is part of is displayed.
```
chrisdavidmills:
```suggestion
The **`autofocus`** [global attribute](/en-US/docs/Web/HTML/Reference/Global_attributes) is a Boolean attribute indicating whether the element should be focused on page load or, if it is nested inside a {{htmlelement("dialog")}} or [`popover`](/en-US/docs/Web/HTML/Reference/Global_attributes/popover) element, when the `<dialog>` or popover is shown.
```

files/en-us/web/html/reference/global_attributes/autofocus/index.md
```diff
@@ -7,17 +7,20 @@ browser-compat: html.global_attributes.autofocus
...
 ```
 
-No more than one element in the document or dialog may have the autofocus attribute. If applied to multiple elements the first one will receive focus.
+Only one element in the document, or part of a dialog or popover element, may have this attribute specified. If applied to multiple elements the first focusable one will receive focus.
 
 > [!NOTE]
 > The `autofocus` attribute applies to all elements, not just form controls. For example, it might be used on a [contenteditable](/en-US/docs/Web/HTML/Reference/Global_attributes/contenteditable) area.
 
+> [!NOTE]
+> On page load, when a [URI fragment](/en-US/docs/Web/URI/Reference/Fragment) points to a focusable element, that element usually receives focus instead of the one indicated by the `autofocus` attribute.
```
chrisdavidmills:
```suggestion
> On page load, if a [URI fragment](/en-US/docs/Web/URI/Reference/Fragment) is specified that points to a focusable element, _and_ an `autofocus` attribute is set on a different focusable element, the element pointed to by the URI fragment usually receives the focus.
```

chrisdavidmills: Very interesting; I didn't know this.

One small point — you say "usually". In what cases is this not true?

chrisdavidmills (changes requested):
Seems like a reasonable update; thanks @sb3nder.

I've got a few suggestions and comments for you to look through.

files/en-us/web/html/reference/global_attributes/autofocus/index.md
sb3nder: From what I can see from testing:
- for a page load it’s true.
- for a regular reload, focus doesn’t go to either `:target` or `autofocus`, so in that case it’s false.
- for a page reload by pressing Enter in the address bar or using Ctrl+F5 it’s also true, I found this related issue https://github.com/whatwg/html/issues/10597 .
- using <iframe>, fragments and autofocus at the same time, give me results that aren’t interoperable, so there may still be some spec work needed on fragment behavior in this case.

What I found in the spec:
[no autofocus when `:target` exists](https://html.spec.whatwg.org/multipage/interaction.html#the-autofocus-attribute:~:text=If%20topDocument%27s%20focused%20area%20is%20not%20topDocument%20itself%2C%20or%20topDocument%20has%20non%2Dnull%20target%20element%2C%20then%3A)
[scroll to fragment when not reloading](https://html.spec.whatwg.org/multipage/nav-history-apis.html#navigate-event-scroll-focus:scroll-to-the-fragment-identifier:~:text=Otherwise%2C%20scroll%20to%20the%20fragment%20given%20document.)->[focus fragment](https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-the-fragment-identifier:~:text=Run%20the%20focusing%20steps%20for%20target%2C)

chrisdavidmills: OK, so this isn't as simple as I'd hoped. In that case, I think you should say something to qualify the "usually", but I'd be happy for you to gloss over it a bit. Maybe just a single sentence added to the paragraph, along the lines of "For example, this behavior is true when ... , but in some cases such as ... , it is false."

chrisdavidmills (changes requested):
Thanks for the continued work on this, @sb3nder. I've provided a suggestion of what to do about the remaining issue, but I think this is very close to complete now.

sb3nder:
Ok, I removed the part about the target element needing to be focusable (since it doesn’t) and simplified the note. Please take a look.

files/en-us/web/html/reference/global_attributes/autofocus/index.md
```diff
@@ -7,17 +7,21 @@ browser-compat: html.global_attributes.autofocus
...
 
-No more than one element in the document or dialog may have the autofocus attribute. If applied to multiple elements the first one will receive focus.
+Only one element inside a document, `<dialog>` element, or popover may have this attribute specified. If applied to multiple elements, the first focusable one will receive focus.
 
 > [!NOTE]
 > The `autofocus` attribute applies to all elements, not just form controls. For example, it might be used on a [contenteditable](/en-US/docs/Web/HTML/Reference/Global_attributes/contenteditable) area.
 
+> [!NOTE]
+> On page load, if a [URI fragment identifier](/en-US/docs/Web/URI/Reference/Fragment#fragment) is specified and identifies an element,
+> the element with the `autofocus` attribute does not receive focus via the `autofocus` attribute.
```
chrisdavidmills: As you've written this now, the note explains what _doesn't_ happen, but not what does happen. It really feels like it could do with an extra sentence on the end, along the lines of "Generally, the element referenced by the fragment receives focus instead."

Given the behavior your testing pointed to, I appreciate this might be seen as somewhat vague, but perhaps we should just go with something like this for now, in the interests of getting this merged. We can always modify it again later if new comments are made on it.

chrisdavidmills (changes requested):
@sb3nder sorry to keep labouring the point on this, but I had one more comment on that note.

files/en-us/web/html/reference/global_attributes/autofocus/index.md
sb3nder: Done

chrisdavidmills (approved):
Awesome, let's get this one merged. Thanks a lot for all your work on this, @sb3nder.

sb3nder:
@chrisdavidmills Thank you for your guidance.

Issue #42767, opened 2026-01-13
[CSS] Firefox 147 'anchor-center' slider demo isn't working


ghiscoding:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/position-anchor#examples

### What specific section or headline is this issue about?

anchor-name on input slider thumb

### What information was incorrect, unhelpful, or incomplete?

The new Firefox 147 is supposed to have position-anchor all implemented but 1 of the demo with the input slider isn't working as it should. The tooltip is supposed to follow the slider thumb but it doesn't. The demo uses specific Chromium code `input::-webkit-slider-thumb` 

### What did you expect to see?

I would expect the tooltip to follow the input slider thumb.

I tried changing the Chromium specific code to make it work in Firefox too but that still doesn't work:

```css
input::-webkit-slider-thumb,
input::-moz-range-thumb {
  anchor-name: --thumb;
}
```

<img width="1510" height="675" alt="Image" src="https://github.com/user-attachments/assets/a5c739d3-3eb7-4bde-9236-ee8303b6df11" />

chrisdavidmills:
Hi @ghiscoding, and thanks for reporting this.

This is an interesting problem — it looks like the code you are suggesting should work, but it doesn't. However:

- Firefox supports anchor positioning. If you look at other examples for this module, they work in Firefox.
- `input::-moz-range-thumb` also works, to some degree. If you try, for example, adding `background-color: red` or `border: 1px solid red` to the ruleset, those styles are applied to the thumb, but the rest of the slider (minus the thumb) disappears.

There are multiple buggy things going on here. I would suggest filing a bug on https://bugzilla.mozilla.org to see if the Mozilla engineers can explore this.

ghiscoding:
Yes I know that the other examples do work, but this is the first visible example so it's not a good start for Firefox... Also I know the version 147 is fairly new but technically speaking this page also requires more update since it shows as experimental in Firefox but that is no longer true since it shipped yesterday

<img width="812" height="508" alt="Image" src="https://github.com/user-attachments/assets/6c2ab4d4-b3b9-462a-b6e0-5e5c6fd18308" />

Also I tried to fill a Firefox report anonymously via the link you sent, so I have no idea if they will take it or not. I have never filled a report on their website and this was my first time ever. 

[...]

PR #42797, opened 2026-01-16
position-anchor slider demo: adjust position and add support note

chrisdavidmills:
### Description

As per https://github.com/mdn/content/issues/42767, The thumb-slider anchor positioning demo on the `position-anchor` reference page doesn't work properly. This is because the -webkit-prefixed thumb pseudo-element doesn't work in Firefox. We could add in the -moz-prefixed version, but that still doesn't fix it because it is currently not anchorable (see https://bugzilla.mozilla.org/show_bug.cgi?id=1993699).

As a temporary fix, I have elected to move that demo to the end of the demos list, so it isn't quite so prominent, and add a note to explain the current support issue. I don't think we can really add this into BCD comfortably without lots of hassle, and hopefully we can just remove the note and update the demo once the bug is fixed.

Fixes https://github.com/mdn/content/issues/42767

### DIFF
--- a/files/en-us/web/css/reference/properties/position-anchor/index.md
+++ b/files/en-us/web/css/reference/properties/position-anchor/index.md
@@ -237,7 +237,7 @@ We include an [`<input type="range">`](/en-US/docs/Web/HTML/Reference/Elements/i
 
 #### CSS
 
-We give the thumb, represented by the {{cssxref("::-webkit-slider-thumb")}} pseudo-element, an anchor name of `--thumb`. We then set that name as the value of the `<output>` element's `position-anchor` property, and give it a {{cssxref("position")}} value of `fixed`. These steps associated the `<output>` with the thumb.
+We give the thumb, represented by the {{cssxref("::-webkit-slider-thumb")}} and {{cssxref("::-moz-range-thumb")}} pseudo-elements, an anchor name of `--thumb`. We then set that name as the value of the `<output>` element's `position-anchor` property, and give it a {{cssxref("position")}} value of `fixed`. These steps associated the `<output>` with the thumb.
 
 Finally, we use {{cssxref("left")}} and {{cssxref("top")}} properties with {{cssxref("anchor()")}} values to position the `<output>` relative to the thumb.
 
@@ -263,6 +263,10 @@ output {
 
 ```css
 input::-webkit-slider-thumb {
+  anchor-name: --thumb;
+}
+
+input::-moz-range-thumb {
   anchor-name: --thumb;
 }
 


files/en-us/web/css/reference/properties/position-anchor/index.md
```diff
@@ -293,6 +221,81 @@ Select different values from the drop-down menus to change the anchors that the
 
 {{ EmbedLiveSample("Multiple positioned elements and anchors", "100%", "400") }}
 
+### Using a slider thumb as an anchor
+
+In this example, an {{htmlelement("output")}} is positioned relative to an anchor that is the thumb of a range slider.
+
+#### HTML
+
+We include an [`<input type="range">`](/en-US/docs/Web/HTML/Reference/Elements/input/range) element and an [`<output>`](/en-US/docs/Web/HTML/Reference/Elements/output) element to display the value of the range. The value displayed in the `<output>` element is updated via JavaScript as the slider value changes.
```
estelle:
```suggestion
We include an [`<input type="range">`](/en-US/docs/Web/HTML/Reference/Elements/input/range) element and an {{htmlelement("output")}} element to display the value of the range. The value displayed in the `<output>` element is updated via JavaScript as the slider value changes.
```

files/en-us/web/css/reference/properties/position-anchor/index.md
```diff
@@ -293,6 +221,81 @@ Select different values from the drop-down menus to change the anchors that the
...
+
+```html
+<label for="slider">Change the value:</label>
+<input type="range" min="0" max="100" value="25" id="slider" />
+<output>25</output>
+```
+
+#### CSS
+
+We give the thumb, represented by the {{cssxref("::-webkit-slider-thumb")}} pseudo-element, an anchor name of `--thumb`. We then set that name as the value of the `<output>` element's `position-anchor` property, and give it a {{cssxref("position")}} value of `fixed`. These steps associated the `<output>` with the thumb.
```
estelle:
```suggestion
We give the thumb, represented by the {{cssxref("::-webkit-slider-thumb")}} and {{cssxref("[::-moz-range-thumb")}} pseudo-elements, an anchor name of `--thumb`. We then set that name as the value of the `<output>` element's `position-anchor` property, and give it a {{cssxref("position")}} value of `fixed`. These steps associated the `<output>` with the thumb.
```

files/en-us/web/css/reference/properties/position-anchor/index.md
```diff
@@ -293,6 +221,81 @@ Select different values from the drop-down menus to change the anchors that the
...
+  background: cyan;
+  border: 2px solid darkblue;
+  padding: 3px 6px;
+  border-radius: 10px;
+  font-size: 1.2rem;
+}
+```
+
+```css
+input::-webkit-slider-thumb {
```
estelle, line 265: don't just add the moz here. use where or is, which is forgiving. https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Selector_list#forgiving_selector_list

files/en-us/web/css/reference/properties/position-anchor/index.md
```diff
@@ -293,6 +221,81 @@ Select different values from the drop-down menus to change the anchors that the
...
+output {
+  position-anchor: --thumb;
+  position: absolute;
+  left: anchor(right);
+  bottom: anchor(top);
+}
+```
+
+> [!NOTE]
+> The anchor positioning CSS for this demo currently doesn't work in Firefox. The `::-webkit-slider-thumb` pseudo-element isn't supported in Firefox and, while there is a Firefox-specific equivalent — {{cssxref("::-moz-range-thumb")}} — it is currently not anchorable (see [Firefox bug 1993699](https://bugzil.la/1993699)).
```
estelle:
```suggestion
> The anchor positioning CSS for this demo currently doesn't work in Firefox. The {{cssxref("::-moz-range-thumb")}} is currently not anchorable (see [Firefox bug 1993699](https://bugzil.la/1993699)).
```
(and move this to after the results)

chrisdavidmills: Moved

files/en-us/web/css/reference/properties/position-anchor/index.md
chrisdavidmills: I tried this, and couldn't get it to work. The syntax I tried was this (and several variations):

```css
:where(input::-webkit-slider-thumb, input::-moz-range-thumb) {
  anchor-name: --thumb;
}
```

I've therefore not implemented this, or the associated changes.

Any thoughts on how to make this syntax work?

estelle: fudge. Forgot that pseudo‑elements cannot appear inside the argument of functional pseudo‑classes like `:is()`, `:where()`, and `:not()`, so it has to be repeated.

```css
input::-webkit-slider-thumb {
  anchor-name: --thumb;
}
 input::-moz-range-thumb {
  anchor-name: --thumb;
}
```

we should future-proof with an `@supports`, which is easier said then done. I am currently marinating on how you you can test if the the anchor part, the XXX in this:

```css
@supports selector(::-moz-range-thumb) {
/* only FF sees this */
  @supports not (XXX) {
    body::before {
      content: "Your browser doesn't support anchors on thumbs.";
      background-color: wheat;
      display: block;
      text-align: center;
   }
  }
}
```

what is XXX?

chrisdavidmills: > ```css
> input::-moz-range-thumb {
>   anchor-name: --thumb;
> }
> ```

I've added this in, and explained it in the text. This at least future proofs it in that it will start working when Fx start supporting anchors on thumbs.

> we should future-proof with an `@supports`, which is easier said then done. I am currently marinating on how you you can test if the the anchor part, the XXX in this:

I can understand why you want to do this, but I don't think it is possible. I think we should just get this merged without this.

PR #42812, opened 2026-01-18
Remove a condition which does not really occur.

mfuji09:
### Description

Remove a condition which does not really occur.

> If the key can't be found, `undefined` is returned.

### Motivation

If the given key is not found within the `WeakMap`, this method inserts a new entry with that key. Therefore, unless an exception occurs, the key will always be found when the value is returned.
Of course, if the value of the entry for the given key is `undefined`, this method may return `undefined`, but this is unrelated to the condition described above.

### Additional details

https://tc39.es/proposal-upsert/#sec-weakmap.prototype.getOrInsert

The spec doesn't define that the method returns `undefined` when the given key isn't found.

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/weakmap/getorinsert/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/weakmap/getorinsert/index.md
@@ -37,7 +37,7 @@ getOrInsert(key, defaultValue)
 
 ### Return value
 
-The value associated with the specified key in the `WeakMap` object.
+The value associated with the specified key in the `WeakMap` object. If the key can't be found, `defaultValue` is inserted and returned.
 
 ### Exceptions
 


files/en-us/web/javascript/reference/global_objects/weakmap/getorinsert/index.md
```diff
@@ -37,7 +37,7 @@ getOrInsert(key, defaultValue)
 
 ### Return value
 
-The value associated with the specified key in the `WeakMap` object. If the key can't be found, {{jsxref("undefined")}} is returned.
+The value associated with the specified key in the `WeakMap` object.
```
Josh-Cena:
```suggestion
The value associated with the specified key in the `WeakMap` object. If the key can't be found, `defaultValue` is inserted and returned.
```

PR #42816, opened 2026-01-18
Enhance cookie documentation with Domain and Path notes

binarygit:
Added notes about Domain and Path attributes for cookies.

### Description
When writing cookies with `document.cookies` if you don't specify a Domain or Path attribute, it's set for you by the browser. This PR adds that information to the docs.

### Motivation
They inform readers that a Domain and Path attribute is set on the cookie even if they don't specify it.

### DIFF
--- a/files/en-us/web/api/document/cookie/index.md
+++ b/files/en-us/web/api/document/cookie/index.md
@@ -49,7 +49,6 @@ You can also assign to this property a string of the form `"key=value"`, specify
 
   > [!NOTE]
   > If the `Path` attribute is omitted, the browser assigns a default path derived from the request URL.
-
   - `;samesite`: The `SameSite` attribute of a {{httpheader("Set-Cookie")}} header can be set by a server to specify when the cookie will be sent. Possible values are `lax`, `strict` or `none` (see also [Controlling third-party cookies with `SameSite`](/en-US/docs/Web/HTTP/Guides/Cookies#controlling_third-party_cookies_with_samesite)).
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/document/cookie/index.md
```diff
@@ -31,6 +31,8 @@ You can also assign to this property a string of the form `"key=value"`, specify
     > The domain _must_ match the domain of the JavaScript origin.
     > Setting cookies to foreign domains will be silently ignored.
 
+    > [!NOTE]
```
hamishwillee, line 34: This is already covered above

> If not specified, this defaults to the host portion of the current document location and the cookie is not available on subdomains.
    If a domain is specified, subdomains are always included.
    Contrary to earlier specifications, leading dots in domain names are ignored, but browsers may decline to set the cookie containing such dots.

files/en-us/web/api/document/cookie/index.md
```diff
@@ -45,6 +47,8 @@ You can also assign to this property a string of the form `"key=value"`, specify
 
   - `;path=path`: The value of the cookie's `Path` attribute (See [Define where cookies are sent](/en-US/docs/Web/HTTP/Guides/Cookies#define_where_cookies_are_sent) for more information).
 
+  > [!NOTE]
```
hamishwillee, line 50: This is not explicitly mentioned, but is covered in the link above.

hamishwillee:
Thanks, but this information either duplicates information already in the document, or that is readily accessible.

PR #42817, opened 2026-01-18
Add note about list-style inheritance interruption (#42806)

### LINKED ISSUE #42806: list-style: add a note that ul/ol/... usually interrupt inheritance
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/list-style

Try it

The text says:
> Because this property is inherited, it can be set on a parent element (normally `<ol>` or `<ul>`) to make the same list styling apply to all the nested items.

This may give the (wrong) impression that `list-style` would also be inherited by nested lists.  But this is usually not true since the wrapper elements of the inner lists (`<ol>` or `<ul>` or ...) override the inherited values already in the browser's default style sheets.

There should be a hint mentioning that inheritance is actually interrupted by `<ol>`/`<ul>`/...

_No response_

This also applies to `list-style`'s constituent properties.

### MDN metadata

<details>

* Folder: `en-us/web/css/reference/properties/list-style`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/list-style
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/reference/properties/list-style/index.md
* Last commit: https://github.com/mdn/content/commit/1dbba9f7a2c2e35c6e01e8a63159e2aac64b601b
* Document last modified: 2025-12-05T16:09:35.000Z

</details>

AbhishekGiri04:
### Description

Added a note explaining that HTML list elements (`<ul>`, `<ol>`, `<menu>`) reset the `list-style` property in their user-agent stylesheet, interrupting inheritance from ancestor elements.

### Motivation

Users were confused about why `list-style` inheritance doesn't work as expected with nested lists. This note clarifies that list elements reset the property and explains how to apply inherited styles explicitly.

### Additional details

- Added NOTE block after the inheritance explanation
- Explains user-agent stylesheet behavior
- Provides guidance on how to work with this behavior

Fixes #42806


chrisdavidmills (changes requested):
Hi @AbhishekGiri04, and thank you for your work towards fixing this issue. However, I think some more work and research is needed here. Some points to think about:

1. I did a bunch of testing, and observed that it only seems to be `list-style-type` that this behavior is true for. `list-style-position` and `list-style-image` seem to inherit normally.
2. The way you've written this, the note seems to contradict the earlier statement about inheritance, so it reads strangely. I would put the sentence starting "Because this property is inherited..." into its own paragraph, and then make the note a part of that paragraph (remove the note styling). Maybe start out by saying that HTML list elements always take their `list-style-type` from the initial value set in the user-agent stylesheets, then go on to say this means that even though `list-style` inherits, only the `list-style-position` and `list-style-image` values will be inherited by nested list elements. If you want nested list elements to have the same list-style-type values as the parent list element, you need to set it explicitly on each nested list element. Something like that, anyway.
3. You also need to fix this issue on the `list-style-type` reference page. The fix will be similar, but you only need to talk about `list-style-type` on this page.

Josh-Cena:
@AbhishekGiri04?

chrisdavidmills (commented):
Looking at this again months later, I'm unclear on its intent.

We are basically saying that the UA stylesheet sets a default `list-style-type` on list elements such as `<ul>`, and you can't set `list-style` or `list-style-type` on one of these elements and have its `<li>` children inherit that list style without setting `list-style-type: inherit` on them.

This sounds untrue to me. If you look at this demo I have created:

https://jsbin.com/rekeyolida/edit?html,css,output

The children inherit `list-style: none;`

So I'm thinking that we should remove the bit starting "However, user-agent..." on both pages, and get it merged.

Or am I missing the point? Any thoughts, @Josh-Cena ?

Josh-Cena:
Your example has no nested lists. The note is specifically about how list containers (`ul`, `ol`, `menu`) explicitly set `list-style-type` and therefore doesn't inherit.

```html
<ul style="list-style: none">
  <li>No marker: inherits none from the outer ul</li>
  <li>
    Also no marker
    <ul>
      <li>Circle marker: inherits circle from the inner ul</li>
    </ul>
  </li>
</ul>
```

chrisdavidmills:
> Your example has no nested lists. The note is specifically about how list containers (`ul`, `ol`, `menu`) explicitly set `list-style-type` and therefore doesn't inherit.

Gah, in my rush to look at this, I somehow missed the word "nested".

I've added a brief example to both pages to make sure it is crystal clear.

chrisdavidmills (approved):
LGTM now. @Josh-Cena, can you give the new stuff a quick check before merging?

Josh-Cena:
Sure :) This LGTM

PR #42820, opened 2026-01-18
Remove empty table row in named color index

Bishops-exe:
Removed an unnecessary table row from the named color reference.

### Description

I removed an unclosed and unneccesary table tag

### Motivation

Idk


chrisdavidmills (approved):
Thanks for the fix, @AdrianDoesCoding. It is weird to end up with a superfluous opening `<tr>` tag in that table. Good to get rid of it.

PR #42824, opened 2026-01-19
Server timing duration - mention unit

VojtaStanek:
Based on [note in W3C spec](https://w3c.github.io/server-timing/#duration-attribute):
> Since duration is a DOMHighResTimeStamp, it usually represents a duration in milliseconds. Since this is not enforcable in practice, duration can represent any unit of time, and having it represent a duration in milliseconds is a recommendation.

### Motivation

Currently there is no note about the unit of duration, it is helpful to the reader to have a note about thi. 

### Additional details

Spec: https://w3c.github.io/server-timing/#duration-attribute


sideshowbarker:
Vojta, congrats on landing your first docs change here — welcome aboard 🎉

PR #42827, opened 2026-01-19
clarify common use cases of the event interface

Asha-punia:
### Description

This PR adds a short clarification describing common real world uses of Event interface to improve readability for beginners.


hamishwillee:
Thanks, but the following paragraph covers this, and also that that the event might also be programmatically created. Closing as not planned.

PR #42834, opened 2026-01-20
SVG: Add examples for repeating angled gradients

estelle:
i spend a long time scouring the web to figure out how to create angles that stayed constant regardless of aspect ratio

### DIFF
--- a/files/en-us/web/svg/reference/element/lineargradient/index.md
+++ b/files/en-us/web/svg/reference/element/lineargradient/index.md
@@ -99,8 +99,8 @@ svg {
       id="grad"
       x1="0"
       y1="0"
-      x2="10"
-      y2="10"
+      x2="20"
+      y2="20"
       spreadMethod="repeat"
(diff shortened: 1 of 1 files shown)


caugner:
FYI

> * `Macro svginfo produces link /en-US/docs/Web/SVG/Element which is a redirect`

This issue will be fixed via https://github.com/mdn/rari/pull/490.

> * `Macro svginfo produces link /en-US/docs/Web/SVG/Reference/Element/discard which is a redirect`

This issue will be fixed via https://github.com/mdn/content/pull/42836.

estelle:
> Thanks @estelle - it works and looks pretty cool.
> 
> FWIW I do prefer examples that explain how things "work" as well as what they acheive. i.e. I can see that this works but I don't know if I could do it myself in any other context from this, or what changes would be OK vs what would break the aspect ratios being the same.

I agree, but the supporting docs on mdn don't explain things. When i have time, i should update the `gradientUnits` attribute page, so we would have docs to link to.

PR #42836, opened 2026-01-20
fix(jsondata/SVGData): remove discard

caugner:
### Description

Removes the SVG `<discard>` element from `SVGData.json`.

### Motivation

This was causing issues/flaws, because the discard element page got removed, and now redirects to the SVG Element overview.

(This also means `<discard>` in https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/linearGradient#usage_context points to the overview.)

Noticed in https://github.com/mdn/content/pull/42834.

Related to https://github.com/mdn/content/issues/39618.

### DIFF
--- a/files/jsondata/SVGData.json
+++ b/files/jsondata/SVGData.json
@@ -203,21 +203,6 @@
       "attributes": ["coreAttributes", "'class'", "'style'"],
       "interfaces": ["SVGDescElement"]
     },
-    "discard": {
-      "categories": ["animationElement"],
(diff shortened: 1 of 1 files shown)


caugner:
> I wonder if we could’ve done anything differently when removing `<discard>` so this data would be cleaned up as well.

In theory, we could lint the `jsondata` files, similar to how we lint the `sidebars`. But it would be more complex, as each `jsondata` file is different.

PR #42841, opened 2026-01-20
Editorial review: Document WebAssembly reference types

chrisdavidmills:
### Description

This PR aims to document [WebAssembly reference types](https://github.com/WebAssembly/spec/blob/main/proposals/reference-types/Overview.md), as part of the "Fill the gaps" project.

Specifically, it documents:

- [x] The externref type
- [x] The funcref type
- [x] table.get
- [x] table.set
- [x] table.size
- [x] table.grow
- [x] table.fill
- [x] The fact that multiple tables are allowed.
- [x] Declared element segments, e.g. `(elem declare func $f1 $f2)`

### DIFF
--- a/files/en-us/webassembly/reference/definitions/table/index.md
+++ b/files/en-us/webassembly/reference/definitions/table/index.md
@@ -63,7 +63,7 @@ table name initial_size max_size type
     - [`funcref`](/en-US/docs/WebAssembly/Reference/Types/funcref)
       - : Stores references to functions defined inside Wasm.
     - [`externref`](/en-US/docs/WebAssembly/Reference/Types/externref)
-      - : Store references to external functions defined inside JavaScript.
+      - : Store references to external values defined inside JavaScript.
 
 ## Description
 
@@ -81,7 +81,7 @@ Optionally, you can also provide an identifier, which can be used to identify th
 (table $mytable 2 10 funcref)
 ```
 
-The following defines a function type, defines a basic function with that type that returns an `i32`, and and forward-declares it using `(elem declare func $f1)` so it can be referenced later on.
+The following defines a function type, defines a basic function with that type that returns an `i32`, and forward-declares it using `(elem declare func $f1)` so it can be referenced later on.
 
 ```wat
 (type $ret_i32 (func (result i32)))
@@ -99,7 +99,7 @@ To call a function referenced in a table, you have to reference the table and th
 (call_indirect (type $ret_i32) (local.get $index))
 ```
 
-It is possible to mutate tables at runtime using instructions like [`table.set`](/en-US/docs/WebAssembly/Reference/Table/get) and [`table.fill`](/en-US/docs/WebAssembly/Reference/Table/get), and retrieve values using [`table.get`](/en-US/docs/WebAssembly/Reference/Table/get).
+It is possible to mutate tables at runtime using instructions like [`table.set`](/en-US/docs/WebAssembly/Reference/Table/set) and [`table.fill`](/en-US/docs/WebAssembly/Reference/Table/fill), and retrieve values using [`table.get`](/en-US/docs/WebAssembly/Reference/Table/get).
 
 ### External references
 
@@ -153,7 +153,7 @@ But you could instead reference the tables by their index values (`0` specifies 
 (call_indirect 1 (type $ret_i32) (i32.const 0))
 ```
 
-If you don't specify an indentifying name _or_ an index, the index `0` is assumed:
+If you don't specify an identifying name _or_ an index, the index `0` is assumed:
 
 ```wat
 ;; Accesses the table with index 0
--- a/files/en-us/webassembly/reference/table/size/index.md
+++ b/files/en-us/webassembly/reference/table/size/index.md
@@ -8,7 +8,7 @@ spec-urls: https://webassembly.github.io/spec/core/syntax/instructions.html#synt
 sidebar: webassemblysidebar
 ---
 
-The **`table.size`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) returns the current number of elements contained in a table.
+The **`table.size`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) returns the current size of the table.
 
 {{InteractiveExample("Wat Demo: table.size", "tabbed-taller")}}
 
--- a/files/en-us/webassembly/reference/types/externref/index.md
+++ b/files/en-us/webassembly/reference/types/externref/index.md
@@ -8,7 +8,7 @@ spec-urls: https://webassembly.github.io/spec/core/syntax/types.html#reference-t
 sidebar: webassemblysidebar
 ---
 
-The **`externref`** value type holds a JavaScript value, enabling it to be given to a Wasm module without the need for copying or serializing.
+The **`externref`** value type references a JavaScript value, enabling it to be given to a Wasm module without the need for copying or serializing.
 
 {{InteractiveExample("Wat Demo: externref", "tabbed-taller")}}
 
@@ -147,4 +147,5 @@ The value `16` is displayed inside the output `<p>` element. This makes sense, a
 
 ## See also
 
-xx
+- [`funcref`](/en-US/docs/WebAssembly/Reference/Types/funcref)
+- [`table`](/en-US/docs/WebAssembly/Reference/Definitions/table) definition
--- a/files/en-us/webassembly/reference/types/funcref/index.md
+++ b/files/en-us/webassembly/reference/types/funcref/index.md
@@ -154,4 +154,5 @@ The original array has all its values multiplied by `2`, and the resulting new a
 
 ## See also
 
-xx
+- [`externref`](/en-US/docs/WebAssembly/Reference/Types/externref)
+- [`table`](/en-US/docs/WebAssembly/Reference/Definitions/table) definition
--- a/files/en-us/webassembly/reference/types/index.md
+++ b/files/en-us/webassembly/reference/types/index.md
@@ -7,7 +7,7 @@ sidebar: webassemblysidebar
 
 This section documents the different WebAssembly types.
 
-## Value types
+## Reference types
 
 - [`externref`](/en-US/docs/WebAssembly/Reference/Types/externref)
   - : References a JavaScript value, enabling it to be given to a Wasm module without the need for copying or serializing.
--- a/files/en-us/webassembly/guides/index.md
+++ b/files/en-us/webassembly/guides/index.md
@@ -8,7 +8,37 @@ sidebar: webassemblysidebar
 
 The [WebAssembly](/en-US/docs/WebAssembly) guides cover topics such as high-level concepts, compiling from different languages, the textual representation of the Wasm binary format, and how to run WebAssembly.
 
-{{SubPagesWithSummaries}}
+- [WebAssembly concepts](/en-US/docs/WebAssembly/Guides/Concepts)
+  - : This article explains the concepts behind how WebAssembly works including its goals, the problems it solves, and how it runs inside the web browser's JavaScript engine.
+
+## Compiling WebAssembly
+
+- [Compiling a new C/C++ module to WebAssembly](/en-US/docs/WebAssembly/Guides/C_to_Wasm)
+  - : When you've written a new code module in a language like C/C++, you can compile it into WebAssembly using a tool like [Emscripten](https://emscripten.org/). Let's look at how it works.
+- [Compiling an existing C module to WebAssembly](/en-US/docs/WebAssembly/Guides/Existing_C_to_Wasm)
+  - : A core use-case for WebAssembly is to take the existing ecosystem of C libraries and allow developers to use them on the web.
+- [Compiling from Rust to WebAssembly](/en-US/docs/WebAssembly/Guides/Rust_to_Wasm)
+  - : This tutorial shows you how to compile a Rust project into WebAssembly and use it in an existing web app.
+
+## WebAssembly language guide
+
+- [Understanding WebAssembly text format](/en-US/docs/WebAssembly/Guides/Understanding_the_text_format)
+  - : To enable WebAssembly to be read and edited by humans, there is a textual representation of the Wasm binary format. This is an intermediate form designed to be displayed in text editors, browser developer tools, and other similar environments. This article explains how the text format works in terms of its raw syntax, and how it relates to the underlying bytecode it represents and the wrapper objects that represent Wasm in JavaScript.
+- [Converting WebAssembly text format to binary](/en-US/docs/WebAssembly/Guides/Text_format_to_Wasm)
+  - : WebAssembly has an S-expression-based textual representation, an intermediate form designed to be exposed in text editors, browser developer tools, etc. This article explains a little bit about how it works, and how to use available tools to convert text format files to the Wasm format.
+- [Exported WebAssembly functions](/en-US/docs/WebAssembly/Guides/Exported_functions)
+  - : Exported WebAssembly functions are how WebAssembly functions are represented in JavaScript. This article describes what they are in a little more detail.
+- [WebAssembly JavaScript builtins](/en-US/docs/WebAssembly/Guides/JavaScript_builtins)
+  - : WebAssembly JavaScript builtins are Wasm equivalents of JavaScript operations that provide a way to use JavaScript features inside Wasm modules without having to import JavaScript glue code to provide a bridge between JavaScript and WebAssembly values and calling conventions.
+- [WebAssembly Imported global string constants](/en-US/docs/WebAssembly/Guides/Imported_string_constants)
+  - : WebAssembly imported global string constants make working with JavaScript strings inside Wasm modules easier by removing the need for a lot of the boilerplate associated with traditional string imports.
+
+## JavaScript API guide
+
+- [Using the WebAssembly JavaScript API](/en-US/docs/WebAssembly/Guides/Using_the_JavaScript_API)
+  - : This article teaches you how to use the key features of the WebAssembly JavaScript API, including loading Wasm modules and manipulating WebAssembly memories, tables, and globals.
+- [Loading and running WebAssembly code](/en-US/docs/WebAssembly/Guides/Loading_and_running)
+  - : To use WebAssembly in JavaScript, you first need to pull your module into memory before compilation/instantiation. This article provides a reference for the different mechanisms that can be used to fetch WebAssembly bytecode, as well as how to compile/instantiate then run it.
 
 ## See also
 
--- a/files/en-us/webassembly/guides/rust_to_wasm/index.md
+++ b/files/en-us/webassembly/guides/rust_to_wasm/index.md
@@ -5,7 +5,7 @@ page-type: guide
 sidebar: webassemblysidebar
 ---
 
-If you have some Rust code, you can compile it into [WebAssembly](/en-US/docs/WebAssembly) (Wasm). This tutorial will show you how to compile a Rust project into WebAssembly and use it in an existing web app.
+This tutorial shows you how to compile a Rust project into WebAssembly and use it in an existing web app.
 
 ## Rust and WebAssembly use cases
 
--- a/files/en-us/webassembly/guides/using_the_javascript_api/index.md
+++ b/files/en-us/webassembly/guides/using_the_javascript_api/index.md
@@ -5,12 +5,12 @@ page-type: guide
 sidebar: webassemblysidebar
 ---
 
-If you have already [compiled a module from another language using tools like Emscripten](/en-US/docs/WebAssembly/Guides/C_to_Wasm), or [loaded and run the code yourself](/en-US/docs/WebAssembly/Guides/Loading_and_running), the next step is to learn more about using the other features of the WebAssembly JavaScript API. This article teaches you what you'll need to know.
+This article teaches you how to use the key features of the WebAssembly JavaScript API, including loading Wasm modules and manipulating WebAssembly memories, tables, and globals.
 
 > [!NOTE]
 > If you are unfamiliar with the basic concepts mentioned in this article and need more explanation, read [WebAssembly concepts](/en-US/docs/WebAssembly/Guides/Concepts) first, then come back.
 
-## Some examples
+## Loading Wasm modules in JavaScript
 
 Let's run through some examples that explain how to use the WebAssembly JavaScript API, and how to use it to load a Wasm module in a web page.
 
--- a/files/en-us/webassembly/reference/table/index.md
+++ b/files/en-us/webassembly/reference/table/index.md
@@ -23,7 +23,7 @@ This set of pages details the table instructions available in Wasm to create and
 - [`table.set`](/en-US/docs/WebAssembly/Reference/Table/set)
   - : Changes the value stored in a particular table element.
 - [`table.size`](/en-US/docs/WebAssembly/Reference/Table/size)
-  - : Returns the current number of elements contained in a table.
+  - : Returns the current size of the table.
 
 ## Specifications
 
--- a/files/sidebars/webassemblysidebar.yaml
+++ b/files/sidebars/webassemblysidebar.yaml
@@ -9,26 +9,35 @@ sidebar:
     title: Guides
   - link: /WebAssembly/Guides/Concepts
     title: WebAssembly_concepts
-  - link: /WebAssembly/Guides/C_to_Wasm
-    title: Compiling_to_Wasm
-  - link: /WebAssembly/Guides/Existing_C_to_Wasm
-    title: Existing_C_to_Wasm
-  - link: /WebAssembly/Guides/Rust_to_Wasm
-    title: Compiling_Rust_to_Wasm
-  - link: /WebAssembly/Guides/Using_the_JavaScript_API
-    title: JavaScript_API
-  - link: /WebAssembly/Guides/Understanding_the_text_format
-    title: Text_format
-  - link: /WebAssembly/Guides/Text_format_to_Wasm
-    title: Text_format_to_Wasm
-  - link: /WebAssembly/Guides/Loading_and_running
-    title: Loading_and_running
-  - link: /WebAssembly/Guides/Exported_functions
-    title: Exported_functions
-  - link: /WebAssembly/Guides/JavaScript_builtins
-    title: JavaScript_builtins
-  - link: /WebAssembly/Guides/Imported_string_constants
-    title: Imported_string_constants
+  - title: Compiling_WebAssembly
+    details: closed
+    children:
+      - link: /WebAssembly/Guides/C_to_Wasm
+        title: Compiling_to_Wasm
+      - link: /WebAssembly/Guides/Existing_C_to_Wasm
+        title: Existing_C_to_Wasm
+      - link: /WebAssembly/Guides/Rust_to_Wasm
+        title: Compiling_Rust_to_Wasm
+  - title: Language_guide
+    details: closed
+    children:
+      - link: /WebAssembly/Guides/Understanding_the_text_format
+        title: Text_format
+      - link: /WebAssembly/Guides/Text_format_to_Wasm
+        title: Text_format_to_Wasm
+      - link: /WebAssembly/Guides/Exported_functions
+        title: Exported_functions
+      - link: /WebAssembly/Guides/JavaScript_builtins
+        title: JavaScript_builtins
+      - link: /WebAssembly/Guides/Imported_string_constants
+        title: Imported_string_constants
+  - title: API_guide
+    details: closed
+    children:
+      - link: /WebAssembly/Guides/Using_the_JavaScript_API
+        title: JavaScript_API
+      - link: /WebAssembly/Guides/Loading_and_running
+        title: Loading_and_running
   - type: section
     link: /WebAssembly/Reference
     title: Reference
@@ -156,12 +165,15 @@ l10n:
     JavaScript_interface: JavaScript-Schnittstelle
   en-US:
     WebAssembly_concepts: WebAssembly concepts
-    Compiling_to_Wasm: Compiling from C/C++ to WebAssembly
+    Compiling_WebAssembly: Compiling WebAssembly
+    Language_guide: WebAssembly language
+    API_guide: JavaScript API guide
+    Compiling_to_Wasm: Compiling a new C/C++ module to WebAssembly
     Existing_C_to_Wasm: Compiling an existing C module to WebAssembly
     Compiling_Rust_to_Wasm: Compiling from Rust to WebAssembly
     JavaScript_API: Using the WebAssembly JavaScript API
     Text_format: Understanding WebAssembly text format
-    Text_format_to_Wasm: Converting WebAssembly text format
+    Text_format_to_Wasm: Converting WebAssembly text format to binary
     Loading_and_running: Loading and running WebAssembly code
     Exported_functions: Exported WebAssembly functions
     JavaScript_builtins: JavaScript builtins


hamishwillee:
@MendyBerger This is still draft, but I've added you as a review in case you have some bandwidth to look at it when the time comes.

files/en-us/webassembly/reference/table/grow/index.md
```diff
@@ -0,0 +1,226 @@
...
+title: "grow: Wasm table instruction"
+short-title: grow
+slug: WebAssembly/Reference/Table/grow
+page-type: webassembly-instruction
+browser-compat: webassembly.reference-types
+spec-urls: https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-table
+sidebar: webassemblysidebar
+---
+
+The **`grow`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) increases the size of a table by a specified number of segments.
```
eqrion: Could we use the full instruction name `table.grow`? Abbreviating it runs the risk of colliding with other things like `memory.grow`. In general existing wasm docs always uses the full names for instructions. It looks like the reference in MDN doesn't do that, but I think that'd be good to change.

eqrion:
```suggestion
The **`grow`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) increases the size of a table by a specified number of elements.
```

files/en-us/webassembly/reference/table/grow/index.md
```diff
@@ -0,0 +1,226 @@
...
+## Syntax
+
+```wat
+table.grow identifier initial_value grow_amount
+```
+
+- `table.grow`
+  - : The `table.grow` instruction type. Must always be included first.
+- `identifier`
+  - : The identifier for the table you want to grow; must begin with a `$` symbol. [EDITORIAL: is this optional?]
```
eqrion: Yes, [the identifier is optional](https://webassembly.github.io/spec/core/text/instructions.html#id1). We originally only supported a single table in a module, which meant you didn't need to say which one. So if you omit an identifier, it default to a '0' index for the first table in the module.

Also technically this is either an 'index' or an 'identifier'.

All wasm definitions (tables, globals, memories, etc) have an index space. The first table defined is 0, the second is 1, etc. The instructions then need to know if they're referencing table 0/1/2/3/etc.

When they're defined you can also give them an identifier using $foo, which you can then use when referencing that table to avoid needing to hardcode the numeric index.

Most handwritten wat will use identifiers to be readable, but if you print a wasm binary to wat, you won't get any identifiers only indices (binary format doesn't keep that around by default).

files/en-us/webassembly/reference/table/grow/index.md
```diff
@@ -0,0 +1,226 @@
...
+- `initial_value`
+  - : The initial value to set for the new table segments. This can be one of the following:
+    - `ref.null func`
+      - : Sets the initial values to `null`.
+    - OTHER
+      - : EDITORIAL: what other values can be set for the initial value?
+- `grow_amount`
+  - : The number of segments to grow the table by. This must be an `i32` value, for example `(i32.const 1)`.
```
eqrion:
```suggestion
  - : The number of elements to grow the table by. This must be an `i32` value, for example `(i32.const 1)`.
```

files/en-us/webassembly/reference/table/grow/index.md
```diff
@@ -0,0 +1,226 @@
...
+
+An `i32` equal to the size of the table before the `grow` instruction is applied to it, or `-1` if the table failed to grow, for example due to an out-of-memory (OOM) error or the new size being greater than the [table's maximum size](/en-US/docs/WebAssembly/Reference/Table/table#max_size).
+
+To retrieve the new table size after the `grow` instruction is applied to it, use the [`table.size`](/en-US/docs/WebAssembly/Reference/Table/size) instruction.
+
+### Opcodes
+
+| Instruction  | Binary opcode |
+| ------------ | ------------- |
+| `table.grow` | `0xFC`        |
```
eqrion:
```suggestion
| `table.grow` | `0xFC` `0x0F`       |
```

table.grow technically has two opcode bytes.

Image

There's the initial 0xFC followed by a hardcoded 15:u32 (which is a fancy way of saying 0x0F, 15=0x0F in hex).

files/en-us/webassembly/reference/table/grow/index.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Description
+
+`table.grow` is used to increase the size of a table by a specified number of segments.
+
+Note that `table.grow` must be included inside a function. Trying to include the instruction at the top level of a module results in an error.
```
eqrion: All instructions must be used within a function, or as an initializer expression for globals. So this warning might get a little repetitive and be worthwhile somewhere else?

files/en-us/webassembly/reference/table/table/index.md
```diff
@@ -0,0 +1,153 @@
...
+title: "table: Wasm table instruction"
+short-title: table
+slug: WebAssembly/Reference/Table/table
+page-type: webassembly-instruction
+browser-compat: webassembly.reference-types
+spec-urls: https://webassembly.github.io/spec/core/syntax/modules.html#syntax-table
+sidebar: webassemblysidebar
+---
+
+The **`table`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) creates a new table.
```
eqrion: Technically, a `(table)` is a top-level definition in a module, not an instruction. The other definitions a module can have are 'type', 'memory', 'global', 'func', 'elem', and 'data'. I think they could go into their own section of the reference.

files/en-us/webassembly/reference/table/table/index.md
```diff
@@ -0,0 +1,153 @@
...
+- `storage_type`
+  - : The name of the function type to store. Possible values are:
+    - `funcref`
+      - : Stores references to functions defined inside Wasm.
+    - `externref`
+      - : Store references to external functions defined inside JavaScript.
+
+## Description
+
+WebAssembly tables allow storage function references, which can be later retrieved for efficient, indirect function access when needed. The `table` instruction creates a new table.
```
eqrion:
```suggestion
WebAssembly tables allow storage of reference values separate from byte-oriented WebAssembly memories. The primary use-case is for storing function references that can be used with call_indirect to support indirect function calls for languages that have them. The `table` instruction creates a new table.
```

files/en-us/webassembly/reference/types/externref/index.md
```diff
@@ -0,0 +1,150 @@
...
+title: "externref: Wasm type"
+short-title: externref
+slug: WebAssembly/Reference/Types/externref
+page-type: webassembly-instruction
+browser-compat: webassembly.reference-types
+spec-urls: https://webassembly.github.io/spec/core/syntax/types.html#reference-types
+sidebar: webassemblysidebar
+---
+
+The **`externref`** type references an object defined in JavaScript, enabling it to be used inside a Wasm module without the need for copying or serializing.
```
eqrion:
```suggestion
The **`externref`** value type holds a JavaScript value enabling it to be given to a Wasm module without the need for copying or serializing. WebAssembly code cannot directly access the JS value and must either store it in a global/table, or pass it to an imported JS function.
```

eqrion: `externref` can hold and JS value, not just objects. So you can pass JS numbers, strings, symbols, etc.

files/en-us/webassembly/reference/types/externref/index.md
```diff
@@ -0,0 +1,150 @@
...
+
+;; Define named function parameter
+(param $num externref)
+```
+
+## Description
+
+The `externref` type is used to reference objects defined in JavaScript from inside Wasm modules. Any object type can be referenced, which is very useful when importing functions into WebAssembly that manipulate DOM nodes, write to canvas contexts, or manipulate image data. There is no longer the need to represent data in appropriate formats (for example, serialize objects) before it is sent over to Wasm.
+
+Each `externref` introduces a performance cost to an application, so you need to use them as sparingly as possible. For example, if you want to do a load of calculations using a Wasm function, you should pass all the data over to the Wasm module once (for example, in an array), perform the calculations, end then pass the array back again. You should avoid using multiple exported function calls for individual calculations. If you are passing single number values into your Wasm module, you can avoid using `externref` altogether.
```
eqrion: I think the cost of an `externref` by itself is pretty low and doesn't need to be called out. It just lets wasm holds JS values given to it by JS.

files/en-us/webassembly/reference/types/funcref/index.md
```diff
@@ -0,0 +1,157 @@
...
+title: "funcref: Wasm type"
+short-title: funcref
+slug: WebAssembly/Reference/Types/funcref
+page-type: webassembly-instruction
+browser-compat: webassembly.reference-types
+spec-urls: https://webassembly.github.io/spec/core/syntax/types.html#reference-types
+sidebar: webassemblysidebar
+---
+
+The **`funcref`** type references a function defined in Wasm, enabling higher-order functions to be used across the Wasm and JavaScript language boundaries.
```
eqrion:
```suggestion
The **`funcref`** value type references a function defined in Wasm, enabling higher-order functions to be used across the Wasm and JavaScript language boundaries.
```

files/en-us/webassembly/reference/types/index.md
```diff
@@ -0,0 +1,13 @@
...
+page-type: landing-page
+sidebar: webassemblysidebar
+---
+
+This section documents the different WebAssembly types.
+
+- [`externref`](/en-US/docs/WebAssembly/Reference/Types/externref)
+  - : References an object defined in JavaScript, enabling it to be used inside a Wasm module without the need for copying or serializing.
+- [`funcref`](/en-US/docs/WebAssembly/Reference/Types/funcref)
+  - : References a function defined in Wasm, enabling higher-order functions to be used across the Wasm and JavaScript language boundaries.
```
eqrion, line 15: Technically these are both 'value types', which just means they're the types of things wasm can use for values in computations.

Another important kind of type are 'defined types' or 'composite types' [1]. These are 'func types', 'struct types' and 'array types'. struct/arrays are added in the GC proposal. Func types have been around since 1.0, and are used to define what params/results a function can have. Every function defined in a module has a type associated with it, even if you don't write it out.

So if you write:
```
(func (param i32))
```
That's actually syntax sugar for:
```
(type $F
  (func (param i32))
)
(func (type $F))
```

I say all that just to say that the structure of the 'types' section in the reference should probably have a section for 'value types' and also 'composite types'.

[1] https://webassembly.github.io/spec/core/syntax/types.html#composite-types

files/en-us/webassembly/reference/table/size/index.md
```diff
@@ -0,0 +1,175 @@
...
+    (call $output
+      (local.get $elem)
+      (table.size)
+    )
+  )
+)
+```
+
+> [!NOTE]
+> In this code, we haven't specified the table identifier (`$my_table`) after the `table.size` instructions. This is OK, as we've only got one table defined, so the compiler can work out which table the instruction is aimed at.
```
eqrion: I think a lot of instructions default to guessing an index of `0` when you don't specify anything, which works if there is only 1 thing defined for that index space.

files/en-us/webassembly/reference/table/grow/index.md
```diff
@@ -0,0 +1,226 @@
...
+
+```wat
+table.grow identifier initial_value grow_amount
+```
+
+- `table.grow`
+  - : The `table.grow` instruction type. Must always be included first.
+- `identifier`
+  - : The identifier for the table you want to grow; must begin with a `$` symbol. [EDITORIAL: is this optional?]
+- `initial_value`
```
eqrion: For a table with element type X (e.g. 'funcref'), it can be any X value. The set of instructions that can create that can vary quite a bit. It could also receive the initial value from a function param or global or something.

```
(table externref)

(func (param externref)
  (table.grow
    i32.const 0
    local.get 0 ;; use the externref param as the initial_value
  )
)
```

So I think saying something like this, would be accurate.
```
The initial value to set for the new table segments. This must be the element type of the table.
```

eqrion (commented):
Thanks for working on this, hopefully these aren't too pedantic. I've not looked at all of the instructions thoroughly, mostly at a medium level.

files/en-us/webassembly/reference/table/size/index.md
bvisness: Yes, to be specific this is an abbreviation defined [here](https://webassembly.github.io/spec/core/text/instructions.html#id1). That style of abbreviation applies to table and memory instructions but nothing else.

files/en-us/webassembly/reference/table/grow/index.md
chrisdavidmills: This makes sense to me. In my next commit, I've updated all new instruction reference pages to use this style, and also updated a few wrong memory ones.

chrisdavidmills: I'm glad you highlighted this, as I was wondering about the correct terminology. I've updated this throughout.

files/en-us/webassembly/reference/table/grow/index.md
chrisdavidmills: This is really useful information, thanks! I've kept this information in a place where I'll be able to find it for when I write wasm guides.

For now, I have updated all the ref pages to include a full description of the possible values for the identifier, it's optional status, and the default value if it is omitted.

files/en-us/webassembly/reference/table/grow/index.md
chrisdavidmills: updated

files/en-us/webassembly/reference/table/grow/index.md
chrisdavidmills: Cool, updated. I've also filled in the correct second value for `table.size`.

Question — would it be more accurate to put "Hex opcode" for the table header (or similar), if we are expressing these in hex?

files/en-us/webassembly/reference/table/grow/index.md
chrisdavidmills: Yeah, good point. I've deleted all of these for now, and made a note to cover this in an appropriate guide somewhere.

files/en-us/webassembly/reference/table/table/index.md
chrisdavidmills: Ah, good to clear this up. I have created a new reference section called "Module definitions", and placed this page underneath it. You'll have to let me know what you think of the new structure.

files/en-us/webassembly/reference/table/grow/index.md
bvisness: Actually, this is a weird case because the 0xFC must be exactly a single byte, but the 15 is a `u32`, a variable-width LEB128 integer. So there are actually many ways to express that 15 value. Of course, a single byte is the most reasonable way to do that, but for some reason the spec authors decided to allow other encodings too. 

I'm not sure what the best way is to describe it. Probably, for correctness, it should be expressed more or less how it is in the spec.

files/en-us/webassembly/reference/table/table/index.md
chrisdavidmills: updated

files/en-us/webassembly/reference/types/externref/index.md
chrisdavidmills: Good to know, thanks. I've used your text on the page, but restructured it a bit.

files/en-us/webassembly/reference/types/externref/index.md
chrisdavidmills: OK, I've removed this paragraph.

files/en-us/webassembly/reference/types/funcref/index.md
chrisdavidmills: updated

files/en-us/webassembly/reference/types/index.md
chrisdavidmills: This is all useful, thanks. I've noted down these details to use in the guides, later on. For the reference, just for now, I've decided to just add a heading of "Value types" to the main "WebAssembly types" page. I'll then add the composite types under a "Composite types" heading, when we start to document those.

I don't think there's enough types to require a more complex structure than this.

files/en-us/webassembly/reference/table/size/index.md
chrisdavidmills: I've updated the note to

> [!NOTE]
> In this code, we haven't specified the table identifier (`$my_table`) after the `table.size` instructions. This is OK, as when you omit the identifying name, an index value of `0` is used by default, which refers to the first table in the Wasm module.

However, I had a question — I thought I'd try getting rid of the table identifier altogether and just using index values for this example. However, when you try defining the table without a name:

> `(table 0 funcref)`

The code doesn't work. 

Is there a way to define the table without an identifier, or can it only be omitted when using instructions like `table.size`?

files/en-us/webassembly/reference/table/grow/index.md
chrisdavidmills: Ah, so it has to have the same type as the `storage_type` specified when you first create the table. In that case, I've updated the description to:

> The initial value to set for the new table elements. Its type must be the same as the [`storage_type`](/en-US/docs/WebAssembly/Reference/Module_definitions/table#storage_type) initially set when the table was created.

files/en-us/webassembly/reference/table/size/index.md
bvisness: When defining a table (or anything else), you don't give it an explicit numeric index. The definition just gets an index based on its position in the module. (The first table will always be index 0, the second index 1, etc.) So in this case you just omit the zero.

files/en-us/webassembly/reference/table/grow/index.md
chrisdavidmills: OK, so if I said `𝟶𝚡𝙵𝙲  15:𝚞𝟹𝟸`, would that work?

bvisness: Yeah, as long as you can link to an appropriate reference for the `u32` encoding. I don't know if you have anything about that in your MDN content yet, or where exactly that would best fit. (LEB128 is used all over the wasm spec, and the spec's definition is [here](https://webassembly.github.io/spec/core/binary/values.html#binary-int). But as always it's a very technical definition that might not be very helpful.)

files/en-us/webassembly/reference/table/size/index.md
chrisdavidmills: You mean, as in `(table funcref)`? I've tried it, and that doesn't work either.

files/en-us/webassembly/reference/table/grow/index.md
chrisdavidmills: OK. I've updated the opcode references on my new pages to a structure like this:

> `𝟶𝚡𝙵𝙲 15:𝚞𝟹𝟸` ([variable-width LEB128](https://webassembly.github.io/spec/core/binary/values.html#binary-int))

If that works for you, I can roll this change out everywhere later on.

files/en-us/webassembly/reference/table/size/index.md
bvisness: Oh, apologies, I misled you there. So, the text grammar for a table definition is basically:

```
(table <identifier>? <address type=i32>? <min size> <max size>? <elem type>)
```

If you just do `(table funcref)`, you'll get an error because the min size of the table is required.

To give a concise example with and without identifiers:

```
;; with identifiers
(module
  (table $a 10 funcref)
  (table $b 20 externref)
  (func (result i32)
    table.size $a
    table.size $b
    i32.add
  )
)

;; without identifiers
(module
  (table 10 funcref)
  (table 20 externref)
  (func (result i32)
    table.size    ;; `table.size 0` would also be valid
    table.size 1
    i32.add
  )
)
```

chrisdavidmills: This makes total logical sense to me. However, looking at the example at https://pr42841.review.mdn.allizom.net/en-US/docs/WebAssembly/Reference/Table/size#observing_increases_in_table_size, if I replace

```
(table $my_table 0 funcref)
```

with

```
(table 0 funcref)
```

The example no longer works (or at least, it no longer compiles successfully, and the live example no longer displays).

Any thoughts on why?

bvisness: Hm, not sure. When I remove `$my_table` from the table declaration, and remove it from other places in the module, it works just fine. For example, here's your code running with minor modifications in the SpiderMonkey shell locally:

```
const m = new WebAssembly.Module(wasmTextToBinary(`(module
  ;; Import output function
  (import "obj" "output" (func $output (param externref) (param i32)))

  ;; Define an initially empty table of funcrefs
  (table 0 funcref)

  (func (export "run") (param $elem externref)
    ;; Grow the table by 1, setting the initial values to null.
    (table.grow
      ref.null func
      (i32.const 1)
    )
    (drop)

    ;; Call the output function, to output the table size to the DOM
    (call $output
      (local.get $elem)
      (table.size)
    )

    ;; Grow the table by 1, setting the initial values to null.
    (table.grow
      ref.null func
      (i32.const 1)
    )
    (drop)

    ;; Call the output function, to output the table size to the DOM
    (call $output
      (local.get $elem)
      (table.size)
    )
  )
)`));
const { run } = new WebAssembly.Instance(m, {
  "obj": {
    output(elem, size) {
      print(elem, size);
    },
  } },
).exports;
run("my table");

// Prints:
// my table 1
// my table 2
```

chrisdavidmills: @bvisness my turn to apologise now; I'd forgotten to remove the identifiers from the `table.grow` instructions ;-|

files/en-us/webassembly/reference/index.md
```diff
@@ -1,14 +1,40 @@
...
-title: WebAssembly instruction reference
+title: WebAssembly reference
 short-title: Reference
 slug: WebAssembly/Reference
 page-type: listing-page
 sidebar: webassemblysidebar
 ---
 
-This page lists instruction references for [WebAssembly](/en-US/docs/WebAssembly) usage on the web.
+This page lists references for [WebAssembly](/en-US/docs/WebAssembly) usage on the web.
```
eqrion:
```suggestion
This page lists references for [WebAssembly](/en-US/docs/WebAssembly).
```

Pre-existing, but I'm not sure what 'usage on the web' here means? Or maybe I just misunderstand.

files/en-us/webassembly/reference/module_definitions/table/index.md
```diff
@@ -0,0 +1,219 @@
...
+## Syntax
+
+```wat
+table identifying_name initial_size max_size storage_type
+```
+
+- `table`
+  - : The `table` instruction type. Must always be included first.
+- `identifying_name` {{optional_inline}}
+  - : An optional identifying name for the table. This must begin with a `$` symbol, for example `$my_table`. If this is omitted, the table can be identified by its index, for example `0` for the first table in the wasm script, `1` for the second, etc.
```
eqrion:
```suggestion
- `name` {{optional_inline}}
  - : An optional name for the table. This must begin with a `$` symbol, for example `$my_table`. If this is omitted, the table can be identified by its index, for example `0` for the first table in the wasm script, `1` for the second, etc.
```

Do you think we could just shorten this to 'name'?

files/en-us/webassembly/reference/module_definitions/table/index.md
```diff
@@ -0,0 +1,219 @@
...
+- `initial_size`
+  - : An integer representing the initial size of the table.
+- `max_size` {{optional_inline}}
+  - : An integer representing the maximum size the table is allowed to grow to. If this is not included, the table has no maximum size, and its growth is limited only by system constraints such as available memory.
+- `storage_type`
+  - : The name of the function type to store. Possible values are:
+    - `funcref`
+      - : Stores references to functions defined inside Wasm.
+    - `externref`
+      - : Store references to external functions defined inside JavaScript.
```
eqrion: Can we link to corresponding pages in the types section?

files/en-us/webassembly/reference/module_definitions/table/index.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Description
+
+WebAssembly tables allow storage of reference values separate from byte-oriented WebAssembly memories. The primary use-case is for storing function references that can be used with `call_indirect` to support indirect function calls for languages that have them. The `table` instruction creates a new table.
```
eqrion:
```suggestion
WebAssembly tables allow storage of reference values separate from byte-oriented WebAssembly memories. The primary use-case is for storing function references that can be used with `call_indirect` to support indirect function calls for languages that have them. The `table` definition creates a new table.
```

files/en-us/webassembly/reference/table/fill/index.md
```diff
@@ -0,0 +1,203 @@
...
+title: "fill: Wasm table instruction"
+short-title: fill
+slug: WebAssembly/Reference/Table/fill
+page-type: webassembly-instruction
+browser-compat: webassembly.reference-types
+spec-urls: https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-table
+sidebar: webassemblysidebar
+---
+
+The **`table.fill`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) sets the function reference stored in a series of table elements to the same reference.
```
eqrion:
```suggestion
The **`table.fill`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) sets a range of table elements to the same value.
```

Tables can store more than just funcrefs, so it'd be good to make the description here generic.

files/en-us/webassembly/reference/index.md
```diff
@@ -1,14 +1,40 @@
...
+## JavaScript reference
+
+- [WebAssembly](/en-US/docs/WebAssembly/Reference/JavaScript_interface)
+  - : The WebAssembly JavaScript object acts as the namespace for all WebAssembly-related functionality.
+
+## Module definitions
+
+- [WebAssembly module definitions](/en-US/docs/WebAssembly/Reference/Module_definitions)
+  - : The top-level module definitions available in Wasm for defining module features such as tables, types, memories, functions, etc.
+
```
eqrion, line 20:
```suggestion
## Definitions

- [WebAssembly definitions](/en-US/docs/WebAssembly/Reference/Definitions)
  - : The top-level definitions available in Wasm for defining module features such as tables, types, memories, functions, etc.
```

I know I referred to these as 'module definitions', but seeing it in writing now I think it'd be cleaner to just call them 'definitions'. Technically everything in wasm exists within a 'module' and so prefixing it is gonna be a bit redundant.

files/en-us/webassembly/reference/table/fill/index.md
```diff
@@ -0,0 +1,203 @@
...
+      - : The table's index number, for example `0` for the first table in the wasm script, `1` for the second, etc.
+
+    If the `identifier` is omitted, it will default to `0`.
+
+- `starting_index`
+  - : The index of the first element to store the function reference in. This must be an `i32` value, for example `(i32.const 0)`.
+- `function_reference`
+  - : The function reference to store in the table. This should be a `ref.func`, for example `(ref.func $f1)`.
+- `element_span`
+  - : The number of elements to store the function reference in.
```
eqrion: Sorry this is a long one, Ben and I thought about this a lot.

Technically, this description is mixing up two separate things that I think could be clearer if they were separated.
  1. Syntax - which is how to write the instruction, this includes the instruction 'immediates'.
  2. Type - which defines the 'operands' the instruction operates on, and the 'results' it creates.
     - in JS reference material this is equivalent to the 'Parameters' and 'Return value' section, just joined    together. Wasm usually thinks of them together, so I think that's clearer. But we could split it into two separate sections.

So for example, with table.fill the 'identifier' is an 'immediate' and defines which table the instruction operates on. 'immediates' always need to go on the instruction.

While the `starting_index function_reference element_span` are instruction operands that are inputs to the instruction, but come from elsewhere (the value stack).

The following i32.add's are all equivalent and show how there are different ways to give operands to an instruction. This is all because wasm is a 'stack machine' where instructions push and pop values from it (I'm not sure if this is documented anywhere in MDN so far, but I do see both 'folded' and 'stack' ways of writing examples).

```wat
;; push the operands manually
i32.const 1
i32.const 2
;; this pops the operands operands above, and adds them, pushing the result back on the stack
i32.add

;; use the 'folded' form, which is just a text format nicety that compiles to the previous stack based version
(i32.add
  i32.const 1
  i32.const 2)

;; the same as before but using locals
(local.set $a i32.const 1)
(local.set $b i32.const 2)
(i32.add
  local.get $a
  local.get $b)
```

Github's UI for applying suggestions is broken for me when the suggestion contains a code block, so I'm trying to workaround that here.

I tried my hand at breaking out a separate 'type' section which describes the operands/results. It uses a `[a, b, ..] -> [c, ..]` semi-standard pseudo code which describes how the instruction manipulates the wasm value stack. Basically all the values on the left side are what is popped from the stack as inputs to the instruction, all the values on the right are the result values that are pushed.

It's a simplified version of how the spec talks about `instruction types` in the their [validation rules](https://webassembly.github.io/spec/core/valid/instructions.html#xref-syntax-instructions-syntax-instr-table-mathsf-table-fill-x), which is why I call this section 'Type' (but open to a better name/organization)

Image

```
## Syntax

table.fill identifier

- `table.fill`
  - : The `table.fill` instruction type. Must always be included first.
- `identifier` {{optional_inline}}
  - : The identifier for the table you want to fill. This can be one of the following:
    - `identifying_name`
      - : An identifying name [set for the table](/en-US/docs/WebAssembly/Reference/Module_definitions/table#identifying_name) when it was first created. This must begin with a `$` symbol, for example `$my_table`.
    - `index` {{optional_inline}}
      - : The table's index number, for example `0` for the first table in the wasm module, `1` for the second, etc.

    If the `identifier` is omitted, it will default to `0`.

## Type

[index, value, length] -> []

- `index`
  - : The index of the first element to store the reference in. This must be an `i32` value, for example `(i32.const 0)`.
- `value`
  - : The reference to store in the table. This must be the same reference type that the table is defined with. (link to the table definition page maybe?)
- `length`
  - : The number of elements to store the value to, starting at `index`. This must be an `i32` value.
```

If we got with the above, we could replace 'Result' section, as it'd be covered by the 'Type' section.

files/en-us/webassembly/reference/table/fill/index.md
```diff
@@ -0,0 +1,203 @@
...
+### Opcodes
+
+| Instruction  | Binary opcode                                                                                                  |
+| ------------ | -------------------------------------------------------------------------------------------------------------- |
+| `table.fill` | `𝟶𝚡𝙵𝙲 17:𝚞𝟹𝟸` ([variable-width LEB128](https://webassembly.github.io/spec/core/binary/values.html#binary-int)) |
+
+## Description
+
+`table.fill` is used to fill multiple table elements with the same function reference.
+
```
eqrion: Is this redundant with the description at the very beginning? If we do keep it, I think it'd go well next to the 'Syntax' section, either before or after.

So we'd have a template for instructions of:
  1. Description
  2. Syntax
  3. Type
  4. (could be added someday) Traps
    - Equivalent to JS Exceptions section which describes when the instruction performs a trap (which aborts wasm execution for things like OOBs or null).

files/en-us/webassembly/reference/table/get/index.md
```diff
@@ -0,0 +1,192 @@
...
+  - : An identifier for the table you want to retrieve a reference from. This can be one of the following:
+    - `identifying_name`
+      - : An identifying name [set for the table](/en-US/docs/WebAssembly/Reference/Module_definitions/table#identifying_name) when it was first created. This must begin with a `$` symbol, for example `$my_table`.
+    - `index`
+      - : The table's index number, for example `0` for the first table in the wasm script, `1` for the second, etc.
+
+    If the `identifier` is omitted, it will default to `0`.
+
+- `element_number`
+  - : The element number to retrieve the function reference from. This must be an `i32` value, for example `(i32.const 1)`.
```
eqrion: Using the above suggested template layout for instructions:

```
## Type

[index] -> [value]

- `index`
  - : The index in the table to retrieve the value of. This must be an `i32` value, for example `(i32.const 1)`.
- `value`
  - : The value stored in the table at the index.
```

And we could drop the 'result' section because that's covered here.

files/en-us/webassembly/reference/table/grow/index.md
```diff
@@ -0,0 +1,227 @@
...
+  - : The initial value to set for the new table elements. Its type must be the same as the [`storage_type`](/en-US/docs/WebAssembly/Reference/Module_definitions/table#storage_type) initially set when the table was created.
+- `grow_amount`
+  - : The number of elements to grow the table by. This must be an `i32` value, for example `(i32.const 1)`.
+
+### Result
+
+An `i32` equal to the size of the table before the `grow` instruction is applied to it, or `-1` if the table failed to grow, for example due to an out-of-memory (OOM) error or the new size being greater than the [table's maximum size](/en-US/docs/WebAssembly/Reference/Module_definitions/table#max_size).
+
+To retrieve the new table size after the `grow` instruction is applied to it, use the [`table.size`](/en-US/docs/WebAssembly/Reference/Table/size) instruction.
```
eqrion:
```
## Type

[initial_value, grow_amount] -> [previous_length]

- `initial_value`
  - : The initial value to set for the new table elements. Its type must be the same as the [`storage_type`](/en-US/docs/WebAssembly/Reference/Module_definitions/table#storage_type) initially set when the table was created.
- `grow_amount`
  - : The number of elements to grow the table by. This must be an `i32` value, for example `(i32.const 1)`.
- `previous_length`
  - : An `i32` equal to the size of the table before the `grow` instruction is applied to it, or `-1` if the table failed to grow, for example due to an out-of-memory (OOM) error or the new size being greater than the [table's maximum size](/en-US/docs/WebAssembly/Reference/Module_definitions/table#max_size).

To retrieve the new table size after the `grow` instruction is applied to it, use the [`table.size`](/en-US/docs/WebAssembly/Reference/Table/size) instruction.
```

files/en-us/webassembly/reference/table/set/index.md
```diff
@@ -0,0 +1,271 @@
...
+title: "set: Wasm table instruction"
+short-title: set
+slug: WebAssembly/Reference/Table/set
+page-type: webassembly-instruction
+browser-compat: webassembly.reference-types
+spec-urls: https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-table
+sidebar: webassemblysidebar
+---
+
+The **`table.set`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) changes the function reference stored in a particular table element.
```
eqrion:
```suggestion
The **`table.set`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) changes the value stored in a particular table element.
```

files/en-us/webassembly/reference/table/set/index.md
```diff
@@ -0,0 +1,271 @@
...
+    If the `identifier` is omitted, it will default to `0`.
+
+- `element_number`
+  - : The element number to store the function reference in. This must be an `i32` value, for example `(i32.const 1)`.
+- `function_reference`
+  - : The function reference to store in the table. This should be a `ref.func`, for example `(ref.func $f1)`.
+
+### Result
+
+No result.
```
eqrion:
```
### Type

[index, value] -> []

- `index`
  - : The element index to store the value in. This must be an `i32` value, for example `(i32.const 1)`.
- `value`
  - : The value to store in the table. This must match the element type the table was defined with.
```

files/en-us/webassembly/reference/table/size/index.md
```diff
@@ -0,0 +1,176 @@
...
+      - : An identifying name [set for the table](/en-US/docs/WebAssembly/Reference/Module_definitions/table#identifying_name) when it was first created. This must begin with a `$` symbol, for example `$my_table`.
+    - `index`
+      - : The table's index number, for example `0` for the first table in the wasm script, `1` for the second, etc.
+
+    If the `identifier` is omitted, it will default to `0`.
+
+### Result
+
+An `i32` equal to the current number of elements contained inside the table.
+
```
eqrion:
```suggestion
### Type

[] -> [length]

- length: An `i32` equal to the current number of elements contained inside the table.

```

eqrion (commented):
Apologies for leaving more big comments. Ben and I are figuring out as we go what we think a good layout for instruction reference would be. It's difficult to balance being accurate but also not going into so much detail that it becomes the spec. I think we have a good idea now, but open to your thoughts on it.

files/en-us/webassembly/reference/index.md
chrisdavidmills: Yes, I agree with this change. It is unnecessary and inaccurate.

files/en-us/webassembly/reference/index.md
chrisdavidmills: No worries, this makes sense to me. I've updated the PR so that this is changed to "Definitions" everywhere.

files/en-us/webassembly/reference/table/fill/index.md
chrisdavidmills: Thanks so much for this, @eqrion, and @bvisness — this makes sense to me. We sort of cover these syntax details here: https://developer.mozilla.org/en-US/docs/WebAssembly/Guides/Understanding_the_text_format#signatures_and_parameters. But it is not really that clear.

If this structure makes more sense for Wasm reference docs, then I am happy to implement it

chrisdavidmills: Question — Is the same terminology suitable for the Definitions reference pages?

files/en-us/webassembly/reference/table/fill/index.md
chrisdavidmills: This is usually where the "Description" section goes in MDN reference pages, and I'd prefer not to deviate from the standard structure more than is necessary. However, saying that, in this case the "Description" is not needed — it is just repeating what is in the page intro. I will delete it from here, and from other pages where it is redundant. In some cases, it provides more information.

I am quite interested in including the "Traps" section where appropriate (I'm assuming that not all instructions have traps?). Where can I find out detailed information about each instruction, its syntax, and its traps? The Wasm spec isn't as obviously browsable as JS API specs, but maybe this has more to do with my knowledge limitations.

files/en-us/webassembly/reference/table/get/index.md
chrisdavidmills: updated, thanks.

files/en-us/webassembly/reference/table/grow/index.md
chrisdavidmills: looking good; updated.

files/en-us/webassembly/reference/table/set/index.md
chrisdavidmills: updated, thanks

files/en-us/webassembly/reference/table/set/index.md
chrisdavidmills: updated

files/en-us/webassembly/reference/table/size/index.md
chrisdavidmills: updated, thanks!

files/en-us/webassembly/reference/table/fill/index.md
chrisdavidmills: done

files/en-us/webassembly/reference/module_definitions/table/index.md
chrisdavidmills: Yes, done.

files/en-us/webassembly/reference/module_definitions/table/index.md
chrisdavidmills: Yes. I've updated it here, and I've also updated all the links to it in other pages.

files/en-us/webassembly/reference/table/get/index.md
```diff
@@ -0,0 +1,198 @@
...
+### Type
+
+```plain
+[index] -> [value]
+```
+
+- `index`
+  - : The index in the table to retrieve the value of. This must be an `i32` value, for example `(i32.const 1)`.
+- `value`
+  - : The value stored in the table at the index.
```
eqrion:
```suggestion
  - : The value stored in the table at the index. This will be the element type of the table.
```

files/en-us/webassembly/reference/table/fill/index.md
eqrion: > Question — Is the same terminology suitable for the Definitions reference pages?

The short answer is that I think the Definitions reference pages use of 'Syntax' is correct. All the things you're listing there are a part of the 'syntax' of defining a table in wasm.

To be very precise, the definitions in a module also have a 'type' that you could talk about, but it's a little different than instruction types. Tables, for example, have a type which can be written as `(table min max elem-type)`. When you write a table definition, you're technically writing the type inline.

Where this matters is imports. A table can be defined inside a module, or imported. When you import a table, you just specify the type of table you want. When you define a table, you specify the type of the table and also any other details that aren't part of the table type. In Wasm 1.0, I don't think this was anything, but with the function-references proposal, tables could specify a default value for table elements on the table.

```
;; importing a table with a type that has an initial 10 funcref elements
(import "" "" (table 10 funcref))

;; define a table with the same type as the import
;; but also specify that the default value of the 10 funcrefs isn't null, but actually a specific function
;; this extra info isn't part of the type, it's a 'private' detail of the table that doesn't affect imports
(table 10 funcref (ref.func $defaultValue))
```

For right now, I don't think this level of detail is needed on MDN (unless you're interested in doing it!). But if we did want to document it we would add a 'Type' section on Definition pages.

files/en-us/webassembly/reference/table/fill/index.md
eqrion: That makes sense, sounds good.

Yes, not all instructions have traps. I'll try to call out any traps when reviewing things.

This information is all in the core wasm spec, but it's spread out in different sections and not easily accessible unless you know the PL formalisms. For traps, they are part of the runtime behavior for an instruction and so you need to go to 'Execution > Instructions' to find it. For example, call_ref will trap if the input is null [1]. There are step-by-step instructions for how execution works, and you'll see 'Trap' clauses there.

[1] https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-control-mathsf-call-ref-x

files/en-us/webassembly/reference/table/fill/index.md
chrisdavidmills: Thanks for the explanation. I think I'll leave this for now and add some details if people feel they really need it.

chrisdavidmills:
> This looks great. I can't approve it though because I don't have permissions, otherwise I would.

@eqrion No worries, this is approval enough; now I will move it onto the editorial review stage.

I will also have a go at adding the traps sections to the table instruction pages. I'll let you know when that is done so you cna have a look.

chrisdavidmills:
@eqrion OK, "Traps" sections added to `table.[fill|get|set]`

eqrion:
I double checked and those are accurate.

chrisdavidmills:
@MendyBerger OK, I've opened this up for editorial review. Let me know if you have time to look this stuff over. No pressure if not.

files/en-us/webassembly/reference/table/size/index.md
```diff
@@ -0,0 +1,181 @@
...
+title: "size: Wasm table instruction"
+short-title: size
+slug: WebAssembly/Reference/Table/size
+page-type: webassembly-instruction
+browser-compat: webassembly.reference-types
+spec-urls: https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-table
+sidebar: webassemblysidebar
+---
+
+The **`table.size`** [Table instruction](/en-US/docs/WebAssembly/Reference/Table) returns the current number of elements contained in a table.
```
MendyBerger: Nitpick: "number of elements contained in a table" might read like the non-null elements of the table. Consider if something like "size of the table" is clearer.

files/en-us/webassembly/reference/types/index.md
```diff
@@ -0,0 +1,15 @@
+---
+title: WebAssembly types
+slug: WebAssembly/Reference/Types
+page-type: landing-page
+sidebar: webassemblysidebar
+---
+
+This section documents the different WebAssembly types.
+
+## Value types
```
MendyBerger: These only include reference types. Consider changing the title to "Reference types" unless you'd like to see primitive types listed here as well.

files/en-us/webassembly/reference/definitions/table/index.md
```diff
@@ -0,0 +1,229 @@
...
+(elem declare func $f1)
+```
+
+To call a function referenced in a table, you have to reference the table and the index value the function reference is stored at. The following example uses `call_indirect`:
+
+```wat
+(call_indirect (type $ret_i32) (local.get $index))
+```
+
+It is possible to mutate tables at runtime using instructions like [`table.set`](/en-US/docs/WebAssembly/Reference/Table/get) and [`table.fill`](/en-US/docs/WebAssembly/Reference/Table/get), and retrieve values using [`table.get`](/en-US/docs/WebAssembly/Reference/Table/get).
```
MendyBerger: The links all point to table.get

files/en-us/webassembly/reference/types/externref/index.md
```diff
@@ -0,0 +1,150 @@
...
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
+xx
```
MendyBerger: What's this?

files/en-us/webassembly/reference/types/funcref/index.md
```diff
@@ -0,0 +1,157 @@
...
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
+xx
```
MendyBerger: What's this?

files/en-us/webassembly/reference/definitions/table/index.md
```diff
@@ -0,0 +1,229 @@
...
+- `initial_size`
+  - : An integer representing the initial size of the table.
+- `max_size` {{optional_inline}}
+  - : An integer representing the maximum size the table is allowed to grow to. If this is not included, the table has no maximum size, and its growth is limited only by system constraints such as available memory.
+- `type`
+  - : The name of the function type to store. Possible values are:
+    - [`funcref`](/en-US/docs/WebAssembly/Reference/Types/funcref)
+      - : Stores references to functions defined inside Wasm.
+    - [`externref`](/en-US/docs/WebAssembly/Reference/Types/externref)
+      - : Store references to external functions defined inside JavaScript.
```
MendyBerger: You mean external 'values', right?

files/en-us/webassembly/reference/definitions/table/index.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+But you could instead reference the tables by their index values (`0` specifies the first table in the module, `1` the second table, etc.):
+
+```wat
+(call_indirect 0 (type $ret_i32) (i32.const 0))
+(call_indirect 1 (type $ret_i32) (i32.const 0))
+```
+
+If you don't specify an indentifying name _or_ an index, the index `0` is assumed:
```
MendyBerger: Typo: "indentifying"

files/en-us/webassembly/reference/definitions/table/index.md
```diff
@@ -0,0 +1,229 @@
...
+(table 2 funcref)
+```
+
+Optionally, you can also provide an identifier, which can be used to identify the table elsewhere, and a maximum growth size. For example:
+
+```wat
+(table $mytable 2 10 funcref)
+```
+
+The following defines a function type, defines a basic function with that type that returns an `i32`, and and forward-declares it using `(elem declare func $f1)` so it can be referenced later on.
```
MendyBerger: Typo: "and and"

MendyBerger (approved):
This is a really well written PR, thanks @chrisdavidmills for working on this!

I didn't do a line by line deep dive, but I read thought all of it at least once and it looks great!

Added some mostly nitpick comments. Overall, seems ready to go.

As an aside: some instructions/keywords like `elem`, `declare`, and `func.ref` are mentioned in passing but don't have dedicated pages. Might be worth adding pages for them at a later point.

chrisdavidmills:
Thanks so much for the review, @MendyBerger.

> As an aside: some instructions/keywords like `elem`, `declare`, and `func.ref` are mentioned in passing but don't have dedicated pages. Might be worth adding pages for them at a later point.

Yes, definitely! The plan is to tackle all the missing "new" Wasm stuff first, but after that, we're intending to go back and give all the missing fundamental stuff the same treatment.

files/en-us/webassembly/reference/table/size/index.md
chrisdavidmills: Sure, makes sense. Updated to "...returns the current size of the table." in my next commit.

files/en-us/webassembly/reference/types/index.md
chrisdavidmills: Changed to "Reference types" in the next commit. I think we can add a separate heading for "Primitive types" in a future PR, and it will work out fine.

files/en-us/webassembly/reference/definitions/table/index.md
chrisdavidmills: Ack! Good spot, fixed!

files/en-us/webassembly/reference/types/externref/index.md
chrisdavidmills: Ooops, I forgot to fill in any links there. I've added some minimal suitable links in.

files/en-us/webassembly/reference/types/funcref/index.md
chrisdavidmills: Same here.  I've added some minimal suitable links in.

I've also checked the other new pages to make sure there aren't any placeholders left.

files/en-us/webassembly/reference/definitions/table/index.md
chrisdavidmills: Yes! Good spot, changed in next commit.

files/en-us/webassembly/reference/definitions/table/index.md
chrisdavidmills: Fixed, ta!

files/en-us/webassembly/reference/definitions/table/index.md
chrisdavidmills: Fixed, ta!

chrisdavidmills:
Note: I also took the opportunity in this PR to rework the structure of the "Guides" part of the sidebar, as it was unorganized and annoying me. We will, of course, rework it further when we come to write new guides, but this looks better for now.

PR #42848, opened 2026-01-21
Update timeline-range-name to include keyframe selectors

estelle:
https://drafts.csswg.org/scroll-animations/#named-range-keyframes


chrisdavidmills (approved):
LGTM! Thanks @estelle.

PR #42858, opened 2026-01-21
Clarify that arrow functions do not need to return a value

### LINKED ISSUE #42839: Page implies that arrow functions MUST always return a value
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

"Function Body" and others / generally

As a newcomer to arrow functions, after reading the entire page, I was left with the impression that an arrow function must always return a value.  I did not find any place that helped me to understand that it is perfectly legal and valid to omit "return" from the expression block to have a purely bahavioural function.

From "Function body" section:

"In an expression body, only a single expression is specified, which becomes the implicit return value. In a block body, you must use an explicit return statement."

In the "Description" section, there is no purely behavioural example.

In the "Examples" section, there are a couple examples, but I didn't notice them and was left with the impression that it's illegal NOT to return a value.

This could be considered subjective, and that I just read it wrong (and i accept that it might be true!).  But the impression that arrow functions must always return a value was left so strongly with me, and it took me awhile to find other sources to suggest that was wrong, that I thought it was worth reporting.

An example near the top of the page showing an arrow function using an expression block with braces and no return statement I think would be very helpful. Or even explicitly written that it's valid to use braces with no return value.

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/javascript/reference/functions/arrow_functions`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/functions/arrow_functions/index.md

</details>

SURYAS1306:
### Description
Adds a clarification and example showing that arrow functions with block bodies are not required to return a value and will return 'undefined' if no return statement is present.

### Motivation
The current page can give newcomers the impression that arrow functions must always return a value. This change directly addresses the confusion reported in issue #42839.

### Related issues
Fixes #42839

### DIFF
--- a/files/en-us/web/javascript/reference/functions/arrow_functions/index.md
+++ b/files/en-us/web/javascript/reference/functions/arrow_functions/index.md
@@ -113,7 +113,7 @@ const b = 2;
 () => a + b + 100;
 ```
 
-The braces can only be omitted if the function directly returns an expression. If the body has statements, the braces are required — and so is the `return` keyword. Arrow functions cannot guess what or when you want to return.
+The braces can only be omitted if the function directly returns an expression. If the body has statements, the braces are required. In this case, return values must be explicitly specified with the `return` keyword. Arrow functions cannot guess what or when you want to return.
 
 ```js
 // Traditional anonymous function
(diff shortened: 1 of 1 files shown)


SURYAS1306:
Hi MDN team, 
This PR adds a short clarification and example showing that arrow functions with block bodies do not have to return a value and will return 'undefined' if no return is present, as discussed in #42839.  

Thanks for reviewing.

Josh-Cena (approved):
Thanks

PR #42866, opened 2026-01-22
Fix allowable properties for ::highlight pseudo-element

### LINKED ISSUE #42863: ::highlight() page mentions `-webkit-*` as allowable property, but spec doesn't mention these
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::highlight

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::highlight#allowable_properties

The section includes

> [-webkit-text-stroke-color](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/-webkit-text-stroke-color), [-webkit-text-fill-color](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/-webkit-text-fill-color) and [-webkit-text-stroke-width](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/-webkit-text-stroke-width)

But these don't seem to be mentioned [in the spec](https://drafts.csswg.org/css-pseudo/#highlight-styling).

Expected only properties mentioned in the spec to be listed.

Originally reported by @flying-sheep via https://github.com/mdn/browser-compat-data/issues/28852.

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/css/reference/selectors/_doublecolon_highlight`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::highlight
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/reference/selectors/_doublecolon_highlight/index.md

</details>

SURYAS1306:
###Description

This PR corrects the “Allowable properties” section for the ::highlight pseudo-element.
It removes vendor-prefixed properties and updates the list to use the standard stroke-color, fill-color, and stroke-width properties.

###Motivation

The MDN page currently lists -webkit-text-* properties as allowable, which does not match the CSS specification.
This change aligns the documentation with the spec and prevents developers from relying on unsupported properties.

###Related issues and pull requests

Fixes #42863


files/en-us/web/css/reference/selectors/_doublecolon_highlight/index.md
```diff
@@ -20,7 +20,9 @@ Only certain CSS properties can be used with `::highlight()`:
 - {{CSSxRef("background-color")}}
 - {{CSSxRef("text-decoration")}} and its associated properties
 - {{CSSxRef("text-shadow")}}
-- {{CSSxRef("-webkit-text-stroke-color")}}, {{CSSxRef("-webkit-text-fill-color")}} and {{CSSxRef("-webkit-text-stroke-width")}}
+- {{CSSxRef("stroke")}}, {{CSSxRef("fill")}}, and {{CSSxRef("stroke-width")}}
+
+Vendor-prefixed properties such as {{CSSxRef("-webkit-text-stroke-color")}} or {{CSSxRef("-webkit-text-fill-color")}} do not apply to highlight pseudo-elements.
```
estelle:
```suggestion
- {{CSSxRef("stroke-color")}}, {{CSSxRef("fill-color")}}, and {{CSSxRef("stroke-width")}}
- Custom properties ({{cssxref("--*"))}}
```

estelle (changes requested):
made a suggestion. 
if the -webkit- aren't in the spec, we don't need to mention them

SURYAS1306:
Hi @estelle , I’ve removed the vendor-prefixed properties and kept only the spec-defined allowable properties. Please let me know if this looks good now. Thanks!

estelle:
> Hi @estelle , I’ve removed the vendor-prefixed properties and kept only the spec-defined allowable properties. Please let me know if this looks good now. Thanks!

I haven't tested them since they props only apply to svg, but the spec reads `stroke-color` and the pr has `stroke`. Also, the spec lists custom properties as supported props.

flying-sheep:
@SURYAS1306 can you make these updates or should someone else fix this?

SURYAS1306:
Hi @estelle, 
I've updated the properties to stroke-color and fill-color as per the spec, and I also added the line for Custom Properties. Let me know if there's anything else.

@flying-sheep I have applied the updates

pepelsbey:
Hey! Thank you for contributing 🙂

Unfortunately, it’s a little complicated:

- Only Safari supports the `stroke-color` property (currently not documented on MDN, unfortunately)
- No browser supports the `fill-color` property. It only exists in the spec for now. 
- Only the `stroke-width` property is fine, well-supported, and documented.

So I wonder if we shouldn’t mention the `fill-color` property at all in this list, as it’s purely theoretical.

@estelle what do you think?

files/en-us/web/css/reference/selectors/_doublecolon_highlight/index.md
```diff
@@ -20,7 +20,8 @@ Only certain CSS properties can be used with `::highlight()`:
 - {{CSSxRef("background-color")}}
 - {{CSSxRef("text-decoration")}} and its associated properties
 - {{CSSxRef("text-shadow")}}
-- {{CSSxRef("-webkit-text-stroke-color")}}, {{CSSxRef("-webkit-text-fill-color")}} and {{CSSxRef("-webkit-text-stroke-width")}}
+- {{CSSxRef("stroke-color")}}, {{CSSxRef("fill-color")}}, and {{CSSxRef("stroke-width")}}
```
estelle, line 24:
```suggestion
- {{CSSxRef("stroke-color")}} and {{CSSxRef("stroke-width")}}
```
I tested a bit.
https://codepen.io/estelle/pen/LEZqKxJ
-webkit-text-fill-color doesn't work anywhere either i don't think, so removing what is currently there is OK.
should we mention `fill-color` either? I say No.
made a suggestion to remove it.

flying-sheep: As said, the spec explicitly says that vendor-prefixed properties (like `-webkit-*`) are *not* supported for `::highlight`, so removing it is not only OK but definitely the correct thing to do.

pepelsbey: > As said, the spec explicitly says that vendor-prefixed properties (like -webkit-*) are not supported

It's always good to double check whether it actually works. MDN is documenting browser behavior, not the spec.

flying-sheep: Got it. Still, the current state (before this PR) is very confusing, as it advertises something that neither exists nor is supposed to exist.

flying-sheep:
can this be merged? the MDN page continues to be confusing and misleading without this.

Josh-Cena:
Hi @pepelsbey @estelle. I'm not sure exactly what this PR is blocked on. As Vadim already pointed out, `fill-color` should definitely be in the list because it's in the spec. It's "unsupported" because the entire property is unsupported. So really there's no problem with aligning with the spec.

The problem is whether we want to "document the spec" or "document the browser". In fact our stance is nearly always "document the spec" unless something is a new spec thing that no browser has caught up to yet, which vendor prefixes by definition cannot be because they aren't baseline to start with. I have pushed a version with a big fat note containing all the gory details of actual browser support; IMO none of this should be in content, but I don't know where we can put it in BCD. The support reality may be more complicated than you would think. But if "document the browser" is what you want then that is it.

Anyway, my personal vote is to remove that note and move on. Pretend all the browser reality doesn't exist and hope they catch up to the spec. Any opinions?

pepelsbey:
> Pretend all the browser reality doesn't exist and hope they catch up to the spec

Huh, no. I strongly disagree. I think we should document web platform's state that's actually useful for developers. That's why we have the BCD table and Baseline banners and labels.

And as far as I remember this is what we've been doing, at least in the last few years since I joined the project.

The only spec-driven pages we have in CSS are the module pages, otherwise we don't document things that are not supported. That's [what we say](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/What_we_write#topics_that_belong_on_mdn_web_docs) in writing guidelines.

chrisdavidmills:
I agree with @pepelsbey here. I've been on the project for 13 years, and this is always what we've leaned towards, in my experience.

Josh-Cena:
Let me rephrase: we document *interoperable* browser realities. If all browsers do A but spec says B, we document A, optionally mentioning B in a note. If some browsers do A but others do B, then we document B and put A in BCD.

Does this note do what you would like to see?

Josh-Cena:
I don't think any of these principles can translate to "we should remove `fill-color` from the list" anyway, despite no browser supporting it as a CSS property. This section essentially means "if X is supported as a property, then X should be supported in `::highlight()`". The precondition being false makes the whole statement vacuously true. Practically, removing it makes it very hard to update in the future.

Issue #42871, opened 2026-01-22
Decription of behaviour without the `X-Frame-Options` header is missing


doerwalter:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options

### What information was incorrect, unhelpful, or incomplete?

The page is missing a description of the browser behaviour when the `X-Frame-Options` header is not present.

### What did you expect to see?

An explanation of the behaviour when the `X-Frame-Options` header is not present.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/http/reference/headers/x-frame-options`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/reference/headers/x-frame-options/index.md
* Last commit: https://github.com/mdn/content/commit/7f6778934020a9b5b82b4dd8ca79a99bc9950c2a
* Document last modified: 2025-11-21T00:51:23.000Z

</details>

PR #42874, opened 2026-01-22
Fix issue 42871

wbamberg:
Fixes https://github.com/mdn/content/issues/42871.

@doerwalter, does this address the issue for you?

I took out the bit about "it only works if the browser supports it" because surely that's obvious.

### DIFF
--- a/files/en-us/web/http/reference/headers/x-frame-options/index.md
+++ b/files/en-us/web/http/reference/headers/x-frame-options/index.md
@@ -33,9 +33,9 @@ X-Frame-Options: SAMEORIGIN
 ### Directives
 
 - `DENY`
-  - : The page cannot be displayed in a frame, regardless of the site attempting to do so. Not only will the browser attempt to load the page in a frame fail when loaded from other sites, attempts to do so will fail when loaded from the same site.
+  - : The document cannot be loaded in any frame, regardless of origin (both same- and cross-origin embedding is blocked).
 - `SAMEORIGIN`
-  - : The page can only be displayed if all ancestor frames have the same {{glossary("origin")}} as the page itself. You can still use the page in a frame as long as the site including it in a frame is the same as the one serving the page.
+  - : The document can only be embedded if all ancestor frames have the same {{glossary("origin")}} as the page itself.
 - `ALLOW-FROM origin` {{deprecated_inline}}
   - : This is an obsolete directive. Modern browsers that encounter response headers with this directive will ignore the header completely. The {{HTTPHeader("Content-Security-Policy")}} HTTP header has a {{HTTPHeader("Content-Security-Policy/frame-ancestors", "frame-ancestors")}} directive which you should use instead.
 


files/en-us/web/http/reference/headers/x-frame-options/index.md
```diff
@@ -10,9 +10,9 @@ sidebar: http
...
+The HTTP **`X-Frame-Options`** {{Glossary("response header")}} can be used to indicate whether a browser should be allowed to render the document in a {{HTMLElement("frame")}}, {{HTMLElement("iframe")}}, {{HTMLElement("embed")}} or {{HTMLElement("object")}}. Sites can use this to avoid [clickjacking](/en-US/docs/Web/Security/Attacks/Clickjacking) attacks, by ensuring that their content is not embedded into other sites.
 
-The added security is provided only if the user accessing the document is using a browser that supports `X-Frame-Options`.
+If this header is not sent, and the website has not implement any other mechanisms to restrict embedding (such as the {{HTTPHeader("Content-Security-Policy/frame-ancestors", "frame-ancestors")}} CSP directive), then the browser will allow other sites to embed this document.
```
wbamberg:
```suggestion
If this header is not sent, and the website has not implemented any other mechanisms to restrict embedding (such as the {{HTTPHeader("Content-Security-Policy/frame-ancestors", "frame-ancestors")}} CSP directive), then the browser will allow other sites to embed this document.
```

doerwalter:
This looks good to me. Thanks!

files/en-us/web/http/reference/headers/x-frame-options/index.md
```diff
@@ -33,9 +33,9 @@ X-Frame-Options: SAMEORIGIN
 ### Directives
 
 - `DENY`
-  - : The page cannot be displayed in a frame, regardless of the site attempting to do so. Not only will the browser attempt to load the page in a frame fail when loaded from other sites, attempts to do so will fail when loaded from the same site.
+  - : The document cannot be loaded in any frame, regardless of origin (both same- and cross-origin embedding is blocked).
```
hamishwillee, line 36: Additional text was redundant.

files/en-us/web/http/reference/headers/x-frame-options/index.md
```diff
@@ -33,9 +33,9 @@ X-Frame-Options: SAMEORIGIN
...
 - `SAMEORIGIN`
-  - : The page can only be displayed if all ancestor frames have the same {{glossary("origin")}} as the page itself. You can still use the page in a frame as long as the site including it in a frame is the same as the one serving the page.
+  - : The document can only be embedded if all ancestor frames have the same {{glossary("origin")}} as the page itself.
```
hamishwillee, line 38: Additional text was inconsistent with the first statement - you can only embed content if **all** ancestors are sameorigin, not just the parent.

wbamberg:
> @wbamberg I would probably have argued that it is obvious that not having the header would allow embedding - otherwise why would you have it. Does no harm though and definitely not worth mentioning the "if supported" thing.

Yeah, I don't think it hurts to indicate what happens without it. Your other edits look great to me too, I didn't even check that.

PR #42876, opened 2026-01-22
New guide: CSS @property

estelle:
Made the `@property` ref page be more like the other reference pages.
created a new guide explaining the at-rule and registering custom props in general
added links to the new page.

### DIFF
--- a/files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
+++ b/files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
@@ -9,7 +9,7 @@ The `@property` at-rule, part of the [CSS Houdini](/en-US/docs/Web/API/Houdini_A
 
 ## Why register custom properties?
 
-CSS custom properties, commonly called CSS variables, can be defined by simple declaration. Generally, these are declared on the {{cssxref(":root")}}, and can be overridden with a different value in subsequent rulesets, including in ruled based on UI state and within container and media queries.
+CSS custom properties, commonly called CSS variables, can be defined by basic declaration. Generally, these are declared on the {{cssxref(":root")}}, and can be overridden with a different value in subsequent rulesets — including in rules based on UI state and within container and media queries.
 
 In this example, `--myColor` is assigned the value `#bada55` on `:root`, then set to a different value on `main`, which is inherited by all descendants of `main`. That value is overridden when `main` is hovered or is in a viewport that is less than `750px` wide. Within elements with a class of `circles`, `--myColor` is set to `45deg`, which is a different data type.
 
@@ -123,7 +123,7 @@ Here we use the CSS `@property` at-rule to define the `--item-size` custom prope
 }
 ```
 
-You can also use [JavaScript](/en-US/docs/Web/JavaScript) to register properties. JavaScript's {{domxref('CSS.registerProperty_static', 'CSS.registerProperty()')}} method is equivalent to `@property` at-rule. Here we defined our second custom property, `--item-color`, to have an initial value of `aqua`, to accept only {{cssxref("&lt;color&gt;")}} values, and to not be inherited.
+You can also use [JavaScript](/en-US/docs/Web/JavaScript) to register properties. JavaScript's {{domxref('CSS.registerProperty_static', 'CSS.registerProperty()')}} method is equivalent to the `@property` at-rule. Here we defined our second custom property, `--item-color`, to have an initial value of `aqua`, to accept only {{cssxref("&lt;color&gt;")}} values, and to not be inherited.
 
 ```js
 window.CSS.registerProperty({


files/en-us/web/api/css_properties_and_values_api/guide/index.md
```diff
@@ -110,3 +110,7 @@ Second, unlike standard properties, registered properties aren't validated when
 ## Browser compatibility
 
 {{Compat}}
+
+## See also
+
+- [Registering custom properties with CSS](/en-US/docs/Web/CSS/Guides/Properties_and_values_API/Registering_properties)
```
chrisdavidmills, line 116: Just a little warning note that the actual title on the article itself is "Registering custom properties in CSS", whereas here you're using "Registering custom properties with CSS", and then in the next article the link text is "Registering CSS custom properties" ;-)

I won't bring this one up again; just decide on one and stick to it throughout. My vote would be for "Registering CSS custom properties", because during the article, you also discuss how to register them via JS.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
+---
+title: Registering custom properties in CSS
```
chrisdavidmills, line 2: Question — this new guide overlaps somewhat with the existing guide at https://pr42876.review.mdn.allizom.net/en-US/docs/Web/API/CSS_Properties_and_Values_API/guide. Do we need both? Are you intending to retire the older one once this newer one is published?

files/en-us/web/css/guides/properties_and_values_api/index.md
```diff
@@ -80,6 +80,8 @@ The value of `--stop-color` is set to `cornflowerblue` at first, but when you ho
 
 ## Guides
 
+- [Registering CSS custom properties](/en-US/docs/Web/CSS/Guides/Properties_and_values_API/Registering_properties)
+  - : The usages and benefits of registering CSS custom properties with the `@property` at-rule and its descriptors.
```
chrisdavidmills:
```suggestion
  - : Covers how to register CSS custom properties with the `@property` at-rule and its descriptors and the associated JavaScript API, and explains the benefits of doing so.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+slug: Web/CSS/Guides/Properties_and_values_API/Registering_properties
+page-type: guide
+sidebar: cssref
+---
+
+The `@property` at-rule, part of the [CSS Houdini](/en-US/docs/Web/API/Houdini_APIs) set of APIs, allows developers to explicitly define [CSS custom properties](/en-US/docs/Web/CSS/Reference/Properties/--*). In this guide, we learn how to use `@property` for type checking and constraining, using CSS to set default values, and define whether a custom property can inherit values or not.
```
chrisdavidmills:
```suggestion
The `@property` at-rule, part of the [CSS Houdini](/en-US/docs/Web/API/Houdini_APIs) set of APIs, allows developers to explicitly define [CSS custom properties](/en-US/docs/Web/CSS/Reference/Properties/--*). In this guide, we learn how to use `@property` for type checking and constraining, set default custom property values in CSS, and define whether a custom property can inherit values or not.
```

Again, mention the JavaScript API? You do cover it below.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Why register custom properties
+
+CSS custom properties, commonly called CSS variables, can be defined by simple declaration. Generally, these are declared on the {{cssxref(":root")}}, and can be overridden with a different value scoped to the content matched by the selector, even based state or by container or media queries:
```
chrisdavidmills:
```suggestion
CSS custom properties, commonly called CSS variables, can be defined by simple declaration. Generally, these are declared on the {{cssxref(":root")}}, and can be overridden with a different value in subsequent rulesets, even based state or by container or media queries:
```

Two things:

1. "scoped to the content matched by the selector" seems wordy and tricky to parse, but also unnecessarily obvious, hence suggesting a change. It's just how selectors work anyway?
2.  I'm not sure what you mean by "even based state or by container or media queries". Is there a word missing here? Seems like it needs rephrasing.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
+---
+title: Registering custom properties in CSS
+slug: Web/CSS/Guides/Properties_and_values_API/Registering_properties
+page-type: guide
+sidebar: cssref
+---
+
+The `@property` at-rule, part of the [CSS Houdini](/en-US/docs/Web/API/Houdini_APIs) set of APIs, allows developers to explicitly define [CSS custom properties](/en-US/docs/Web/CSS/Reference/Properties/--*). In this guide, we learn how to use `@property` for type checking and constraining, using CSS to set default values, and define whether a custom property can inherit values or not.
+
+## Why register custom properties
```
chrisdavidmills:
```suggestion
## Why register custom properties?
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+  --myColor: #abba;
+}
+@media (width < 750px) {
+  main {
+    --myColor: #abba;
+  }
+}
+```
+
+The [CSS properties and values API](/en-US/docs/Web/CSS/Guides/Properties_and_values_api/) provides a method of registering custom properties using the the CSS `@property` at-rule. Registering CSS custom properties with `@property` sets the variable's type, inheritance, and default value, which makes them more predictable and more performant.
```
chrisdavidmills:
```suggestion
The [CSS properties and values API](/en-US/docs/Web/CSS/Guides/Properties_and_values_api/) provides a method of registering custom properties using the CSS `@property` at-rule. Registering CSS custom properties with `@property` sets the variable's type, inheritance, and default value, making them more predictable and performant.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+Registered properties are validated when they're computed rather than when they are parsed. That means both that invalid values won't appear as invalid when inspecting the element's properties, and that including an invalid property after a valid one won't fall back to the valid property. An invalid property will, however, fall back to its registered default.
```
chrisdavidmills:
```suggestion
Registered properties are validated when they're computed rather than when they are parsed. This means that:

 - Invalid values won't appear as invalid when inspecting the element's properties.
 - An invalid property included after a valid one won't fall back to the valid property. An invalid property will, however, fall back to its registered default.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+By default, all CSS variables inherit. When you change a non-registered variable value, the browser re-parses the DOM to check which elements in the DOM tree are effected. With registered properties, setting the value `inherits: false;` tells the browser that, if the value changes, the children don't need to be reparsed, narrowing the scope of style recalculation.
```
chrisdavidmills:
```suggestion
By default, all CSS variables inherit. When you change a non-registered variable value, the browser re-parses the DOM to check which elements in the DOM tree are affected. With registered properties, setting the value `inherits: false;` tells the browser that, if the value changes, the children don't need to be reparsed, narrowing the scope of style recalculation.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+Generally, you'll want to register your custom properties is you want to animate the variable, you want type checking, or you want predictable inheritance behavior. If you're creating a CSS framework, you'll want to namespace and register your custom properties to ensure that they always have a default value and should a framework user change the value, they can't assign it an invalid data type.
```
chrisdavidmills:
```suggestion
Generally, you'll want to register your custom properties if you are animating the variable, implementing type checking, or guaranteeing predictable inheritance behavior. If you're creating a CSS framework, namespace and register your custom properties to ensure that they always have a default value, and prevent framework users from assigning invalid data types to them if they change their values.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+sidebar: cssref
+---
+
+The `@property` at-rule, part of the [CSS Houdini](/en-US/docs/Web/API/Houdini_APIs) set of APIs, allows developers to explicitly define [CSS custom properties](/en-US/docs/Web/CSS/Reference/Properties/--*). In this guide, we learn how to use `@property` for type checking and constraining, using CSS to set default values, and define whether a custom property can inherit values or not.
+
+## Why register custom properties
+
+CSS custom properties, commonly called CSS variables, can be defined by simple declaration. Generally, these are declared on the {{cssxref(":root")}}, and can be overridden with a different value scoped to the content matched by the selector, even based state or by container or media queries:
+
+```css
```
chrisdavidmills, line 16: It feels like this example needs a small bit of intro text to explain it, just for consistency with other examples later on, although granted, it is pretty obvious. I'll leave it up to you.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+  animation: colorChange 2s linear forwards;
+}
+@keyframes colorChange {
+  to {
+    color: red;
+  }
+}
+```
+
+In this example, `--myColor` is registered as an inheritable color with the value `#bada55`. The `:root` assignment is ignored because the property's data type is registered as `<color>`, not `<angle>`. The registration improves the animation performance because the value is of the correct type, and the browser doesn't need to parse the value before animating it.
```
chrisdavidmills:
```suggestion
In this example, `--myColor` is registered as an inheritable color with the initial value `#bada55`. The `:root` assignment is ignored because the property's data type is registered as `<color>`, not `<angle>`. The registration improves the animation performance because the value is of the correct type, and the browser doesn't need to parse the value before animating it.
```

Two things:

1. I think this explanation would work better placed before the example code, where it would also act as a lead-in to the example.
2. Make it a live example, so the reader can play with it?
3.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+## The @property and descriptors
+
+The `@property` rule represents a custom property registration directly in a stylesheet without having to run any JavaScript. Valid `@property` rules result in a registered custom property, which is similar to calling {{domxref('CSS.registerProperty_static', 'registerProperty()')}} with equivalent parameters.
```
chrisdavidmills:
```suggestion
The `@property` rule enables custom property registration directly inside stylesheets without requiring any JavaScript. Valid `@property` rules produce registered custom properties, producing the same effect as calling {{domxref('CSS.registerProperty_static', 'registerProperty()')}} with equivalent parameters.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+@property --rotation {
+  syntax: "<angle>";
+  inherits: false;
+  initial-value: 45deg;
+}
+```
+
+### @property descriptors
+
+For the `@property` rule to be valid, it must include both the {{cssxref("@property/syntax","syntax")}} and {{cssxref("@property/inherits","inherits")}} descriptors.
```
chrisdavidmills:
```suggestion
The `@property` at-rule can include the following descriptors. For it to be valid, it must include both the {{cssxref("@property/syntax","syntax")}} and {{cssxref("@property/inherits","inherits")}} descriptors.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+- `syntax`
+  - : The {{cssxref("@property/syntax","syntax")}} descriptor is a string that describes the allowed value types for the registered custom property. The value of the descriptor may be a data type name, such as `<color>`, `<length>`, or `<number>`, with multipliers (`+`, `#`) and combinators (`|`), or a custom ident. In our example, the `--rotation` property is set to `<angle>`, meaning it can only be assigned an {{cssxref("angle")}} value.
```
chrisdavidmills, line 94: I'm guessing this can also be a [`type()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/type) function? Or no?

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+- `inherits`
+  - : The {{cssxref("@property/inherits","inherits")}} descriptor is a {{glossary("boolean")}} value that controls whether the custom property registration specified by `@property` inherits by default: set it to `true` to have it inherit, `false` otherwise. See [Registering and using custom properties](#registering-and-using-custom-properties) for an example of both values.
```
chrisdavidmills:
```suggestion
  - : The {{cssxref("@property/inherits","inherits")}} descriptor is a {{glossary("boolean")}} value that controls whether the custom property inherits by default: set it to `true` to have it inherit, `false` to not inherit. See [Registering and using custom properties](#registering_and_using_custom_properties) for an example of both values.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+If either the `syntax` or `inherits` descriptor is missing, the entire `@property` rule is invalid and ignored. If the `initial-value` descriptor is required but omitted, the entire `@property` rule is invalid and ignored. Unknown descriptors are invalid and ignored, but do not invalidate the `@property` rule.
+
+## Registering and using custom properties
+
+Custom properties can be registered in CSS or JavaScript. In this example, we will be define two custom properties, `--item-size` will be defined with CSS and `--item-color` will be defined with JavaScript. We'll then use these custom properties to define the size (width and height) and background color of the three following items.
```
chrisdavidmills:
```suggestion
Custom properties can be registered in CSS or JavaScript. In this example, we will define two custom properties: `--item-size` will be defined with CSS, and `--item-color` will be defined with JavaScript. We'll then use these custom properties to define the size (width and height) and background color of the three following items.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+<div class="container">
+  <div class="item one">Item one</div>
+  <div class="item two">Item two</div>
+  <div class="item three">Item three</div>
+</div>
+```
+
+### Registering properties with CSS
+
+Here we use the CSS `@property` at-rule to define a custom property named `--item-size`. The property sets the initial value to `40%`, limiting valid values to {{cssxref("percentage")}} values only. This means, when used as the value for an item's size, its size will always be relative to its parent's size. The property is inheritable.
```
chrisdavidmills:
```suggestion
Here we use the CSS `@property` at-rule to define the `--item-size` custom property. We set its initial value to `40%` and limit valid values to {{cssxref("percentage")}} values only. This means that when this property is used to size an element, the element's size will always be relative to its parent's size. The property is inheritable.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+@property --item-size {
+  syntax: "<percentage>";
+  inherits: true;
+  initial-value: 40%;
+}
+```
+
+### Registering a property with JavaScript
+
+We define a second custom property, `--item-color`, using [JavaScript](/en-US/docs/Web/JavaScript) instead of CSS. The JavaScript {{domxref('CSS.registerProperty_static', 'registerProperty()')}} method is equivalent to `@property` at-rule. The property is defined to have an initial value of `aqua`, to accept only {{cssxref("&lt;color&gt;")}} values, and is not inherited.
```
chrisdavidmills:
```suggestion
We define our second custom property, `--item-color`, using [JavaScript](/en-US/docs/Web/JavaScript) instead of CSS. The JavaScript {{domxref('CSS.registerProperty_static', 'registerProperty()')}} method is equivalent to the `@property` at-rule. The property is defined to have an initial value of `aqua`, to accept only {{cssxref("&lt;color&gt;")}} values, and not be inherited.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+  --item-size: initial;
+  --item-color: inherit;
+}
+.three {
+  --item-size: 1000px;
+  --item-color: xyz;
+}
+```
+
+We use the two custom properties to style the items, setting the size and background color for all the items at once:
```
chrisdavidmills:
```suggestion
We use the two custom properties as values for the items' {{cssxref("width")}}, {{cssxref("height")}}, and {{cssxref("background-color")}} properties:
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+{{ EmbedLiveSample('Registering and using custom properties', '100%', '250px') }}
+
+### Inheritance
+
+The `inherits` descriptor controls whether or not the registered CSS custom property inherits by default.
+
+In our example, the size property was set to be inheritable; the color was not. The two custom properties, `--item-size: 20%` and `--item-color: orange;` were set on the `container` parent, overriding the `40%` and `aqua` default values set when these custom properties were defined.
+
+For item one, none of these custom properties were set. Because the `--item-size` is inheritable, the value `20%` set on its parent `container` was inherited and used. The property `--item-color` was not inheritable, so the value `orange` set on the parent was not considered. Instead the default initial value `aqua` was used.
```
chrisdavidmills:
```suggestion
For item one, none of these custom properties were set. Because the `--item-size` is inheritable, the value `20%` set on its parent `container` was inherited and used. The property `--item-color` was not inheritable, so the value `orange` set on the parent was not considered. Instead, the default initial value `aqua` was used.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+  name: "--item-color",
+  syntax: "<color>",
+  inherits: false,
+  initialValue: "aqua",
+});
+```
+
+### Using registered custom properties
+
+We can assign new values to registered properties. Here we custom property values on the parent container. We then set custom property values on the element themselves. The values set on `two` are valid while the values set on `three` are invalid.
```
chrisdavidmills:
```suggestion
We can assign new values to registered properties. Here, we set custom property values on the parent container, then set custom property values on the items themselves. The values set on `two` are valid while the values set on `three` are invalid.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+In our example, the size property was set to be inheritable; the color was not. The two custom properties, `--item-size: 20%` and `--item-color: orange;` were set on the `container` parent, overriding the `40%` and `aqua` default values set when these custom properties were defined.
+
+For item one, none of these custom properties were set. Because the `--item-size` is inheritable, the value `20%` set on its parent `container` was inherited and used. The property `--item-color` was not inheritable, so the value `orange` set on the parent was not considered. Instead the default initial value `aqua` was used.
+
+The `inherits` descriptor is a required descriptor; if it were missing or invalid, the entire `@property` rule would have been invalid and ignored.
+
+### Valid and invalid values
+
+A benefit of registering your custom properties (versus using simple declarations) is that they can be set to only be re-assigned to valid values. The `syntax` descriptor defines which values are valid. Any declaration with a value that doesn't meet those contstraints will be ignored.
```
chrisdavidmills:
```suggestion
A benefit of registering your custom properties (versus using simple declarations) is that they can be set to be reassigned only to valid values. The `syntax` descriptor defines which values are valid. Any declaration with a value that doesn't meet those constraints will be ignored.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+For item two, CSS global keywords were set for both custom properties. The CSS global keyword values, {{cssxref("initial")}}, {{cssxref("inherit")}}, {{cssxref("unset")}}, {{cssxref("revert")}}, {{cssxref("revert-layer")}}, are valid values for all value types and therefore valid no matter the `syntax` descriptor value. The `--item-size` was set to `initial` and used the `initial-value: 40%;` set in the `@property` declaration. The `initial` value means the `initialValue` value for the property is used. The `--item-color` was set to `inherit`, explicitly inheriting the `orange` value from its parent even though the custom property was set to otherwise not be inherited. This is why item two was orange.
```
chrisdavidmills:
```suggestion
For item two, CSS global keywords were set for both custom properties. The CSS global keyword values — {{cssxref("initial")}}, {{cssxref("inherit")}}, {{cssxref("unset")}}, {{cssxref("revert")}}, and {{cssxref("revert-layer")}} — are valid values for all value types and therefore valid no matter the `syntax` descriptor value. The `--item-size` was set to `initial` and therefore used the `initial-value: 40%;` set in the `@property` declaration. The `--item-color` was set to `inherit`, explicitly inheriting the `orange` value from its parent even though the custom property was set to otherwise not be inherited. This is why item two was orange.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+For item three, the `--item-size` value was set to `1000px`. While `1000px` is a {{cssxref("length")}} value, the `@property` declaration requires the value be a `<percentage>`, so the declaration was not valid and was ignored, meaning the inheritable `20%` set on the parent was used. The `xyz` value was also invalid. As `registerProperty()` set `--item-color` to not be inherited, the default initial value of `aqua` was used and not the parent's `orange` value.
+
+### Animating custom property values
+
+Registered CSS custom properties offer a performance edge because they are treated as typed data rather than strings, providing the browser's rendering engine with a defined syntax and inheritance rules. The browser parses the value once. By declaring a syntax, when that variable is used elsewhere, the type and internal representation are known, so it doesn't need to be re-parsed.
```
chrisdavidmills:
```suggestion
Registered CSS custom properties offer a performance edge because they are treated as typed data rather than strings, providing the browser's rendering engine with a defined syntax and inheritance rules. The browser parses the defined syntax once; when that variable is used elsewhere, the type and internal representation are already known, so it doesn't need to be re-parsed.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+Because non-registered custom properties are just strings, the browser may not {{glossary("interpolation", "interpolate")}} them by default, so can no animation them. Non-registered custom property may make a normally animatable property animate discretely. Registered variables allow the browser to interpolate the variable directly which can offload interpolation calculations to the GPU, which is more performant than animations done on the main thread.
```
chrisdavidmills:
```suggestion
Because non-registered custom properties are just strings, the browser cannot {{glossary("interpolation", "interpolate")}} them, so it cannot smoothly animate them. Instead, non-registered custom property values are animated discretely. Registered variables can be interpolated and therefore animated; their interpolation calculations are offloaded to the GPU, which is more performant than animations done on the main thread.
```

files/en-us/web/css/reference/at-rules/@property/index.md
```diff
@@ -6,18 +6,30 @@ browser-compat: css.at-rules.property
...
+The **`@property`** [CSS](/en-US/docs/Web/CSS) [at-rule](/en-US/docs/Web/CSS/Guides/Syntax/At-rules) is used to explicitly define [CSS custom properties](/en-US/docs/Web/CSS/Reference/Properties/--*), allowing for property type checking and constraining, setting default values, and defining whether a custom property can inherit values or not.
 
-The `@property` rule represents a custom property registration directly in a stylesheet without having to run any JavaScript. Valid `@property` rules result in a registered custom property, which is similar to calling {{domxref('CSS.registerProperty_static', 'registerProperty()')}} with equivalent parameters.
+> [!NOTE]
+> The JavaScript {{domxref('CSS.registerProperty_static', 'registerProperty()')}} method is equivalent to `@property` at-rule.
```
chrisdavidmills:
```suggestion
> The JavaScript {{domxref('CSS.registerProperty_static', 'registerProperty()')}} method is equivalent to the `@property` at-rule.
```

files/en-us/web/css/reference/at-rules/@property/index.md
```diff
@@ -26,116 +38,87 @@ The custom property name is a {{cssxref("dashed-ident")}} that starts with `--`
...
 - {{cssxref("@property/inherits","inherits")}}
   - : A boolean value that controls whether the custom property registration specified by `@property` inherits by default.
 - {{cssxref("@property/initial-value","initial-value")}}
   - : A value that sets the starting value for the property.
 
 ## Description
 
+The `@property` at-rule is part of the [CSS Houdini](/en-US/docs/Web/API/Houdini_APIs) set of APIs. It allows developers to explicitly define [CSS custom properties](/en-US/docs/Web/CSS/Reference/Properties/--*), allowing for property type checking and constraining, setting default values, and defining whether a custom property can inherit values or not.
+
+The `@property` rule represents a custom property registration directly in a stylesheet without having to run any JavaScript. Valid `@property` rules result in a registered custom property, which is similar to calling {{domxref('CSS.registerProperty_static', 'registerProperty()')}} with equivalent parameters.
```
chrisdavidmills:
```suggestion
The `@property` rule enables custom property registration directly inside stylesheets without requiring any JavaScript. Valid `@property` rules produce registered custom properties, producing the same effect as calling {{domxref('CSS.registerProperty_static', 'registerProperty()')}} with equivalent parameters.
```

files/en-us/web/css/reference/at-rules/@property/index.md
```diff
@@ -26,116 +38,87 @@ The custom property name is a {{cssxref("dashed-ident")}} that starts with `--`
...
+
 The following conditions must be met for the `@property` rule to be valid:
 
 - The `@property` rule must include both the {{cssxref("@property/syntax","syntax")}} and {{cssxref("@property/inherits","inherits")}} descriptors.
   If either is missing, the entire `@property` rule is invalid and ignored.
-- The {{cssxref("@property/initial-value","initial-value")}} descriptor is optional if the value of the `syntax` descriptor is the universal syntax definition (that is, `syntax: "*"`).
+- The `syntax` may be a data type name (such as `<color>`, `<length>`, or `<number>`, etc.), with multipliers (`+`, `#`) and combinators (`|`), a custom ident, or the universal syntax definition (`*`), meaning the syntax can be any valid token stream. The value is a {{cssxref("string")}}. As such, it must be in quotes.
```
chrisdavidmills, line 56: Again, can it accept a `type()` function?

files/en-us/web/css/reference/at-rules/@property/index.md
```diff
@@ -26,116 +38,87 @@ The custom property name is a {{cssxref("dashed-ident")}} that starts with `--`
...
 - If the value of the `syntax` descriptor is not the universal syntax definition, the {{cssxref("@property/initial-value","initial-value")}} descriptor has to be a [computationally independent](https://drafts.css-houdini.org/css-properties-values-api-1/#computationally-independent) value.
   This means the value can be converted into a computed value without depending on other values, except for "global" definitions independent of CSS.
   For example, `10px` is computationally independent—it doesn't change when converted to a computed value. `2in` is also valid, because `1in` is always equivalent to `96px`. However, `3em` is not valid, because the value of an `em` is dependent on the parent's {{cssxref("font-size")}}.
 - Unknown descriptors are invalid and ignored, but do not invalidate the `@property` rule.
 
+If multiple valid `@property` rules are defined using the same name, the last one in stylesheet order "wins". If custom property registration with the same name is registered using the CSS `@property` and the JavaScript `CSS.registerProperty()`, the JavaScript registration wins.
```
chrisdavidmills:
```suggestion
If multiple valid `@property` rules are defined using the same name, the last one in stylesheet order "wins". If custom properties are registered with the same name using `@property` in CSS and `CSS.registerProperty()` in JavaScript, the JavaScript registration wins.
```

files/en-us/web/css/reference/at-rules/@property/index.md
```diff
@@ -26,116 +38,87 @@ The custom property name is a {{cssxref("dashed-ident")}} that starts with `--`
...
 ## Formal syntax
 
 {{csssyntax}}
 
 ## Examples
 
-### Using `@property` to register and use a custom property
+### Basic example
+
+In this example, we use the `@property` to declare two custom properties, and then use those properties in our style declarations.
```
chrisdavidmills:
```suggestion
In this example, we use the `@property` at-rule to declare two custom properties, and then use those properties in our style declarations.
```

files/en-us/web/api/css_properties_and_values_api/guide/index.md
estelle: in this one case, because it's a JS page, about Registering custom properties with JS, and the article I'm linking to is about the exact same thing but in CSS, i think it's important to be clear even if it doesn't match the title, but the other occurrences, yup.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
estelle: I will be editing the API one from a JS perspective. We need to have this in both places.

files/en-us/web/css/guides/properties_and_values_api/index.md
estelle: went with:
> Covers how to register CSS custom properties with the `@property` at-rule and explains the benefits of doing so.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
estelle: went with:

> CSS custom properties, commonly called CSS variables, can be defined by simple declaration. Generally, these are declared on the {{cssxref(":root")}}, and can be overridden with a different value in subsequent rulesets, including in ruled based on UI state and within container and media queries:

 `:hover` and `@media` examples follow

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
estelle: added "for example"

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
estelle: d'oh

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
estelle: i changed the two examples adding "circles" with an angle.
there's not much to play with, so not live exampling it.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
estelle: i included this before the previous example.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
estelle: no. at least, not yet.

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
estelle: pulled out the js part and moved it down

files/en-us/web/css/reference/at-rules/@property/index.md
estelle: nope. at least not yet. the spec delimits the values too, and it's not listed.

files/en-us/web/api/css_properties_and_values_api/guide/index.md
chrisdavidmills: ok

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
chrisdavidmills: righty-ho

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,211 @@
...
+slug: Web/CSS/Guides/Properties_and_values_API/Registering_properties
+page-type: guide
+sidebar: cssref
+---
+
+The `@property` at-rule, part of the [CSS Houdini](/en-US/docs/Web/API/Houdini_APIs) set of APIs, allows developers to explicitly define [CSS custom properties](/en-US/docs/Web/CSS/Reference/Properties/--*). In this guide, we learn how to use `@property` for type checking and constraining, set default custom property values in CSS, and define whether a custom property can inherit values or not.
+
+## Why register custom properties?
+
+CSS custom properties, commonly called CSS variables, can be defined by simple declaration. Generally, these are declared on the {{cssxref(":root")}}, and can be overridden with a different value in subsequent rulesets, including in ruled based on UI state and within container and media queries.
```
chrisdavidmills:
```suggestion
CSS custom properties, commonly called CSS variables, can be defined by basic declaration. Generally, these are declared on the {{cssxref(":root")}}, and can be overridden with a different value in subsequent rulesets — including in rules based on UI state and within container and media queries.
```

files/en-us/web/css/guides/properties_and_values_api/registering_properties/index.md
```diff
@@ -0,0 +1,211 @@
...
+
+```css
+@property --item-size {
+  syntax: "<percentage>";
+  inherits: true;
+  initial-value: 40%;
+}
+```
+
+You can also use [JavaScript](/en-US/docs/Web/JavaScript) to register properties. JavaScript's {{domxref('CSS.registerProperty_static', 'CSS.registerProperty()')}} method is equivalent to `@property` at-rule. Here we defined our second custom property, `--item-color`, to have an initial value of `aqua`, to accept only {{cssxref("&lt;color&gt;")}} values, and to not be inherited.
```
chrisdavidmills:
```suggestion
You can also use [JavaScript](/en-US/docs/Web/JavaScript) to register properties. JavaScript's {{domxref('CSS.registerProperty_static', 'CSS.registerProperty()')}} method is equivalent to the `@property` at-rule. Here we defined our second custom property, `--item-color`, to have an initial value of `aqua`, to accept only {{cssxref("&lt;color&gt;")}} values, and to not be inherited.
```

PR #42884, opened 2026-01-23
Remove MV3 permissions note from Content Scripts article

rebloor:
### Description

The behavior of `host_permissions` for MV3 extensions was changed in Firefox 127, see [Bug 1889402](https://bugzilla.mozilla.org/show_bug.cgi?id=1889402) Grant origin permissions during install for mv3. However, when other changes were made for this, the note on the content scripts article wasn't updated.

Given that the MV3 behavior is now consistent with MV2, the note is dropped.

However, some additional information has been added to the `optional_host_permissions` page to cover users' ability to grant and revoke optional permissions.

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/content_scripts/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/content_scripts/index.md
@@ -49,7 +49,7 @@ Registered content scripts are only executed if the extension is granted [host p
 
 To inject scripts programmatically, the extension needs either the [`activeTab` permission](/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#activetab_permission) or [host permissions](/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#host_permissions). The `scripting` permission is required to use methods from the {{WebExtAPIRef("scripting")}} API.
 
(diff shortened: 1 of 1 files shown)


files/en-us/mozilla/add-ons/webextensions/content_scripts/index.md
```diff
@@ -49,8 +49,6 @@ Registered content scripts are only executed if the extension is granted [host p
 
 To inject scripts programmatically, the extension needs either the [`activeTab` permission](/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#activetab_permission) or [host permissions](/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#host_permissions). The `scripting` permission is required to use methods from the {{WebExtAPIRef("scripting")}} API.
 
-Starting with Manifest V3, host permissions are not automatically granted at install time. Users may opt in or out of host permissions after installing the extension.
```
Rob--W, line 52:
```suggestion
Host permissions may be requested at install time for hosts in the `matches` lists of the `content_scripts` manifest key. Users may opt in or out of host permissions after installing the extension.
```

Included a note to continue emphasizing that the permission can be revoked later. Also added a note about `matches` because that is currently not documented as a way to trigger permission requests from the articles linked in this section.

PR #42890, opened 2026-01-23
Unexpected output error fixed in an example

gabriel-1607:
### Description
I modified one line and added another line.
The incrementDay function was **intended to return a Date instance**, as stated in the inline JS comment; however, it just returns a Unix time number. My small fix is returning the Date instance instead of returning the return value of the setDate method of Date.

The current inline JS comment shows `2019-06-20` as the return value of the function, despite the real output, on testing, would be Unix and, thus, something like `1561006800`.

### Motivation
This small intended improvement to the JS example is an opportunity to showcase my progress in the understanding and mastery of web technologies such as JavaScript, which will serve my professional career or even my personal satisfaction, because I will be able to display this as my accomplisment of contributing to MDN docs, which are an authoritative and prestigious source of information.

### Additional details
[Here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_classes#why_classes) is the link to the section of the page where the example is found.

There are no related issues or pull requests to this single line fix, as far as I know.

### DIFF
--- a/files/en-us/web/javascript/guide/using_classes/index.md
+++ b/files/en-us/web/javascript/guide/using_classes/index.md
@@ -725,8 +725,7 @@ For example, one thing that makes `Date` objects infamous is that they're _mutab
 
 ```js
 function incrementDay(date) {
-  date.setDate(date.getDate() + 1);
-  return date;
+  return new Date(date.setDate(date.getDate() + 1));
 }
 const date = new Date(); // 2019-06-19
 const newDay = incrementDay(date);


gabriel-1607:
Josh-Cena, the point of this example, what it is intended to show, is that the original instance of Date is modified.

Your fix to the code makes the example out of place. Another example could be made to show the correct way avoid the mutability of class instances.

EDIT: sorry, your fix still changes the original instance. Forget about this comment.

gabriel-1607:
Thank you too, Josh!

Josh-Cena:
Yes, the intention is to show that it "looks like" a new instance is returned but in fact the original is still modified, so your confusion is on point :)

PR #42900, opened 2026-01-24
Synchronize with BCD v7.2.5

OnkarRuikar:
[BCD release v7.2.5](https://github.com/mdn/browser-compat-data/releases/tag/v7.2.5)

/cc cc/ @bsmth @wbamberg

### DIFF
--- a/files/en-us/web/api/shadowroot/index.md
+++ b/files/en-us/web/api/shadowroot/index.md
@@ -38,7 +38,7 @@ You can retrieve a reference to an element's shadow root using its {{domxref("El
 - {{DOMxRef("ShadowRoot.pointerLockElement")}} {{ReadOnlyInline}}
   - : Returns the {{DOMxRef('Element')}} set as the target for mouse events while the pointer is locked.
     `null` if lock is pending, pointer is unlocked, or if the target is in another tree.
-- `ShadowRoot.referenceTarget` {{ReadOnlyInline}} {{Experimental_Inline}}
+- `ShadowRoot.referenceTarget` {{ReadOnlyInline}} {{Experimental_Inline}} {{non-standard_inline}}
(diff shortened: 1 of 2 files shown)


caugner:
@OnkarRuikar Is this script fully automated, or does it require manual intervention? I'm asking to understand if it could be added to this repo, then we could open these PRs with a daily scheduled workflow.

OnkarRuikar:
> @OnkarRuikar Thanks - are these ever "wrong"? I mean what level of checking do we need to do on them?

@hamishwillee For quick reviews, look for only markdown syntax errors. If you have more time, then you can compare the changes with BCD[ release diff](https://github.com/mdn/browser-compat-data/compare/v7.2.4...v7.2.5#diff-16f116d4feb7d4f359a035126d9929a0c028f708638d41ece75012d32064325bR7344). The sync automation has been running for more than a year, synchronizing every Tuesday and Saturday. So I would say it is stable. Reviewers raise concerns mostly for the following reasons:
- PR got merged in BCD, but it doesn't reflect in this sync PRs. The reason could be that the PR didn't make it in time to be in the corresponding BCD release. The solution is to simply wait till the next BCD release. 
- Statues are not reflected on CSS pages. The reason is that we don't update inline status macros on CSS pages anymore.
- Sometimes the statuses are not reflected correctly on API landing pages. The reason is that the script picks only the first provided BCD query. We haven't figured out how to handle landing pages with multiple BCD queries. The discussion never concluded.


> @OnkarRuikar Is this script fully automated, or does it require manual intervention? I'm asking to understand if it could be added to this repo, then we could open these PRs with a daily scheduled workflow.

@caugner Yes, it's fully automated, and no manual intervention is required. It already [runs at every GItHub's midnight](https://github.com/OnkarRuikar/temp/actions). Unless you wish to trigger it immediately after BCD releases, no change is required.
This sync automation is a stopgap solution till the platform handles the status badges/banners at build/render time.  When that happens, we won't need the status macros and the sync automation. How far are we from implementing this in the platform code?

wbamberg:
Personally when I review these I just look for things that seem wildly wrong (i.e. the bot went completely off the rails). If it's syntactically correct, I don't check the data. The bot is pretty reliable :).

PR #42901, opened 2026-01-25
Fix grammar

RizzWan-Codes:
### Description

This PR fixes several minor grammar and clarity issues across multiple Web API documentation pages.

### Motivation

These changes improve the readability and professionalism of the MDN Web Docs without altering any technical meaning.

Fixes #42899

PR #42902, opened 2026-01-25
Fix capitalization in hosting and domain name section

gmch86:
“If” should be lowercase “if” (since it’s mid-sentence).

### Description
Corrected a minor capitalization error in the text.

### Motivation
To improve readability.


Josh-Cena:
Thanks for the PR! This appears to have already been fixed in https://github.com/mdn/content/pull/42901.

PR #42907, opened 2026-01-26
Update margin-block-start values reference from margin-left to margin-top

### LINKED ISSUE #42906: Syntax main section - Values subsection: Definition could reference most used physical mapped property instead of margin-left to improve context reading.
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/margin-block-start#values

Syntax main section - Values subsection content reference.

The present text references that the property takes the same values as the physical property margin-left, which is true. But because the margin-block-start maps to margin-top in most websites that are written in horizontal left to right writing I think it makes sense to reference that property instead to improve context reading.

The definition could reference the most used mapped physical property for margin-block-start which is margin-top in horizontal left to right writing.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/margin-block-start#values

Same is happening for the rest of margin-related properties.

### MDN metadata

<details>

* Folder: `en-us/web/css/reference/properties/margin-block-start`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/margin-block-start
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/reference/properties/margin-block-start/index.md

</details>

Buenaventura0:
…-top.

The logical margin-block-start maps to the physical margin-top in horizontal left to right writing, so it makes sense to reference that property in values section instead of margin-left, to avoid any misleading confusion for most developers.

### Description

The present text references that the property takes the same values as the physical property margin-left, which is true. But because the margin-block-start maps to margin-top in most websites that are written in horizontal left to right writing I think it makes sense to reference that property instead to improve context reading.

### Motivation

I am suggesting a context reading improvement based on my teaching background. This reference improves cohesion to the article and avoids possible confusion regarding which property is being mapped to most of the time.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/margin-block-start#values

"Fixes #42906 "

### DIFF
--- a/files/en-us/web/css/reference/properties/margin-inline-start/index.md
+++ b/files/en-us/web/css/reference/properties/margin-inline-start/index.md
@@ -85,7 +85,7 @@ It relates to {{cssxref("margin-block-start")}}, {{cssxref("margin-block-end")}}
 
 ### Values
 
-The `margin-inline-start` property takes the same values as the {{cssxref("margin-left")}} property.
+The `margin-inline-start` property takes the same values as the {{cssxref("margin-top")}} property.
 
 ## Formal definition
 


Josh-Cena:
If you feel so strongly, then we should also update `margin-block-end`, `margin-inline-start`, and `margin-inline-end`. FYI, the spec refers to `margin-top` for all these four values (which are specified together anyway), so my recommendation is to change all of these to `margin-top`, if we want to make any changes at all. I personally think that sticking to one reference property, whether `margin-left` or `margin-top`, is better than using different reference properties because that creates the false illusion that `margin-block-start` has anything to do with `margin-top`.

Buenaventura0:
> If you feel so strongly, then we should also update `margin-block-end`, `margin-inline-start`, and `margin-inline-end`. FYI, the spec refers to `margin-top` for all these four values (which are specified together anyway), so my recommendation is to change all of these to `margin-top`, if we want to make any changes at all. I personally think that sticking to one reference property, whether `margin-left` or `margin-top`, is better than using different reference properties because that creates the false illusion that `margin-block-start` has anything to do with `margin-top`.

Hi Josh! Thanks for your answer and for taking the time to review the pull request and the issue. That's absolutely right, in the issue I suggested to review the rest of the logical properties as well, as they are all related. My take on this is that there are two possible ways to improve the present section:

1. Reference in the syntax values section the physical mapped property that's been replaced most of the time in horizontal rl writing, which is what most devs encounter daily. Example, for margin-block-start -> margin-top and so on.
2. Reference the four values, declaring explicitly that the property takes the same values as the four physical properties.

Looking forward to hearing from Chris and the community.

Happy Sunday.

Josh-Cena:
As I said, (1) is the dangerous option here, because what devs are familiar with is irrelevant to what users actually use, especially since i18n or even browser autotranslation are out of the developer's control. The dev cannot assume that their website is read in a particular direction and must program defensively.

Option (2) is theoretically fine, but IMO unworthwhile. I would much prefer we either do nothing or change everything to be `margin-top` to align with the spec.

Buenaventura0:
Hi @chrisdavidmills and @Josh-Cena. I understand, thank you for your honest and kind feedback. This is part of a learning process involving myself with the MDN and content contribution.  Since I use it so much in my daily routine, I want to give back to the community.

I incorporated the proposed change to the remaining logical properties to align with the spec, if you are okay with it and think this PR should move forward. Or if it makes sense to close the PR and the issue, that's fine too.

Let me know what you think, and thanks again for your time.

Best!

PR #42908, opened 2026-01-26
Add session mgt guide

wbamberg:
Here's a page on session management.

## Terminology

I've talked mostly about two different models, one where the state is stored in the server and the client gets a session ID, and the other where the state is stored as a signed object (JWT) in the client. I've called these "centralized" and "decentralized" but people don't use these terms.

People sometimes use terms like "cookie-based" for the first and the second "JWT-based", but I don't like these, because ISTM that:

1. how you store session information in the client and communicate it to the server is orthogonal to these architectural choices - that is, you don't have to do the first using cookies, and you could do the second using cookies

2. (this is a much weaker objection) although in practice everyone does use JWTs for the second, that's an implementation choice not an architectural one.

I have asked about this and got feedback that my choice here is reasonable, but just flagging it here. We could call them "server-maintained state" and "client-maintained state" which is more descriptive but a real mouthful.

## Frameworks and libraries

Regarding the section on "Frameworks and libraries", I do think we need to say this but went back and forth on whether to say it at the start or at the end, and how much detail to go into. FWIW although it might seem to negate the point of all this if we just tell people to use a framework, I really don't think it does - it is important to understand the principles and good practices, even if your framework is looking after a lot of the details for you.

### DIFF
--- a/files/en-us/web/security/authentication/session_management/index.md
+++ b/files/en-us/web/security/authentication/session_management/index.md
@@ -148,7 +148,7 @@ In some situations, a website might want to invalidate a user's sessions and req
 
 ## Decentralized session management
 
-The model we've described so far might be called "centralized session management": the session state is maintained in a database on the server, and the client is given an identifier for the state, which it gives to the server when it makes a request.
+In the "centralized session management" model that we've described so far, the session state is maintained in a database on the server, and the client is given an identifier for the state, which it gives to the server when it makes a request.
 
 An alternative model might be called "decentralized session management". In this model, the session state is maintained as a {{glossary("digital signature", "digitally signed")}} object in the client. These signed objects are typically represented as [JSON Web Tokens (JWTs)](https://www.jwt.io/).
 


files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+An attacker can hijack a session if it can execute its code in the user's browser, in the context of the target website, for example in a [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attack.
+
+If the session ID is available to JavaScript (for example, because it is stored in [local storage](/en-US/docs/Web/API/Web_Storage_API)), then the attacker can steal the ID and use it to impersonate the user.
+
+The defense against this is to implement the usual [defenses against XSS](/en-US/docs/Web/Security/Attacks/XSS#defenses_against_xss).
+
+### Session fixation
+
+Ina session fixation attack, the attacker chooses a session ID value, and then convinces the website to use this value as the session ID for the target user. The attacker can then impersonate the user, because they know the ID. In this attack, the attacker doesn't have to steal the ID, because they knew it all along.
```
hamishwillee:
```suggestion
In a session fixation attack, the attacker chooses a session ID value, and then convinces the website to use this value as the session ID for the target user. The attacker can then impersonate the user, because they know the ID. In this attack, the attacker doesn't have to steal the ID, because they knew it all along.
```

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+In this guide we'll first describe the most common architecture for session management, in which the user's session state is kept in the server:
+
+- When the user is authenticated, the server records their state and generates a _session ID_, which it associates with this state. The server returns a copy of the session ID to the client. The client stores the session ID.
+
+- When the client makes a request to the server, the client includes the ID. The server uses this to look up the user's session state, to decide what the client is allowed to do.
+
+![Diagram showing session management using server-maintained session state](session-mgt-centralized.svg)
+
+## Attacks
```
hamishwillee: I'd add structure, with this and subsequent under centralized session management.  The "In this guide" would then explain that "we'll describe the two main architectures for session management, common attacks, etc etc.

wbamberg: You mean something like:

```
## Centralized session management

### Session ID values

### Session ID storage

### ...

## Decentralized session management
```

I did consider this, but:

1. I wasn't that committed to my terminology to want to make the whole doc structure based on it.
2. I want to present centralized as "the main/default way" and decentralized as "this alternative, which most people should not use".
3. Because MDN's ToC only shows H2, we would have a useless ToC with that structure

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
+---
+title: Session management
+slug: Web/Security/Authentication/Session_management
+page-type: guide
+sidebar: security
+---
+
+Once a website has authenticated a user, it typically needs to remember that user's identity across the subsequent requests that the user makes to the site. That's the function of a session management system.
```
hamishwillee: This just feels a little awkward. I don't love my words, but have a think about it.

```suggestion
HTTP is a stateless protocol, which means that each request and response pair beween the browser and the website does not depend on previous interactions.
Session management systems provide a mechanism for remembering a user's identity across multiple requests to a site.
```

files/en-us/web/security/authentication/session_management/index.md
hamishwillee: I hate that our infrastructure forces this - if not for that I'd still hope for the structure.

Anyway, you have reasons - it doesn't "not work".

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+Session management is not exclusive to authentication: for example, a site might let any user choose a light or dark theme, and then their choice would represent state that should persist across subsequent requests to the site. However, it's especially relevant to authentication, because an authenticated user's identity on the site represents state that is especially valuable to an attacker, so a site author has to be especially careful when implementing a session management solution for authenticated users.
+
+In this guide we'll first describe the most common architecture for session management, in which the user's session state is kept in the server:
+
+- When the user is authenticated, the server records their state and generates a _session ID_, which it associates with this state. The server returns a copy of the session ID to the client. The client stores the session ID.
+
+- When the client makes a request to the server, the client includes the ID. The server uses this to look up the user's session state, to decide what the client is allowed to do.
```
hamishwillee, line 24: Below you say

> In targeting session management, an attacker's goal is to be able to impersonate a legitimate user, without having to compromise the website's authentication system itself. In this section we'll describe the two main ways in which an attacker can do this: [_session hijacking_](#session_hijacking) and [_session fixation_](#session_fixation). 

As I understand it now, and what I think you are trying to show here, is that authentication is (can be?) separate to session. The authentication follows some process that confirms identity. The representation of identity is then a session cookie (in this first case). The authentication process might use some credential to identify the user, but the session _becomes_ the credential that is used for subsequently identifying the user

So in that sentence you're saying that once the session is what represents the user credential, you can attack that rather than the authentication mechanism.

It might be worth spelling out in a sentence after this bullet point something along the lines of "Therefore after authentication the session id is the credential used to identify the user, and that defines their accesses".

FWIW thanks!  One of my big problems with sessions and authentication is that I never really understood how they interacted. We talk about cookies and authentication and credentials, and we separately talk about all sorts of authentication mechanisms that don't use cookies and it wasn't clear how they all fit together (at least for me)

wbamberg: > Below you say
> 
> > In targeting session management, an attacker's goal is to be able to impersonate a legitimate user, without having to compromise the website's authentication system itself. In this section we'll describe the two main ways in which an attacker can do this: [_session hijacking_](#session_hijacking) and [_session fixation_](#session_fixation).
> 
> As I understand it now, and what I think you are trying to show here, is that authentication is (can be?) separate to session. The authentication follows some process that confirms identity. The representation of identity is then a session cookie (in this first case). The authentication process might use some credential to identify the user, but the session _becomes_ the credential that is used for subsequently identifying the user
> 
> So in that sentence you're saying that once the session is what represents the user credential, you can attack that rather than the authentication mechanism.

Yes, this is how I understand things.

> 
> It might be worth spelling out in a sentence after this bullet point something along the lines of "Therefore after authentication the session id is the credential used to identify the user, and that defines their accesses".
> 

I will think about this.

> FWIW thanks! One of my big problems with sessions and authentication is that I never really understood how they interacted. We talk about cookies and authentication and credentials, and we separately talk about all sorts of authentication mechanisms that don't use cookies and it wasn't clear how they all fit together (at least for me)

Thanks! It's nice to hear that this is helpful (and I hope it is also correct!)

files/en-us/web/security/authentication/session_management/index.md
wbamberg: hm, I wanted to start by tying session management into authentication, since that's the context here ("you're reading about authentication, and part of that is knowing about session mgt". I don't particularly mind starting with the generic "HTTP is stateless" approach but it would need some additional rework in the intro, or things get incoherent.

In particular, the "Session management systems provide a mechanism for remembering a user's identity..." doesn't follow well from a general statement about statelessness. Why are suddenly talking about identity here?

I think it could work to say something like:

1. HTTP is stateless
2. session management (in general) is a way to add state (in general)
3. a major application for this is remembering that a particular client has been authenticated as a particular user, which is where it's relevant to authentication.

hamishwillee: Yes, I saw the intended approach. I do prefer your "could work something like" structure, but you could ignore me as this does do what you want.

wbamberg: does this -> https://github.com/mdn/content/pull/42908/changes/63570e4e3ec82f3b7903a92fd3dae3896c320dd2 make more sense?

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+title: Session management
+slug: Web/Security/Authentication/Session_management
+page-type: guide
+sidebar: security
+---
+
+HTTP is commonly called a _stateless_ protocol, meaning that it provides no built-in mechanism to correlate a series of individual HTTP requests. This is a problem for websites that want to maintain some client-specific state. For example, a site might let any user choose a light or dark theme, and then their choice would represent state that should persist across subsequent requests to the site. To support this, the website needs to treat a series of requests from a single client as representing a _session_, which can be associated with this persistent state.
+
+When a website has the ability to authenticate users, so as to grant individuals access to specific data or actions, it typically wants to treat an authenticated user's identity as state that persists across multiple requests from the client. Unlike a choice of theme, though, an authenticated user's identity on the site represents state that is especially valuable to an attacker, so a site author has to be especially careful when implementing a session management solution for authenticated users.
```
hamishwillee: Too especially :-)
```suggestion
When a website has the ability to authenticate users, so as to grant individuals access to specific data or actions, it typically wants to treat an authenticated user's identity as state that persists across multiple requests from the client. Unlike a choice of theme, though, an authenticated user's identity on the site represents state that is especially valuable to an attacker, so a site author has to be very careful when implementing a session management solution for authenticated users.
```

files/en-us/web/security/authentication/session_management/index.md
hamishwillee: It always made sense, but yes, I like the update a lot.

hamishwillee (approved):

files/en-us/web/security/authentication/session_management/index.md
wbamberg: 👍 thanks Hamish.

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+The attacker could steal a session ID while it is being sent between client and server, in a [manipulator in the middle (MITM)](/en-US/docs/Web/Security/Attacks/MITM) attack.
+
+The defense against this is to serve the site over [TLS](/en-US/docs/Web/Security/Defenses/Transport_Layer_Security).
+
+#### Predicting the session ID
+
+If an attacker can predict, guess, or brute-force the value of the session ID that's assigned to a legitimate user, then they can use that ID to impersonate them, without having to steal the user's copy.
+
+The defense against this is to [ensure that session ID values are long enough and random enough to be unpredictable](#session_id_recommendations).
```
Elchi3:
```suggestion
The defense against this is to [ensure that session ID values are long enough and random enough to be unpredictable](#session_id_values).
```

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+If the session ID is in a cookie, then it will be automatically sent when the client makes requests. Again, there are attributes that control which requests include the cookie.
+
+### Secure transmission
+
+If the [`Secure`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure) attribute is set, then the cookie will not be sent over unencrypted connections. This is a defense against a [Manipulator in the Middle (MITM)](/en-US/docs/Web/Security/Attacks/MITM) attack stealing a session identifier while it is in transit.
+
+### Controlling cookie destinations
+
+The [`Domain`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#domain) and [`Path`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#path) attributes control the URLs to which the cookie will be sent, and should be set to the most restrictive values possible, given the needs of your site.
```
Elchi3:
```suggestion
The [`Domain`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#domaindomain-value) and [`Path`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#pathpath-value) attributes control the URLs to which the cookie will be sent, and should be set to the most restrictive values possible, given the needs of your site.
```

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+If the website uses cookies, instead of URL parameters, to send the session ID, then this attack is harder: the attacker needs to be able to execute an XSS attack on the target website. This is one reason cookies are a much better method for storing and communicating session IDs than URL parameters.
+
+The key defense against session fixation, though, is that the server must always generate a new session ID, and invalidate any existing value, every time the user signs in.
+
+## Session ID values
+
+The session ID should:
+
+- Contain enough entropy to protect against [guessing or prediction attacks](predicting_the_session_id). The [OWASP guide to session management](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#session-id-properties) recommends at least 64 bits of entropy.
```
Elchi3:
```suggestion
- Contain enough entropy to protect against [guessing or prediction attacks](#predicting_the_session_id). The [OWASP guide to session management](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#session-id-properties) recommends at least 64 bits of entropy.
```

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+### CSRF considerations
+
+If a website uses cookies to transmit session identifiers, it must implement protection against [cross-site request forgery (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF) attacks. The risk here is that the browser may include cookies in a request from the attacker's site to the target website, and if these cookies include a valid session ID, then the server will treat the request as if it came from a legitimate user, and carry out the attacker's request.
+
+Our CSRF guide describes [recommended defenses](/en-US/docs/Web/Security/Attacks/CSRF#defenses_against_csrf) here: note that setting the [`SameSite`](#samesite) cookie attribute is only part of a complete defense.
```
Elchi3:
```suggestion
Our CSRF guide describes [recommended defenses](/en-US/docs/Web/Security/Attacks/CSRF#defenses_against_csrf) here: note that setting the [`SameSite`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) cookie attribute is only part of a complete defense.
```

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+The refresh endpoint gives the application a centralized place to determine whether or not the user's session should be invalidated. If it does choose to invalidate the session, is does so by refusing to issue new access tokens.
+
+## Frameworks and libraries
+
+Rather than implement all the details of session management yourself, we recommend that you use a well-regarded framework or library, and use the facilities it provides for session management.
+
+## Session management checklist
+
+We can summarise the recommendations above as follows:
```
Elchi3:
```suggestion
We can summarize the recommendations above as follows:
```

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,231 @@
...
+title: Session management
+slug: Web/Security/Authentication/Session_management
+page-type: guide
+sidebar: security
+---
+
+HTTP is commonly called a _stateless_ protocol, meaning that it provides no built-in mechanism to correlate a series of individual HTTP requests. This is a problem for websites that want to maintain some client-specific state. For example, a site might let any user choose a light or dark theme, and then their choice would represent state that should persist across subsequent requests to the site. To support this, the website needs to treat a series of requests from a single client as representing a _session_, which can be associated with this persistent state.
+
+When a website has the ability to authenticate users, so as to grant individuals access to specific data or actions, it typically wants to treat an authenticated user's identity as state that persists across multiple requests from the client. Unlike a choice of theme, though, an authenticated user's identity on the site represents state that is especially valuable to an attacker, so a site author has to be very careful when implementing a session management solution for authenticated users.
+
```
Elchi3, line 11: Should we have a heading here?

```suggestion

## Centralized session management

```

wbamberg: The reason I didn't do this is that the subsequent H2s (Session ID values etc) are specifically about "centralized". If MDN's ToC worked properly I would put them as H3s under this H2, but they don't. 

This is essentially the same as Hamish's https://github.com/mdn/content/pull/42908#discussion_r2772084243.

But since it has come up twice I will think again.

wbamberg: All right, I have been back and forth a bit on this today but have settled on something small, that just adds a bit about the two models at the start and outlines what we're going to say about them. I think anything more extensive would get pretty complicated.

-> https://github.com/mdn/content/pull/42908/commits/ef5635f14475274e5aaee50d6fe052f7d3b242b7

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,237 @@
...
+In some situations, a website might want to invalidate a user's sessions and require reauthentication:
+
+1. When the client attempts some high-risk operation, such as attempting to change, or actually changing, the user's credentials on the site, or triggering the account recovery process (for example, password reset).
+
+2. When the server has some grounds for thinking that the session ID might have been stolen. This could include, for example, a sign-in from a new IP address or device.
+
+## Decentralized session management
+
+The model we've described so far might be called "centralized session management": the session state is maintained in a database on the server, and the client is given an identifier for the state, which it gives to the server when it makes a request.
```
hamishwillee: No longer just "might be called"

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,237 @@
...
+
+In this guide we'll first describe the most common model for session management, in which the user's session state is stored in the server. We call this [centralized session management](#centralized_session_management). We'll outline this model and then describe potential attacks and implementation considerations.
+
+In an alternative model, which is useful in certain web application architectures, the client stores the session state as an object that's been {{glossary("digital signature", "digitally signed")}} by the server. We call this [decentralized session management](#decentralized_session_management). We'll outline this model, compare it to the more common approach, and discuss when to use which.
```
hamishwillee: I like the new heading structure - it's a good compromise. I find these paragraphs a little clunky. I think perhaps because you're being so careful to say "we're calling this". I don't think that is necessary unless you also explain what others call it, and the structure of the two paragraphs seems repetetive. 

Just a suggestion though.

```suggestion
The most common model for session management is [centralized session management](#centralized_session_management), in which the user's session state is stored in the server.
An alternative model, which is useful in certain web application architectures, is [decentralized session management](#decentralized_session_management) (sometimes referred to as JWT Server tokens). In this model the client stores the session state as an object that's been {{glossary("digital signature", "digitally signed")}} by the server.

In the following sections we first explain the centralized model, then describe potential attacks and implementation considerations.
We'll then describe the decentralized model, compare the two methods, and discuss when to use each approach.
```

files/en-us/web/security/authentication/session_management/index.md
```diff
@@ -0,0 +1,239 @@
...
+
+In some situations, a website might want to invalidate a user's sessions and require reauthentication:
+
+1. When the client attempts some high-risk operation, such as attempting to change, or actually changing, the user's credentials on the site, or triggering the account recovery process (for example, password reset).
+
+2. When the server has some grounds for thinking that the session ID might have been stolen. This could include, for example, a sign-in from a new IP address or device.
+
+## Decentralized session management
+
+The model we've described so far might be called "centralized session management": the session state is maintained in a database on the server, and the client is given an identifier for the state, which it gives to the server when it makes a request.
```
wbamberg:
```suggestion
In the "centralized session management" model that we've described so far, the session state is maintained in a database on the server, and the client is given an identifier for the state, which it gives to the server when it makes a request.
```

hamishwillee:
What he said ^^^

PR #42929, opened 2026-01-27
MouseEvent.movementX/Y - update events that set a value

hamishwillee:
`MouseEvent.movementX` and `MouseEvent.movementY` state that the value is only set for `mousemove` event.  In fact it is also set for `pointermove` on Chrome and FF, and `pointerrawupdate` on Chrome and FF from version 148.

This adds the clarification. 

Related docs work can be tracked in #42752


hamishwillee:
> but those two properties are actually part of `MouseEvent` and only inherited, so it is ok as is.

Yes. MDN policy is to avoid duplication - we don't document the inherited methods in the derived classes. It makes sense in that context, but sometimes it does mean that information is hidden.

Thank you - and to @wbamberg for review.

PR #42943, opened 2026-01-28
Synchronize code with description

Rosuav:
### Description

Simple change to one of the examples for the Meter element

### Motivation

The descriptive text says that the min attribute is omitted, but it was included. The min attribute can in fact be omitted, and having an example showing this is beneficial.

### Additional details

The alternative fix would be to remove the paragraph above the example stating that the min attribute is omitted. Either would be acceptable, but keeping the note and the distinction showcases another aspect of the meter's functionality.


chrisdavidmills (approved):
LGTM; thanks, @Rosuav.

I agree that removing the attribute is the best option.

Rosuav:
Thank you! Fast turnaround, appreciate it.

PR #42944, opened 2026-01-28
Annotate example of using with null with good/bad

dontcallmedom:
### Description

Add example-bad and example-good annotations to JS examples of `using` with a null value.

### Motivation

To clarify which example is good to follow


chrisdavidmills (approved):
Thanks @dontcallmedom, that change makes sense to me.

However, before merging, I had a look at the surrounding structure, fixed a grammar issue, and reordered the section to put the `acquireResource()` function out into its own code block.

Does that make sense to you — it is neither part of the good or bad code, right?

dontcallmedom:
that makes a lot of sense, thanks!

PR #42947, opened 2026-01-28
Fix: clarify cross-site terminology and SameSite=Lax behavior

### LINKED ISSUE #42945: SameSite docs use cross-origin terminology and are imprecise about Lax behavior
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Cookies

SameSite

The documentation uses the term *cross-origin* when describing SameSite behavior. However, SameSite is defined in terms of *site* (scheme + eTLD+1), not origin. Using cross-origin terminology is misleading and can cause confusion.

Additionally the description of `SameSite=Lax` is technically correct but imprecise. It does not clearly distinguish between:

- cross-site top-level navigations vs cross-site subresource requests  
- safe HTTP methods (such as GET) vs unsafe/state-changing methods (such as POST)

- Consistent use of *cross-site* terminology instead of cross-origin
- A more precise explanation of `SameSite=Lax` behavior that explicitly mentions top-level navigations and safe HTTP methods

https://web.dev/articles/same-site-same-origin?hl=en
https://web.dev/articles/samesite-cookies-explained?hl=en

_No response_

andrealungh1:
### Description

Clarifies the `SameSite` cookie attribute documentation by replacing incorrect *cross-origin* terminology with *cross-site* and by making the description of `SameSite=Lax` more precise.

### Motivation

The SameSite attribute is defined in terms of *site* (scheme + eTLD+1), not origin. Using cross-origin terminology can mislead readers and make it harder to understand real browser behavior.

Additionally, while the existing description of `SameSite=Lax` was technically correct, it was imprecise. This update clarifies when cookies are included or omitted by explicitly distinguishing top-level navigations from subresource requests and safe HTTP methods from state-changing ones.

### Additional details

- https://web.dev/articles/same-site-same-origin
- https://web.dev/articles/samesite-cookies-explained

Fixes #42945

### DIFF
--- a/files/en-us/web/security/practical_implementation_guides/cookies/index.md
+++ b/files/en-us/web/security/practical_implementation_guides/cookies/index.md
@@ -34,9 +34,9 @@ To minimize the scope for cookie vulnerabilities on your site, limit access to c
 - `Path`
   - : Cookies should be set to the most restrictive `Path` possible.
 - `SameSite`
-  - : Forbid sending cookies via cross-origin requests (for example from {{htmlelement("img")}} elements) using `SameSite`. You should use one of the following two values:
+  - : Forbid sending cookies via cross-site requests (for example from {{htmlelement("img")}} elements) using `SameSite`. You should use one of the following two values:
     - `SameSite=Strict`: Only send the cookie in same-site contexts (navigations and other requests). Cookies are omitted in cross-site requests (e.g., embedding images or other resources from other sites) and cross-site navigation (e.g., when following a link from a different web page). This is a very strict setting, but it does provide strong [CSRF](/en-US/docs/Web/Security/Attacks/CSRF) protection, so use this value if possible.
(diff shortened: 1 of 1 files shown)


wbamberg (commented):
Thanks for your PR!

I think rather than repeat the definition for SameSite here it would be better to link to our existing docs.

I would replace this whole section with something like:

> Developers should set the [`SameSite`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute to `Strict` or `Lax`. This is a partial defense against [CSRF](/en-US/docs/Web/Security/Attacks/CSRF#defense_in_depth_samesite_cookies) and [clickjacking](/en-US/docs/Web/Security/Attacks/Clickjacking#clickjacking_defenses) attacks.

Also note that the "Examples" section repeats the "cross-origin" error.

wbamberg:
Closing in favour of https://github.com/mdn/content/pull/43089.

PR #40674, opened 2025-08-08
URL Pattern API updates

hamishwillee:
FF142 adds support for the [URL Pattern API](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern) in https://bugzilla.mozilla.org/show_bug.cgi?id=1731418

The docs are a little out of date according to the spec (i.e. in https://github.com/whatwg/urlpattern/pull/198/files). This attempts to fix them.

Fixes #30024
Fixes #9327
Fixes #27864
Fixes #17763

Related work can be tracked in #40474

PR #42950, opened 2026-01-28
URL Pattern API incorrect example

Link2Twenty:
### Description

The leading slash should not exist when looking at a hash. The current example will return false.

### Motivation

I was reading the examples and thought it didn't make sense, tested it and it was wrong

### Additional details

N/A

N/A


wbamberg:
I think you are right. It looks like this got introduced in https://github.com/mdn/content/pull/40674, @hamishwillee is there anything extra to consider here?

PR #42955, opened 2026-01-29
Reference for stage 3 joint-iteration

Josh-Cena:
Part of https://github.com/mdn/content/issues/42743

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
@@ -38,7 +38,7 @@ A new {{jsxref("Iterator")}} object. Each of its elements is an array with lengt
 
 ## Description
 
-The `Iterator.zip()` function behaves like a transpose operation. If we represent iterables as arrays, the input may looks like this:
+The `Iterator.zip()` function behaves like a [transpose](https://en.wikipedia.org/wiki/Transpose) operation, yielding arrays containing the elements at matching positions in each of the inputs. If we represent iterables as arrays, the input may looks like this:
 
 ```js
 [
@@ -48,7 +48,7 @@ The `Iterator.zip()` function behaves like a transpose operation. If we represen
 ];
 ```
 
-The resulting iterator will yield the following arrays:
+The resulting iterator, regardless of the options, will start by yielding the following arrays:
 
 ```js
 [a1, b1, c1];
@@ -56,7 +56,7 @@ The resulting iterator will yield the following arrays:
 [a3, b3, c3];
 ```
 
-After the first three arrays are yielded, the input iterable `b` is exhausted on the fourth `next()` call—it returns `{ done: true }`. What happens next depends on the `mode` option. If `mode` is `"shortest"` (the default), the resulting iterator stops here: the other two input iterators are closed. If `mode` is `"strict"`, an error is thrown because the other two iterables are _not_ finished when the second one yields the result `{ done: true }`. If `mode` is `"longest"`, the resulting iterator continues yielding arrays, filling missing values. For example, if `padding` is not provided, it defaults to `undefined`:
+After the first three arrays are yielded, the input iterable `b` is exhausted on the fourth `next()` call—it returns `{ done: true }`. What happens next depends on the `mode` option. If `mode` is `"shortest"` (the default), the resulting iterator stops here: the other two input iterators are [closed](/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#errors_during_iteration). If `mode` is `"strict"`, an error is thrown because the other two iterables are _not_ finished when the second one yields the result `{ done: true }`. If `mode` is `"longest"`, the resulting iterator continues yielding arrays, filling missing values. For example, if `padding` is not provided, it defaults to `undefined`:
 
 ```js
 [a4, undefined, c4];
@@ -76,7 +76,7 @@ If the `padding` iterable has fewer than three values, the remaining missing val
 
 ### Iteration over a map with indices
 
-Using `Iterator.zip()`, you can iterate over any iterable while also having access to an incrementing counter:
+Using `Iterator.zip()`, you can iterate over any iterable object (strings aren't supported by default) while also having access to an incrementing counter:
 
 ```js
 const ages = new Map([
@@ -101,7 +101,7 @@ for (const [index, [name, age]] of Iterator.zip([numbers(), ages])) {
 // 2: Evelyn is 35 years old.
 ```
 
-`numbers()` is an infinite iterator that generates incrementing numbers starting from `0`. Because `Iterator.zip()` stops when the shortest input iterable is exhausted, the loop iterates exactly three times. The `numbers()` iterator is properly closed after the loop ends; it doesn't cause an infinite loop.
+`numbers()` is an infinite iterator that generates incrementing numbers starting from `0`. Because `Iterator.zip()` stops by default when the shortest input iterable is exhausted, the loop iterates exactly three times. The `numbers()` iterator is properly closed after the loop ends; it doesn't cause an infinite loop.
 
 ### Creating a Map from lists of keys and values
 
--- a/files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
@@ -179,7 +179,7 @@ In some cases, you may wish to split by [grapheme clusters](/en-US/docs/Web/Java
 
 ```js
 const segmenter = new Intl.Segmenter("en-US", { granularity: "grapheme" });
-const str1 = "😁😭😡";
+const str1 = "🤷‍♂️🤷‍♀️🤷";
 const str2 = "123";
 const it = Iterator.zip([
   segmenter.segment(str1).map(({ segment }) => segment),
@@ -189,9 +189,9 @@ for (const [char1, char2] of it) {
   console.log(`${char1} - ${char2}`);
 }
 // Output:
-// 😁 - 1
-// 😭 - 2
-// 😡 - 3
+// 🤷‍♂️ - 1
+// 🤷‍♀️ - 2
+// 🤷 - 3
 ```
 
 ## Specifications
@@ -204,6 +204,8 @@ for (const [char1, char2] of it) {
 
 ## See also
 
+- [Polyfill of `Iterator.zip` in `core-js`](https://core-js.io/docs/features/proposals/joint-iteration)
+- [es-shims polyfill of `Iterator.zip`](https://www.npmjs.com/package/es-iterator-helpers)
 - {{jsxref("Iterator")}}
 - {{jsxref("Iterator.zipKeyed()")}}
 - {{jsxref("Iterator.from()")}}
--- a/files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
@@ -1,5 +1,6 @@
 ---
 title: Iterator.zip()
+short-title: zip()
 slug: Web/JavaScript/Reference/Global_Objects/Iterator/zip
 page-type: javascript-static-method
 browser-compat: javascript.builtins.Iterator.zip
--- a/files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
@@ -31,7 +31,7 @@ Iterator.zip(iterables, options)
         - `"longest"`: The resulting iterator stops when all input iterables are exhausted. Missing values from shorter iterables are filled according to the `padding` option.
         - `"strict"`: A {{jsxref("TypeError")}} is thrown if not all input iterables finish at the same time.
     - `padding` {{optional_inline}}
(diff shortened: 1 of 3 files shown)


files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
```diff
@@ -0,0 +1,113 @@
...
+
+- `iterables`
+  - : An object. Each property's key is used as the key in the resulting objects. The property's value must implement either the [iterable](/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) protocol or, failing that, the [iterator](/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol) protocol. These iterables may be infinite. Strings are rejected: to zip strings, convert them to iterators explicitly using {{jsxref("Iterator.from()")}}.
+- `options` {{optional_inline}}
+  - : An object specifying behavior in case of inconsistent input lengths. It may have the following properties:
+    - `mode` {{optional_inline}}
+      - : One of the following:
+        - `"shortest"` (default): The resulting iterator stops when one input iterable is exhausted.
+        - `"longest"`: The resulting iterator stops when all input iterables are exhausted. Missing values from shorter iterables are filled according to the `padding` option.
+        - `"strict"`: A {{jsxref("TypeError")}} is thrown if not all input iterables finish at the same time.
```
hamishwillee, line 32: Shouldn't this also be in an Exceptions section?

files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
```diff
@@ -0,0 +1,113 @@
...
+### Parameters
+
+- `iterables`
+  - : An object. Each property's key is used as the key in the resulting objects. The property's value must implement either the [iterable](/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol) protocol or, failing that, the [iterator](/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol) protocol. These iterables may be infinite. Strings are rejected: to zip strings, convert them to iterators explicitly using {{jsxref("Iterator.from()")}}.
+- `options` {{optional_inline}}
+  - : An object specifying behavior in case of inconsistent input lengths. It may have the following properties:
+    - `mode` {{optional_inline}}
+      - : One of the following:
+        - `"shortest"` (default): The resulting iterator stops when one input iterable is exhausted.
+        - `"longest"`: The resulting iterator stops when all input iterables are exhausted. Missing values from shorter iterables are filled according to the `padding` option.
```
hamishwillee, line 31:
```suggestion
        - `"longest"`: The resulting iterator stops when all input iterables are exhausted. Missing values from shorter iterables may be filled according to the `padding` option.
```

files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
```diff
@@ -0,0 +1,113 @@
...
+        - `"strict"`: A {{jsxref("TypeError")}} is thrown if not all input iterables finish at the same time.
+    - `padding` {{optional_inline}}
+      - : Either `undefined` (default) or an object. Only retrieved and validated when `mode` is `"longest"`. If `undefined`, missing values from shorter iterables are filled with `undefined`. If an object is provided, each key of the `iterables` argument is retrieved _as soon as `Iterator.zipKeyed()` is called_. `padding[key]` is used for missing values for `iterables[key]`. If the `padding` object is missing some keys, those keys are filled with `undefined`.
```
hamishwillee: It may be that this and https://github.com/mdn/content/pull/42955/files#r2752150527  are strictly correct but they feel odd. Normally when you look at optional options you would say "set this value if you want to control the padding, and otherwise it will default", whereas because of the first sentence here it reads like "set undefined or a value, and by the way you can also choose not o set undefined because it is the default".

In other words, it might not be wrong, but I'd be tempted to word it like

```suggestion
    - `padding` {{optional_inline}}
      - : An object. Only retrieved and validated when `mode` is `"longest"`. If `undefined`, missing values from shorter iterables are filled with `undefined`. If an object is provided, each key of the `iterables` argument is retrieved _as soon as `Iterator.zipKeyed()` is called_. `padding[key]` is used for missing values for `iterables[key]`. If the `padding` object is missing some keys, those keys are filled with `undefined`.
```

files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
hamishwillee: Comments apply to other doc too.

files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
```diff
@@ -0,0 +1,209 @@
...
+    - `padding` {{optional_inline}}
+      - : Either `undefined` (default) or an iterable object (not iterator). Only retrieved and validated when `mode` is `"longest"`. If `undefined`, missing values from shorter iterables are filled with `undefined`. If an iterable is provided, it is iterated for the number of times equal to the number of elements in `iterables` _as soon as `Iterator.zip()` is called_. `padding[i]` is used for missing values for `iterables[i]` (assuming `padding` and `iterables` are provided as arrays; they don't have to be). If `padding` is shorter than `iterables`, `undefined` is used for the remaining iterables.
+
+### Return value
+
+A new {{jsxref("Iterator")}} object. Each of its elements is an array with length equal to the number of input iterables, containing the elements from each input iterable at the corresponding position. If the `iterables` object is empty, the resulting iterator is created as completed.
+
+## Description
+
+The `Iterator.zip()` function behaves like a transpose operation. If we represent iterables as arrays, the input may looks like this:
```
hamishwillee: FWIW My maths is years old so I had to look up transpose. You might want to provide a link. Up to you. Also a typo "looks"

hamishwillee:
```suggestion
The `Iterator.zip()` function behaves like a transpose operation, yielding arrays containing the elements at matching positions in each of the inputs.
If we represent iterables as arrays, the input may look like this:
```

files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
```diff
@@ -0,0 +1,209 @@
...
+
+```js
+[
+  [a1, a2, a3, a4], // Iterable a
+  [b1, b2, b3], // Iterable b
+  [c1, c2, c3, c4, c5], // Iterable c
+];
+```
+
+The resulting iterator will yield the following arrays:
```
hamishwillee: Even if `strict` is set?

files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
```diff
@@ -0,0 +1,209 @@
...
+
+```js
+[a1, b1, c1];
+[a2, b2, c2];
+[a3, b3, c3];
+```
+
+After the first three arrays are yielded, the input iterable `b` is exhausted on the fourth `next()` call—it returns `{ done: true }`. What happens next depends on the `mode` option. If `mode` is `"shortest"` (the default), the resulting iterator stops here: the other two input iterators are closed. If `mode` is `"strict"`, an error is thrown because the other two iterables are _not_ finished when the second one yields the result `{ done: true }`. If `mode` is `"longest"`, the resulting iterator continues yielding arrays, filling missing values. For example, if `padding` is not provided, it defaults to `undefined`:
```
hamishwillee: What does it mean " the other two input iterators are closed." ? 

EDITED. I had some false starts here, but wasn't sure why ^^^ is relevant.

files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
```diff
@@ -0,0 +1,209 @@
...
+[p1, p2, c5];
+```
+
+If the `padding` iterable has fewer than three values, the remaining missing values are filled with `undefined`.
+
+## Examples
+
+### Iteration over a map with indices
+
+Using `Iterator.zip()`, you can iterate over any iterable while also having access to an incrementing counter:
```
hamishwillee: Pedantic, do we need to say "explicit iterable" - i.e. do we need to capture here that this doesn't work for strings?

files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
```diff
@@ -0,0 +1,209 @@
...
+  console.log(`${index}: ${name} is ${age} years old.`);
+}
+
+// Output:
+// 0: Caroline is 30 years old.
+// 1: Danielle is 25 years old.
+// 2: Evelyn is 35 years old.
+```
+
+`numbers()` is an infinite iterator that generates incrementing numbers starting from `0`. Because `Iterator.zip()` stops when the shortest input iterable is exhausted, the loop iterates exactly three times. The `numbers()` iterator is properly closed after the loop ends; it doesn't cause an infinite loop.
```
hamishwillee:
```suggestion
`numbers()` is an infinite iterator that generates incrementing numbers starting from `0`. Because `Iterator.zip()` stops when the shortest input iterable is exhausted (by default), the loop iterates exactly three times. The `numbers()` iterator is properly closed after the loop ends; it doesn't cause an infinite loop.
```

files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
```diff
@@ -0,0 +1,209 @@
+---
+title: Iterator.zip()
+slug: Web/JavaScript/Reference/Global_Objects/Iterator/zip
+page-type: javascript-static-method
+browser-compat: javascript.builtins.Iterator.zip
```
hamishwillee, line 6: Missing spec link. Are you planning on doing this BCD or deferring to me?

files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
hamishwillee: PS Good examples.

files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
```diff
@@ -0,0 +1,113 @@
...
+};
+
+for (const { name, age, city } of Iterator.zipKeyed(table)) {
+  console.log(`${name}, aged ${age}, lives in ${city}.`);
+}
+
+// Output:
+// Caroline, aged 30, lives in New York.
+// Danielle, aged 25, lives in London.
+// Evelyn, aged 35, lives in Hong Kong.
```
hamishwillee, line 96: Nit. I was expecting to see an object - from "This example shows how to convert from the former to the latter using `Iterator.zipKeyed()`."

files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
Josh-Cena: I'm ambivalent because this error is not thrown at _construction_ time but at _iteration_ time, so it technically doesn't belong to the function itself but to the iterator it produces :)

files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
Josh-Cena: You can also explicitly set it to `undefined`, though. Not all options have `undefined` and "missing" mean the same thing, so it's better to be explicit.

files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
Josh-Cena: Yeah I'll defer BCD to you 😄

files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
Josh-Cena: I prefer "are" because it's always the case; the "missing" state is also an explicit choice!

files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
hamishwillee: Thanks. Your decision makes sense to me.

files/en-us/web/javascript/reference/global_objects/iterator/zipkeyed/index.md
hamishwillee: I stand by the fact it is weird to read. Your call though.

hamishwillee (approved):
Thanks very much.

files/en-us/web/javascript/reference/global_objects/iterator/zip/index.md
```diff
@@ -0,0 +1,210 @@
...
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
+- {{jsxref("Iterator")}}
+- {{jsxref("Iterator.zipKeyed()")}}
+- {{jsxref("Iterator.from()")}}
+- {{jsxref("Iterator.concat()")}}
```
zloirock, line 212: Please, add links to `core-js` polyfills of those methods, https://core-js.io/docs/features/proposals/joint-iteration

Josh-Cena: Nice website. Perhaps you want to submit a PR to update all other pages to the website as well?

zloirock: Thanks. Yep, I'm thinking about some changes and making a PR with updating links after that.

hamishwillee:
@Josh-Cena Note, approved this so you can merge when you are ready.

Josh-Cena:
Thanks, I gave some thought to the suggestion and I think it makes sense, so reworded a bit.

PR #42959, opened 2026-01-29
Media Capabilities API: Fix Incorrect Word Choice

acruzg:
### Description

Surgical replacement of incorrect word choice.

### Motivation

Original words were "_a pick_" (choice, selection), but from the surrounding text (CPU/GPU usage), probably the intention was to use "_peak_" (relating to the highest level, price, rate, etc. that something reaches).

### DIFF
--- a/files/en-us/web/api/media_capabilities_api/index.md
+++ b/files/en-us/web/api/media_capabilities_api/index.md
@@ -19,7 +19,7 @@ The Media Capabilities API provide more powerful features than say {{DOMxref("Me
 
 To test support, smoothness, and power efficiency for encoding and decoding video or audio content, you use the {{DOMxRef("MediaCapabilities")}} interface's {{DOMxRef("MediaCapabilities.encodingInfo()","encodingInfo()")}} and {{DOMxRef("MediaCapabilities.decodingInfo()","decodingInfo()")}} methods.
 
-Media capabilities information enables websites to enable adaptive streaming to alter the quality of content based on actual user-perceived quality, and react to a pick of CPU/GPU usage in real time.
+Media capabilities information enables websites to enable adaptive streaming to alter the quality of content based on actual user-perceived quality, and react to peaks of CPU/GPU usage in real time.
 
 ## Interfaces
 


wbamberg:
Thanks for your PR!

The current wording is definitely strange.

Are you familiar with this API? The bit you've edited:

> Media capabilities information enables websites to enable adaptive streaming to alter the quality of content based on actual user-perceived quality, and react to peaks of CPU/GPU usage in real time.

...is this intended to mean something like: a website can start playing video, and monitor CPU/GPU usage, and if it gets above some threshold, can switch to a different less intensive format? If so I don't really see anywhere in this API that this complete use case is supported or anywhere where it's explained how to do it in combination with other APIs, so this is left looking like an unsupported claim.

Similarly the [guide page](https://developer.mozilla.org/en-US/docs/Web/API/Media_Capabilities_API/Using_the_Media_Capabilities_API) has something which looks relevant to the bit you have edited:

> Support for getting real-time feedback about the playback of media, so your code can make informed decisions about adapting the stream's quality or other settings to manage the user's perceived media performance and quality. One feature of this is the ability to detect when the device switches GPUs, so you can make appropriate adjustments based on the new GPU's capabilities.

...the bit about "switches GPUs" might support the reading of "pick" in the original? But I don't see anything like that feature in the API, nor is it explained how we would do this in the guide, which is very basic.

That para is followed by this note:

> Note: The display capabilities functionality mentioned in the third point above have not yet appeared in any browser. They will be a useful feature of the API once available, but there is a high probability of the display capabilities functionality changing a great deal before browser implementations arrive.

...which is double weird, because the para doesn't appear to be about "display capabilities" (but the second para does).

So... it seems to me that there is more work to do, if we are to make this story coherent. Do you have experience with this API that could help with that?

tidoust:
Not sure where it comes from, but I agree that the wording in both this page and the guide page seems wrong. The Media Capabilities API explainer notes that [real time experience management](https://github.com/w3c/media-capabilities/blob/main/explainer.md#real-time-experience-management) is a non-goal, pointing at the [Media Playback Quality](https://w3c.github.io/media-playback-quality/) as a possible way to detect perceived playback quality.

wbamberg:
Closing in favour of https://github.com/mdn/content/pull/43046.

PR #42967, opened 2026-01-29
ShadowRoot.referenceTarget is not read-only

alice:
### Description

Remove the ReadOnly marker, and also note that it's nullable.

### Additional details

This is the PR against the DOM standard which specifies the `referenceTarget` IDL attribute: https://whatpr.org/dom/1353.html#interface-shadowroot


sideshowbarker (approved):

sideshowbarker:
It seems the ReadOnly marker already got removed on main — but the `nullable` part didn’t. So I rebased and reduced the patch to just that.

alice:
Thank you @sideshowbarker!

PR #42974, opened 2026-01-29
docs: clarify that Array.map does not mutate the original array

SihamNafie24:
### Description

This pull request clarifies that Array.prototype.map() does not mutate the original array and instead returns a new array.

### Motivation

This clarification helps avoid confusion for readers who may assume that map() modifies the existing array, especially beginners learning JavaScript array methods.

### Additional details

No functional changes were made. This is a documentation-only improvement.


Josh-Cena:
Hi there, thanks for the PR and sorry for the wait!

However in fact this PR moves in the _opposite_ order of what I have done. In the past, our array method pages were very bloated and filled with callouts like this one. I cleared them all out and condensed them into a holistic overview on the `Array` landing page, specifically: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#copying_methods_and_mutating_methods. What's left on each of these pages is a single paragraph:

> The `map()` method is an [iterative method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#iterative_methods). ... Read the [iterative methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#iterative_methods) section for more information about how these methods work in general.

I think this is sufficient signposting to the correct place for information. There are also examples further down this page that illustrate the point.

PR #42975, opened 2026-01-29
docs: clarify <dialog> Esc close request behavior

### LINKED ISSUE #41656: Seemingly incorrect or outdated information regarding dialog closure via Escape key
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog

Handling the return value from the dialog

"If the dialog is closed by pressing the Esc key, the returnValue is not updated, and the close event doesn't occur"

If this were true, in the live example that follows, closing the dialog with Escape should not change the Output content.  However, at least for Windows on Chrome, the Output is set to "ReturnValue: ." indicating that the close event is getting triggered and the dialog returnValue is getting set to empty string.

I don't know if this might be specific to Chrome or Windows or something.  Maybe its behavior is against spec.  I certainly think that Escape triggering the close event and setting the returnValue is much preferable to it doing neither, at any rate.

But even if it were somehow due to the spec not being appropriately adhered to, it would still be very important to clarify in this documentation that the statement "If the dialog is closed by pressing the Esc key, the returnValue is not updated, and the close event doesn't occur" does not necessarily hold true, and should not be relied on for production code.

And if it turns out the statement is just entirely outdated or incorrect, then it can simply be removed or appropriately edited.

Only my own firsthand observation.

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/html/reference/elements/dialog`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/reference/elements/dialog/index.md
* Last commit: https://github.com/mdn/content/commit/f2d281d86396bcd2dcecfdabd5837b1590132aa6
* Document last modified: 2025-08-06T09:01:21.000Z

</details>

ross-u:
### Description

Corrects the explanation of how a modal `<dialog>` is dismissed when the user presses the Esc key. It clarifies the sequence of events, `cancel` followed by `close`, if not prevented, and explains that no explicit return value is provided in this case leaving `dialog.returnValue` as an empty string.

### Motivation

I would like to help and contribute to the MDN documentation and thought to give it a shot with this small issue first. 🙂 

Fixes #41656

### DIFF
--- a/files/en-us/web/html/reference/elements/dialog/index.md
+++ b/files/en-us/web/html/reference/elements/dialog/index.md
@@ -152,7 +152,7 @@ This example demonstrates the [`returnValue`](/en-US/docs/Web/API/HTMLDialogElem
 
 This example opens a modal dialog when the "Show the dialog" button is activated. The dialog contains a form with a {{HTMLElement("select")}} and two {{HTMLElement("button")}} elements, which default to `type="submit"`. An event listener updates the value of the "Confirm" button when the select option changes. If the "Confirm" button is activated to close the dialog, the current value of the button is the return value. If the dialog is closed by pressing the "Cancel" button, the `returnValue` is `cancel`.
 
-When the dialog is closed, the return value is displayed under the "Show the dialog" button. If the dialog is closed by pressing the <kbd>Esc</kbd> key, the `returnValue` is not updated, and the `close` event doesn't occur, so the text in the {{HTMLElement("output")}} is not updated.
+When the dialog is closed, the return value is displayed under the "Show the dialog" button. If the dialog is closed by pressing the <kbd>Esc</kbd> key, the dialog first triggers the [`cancel`](/en-US/docs/Web/API/HTMLDialogElement/cancel_event) event and then the [`close`](/en-US/docs/Web/API/HTMLDialogElement/close_event) event, leaving the `returnValue` as an empty string, so the {{HTMLElement("output")}} displays an empty return value.
 
 #### HTML
 
--- a/files/en-us/web/html/reference/elements/dialog/index.md
+++ b/files/en-us/web/html/reference/elements/dialog/index.md
@@ -152,7 +152,7 @@ This example demonstrates the [`returnValue`](/en-US/docs/Web/API/HTMLDialogElem
 
 This example opens a modal dialog when the "Show the dialog" button is activated. The dialog contains a form with a {{HTMLElement("select")}} and two {{HTMLElement("button")}} elements, which default to `type="submit"`. An event listener updates the value of the "Confirm" button when the select option changes. If the "Confirm" button is activated to close the dialog, the current value of the button is the return value. If the dialog is closed by pressing the "Cancel" button, the `returnValue` is `cancel`.
 
-When the dialog is closed, the return value is displayed under the "Show the dialog" button. If the dialog is closed by pressing the <kbd>Esc</kbd> key, the dialog first triggers the [`cancel`](/en-US/docs/Web/API/HTMLDialogElement/cancel_event) event and then the [`close`](/en-US/docs/Web/API/HTMLDialogElement/close_event) event, leaving the `returnValue` as an empty string, so the {{HTMLElement("output")}} displays an empty return value.
+When the dialog is closed, the `returnValue` is displayed under the "Show the dialog" button. If the dialog is closed by pressing the <kbd>Esc</kbd> key, the dialog first triggers the [`cancel`](/en-US/docs/Web/API/HTMLDialogElement/cancel_event) event and then the [`close`](/en-US/docs/Web/API/HTMLDialogElement/close_event) event, leaving the `returnValue` as an empty string.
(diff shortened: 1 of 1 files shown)


files/en-us/web/html/reference/elements/dialog/index.md
```diff
@@ -152,7 +152,7 @@ This example demonstrates the [`returnValue`](/en-US/docs/Web/API/HTMLDialogElem
...
 
-When the dialog is closed, the return value is displayed under the "Show the dialog" button. If the dialog is closed by pressing the <kbd>Esc</kbd> key, the `returnValue` is not updated, and the `close` event doesn't occur, so the text in the {{HTMLElement("output")}} is not updated.
+When the dialog is closed, the return value is displayed under the "Show the dialog" button. If the dialog is closed by pressing the <kbd>Esc</kbd> key, the dialog first triggers the [`cancel`](/en-US/docs/Web/API/HTMLDialogElement/cancel_event) event and then the [`close`](/en-US/docs/Web/API/HTMLDialogElement/close_event) event, leaving the `returnValue` as an empty string, so the {{HTMLElement("output")}} displays an empty return value.
```
chrisdavidmills:
```suggestion
When the dialog is closed, the `returnValue` is displayed under the "Show the dialog" button. If the dialog is closed by pressing the <kbd>Esc</kbd> key, the dialog first triggers the [`cancel`](/en-US/docs/Web/API/HTMLDialogElement/cancel_event) event and then the [`close`](/en-US/docs/Web/API/HTMLDialogElement/close_event) event, leaving the `returnValue` as an empty string.
```

chrisdavidmills (changes requested):
Hi @ross-u, and thank you for your work in fixing this issue.

I've provided a suggestion to tighten up the text a bit, however, I think there is a bigger concern here — the behavior works as described, but only in Chrome.

If you try it in Firefox and Safari, the `returnValue` is set to an empty string when you press Esc, only if you haven't already selected a value from the dialog dropdown or pressed the dialog "Cancel" button. If you've done one of those two things, then on subsequent presses of the "Show the dialog" button, when you press the Esc key, the `returnValue` remains set to the previously selected dialog value.

I don't know if this is old behavior, and Chrome's behavior has been updated to spec behavior, or something else. But it is worth investigating this and thinking about how to communicate it in the doc.

Josh-Cena:
Asked Codex to investigate this.

**Firefox’s behavior is correct for the `returnValue` difference reported in the PR.** Escape should preserve the previous value; Chrome’s reset to `""` is a bug.

The specification separates two things:

- **Events:** Escape initiates a close request. The dialog fires `cancel`; if the request isn’t canceled, it closes and queues a `close` event. The [[dialog event test](https://github.com/web-platform-tests/wpt/blob/master/html/semantics/interactive-elements/the-dialog-element/dialog-cancel-events.html)](https://github.com/web-platform-tests/wpt/blob/master/html/semantics/interactive-elements/the-dialog-element/dialog-cancel-events.html) explicitly checks that order.
- **Return value:** `returnValue` starts as `""` **when the element is created**, not each time it opens. The close watcher supplies an internal “request close return value,” initially `null`. The closing algorithm updates `returnValue` only when its result is non-null. Thus, ordinary Escape dismissal preserves the existing value. See the [[returnValue definition](https://html.spec.whatwg.org/multipage/interactive-elements.html#dom-dialog-returnvalue)](https://html.spec.whatwg.org/multipage/interactive-elements.html#dom-dialog-returnvalue) and [[closing algorithm, step 9](https://html.spec.whatwg.org/multipage/interactive-elements.html#close-the-dialog)](https://html.spec.whatwg.org/multipage/interactive-elements.html#close-the-dialog).

For example, after `dialog.close("cancel")`, reopening the same dialog and pressing Escape should leave `dialog.returnValue === "cancel"`.

**Chromium’s implementation explains the discrepancy.** Its constructor initializes `request_close_return_value_` to `""`. `CloseWatcherFiredClose()` passes that to `close()`, which overwrites `return_value_` because the string isn’t null. That directly explains the reported reset. See [[Chromium’s dialog implementation](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/core/html/html_dialog_element.cc)](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/core/html/html_dialog_element.cc). This is source inspection, not a fresh browser reproduction.

The [[PR](https://github.com/mdn/content/pull/42975)](https://github.com/mdn/content/pull/42975) therefore fixes the incorrect assertion that `close` doesn’t fire, but introduces an incorrect assertion about an empty return value. I’d suggest:

> Pressing Esc requests that the dialog close, firing a `cancel` event. If the request is not canceled, the dialog closes and fires a `close` event. The `returnValue` remains unchanged, retaining its previous value or its initial empty string.

Any note about Chrome resetting the value should describe it as a browser discrepancy, rather than the specified behavior.

PR #42977, opened 2026-01-30
Fix error in comment

Aurora2500:
### Description

I've updated the example in which a comment indicated a line begins a render pass when it actually begins a compute pass.

### Motivation

I think the documentation examples being consistent is important to not cause confusion when learning new concepts.


sideshowbarker:
Aurora, congrats on landing your first docs change here — welcome aboard 🎉

PR #42979, opened 2026-01-30
Update Fx148 relnotes to include initial about:blank

vinhill:
### Description

Quoting from the change:
> The initial `about:blank` document now loads synchronously. A browsing context's first navigation may resolve to `about:blank` (for example, when the initial URL is empty or explicitly set to `about:blank`). In these cases, Firefox no longer replaces the initial empty document with a second, asynchronously loaded one, and instead fires the `load` event synchronously on the initial document.

### Motivation

Changing the timing of `about:blank` loads solves a long-standing web-compat issue. But we've seen web pages or addons break, so developers might want to know about this change.

To name some examples of what changed:
- `window.open("about:blank").onload = handler` the handler won't run anymore because load already fired
- The same with `ifr = document.createElement("iframe"); document.body.append(ifr); ifr.onload = handler;`
- If an empty iframe is appended and synchronously removed, declarative extension content scripts could execute in it anyway.
- `ifr = document.createElement("iframe"); ifr.onload = handler; document.body.append(ifr); ifr.src = other` will observe two load events now because `src` was set too late.

- https://bugzilla.mozilla.org/show_bug.cgi?id=543435
- https://www.firefox.com/en-US/firefox/148.0beta/releasenotes/

### DIFF
--- a/files/en-us/mozilla/firefox/releases/148/index.md
+++ b/files/en-us/mozilla/firefox/releases/148/index.md
@@ -18,9 +18,9 @@ Firefox 148 is the current [Beta version of Firefox](https://www.firefox.com/en-
 
 <!-- ### Developer Tools -->
 
-<!-- ### HTML -->
+### HTML
 
-<!-- No notable changes. -->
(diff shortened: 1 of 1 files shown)


pepelsbey:
Hey! Thank you, it looks very good. The only question I have is about the section. Looking back to how we used the HTML section, it usually has something to do with markup, elements, attributes, etc.

This change looks a bit different. More like API or DOM. Would you consider other sections for this change?

vinhill:
Hey!
I think HTML is the better fit as this is about the document/navigation lifecycle. I only see DOM as subsection of API and this is not an API change.

PR #42981, opened 2026-01-30
Add a couple of headings in the clickjacking guide

wbamberg:
This just adds subheadings for the different clickjacking defenses, to make it easier to scan and to allow us to link to individual defenses. Happy to take suggestions for the wording here, I'm not in love with "controlling cookie transmission".

### DIFF
--- a/files/en-us/web/security/attacks/clickjacking/index.md
+++ b/files/en-us/web/security/attacks/clickjacking/index.md
@@ -63,7 +63,7 @@ The `frame-ancestors` directive is a replacement for `X-Frame-Options`. By setti
 
 If `frame-ancestors` and `X-Frame-Options` are both set, then browsers that support `frame-ancestors` will ignore `X-Frame-Options`.
 
-### Restricting cross-site cookie transmission
+### Restricting cross-site cookies
 
 As an additional partial mitigation, sites should set the [`SameSite`](/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) cookie attribute for session cookies to `Lax` or `Strict`. Requests from embedded contexts such as `<iframe>` elements that are not {{glossary("Site", "same-site")}} with the top-level document will not include these cookies, and the server will therefore not treat the request as coming from a logged-in user.
 


files/en-us/web/security/attacks/clickjacking/index.md
```diff
@@ -50,6 +50,8 @@ If the user tries to click "Click here for a free kitten!", they will actually b
 
 ## Clickjacking defenses
 
+### Controlling embedding
+
 Clickjacking depends on the target website being embedded in the attacker's decoy site inside an `<iframe>`. The main defense is to disallow or at least control this capability.
```
hamishwillee: This says "at least control this capability" but all we talk about is disallowing it. I am not sure what controlling would mean unless you mean "only allow embedding in same-origin" or similar, in which case we should say that.

Perhaps consider `### Disallowing embedding`?

files/en-us/web/security/attacks/clickjacking/index.md
```diff
@@ -61,6 +63,8 @@ The `frame-ancestors` directive is a replacement for `X-Frame-Options`. By setti
 
 If `frame-ancestors` and `X-Frame-Options` are both set, then browsers that support `frame-ancestors` will ignore `X-Frame-Options`.
 
+### Controlling cookie transmission
```
hamishwillee: Maybe `### Restrict cross-site cookies` (since you don't love this).

hamishwillee: FWIW Adding these headings is IMO good, even if they are not optimal.

files/en-us/web/security/attacks/clickjacking/index.md
wbamberg: The reason to use "controlling" is that frame-ancestors gives you the ability to exercise fine-grained control, not just to disallow embedding. But I see that the text does not mention this. So perhaps it would be better to update the first bullet, to mention this?

I like "restricting" better than "controlling" though.

wbamberg: I mean otherwise people might think that they can _only_ disallow, so if they do need framing in some specific cases, they might think they can't defend against clickjacking.

hamishwillee: >  So perhaps it would be better to update the first bullet, to mention this?

I'd change "The main defense is to disallow or at least control this capability." to use restrict here too.

Maybe instead of the first bullet update  "The `frame-ancestors` directive is a replacement for `X-Frame-Options`." with something like " (which additionally allows fine grain control over ...)

wbamberg: -> https://github.com/mdn/content/pull/42981/commits/134a1d9bd533eb0d771e0f32fb1da7b4d54e5e5f

files/en-us/web/security/attacks/clickjacking/index.md
wbamberg: -> https://github.com/mdn/content/pull/42981/commits/134a1d9bd533eb0d771e0f32fb1da7b4d54e5e5f

I combined them! Not sure if this is better tbh but I agree, even if the wording isn't ideal it is better to have them.

files/en-us/web/security/attacks/clickjacking/index.md
hamishwillee: Cool. Ignore my comment.

files/en-us/web/security/attacks/clickjacking/index.md
```diff
@@ -50,17 +50,21 @@ If the user tries to click "Click here for a free kitten!", they will actually b
...
-- The {{httpheader("X-Frame-Options")}} response header.
+- The [`frame-ancestors` directive](/en-US/docs/Web/HTTP/Guides/CSP#clickjacking_protection) in a [content security policy](/en-US/docs/Web/HTTP/Guides/CSP). With this directive, you can control exactly which other documents are allowed to embed yours.
+- The {{httpheader("X-Frame-Options")}} response header. This header is less fine-grained: it just enables you to disable embedding entirely, or allow it only for [same-origin](/en-US/docs/Web/Security/Defenses/Same-origin_policy) documents.
 
 The `frame-ancestors` directive is a replacement for `X-Frame-Options`. By setting `X-Frame-Options` as well as `frame-ancestors`, you can prevent embedding in browsers that don't support `frame-ancestors`. As [browser support for `frame-ancestors` is very good](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors#browser_compatibility), this is not a major concern.
 
 If `frame-ancestors` and `X-Frame-Options` are both set, then browsers that support `frame-ancestors` will ignore `X-Frame-Options`.
 
+### Restricting cross-site cookie transmission
```
hamishwillee:
```suggestion
### Restricting cross-site cookies
```

hamishwillee (approved):
See https://github.com/mdn/content/pull/42981/files#r2752220645 but otherwise approved and you can merge at will.

PR #42982, opened 2026-01-30
Update template element description for CSS scoping

SuperMo0:
Clarified why the scoping of CSS styles in the template element happens.

### Description

Added a clarification of why the styles are scoped to the template fragment.

### Motivation

I was reading the documentation and I got confused as the JavaScript follow up was still below.


sideshowbarker:
Mowafak, congrats on landing your first docs change here — welcome aboard 🎉

PR #42986, opened 2026-02-01
Update wording around multi-keyword syntax for the `display` property

romainmenke:
### Description

The current docs for `display` warn slightly against using the multi-keyword syntax and have some incorrect content (e.g. `block` is not a precomposed value, only `inline-block` is).

### Motivation

Given that multi-keyword syntax is now Baseline Widely Available it seems a good time to update the wording to promote this syntax a bit more.

The multi-keyword syntax makes it easier to fully understand how both inner and outer display modes function.

### Additional details

https://drafts.csswg.org/css-display-3/#the-display-properties

None

### DIFF
--- a/files/en-us/web/css/reference/properties/display/index.md
+++ b/files/en-us/web/css/reference/properties/display/index.md
@@ -130,7 +130,7 @@ The keyword values can be grouped into six value categories.
       - : The element generates one or more inline boxes that do not generate line breaks before or after themselves. In normal flow, the next element will be on the same line if there is space.
 
 > [!NOTE]
-> When a display property is set with only an **outer** value (e.g., `display: block` or `display: inline`), the inner value defaults to `flow` (e.g., `display: block flow` and `display: inline flow`).
+> When a display property is specified with only an **outer** value (e.g., `display: block` or `display: inline`), the inner value defaults to `flow` (e.g., `display: block flow` and `display: inline flow`).
 
 > [!NOTE]
 > You may use the single-value syntax as a fallback for multi-keyword syntax, for example `display: inline flex` could have the following fallback
@@ -167,7 +167,7 @@ The keyword values can be grouped into six value categories.
       - : The element behaves like an inline-level element and lays out its content according to the ruby formatting model. It behaves like the corresponding HTML {{HTMLElement("ruby")}} elements.
 
 > [!NOTE]
-> When a display property is set with only an **inner** value (e.g., `display: flex` or `display: grid`), the outer value defaults to `block` (e.g., `display: block flex` and `display: block grid`).
+> When a display property is specified with only an **inner** value (e.g., `display: flex` or `display: grid`), the outer value defaults to `block` (e.g., `display: block flex` and `display: block grid`).
 
 ### List Item
 


files/en-us/web/css/reference/properties/display/index.md
```diff
@@ -129,10 +137,10 @@ The keyword values can be grouped into six value categories.
       - : The element generates one or more inline boxes that do not generate line breaks before or after themselves. In normal flow, the next element will be on the same line if there is space.
 
 > [!NOTE]
-> When browsers that support multi-keyword syntax encounter a display property that only has an **outer** value (e.g., `display: block` or `display: inline`), the inner value is set to `flow` (e.g., `display: block flow` and `display: inline flow`).
+> When browsers encounter a display property that only has an **outer** value (e.g., `display: block` or `display: inline`), the inner value is set to `flow` (e.g., `display: block flow` and `display: inline flow`).
 
 > [!NOTE]
-> To be sure layouts work on older browsers, you may use single-value syntax, for example `display: inline flex` could have the following fallback
+> You may use single-value syntax as fallback for multi-keyword syntax, for example `display: inline flex` could have the following fallback
```
romainmenke:
```suggestion
> You may use the more widely supported single-value syntax as a fallback for multi-keyword syntax, for example `display: inline flex` could have the following fallback
```

files/en-us/web/css/reference/properties/display/index.md
```diff
@@ -76,34 +76,42 @@ code {
...
 display: grid;
-display: inline-grid;
-display: flow-root;
-
-/* Box suppression */
-display: none;
-display: contents;
+display: table;
 
 /* multi-keyword syntax */
```
estelle: we should either use "short syntax" and "full syntax" OR "single-keyword" and "multi-keyword".

`display: inline list-item` is both "short syntax" and "multi-keyword".

As the values are broken down by data-type under values, I  suggest we keep it to three headers in the syntax section:

1. short display
2. full display
3. global values

files/en-us/web/css/reference/properties/display/index.md
```diff
@@ -129,10 +137,10 @@ The keyword values can be grouped into six value categories.
       - : The element generates one or more inline boxes that do not generate line breaks before or after themselves. In normal flow, the next element will be on the same line if there is space.
 
 > [!NOTE]
-> When browsers that support multi-keyword syntax encounter a display property that only has an **outer** value (e.g., `display: block` or `display: inline`), the inner value is set to `flow` (e.g., `display: block flow` and `display: inline flow`).
+> When browsers encounter a display property that only has an **outer** value (e.g., `display: block` or `display: inline`), the inner value is set to `flow` (e.g., `display: block flow` and `display: inline flow`).
```
estelle: is this true?


```js
let styleMap = el.computedStyleMap();
styleMap.get("display").value /* returns outer without inner */
```

I think it may be more accurate to write:

```suggestion
When a display property has only an **outer** value set (e.g., `display: block` or `display: inline`), the inner value defaults to `flow` (e.g., `display: block flow` and `display: inline flow`).
```

files/en-us/web/css/reference/properties/display/index.md
```diff
@@ -129,10 +137,10 @@ The keyword values can be grouped into six value categories.
...
 
 > [!NOTE]
-> To be sure layouts work on older browsers, you may use single-value syntax, for example `display: inline flex` could have the following fallback
+> You may use the more widely supported single-value syntax as a fallback for multi-keyword syntax, for example `display: inline flex` could have the following fallback
```
estelle:
```suggestion
> You may use the single-value syntax as a fallback for multi-keyword syntax, for example `display: inline flex` could have the following fallback
```
multi-keyword is baseline widely available, so it's all well supported.

files/en-us/web/css/reference/properties/display/index.md
```diff
@@ -166,7 +174,7 @@ The keyword values can be grouped into six value categories.
       - : The element behaves like an inline-level element and lays out its content according to the ruby formatting model. It behaves like the corresponding HTML {{HTMLElement("ruby")}} elements.
 
 > [!NOTE]
-> When browsers that support multi-keyword syntax encounter a display property that only has an **inner** value (e.g., `display: flex` or `display: grid`), the outer value is set to `block` (e.g., `display: block flex` and `display: block grid`).
+> When browsers encounter a display property that only has an **inner** value (e.g., `display: flex` or `display: grid`), the outer value is set to `block` (e.g., `display: block flex` and `display: block grid`).
```
estelle: same comment as above. it defaults, but it isn't set (as in, there is no effect if querying the `display` value)

files/en-us/web/css/reference/properties/display/index.md
romainmenke: I agree! Good catch

romainmenke:
```suggestion
> When a display property has only an **outer** value (e.g., `display: block` or `display: inline`), the inner value defaults to `flow` (e.g., `display: block flow` and `display: inline flow`).
```

romainmenke: I've applied a slight variant of your suggestions. Please let me know if this doesn't cover your feedback.

files/en-us/web/css/reference/properties/display/index.md
```diff
@@ -129,10 +137,10 @@ The keyword values can be grouped into six value categories.
       - : The element generates one or more inline boxes that do not generate line breaks before or after themselves. In normal flow, the next element will be on the same line if there is space.
 
 > [!NOTE]
-> When browsers that support multi-keyword syntax encounter a display property that only has an **outer** value (e.g., `display: block` or `display: inline`), the inner value is set to `flow` (e.g., `display: block flow` and `display: inline flow`).
+> When a display property has only an **outer** value (e.g., `display: block` or `display: inline`), the inner value defaults to `flow` (e.g., `display: block flow` and `display: inline flow`).
```
romainmenke:
```suggestion
> When a display property is specified with only an **outer** value (e.g., `display: block` or `display: inline`), the inner value defaults to `flow` (e.g., `display: block flow` and `display: inline flow`).
```

`When a display property is specified with only an **outer** value`

vs.

`When a display property has only an **outer** value set` as suggested by @estelle

`value set` could be read as a set of values, which was confusing to me.

This wording matches: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Display/Multi-keyword_syntax#:~:text=%22If%20a%20%3Cdisplay%2Doutside%3E%20value%20is%20specified%20but%20%3Cdisplay%2Dinside%3E%20is%20omitted%2C%20the%20element's%20inner%20display%20type%20defaults%20to%20flow.%22

files/en-us/web/css/reference/properties/display/index.md
romainmenke:
```suggestion
> When a display property is set with only an **inner** value (e.g., `display: flex` or `display: grid`), the outer value defaults to `block` (e.g., `display: block flex` and `display: block grid`).
```

files/en-us/web/css/reference/properties/display/index.md
romainmenke: I've applied this suggestion, please let me know if I misunderstood or if further changes are needed.

- moved `precomposed values` into `short display`
- moved `none` and `contents` into `short display` ([as seen in the table here](https://drafts.csswg.org/css-display-3/#the-display-properties))
- omitted `table-row` from this section, it doesn't fit in any of these three sections

romainmenke:
Thank you for the review @estelle 🙇 
This is ready for a second look

estelle (approved):
Thanks! 🎉

PR #42992, opened 2026-02-02
Fix the <em> and <i> demos

Finesse:
### Description

Fixes the demos on https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/em and https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/i.

### Motivation

The demos claim to how the elements mark text that has stress emphasis, but in fact all the text in the demos looks the same. At least on macOS.

image

image

### Additional details

The solution is not very good because it's ad-hoc. That is, the demos receive an irrelevant `p { ... }` CSS rule just to fix the bug. Ideally, the default `font-family` of the demos should be modified to support all font styles.

### DIFF
--- a/files/en-us/web/html/reference/elements/em/index.md
+++ b/files/en-us/web/html/reference/elements/em/index.md
@@ -22,6 +22,10 @@ The **`<em>`** [HTML](/en-US/docs/Web/HTML) element marks text that has stress e
 em {
   /* Add your styles here */
 }
+
+p {
+  font-family: reset;
+}
 ```
 
 ## Attributes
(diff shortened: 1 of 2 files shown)


Josh-Cena:
This is a known problem with Safari, not macOS. It works in Chrome:

image

I think this relates to https://github.com/mdn/yari/issues/7203. It was fixed in the main content but not in the playground. I don't think we should fix it in content though, because this could happen for literally every page.

chrisdavidmills:
I agree with @Josh-Cena here. We should try to fix this at the system level.

@caugner, is it worth opening an issue against fred for this, so it can be investigated?

caugner:
> don't think we should fix it in content though, because this could happen for literally every page.

Agree, this is best fixed on the platform level (dex/fred).

> is it worth opening an issue against fred for this, so it can be investigated?

Filed https://github.com/mdn/fred/issues/1269.

chrisdavidmills:
> Filed [mdn/fred#1269](https://github.com/mdn/fred/issues/1269).

Thanks @caugner. I'll close this one.

PR #43016, opened 2026-02-04
Document trusted types directives in the CSP guide

wbamberg:
The [CSP guide](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) didn't mention the trusted types directives, this PR adds a section on that. I didn't want to get too much into the details, since we talk about it quite extensively in the TT API page.

### DIFF
--- a/files/en-us/web/http/guides/csp/index.md
+++ b/files/en-us/web/http/guides/csp/index.md
@@ -531,7 +531,7 @@ Some APIs in the web platform are known as _injection sinks_. These are APIs tha
 
 If an attacker can supply some specially crafted input to your website, and your website passes it to one of these injection sinks, then the attacker can execute malcious code.
 
-We've seen that we can use a CSP to disable APIs which are almost always dangerous, like `eval()`, but we might want to be able to use many injection sinks such as `innerHTML`, while ensuring that the input is safe. The practice of ensuring that the input is safe is called [_sanitizing_](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input.
+Some injection sinks, like `eval()`, are very hard to use safely, and we've seen that a CSP will typically [block them entirely](#eval_and_similar_apis). Others can be made safer if the input to them is processed so as to remove unsafe elements. This practice is called [_sanitization_](/en-US/docs/Web/Security/Attacks/XSS#sanitization).
 
 ### The Trusted Types API
 


files/en-us/web/http/guides/csp/index.md
```diff
@@ -58,9 +65,15 @@ A cross-site scripting (XSS) attack is one in which an attacker is able to execu
...
 
-Websites should protect themselves against XSS by sanitizing this input before including it in the page. A CSP provides a complementary protection, which can protect the website even if sanitization fails.
+Websites should protect themselves against XSS by sanitizing this input before including it in the page.
 
-If sanitization does fail, there are various forms the injected malicious code can take in the document, including:
+> [!NOTE]
+> a CSP can actually help protect against XSS in two different ways:
+>
+> 1. It can help ensure that input is sanitized before being used in the client: we discuss this later on in [Requiring trusted types](#requiring_trusted_types).
+> 2. By controlling resource loads, a CSP can provide a defense in depth against XSS, protecting the website even if sanitization fails. This is the XSS defense that we will discuss in this section.
```
hamishwillee:
```suggestion
> A CSP can actually help protect against XSS in two different ways:
>
> - It can help ensure that input is sanitized before being used in the client: we discuss this later on in [Requiring trusted types](#requiring_trusted_types).
> - By controlling resource loads, a CSP can provide a defense in depth against XSS, protecting the website even if sanitization fails. This is the XSS defense that we will discuss in this section.
```

files/en-us/web/http/guides/csp/index.md
```diff
@@ -506,6 +519,34 @@ The browser will upgrade the first link to HTTPS, but not the second, as it is n
 
 This directive is not a substitute for the {{httpheader("Strict-Transport-Security")}} header (also known as HSTS), because it does not upgrade external links to a site. Sites should include this directive and the `Strict-Transport-Security` header.
 
+## Requiring trusted types
```
hamishwillee, line 522: FYI only, CSP also has the keywor https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#trusted-types-eval - I don't think it should be mentioned here _unless_ you update `eval` above to point down here.

files/en-us/web/http/guides/csp/index.md
```diff
@@ -506,6 +519,34 @@ The browser will upgrade the first link to HTTPS, but not the second, as it is n
...
+
+The [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) and [`trusted-types`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/trusted-types) directives enable you to defend against client-side [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, by ensuring that any input has been passed through a transformation to make it safe before it is passed to a web platform API that might otherwise execute it as code.
```
hamishwillee, line 524: This feels like an overlong sentence. Too much use of "passed" in second sentence. The "to make it safe" is an intent.

```suggestion
The [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) and [`trusted-types`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/trusted-types) directives enable you to defend against client-side [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks.
They ensure that any input must be passed through a transformation function, providing an opportunity to make it safe before it is sent to a web platform API that might otherwise execute it as code.
```

files/en-us/web/http/guides/csp/index.md
```diff
@@ -506,6 +519,34 @@ The browser will upgrade the first link to HTTPS, but not the second, as it is n
...
+
+### Injection sinks and sanitization
+
+Some APIs in the web platform are known as _injection sinks_. These are APIs that can be passed some input, usually in the form of a string, and that can interpret that input as code. In this guide we've already seen `eval()`, but there are many other injection sinks, such as {{domxref("Element.innerHTML")}} or {{domxref("Document.write()")}}.
+
+If an attacker can supply some specially crafted input to your website, and your website passes it to one of these injection sinks, then the attacker can execute a client-side [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attack.
```
hamishwillee: We already mention XSS just above. An option

```suggestion
If an attacker can supply some specially crafted input to your website, and your website passes it to one of these injection sinks, then the attacker can execute malcious code.
```

files/en-us/web/http/guides/csp/index.md
```diff
@@ -506,6 +519,34 @@ The browser will upgrade the first link to HTTPS, but not the second, as it is n
...
+
+We've seen that we can use a CSP to disable APIs which are almost always dangerous, like `eval()`, but we might want to be able to use many injection sinks such as `innerHTML`, while ensuring that the input is safe. The practice of ensuring that the input is safe is called [_sanitizing_](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input.
```
hamishwillee: The "But we might" clause doesn't feel quite like it follows on. I think I was looking for words like "But sometimes we might want to control the use of APIs that are only potentially dangerous", or perhaps a mention that CSP can be used for this purpose. 

I don't have a proposal here, but I can think more needed.

files/en-us/web/http/guides/csp/index.md
```diff
@@ -506,6 +519,34 @@ The browser will upgrade the first link to HTTPS, but not the second, as it is n
...
+
+### The Trusted Types API
+
+With the [Trusted Types API](/en-US/docs/Web/API/Trusted_Types_API), you can pass _trusted types_ into injection sinks, instead of strings. Trusted types are the result of passing potentially dangerous input through a transformation function. This transformation typically sanitizes the input, by removing any elements that might make it executable (such as {{htmlelement("script")}} tags).
```
hamishwillee:
```suggestion
With the [Trusted Types API](/en-US/docs/Web/API/Trusted_Types_API), you can pass _trusted types_ into injection sinks, instead of strings. Trusted types are objects that result from passing potentially dangerous input through a transformation function. This transformation typically sanitizes the input, by removing any elements that might make it executable (such as {{htmlelement("script")}} tags).
```

files/en-us/web/http/guides/csp/index.md
```diff
@@ -506,6 +519,34 @@ The browser will upgrade the first link to HTTPS, but not the second, as it is n
 
 This directive is not a substitute for the {{httpheader("Strict-Transport-Security")}} header (also known as HSTS), because it does not upgrade external links to a site. Sites should include this directive and the `Strict-Transport-Security` header.
 
+## Requiring trusted types
+
+The [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) and [`trusted-types`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/trusted-types) directives enable you to defend against client-side [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks, by ensuring that any input has been passed through a transformation to make it safe before it is passed to a web platform API that might otherwise execute it as code.
+
```
hamishwillee, line 527: I also wonder if we should be even more direct here and start by mentioning the API Something like

```suggestion
The [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) and [`trusted-types`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/trusted-types) directives can be used to enforce the [Trusted Types API](/en-US/docs/Web/API/Trusted_Types_API).
This allows to defend against client-side [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks by mandating that any input must be passed through a transformation function, providing an opportunity to make it safe before it is sent to a web platform API that might otherwise execute it as code.
```

hamishwillee (commented):
Thanks, seems about the right level to me. A few comments/suggestions.

files/en-us/web/http/guides/csp/index.md
wbamberg: I like this but can I say:

```suggestion
The [`require-trusted-types-for`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for) and [`trusted-types`](/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/trusted-types) directives can be used to enforce the [Trusted Types API](/en-US/docs/Web/API/Trusted_Types_API).
This allows you to defend against client-side [cross-site scripting (XSS)](/en-US/docs/Web/Security/Attacks/XSS) attacks by mandating that any input must be passed through a transformation function, providing an opportunity to make it safe before it is sent to a web platform API that might otherwise execute it as code.
```

I don't know if grammar like "this allows to", omitting the pronoun, is correct but it's nails on a chalkboard for me.

files/en-us/web/http/guides/csp/index.md
wbamberg: It's a little tricky to talk about because we want to say:

1. some sinks, like eval(), are almost always dangerous, and as we have seen, CSP allows us to disable them entirely
2. other sinks, like innerHTML, have legitimate uses but are dangerous unless their input is sanitized

...but that might lead someone to think that TT only applies to the second type of sink. But if we don't say (1) people might think, hang on, didn't we already talk about eval() and CSPs, why are we talking about them here as well. But also we don't really want to encourage people to allow `eval()`, which talking about how TTs apply to them might come across like ("it's OK to use eval() because of TTs").

files/en-us/web/http/guides/csp/index.md
hamishwillee: Me too. A typo.

files/en-us/web/http/guides/csp/index.md
hamishwillee: That's enough for a suggestion - something like

```suggestion
We've seen that we can use a CSP to disable APIs unless they are hosted on permitted locations ( or always for APIs that are almost always dangerous, like `eval()`).
Sometimes though we want to be able to use APIs that might be dangerous, while ensuring that the input is safe. The practice of ensuring that the input is safe is called [_sanitizing_](/en-US/docs/Web/Security/Attacks/XSS#sanitization) the input.
```

wbamberg: I'm not sure I like this though.

The context of this is injection sinks, so the implication here is that a CSP can disable injection sinks "unless they are hosted on permitted locations", but it can't. I think it is also confusing to start talking about "APIs" in general when we have been talking about injection sinks.

Maybe this?

```suggestion
Some injection sinks, like `eval()`, are very hard to use safely, and we've seen that a CSP will typically [block them entirely](#eval_and_similar_apis). Others can be made safer if the input to them is processed so as to remove unsafe elements. This practice is called [_sanitization_](/en-US/docs/Web/Security/Attacks/XSS#sanitization).
```

This is _slightly_ misleading as it might imply that eval() is not in scope for sanitization, but that is sort of true: even in a TT world, we generally recommend disabling eval(), and there's not really any generic "sanitization" you can do to a JS sink, to make it safe - most of our examples are either blocking it entirely or only allowing a specific set of static values.

hamishwillee: I'm accepting this. Mostly because the slight mislead is beneficial - script sources are never really safer even with sanitization.

hamishwillee (approved):
Approving. Not merging in case you want other reviews.

PR #43019, opened 2026-02-05
revert bot errors

estelle:
moved a section demonstrating backgrounds, borders, and box shadows with examples and code snippets to where it is supposed to be

https://github.com/mdn/content/commit/0c81cbce5f95a0be935724bcd936f5592774eb3a#diff-289a1935e45dfd140c9ff4e9b7f2733a7e499bc2c6fe618df6adeb8dc031c6b7

### DIFF
--- a/files/en-us/web/css/guides/backgrounds_and_borders/index.md
+++ b/files/en-us/web/css/guides/backgrounds_and_borders/index.md
@@ -17,7 +17,7 @@ Box shadows include inset and outset shadow, single or multiple shadows, and sol
 
 The properties in this module also let you define whether cells inside a {{HTMLElement("table")}} should have shared or separate borders.
 
-### Backgrounds, borders, and box shadows in action
+## Backgrounds, borders, and box shadows in action
 
 This sample of borders, backgrounds, and box shadows consists of centered background images made of linear and radial gradients. A series of box shadows make the border appear to "pop". The element on the left has a border image set. The element on the right has a rounded dotted border.
 


Josh-Cena:
The fix was intentional: it's not allowed to start the heading sequence with an H3. You have to either promote it to an H2 (which I would discourage, because H2 headings should be consistent across pages), add a leading H2 heading, or keep the current approach of using "Examples" (which I find reasonable because this is, indeed, an example).

files/en-us/web/css/guides/backgrounds_and_borders/index.md
```diff
@@ -17,6 +17,67 @@ Box shadows include inset and outset shadow, single or multiple shadows, and sol
 
 The properties in this module also let you define whether cells inside a {{HTMLElement("table")}} should have shared or separate borders.
 
+### Backgrounds, borders, and box shadows in action
```
estelle:
```suggestion
## Backgrounds, borders, and box shadows in action
```

estelle:
> The fix was intentional: it's not allowed to start the heading sequence with an H3. You have to either promote it to an H2 (which I would discourage, because H2 headings should be consistent across pages), add a leading H2 heading, or keep the current approach of using "Examples" (which I find reasonable because this is, indeed, an example).

The "in actions" are demos of the features at the top of the CSS module pages. The code is generally hidden and not explained. See https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Page_structures/Page_types/CSS_module_page_template#nameofthemodule_in_action

the bot changed the h2 to an h3 as it added the example header. not sure how or why the bot changed this one only, but glad it didn;t change them all.

PR #43020, opened 2026-02-05
related concepts on css custom function mod page

estelle:
Added related concepts  to the guide.

### DIFF
--- a/files/en-us/web/css/guides/custom_functions_and_mixins/index.md
+++ b/files/en-us/web/css/guides/custom_functions_and_mixins/index.md
@@ -51,8 +51,13 @@ The CSS custom functions and mixins module also introduces the `@mixin`, `@apply
 [CSS values and units](/en-US/docs/Web/CSS/Guides/Values_and_units)
 
 - {{cssxref("abs()")}}
+- {{cssxref("acos()")}}
+- {{cssxref("asin()")}}
+- {{cssxref("atan()")}}
+- {{cssxref("atan2()")}}
 - {{cssxref("calc()")}}
 - {{cssxref("clamp()")}}
+- {{cssxref("cos()")}}
 - {{cssxref("exp()")}}
 - {{cssxref("log()")}}
 - {{cssxref("max()")}}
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/guides/custom_functions_and_mixins/index.md
```diff
@@ -45,6 +45,32 @@ The CSS custom functions and mixins module also introduces the `@mixin`, `@apply
 
 - [Using CSS custom functions](/en-US/docs/Web/CSS/Guides/Custom_functions_and_mixins/Using_custom_functions)
   - : This guide teaches you how to use CSS custom functions and presents some typical use cases.
+ 
+## Related concepts
+
+- {{cssxref("abs()")}}
```
chrisdavidmills, line 53: From what I can see from other module pages, this section should be presented as top-level category bullets (with a link to a page or section about that category, if available), and then second-level bullets for each property, or function, or whatever. For example, https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Anchor_positioning#related_concepts.

So, I think there should be a top-level bullet "Math functions" over the top of these function links, which should be second-level bullets.

files/en-us/web/css/guides/custom_functions_and_mixins/index.md
```diff
@@ -45,6 +45,32 @@ The CSS custom functions and mixins module also introduces the `@mixin`, `@apply
...
+- {{cssxref("max()")}}
+- {{cssxref("min()")}}
+- {{cssxref("mod()")}}
+- {{cssxref("pow()")}}
+- {{cssxref("rem()")}}
+- {{cssxref("round()")}}
+- {{cssxref("sign()")}}
+- {{cssxref("sqrt()")}}
+
+Trigonometric functions
```
chrisdavidmills: Again, make this a top-level bullet, and then have the following links a level below.

files/en-us/web/css/guides/custom_functions_and_mixins/index.md
estelle: they're all in the same spec. i was hoping to separate out trig functions.... but you are right. updated.

PR #43026, opened 2026-02-05
Editorial review: scroll-state scrolled descriptor

chrisdavidmills:
### Description

Chrome 144 adds support for the `@container` at-rule scroll-state `scrolled` descriptor; see https://chromestatus.com/feature/5083137520173056.

This PR adds documentation for this new descriptor.

### DIFF
--- a/files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
+++ b/files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
@@ -1123,7 +1123,7 @@ Next, we define a {{cssxref("@container")}} block that sets the container name w
 ```css hidden live-sample___scrollable live-sample___scrolled live-sample___snapped live-sample___stuck
 @supports not (container-type: scroll-state) {
   body::before {
-    content: "Your browser does not support the `scroll-state` value of the 'container-type' property.";
+    content: "Your browser does not support `scroll-state` container queries.";
     color: black;
     background-color: wheat;
     position: fixed;


files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +326,297 @@ We've hidden the rest of the example CSS for brevity.
...
+}
+```
+
+The {{glossary("scroll container")}} in this example is the `<html>` element itself, denoted as a scroll-state query container with a {{cssxref("container-type")}} value of `scroll-state`. The {{cssxref("container-name")}} isn't strictly necessary but is useful in cases where the code is added to a codebase with multiple scroll-state query containers targeted with different queries.
+
+```css live-sample___scrolled
+html {
+  container-type: scroll-state;
+  container-name: scroller;
+}
```
estelle, line 595: Drive by comment: 

FF and Safari don't support `container-type: scroll-state;`, so add an `@support`

```
@layer supports {
  @supports not (container-type: scroll-state) {
    body::before {
      content: "Your browser does not support the `scroll-state` value for the 'container-type' property.";
      background-color: wheat;
      display: block;
      text-align: center;
      padding: 1rem;
    }
  }
}
```

chrisdavidmills: Good call; I've added a universal banner to the page and included it in all examples (given that all of them rely on this same property for feature detection), so all of the live samples get a non-support banner in non-supporting browsers.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -264,6 +265,23 @@ img {
 }
 ```
 
+```css hidden live-sample___scrollable live-sample___scrolled live-sample___snapped live-sample___stuck
```
estelle: Let's put the "no support" hidden CSS after all other CSS for the all the examples using it so if the reader hits "play", this less relevant CSS is at the bottom of the CSS tab.

It also makes sense to put all hidden CSS after the shown CSS for the same reason. If there's ever a concern over cascade/specificity, put all the hidden CSS in a layer; anything not in a layer will always take precedence (unless !important is used, which it won't be)

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
 
 Try scrolling the document down, and note how the "back-to-top" link appears as a result, animating smoothly from the right side of the viewport due to the `transition`. If you scroll back to the top by activating the link or manually scrolling, the "back-to-top" link transitions off-screen.
 
+## Using `scrolled` queries
+
+Scroll-state [`scrolled`](/en-US/docs/Web/CSS/Reference/At-rules/@container#scrolled) queries, written as `scroll-state(scrolled: value)`, test whether a container's scrolling ancestor was most recently scrolled in the given direction. If not, the query returns false.
+
+The `value` indicates the direction you are testing for, for example:
```
estelle:
```suggestion
Scroll-state [`scrolled`](/en-US/docs/Web/CSS/Reference/At-rules/@container#scrolled) queries, written as `scroll-state(scrolled: <keyword>)`, test whether a container's scrolling ancestor was most recently scrolled in the given direction. If not, the query returns false.

The keyword value indicates the direction you are testing for, for example:
```

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+
+- `block-start`: Tests whether the container was most recently scrolled towards its block-start edge.
+- `right`: Tests whether the container was most recently scrolled towards its right-hand edge.
+- `y`: Tests whether the container was most recently scrolled in either direction along its y axis.
```
estelle:
```suggestion
- `y`: Tests whether the container was most recently scrolled in either direction along its vertical y-axis.
```
or
```suggestion
- `y`: Tests whether the container was most recently scrolled up or down along the y-axis.
```

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+
+If the test passes, the rules inside the `@container` block are applied to descendants of the matching scroll container.
```
estelle:
```suggestion
If the test returns true, the rules inside the `@container` block are applied to the descendants of the matching scroll container.
```

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+
+Let's look at an example with a scrolling container full of content, in which we use a `scrolled` query to display top and bottom content "bars", only when the user is scrolling up or down, respectively.
```
estelle:
```suggestion
Let's look at an example of a scroll container with a `scrolled` query that displays top and bottom content "bars" only when the user is scrolling up or down, respectively.
```

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+
+### HTML
+
+In the HTML we have an {{htmlelement("article")}} element containing enough content to cause the document to scroll, preceded by a couple of {{htmlelement("div")}} elements that represent our top and bottom "bars":
```
estelle:
```suggestion
In our HTML, we have an {{htmlelement("article")}} element containing enough content to cause the document to scroll, preceded by two {{htmlelement("div")}} elements that represent our top and bottom "bars":
```

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+      Aliquam sed posuere ligula, at auctor ipsum. Morbi dignissim accumsan
+      tellus pretium iaculis.
+    </p>
+  </section>
+</article>
+```
+
+### CSS
+
+The "bars" are given some rudimentary styling. Most significantly, they are given a {{cssxref("position")}} value of `fixed`, and given a little bit of spacing on either side using {{cssxref("left")}} and {{cssxref("right")}} values.
```
estelle:
```suggestion
The "bars" are given some rudimentary styling. Most significantly, they are given a {{cssxref("position")}} value of `fixed`, which we offset from either side using {{cssxref("left")}} and {{cssxref("right")}}.
```
only the "fixed" is significant, so tried to downplay the left and right, while still mentioning it.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+.bar {
+  border-radius: 10px;
+  border: 1px solid #000;
+  background: #0009;
+  padding: 10px;
+  color: white;
+  text-shadow: 1px 1px 1px black;
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
estelle, line 567:
```suggestion
  align-items: center;
  
```
add a line of space before the important part to make it easier to spot as there is so much here

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+  border: 2px solid gray;
+  aspect-ratio: 3/2;
+}
+```
+
+```css live-sample___scrolled
+.bar {
+  border-radius: 10px;
+  border: 1px solid #000;
+  background: #0009;
```
estelle:
```suggestion
  background-color: #0009;
```

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  position: fixed;
+  left: 5px;
+  right: 5px;
+}
+```
+
+Next, we give the top "bar" a {{cssxref("top")}} value of `-50px`, and the bottom "bar" a {{cssxref("bottom")}} value of `-50px` to hide them off the top and bottom of the screen by default. We also give them both a {{cssxref("transition")}} so that they smoothly animate when their `top`/`bottom` values are changed.
```
estelle:
```suggestion
Next, we set negative {{cssxref("top")}} and {{cssxref("bottom")}} length values on the top and bottom bars so they are hidden above and below the viewport by default. We add a {{cssxref("transition")}} to smoothly animate them into view when their `top` or `bottom` value changes.
```

estelle: That said, we likely should translate them in, since animating box model properties is not a best practice. We can still keep the -50px, but we can translate them by 55px in the visible state.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+
+```css live-sample___scrolled
+#top-bar {
+  top: -50px;
+  transition: 0.6s top;
+}
+
+#bottom-bar {
+  bottom: -50px;
+  transition: 0.6s bottom;
```
estelle: maybe do `transition: translate 0.6s` on both.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+  #bottom-bar {
+    bottom: 5px;
+  }
+}
+
+@container scroller scroll-state(scrolled: block-start) {
+  #top-bar {
+    top: 5px;
+  }
+}
```
estelle:
```suggestion
@container scroller scroll-state(scrolled: block-start) {
  #top-bar {
    top: 5px;
  }
}

@container scroller scroll-state(scrolled: block-end) {
  #bottom-bar {
    bottom: 5px;
  }
}
```
just. to match the order of the previous code block.
though, i think we should use translate.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+The {{glossary("scroll container")}} in this example is the `<html>` element itself, denoted as a scroll-state query container with a {{cssxref("container-type")}} value of `scroll-state`. The {{cssxref("container-name")}} isn't strictly necessary but is useful in cases where the code is added to a codebase with multiple scroll-state query containers targeted with different queries.
+
+```css live-sample___scrolled
+html {
+  container-type: scroll-state;
+  container-name: scroller;
+}
+```
+
+Next, we define two {{cssxref("@container")}} blocks that set the container name targeted by these queries, and queries of `scrolled: block-end` and `scrolled: block-start`. Respectively, these queries apply the rules contained within the block only if the `<html>` element was most recently scrolled toward its block-end edge or block-start edge — in other words, when the container is scrolled downwards or upwards. When those conditions become true, the bar referenced inside the block has a new `bottom`/`top` value set to cause it to transition on-screen.
```
estelle:
```suggestion
Next, we define two {{cssxref("@container")}} blocks that set the container name targeted by these queries, and queries of `scrolled: block-end` and `scrolled: block-start`. These queries apply the rules contained within the block only if the `<html>` element was most recently scrolled toward its block-end edge or block-start edge — in other words, when the container is scrolled down or up — respectively. When either condition becomes true, the bar referenced inside the block has a new `bottom` or `top` value, which causes it to transition on-screen. The bar referenced in the condition that is no longer true transitions off the screen.
```

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,6 +343,297 @@ We've hidden the rest of the example CSS for brevity.
...
+
+### Result
+
+{{EmbedLiveSample("scrolled", "100%", "400px")}}
+
+Try scrolling the document up and down, and note how the different bars appear as a result, animating smoothly on- and off-screen.
+
 ## Using `snapped` queries
 
 Relevant only when [scroll snapping](/en-US/docs/Web/CSS/Guides/Scroll_snap) is implemented, scroll-state [`snapped`](/en-US/docs/Web/CSS/Reference/At-rules/@container#snapped) queries (written as `scroll-state(snapped: value)`) test whether a container is going to be snapped to a [scroll snap container](/en-US/docs/Glossary/Scroll_snap#scroll_snap_container) ancestor along the given axis. If not, the query returns false.
```
estelle:
```suggestion
Relevant only when [scroll snapping](/en-US/docs/Web/CSS/Guides/Scroll_snap) is implemented, scroll-state [`snapped`](/en-US/docs/Web/CSS/Reference/At-rules/@container#snapped) queries, written as `scroll-state(snapped: value)`, test whether a container is going to be snapped to a [scroll snap container](/en-US/docs/Glossary/Scroll_snap#scroll_snap_container) ancestor along the given axis. If not, the query returns false.
```

files/en-us/web/css/reference/at-rules/@container/index.md
```diff
@@ -175,7 +178,7 @@ Scroll-state container descriptors are specified inside the `<container-conditio
 }
 ```
 
-Supported keywords for scroll-state container descriptors include physical and {{glossary("flow relative values")}}
+Supported keywords for scroll-state container descriptors include physical and {{glossary("flow relative values")}}.
```
estelle:
```suggestion
Supported keywords for scroll-state container descriptors include {{glossary("physical properties", "physical")}} and {{glossary("flow relative values")}}.
```

files/en-us/web/css/reference/at-rules/@container/index.md
```diff
@@ -216,6 +219,45 @@ Supported keywords for scroll-state container descriptors include physical and {
     }
     ```
 
+- `scrolled`
+  - : Queries whether the container was most recently scrolled in a specified direction. Valid `scrolled` values include the following keywords:
+    - `none`
+      - : The container is not a {{glossary("scroll container")}} or otherwise has not recently been scrolled in any direction.
```
estelle: what does "recently" mean. Does it change from "top" to "none" after a period of time?

files/en-us/web/css/reference/at-rules/@container/index.md
```diff
@@ -216,6 +219,45 @@ Supported keywords for scroll-state container descriptors include physical and {
...
+    - `inline-start`
+      - : The container was most recently scrolled towards its inline-start edge.
+    - `inline-end`
+      - : The container was most recently scrolled towards its inline-end edge.
+    - `block`
+      - : The container was most recently scrolled towards either its block-start or block-end edges.
+    - `inline`
+      - : The container was most recently scrolled towards either its inline-start or inline-end edges.
+
+    If the test passes, the rules inside the `@container` block are applied to descendants of the scroll container.
```
estelle:
```suggestion
    If the test returns true, the rules nested in the `@container` block are applied to the descendants of the scroll container.
```

files/en-us/web/css/reference/at-rules/@container/index.md
```diff
@@ -216,6 +219,45 @@ Supported keywords for scroll-state container descriptors include physical and {
...
+
+    To evaluate whether a container has recently been scrolled, without being concerned about the direction, use the `none` value with the `not` operator:
+
+    ```css
+    @container not scroll-state(scrolled: none) {
+      /* … */
+    }
+    ```
```
estelle, line 259: line 251 to here should be in a description section, not in the values, but since this page is already including description content in a list of descriptor values in a list of descriptors that should be linking to individual descriptor pages, I get it. This whole page should be reworked. Up for it 😉 ? 😛

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: OK, I've moved the `@supports` block to the end of the articles, and in the future, I'm going to bear what you've said about this in mind. That said, I'm not going to spend considerable time restructuring all the rest of the CSS source of this article. It is fairly complicated, there are quite a few shared bits, and I don't think the benefit to MDN Playground users outweighs the effort required, and the issue that it'll probably make it harder to edit for future editors.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: I agree with this change, but I've kept this as a reminder rather than committing it, as there are four instances to update in the article. I've made those changes in my next commit.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: updated in my next commit. Again, I kept it as a reminder rather than committing it, as I thought there'd be multiple instances, but there was only one in the end.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: Done in my commit; I managed to miss this in my first pass.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: done

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: done

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: Good call. I've used your suggestion here, but I've also updated the code and descriptions to use `translate` transitions rather than `top/`bottom` transitions.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: done

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: order changed

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: I've used some of these suggestions.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
chrisdavidmills: Updated in my next commit. I kept this as a reminder in case there were other instances to fix.

files/en-us/web/css/reference/at-rules/@container/index.md
chrisdavidmills: Agree with the change, but I also did some additional editing in my next commit.

files/en-us/web/css/reference/at-rules/@container/index.md
chrisdavidmills: No, it doesn't do that, but I can see how the wording would be a little confusing. In this case, I've changed it to "previously".

I think "most recently" is still suitable for the other bullets — the test is what the most recent scroll direction was, if any.

files/en-us/web/css/reference/at-rules/@container/index.md
chrisdavidmills: LOL, no! ;-)

chrisdavidmills: Seriously, though, I think that should be a separate PR, whoever ends up doing it.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,11 +326,303 @@ We've hidden the rest of the example CSS for brevity.
 
 Try scrolling the document down, and note how the "back-to-top" link appears as a result, animating smoothly from the right side of the viewport due to the `transition`. If you scroll back to the top by activating the link or manually scrolling, the "back-to-top" link transitions off-screen.
 
+## Using `scrolled` queries
+
+Scroll-state [`scrolled`](/en-US/docs/Web/CSS/Reference/At-rules/@container#scrolled) queries, written as `scroll-state(scrolled: <keyword>)`, test whether a container's scrolling ancestor was most recently scrolled in the given direction. If not, the query returns false.
+
+The keyword value indicates the direction you are testing for, for example:
```
estelle:
```suggestion
The keyword value indicates the direction you are testing. For example:
```

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,11 +326,303 @@ We've hidden the rest of the example CSS for brevity.
...
+
+- `block-start`: Tests whether the container was most recently scrolled towards its block-start edge.
+- `right`: Tests whether the container was most recently scrolled towards its right-hand edge.
+- `y`: Tests whether the container was most recently scrolled up or down along the y axis.
```
estelle, line 337:
```suggestion
- `y`: Tests whether the container was most recently scrolled up or down along the y axis.
- `none`: Tests whether the container is not a {{glossary("scroll container")}} or has not been scrolled in any direction since rendering.
```
none is cool, and likely something people won't think about, but if they heard about it, would definitely use.

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,11 +326,303 @@ We've hidden the rest of the example CSS for brevity.
...
+  transition: 0.6s translate;
+}
+
+#bottom-bar {
+  bottom: -50px;
+  transition: 0.6s translate;
+}
+```
+
+The {{glossary("scroll container")}} in this example is the `<html>` element itself, denoted as a scroll-state query container with a {{cssxref("container-type")}} value of `scroll-state`. The {{cssxref("container-name")}} isn't strictly necessary but is useful in cases where the code is added to a codebase with multiple scroll-state query containers targeted with different queries.
```
estelle:
```suggestion
The {{glossary("scroll container")}} in this example is the `<html>` element itself, denoted as a scroll-state query container with a {{cssxref("container-type")}} value of `scroll-state`. The {{cssxref("container-name")}} isn't strictly necessary, but it is useful when a codebase has multiple scroll-state query containers targeted with different queries.
```

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -325,11 +326,303 @@ We've hidden the rest of the example CSS for brevity.
...
+
+```css live-sample___scrolled
+html {
+  container-type: scroll-state;
+  container-name: scroller;
+}
+```
+
+Next, we define two {{cssxref("@container")}} blocks that set the container name targeted by these queries, and queries of `scrolled: block-end` and `scrolled: block-start`. These queries apply the rules contained within the block only if the `<html>` element was most recently scrolled toward its block-end edge or block-start edge — in other words, when the container is scrolled down or up. When either condition becomes true, the bar referenced inside the block has a `translate` value set to cause it to transition on-screen. The bar referenced in the `@condition` that is no longer true transitions off the screen.
```
estelle:
```suggestion
Next, we define two {{cssxref("@container")}} blocks, both of which target the `scroller` container name. The first block defines a query `scrolled: block-end` and the second defines a query `scrolled: block-start`. Respectively, these queries apply the rules contained within their block only if the `<html>` element was most recently scrolled toward its block-end edge or block-start edge. In other words, when the container is scrolled down or up. When either condition becomes true, the bar referenced inside the block has a `translate` value set to cause it to transition on-screen. The bar referenced in the `@condition` that is no longer true transitions off-screen.
```

i got a little confused with "set the container name targeted by these queries" and then went overboard on the edits. If you disagree with the suggestion, the part that made me go huh was mostly "container name targeted"

files/en-us/web/css/guides/conditional_rules/container_scroll-state_queries/index.md
```diff
@@ -826,6 +1119,23 @@ Next, we define a {{cssxref("@container")}} block that sets the container name w
 }
 ```
 
+```css hidden live-sample___scrollable live-sample___scrolled live-sample___snapped live-sample___stuck
+@supports not (container-type: scroll-state) {
+  body::before {
+    content: "Your browser does not support the `scroll-state` value of the 'container-type' property.";
```
estelle:
```suggestion
    content: "Your browser does not support `scroll-state` container queries.";
```
THis is the basic test of supporting scroll-state container queries, right? let's make it a bit more generic so they don't start trying to figure out if this is the only value that's not supported, but it's a general feature not being supported, not just a value.

PR #43043, opened 2026-02-06
Update outdated information on border-width keywords

Psychpsyo:
### Description

This changes the note about the meaning of the border-width values `thin`, `medium` and `thick` being implementation-defined to instead note their spec widths.

### Motivation

The spec defines this nowadays.

### Additional details

See https://drafts.csswg.org/css-backgrounds/#border-width

None that I am aware of.


Psychpsyo:
Actually, https://github.com/mdn/content/pull/43040 from an hour ago does this better than I did...

PR #43046, opened 2026-02-06
Remove some incorrect statements about the Media Capabilities API

wbamberg:
[As far as we can tell](https://github.com/mdn/content/pull/42959#issuecomment-3819637295), the [Media Capabilities API page](https://developer.mozilla.org/en-US/docs/Web/API/Media_Capabilities_API) and [guide](https://developer.mozilla.org/en-US/docs/Web/API/Media_Capabilities_API/Using_the_Media_Capabilities_API) make some claims about the API that are unsupported. This PR removes them.

### DIFF
--- a/files/en-us/web/api/media_capabilities_api/using_the_media_capabilities_api/index.md
+++ b/files/en-us/web/api/media_capabilities_api/using_the_media_capabilities_api/index.md
@@ -7,7 +7,7 @@ browser-compat: api.MediaCapabilities
 
 {{DefaultAPISidebar("Media Capabilities API")}}
 
-The [Media Capabilities API](/en-US/docs/Web/API/Media_Capabilities_API) enabled you to query the browser to determine its ability to encode or decode media given a specified set of encoding parameters. These parameters may include the codecs, resolutions, bit rates, frame rates, and other such details.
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/media_capabilities_api/using_the_media_capabilities_api/index.md
```diff
@@ -7,16 +7,9 @@ browser-compat: api.MediaCapabilities
 
 {{DefaultAPISidebar("Media Capabilities API")}}
 
-The [Media Capabilities API](/en-US/docs/Web/API/Media_Capabilities_API) provides several key features to help you better decide how to handle media, but also to determine how well media is being handled, in real time.
+The [Media Capabilities API](/en-US/docs/Web/API/Media_Capabilities_API) enabled you to query the browser to determine its ability to encode or decode media given a specified set of encoding parameters. These parameters may include the codecs, resolutions, bit rates, frame rates, and other such details.
```
wbamberg:
```suggestion
The [Media Capabilities API](/en-US/docs/Web/API/Media_Capabilities_API) enables you to query the browser to determine its ability to encode or decode media given a specified set of encoding parameters. These parameters may include the codecs, resolutions, bit rates, frame rates, and other such details.
```

hamishwillee (approved):
Excellent thank you.

I was going to comment that there is quite a big caveat on "if playback should be smooth and power efficient." - the method relies on statistics being gathered, so when you first try it the method really only allows checking for codec type (all supported configs report as performant and smooth until after the connection has run for a bit).

But this is corrections, so out of scope.

PR #43050, opened 2026-02-07
Fix WeakMap equivalent emulating private members with private fields …

BrettKulp:
…example missing Thing instantiation

### Description

Add the missing `Thing` instantiation to the example of emulating private members with private fields. Without it, the example on this page https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap throws a ReferenceError when run because thing is not defined

### Motivation

Users copying the example as-is will encounter a runtime error


Josh-Cena (approved):
Nice catch, thank you

PR #43053, opened 2026-02-08
Update index.md

ruchi-mish90:
docs: improve wording in HTML index page


pepelsbey:
Hey! Thank you for the suggestion, but it's not an improvement, more like a stylistic difference.

If you'd really like to help, I would suggest looking at the open issues and starting from there.

PR #43060, opened 2026-02-09
SharedArrayBuffer and Atomic - remove interactive examples

hamishwillee:
The interactive examples in [`Atomics`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics) and  [`SharedArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer) don't work, because they depend on (and make sense for) `SharedArrayBuffer` and that has security concerns that mean you can't run the examples in an interactive example, or even copy them into the console.
The Rari team say that the [interactive examples infrastructure will not be fixed](https://github.com/mdn/fred/issues/971#issuecomment-3750043131), so this PR attempts to remove the examples in a sensible way.

Specifically I have added note about  the security concerns between the Examples and first example in all cases:

> Note that these examples cannot be run directly from the console or an arbitrary web page, because `SharedArrayBuffer` is not defined unless its [security requirements](/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements) are met.

I have removed the interactive examples. However IMO these are better than the examples in that they have a little more documentation alongside explaining what is going on. So where sensible, I have copied information or patterns from the interactive example into the first example in the body.

Lastly, have fixed up a spurious line endings in the first commit.

### DIFF
--- a/files/en-us/web/javascript/reference/global_objects/atomics/add/index.md
+++ b/files/en-us/web/javascript/reference/global_objects/atomics/add/index.md
@@ -7,8 +7,7 @@ browser-compat: javascript.builtins.Atomics.add
 sidebar: jsref
 ---
 
-The **`Atomics.add()`** static method adds a given value at a given position in the array and returns the old value at that position.
-This atomic operation guarantees that no other write happens until the modified value is written back.
+The **`Atomics.add()`** static method adds a given value at a given position in the array, and returns the old value at that position. This atomic operation guarantees that no other write happens until the modified value is written back.
 
 ## Syntax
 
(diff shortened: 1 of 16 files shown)


hamishwillee:
> 👍 ok
> 
> Slightly disappointed with this resolution but I see that it makes sense

Me too. It sux. I don't have any insight into a better way forward though.

EDIT. And I prefer this to all these examples just not working, or working but not copyable or clear why.

PR #43062, opened 2026-02-09
Revert "chore(npm): enable engine-strict, disable audit"

porucznikswext-source:
Reverts mdn/content#42211

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,0 @@
-audit=false
-engine-strict=true


.npmrc
```diff
@@ -1,2 +0,0 @@
-audit=false
```
porucznikswext-source, line 1: true

.npmrc
```diff
@@ -1,2 +0,0 @@
...
-engine-strict=true
```
porucznikswext-source, line 2: true

porucznikswext-source (commented):
### ### ### ### ### ### ### ### ###

.npmrc
porucznikswext-source: true is true

PR #43064, opened 2026-02-09
docs: fix local preview URL casing

pranjal2004838:
## Summary
- Update the local preview example URL in CONTRIBUTING.md to use the correct locale casing (en-US).

## Why
- The previous example used en-us, which does not match MDN locale casing and can lead to 404s when contributors copy/paste the URL.

## Testing
- Not required (docs-only change).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -198,7 +198,7 @@ This will make it easier to submit a pull request for your changes.
 1. When the server is running, make the changes you would like to make to one or more `index.md` files.
 
 2. Open a browser and navigate to the equivalent pages you've changed.
-   If you changed `files/en-us/web/javascript/index.md`, you would navigate to `http://localhost:5042/en-us/docs/web/javascript` in your browser, for example.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -198,7 +198,7 @@ This will make it easier to submit a pull request for your changes.
 1. When the server is running, make the changes you would like to make to one or more `index.md` files.
 
 2. Open a browser and navigate to the equivalent pages you've changed.
-   If you changed `files/en-us/web/javascript/index.md`, you would navigate to `http://localhost:5042/en-us/docs/web/javascript` in your browser, for example.
+   If you changed `files/en-us/web/javascript/index.md`, you would navigate to `http://localhost:5042/en-US/docs/web/javascript` in your browser, for example.
```
Josh-Cena, line 201: @pranjal2004838 @pepelsbey it's still not correct though. It was made that way so people can easily copy the relative file path and past into the browser. But the pedantic URL is  `http://localhost:5042/en-US/docs/Web/JavaScript`. Now we are stuck in a weird place in the middle.

PR #43067, opened 2026-02-09
Document Touch altitudeAngle, azimuthAngle, touchType properties

pepelsbey:
### Description

**New pages**

- `/en-US/docs/Web/API/Touch/altitudeAngle` — The `altitudeAngle` property
- `/en-US/docs/Web/API/Touch/azimuthAngle` — The `azimuthAngle` property
- `/en-US/docs/Web/API/Touch/touchType` — The `touchType` property

**Updated pages**

- `/en-US/docs/Web/API/Touch` — Added a new "Touch type and angle" subsection
- `/en-US/docs/Web/API/PointerEvent/altitudeAngle` — Cross-referenced similar properties
- `/en-US/docs/Web/API/PointerEvent/azimuthAngle` — Cross-referenced similar properties

### Motivation

Document properties that are supported since Safari iOS 10, according to BCD.

### Additional details

- [BCD: Touch.altitudeAngle](https://github.com/mdn/browser-compat-data/blob/main/api/Touch.json#L94)
- [BCD: Touch.azimuthAngle](https://github.com/mdn/browser-compat-data/blob/main/api/Touch.json#L130)
- [BCD: Touch.touchType](https://github.com/mdn/browser-compat-data/blob/main/api/Touch.json#L786)

### DIFF
--- a/files/en-us/web/api/touch/altitudeangle/index.md
+++ b/files/en-us/web/api/touch/altitudeangle/index.md
@@ -17,9 +17,11 @@ For hardware and platforms that do not report tilt or angle, the value is `0`.
 
 ## Value
 
-An angle in radians between `0` and `π/2` where `0` is parallel to the device surface (X-Y plane), and `π/2` is perpendicular to the surface.
+An angle in radians between `0` and `π/2`, where `0` is parallel to the device surface (X-Y plane), and `π/2` is perpendicular to the surface.
 
 ## Example
+
+### Basic usage
 
 ```js
 someElement.addEventListener(
--- a/files/en-us/web/api/touch/azimuthangle/index.md
+++ b/files/en-us/web/api/touch/azimuthangle/index.md
@@ -14,9 +14,11 @@ For hardware and platforms that do not report tilt or angle, the value is `0`. W
 
 ## Value
 
-An angle in radians between `0` and `2π` where `0` represents a transducer whose cap is pointing in the direction of increasing X values (point to "3 o'clock" if looking straight down) on the X-Y plane, and the values progressively increase when going clockwise (`π/2` at "6 o'clock", `π` at "9 o'clock", `3π/2` at "12 o'clock").
+An angle in radians between `0` and `2π`, where `0` represents a transducer whose cap is pointing in the direction of increasing X values (point to "3 o'clock" if looking straight down) on the X-Y plane, and the values progressively increase when going clockwise (`π/2` at "6 o'clock", `π` at "9 o'clock", `3π/2` at "12 o'clock").
 
 ## Example
+
+### Basic usage
 
 ```js
 someElement.addEventListener(
--- a/files/en-us/web/api/touch/touchtype/index.md
+++ b/files/en-us/web/api/touch/touchtype/index.md
@@ -8,7 +8,7 @@ browser-compat: api.Touch.touchType
 
 {{ APIRef("Touch Events") }}
 
-The **`touchType`** read-only property of the {{domxref("Touch")}} interface returns the type of device that triggered the touch, indicating whether it was a direct touch from a finger or from a stylus.
+The **`touchType`** read-only property of the {{domxref("Touch")}} interface returns the type of device that triggered the touch, such as a stylus, or direct touch from a finger.
 
 ## Value
 
@@ -20,6 +20,8 @@ A string from the `TouchType` enumeration. Possible values are:
   - : The touch was made using a stylus or pen device.
 
 ## Example
+
+### Basic usage
 
 ```js
 someElement.addEventListener(


files/en-us/web/api/touch/altitudeangle/index.md
```diff
@@ -0,0 +1,53 @@
...
+The **`altitudeAngle`** read-only property of the {{domxref("Touch")}} interface returns the angle between a transducer (a pointer or stylus) axis and the X-Y plane of a device screen. The altitude angle describes whether the transducer is perpendicular to the screen, parallel, or at some angle in between.
+
+> [!NOTE]
+> The default value of `altitudeAngle` is `0` (parallel to the device surface), which differs from the {{domxref("PointerEvent.altitudeAngle")}} property, which defaults to `π/2` (perpendicular to the surface).
+
+For hardware and platforms that do not report tilt or angle, the value is `0`.
+
+## Value
+
+An angle in radians between `0` and `π/2` where `0` is parallel to the device surface (X-Y plane), and `π/2` is perpendicular to the surface.
```
hamishwillee:
```suggestion
An angle in radians between `0` and `π/2`, where `0` is parallel to the device surface (X-Y plane), and `π/2` is perpendicular to the surface.
```

files/en-us/web/api/touch/altitudeangle/index.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Example
+
```
hamishwillee, line 23: Web API standard suggests level 3 intermediate heading


```suggestion
## Example

### Basic usage

```

files/en-us/web/api/touch/azimuthangle/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+{{ APIRef("Touch Events") }}
+
+The **`azimuthAngle`** read-only property of the {{domxref("Touch")}} interface returns the angle between the Y-Z plane and the plane containing both the transducer (pointer or stylus) axis and the Y axis. This represents the compass direction that the transducer is pointing in relative to the device surface.
+
+For hardware and platforms that do not report tilt or angle, the value is `0`. When the transducer is perpendicular to the surface ({{domxref("Touch.altitudeAngle", "altitudeAngle")}} of `π/2`), the value is `0`.
+
+## Value
+
+An angle in radians between `0` and `2π` where `0` represents a transducer whose cap is pointing in the direction of increasing X values (point to "3 o'clock" if looking straight down) on the X-Y plane, and the values progressively increase when going clockwise (`π/2` at "6 o'clock", `π` at "9 o'clock", `3π/2` at "12 o'clock").
```
hamishwillee:
```suggestion
An angle in radians between `0` and `2π`, where `0` represents a transducer whose cap is pointing in the direction of increasing X values (point to "3 o'clock" if looking straight down) on the X-Y plane, and the values progressively increase when going clockwise (`π/2` at "6 o'clock", `π` at "9 o'clock", `3π/2` at "12 o'clock").
```

files/en-us/web/api/touch/azimuthangle/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Example
+
```
hamishwillee, line 20:
```suggestion
## Example

### Basic usage

```

files/en-us/web/api/touch/touchtype/index.md
```diff
@@ -0,0 +1,54 @@
...
+title: "Touch: touchType property"
+short-title: touchType
+slug: Web/API/Touch/touchType
+page-type: web-api-instance-property
+browser-compat: api.Touch.touchType
+---
+
+{{ APIRef("Touch Events") }}
+
+The **`touchType`** read-only property of the {{domxref("Touch")}} interface returns the type of device that triggered the touch, indicating whether it was a direct touch from a finger or from a stylus.
```
hamishwillee: I'm going to say "such as" because this might extend in future.


```suggestion
The **`touchType`** read-only property of the {{domxref("Touch")}} interface returns the type of device that triggered the touch, such as a stylus, or direct touch from a finger.
```

files/en-us/web/api/touch/touchtype/index.md
```diff
@@ -0,0 +1,54 @@
...
+
+A string from the `TouchType` enumeration. Possible values are:
+
+- `"direct"`
+  - : The touch was made by a direct contact, such as a finger on the screen.
+- `"stylus"`
+  - : The touch was made using a stylus or pen device.
+
+## Example
+
```
hamishwillee, line 23:
```suggestion
## Example

### Basic usage

```

pepelsbey:
This is a part of the fill-the-gaps project where I'm mostly trying to add barebones docs for missing WebKit-only APIs. Not the most popular ones, either. So I'd rather move on to the next one 😉

hamishwillee (approved):
> This is a part of the fill-the-gaps project where I'm mostly trying to add barebones docs for missing WebKit-only APIs. Not the most popular ones, either. So I'd rather move on to the next one 😉

@pepelsbey Just to be pedantic, "being a fill in the gaps job" is not a justification for not doing a great doc. As I understand it the reviewer rule is that you accept updates to existing docs that are just "better" but that new docs should meet all reasonable reviewer standards to be "good".

That said, I wouldn't want to have this requirement added on me because I am poor at diagramming, and I agree it isn't vital. Ditto a live example, though I do really like those because they make it much easier to be sure things work properly - they save a lot of time if you ever have to come back to an interface.

Anyway, a very long way of say "sure". Thanks very much.

pepelsbey:
> "being a fill in the gaps job" is not a justification for not doing a great doc

Fully agree! I think you would also agree that any docs on MDN could be improved almost indefinitely, but we need to focus our effort on what’s more important and the most impactful. I will consider your “sure” as the highest praise 😌

Thank you!

chrisdavidmills:
Just to chime in quickly here, I remember a discussion a little while ago where it was agreed that it would be better to have minimal docs for some of the older, undocumented stuff than no docs at all. We can then experiment to see whether having a minimal doc in existence would make it more likely that volunteers will fill in more details on those pages (rather than expecting them to create a doc from scratch). 

I believe Vadim's work here is part of that initiative.

I also agree that "great docs everywhere" is really what we are after, but time is not on our side here.

hamishwillee:
FWIW My suggestions are based on what I would do if I was assigned this job. I'm lazy, so I don't target "perfect". On the othe rhand Vadim knows what he is doing, which is why I'm fine with this.

I think the minimal docs suggestion for older stuff probably makes sense - it was never going to reach the top of the pile and probably isn't that important anyway.  I am a bit concerned that we start doing this more widely though.

chrisdavidmills:
> I think the minimal docs suggestion for older stuff probably makes sense - it was never going to reach the top of the pile and probably isn't that important anyway. I am a bit concerned that we start doing this more widely though.

You've absolutely hit the nail on the head with this line. Older issues, such as Safari-only gaps, could be left undocumented forever without an initiative like this. At the same time, I am also in full agreement that we shouldn't apply this mindset to more important docs, like new web platform features.

PR #43076, opened 2026-02-10
CSS Reference - Align shorthand logical spacing properties values section with the spec, referencing the correct value.

Buenaventura0:
### Description

Align padding-block, margin-block, and margin-inline values section with the spec, referencing the correct value. 

### Motivation

This change was done in a batch in a previous pull request for margin-related logical properties. Padding-related logical properties use a different structure for values content. So I decided to leave those intact, but change the ones similar to the previous changes.

https://github.com/mdn/content/pull/42907

### Additional details

https://drafts.csswg.org/css-logical/#padding-properties

https://drafts.csswg.org/css-logical/#margin-properties

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/padding-block#values

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/margin-block#values

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/margin-inline#values

Relates to #42907


hamishwillee:
@chrisdavidmills Replacing me with you as reviewer, since you looked at the very similar #42907

Issue #43088, opened 2026-02-10
Remove external links related to paid courses


wpiron10:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*

### What specific section or headline is this issue about?

Description

### What information was incorrect, unhelpful, or incomplete?

In the "Description" 's section of generators function*, there is a link explaning inversion of control concept. It redirects to a paid course from  frontendmasters. Not only it is not useful as we don't get the meaning, but we have to pay the course to know what the subject is about...

Here is the section discussed below (english version) :
> Generators used to be a paradigm for asynchronous programming, avoiding [Callback Hell](https://medium.com/@raihan_tazdid/callback-hell-in-javascript-all-you-need-to-know-296f7f5d3c1) by achieving [Inversion of Control](https://frontendmasters.com/courses/rethinking-async-js/callback-problems-inversion-of-control/). Nowadays, this use case is solved with the simpler [async functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) model and the [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) object. However, generators are still useful for many other tasks, such as defining [iterators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_generators) in a straightforward way.

### What did you expect to see?

A section of description without external link that leads to paid course but instead a link to a clear explanation of the inversion of control.

### Do you have any supporting links, references, or citations?

https://frontendmasters.com/courses/rethinking-async-js/callback-problems-inversion-of-control/

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/javascript/reference/statements/function_star_`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/statements/function_star_/index.md
* Last commit: https://github.com/mdn/content/commit/042b7d8ab67ddb6416da7772a789fd452441b6f6
* Document last modified: 2025-09-06T15:12:50.000Z

</details>

PR #43090, opened 2026-02-10
Replace IoC link with Wikipedia

Josh-Cena:
Fixes https://github.com/mdn/content/issues/43088

We already do so on the "Using promises" page so it makes sense to be consistent.


wbamberg (approved):
👍

PR #43100, opened 2026-02-12
CSS Reference - Align all inset logical positioning related properties to the spec in the values syntax section.

Buenaventura0:
### Description

This PR is part of a batch of related pull requests where spacing and positioning properties were referencing a different value from the spec in the values syntax section. We are changing the reference value progressively whenever I find it, as in the previous pull requests.

### Motivation

Align the docs to the CSS spec for overall consistency, changing the present reference "left" value to "top" as described in the document: https://drafts.csswg.org/css-position-3/#inset-shorthands

### Additional details

https://drafts.csswg.org/css-position-3/#inset-shorthands
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset#values
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-block#values
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-block-start#values
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-block-end#values
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline#values
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline-start#values
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline-end#values

Relates to #43076 and #42907


chrisdavidmills (approved):
I like it; good to see consistency here. Thanks @Buenaventura0.

PR #43103, opened 2026-02-12
update examples for background-attachment 

estelle:
added text explaining examples
updated the examples to ensure something noticeable would happen

### DIFF
--- a/files/en-us/web/css/reference/properties/background-attachment/index.md
+++ b/files/en-us/web/css/reference/properties/background-attachment/index.md
@@ -87,12 +87,12 @@ background-attachment: revert-layer;
 background-attachment: unset;
 ```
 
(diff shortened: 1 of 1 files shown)


estelle:
> Hey! Thank you for updating the examples. But I noticed that there’s more than just CSS here.
> 
> We discussed it in the content team and agreed that changing example content doesn’t add much value (it wasn’t broken), but the opposite: it changes relatively neutral text into heavy and somewhat disturbing one. This change is also missing from the PR description, although it takes up to half of the changed lines.
> 
> I would suggest reverting the old demo text and focusing on the code as the most valuable part of this example.

I changed it to Dr. Seuss, so the user can better see which line they are on as it scrolls

PR #43109, opened 2026-02-12
[WebDriver Classic] Add pages for Back, Forward, Refresh, Get Title, and New Session commands

dipikabh:
### Description

This work is part of the "Fill the gaps" project.

This PR adds documentation for _Back_, _Forward_, _Refresh_, and _Get Title_ commands.

**Note**: While there is an ongoing discussion about documentation priorities (Classic vs BiDi), this PR contains work that's already complete and ready for review. These pages can be reviewed and merged while that discussion is being finalized.

### Note about example code blocks

- _Back_, _Forward_, and _Refresh_ pages: These use Python examples using selenium, to keep them consistent with code blocks on our existing command pages ([example](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/Classic/Commands/CloseWindow#example)).
- _Get Title_ page: For this, I'm trying out example code using curl and HTTP protocol format (based on a suggestion from @whimboo).

I would appreciate feedback to understand which of these two formats is better for our audience.

### Motivation

To improve WebDriver documentation coverage on MDN

### Related links

Spec links:
  - Back: https://w3c.github.io/webdriver/#dfn-back
  - Forward: https://w3c.github.io/webdriver/#dfn-forward
  - Refresh: https://w3c.github.io/webdriver/#dfn-refresh
  - Get Title: https://w3c.github.io/webdriver/#dfn-get-title

### DIFF
--- a/files/en-us/web/webdriver/reference/classic/commands/newsession/index.md
+++ b/files/en-us/web/webdriver/reference/classic/commands/newsession/index.md
@@ -8,6 +8,7 @@ sidebar: webdriver
 ---
 
 The _New Session_ [command](/en-US/docs/Web/WebDriver/Reference/Classic/Commands) of the [WebDriver](/en-US/docs/Web/WebDriver) API creates a new WebDriver session with the browser.
+The session identifier returned in the response is required for all subsequent commands.
 
 ## Syntax
 


files/en-us/web/webdriver/reference/classic/commands/gettitle/index.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Examples
+
+### Retrieving the page title
+
+With a WebDriver server running on `localhost:4444`, assume an active session has navigated to `https://example.org` and you have the `sessionId` from the response. You can retrieve the document title by using the following command:
+
+```bash
+curl -i http://localhost:4444/session/16005d0d-b0dd-4e71-ab38-025329c3baef/title
+```
```
pepelsbey, line 42: It might not be clear that `16005d0d-b0dd-4e71-ab38-025329c3baef` is just an example ID. It would also be easier to replace ID with your specific one, compared to this long string, when you copy-paste the example.

````suggestion
```bash
curl -i http://localhost:4444/session/ID/title
```

Where `ID` is the `sessionId` you got from the response.
````

File: files/en-us/web/webdriver/reference/classic/commands/back/index.md [hunk 2]
```diff
@@ -0,0 +1,89 @@
...
+### Errors
+
+- [Invalid session id](/en-US/docs/Web/WebDriver/Reference/Classic/Errors/InvalidSessionID)
+  - : Session does not exist.
+- No such window
+  - : The [`window`](/en-US/docs/Web/API/Window) object has been discarded, indicating that the tab or window has been closed.
+- Timeout
+  - : The navigation did not complete before its timeout expired.
+
+### Return value
```
whimboo: I wonder if we should list the return value before the errors section. Or is that a common order as well used by other sections?

files/en-us/web/webdriver/reference/classic/commands/back/index.md
```diff
@@ -0,0 +1,89 @@
...
+### Navigating back in browsing history
+
+This example demonstrates navigating from one page to the next and then using the `back` command to return to the first page.
+
+```python
+from selenium import webdriver
+
+session = webdriver.Firefox()
+
+# Navigate to the first page
```
whimboo: I think that we should not use Selenium for the examples given that it hides the actual WebDriver API surface which Selenium itself (as client) implements. We should really demonstrate what payload to send via the HTTP protocol. This as well would not bind us to a single language like Python in this case.

files/en-us/web/webdriver/reference/classic/commands/back/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
+- [Forward command](/en-US/docs/Web/WebDriver/Reference/Classic/Commands/Forward)
```
whimboo, line 65: Should we also add the Navigate and Refresh commands here? Same would apply for the Forward command. Then we would have the same pages referenced as within the Refresh page.

files/en-us/web/webdriver/reference/classic/commands/gettitle/index.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Get Title
+short-title: Get Title
+slug: Web/WebDriver/Reference/Classic/Commands/GetTitle
+page-type: webdriver-command
+browser-compat: webdriver.classic.GetTitle
+sidebar: webdriver
+---
+
+The _Get Title_ [command](/en-US/docs/Web/WebDriver/Reference/Classic/Commands) of the [WebDriver](/en-US/docs/Web/WebDriver) API returns the document title of the current top-level browsing context (window or tab). This is equivalent to reading the {{domxref("Document.title")}} property in JavaScript.
```
whimboo: By default we only operate in content scope which means it returns the title of the top-level browsing context within the tab.

Returning the title of the chrome (browser) window is Mozilla specific and we should potentially cover all that logic outside of MDN?

files/en-us/web/webdriver/reference/classic/commands/back/index.md
```diff
@@ -0,0 +1,89 @@
...
+### URL parameters
+
+- `session id`
+  - : Identifier of the session.
+
+### Errors
+
+- [Invalid session id](/en-US/docs/Web/WebDriver/Reference/Classic/Errors/InvalidSessionID)
+  - : Session does not exist.
+- No such window
```
whimboo, line 31: Maybe we should just define this error given that it would be present for near all the commands. Then we could have a link here as well.

files/en-us/web/webdriver/reference/classic/commands/gettitle/index.md
whimboo: As mentioned as a comment on the WebDriver on MDN document I wonder if we should have at least the `New Session` command implemented. That way we could link to the example of that command in how to create a session. That way we wouldn't have to repeat the same paragraph for each command to add.

files/en-us/web/webdriver/reference/classic/commands/back/index.md
dipikabh: Ah, that's right and good catch! I've moved the "Return value" section before the "Errors" section.
Fixed in https://github.com/mdn/content/pull/43109/commits/c61576dc58eb7e7d5976e184b62a4a664563ebcb

files/en-us/web/webdriver/reference/classic/commands/back/index.md
dipikabh: Yes! Done in https://github.com/mdn/content/pull/43109/commits/c61576dc58eb7e7d5976e184b62a4a664563ebcb

files/en-us/web/webdriver/reference/classic/commands/back/index.md
dipikabh: This is a good point. However, creating a dedicated error page is slightly out of scope for this PR.
I'll try to come back to these later.

files/en-us/web/webdriver/reference/classic/commands/gettitle/index.md
dipikabh: Fixed, removed "(window or tab)"

files/en-us/web/webdriver/reference/classic/commands/back/index.md
dipikabh: I've updated the examples in `back`, `forward`, and `refresh`.

files/en-us/web/webdriver/reference/classic/commands/gettitle/index.md
dipikabh: Accepted both suggestions:
- replaced the session Id value with `ID` in the example.
- added the documentation for "New Session" command and added link to it to get the `sessionId`.

dipikabh:
@pepelsbey @whimboo thank you both for reviewing!

I've addressed the comments and added the New Session command page.
Could you take another look? 🙏

files/en-us/web/webdriver/reference/classic/commands/newsession/index.md
```diff
@@ -0,0 +1,67 @@
+---
+title: New Session
+short-title: New Session
+slug: Web/WebDriver/Reference/Classic/Commands/NewSession
+page-type: webdriver-command
+browser-compat: webdriver.classic.NewSession
+sidebar: webdriver
+---
+
+The _New Session_ [command](/en-US/docs/Web/WebDriver/Reference/Classic/Commands) of the [WebDriver](/en-US/docs/Web/WebDriver) API creates a new WebDriver session with the browser.
```
dipikabh, line 10:
```suggestion
The _New Session_ [command](/en-US/docs/Web/WebDriver/Reference/Classic/Commands) of the [WebDriver](/en-US/docs/Web/WebDriver) API creates a new WebDriver session with the browser.
The session identifier returned in the response is required for all subsequent commands.
```

files/en-us/web/webdriver/reference/classic/commands/back/index.md
whimboo: Sounds fine. Thanks.

files/en-us/web/webdriver/reference/classic/commands/back/index.md
whimboo: Hm, I actually miss the navigate command. Did you miss it here and for the other commands?

files/en-us/web/webdriver/reference/classic/commands/gettitle/index.md
whimboo: Looks good now. Thanks.

files/en-us/web/webdriver/reference/classic/commands/back/index.md
dipikabh: I didn't include the link to the "Navigate To" command since we don't have a page for it yet.
I'm trying to avoid introducing broken links.

We can add it to the "See also" sections once that page is created. Does that work?

whimboo: Ok, that's absolutely fine. Lets ignore it for now.

dipikabh:
> Maybe you can update the commit message a bit so that it is clear that this PR is not about all navigation commands given that the most important `Navigate To` command is not included.

Sure, I've updated the title to clarify that this does not cover all navigation commands

dipikabh:
Ran linters locally on the five new files and all pass.
The CI failure is a known workflow bug and is being fixed in https://github.com/mdn/content/pull/43282.

PR #43110, opened 2026-02-12
Object

### LINKED ISSUE #43095: "Each property name before colons is an identifier (either a name, a number, or a string literal)"
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_objects

Using object initializers

The text contains:

> Each property name before colons is an identifier (either a name, a number, or a string literal)

That is not correct in at least two ways:

1. In the example preceding the text, not every property name is an identifier, only one is.
1. An identifier is not either a name, a number, or a string literal. In this context, the literals are a certain subset of the strings, and those can be written without `""` in this context.

"literal" instead of "identifier" probably would work.

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/javascript/guide/working_with_objects`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_objects
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/guide/working_with_objects/index.md
* Last commit: https://github.com/mdn/content/commit/67bcc0fe90eda964606523670f8688dc1d4a838c
* Document last modified: 2025-12-17T18:54:16.000Z

</details>

Iyeroo:
### Description

Replaced inaccurate wording that described all property names as identifiers.
Clarify that property names in object literals can be an unquoted name
(valid JavaScript identifier), a string literal, a numeric literal,
or a computed property name.

Also simplify the explanation of values to improve clarity for beginners.

### Motivation

The original phrasing conflated different syntactic categories and could cause confusion, especially for readers trying to understand the distinction between identifiers and literals.
Fixes #43095

### DIFF
--- a/files/en-us/web/javascript/guide/working_with_objects/index.md
+++ b/files/en-us/web/javascript/guide/working_with_objects/index.md
@@ -27,7 +27,7 @@ The syntax for an object using an object initializer is:
 const key = "color";
 
 const example = {
-  id: 1, // unqouted name(valid Javascript identifier)
+  id: 2, // unqouted name(valid Javascript identifier)
   "user name": "Alex", // String literal
   100: "numeric key", // Numeric literal
   [key]: "blue", // Computed property name


files/en-us/mozilla/add-ons/webextensions/api/cookies/cookie/index.md
rebloor, line 1: @Iyeroo I don't believe these changes are appropriate. The `cookies.Cookie`object describes details returned in the promise of various methods in the `cookiet` API. Therefore, the returned values reflect the cookie's contents and cannot modify the cookie itself. For example, it is correct that `expirationDate` is "Not provided in for session cookies", its absence doesn't cause a cookie to become a session cookie.
With regard to the wording in `expirationDate`: "Includes milliseconds in the fractional part." This is appropriate given that the milliseconds value is always returned, even if that value is zero.

wbamberg:
Thanks for your PR. I'm closing it because:

- it seems to contain changes that are unrelated to the issue it aims to fix (https://github.com/mdn/content/issues/43095)
- it doesn't fix https://github.com/mdn/content/issues/43095 in the way suggested
- it introduces certain other errors into the pages it changes

PR #43112, opened 2026-02-12
Fix a line of text at the wrong place

ariasuni:
### Description

A part of a sentence was misplaced!! Let them get reunited and makes sense together :heart: 

### Motivation

n/a

### Additional details

n/a

n/a


rebloor (approved):
@ariasuni thanks for spotting and fixing this one

Issue #29326, opened 2023-09-27
Why not use delete and put methods with router object in "Express Tutorial Part 4: Routes and controllers"?


sazk07:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn/Server-side/Express_Nodejs/routes

### What specific section or headline is this issue about?

Create the catalog route module

### What information was incorrect, unhelpful, or incomplete?

the ExpressJS tutorial uses `post` method to delete and update the various resources

```js
// POST request to delete Book.
router.post("/book/:id/delete", book_controller.book_delete_post);

// POST request to update Book.
router.post("/book/:id/update", book_controller.book_update_post);

// POST request to delete Author.
router.post("/author/:id/delete", author_controller.author_delete_post);

// POST request to update Author.
router.post("/author/:id/update", author_controller.author_update_post);

// POST request to delete Genre.
router.post("/genre/:id/delete", genre_controller.genre_delete_post);

// POST request to update Genre.
router.post("/genre/:id/update", genre_controller.genre_update_post);

// POST request to delete BookInstance.
router.post(
  "/bookinstance/:id/delete",
  book_instance_controller.bookinstance_delete_post,
);

// POST request to update BookInstance.
router.post(
  "/bookinstance/:id/update",
  book_instance_controller.bookinstance_update_post,
);
```

### What did you expect to see?

I am just learning web development so this is a question. Wouldn't it be more appropriate/accurate/best practice to use `router.delete(...)` and `router.put(...)` for the delete and update methods?

```js
// DELETE request to delete Book.
router.delete(("/book/:id/delete", book_controller.book_delete_post);

// PUT request to update Book.
router.put("/book/:id/update", book_controller.book_update_post);

// DELETE request to delete Author.
router.delete("/author/:id/delete", author_controller.author_delete_post);

// PUT request to update Author.
router.put("/author/:id/update", author_controller.author_update_post);

// DELETE request to delete Genre.
router.delete("/genre/:id/delete", genre_controller.genre_delete_post);

// PUTT request to update Genre.
router.put("/genre/:id/update", genre_controller.genre_update_post);

// DELETE request to delete BookInstance.
router.delete(
  "/bookinstance/:id/delete",
  book_instance_controller.bookinstance_delete_post,
);

// PUT request to update BookInstance.
router.put(
  "/bookinstance/:id/update",
  book_instance_controller.bookinstance_update_post,
);
```

### Do you have any supporting links, references, or citations?

supporting links:

[...]

PR #43115, opened 2026-02-13
Express: CRUD routes use POST by default

hamishwillee:
Fixes https://github.com/mdn/content/issues/29326

Makes it clear why we use `post` in the Express tutorial for all database update operations, rather than using appropriate HTTP verbs, such as DELETE for deleting records. Essentially, common because it is much harder to do on server rendered sites.

### DIFF
--- a/files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
+++ b/files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
@@ -132,9 +132,9 @@ router.post("/about", (req, res) => {
 ```
 
 Websites should ideally use the route method (and HTTP method) that best corresponds to the operation being performed.
-For example, a client rendered application should use `Router.get()` for reading from the database, `Router.post()` for creating new records, `Router.put()` or `Router.patch()` for updating records, and `Router.delete()` for deleting data.
+For example, a client-rendered application should use `Router.get()` for reading from the database, `Router.post()` for creating new records, `Router.put()` or `Router.patch()` for updating records, and `Router.delete()` for deleting data.
 
-Note however, that Server-rendered applications, such as the one demonstrated by this tutorial, commonly use `Router.post()` for all routes that modify data.
+Note, however, that server-rendered applications, such as the one demonstrated by this tutorial, commonly use `Router.post()` for all routes that modify data.
 The reason for this is that, by default, HTML `<form>` elements are only able to send [`GET`](/en-US/docs/Web/HTTP/Reference/Methods/GET) and [`POST`](/en-US/docs/Web/HTTP/Reference/Methods/POST) requests.
 
 There are various workarounds for this limitation, such as encoding the "desired" HTTP verb in a `POST` request and using the [method-override](https://www.npmjs.com/package/method-override) Express middleware to modify the request to the appropriate HTTP verb before it is passed to the router.


files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
```diff
@@ -685,11 +685,22 @@ module.exports = router;
...
 Routes that act on some specific resource (e.g., book) use path parameters to get the object id from the URL.
 
 The handler functions are all imported from the controller modules we created in the previous section.
 
+> [!NOTE]
+> The routes defined above use the `.post()` method to handle all the requests to create, update, and delete library data.
+> This is common for server-rendered applications, because by default HTML `<form>` elements only send `GET` and `POST` requests.
+>
+> If writing a client-rendered application, which does not need to use `<form>` elements, you should instead use the appropriate method for each operation.
+> For example, your client would send requests to delete objects using the DELETE HTTP method, and your router would specify the `.delete()` method.
```
chrisdavidmills:
```suggestion
> For example, your client would send requests to delete objects using the [`DELETE` HTTP method](/en-US/docs/Web/HTTP/Reference/Methods/DELETE), and your router would specify the `.delete()` method.
```

Not sure if you want to link the HTTP method verbs to their appropriate ref pages? I'm not convinced it's that necessary, but flagging it just in case.

files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
```diff
@@ -685,11 +685,22 @@ module.exports = router;
...
+>
+> It is also possible to use middleware such as [method-override](https://www.npmjs.com/package/method-override) to encode the "desired" HTTP verb in a POST request, which can then be modified to the the desired HTTP verb before the request is passed to the router.
```
chrisdavidmills:
```suggestion
> It is also possible to use middleware such as [method-override](https://www.npmjs.com/package/method-override) to encode the "desired" HTTP verb in a `POST` request, which can then be modified to the desired HTTP verb before the request is passed to the router.
```

files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
```diff
@@ -685,11 +685,22 @@ module.exports = router;
...
+> This may be worthwhile to improve server logging, or if the server needs to handle both server and client-side rendered content through the same endpoint.
```
chrisdavidmills: Maybe also precede this with something like "For basic applications, rewriting your HTTP verbs for the sake of correctness is overkill..."?

files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
hamishwillee: Perhaps not necessary, but a good idea. Updated in your comment.

files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
```diff
@@ -685,11 +685,22 @@ module.exports = router;
...
-The routes are defined either using `.get()` or `.post()` methods on the router object. All the paths are defined using strings (we don't use string patterns or regular expressions).
+The routes are defined either using `.get()` or `.post()` methods on the router object.
+All the paths are defined using strings (we don't use string patterns or regular expressions).
 Routes that act on some specific resource (e.g., book) use path parameters to get the object id from the URL.
 
 The handler functions are all imported from the controller modules we created in the previous section.
 
+> [!NOTE]
+> The routes defined above use the `.post()` method to handle all the requests to create, update, and delete library data.
+> This is common for server-rendered applications, because by default HTML `<form>` elements only send `GET` and `POST` requests.
```
hamishwillee:
```suggestion
> This is common for server-rendered applications, because by default HTML `<form>` elements only send [`GET`](/en-US/docs/Web/HTTP/Reference/Methods/GET) and [`POST`](/en-US/docs/Web/HTTP/Reference/Methods/POST) requests.
```

files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
hamishwillee: I like that.


```suggestion
> For basic applications, rewriting code just to use the correct HTTP verbs is overkill.
> It may be worthwhile to improve server logging, or if the server needs to handle both server and client-side rendered content through the same endpoint.
```

hamishwillee:
> I also wondered whether this addition should be presented as a non-note section with its own subheading, rather than a note, as it is quite a long section to be in a single note. But I'm not going to block on that.

To do this I have had to move the content, and invert it a bit. Essentially a rewrite - worth it because I think it is more prominent and appropriate to talk about this in the HTTP verb section

Thanks!

files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
```diff
@@ -131,6 +131,16 @@ router.post("/about", (req, res) => {
 });
 ```
 
+Websites should ideally use the route method (and HTTP method) that best corresponds to the operation being performed.
+For example, a client rendered application should use `Router.get()` for reading from the database, `Router.post()` for creating new records, `Router.put()` or `Router.patch()` for updating records, and `Router.delete()` for deleting data.
```
chrisdavidmills:
```suggestion
For example, a client-rendered application should use `Router.get()` for reading from the database, `Router.post()` for creating new records, `Router.put()` or `Router.patch()` for updating records, and `Router.delete()` for deleting data.
```

files/en-us/learn_web_development/extensions/server-side/express_nodejs/routes/index.md
```diff
@@ -131,6 +131,16 @@ router.post("/about", (req, res) => {
...
+
+Note however, that Server-rendered applications, such as the one demonstrated by this tutorial, commonly use `Router.post()` for all routes that modify data.
```
chrisdavidmills:
```suggestion
Note, however, that server-rendered applications, such as the one demonstrated by this tutorial, commonly use `Router.post()` for all routes that modify data.
```

hamishwillee:
Thanks for the reviews @chrisdavidmills !

PR #43124, opened 2026-02-14
Update index.md

byjadams:
### Description

Changed the word "simultaneously" to "in one declaration" in the overview.

Removed the extra examples of a shorthand.

Added the shorthand properties that use the top, right, bottom, and right motions so they are all in one place.

Added the special considerations for `border` and `outline`.

Changed the words from "edge" to "side" when referring to the box model.

Removed the words "horizontal" and "vertical".

Hyphenated the top-left, top-right, etc.

### Motivation

Simplifies and clarifies the language so it may be translated easier.

One example is enough. The rest of the article is how to use and what can be used.

Since the clockwise motion is special, all the shorthands are listed at the bottom of the page, and the properties mostly follow the box model it was simple enough to list which ones are part of a similar theme.

Someone could assume that since `margin` and `padding` shorthands use the clockwise motion, that `border` and `outline` would follow. However they have special considerations that should be mentioned.

Although it was funny word play, "edge cases" and "edge of the box", the common usage would be side of a box.

Again, simplifying the language.

Proper grammar. It isn't "top left", rather "top-left". Minor point.

### DIFF
--- a/files/en-us/web/css/guides/cascade/shorthand_properties/index.md
+++ b/files/en-us/web/css/guides/cascade/shorthand_properties/index.md
@@ -5,15 +5,50 @@ page-type: guide
 sidebar: cssref
 ---
 
-**_Shorthand properties_** are CSS properties that let you set the values of multiple other CSS properties simultaneously. Using a shorthand property, you can write more concise (and often more readable) style sheets, saving time and energy.
+**_Shorthand properties_** are CSS properties that let you set the values of multiple other CSS properties in one declaration. Using a shorthand property, you can write more concise (and often more readable) style sheets, saving time and energy.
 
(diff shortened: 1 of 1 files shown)


pepelsbey:
Hello! Thank you for trying to help.

Unfortunately, in this case I don't think it's something we could accept. We don't just rewrite whole guides because we feel like it. Instead, we try to approach it iteratively: first stating the problem in an issue and then solving it one step at a time.

If you really like to help, I would suggest starting from existing issues ([Help wanted label](https://github.com/mdn/content/issues?q=state%3Aopen%20label%3A%22help%20wanted%22)) to get more experience contributing. Or if you have a strong opinion about this guide, please file an issue, and we can discuss it. Feel free to tag me there if you like.

Thank you again, and I hope to see your contributions to MDN!

PR #43127, opened 2026-02-14
Change “simultaneously” to “in one declaration”

byjadams:
Changed the word "simultaneously" to "in one declaration" to make it simplify the language and make it easier for translation.

### Description

Changed the word "simultaneously" to "in one declaration".

### Motivation

Simplify the language and make it easier for translation.


byjadams:
This makes it logically consistent with the rest of the page that says, "...shortened to just one" or "... equivalent to the longhand declarations..."

PR #43128, opened 2026-02-14
Update index.md

byjadams:
Changed "for instance" to "for example" to make it easier to translate and also removed extra examples. One is enough to get the point across.

### Description

Changed "for instance" to "for example".

Removed extra examples.

### Motivation

Simplify the language to make it easier to translate.

One example is enough to make the point of what a shorthand property does.


hamishwillee (approved):
Thanks @byjadams - and welcome to MDN.

FYI, normally MDN will  reject this kind of PR. Yes, it is an improvement, but the replaced text is neither incorrect or ambiguous. Since we have a limited reviewer pool we simply don't have the resource for grammatical improvements unless they accompany errors or new technical content.

PR #43138, opened 2026-02-16
Express: Updates to mongoose

hamishwillee:
I got a note from MongoDb hosting that we need to update the driver. This appears to be incorrect (based on the driver mentioned) but nevertheless am updating to the latest version of mongoose to make sure we are up to date.

In code, main change is to remove mongoose.set("strictQuery", false); which is set by default in current driver. Also did some minor tidy suggested by AI, which makes sense.

This updates the docs. Code changes in https://github.com/mdn/express-locallibrary-tutorial/pull/344


chrisdavidmills (approved):
LGTM; thanks @hamishwillee

PR #43141, opened 2026-02-16
Update CPG email

couci:
Updating the email for the MDN CPG form. In the text there is a reference for mdn-cpg-report@mozilla.com but such an email doesn't exist. I am proposign to send to the regular cpg-report instead

### Description

Updated the email for reporting cpg cases in MDN

### Motivation

The current email doesn't work. A user reported this on SUMO forums

### DIFF
--- a/files/en-us/mdn/community/community_participation_guidelines/index.md
+++ b/files/en-us/mdn/community/community_participation_guidelines/index.md
@@ -70,7 +70,7 @@ Below is a summary of severity levels and enforcement actions.
 
 ### Reporting violations
 
-Reports can be sent to mdn-cpg-report@mozilla.com.
+Reports can be sent to cpg-report@mozilla.com.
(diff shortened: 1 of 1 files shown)


wsmwk:
localized pages are also affected, such as 

- https://developer.mozilla.org/de/docs/MDN/Community/Community_Participation_Guidelines 
- https://developer.mozilla.org/ja/docs/MDN/Community/Community_Participation_Guidelines
- https://developer.mozilla.org/fr/docs/MDN/Community/Community_Participation_Guidelines

pransh15:
Thanks for catching this and opening the PR, @couci. There was indeed an issue with the group delivery and moderation settings on the original address, which I’ve now updated, this should resolve the delivery problem going forward. I appreciate you flagging it, and referencing the SUMO report.

Thanks for checking localized pages, @wsmwk. After testing, I can safely say that the email should work now, and is not an issue anymore.

PR #43146, opened 2026-02-16
Change event listener from `onchange` to `oninput`

artlung:
### Description

Switching from the `change` event to the `input` event allows the visitor to see color contrast changes in real time as they edit the color.

### Motivation

More immediate visual feedback to the user!

### DIFF
--- a/files/en-us/web/css/reference/values/color_value/contrast-color/index.md
+++ b/files/en-us/web/css/reference/values/color_value/contrast-color/index.md
@@ -37,7 +37,7 @@ The `contrast-color()` function returns a value of `white` or `black`, depending
 
 ### Contrasting text for a button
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/reference/values/color_value/contrast-color/index.md
hamishwillee:
```suggestion
In the following example, the browser automatically applies the selected color to the background of the "Button" {{htmlelement("button")}}, and the contrasting color (black or white) to its text.
```

hamishwillee:
```suggestion
<button>Button</button>
```

hamishwillee (approved):
Thanks @artlung - that is better. I made some minor modifications because it was a bit confusing to name it submit when it wasn't used as an update trigger for the color, and because the fact that the text could only turn black/white wasn't clear just at this point.

PR #43148, opened 2026-02-17
Clarify finally block control-flow timing in try...catch

### LINKED ISSUE #43122: Clarify execution order of control flow statements with finally
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

The finally block

The third bullet point states: “Immediately before the execution of a control-flow statement (return, throw, break, continue) in the try block or catch block that would exit the block.” However, if control flow reaches a return statement in the try block, that statement is still evaluated (including any side effects it may produce). The finally block runs afterward, but it does not prevent the return statement itself from being evaluated. Only if the finally block contains its own return statement does it override the previous one, causing the function to return the value associated with the return statement in finally.

### The finally block

The `finally` block contains statements to execute after the `try` block and `catch` block(s) execute, but before the statements following the `try...catch...finally` block. Control flow will always enter the `finally` block, which can proceed in one of the following ways:

- Immediately after the control flow exits the `try` block in a `try...finally` construct (e.g., after the last statement or a `throw` or `return` statement);
- Immediately after the control flow exits the `catch` block in a `try...catch...finally` construct;
- Immediately after the control flow exits the `try` block in a `try...catch...finally` construct, unless it exits via a `throw` statement (in which case control flow enters the `catch` block first);

The following example shows one use case for the `finally` block. The code opens a file and then executes statements that use the file; the `finally` block makes sure the file always closes after it is used even if an exception was thrown.

```js
openMyFile();
try {
  // tie up a resource
  writeMyFile(theData);
} finally {
  closeMyFile(); // always close the resource
  // any uncaught exception is propagated here
}
```

If control flow reaches a `return` statement in the `try` block and no `return` statement is present in the `finally` block, all statements in the `finally` block are still executed before the function returns the value associated with this statement.

```js
function doIt() {
  try {
    return 1; // evaluated first
  } finally {
    console.log("enter finally"); // executed after
    // the function now returns 1
  }
}
doIt(); // returns 1
```

It is generally a bad idea to use control flow statements (return, throw, break, continue) in the `finally` block. The `finally` block always runs after the `try` and `catch` blocks, even if they exit because of a control flow statement. If `finally` itself contains a control flow statement, it takes precedence over the one from `try` or `catch`. This means it can change a returned value or replace a thrown exception. For this reason, the `finally` block should generally be reserved for cleanup code.

The following examples illustrate how this happens in practice.

If control flow reaches two `return` statements (possibly in the `try` and `finally` blocks or in the `catch` and `finally` blocks), both statements are evaluated in the correct order (which matters in the case of side effects). Only the function's returned value will be the one associated with the last `return` statement evaluated (which is the one in the `finally` block).

```js
function doIt() {
  let order = ['z'];
  try {
    order.push('try');
    return order.sort(); // 'z' is now after 'try'
  } finally {
    order.push('finally');
    return order; // overrides previously returned value
  }
}
doIt(); // returns ['try', 'z', 'finally']
```

The same applies to other control flow statements (`throw`, `break`, `continue`).

```js
function doIt() {
  try {
    throw 'try'; // makes control flow enter the `catch` block
  } catch {
    throw 'catch'; // makes control flow enter the `finally` block
  } finally {
    throw 'finally'; // changes the thrown value from 'catch' to 'finally'
  }
}
doIt(); // throws 'finally'
```

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/javascript/reference/statements/try...catch`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/statements/try...catch/index.md

</details>

it-education-md:
### Description

This PR makes a small wording clarification in the **“The finally block”** section of the `try...catch` reference to better reflect control-flow timing.

It updates one bullet in `files/en-us/web/javascript/reference/statements/try...catch/index.md`.

* From: “Immediately before the execution of a control-flow statement (`return`, `throw`, `break`, `continue`) …”
* To: “When control flow exits the `try` block or `catch` block through a control-flow statement (`return`, `throw`, `break`, `continue`).”

### Motivation

The previous wording can read as if `finally` runs *before* the control-flow statement executes.

For `return`, the expression is evaluated first (including side effects), and then `finally` runs before completion leaves the `try...catch...finally` construct.

This tweak keeps the explanation accurate while staying concise.

### Additional details

- No API or behavior changes.
- No example updates needed.

Fixes #43122

### DIFF
--- a/files/en-us/web/javascript/reference/statements/try...catch/index.md
+++ b/files/en-us/web/javascript/reference/statements/try...catch/index.md
@@ -122,7 +122,7 @@ The `finally` block contains statements to execute after the `try` block and `ca
 
 - Immediately after the `try` block finishes execution normally (and no exceptions were thrown);
 - Immediately after the `catch` block finishes execution normally;
-- Immediately before the execution of a control-flow statement (`return`, `throw`, `break`, `continue`) in the `try` block or `catch` block that would exit the block.
+- When a control-flow statement (`return`, `throw`, `break`, or `continue`) in the `try` or `catch` block would transfer control out of the surrounding `try...finally` or `try...catch...finally` statement.
 
 If an exception is thrown from the `try` block, even when there's no `catch` block to handle the exception, the `finally` block still executes, in which case the exception is still thrown immediately after the `finally` block finishes executing.
 


files/en-us/web/javascript/reference/statements/try...catch/index.md
```diff
@@ -122,7 +122,7 @@ The `finally` block contains statements to execute after the `try` block and `ca
 
 - Immediately after the `try` block finishes execution normally (and no exceptions were thrown);
 - Immediately after the `catch` block finishes execution normally;
-- Immediately before the execution of a control-flow statement (`return`, `throw`, `break`, `continue`) in the `try` block or `catch` block that would exit the block.
+- When control flow exits the `try` block or `catch` block through a control-flow statement (`return`, `throw`, `break`, `continue`).
```
s-egea: Removing the "that would exit the block" and using "`try` block or `catch` block through" makes the sentence inaccurate since exiting the `try` block through a `throw` statement would result in entering the `catch` block

s-egea: What about using "Through the execution of a control-flow statement (`return`, `throw`, `break`, `continue`) that would exit the whole block (in the `try` or `catch` block)." ?

it-education-md: @s-egea Thanks, this direction makes sense, but I think this phrasing is still a bit ambiguous (`through the execution / whole block`). I think it’s clearer to say:

When a control-flow statement (`return, throw, break, or continue`) in the try or catch block would transfer control out of the surrounding `try...finally` or `try...catch...finally` statement.

s-egea: Yes I like it ! (maybe just replacing the last "statement" by "block") :)

it-education-md: @s-egea Great but `try...finally` and `try...catch...finally` are `JavaScript` statements; using block can be confusing because try and catch already have their own blocks.

s-egea: My bad I am not really familiar with the terminology ! :)

Josh-Cena:
Thanks for your PR! I will go ahead with https://github.com/mdn/content/pull/43152's version because I do think there's some merit in rewriting the logic.

PR #43152, opened 2026-02-17
rewrite finally section

### LINKED ISSUE #43122: Clarify execution order of control flow statements with finally

s-egea:
### Description

Re-write `The finally block` section in the `try...catch` statement documentation file trying to clarify control flow statements execution order while also keeping original structure and examples.

### Motivation

The main point to clarify is the combination of the 3rd bullet points and the following sentence:
- "Immediately before the execution of a control-flow statement (`return`, `throw`, `break`, `continue`) in the `try` block or `catch` block that would exit the block."
- "If an exception is thrown from the `try` block, even when there's no `catch` block to handle the exception, the `finally` block still executes, in which case the exception is still thrown immediately after the `finally` block finishes executing."

The first part says that the `finally` block is entered **before** the execution of the control flow statement that would exit the `try` or `catch` block. The second implies that the `finally` block still executes **after** an exception is thrown (or the execution of a `throw` statement).

The proposed re-writing try to fix this inconsistency by adopting a point of view in which:
- The `finally` block is always entered **after** the control flow statement that would exit the `try` or `catch` block
- The effect of this control flow statement is deferred at the end of the `finally` block unless the `finally` block also ends with a control flow statement.

Fixes #43122

### DIFF
--- a/files/en-us/web/javascript/reference/statements/try...catch/index.md
+++ b/files/en-us/web/javascript/reference/statements/try...catch/index.md
@@ -120,73 +120,13 @@ function isValidJSON(text) {
 
 The `finally` block contains statements to execute after the `try` block and `catch` block(s) execute, but before the statements following the `try...catch...finally` block. Control flow will always enter the `finally` block, which can proceed in one of the following ways:
 
-- Immediately after the control flow exits the `try` block in a `try...finally` construct (e.g., after the last statement or a `throw`, `return`, `break`, or `continue` statement);
+- Immediately after the control flow exits the `try` block in a `try...finally` construct (either after the last statement or a `throw`, `return`, `break`, or `continue` statement);
 - Immediately after the control flow exits the `catch` block in a `try...catch...finally` construct;
 - Immediately after the control flow exits the `try` block in a `try...catch...finally` construct, unless it exits via a `throw` statement (in which case control flow enters the `catch` block first).
 
-If the `finally` block is entered after a control flow statement (`return`, `throw`, `break`, `continue`), the effect of this statement is deferred until after the last statement executed in the `finally` block. For example, if an exception is thrown from the `try` block, even when there's no `catch` block to handle the exception, the `finally` block still executes, and the exception is thrown immediately after the `finally` block finishes executing.
+If the `finally` block is entered after a control flow statement (`return`, `throw`, `break`, `continue`) in the `try` or `catch` block, the effect of this statement is deferred until after the last statement executed in the `finally` block. For example, if an exception is thrown from the `try` block, even when there's no `catch` block to handle the exception, the `finally` block still executes, and the exception is thrown immediately after the `finally` block finishes executing.
 
-However, there is one exception (discussed in more detail below) to this rule: if the last statement executed in the `finally` block is itself a control flow statement, that statement will override the effect of the previous one (no deferral).
-
-The following example shows one use case for the `finally` block. The code opens a file and then executes statements that use the file; the `finally` block makes sure the file always closes after it is used even if an exception was thrown.
-
-```js
-openMyFile();
-try {
-  // tie up a resource
-  writeMyFile(theData);
-} finally {
-  closeMyFile(); // always close the resource
-  // any uncaught exception is deferred here
-}
-```
-
-In the same way, the effect of any `return` statement in the `try` block is deferred at the end of the `finally` block, although the return value expression is evaluated before entering the `finally` block.
-
-```js
-function safeWriteMyFile() {
-  openMyFile();
-  try {
-    return writeMyFile(theData); // function call is evaluated
-  } finally {
-    closeMyFile(); // always close the resource
-    // return is deferred here
-  }
-}
-```
-
-It is generally a bad idea to use control flow statements (`return`, `throw`, `break`, `continue`) in the `finally` block because they can override the effect of previously executed control flow statements, which is rarely intended. Most of the time, the `finally` block should be reserved for cleanup code.
-
-The following example illustrates how a `return` statement in the `finally` block can override the effect of an earlier `return` statement. When control flow exits the `try` block via the first `return` statement, the return value expression (`order.sort()`) is evaluated before entering the `finally` block, and the function is planned to return that value after the `finally` block finishes executing. Because the `finally` block also exits via a `return` statement, the function returns the value evaluated in this last statement instead.
-
-```js
-function doIt() {
-  let order = ["z"];
-  try {
-    order.push("try");
-    return order.sort(); // "z" is now after "try"
-  } finally {
-    order.push("finally");
-    return order; // overrides previously returned value
-  }
-}
-doIt(); // returns ["try", "z", "finally"]
-```
-
-The same logic applies to other control flow statements. Here, the function is first planned to throw the value `"catch"` but instead returns the value `"finally"`.
-
-```js
-function doIt() {
-  try {
-    throw "try"; // makes control flow enter the `catch` block
-  } catch {
-    throw "catch"; // makes control flow enter the `finally` block
-  } finally {
-    return "finally"; // returns "finally" instead of throwing "catch"
-  }
-}
-doIt(); // returns "finally"
-```
+However, there is one exception to this rule: if the last statement executed in the `finally` block is itself a control flow statement, that statement will override the effect of the previous one (no deferral); see [returning from a `finally` block](#returning_from_a_finally_block) for examples. It is generally a bad idea to use control flow statements (`return`, `throw`, `break`, `continue`) in the `finally` block because they can override the effect of previously executed control flow statements, which is rarely intended. Most of the time, the `finally` block should be reserved for cleanup code that does not modify the main logic.
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


files/en-us/web/javascript/reference/statements/try...catch/index.md
```diff
@@ -120,11 +120,11 @@ function isValidJSON(text) {
 
 The `finally` block contains statements to execute after the `try` block and `catch` block(s) execute, but before the statements following the `try...catch...finally` block. Control flow will always enter the `finally` block, which can proceed in one of the following ways:
 
-- Immediately after the `try` block finishes execution normally (and no exceptions were thrown);
-- Immediately after the `catch` block finishes execution normally;
-- Immediately before the execution of a control-flow statement (`return`, `throw`, `break`, `continue`) in the `try` block or `catch` block that would exit the block.
+- Immediately after the control flow exits the `try` block in a `try...finally` construct (e.g., after the last statement or a `throw` or `return` statement);
```
Josh-Cena:
```suggestion
- Immediately after the control flow exits the `try` block in a `try...finally` construct (e.g., after the last statement or a `throw`, `return`, `break`, or `continue` statement);
```

files/en-us/web/javascript/reference/statements/try...catch/index.md
```diff
@@ -120,11 +120,11 @@ function isValidJSON(text) {
...
+- Immediately after the control flow exits the `catch` block in a `try...catch...finally` construct;
+- Immediately after the control flow exits the `try` block in a `try...catch...finally` construct, unless it exits via a `throw` statement (in which case control flow enters the `catch` block first);
```
Josh-Cena:
```suggestion
- Immediately after the control flow exits the `try` block in a `try...catch...finally` construct, unless it exits via a `throw` statement (in which case control flow enters the `catch` block first).
```

files/en-us/web/javascript/reference/statements/try...catch/index.md
```diff
@@ -120,11 +120,11 @@ function isValidJSON(text) {
...
 
-If an exception is thrown from the `try` block, even when there's no `catch` block to handle the exception, the `finally` block still executes, in which case the exception is still thrown immediately after the `finally` block finishes executing.
+In most cases, if the `finally` block is entered after a control flow statement (`return`, `throw`, `break`, `continue`), the effect of this statement is deferred until after the last statement executed in the `finally` block. However, if the last statement executed in the `finally` block is itself a control flow statement, that statement will override the effect of the previous one (no deferral).
```
Josh-Cena:
```suggestion
If the `finally` block is entered after a control flow statement (`return`, `throw`, `break`, `continue`), the effect of this statement is deferred until after the last statement executed in the `finally` block. For example, if an exception is thrown from the `try` block, even when there's no `catch` block to handle the exception, the `finally` block still executes, in which case the exception is still thrown immediately after the `finally` block finishes executing and no other control flow statements were encountered.

However, if the last statement executed in the `finally` block is itself a control flow statement, that statement will override the effect of the previous one (no deferral).
```

files/en-us/web/javascript/reference/statements/try...catch/index.md
```diff
@@ -135,24 +135,58 @@ try {
   writeMyFile(theData);
 } finally {
   closeMyFile(); // always close the resource
+  // any uncaught exception is deferred here
 }
 ```
 
-Control flow statements (`return`, `throw`, `break`, `continue`) in the `finally` block will "mask" any completion value of the `try` block or `catch` block. In this example, the `try` block tries to return 1, but before returning, the control flow is yielded to the `finally` block first, so the `finally` block's return value is returned instead.
+In the same way, any `return` statement in the `try` block is deferred at the end of the `finally` block.
```
Josh-Cena:
```suggestion
In the same way, any `return` statement in the `try` block is deferred at the end of the `finally` block, although the return value expression is evaluated before entering the `finally` block.
```

files/en-us/web/javascript/reference/statements/try...catch/index.md
```diff
@@ -135,24 +135,58 @@ try {
...
+    return writeMyFile(theData); // function call is evaluated
   } finally {
-    return 2;
+    closeMyFile(); // always close the resource
+    // return is deferred here
   }
 }
+```
+
+It is generally a bad idea to use control flow statements (`return`, `throw`, `break`, `continue`) in the `finally` block, as it makes the code harder to read. Most of the time, the `finally` block should be reserved for cleanup code.
```
Josh-Cena:
```suggestion
It is generally a bad idea to use control flow statements (`return`, `throw`, `break`, `continue`) in the `finally` block, because they will "mask" the original control flow effect, which is rarely intended. Most of the time, the `finally` block should be reserved for cleanup code.
```

files/en-us/web/javascript/reference/statements/try...catch/index.md
```diff
@@ -135,24 +135,58 @@ try {
...
+
+The following examples illustrate this.
+
+If control flow reaches two `return` statements (possibly in the `try` and `finally` blocks or in the `catch` and `finally` blocks), both statements are executed in the correct order (which matters in the case of side effects). Only the function's returned value will be the one associated with the last `return` statement executed (which is the one in the `finally` block).
```
Josh-Cena:
```suggestion
If control flow exits the `try` or `catch` block via a `return` statement, the return value is evaluated (`order.sort()` in the example below). However, the `return` _statement_'s execution is deferred until the `finally` block finishes executing. Because the `finally` block also executes a `return` statement, the latter statement causes the first `return` to never execute.
```

Josh-Cena (commented):
Thanks, I like this. Just some suggestions to make things more explicit.

files/en-us/web/javascript/reference/statements/try...catch/index.md
s-egea: Done

files/en-us/web/javascript/reference/statements/try...catch/index.md
s-egea: Done

files/en-us/web/javascript/reference/statements/try...catch/index.md
s-egea: Added the canonical example with a few changes
- removed "and no other control flow statements were encountered" as I find it difficult to understand this before the general rule has been described (case where the `finally` ends with a control flow statement)
- changed "in which case the exception is still thrown" to "and the exception is thrown"

files/en-us/web/javascript/reference/statements/try...catch/index.md
s-egea: Applied. Also used "the effect of any `return` statement" for consistent terminology

files/en-us/web/javascript/reference/statements/try...catch/index.md
s-egea: Applied. 
- Kept "which is rarely intended"
- Changed "mask" to "override" for consistent terminology
- Used "the effect of any previously executed control flow statement" for consistent terminology

files/en-us/web/javascript/reference/statements/try...catch/index.md
s-egea: Applied
- Kept the idea of describing the example in more detail
- Introduced "the function is planned to return..." terminology

s-egea:
@Josh-Cena as far as I am concerned, the PR is ready for a 2nd review. I tried to keep the core ideas behind your suggestions while also maintaining wording consistency.

PR #43154, opened 2026-02-18
Fix GainMode's channelInterpretation linking to channelCountMode.

DDR0:
### Description

Looks like the channel count mode xref was copied and not updated for the following channel interpretation xref on the [GainMode](https://developer.mozilla.org/en-US/docs/Web/API/GainNode/GainNode) page.

### Motivation

Fix mis-targeted link.


chrisdavidmills (approved):
Good spot; thanks for the fix, @DDR0!

PR #43155, opened 2026-02-18
CSS property: scroll-initial-target

estelle:
Added the missing CSS property.

part of https://github.com/openwebdocs/project/issues/234

also added the word "module" to the module links and put links in order by date.

### DIFF
--- a/files/en-us/web/css/reference/properties/scroll-initial-target/index.md
+++ b/files/en-us/web/css/reference/properties/scroll-initial-target/index.md
@@ -190,7 +190,8 @@ p {
 {{EmbedLiveSample("Using scroll-initial-target", "100%", "500")}}
 
 The property's effect is demonstrated when the scroll-snap container is drawn to the page.
-Each row snaps to the first element with `nearest` set, in tree order.
+
+Each row snaps to the first element with `nearest` set, if any, in tree order. In the last example, we've overridden the `nearest` value with `none` on the first element, so the first element with `nearest` applied is #3.
 
 ## Specifications
 


files/en-us/web/css/reference/properties/scroll-initial-target/index.md
```diff
@@ -0,0 +1,204 @@
+---
+title: scroll-initial-target
+slug: Web/CSS/Reference/Properties/scroll-initial-target
+page-type: css-property
+browser-compat: css.properties.scroll-initial-target
+sidebar: cssref
+---
+
+The **`scroll-initial-target`** [CSS](/en-US/docs/Web/CSS) property enables defining elements that should be snapped when scroll containers are first rendered.
```
hamishwillee: "enables defining elements that" feels awkward. My preference is "allows you to define elements" but assuming  there is a pattern for these introductions how about.

EDITED. Reading down ...

```suggestion
The **`scroll-initial-target`** [CSS](/en-US/docs/Web/CSS) property enables the definition of elements that are potential snap targets when their ancestor scroll containers is first rendered.
```

hamishwillee: Also, IFF there is a glossary or other good referenc of "snapped" I'd link it.

files/en-us/web/css/reference/properties/scroll-initial-target/index.md
```diff
@@ -0,0 +1,204 @@
...
+### Values
+
+- `none`
+  - : The element is not an initial scroll target.
+- `nearest`
+  - : The element is potentially an initial scroll target for its nearest scroll container ancestor.
+
+## Description
+
+The `scroll-initial-target` property enables defining elements that should be snapped when scroll snap containers are first rendered. Setting the value to `nearest` defines the element as a potential target that should be snapped when the nearest ancestor {{glossary("scroll container")}} what that container first shows up on the page.
```
hamishwillee: As above for first sentence.

hamishwillee: Second sentence is broken - see "what that container".

What I think this means is (something like) below. The nuance I think you are trying to show here is that the value can be inherited and the snap target might not therefore be the immediate parent. Setting nearest means that the immediate parent is a potential target.


```suggestion
The `scroll-initial-target` property allows you to indicate elements that are potential snap targets when their parent scroll snap containers are first rendered. Setting the value to `nearest` defines the element as a potential snap target when the nearest ancestor {{glossary("scroll container")}} shows up on the page.
```

files/en-us/web/css/reference/properties/scroll-initial-target/index.md
```diff
@@ -0,0 +1,204 @@
...
+
+If multiple elements or pseudo-elements in the scroll container are set to `nearest`, the first element in the tree order is the initial scroll-snapping target.
+
+The initial value is `none`, meaning a scroll-snappable element is not by default an initial scroll target. The `none` value can also be set on an element to explicitly make it not be an initial scroll target. When a scroll snap container contains no descendant scroll snap elements set to be the scroll container's initial scroll target,
```
hamishwillee: The second sentence is dangling and doesn't make sense: When a scroll snap container contains no descendant scroll snap elements set to be the scroll container's initial scroll target,

files/en-us/web/css/reference/properties/scroll-initial-target/index.md
```diff
@@ -0,0 +1,204 @@
...
+
+If a scroll container's initial scroll position is potentially set by both the {{cssxref("place-content")}} content-distribution property and by `scroll-initial-target` on any descendants, the first descendant with `scroll-initial-target: nearest` wins.
```
hamishwillee: Comma is not quite right punctuation. I like emdash. 

```suggestion
If a scroll container's initial scroll position is potentially set by both the {{cssxref("place-content")}} content-distribution property and by `scroll-initial-target` on any descendants — the first descendant with `scroll-initial-target: nearest` wins.
```

files/en-us/web/css/reference/properties/scroll-initial-target/index.md
```diff
@@ -0,0 +1,204 @@
...
+    padding: 1em;
+  }
+}
+```
+
+#### Result
+
+{{EmbedLiveSample("Using scroll-initial-target", "100%", "500")}}
+
+The property's effect is demonstrated when the scroll-snap container is drawn to the page. When this example is loaded, the initially targeted elements are 1, 4, 2, 1, 3.
```
hamishwillee: This is a really good example, which for some reason took me agest to work out "what I was seeing". I think you can make the "first nearest selected in tree order more obvious by explicitly listing the selections with nearest on the example."

So first move this above the rendering.

```suggestion
The property's effect is demonstrated when the scroll-snap container is drawn to the page.
Each row snaps to the first element with `nearest` set, in tree order.
```

Then update notes inline

```
none on #4 only > none on #4 only (nearest on 1,2,3,5)
nearest on #4 only > nearest on #4 only
nearest on even elements > nearest on even elements (2, 4)
etc
```

If you do that it makes it clear why each element is snapped in place. Up to you.

files/en-us/web/css/reference/properties/scroll-initial-target/index.md
```diff
@@ -0,0 +1,204 @@
...
+- `nearest`
+  - : The element is potentially an initial scroll target for its nearest scroll container ancestor.
+
+## Description
+
+The `scroll-initial-target` property enables defining elements that should be snapped when {{glossary("scroll snap")}} containers are first rendered. Setting the value to `nearest` defines the element as a potential target that should be snapped when the nearest ancestor {{glossary("scroll container")}} what that container first shows up on the page.
+
+If multiple elements or pseudo-elements in the scroll container are set to `nearest`, the first element in the tree order is the initial scroll-snapping target.
+
+The initial value is `none`, meaning a scroll-snappable element is not by default an initial scroll target. The `none` value can also be set on an element to explicitly make it not be an initial scroll target. When a scroll snap container contains no descendant scroll snap elements set to be the scroll container's initial scroll target,
```
estelle:
```suggestion
The initial value is `none`, meaning a scroll-snappable element is not by default an initial scroll target. The `none` value can also be set on an element to explicitly make it not be an initial scroll target.
```

files/en-us/web/css/reference/properties/scroll-initial-target/index.md
estelle: removed

files/en-us/web/css/reference/properties/scroll-initial-target/index.md
```diff
@@ -0,0 +1,205 @@
...
+### Values
+
+- `none`
+  - : The element is not an initial scroll target.
+- `nearest`
+  - : The element is potentially an initial scroll target for its nearest scroll container ancestor.
+
+## Description
+
+The `scroll-initial-target` property enables defining elements that should be snapped when {{glossary("scroll snap")}} containers are first rendered. Setting the value to `nearest` defines the element as a potential target that should be snapped when the nearest ancestor {{glossary("scroll container")}} what that container first shows up on the page.
```
estelle:
```suggestion
The `scroll-initial-target` property enables defining elements that should be snapped when their parent {{glossary("scroll snap")}} containers are first rendered. Setting the value to `nearest` defines the element as a potential target that should be snapped to when the nearest ancestor {{glossary("scroll container")}} first shows up on the page.
```

PR #43157, opened 2026-02-18
fix(scripts): use console.error for git command failures

pranjal2004838:
## Summary
- replace  calls used in git failure path with 
- improve error output to include the full git command context
- keep existing throw behavior while making failure messages clearer

## Testing
- npm test


pranjal2004838:
This PR is ready and tests pass locally. The failing "block" check is expected because I’m not an admin (policy gate), not due to code/test errors. Could an admin/maintainer please review and proceed with the required merge/bypass workflow? Thanks!

PR #43167, opened 2026-02-19
Fix a duplicated object name

mfuji09:
### Description

Fix a linked object name which is duplicated.
Perhaps one of them should be "DOMMatrixReadOnly".

### DIFF
--- a/files/en-us/web/api/dommatrix/frommatrix_static/index.md
+++ b/files/en-us/web/api/dommatrix/frommatrix_static/index.md
@@ -20,7 +20,7 @@ DOMMatrix.fromMatrix(other)
 ### Parameters
 
 - `other` {{optional_inline}}
-  - : A {{domxref("DOMMatrix")}}, {{domxref("DOMMatrix")}}, or an object with the same properties. All properties default to `0`. The properties are:
+  - : A {{domxref("DOMMatrix")}}, {{domxref("DOMMatrixReadOnly")}}, or an object with the same properties. All properties default to `0`. The properties are:
     - `is2D`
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/dommatrix/frommatrix_static/index.md
```diff
@@ -20,7 +20,7 @@ DOMMatrix.fromMatrix(other)
 ### Parameters
 
 - `other` {{optional_inline}}
-  - : A {{domxref("DOMMatrix")}}, {{domxref("DOMMatrix")}}, or an object with the same properties. All properties default to `0`. The properties are:
+  - : A {{domxref("DOMMatrix")}}, {{domxref("DOMMatrixReadOnly")}}, or an object with the same properties. All properties default to `0`. The properties are:
```
chrisdavidmills, line 23: Looks plausible to me. I'd maybe say something like


```suggestion
  - : A {{domxref("DOMMatrix")}} or {{domxref("DOMMatrixReadOnly")}} object instance, or another object with the same properties. All properties default to `0`. The properties are:
```

Josh-Cena: @chrisdavidmills this is standard verbiage across many pages. It's a disjunction of three object types, and saying "another" implies two coexisting objects.

chrisdavidmills: Well, as written, it seems weird to me, as it implies that `DOMMatrix` and `DOMMatrixReadOnly` aren't objects. But I don't think it's that important. Let's just get it merged.

PR #43186, opened 2026-02-22
Clarify that role="button" does not inherit button functionality

abelmark:
### Description

Clarifies the summary and description paragraphs of the `button` role page to make explicit that `role="button"` does not inherit any native button functionality — event listeners and keyboard handling must be added manually.

### Motivation

The original wording ("provides no button functionality") was ambiguous. It could be read as saying a button with this role has no useful functionality at all, rather than conveying that the role only affects how assistive technology announces the element. The updated wording makes clear that the semantic role is communicated to screen readers, but the interactive behavior (event listeners, keyboard handling) is not inherited and must be implemented by the developer.

### Additional details

The change is applied to both the opening summary sentence and the equivalent sentence in the Description section.

N/A

### DIFF
--- a/files/en-us/web/api/customelementregistry/customelementregistry/index.md
+++ b/files/en-us/web/api/customelementregistry/customelementregistry/index.md
@@ -1,0 +1,83 @@
+---
+title: "CustomElementRegistry: CustomElementRegistry() constructor"
+short-title: CustomElementRegistry()
+slug: Web/API/CustomElementRegistry/CustomElementRegistry
+page-type: web-api-constructor
+browser-compat: api.CustomElementRegistry.CustomElementRegistry
+---
+
+{{APIRef("Web Components")}}
+
+The **`CustomElementRegistry()`** constructor creates a new {{domxref("CustomElementRegistry")}} object for scoped usage.
+
+The constructor is specifically used for creating scoped registries that limit [custom element](/en-US/docs/Web/API/Web_components/Using_custom_elements) definitions to a particular scope, such as an element or {{domxref("ShadowRoot")}}.
+
+> [!NOTE]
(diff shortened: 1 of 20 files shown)


files/en-us/web/accessibility/aria/reference/roles/button_role/index.md
```diff
@@ -9,11 +9,11 @@ spec-urls:
 sidebar: accessibilitysidebar
 ---
 
-The `button` role is for clickable elements that trigger a response when activated by the user. Adding `role="button"` tells the screen reader the element is a button, but provides no button functionality. Use {{HTMLElement("button")}} or {{HTMLElement("input")}} with `type="button"` instead.
+The `button` role is for clickable elements that trigger a response when activated by the user. Adding `role="button"` tells the screen reader the element is a button, but does not inherit any button functionality — event listeners and keyboard handling must be added manually. Use {{HTMLElement("button")}} or {{HTMLElement("input")}} with `type="button"` instead.
```
hamishwillee: Thanks @abelmark . The term inherit is not incorrect here, but it does add semantics we don't need. That said, I agree that spelling out the functionality that is missing is better. So how about this (and if you agree copy below as well).


```suggestion
The `button` role is for clickable elements that trigger a response when activated by the user. Adding `role="button"` tells the screen reader the element is a button, but does not provide other typical button functionality such as click events and keyboard handling. You can add these yourself, but you should generally use {{HTMLElement("button")}} or {{HTMLElement("input")}} with `type="button"` instead.
```

abelmark: That sounds great @hamishwillee!  I will update.

abelmark: Done!

PR #43193, opened 2026-02-22
Editorial review: Add Wasm SIMD reference: type, conversion, extraction, other assorted changes

chrisdavidmills:
### Description

This PR adds reference docs for WebAssembly SIMD: type, conversion, extraction, other assorted changes

### DIFF
--- a/files/en-us/webassembly/reference/types/v128/index.md
+++ b/files/en-us/webassembly/reference/types/v128/index.md
@@ -60,7 +60,7 @@ You can interpret a `v128` value as:
 - `f32x4` : 4 32-bit float values
 - `f64x2` : 2 64-bit float values
 
-For example, the following uses the [`const`](/en-US/docs/WebAssembly/Reference/Numeric/const) instruction to create an `f32x4` value that contains 4 32-bit float values — `65.4`, `780.6`, `1011`, and `3.0`.
+For example, the following uses the [`const`](/en-US/docs/WebAssembly/Reference/Numeric/const) instruction to create an `f32x4` value that contains 4 32-bit float values — `65.4`, `780.9`, `1011.1`, and `3.0`.
 
 ```wat
 v128.const f32x4 65.4 780.9 1011.1 3.0


files/en-us/webassembly/reference/simd/conversion/ceil/index.md
```diff
@@ -0,0 +1,73 @@
...
+
+The **`ceil`** [SIMD conversion instruction](/en-US/docs/WebAssembly/Reference/SIMD/conversion) rounds the value in each lane of a `v128` value interpretation up to the nearest integer above.
+
+{{InteractiveExample("Wat Demo: ceil", "tabbed-taller")}}
+
+```wat interactive-example
+(module
+  (import "console" "log" (func $log (param i32)))
+  (func $main
+    (v128.const f32x4 1.9 2.5 0.5 12.1)
```
yurydelendik: nit: here and more below, parenthesis can be removed to be consistent with instructions below

files/en-us/webassembly/reference/simd/conversion/ceil/index.md
```diff
@@ -0,0 +1,73 @@
...
+
+    f32x4.ceil
+    i32x4.extract_lane 0
```
yurydelendik: not important, but I would reduce amount of `extract_lane 0` usages for log, just to emphasis that there are more than 1 items in the vector. skew random index distribution toward middle-high.

files/en-us/webassembly/reference/simd/extract/extract_lane/index.md
```diff
@@ -0,0 +1,77 @@
...
+    - `i64x2`
+    - `f32x4`
+    - `f64x2`
+- `extract_lane`
+  - : The `extract_lane` instruction. Must always be included after the `value_type` and a period (`.`).
+
+### Type
+
+```plain
+[lane] -> []
```
yurydelendik: this looks off for extract_lane. Ideally it shall be something like `[input, _lane_] => [item]`. The `input` comes from stack, `lane` comes as immediate operand.

files/en-us/webassembly/reference/simd/conversion/convert_i32x4_s/index.md
```diff
@@ -0,0 +1,71 @@
+---
+title: "convert_i32x4_s: Wasm SIMD conversion instruction"
+short-title: convert_i32x4_s
+slug: WebAssembly/Reference/SIMD/conversion/convert_i32x4_s
+page-type: webassembly-instruction
+browser-compat: webassembly.simd.convert_i32x4_s
+sidebar: webassemblysidebar
+---
+
+The **`convert_i32x4_s`** [SIMD conversion instruction](/en-US/docs/WebAssembly/Reference/SIMD/conversion) converts the lanes of a `v128` `f32x4` value interpretation into a signed `i32x4` value interpretation.
```
yurydelendik: This needs to be other way around: i32x4 to f32x4. I think the same problem in all converts.

The provided example outputs 1317469696, instead of expected 255.

Adjusted example.
```
(module
  (import "console" "log" (func $log (param f32)))
  (func $main
    v128.const i32x4 0x3 0x3a 0xa9 0xff

    f32x4.convert_i32x4_s
    f32x4.extract_lane 3
    call $log ;; log the result
  )
  (start $main)
)
```

files/en-us/webassembly/reference/simd/conversion/splat/index.md
```diff
@@ -0,0 +1,95 @@
...
+- `value`
+  - : The value you want to copy to all lanes of the `v128`, for example `(f32.const 16)`.
+- `output`
+  - : The resulting `v128` value after all lanes of the initial `v128` value have been updated with the `value`.
+
+### Binary encoding
+
+| Instruction   | Binary equivalent | Example text => binary                           |
+| ------------- | ----------------- | ------------------------------------------------ |
+| `i8x16.splat` | `0xfd 15:u32`     | `i8x16.splat (f32.const 15)` => `0xfd 0x0f 0x0f` |
```
yurydelendik: The `(f32.const 15)` is not important here -- it comes from the stack. For `i8x16.splat` it will be an instruction with i32 type e.g. `i32.const 15` or `i32.add`. That said, I think the binary encoding is incorrect too.

files/en-us/webassembly/reference/simd/conversion/shuffle/index.md
```diff
@@ -0,0 +1,78 @@
...
+- `input2`
+  - : The second `v128` value interpretation you want to take values from.
+- `output`
+  - : The output `v128` value interpretation.
+
+### Binary encoding
+
+| Instruction     | Binary equivalent          | Example text => binary                 |
+| --------------- | -------------------------- | -------------------------------------- |
+| `i8x16.shuffle` | `0xfd 31:u32 l:laneidx^16` | `i8x16.shuffle ...` => `0xfd 0x0d ...` |
```
yurydelendik: Typo near 31: `0xfd 13:u32 l:laneidx^16`

files/en-us/webassembly/reference/simd/conversion/splat/index.md
```diff
@@ -0,0 +1,95 @@
...
+### Type
+
+```plain
+[value] -> [output]
+```
+
+- `value`
+  - : The value you want to copy to all lanes of the `v128`, for example `(f32.const 16)`.
+- `output`
+  - : The resulting `v128` value after all lanes of the initial `v128` value have been updated with the `value`.
```
yurydelendik: "updated" is a strong word here (implied modifying something), perhaps "set to"?

files/en-us/webassembly/reference/simd/conversion/shl/index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "shl: Wasm SIMD extract instruction"
+short-title: shl
+slug: WebAssembly/Reference/SIMD/conversion/shl
```
yurydelendik: Category confusion? shl/shr are under "conversion", which are bitwise or arithmetic operations. The same problem can be with splat and shuffle.

files/en-us/webassembly/reference/simd/conversion/ceil/index.md
chrisdavidmills: This is good feedback. I've gone through both the PRs I've created so far and removed any unneccessary parens.

files/en-us/webassembly/reference/simd/conversion/ceil/index.md
chrisdavidmills: On the upside, the reader is able to edit the code sample and rerun it, to see what is contained in each lane of the vector. But your feedback is still useful — I've gone through and edited all of the samples to bias the lane values to middle-high.

files/en-us/webassembly/reference/simd/extract/extract_lane/index.md
chrisdavidmills: Ah, I'm glad you spotted this. I got a few of the type sections wrong on the first few that I did. I've corrected them now.

files/en-us/webassembly/reference/simd/conversion/convert_i32x4_s/index.md
chrisdavidmills: I'll be honest — I found these instructions pretty confusing ;-)

Thanks for the explanation; I've corrected all of these now.

files/en-us/webassembly/reference/simd/conversion/splat/index.md
chrisdavidmills: OK, I've updated several of the pages, including this one, to try to remove the values taken from the stack and just include the values that are part of the actual instruction syntax. Let me know if this looks correct now. 

It has been challenging to understand which values are part of the syntax and which are taken from the stack, and therefore what to include in these binary sections. I should make fewer errors here as I go forward.

files/en-us/webassembly/reference/simd/conversion/shuffle/index.md
chrisdavidmills: Good spot; fixed in next commit.

files/en-us/webassembly/reference/simd/conversion/splat/index.md
chrisdavidmills: That's fair. Updated in next commit.

files/en-us/webassembly/reference/simd/conversion/shl/index.md
chrisdavidmills: I wasn't sure what category the instructions should all be in, so I took some educated guesses.

Some questions:

1. Do we need bitwise _and_ arithmetic categories, or are they synonymous in this context?
2. If we only need one, which would be the preferred option?
3. Which other instructions should be in that category?

yurydelendik: shl/shr are belong to bitwise operations (like and, or, not), you there no the bitwise categore then the arithmetic

swizzle, splat and shuffle perhaps something along the lines of "Move/Permute"

files/en-us/webassembly/reference/simd/conversion/splat/index.md
```diff
@@ -0,0 +1,98 @@
...
+    - `i64x2`
+    - `f32x4`
+    - `f64x2`
+- `splat`
+  - : The `splat` instruction. Must always be included after the `value_type` and a period (`.`).
+
+### Type
+
+```plain
+[input, value] -> [output]
```
yurydelendik, line 67: `[value] -> [output]` - there instruction takes value of the specified type e.g. `i32` for `i32x4` and sets it to all lanes

files/en-us/webassembly/reference/simd/conversion/ceil/index.md
```diff
@@ -0,0 +1,73 @@
...
+{{InteractiveExample("Wat Demo: ceil", "tabbed-taller")}}
+
+```wat interactive-example
+(module
+  (import "console" "log" (func $log (param i32)))
+  (func $main
+    v128.const f32x4 1.9 2.5 0.5 12.1
+
+    f32x4.ceil
+    i32x4.extract_lane 3
```
yurydelendik: needs to extract f32 type: `f32x4.extract_lane 3`

notice that the import's type will be changed too `(import "console" "log" (func $log (param f32)))`

files/en-us/webassembly/reference/simd/conversion/floor/index.md
```diff
@@ -0,0 +1,73 @@
...
+{{InteractiveExample("Wat Demo: floor", "tabbed-taller")}}
+
+```wat interactive-example
+(module
+  (import "console" "log" (func $log (param i32)))
+  (func $main
+    v128.const f32x4 1.9 2.5 0.5 12.1
+
+    f32x4.floor
+    i32x4.extract_lane 3
```
yurydelendik: log type needs to be `f32` : `f32x4.extract_lane 3`

files/en-us/webassembly/reference/simd/conversion/convert_i32x4_s/index.md
```diff
@@ -0,0 +1,71 @@
+---
+title: "convert_i32x4_s: Wasm SIMD conversion instruction"
+short-title: convert_i32x4_s
+slug: WebAssembly/Reference/SIMD/conversion/convert_i32x4_s
+page-type: webassembly-instruction
+browser-compat: webassembly.simd.convert_i32x4_s
+sidebar: webassemblysidebar
+---
+
+The **`convert_i32x4_s`** [SIMD conversion instruction](/en-US/docs/WebAssembly/Reference/SIMD/conversion) converts the lanes of a signed `v128` `i32x4` value interpretation into an `f64x2` value interpretation.
```
yurydelendik: "an `f32x4` value interpretation"

files/en-us/webassembly/reference/simd/conversion/convert_low_i32x4_u/index.md
```diff
@@ -0,0 +1,71 @@
+---
+title: "convert_low_i32x4_u: Wasm SIMD conversion instruction"
+short-title: convert_low_i32x4_u
+slug: WebAssembly/Reference/SIMD/conversion/convert_low_i32x4_u
+page-type: webassembly-instruction
+browser-compat: webassembly.simd.convert_low_i32x4_u
+sidebar: webassemblysidebar
+---
+
+The **`convert_low_i32x4_u`** [SIMD conversion instruction](/en-US/docs/WebAssembly/Reference/SIMD/conversion) converts the first two lanes of an unsigned `v128` `i32x4` value interpretation into an `f32x4` value interpretation.
```
yurydelendik: "into a `f64x2` value interpretation"

files/en-us/webassembly/reference/simd/conversion/trunc/index.md
```diff
@@ -0,0 +1,73 @@
...
+- `input`
+  - : The `v128` value interpretation you want to convert.
+- `output`
+  - : The output `v128` value interpretation.
+
+### Binary encoding
+
+| Instruction   | Binary equivalent | Example text => binary       |
+| ------------- | ----------------- | ---------------------------- |
+| `f32x4.trunc` | `0xfd 0x69`       | `f32x4.trunc` => `0xfd 0x69` |
```
yurydelendik: minor: `0xfd 105:u32` will look more consistent in comparison with other pages

files/en-us/webassembly/reference/simd/conversion/shl/index.md
chrisdavidmills: I've moved the shl* instructions to a new "bitwise" directory.

Apologies, but I didn't understand "you there no the bitwise categore then the arithmetic". Does this mean I need an "arithmetic" category, or not? I would assume instructions like `add` would go in "arithmetic" rather than bitwise?

I'm leaving `swizzle`, `splat`, and `shuffle` where they are for now. They seem like conversions to me, and I'd rather not create an additional category just for three items. We can move them later on if needed.

files/en-us/webassembly/reference/simd/conversion/splat/index.md
chrisdavidmills: OK, so it sounds like basically all of these pages need the `input` removed.

In that case, where on the page should we represent the input value? We need to mention it somewhere. Does it go in the syntax bit just below the  "Syntax" heading? Or somewhere else? I could add an "Input" subheading, maybe.

Perhaps this is something we should have a call about, along with @eqrion?

files/en-us/webassembly/reference/simd/conversion/ceil/index.md
chrisdavidmills: Corrected. I also noticed a few other type errors, so I fixed those too.

files/en-us/webassembly/reference/simd/conversion/floor/index.md
chrisdavidmills: fixed

files/en-us/webassembly/reference/simd/conversion/convert_i32x4_s/index.md
chrisdavidmills: fixed

files/en-us/webassembly/reference/simd/conversion/convert_low_i32x4_u/index.md
chrisdavidmills: fixed

files/en-us/webassembly/reference/simd/conversion/trunc/index.md
chrisdavidmills: I've fixed that one. 

Another question: For several others, I was unable to find a binary equivalent written in that form. For example, if I search the spec PDF for `trunc_sat_f64x2_u_zero`, I can only find a binary opcode of `0xFD 0xFD 0x01`, which is what I included in the table on the MDN page.

Is that OK, or should I convert them in some way?

files/en-us/webassembly/reference/simd/conversion/shl/index.md
yurydelendik: You will have "arithmetic" sooner or later. Most of SIMD instruction such as add, mul, etc. are arithmetic, bitwise logic, or comparison.

files/en-us/webassembly/reference/simd/conversion/trunc/index.md
yurydelendik: No preference, which way is easier. The https://webassembly.github.io/spec/core/binary/instructions.html#vector-instructions lists instructions using decimal code.

files/en-us/webassembly/reference/simd/conversion/shl/index.md
chrisdavidmills: OK, that makes sense. Thanks.

files/en-us/webassembly/reference/simd/conversion/trunc/index.md
chrisdavidmills: OK, I think I might just leave it as-is, in the interests of getting this to publication quicker. The page you linked created further confusion. I can't search it, as it is one of those annoying pages we discussed on Slack.

In addition, looking through the page, I can't see `trunc_sat_f64x2_u_zero` listed, but I can see  `𝗍𝗋𝗎𝗇𝖼_𝗌𝖺𝗍_𝗎_𝗓𝖾𝗋𝗈_𝖿𝟨𝟦𝗑𝟤` listed. Is that the same instruction but written a differnt way, or a different instruction?

If it is the same instruction, why is it written differently on this page?

files/en-us/webassembly/reference/simd/conversion/splat/index.md
chrisdavidmills: Having looked at this again, I don't think we need the `input` value represented after all, as this is basically the type the instruction is being run on, i.e., the `value_type` in the previous section.

There needs to be a value of that type previously pushed onto the stack for the instruction to then be run on it, so I think we are all good here. Sorry for overthinking this previously!

chrisdavidmills: I'll go through and update all the pages to remove this excess `input` value where it is included.

chrisdavidmills: OK, all `input` values removed from the Type syntax sections.

files/en-us/webassembly/reference/simd/conversion/trunc/index.md
yurydelendik: https://webassembly.github.io/spec/versions/core/WebAssembly-3.0-draft.pdf fixed that and it is trunc_sat_f64x2_u_zero.

Also, notice that binary bytes are: `0xFD 0xFD 0x01`, though in binary format it is listed as `0xFD 253:u32` (The 253 will be represented as two bytes 0xFD 0x01)

That 0x01 is important, and that means the `convert_i32x4_s`, `convert_i32x4_u`, ` convert_low_i32x4_s` and ` convert_low_i32x4_u` will be missing that too.

chrisdavidmills: I found quite a few others that were wrong, too. This will all be fixed in my next commit.

chrisdavidmills: I now know how to write these — thanks for all the useful info.

chrisdavidmills:
> I think it looks good. Except minor details we can address later.
> 
> After talking to Ryan, the Type section needs to talk only about stack value types (and not immediates). That said, there are minor inaccuracies in splat and extract_lanes, but these can be addressed later.

Thanks, @yurydelendik!

@hamishwillee, any chance you can give this one a quick look and approve it, provided it doesn't give you any red flags? It has already been reviewed pretty thoroughly, including a couple of runs through Claude Code to check for errors and inconsistencies.

files/en-us/webassembly/reference/simd/bitwise/index.md
```diff
@@ -0,0 +1,16 @@
...
+---
+
+WebAssembly SIMD bitwise instructions.
+
+- [`shl`](/en-US/docs/WebAssembly/Reference/SIMD/bitwise/shl)
+  - : Shifts the bits in each lane of a `v128` value interpretation to the left by the same specified amount.
+- [`shr_s`](/en-US/docs/WebAssembly/Reference/SIMD/bitwise/shr_s)
+  - : Shifts the bits in each lane of a `v128` value interpretation to the right by the same specified amount, outputting signed values. This is an arithmetic right shift.
+- [`shr_u`](/en-US/docs/WebAssembly/Reference/SIMD/bitwise/shr_u)
+  - : Shifts the bits in each lane of a `v128` value interpretation to the right by the same specified amount, outputting unsigned values. This is a logical right shift.
```
hamishwillee: Generally I prefer the shortest description first. If it is worth saying at all then giving that info to the user first is the most ueful thing you can do. So if it were me (and everywhere - I won't comment again.)


```suggestion
- [`shl`](/en-US/docs/WebAssembly/Reference/SIMD/bitwise/shl)
  - : Arithmetic left shift. Shifts the bits in each lane of a `v128` value interpretation to the left by the same specified amount.
- [`shr_s`](/en-US/docs/WebAssembly/Reference/SIMD/bitwise/shr_s)
  - : Arithmetic right shift. Shifts the bits in each lane of a `v128` value interpretation to the right by the same specified amount, outputting signed values.
- [`shr_u`](/en-US/docs/WebAssembly/Reference/SIMD/bitwise/shr_u)
  - : Logical right shift. Shifts the bits in each lane of a `v128` value interpretation to the right by the same specified amount, outputting unsigned values.
```

hamishwillee: Also, I don't know/remember what a `v128` is. Perhaps links everywhere, at least on first sight.

hamishwillee: I don't know what a lane is, here or anywhere. Perhaps link. Note, I'm aware that anyone reading this probably already is quite familiar with this stuff. Linking never hurts, but it might not be worth the effort given the audience. So consider these kinds of comments as nits/suggestions.

files/en-us/webassembly/reference/simd/bitwise/shl/index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "shl: Wasm SIMD bitwise instruction"
+short-title: shl
+slug: WebAssembly/Reference/SIMD/bitwise/shl
+page-type: webassembly-instruction
+browser-compat: webassembly.simd.shl
+sidebar: webassemblysidebar
+---
+
+The **`shl`** [SIMD bitwise instruction](/en-US/docs/WebAssembly/Reference/SIMD/bitwise) shifts the bits in each lane of a `v128` value interpretation to the left by the same specified amount.
```
hamishwillee: consider this `v128` being the one to link.

files/en-us/webassembly/reference/simd/bitwise/shl/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+{{InteractiveExample("Wat Demo: shl", "tabbed-taller")}}
+
+```wat interactive-example
+(module
+  (import "console" "log" (func $log (param i32)))
+  (func $main
+    v128.const i32x4 1 2 4 6
+    i32.const 3
```
hamishwillee, line 19: EDIT ... after  re=reading....

So a `v128` is like an array that you can call the same operation on multiple values of it. In this case a bitwise left shift of the specified values by a specified number of bits.

So here we're left shifting 3 bits of _each_ value and outputting the decimal value in the 1st lane.
Since the first lane is 1 this is 1* 2 * 2 * 2 = 8 which is not 16.

Are the lanes 0 indexed?

UP to you, but I think there is a lot of guesswork here. Adding links to the `v128` above might help. Or it might be sufficient for people who have read this before and are just interested in confirmation.

files/en-us/webassembly/reference/simd/conversion/convert_low_i32x4_s/index.md
```diff
@@ -0,0 +1,71 @@
+---
+title: "convert_low_i32x4_s: Wasm SIMD conversion instruction"
+short-title: convert_low_i32x4_s
+slug: WebAssembly/Reference/SIMD/conversion/convert_low_i32x4_s
+page-type: webassembly-instruction
+browser-compat: webassembly.simd.convert_low_i32x4_s
+sidebar: webassemblysidebar
+---
+
+The **`convert_low_i32x4_s`** [SIMD conversion instruction](/en-US/docs/WebAssembly/Reference/SIMD/conversion) converts the first two lanes of a signed `v128` `i32x4` value interpretation into an `f64x2` value interpretation.
```
hamishwillee: Same 
- would be useful to link the first instances of some of these.
- I think this is saying that if you have a `v128` `i32x4` it has four i32 values, referred to as lanes.  This is going to take the values in the first two lanes and convert them into a `f64x2`.
  - So in the example below, is this layout out  0x3 0x3a where the first is the High order bytes, and the second is the low order bytes, then interpreting it as a 64 bit signed real number? 
- Should we say the words "double precision, IEEE 754"?
- So if we do that conversion, does that mean we now have 3 lanes, one for the new real number, and the other two for the two new numbers? 
- Do the unchanged values move down in their lange numbers

I'm so confused.

files/en-us/webassembly/reference/simd/bitwise/shl/index.md
```diff
@@ -0,0 +1,81 @@
...
+- `input`
+  - : The input `v128` value interpretation.
+- `shift_value`
+  - : The value that you want to shift the lanes by.
+- `output`
+  - : The output `v128` value interpretation.
+
+### Binary encoding
+
+| Instruction | Binary format  | Example text => binary          |
```
hamishwillee, line 63: I don't understand what  Example text => binary is showing me.

files/en-us/webassembly/reference/simd/bitwise/index.md
chrisdavidmills: I'll accept the suggestion, as I do think it's a better construction. I'll also search through and see if any other places could take this kind of update.

I will also look through and make sure I'm linking to the `v128` type page at least once from every page that mentions it.

The meaning of "lane" is pretty obvious to anyone in the target audience. An explanation of it should really be put in a guide somewhere, but we are handling the reference material first. What I will do for now is add a explanation of "lanes" to the `v128` page, so people can read about it there if needed.

files/en-us/webassembly/reference/simd/bitwise/shl/index.md
chrisdavidmills: Yes, lanes are zero-indexed, like JS arrays.

Lane position 1 contains the value `2`, so shifting it by three positions to the left gives us `2 * 2 * 2 * 2` = `16`.

I will explain that lanes are zero-indexed in the `v128` page addition.

files/en-us/webassembly/reference/simd/conversion/convert_low_i32x4_s/index.md
chrisdavidmills: This instruction converts the first two lanes of an `i32x4`, which contains four lanes, into a `f64x2`, which only contains two lanes.

The other two lanes are discarded.

files/en-us/webassembly/reference/simd/bitwise/shl/index.md
chrisdavidmills: It is showing you a sample Wasm text instruction and what its binary equivalent is.

chrisdavidmills:
> @chrisdavidmills The docs are very consistent and I think they make sense. I've made some suggestions which might improve this a little.
> 
> I found the examples confusing - but I think that is mostly because I am not sure of the fundamentals. I suspect your core audience for these docs would be able to work from this. Anyway, check out the comments and see if there is anything useful in them.
> 
> Approve anyway, because I think any updates you might do could happily be an addition.

Cheers mate. It is hard to grok at first if you are not used to how Wasm (or general low-level stuff like assembly language) works. It certainly took me quite a long time to understand, as it is so different from languages that I am more familiar with, like JS and Python.

files/en-us/webassembly/reference/simd/conversion/convert_low_i32x4_s/index.md
hamishwillee: @chrisdavidmills OK. The bit about the other two lanes being discarded really should be captured, and what you would get if you were to read that lane.

PR #43227, opened 2026-02-23
Clarify AudioBufferSourceNode.buffer behavior for null values

ahzvenol:
### Description

Clarified that when `buffer` is set to `null`, the node outputs silence and stops immediately if `start()` has been called. The null behavior description has been moved to the Value section for better organization.

### Motivation

According to the [W3C Web Audio API specification](https://webaudio.github.io/web-audio-api/#playback-AudioBufferSourceNode), if `buffer` is `null` and `start()` has been called, the node stops immediately.

The current documentation only stated that the node "generates silence," which could lead readers to assume the node continues playing indefinitely (outputting zeros) until `stop()` is called manually. This update clarifies that the node stops immediately.

### DIFF
--- a/files/en-us/web/api/audiobuffersourcenode/buffer/index.md
+++ b/files/en-us/web/api/audiobuffersourcenode/buffer/index.md
@@ -10,12 +10,11 @@ browser-compat: api.AudioBufferSourceNode.buffer
 
 The **`buffer`** property of the {{domxref("AudioBufferSourceNode")}} interface provides the ability to play back audio using an {{domxref("AudioBuffer")}} as the source of the sound data.
 
-If the `buffer` property is set to `null`, the node outputs a single channel of silence in which every sample is 0, and stops immediately if `start()` has been called.
-
 ## Value
 
-An {{domxref("AudioBuffer")}} which contains the data representing the sound which the
-node will play.
+An {{domxref("AudioBuffer")}} which contains the data representing the sound which the node will play, or `null`.
+
+If set to `null`, the node outputs a single channel of silence in which every sample is 0, and stops immediately if `start()` has been called.
 
 ## Exceptions
 


files/en-us/web/api/audiobuffersourcenode/buffer/index.md
```diff
@@ -10,13 +10,11 @@ browser-compat: api.AudioBufferSourceNode.buffer
...
-If the `buffer` property is set to the value `null`, the node
-generates a single channel containing silence (that is, every sample is 0).
-
 ## Value
 
-An {{domxref("AudioBuffer")}} which contains the data representing the sound which the
-node will play.
+An {{domxref("AudioBuffer")}} which contains the data representing the sound which the node will play, or `null`.
+
+If set to `null`, the node outputs a single channel of silence in which every sample is 0, and stops immediately if `start()` has been called.
```
hamishwillee, line 17: 1. What does stop immediately mean - i.e. how does it differ from outputing a sample of 0?
2. Where in the spec does it say this - I can't infer it from the URL you point to.
3. Did you test it?

What I see in spec is

> Before a source is started (by calling [start()](https://webaudio.github.io/web-audio-api/#dom-audioscheduledsourcenode-start), the source node MUST output silence (0). After a source has been stopped (by calling [stop()](https://webaudio.github.io/web-audio-api/#dom-audioscheduledsourcenode-stop)), the source MUST then output silence (0).

> Any outputs which are not "active" will output silence and would typically not be connected to anything.

ahzvenol: **1.** "Stops immediately" means the node ends its playback state, equivalent to calling `stop()` at that moment, triggering the `ended` event.

Documenting this behavior helps developers handle the edge case of assigning a valid buffer after calling `start()` with a `null` buffer, or starting with a valid buffer and then assigning `null`.

**2.** [Playback algorithm](https://webaudio.github.io/web-audio-api/#playback-AudioBufferSourceNode), lines 86-89 set the internal stop time to `currentTime`:

```javascript
// Handle null buffer case
if (buffer == null) {
    stop = currentTime; // force zero output for all time
}
```

Later in lines 153-156, this causes the node to end its playback state:

```javascript
if (currentTime >= stop) {
    // End playback state of this node.  No further invocations of process()
    // will occur.  Schedule a change to set the number of output channels to 1.
}
```

**3.** I have just tested this behavior across the three major browser engines: Chrome (Blink), Firefox (Gecko), and Safari (WebKit).

Surprisingly, none of the current implementations follow the spec behavior for this edge case. Instead of stopping, the node appears to remain active indefinitely, and the `ended` event is never triggered.

I also created a [test case](https://codepen.io/ahzvenol/pen/KwgpZVR) to demonstrate this: according to the spec, the node should have stopped before the valid buffer is assigned, producing no sound. In practice, current browsers play the audio.

However, this behavior is explicitly defined in the specification and has been present for a considerable time, including in the [W3C Recommendation (2021)](https://www.w3.org/TR/2021/REC-webaudio-20210617/#playback-AudioBufferSourceNode).

Since MDN typically documents the spec behavior, what is MDN's stance on this? Should we mention this implementation gap?

hamishwillee: @ahzvenol Thanks for taking the time to test - much appreciated.

Browsers will often lag spec changes. We don't generally document these kind of changes until at least one browser matches the specification.

No, would not document the gap. We should close this.

hamishwillee:
Closing for reasons given in https://github.com/mdn/content/pull/43227#discussion_r2869972347

Thanks @ahzvenol - appreciate the research and taking the time to fix this.

PR #43232, opened 2026-02-24
Document IntersectionObserverEntry() constructor

pepelsbey:
### Description

**New pages**

- `API/IntersectionObserverEntry/IntersectionObserverEntry` — The `IntersectionObserverEntry()` constructor

**Updated pages**

- `API/IntersectionObserverEntry` — Added Constructor section

### Motivation

Document the `IntersectionObserverEntry()` constructor, supported only in Safari 12.1+.

### Additional details

- [BCD: IntersectionObserverEntry.IntersectionObserverEntry](https://github.com/mdn/browser-compat-data/blob/main/api/IntersectionObserverEntry.json#L39)
- [Spec: The IntersectionObserverEntry interface](https://w3c.github.io/IntersectionObserver/#dom-intersectionobserverentry-intersectionobserverentry)

### DIFF
--- a/files/en-us/web/api/intersectionobserverentry/intersectionobserverentry/index.md
+++ b/files/en-us/web/api/intersectionobserverentry/intersectionobserverentry/index.md
@@ -12,6 +12,9 @@ browser-compat: api.IntersectionObserverEntry.IntersectionObserverEntry
 
 The **`IntersectionObserverEntry()`** constructor creates and returns a new {{domxref("IntersectionObserverEntry")}} object.
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


hamishwillee (commented):
@pepelsbey When would a website developer ever call this? I don't think they usefully can - it gets called by the framework when there is an interesection, and the target is the intersecting element.

We should make that note up top right?
And perhaps also in the example "you can do this but normally wouldnt".

pepelsbey:
@hamishwillee good point! [I added some notes](https://github.com/mdn/content/pull/43232/commits/3189d42539ce79943690b0f46c6a75b22939bd8b).

hamishwillee (approved):
LGTM. Thanks @pepelsbey !

PR #43235, opened 2026-02-24
Fix typo in README

srushtisp09:
Found a documentation error in the README file.  
Original statement: "Since 2005, approximately 45,000 contributors have created the documentation we know and love."  
Updated to: "Since its launch in 2005, more than 45,000 contributors have helped build and improve MDN Web Docs."


Josh-Cena:
Thanks, this is not an improvement.

PR #43238, opened 2026-02-24
Rename Payment Handler API --> Web-based Payment Handler API

stephenmcgruer:
### Description

Rename files referencing the "Payment Handler API" to "Web-based Payment Handler API"

### Motivation

This spec was renamed recently, see
https://w3c.github.io/web-based-payment-handler/

### Additional details

Move was done via:

```
npm run content move web/api/payment_handler_api Web/API/Web_Based_Payment_Handler_API
node scripts/update-moved-file-links.js
```

Then individual edits to files.

### DIFF
--- a/files/en-us/web/api/canmakepaymentevent/canmakepaymentevent/index.md
+++ b/files/en-us/web/api/canmakepaymentevent/canmakepaymentevent/index.md
@@ -8,7 +8,7 @@ status:
 browser-compat: api.CanMakePaymentEvent.CanMakePaymentEvent
 ---
 
-{{APIRef("Web-based Payment Handler API")}}{{SeeCompatTable}}{{AvailableInWorkers("service")}}
+{{APIRef("Web-Based Payment Handler API")}}{{SeeCompatTable}}{{AvailableInWorkers("service")}}
 
 The **`CanMakePaymentEvent()`** constructor creates a new {{domxref("CanMakePaymentEvent")}} object instance.
 
--- a/files/en-us/web/api/canmakepaymentevent/index.md
+++ b/files/en-us/web/api/canmakepaymentevent/index.md
@@ -7,7 +7,7 @@ status:
 browser-compat: api.CanMakePaymentEvent
 ---
 
-{{APIRef("Web-based Payment Handler API")}}{{SeeCompatTable}}{{AvailableInWorkers("service")}}
+{{APIRef("Web-Based Payment Handler API")}}{{SeeCompatTable}}{{AvailableInWorkers("service")}}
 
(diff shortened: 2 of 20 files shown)


stephenmcgruer:
Thanks for the review! I've now done a *much* more thorough pass, as well as learned a bit more about how MDN references work. Not ready for re-review quite yet, I just wanted to be able to examine my changes in the GitHub UI . Will flag for re-review when ready :).

files/en-us/web/api/canmakepaymentevent/canmakepaymentevent/index.md
```diff
@@ -53,7 +53,7 @@ self.addEventListener("canmakepayment", (e) => {
 
 ## See also
 
-- {{domxref("Payment Handler API", "Payment Handler API", "", "nocode")}}
+- {{domxref("Web-based Payment Handler API", "", "", "nocode")}}
```
stephenmcgruer, line 56: A number of files had what I believe is an unnecessary duplication of "Payment Handler API" here, because you only (?) need that if you want the rendered text to be different than the link/reference first parameter. Lmk if that wasn't correct and I can restore the duplication.

files/jsondata/GroupData.json
```diff
@@ -2045,6 +2031,20 @@
       "properties": [],
       "events": []
     },
+    "Web-Based Payment Handler API": {
```
stephenmcgruer, line 2034: The two .json files appear to be in alphabetical order, but I wasn't sure how they treat the order of a dash "-" vs an alphabetical character. Lmk! :)

stephenmcgruer (commented):
Ok, ready for review now I think.

files/en-us/web/api/canmakepaymentevent/canmakepaymentevent/index.md
hamishwillee: Correct - this is if you want to use a different text to the page title. 

FWIW 
- personally I never use domxref for guide links because there is no benefit to it and the rendering is less clear to editors - I do `[Whatever text I want](https://pr43238.review.mdn.allizom.net/en-US/docs/Web/API/Web-Based_Payment_Handler_API)`.
But I don't bother changing it either when I see cases like this.
- More generally, in the Web API section we are ambivalent about using domxref of a bare link. Again though, we don't change them to one or the other in review. CSS section takes a different view and always uses the macro.

files/jsondata/GroupData.json
hamishwillee: I have no idea. Assume `-` is whitespace unless someone explicitly argues.

PR #43248, opened 2026-02-25
fix(sidebars): deduplicate JavaScript RegExp sidebar entry (clean)

pranjal2004838:
Renames the duplicate sidebar key for the Regular expressions reference to Reference_RegExp and adds matching localization entries. Change limited to files/sidebars/jssidebar.yaml.


Josh-Cena:
It is well the intention that these two items share the exact same text because so it is written in English. I suspect you are using AI to generate these PRs; you are advised to stop sending multiple PRs with bad branches and verify that your branch is clean before you submit them.

PR #43251, opened 2026-02-25
[Editorial review] BiDi - Add pages for new, status, and end commands of the session module

dipikabh:
### Description

This PR adds the following pages:
- `session` listing page
- `session.new`
- `session.status`
- `session.end`

It also updates:
- [Modules landing page](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/BiDi/Modules) - adds basic info about how commands and events are structured
-  WebDriver sidebar (`webdriver.yaml`)

**UPDATE**: As part of addressing review feedback, this PR now also [adds stub pages](https://github.com/mdn/content/pull/43251#discussion_r2866264690) for all modules.

### Additional details

Since BiDi  builds on capabilities and errors defined in Classic, we can share a lot of those pages between the two.
So it might make sense to move them to common location under `Web/WebDriver/Reference` and out of the folders under `Web/WebDriver/Reference/Classic` where they currently sit. 

- I plan to place them directly under `Web/WebDriver/Reference` so that both Classic and BiDi commands can link to them.
- On the individual capability and error pages, we can note any differences in the behavior or format between Classic and BiDi.

I'll make these changes in a separate PR. ✅ 
**UPDATE**: Done in #43314

@whimboo pointed out that the top landing page https://developer.mozilla.org/en-US/docs/Web/WebDriver is also in need of an update since the addition of our BiDi section - I'll update this page along with the above structural changes.

### Spec links

- `session.new`: https://w3c.github.io/webdriver-bidi/#command-session-new
- `session.status`: https://w3c.github.io/webdriver-bidi/#command-session-status
- `session.end`: https://w3c.github.io/webdriver-bidi/#command-session-end

### Related issue

Doc issue: mdn/mdn#339


files/sidebars/webdriver.yaml
```diff
@@ -7,30 +7,44 @@ sidebar:
...
-    children:
-      - type: listSubPages
-        path: /Web/WebDriver/Reference/BiDi/Modules
-        link: /Web/WebDriver/Reference/BiDi/Modules
-        details: closed
-  - link: /Web/WebDriver/Reference/Classic
-    title: Classic
+  - type: section
+    link: /Web/WebDriver/Reference/BiDi
+    title: Bidi
```
juliandescottes:
```suggestion
    title: BiDi
```

I would preserve the current capitalization

juliandescottes (commented):
This looks great! I'll let Henrik do the final technical review, just one nitpick on my side.

files/en-us/web/webdriver/reference/bidi/modules/index.md
```diff
@@ -8,15 +8,67 @@ sidebar: webdriver
...
 The WebDriver BiDi protocol is organized into modules.
-Each module represents a collection of related commands and events used in specific browser automation cases.
+Each module represents a collection of related [commands](#commands) and [events](#events) used in specific browser automation cases.
 
 Both command and event names use the module name as a prefix: `module_name.command_name` for commands and `module_name.event_name` for events.
 
 ## List of modules
 
-The following is a list of the available WebDriver BiDi modules:
+- `bluetooth`
```
whimboo: I wonder if we should add a little description along those entries to explain in a short sentence what each module is actually used for. This can be a follow-up PR.

files/en-us/web/webdriver/reference/bidi/modules/index.md
```diff
@@ -8,15 +8,67 @@ sidebar: webdriver
...
+{
+  "id": 1,
+  "method": "session.new",
+  "params": {}
+}
+```
+
+Each command results in either a success response containing a `result` field or an error response containing an `error` field. The structure of `result` is specific to each command.
+
+Most commands require an active session. However, some commands are **static** — they can run without an active session. For example, the [`session.status`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/status) command is static because it can be used to check whether the browser is ready to create a new session before actually creating one.
```
whimboo: So far only `session.status` is static. So I would propose a rewording:

```suggestion
All commands except [`session.new`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/new) and [`session.status`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/status) require an active WebDriver BiDi session in order to be properly targeted.
```

files/en-us/web/webdriver/reference/bidi/modules/index.md
```diff
@@ -8,15 +8,67 @@ sidebar: webdriver
...
+To receive events, the client must first subscribe to them using the `session.subscribe` command. The subscription message has the same `id` and `method` fields as any other command, and the `params` field takes an `events` array containing the event names to subscribe to:
+
+```json
+{
+  "id": 2,
+  "method": "session.subscribe",
+  "params": {
+    "events": ["network.responseCompleted"]
+  }
+}
```
whimboo: I wonder if we should only mention that `session.subscribe` is necessary and move all the rest to the future MDN page. Otherwise two different places would need to be updated.

Note that the subscription can also happen at the module level like `"events": ["browsingContext"]` - then the client will subscribe to all available events of the module.

files/en-us/web/webdriver/reference/bidi/modules/session/index.md
```diff
@@ -0,0 +1,26 @@
+---
+title: session module
+short-title: session
+slug: Web/WebDriver/Reference/BiDi/Modules/session
+page-type: listing-page
+browser-compat: webdriver.bidi.session
+sidebar: webdriver
+---
+
+The **`session`** module contains commands for monitoring the status of the browser and for managing the lifecycle and event subscriptions of a WebDriver BiDi session.
```
whimboo: I would remove the monitoring part which is rarely used compared to the other commands.

```suggestion
The **`session`** module contains commands for managing the lifecycle and event subscriptions of a WebDriver BiDi session.
```

files/en-us/web/webdriver/reference/bidi/modules/session/index.md
```diff
@@ -0,0 +1,26 @@
...
+
+The `session` module has no associated events.
+
+## Specifications
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
```
whimboo, line 26: Interesting that at this level the individual command parameters are included. Can we disable that?

files/sidebars/webdriver.yaml
whimboo: I even wonder if we should use the full name `WebDriver BiDi` to not get in conflicts with other bi-directional things.

files/sidebars/webdriver.yaml
```diff
@@ -7,30 +7,44 @@ sidebar:
...
+    link: /Web/WebDriver/Reference/Classic/Commands
+    details: closed
+  - type: listSubPages
+    path: /Web/WebDriver/Reference/Classic/Capabilities
+    link: /Web/WebDriver/Reference/Classic/Capabilities
+    details: closed
+    code: true
+  - type: listSubPages
+    path: /Web/WebDriver/Reference/Classic/Errors
+    link: /Web/WebDriver/Reference/Classic/Errors
```
whimboo: Note that not all errors are shared between both protocols. Some are only for classic and other only for BiDi.

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Return value
+
+None.
+
+## Examples
+
+### Ending an automation session with the browser
+
+With a WebDriver BiDi connection established, send the following message to end the current session:
```
whimboo, line 33: We need a paragraph somewhere that contains the steps as needed to create a WebDriver BiDi (websocket) connection.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+    - `alwaysMatch` {{optional_inline}}
+      - : An object that specifies features that the session must match.
+        If the browser cannot satisfy all capabilities in this object, the session is not created.
+    - `firstMatch` {{optional_inline}}
+      - : An array of objects that specify alternative capability sets.
+        The browser tries each entry in the order specified and creates a session using the first it can fully satisfy. If none can be satisfied, the session is not created.
+
+Each object may include any of the following optional capabilities:
+
+- `acceptInsecureCerts` {{optional_inline}}
```
whimboo, line 44: I wonder if we want to have a separate page similar to https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/Classic/Capabilities#list_of_capabilities. Note that not all capabilities will be taken care of in both protocols (eg. `webSocketUrl` only works in WebDriver classic).

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+  - : A boolean that indicates whether untrusted TLS certificates (for example, self-signed or expired) are accepted for the duration of the session.
+- `browserName` {{optional_inline}}
+  - : A string that specifies the name of the browser to use (for example, `"firefox"` or `"chrome"`).
+- `browserVersion` {{optional_inline}}
+  - : A string that specifies the browser version to match (for example, `"120.0"`).
+- `platformName` {{optional_inline}}
+  - : A string that specifies the operating system to match (for example, `"windows"`, `"mac"`, or `"linux"`).
```
whimboo: Feel free to add `android` as well.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+- `proxy` {{optional_inline}}
```
whimboo, line 52: What is your plan to further describe this object? Do we need sub pages or would it be fine to have it all in one page when we separate out the content to a Capabilities page like we have in classic?

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+    - `userAgent`
+      - : A string that contains the browser's user agent string (for example, `"Mozilla/5.0 (X11; Linux x86_64; rv:130.0) Gecko/20100101 Firefox/130.0"`).
+    - `proxy` {{optional_inline}}
+      - : An object that describes the active proxy configuration. An empty object (`{}`) indicates no proxy is configured.
+    - `unhandledPromptBehavior` {{optional_inline}}
+      - : An object that describes the default behavior when a user prompt (such as an `alert`, `confirm`, or `prompt` dialog) is encountered during a command. This field is present only when specified in the `capabilities` parameter.
+    - `webSocketUrl` {{optional_inline}}
+      - : A string that contains the WebSocket URL for the session.
+
+The browser may also return vendor-specific capabilities prefixed with a browser identifier (for example, `moz:buildID` for Firefox).
```
whimboo, line 84: I wonder if we can have a separate page for those vendor specific capabilities.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+      "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:147.0) Gecko/20100101 Firefox/147.0",
+      "proxy": {}
+    }
+  }
+}
+```
+
+### Attempting to create a session when one already exists
+
+If a session is already active, sending `session.new` results in an error response:
```
whimboo: FYI that this will not happen for Chrome and derivates because they already have support for multiple sessions (which is not spec'ed yet)

whimboo (changes requested):
That's a great start with the session module @dipikabh! Very well written with all the details and nice examples. I've some questions and suggestions still, which you can find inline. Please let one of us know when something is unclear, and we can help to clarify.

files/sidebars/webdriver.yaml
dipikabh: Good catch! Fixing to "BiDi".

This is in the sidebar, which already does have a heading for "WebDriver". 
"WebDriver" also appears in the breadcrumbs and URL, so there will be context.

<kbd>Image</kbd>

files/en-us/web/webdriver/reference/bidi/modules/index.md
dipikabh: Applied as suggested

files/en-us/web/webdriver/reference/bidi/modules/index.md
dipikabh: I've moved around the sentences in this section so the flow is now slightly different. 
I've retained `session.unsubscribe`, added ability for module-level subscription, and replaced the example with a sample message from browser.

files/en-us/web/webdriver/reference/bidi/modules/index.md
dipikabh: In that case, I've replaced the manual listing of modules with a macro (`{{SubpagesWithSummaries}}`) that will automatically pick the intro sentences from module pages.
I've gone a step further in this PR and added stub pages for all modules so that their descriptions get picked up by the macro here.

Note that in the BCD repo:
- The following spec links are not correct:
  - https://w3c.github.io/webdriver-bidi/#module-bluetooth in [bluetooth.json](https://github.com/mdn/browser-compat-data/blob/main/webdriver/bidi/bluetooth.json)
  - https://w3c.github.io/webdriver-bidi/#module-userAgentClientHints in [userAgentClientHints.json](https://github.com/mdn/browser-compat-data/blob/main/webdriver/bidi/userAgentClientHints.json)
- `permission` (https://github.com/mdn/browser-compat-data/blob/main/webdriver/bidi/permission.json) should be `permissions`

files/en-us/web/webdriver/reference/bidi/modules/session/index.md
dipikabh: I don't think we can - but we can check with @caugner if there's a way to limit the depth (in either rari or browser-compat-data)

whimboo (approved):
I think that looks good for the initial landing of the session module. Great additions to have the placeholders for all modules present as well now. Thanks a lot!

files/sidebars/webdriver.yaml
dipikabh: Thanks for flagging -  yes, as discussed, the plan is to move Capabilities and Errors out of `Classic` (doing it in [#43314](https://github.com/mdn/content/pull/43314)).
In future, each capability and error page will note whether it applies to Classic, BiDi, or both.

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
dipikabh: Agreed, that's a good idea. I think this deserves a dedicated guide page rather than a brief mention in a command reference page. I'll put together a guide page along with my next PR where I was planning to cover the remaining `session` module commands.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: Once https://github.com/mdn/content/pull/43314 is merged, the shared Capabilities pages at `Reference/Capabilities` will cover both protocols, so we won't be needing separate Classic and BiDi capabilities pages. And individual capability pages can then note any protocol-specific behavior.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: Done in https://github.com/mdn/content/pull/43251/commits/f88503e8404a1ca549f1efc9b223c81bb2f01932

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: Yes, I'll use the dedicated capability page to cover the details about this object.
After https://github.com/mdn/content/pull/43314, we'll have a place to add this page. I'll include it in a follow-up.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: There's a precedent with [firefoxOptions](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/Classic/Capabilities/firefoxOptions) for vendor-specific capabilities. However, that's client-sent.

We don't have dedicated pages for browser-sent capabilities. So we might not cover them in that detail. But if important, I can include an example or a note about Firefox-specific metadata included in return value.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: That's good to know! I did test only in Firefox.

We have two options:
- Remove this example section altogether since it's not true for all browsers. And after the support is added, users will not run into this situation.
- Keep it and add caveat about "if multiple sessions not supported" like this:
"In browsers that don't support multiple sessions (e.g., Firefox), sending `session.new` when a session is already active results in this error."

dipikabh:
Thanks a lot @whimboo for the tech review and approval!

I've addressed/answered your comments and questions. There is just [one more minor update](https://github.com/mdn/content/pull/43251#discussion_r2873420996) since your last review.

I also have two open questions ([here](https://github.com/mdn/content/pull/43251#discussion_r2873761765) and [here](https://github.com/mdn/content/pull/43251#discussion_r2873823151)) before I invite the content team for an edit review.

dipikabh:
Rebased this branch to include the changes in #43314, which moves Capabilities and Errors out of `Classic`.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
whimboo: Sounds good. I assume that we will also be able to label individual capabilities with `classic` and `bidi` to indicate in which protocol they are supported?

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
whimboo: I actually like the second proposal given that users can end-up in this situation so it would be good to call it out.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
whimboo: Is it that we don't want to cover vendor specific capabilities or what's the reason why we cannot cover them in such a detail? Maybe we could have a separate section on the firefoxOptions page that handles return-only capabilities?

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: Not labels, but yes in prose, we should be able to mention the protocol that supports them

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: Cool. I've updated it in https://github.com/mdn/content/pull/43251/changes/9c7a4226f6588eaf3cf0dc9ec7efa2e2e465fc3e.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: As I understand it, this metadata, `moz:buildID`, comes under `Extensible` in the spec - it's an implementation detail that can vary by browser. That said, we do already document [firefoxOptions](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/Classic/Capabilities/firefoxOptions).

It might open the door for other browsers to add content for their return capability parameters.

I'm curious - how many such Firefox-specific fields are there in the return value, and what is the expected level of detail?
Is it something your team can help put together as a first draft and I can edit?

> separate section on the firefoxOptions page that handles return-only capabilities

I think if we document return-only capabilities, they would be better as new, separate pages.

dipikabh:
Hi Chris, I'd appreciate if you can provide me an editorial review for this when you get a chance. Thanks!

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
whimboo: Currently we have the following Mozilla specific vendor capabilities:

https://searchfox.org/firefox-main/rev/0894e095dfb46b6dac785365432791a54861fe09/remote/shared/webdriver/Capabilities.sys.mjs#492-497,508-510

Some of them are only returned and others are used for processing capabilities as well (means, they will be fed with the new session command).

files/en-us/web/webdriver/reference/bidi/modules/index.md
whimboo: I forgot to close this thread. I'm fine with it.

files/en-us/web/webdriver/reference/bidi/modules/session/index.md
whimboo: Note that this issue should not block us from merging this PR.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
whimboo: Ok, thanks!

files/en-us/web/webdriver/reference/bidi/modules/bluetooth/index.md
```diff
@@ -0,0 +1,18 @@
+---
+title: bluetooth module
+short-title: bluetooth
+slug: Web/WebDriver/Reference/BiDi/Modules/bluetooth
+page-type: listing-page
+browser-compat: webdriver.bidi.bluetooth
+sidebar: webdriver
+---
+
+The **`bluetooth`** module contains commands for automating Bluetooth device interactions.
```
chrisdavidmills, line 10: To be more in keeping with the standard MDN landing page structures (e.g. Interface landing pages), shouldn't the form of this opening sentence on module pages be something like


```suggestion
The **`bluetooth`** module of the [WebDriver BiDi protocol](/en-US/docs/Web/WebDriver/Reference/BiDi) contains commands for automating Bluetooth device interactions.
```

files/en-us/web/webdriver/reference/bidi/modules/emulation/index.md
```diff
@@ -0,0 +1,18 @@
+---
+title: emulation module
+short-title: emulation
+slug: Web/WebDriver/Reference/BiDi/Modules/emulation
+page-type: listing-page
+browser-compat: webdriver.bidi.emulation
+sidebar: webdriver
+---
+
+The **`emulation`** module contains commands and events related to emulating browser APIs.
```
chrisdavidmills: Are these Web APIs, internal browser APIs, both, or something else?

files/en-us/web/webdriver/reference/bidi/modules/webextension/index.md
```diff
@@ -0,0 +1,18 @@
+---
+title: webExtension module
+short-title: webExtension
+slug: Web/WebDriver/Reference/BiDi/Modules/webExtension
+page-type: listing-page
+browser-compat: webdriver.bidi.webExtension
+sidebar: webdriver
+---
+
+The **`webExtension`** module contains commands for managing browser extensions.
```
chrisdavidmills: Link to https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions?

files/sidebars/webdriver.yaml
```diff
@@ -12,6 +12,20 @@ sidebar:
         path: /Web/WebDriver/Reference/BiDi/Modules
         link: /Web/WebDriver/Reference/BiDi/Modules
         details: closed
+        children:
```
chrisdavidmills: If you just want to list the subpages, and aren't bothered about any particular type of order or customization, you could future-proof this using a `type: listSubPages` menu entry. See the `webassemblysidebar.yaml` file for examples.

files/en-us/web/webdriver/reference/bidi/modules/bluetooth/index.md
chrisdavidmills: Ah, you might not want to do this because you are automatically listing these pages on the modules landing page with a macro.

files/en-us/web/webdriver/reference/bidi/modules/index.md
```diff
@@ -8,15 +8,59 @@ sidebar: webdriver
...
 
 ## List of modules
 
-The following is a list of the available WebDriver BiDi modules:
+{{SubpagesWithSummaries}}
 
-{{ListSubPages}}
+## Commands
+
+A command is an asynchronous operation sent from the client to the browser. Each command message you send to the browser has three fields:
```
chrisdavidmills, line 21: It feels a bit odd having these sections on the modules reference landing page. They feel more like guide material, which probably isn't needed by a developer experienced with WebDriver BiDi, looking at reference material, yet they gloss over the basics enough that it isn't very useful for a WebDriver BiDi beginner like me, who doesn't know how to set up the required environment to send a command. Could you break these out to a more detailed guide or tutorial section, or were you planning to do that in a future PR?

files/en-us/web/webdriver/reference/bidi/modules/index.md
```diff
@@ -8,15 +8,59 @@ sidebar: webdriver
...
+
+```json
+{
+  "id": 1,
+  "method": "session.new",
+  "params": {}
+}
+```
+
+Each command results in either a success response containing a `result` field or an error response containing an `error` field. The structure of `result` is specific to each command.
```
chrisdavidmills, line 37: Would it make sense to use "Result" in the reference pages rather than "Return value" like you currently do?

files/en-us/web/webdriver/reference/bidi/modules/index.md
```diff
@@ -8,15 +8,59 @@ sidebar: webdriver
...
+
+All commands except [`session.new`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/new) and [`session.status`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/status) require an active WebDriver BiDi session in order to be properly targeted.
+
+## Events
+
+An event is a notification sent by the browser to the client when something of interest occurs.
+To receive events, the client must first subscribe to them using the `session.subscribe` command.
+
+The client can subscribe to a specific event or to all events in a module. For example, subscribing to `"browsingContext.contextCreated"` subscribes the client to that single event, while subscribing to `"browsingContext"` subscribes the client to every event in the `browsingContext` module. Once subscribed, the browser sends matching events on its own, without the client having to request each one.
```
chrisdavidmills: The last sentence in this paragraph is a little odd — for a start, isn't it a bit obvious (the whole point of events is that they fire automatically), don't events fire? Or is "sends matching events" terminology specific to WebDriver?

files/en-us/web/webdriver/reference/bidi/modules/session/index.md
```diff
@@ -0,0 +1,26 @@
...
+page-type: listing-page
+browser-compat: webdriver.bidi.session
+sidebar: webdriver
+---
+
+The **`session`** module contains commands for managing the lifecycle and event subscriptions of a WebDriver BiDi session.
+
+## Commands
+
+{{ListSubPages}}
```
chrisdavidmills, line 14: Wouldn't this be better as `{{SubpagesWithSummaries}}`? At the moment, this is just listing the page titles in a bulleted list.

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
```diff
@@ -0,0 +1,66 @@
+---
+title: session.end command
+short-title: session.end
+slug: Web/WebDriver/Reference/BiDi/Modules/session/end
+page-type: webdriver-command
+browser-compat: webdriver.bidi.session.end
+sidebar: webdriver
+---
+
+The `session.end` [command](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#commands) of the [`session`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session) module ends the current BiDi session with the browser.
```
chrisdavidmills: Feels like this could do with being a little more descriptive in terms of who the actors are, who is doing what, that kind of stuff. I know very little about BiDi, but from my knowledge of WebDriver classic, it could/should be something more like


```suggestion
The `session.end` [command](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#commands) of the [`session`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session) module ends the current session that the BiDi driver has established with a browser.
```

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
```diff
@@ -0,0 +1,66 @@
...
+```json-nolint
+{
+  "method": "session.end",
+  "params": {}
+}
+```
+
+### Parameters
+
+None. However, an empty params object `{}` must be sent.
```
chrisdavidmills:
```suggestion
None. However, an empty params object `{}` must be included in the command object.
```

Or something like that?

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+```json
+{
+  "id": 2,
+  "type": "success",
+  "result": {}
+}
+```
+
+The success response is received before the WebSocket connection closes.
```
chrisdavidmills, line 53: Ah, so the success response is different from the return value. Would it make sense to include a "Response" section of some kind in the reference pages to indicate the responses returned on success and failure?

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+short-title: session.new
+slug: Web/WebDriver/Reference/BiDi/Modules/session/new
+page-type: webdriver-command
+browser-compat: webdriver.bidi.session.new
+sidebar: webdriver
+---
+
+The `session.new` [command](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#commands) of the [`session`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session) module creates a new BiDi session with the browser.
+
+This is a static command; it can run without an active session.
```
chrisdavidmills: A couple of points:

- Would it make more sense to say "it runs without an active session", at least for this static command? It sounds weird to state this, given that the whole point of this command is to create an active session.
- I can see from the spec that "static command" is official BiDi terminology for a command that runs without an active session. I'm just thinking forward to the possibility of people new to BiDi who have previous experience with class-based languages getting confused about this terminology. I doubt it will be too much of an issue; probably, a note in a beginner's guide somewhere will suffice.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+### Parameters
+
+The input is an object with the following fields:
+
+- `capabilities`
+  - : An object with the following optional fields:
+    - `alwaysMatch` {{optional_inline}}
+      - : An object that specifies features that the session must match.
```
chrisdavidmills:
```suggestion
      - : An object that specifies features that the session must support.
```

WebGPU has a [similar pattern](https://developer.mozilla.org/en-US/docs/Web/API/GPUAdapter/requestDevice#requiredfeatures) of creating devices that support specific capability sets. I thought "match" was slightly odd phrasing, and "support" would make more sense ad be more consistent.

Of course, if "match" is the correct BiDi terminology, ignore me ;-)

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+Each object may include any of the following optional capabilities:
+
+- `acceptInsecureCerts` {{optional_inline}}
+  - : A boolean that indicates whether untrusted TLS certificates (for example, self-signed or expired) are accepted for the duration of the session.
+- `browserName` {{optional_inline}}
+  - : A string that specifies the name of the browser to use (for example, `"firefox"` or `"chrome"`).
+- `browserVersion` {{optional_inline}}
+  - : A string that specifies the browser version to match (for example, `"120.0"`).
+- `platformName` {{optional_inline}}
+  - : A string that specifies the operating system to match (for example, `"windows"`, `"mac"`, `"android"`, or `"linux"`).
```
chrisdavidmills, line 51: For values like this (and `browserName`), I'd expect to see a list of accepted values, not a short "for example" list. I'd imagine it to be like enums in JS APIs. Is it more the case that you define what these values can be in some kind of config file somewhere?

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+- `proxy` {{optional_inline}}
+  - : An object that specifies the proxy configuration the browser should use for network requests.
+- `unhandledPromptBehavior` {{optional_inline}}
+  - : An object that specifies the default behavior when a user prompt (such as an `alert`, `confirm`, or `prompt` dialog) is encountered during a command.
```
chrisdavidmills, line 55: What fields can this object contain?

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+- `sessionId`
+  - : A string that contains the unique identifier for the newly created session.
+- `capabilities`
+  - : An object that describes the capabilities that were negotiated and are active for the session. It includes the following fields:
+    - `acceptInsecureCerts`
+      - : A boolean that indicates whether untrusted TLS certificates (for example, self-signed or expired) are accepted for the duration of the session.
+    - `browserName`
+      - : A string with the name of the browser.
+    - `browserVersion`
+      - : A string with the version of the browser.
```
chrisdavidmills:
```suggestion
      - : A string containing the browser version.
```

Maybe a similar structure for the other similar ones, too?

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
```diff
@@ -0,0 +1,192 @@
...
+    - `platformName`
+      - : A string with the name of the operating system.
+    - `setWindowRect`
+      - : A boolean that indicates whether the browser supports the `setWindowRect` command.
```
chrisdavidmills: Feels like this needs a cross link. To [here](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/Classic/Commands/SetWindowRect), I'm assuming?

files/en-us/web/webdriver/reference/bidi/modules/session/status/index.md
```diff
@@ -0,0 +1,91 @@
...
+### Return value
+
+An object with the following fields:
+
+- `ready`
+  - : A boolean that indicates whether the browser is ready to create new sessions.
+    - `true`
+      - : The browser is ready to create a new session.
+    - `false`
+      - : The browser cannot accept new sessions because it already has an active session or is otherwise in a state where creating a session would fail.
```
chrisdavidmills, line 36: Could you include some information about what kinds of states can cause the session to fail? Are the exact lists browser/OS-specific, and therefore not really practical to include them all? If so, it might be good to state that.

chrisdavidmills (changes requested):
Hey @dipikabh, here's my editorial review!

I think this is largely useful and well-structured. My comments are fairly picky, and some are definitely a result of my inexperience with WebDriver BiDi. Still worth asking through, IMO.

I was approaching these articles as a beginner (which I am) while trying to put myself in the mindset of someone with previous BiDi experience, and thinking about what they might need.

files/en-us/web/webdriver/reference/bidi/modules/webextension/index.md
dipikabh: done

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
dipikabh: So while in classic, there is a "driver" (like geckodriver and chromedriver), a separate process that sits between the client and the browser, in BiDi, there is no intermediary...the client directly connects to the browser over a WebSocket connection.

However, to your point about making the actors clear, I've updated this to:

> The `session.end` [command](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#commands) of the [`session`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session) module ends the client's current BiDi session with the browser.

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
dipikabh: I do like "included" over "sent". I also went ahead and changed the sentence into active voice for clarity. 

> None. However, you must include a `params` field and set it to an empty object (`{}`).

Also updated in `session.status`.

(I've dropped "in the command object" because it felt a bit redundant given the context of the page.)

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
dipikabh: The JSON response received from the browser has the `"type"` field, and other fields in the response depend on whether `"type"` value is `"success"` or `"error"`.

So yes, there's a success or error type of response and then accordingly a result object or an error message. But they're all part of the same JSON response.

If `"type": "success"`, then there's a `"result"` field in the response.
I've documented the fields of the `"result"` object in the "Return value" section. (for example, fields in the `result` object of [session.new](https://pr43251.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/new#return_value)).
If `"type": "error"`, the response includes the `"error"` type which is covered in the "Errors" section.

> Would it make sense to include a "Response" section of some kind in the reference pages to indicate the responses returned on success and failure?

I've briefly explained the structure of the response message in the [Commands](https://pr43251.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#commands) section.

Does this address your question, or did you have something else in mind?

dipikabh: I've now expanded the return value on `session.end` to:

```
### Return value

The `result` field is an empty object (`{}`).
```

And also clarified on `session.new` that the fields described map to the `result` object in the response:

```
### Return value

A `result` object with the following fields:
```

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
chrisdavidmills: Sounds good, thanks.

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
chrisdavidmills: Great!

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
chrisdavidmills: This all makes more sense now, thanks!

So, when you say "`error` type", is this an `error` property containing the listed error as a string? Taking https://pr43251.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/new#errors as an example?

whimboo: Yes, the error that we return is not a plain number but a unique string. The list of all the available errors can be found at https://w3c.github.io/webdriver-bidi/#errors.

dipikabh: That's right - the error is listed as a string.
The example at https://pr43251.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/new#attempting_to_create_a_session_when_one_already_exists demos a `"type": "error"` situation and the resulting corresponding `error` and `message` fields in the response.

files/en-us/web/webdriver/reference/bidi/modules/bluetooth/index.md
dipikabh: Correct, leaving as is

files/en-us/web/webdriver/reference/bidi/modules/emulation/index.md
dipikabh: Good poke, made me rewrite this as:

> The **`emulation`** module contains commands for overriding the behavior of certain Web APIs, such as the [Geolocation API](/en-US/docs/Web/API/Geolocation_API) and the [Screen Orientation API](/en-US/docs/Web/API/Screen_Orientation_API), to simulate different environments.

Also, I checked again, there are no events in this module

files/en-us/web/webdriver/reference/bidi/modules/session/index.md
dipikabh: The bulleted list is kind of intentional - the summaries macro would produce some repetitive phrases across the descriptions as shown below (unless we hand curate them):

<kbd>image</kbd>


We're quite inconsistent with the style on our landing pages across MDN - HTML uses tables, some CSS pages use lists, and some other CSS pages use hand-maintained definition lists possibly to avoid the repeated phrases.

files/en-us/web/webdriver/reference/bidi/modules/index.md
dipikabh: In my mind, I'm mapping BiDi concepts to Web API equivalents..so modules in BiDi map to interfaces, and commands in BiDi map to methods. This helps me follow the corresponding template/convention.
Using "Result" would be unlike the type of title we use for this content...(we usually use it to show the output for our live examples)

But as a follow-up to https://github.com/mdn/content/pull/43251#discussion_r2908018809, I've refined the sentence further (added "in the response"):

```
### Return value

The `result` object in the response with the following fields:
```

files/en-us/web/webdriver/reference/bidi/modules/index.md
dipikabh: You're right - there's redundancy here. I've removed that sentence.

files/en-us/web/webdriver/reference/bidi/modules/index.md
dipikabh: The reason why I included the "Commands" and "Events" sections on the "Modules" landing page is because I needed a place to explain the generic structure of a command and an event and how responses look.
Every command's first line links to this "Commands" section.

It is unlikely we'll have a Commands landing page, but if we do then we might be able to move this info there.

I don't think this is guide material. I think it's useful for someone starting off with WebDriver BiDi to understand how the protocol is structured into modules that's made up of commands and events, and what's the general structure of command request messages and event notifications because that's what the devs will be working with. So it's short enough for someone to orient themselves before diving into individual command pages.

(fyi - if it helps, I've written a how-to page to set up a BiDi connection; it's part of https://github.com/mdn/content/pull/43343.)

files/en-us/web/webdriver/reference/bidi/modules/session/status/index.md
dipikabh: The spec is not very clear on this.

I'll need to defer to @whimboo for this one - could you help with a list or examples of situations in which session creation would fail?
We already have [one example](https://pr43251.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/status#checking_browser_status_before_creating_a_session) documented for when `ready` is `false`: `"message": "Session already started"`.
Are there more such known situations and are they implementation-specific?

files/sidebars/webdriver.yaml
dipikabh: Uh oh - i noticed that the `session` sub-pages were not appearing in the sidebar at all.

After some tinkering, it looks like `listSubPages` doesn't work at both parent and child level simultaneously - when i added it for the `session` sub-pages nested inside the existing Modules `listSubPages` (line 11), the `session` sub-pages didn't appear in the sidebar. `listSubPages` seems to ignore children of a child.

When I remove `listSubPages` from the outer modules entry and use it at the individual module-level instead, the session sub-pages show up in the sidebar. The downside is that other modules need to be manually listed. This is what's working for now.

I also explored using `depth: 2` and `nested: true` for auto listing the modules and module sub-pages. This produced `session` without a collapsible menu, which means the same would happen for other module sub-pages - all child pages of all modules will be displayed at once. So this was a no-go.

While I was at it, I made a couple of other changes to the sidebar:
- Changed "BiDi" and "Classic" into section headings to reduce the nesting depth
- Added a "Shared" section heading for capabilities and errors (primarily because currently our sidebar indentation doesn't make it visually clear that these are common to BiDi and Classic)

dipikabh:
Thanks for the thorough review, Chris! Lots of food for thought.
I'll tackle the remaining comments in `session/new` tomorrow.

files/en-us/web/webdriver/reference/bidi/modules/session/status/index.md
whimboo: The implementation of the `session.status` command you can find here:
https://searchfox.org/firefox-main/rev/e67dad5c526f22744f22e5e1173213c9c5811df8/remote/webdriver-bidi/WebDriverBiDi.sys.mjs#204

As it can be seen the only factor here is the number of active sessions. While other browsers like Chrome already support multiple session (even they are [not spec'ed yet](https://github.com/w3c/webdriver-bidi/issues/103)), we do not. For more details on the readiness state see the WebDriver classic specification at: https://w3c.github.io/webdriver/#dfn-readiness-state. I hope that helps.

files/en-us/web/webdriver/reference/bidi/modules/session/index.md
chrisdavidmills: Thanks for the explanation, makes sense. I must admit that I mostly use hand-crafted DLs to avoid such repetition issues. I think this is fine.

files/en-us/web/webdriver/reference/bidi/modules/emulation/index.md
chrisdavidmills: Perfect.

files/sidebars/webdriver.yaml
chrisdavidmills: Good work. It is a little fiddly getting those menus right, but what you've done here looks great.

files/en-us/web/webdriver/reference/bidi/modules/index.md
chrisdavidmills: Your explanation makes sense, thanks!

I am looking forward to seeing more guides/how-tos published. It's awesome to see they are on the way.

files/en-us/web/webdriver/reference/bidi/modules/index.md
chrisdavidmills: Perfect, this makes sense.

files/en-us/web/webdriver/reference/bidi/modules/session/end/index.md
chrisdavidmills: Cool, this all makes sense then. Thanks for the responses.

files/sidebars/webdriver.yaml
dipikabh: Awesome, thanks for checking

files/en-us/web/webdriver/reference/bidi/modules/index.md
dipikabh: Definitely, a few guides would really help to connect all the dots

files/en-us/web/webdriver/reference/bidi/modules/session/index.md
caugner: Afaik the old Compat macro had this option, but this was removed at some point. I could imagine hiding nested features where an MDM page exists, i.e. because there is a page about `session.new`, we can omit the subfeatures on the module page, whereas there is no page for the capabilities feature, so we shouldn't hide that row.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: Perhaps I can queue up this task for a future date

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: Hmm..

what do you think if it is said in this way:

> Since this command is used to create a new session, it runs without an already active session. (In BiDi, such a command is called a static command.)

and on https://pr43251.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/status page as:

> Since this command is used to check the status of the browser before creating a new session, it runs without an already active session. (In BiDi, such a command is called a static command.)

Or do you think it is really a nice-to-know piece of info and should be removed?

BTW, I also have this one-liner in the https://pr43251.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#commands section:

> All commands except `session.new` and `session.status` require an active WebDriver BiDi session.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: `alwaysMatch` and so "match" is WebDriver-wide - used both in Classic and BiDi.

Nonetheless, the description doesn't need to use/repeat the word "match" from the capability name and can be simplified. So thanks for flagging.

I also felt there is scope for improving `firstMatch`. I ended up re-writing several paras, specifically lines [30-42](https://github.com/mdn/content/pull/43251/changes#diff-868a56936c614ad689845f49103fd2c95b800a8e8deafb60504dd2fd4d33201fR30-R42). 
In the `capabilities` description, I've clarified that these are features requested by the client. I've rephrased `alwaysMatch` and `firstMatch` to make the descriptions parallel and to highlight the contrast.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: I personally prefer a direct phrasing like "that contains". :)
I've updated other instances as well.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: Good catch! Added the link and also expanded the description:

> A boolean that indicates whether the browser window can be resized and repositioned using the [Set Window Rect](https://pr43251.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/Classic/Commands/SetWindowRect) command.

files/en-us/web/webdriver/reference/bidi/modules/session/status/index.md
dipikabh: Thanks for the links, Henrik!

Chris, I think currently there's no specific list of situations for when session creation would fail and it's very implementation-specific. Perhaps this can be revised if/when there's more info to add here.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: I plan to create dedicated pages for some of these capabilities (similar to the ones we have [here](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/Capabilities#list_of_capabilities)), and so those pages can cover the details

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: From Firefox's [Capabilities.sys.mjs](https://searchfox.org/firefox-main/rev/e67dad5c526f22744f22e5e1173213c9c5811df8/remote/shared/webdriver/Capabilities.sys.mjs) as an example, I'm getting the sense that the values are defined per browser; they're certainly not defined in the spec.
Hopefully I can dig deeper when I create the capability page.

chrisdavidmills: OK

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
chrisdavidmills: This now reads much better — great work!

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
chrisdavidmills: I think you are overthinking it a little, and I prefer the original, albeit with my suggestion ;-)

But, I don't think this is a big deal. I'm not going to block on this.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
chrisdavidmills: Nice, I like that plan.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
chrisdavidmills: WFM

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
chrisdavidmills: Perfect!

files/en-us/web/webdriver/reference/bidi/modules/session/status/index.md
chrisdavidmills: Yeah, this is fine. Again, I'm not going to block on this, and we can always iterate later on if needed.

chrisdavidmills (approved):
OK, I'm pretty happy with this now, @dipikabh, so I'm approving from an editorial standpoint.

You've done a great job on this first set of documentation for WebDriver BiDi. It's obviously an unknown for us on the MDN team, and there will likely be things you'll want to improve on later after you've had some wider feedback. But for now, I think this is more than good enough to publish.

Good work.

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: overthinking - as always :)
Gotcha - i'll revise it in one of my future PRs

files/en-us/web/webdriver/reference/bidi/modules/session/new/index.md
dipikabh: Thanks for checking

dipikabh:
Thanks so much, @chrisdavidmills! Your feedback has helped me think through the phrasing and details to look out for. It'll definitely help as I work through more BiDi pages.
Really appreciate the thoughtful review.

> there will likely be things you'll want to improve on later after you've had some wider feedback.

Agree, that's always the case

PR #43253, opened 2026-02-26
TLS guide update

wbamberg:
We want to refer to the TLS guide quite often, but the current guide could be improved. Currently it talks quite a lot about the history of TLS and how it's implemented internally, which I don't think is super relevant to people using it.

I've removed some of this and added:
- a section on server authentication
- a section on mixed content, that just summarises the situation and links to the [mixed content](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Mixed_content) page
- a section on upgrading HTTP connections, HSTS etc: this last copies a lot of material from the https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/MITM guide, as I think it sits better here. If you agree I'll update that page to point to this one.

Originally I was going to integrate the [mixed content](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Mixed_content) page into here, and still think that might be worth doing either in this PR or a follow up. But I was put off doing that partly by the fact that that page includes a BCD table and it feels off to include that in a page on TLS.

https://github.com/mdn/content/pull/43253/commits/2fdb0fb8c77c40040623b96c25d32e16fc2d386d removes some links to the "TLS 1.3" section in the old page: these are related to early data. While it's true that early data is a feature specific to TLS 1.3, the old page didn't actually mention it, so the links didn't serve to explain it anyway.

### DIFF
--- a/files/en-us/web/security/defenses/transport_layer_security/index.md
+++ b/files/en-us/web/security/defenses/transport_layer_security/index.md
@@ -47,7 +47,7 @@ Modern web hosting services support HTTPS for you, either by default or through 
 
 ## Mixed content
 
-A website should use HTTPS not only for the main document, but also for all subresources that it loads, such as scripts, stylesheets, images, and fonts. If a website loads the main document over HTTPS but then loads any of its subresources over HTTP, this is called _mixed content_.
+A website should use HTTPS not only for the main document, but also for all subresources that it loads, such as scripts, stylesheets, images, and fonts. If a website loads the main document over HTTPS but then loads any of its subresources over HTTP, this is called [mixed content](/en-US/docs/Web/Security/Defenses/Mixed_content).
 
 For example, if a document served from `https://example.org` includes the following content, then it will constitute mixed content:
 


files/en-us/web/http/reference/status/425/index.md
```diff
@@ -8,7 +8,7 @@ sidebar: http
 
 The HTTP **`425 Too Early`** [client error response](/en-US/docs/Web/HTTP/Reference/Status#client_error_responses) status code indicates that the server was unwilling to risk processing a request that might be replayed to avoid potential replay attacks.
 
-If a client has interacted with a server recently, early data (also known as zero round-trip time [(0-RTT) data](/en-US/docs/Web/Security/Defenses/Transport_Layer_Security#tls_1.3)) allows the client to send data to a server in the first round trip of a connection, without waiting for the TLS [handshake](/en-US/docs/Glossary/TCP_handshake) to complete.
+If a client has interacted with a server recently, early data (also known as zero round-trip time (0-RTT) data) allows the client to send data to a server in the first round trip of a connection, without waiting for the TLS [handshake](/en-US/docs/Glossary/TCP_handshake) to complete.
 A client that sends a request in early data does not need to include the `Early-Data` header.
 See {{HTTPHeader("Early-Data")}} for more information.
```
hamishwillee, line 13: Note on you, but it isn't clear to me what this paragraph has to do with 425 too early error (from this text)

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
...
 
-The removal of renegotiation in TLS 1.3 might affect some web servers that rely on client authentication using certificates. Some web servers use renegotiation to either ensure that client certificates are encrypted, or to request client certificates only when certain resources are requested. For the privacy of client certificates, the encryption of the TLS 1.3 handshake ensures that client certificates are encrypted; however this might require some software changes. Reactive client authentication using certificates is supported by TLS 1.3 but not widely implemented. Alternative mechanisms are in the process of being developed, which will also support HTTP/2.
+To reduce the risk of this attack, the server should also send the {{httpheader("Strict-Transport-Security")}} HTTP response header (also known as HSTS): this informs clients that the sites wishes them to use HTTPS, and will cause the browser to connect using HTTPS directly for any subsequent visits, even those made using HTTP URLs.
```
hamishwillee:
```suggestion
To reduce the risk of this attack, the server should also send the {{httpheader("Strict-Transport-Security")}} HTTP response header (also known as HSTS): this informs clients that the site wishes them to use HTTPS, and will cause the browser to connect using HTTPS directly for any subsequent visits, even those made using HTTP URLs.
```

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
...
-
-## TLS handshake timeout values
-
-If the TLS handshake starts to become slow or unresponsive for some reason, the user's experience can be affected significantly. To mitigate this problem, modern browsers have implemented handshake timeouts:
-
-- Since version 58, Firefox implements a TLS handshake timeout with a default value of 30 seconds. The timeout value can be varied by editing the `network.http.tls-handshake-timeout` pref in about:config.
+With HSTS, SSL stripping is prevented except for the first time the browser tries to connect to your site (or, since HSTS has an expiry, the first time after an HSTS record in the browser has expired). To protect the site even on first connection or HSTS record expiry, Chrome maintains a list of domains called the [HSTS preload list](https://hstspreload.org/): if a domain is on this list, then Chrome will always upgrade HTTP request to HTTPS, effectively behaving as if the server has already sent the HSTS header. Safari and Firefox have similar behavior, using a list that is derived from the Chrome list.
```
hamishwillee:
```suggestion
With HSTS, SSL stripping is prevented except for the first time the browser tries to connect to your site (or, since HSTS has an expiry, the first time after an HSTS record in the browser has expired). To protect the site even on first connection or HSTS record expiry, Chrome maintains a list of domains called the [HSTS preload list](https://hstspreload.org/): if a domain is on this list, then Chrome will always upgrade HTTP requests to HTTPS, effectively behaving as if the server has already sent the HSTS header. Safari and Firefox have similar behavior, using a list that is derived from the Chrome list.
```

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
...
-TLS provides three primary services that help ensure the safety and security of data exchanged with it:
+When a client connects to a server using TLS, an initial _handshake_ sets the security parameters for the protocol:
 
-- Authentication
-  - : Authentication lets each party to the communication verify that the other party is who they claim to be.
-- Encryption
-  - : Data is encrypted while being transmitted between the user agent and the server, in order to prevent it from being read and interpreted by unauthorized parties.
-- Integrity
-  - : TLS ensures that between encrypting, transmitting, and decrypting the data, no information is lost, damaged, tampered with, or falsified.
+- Client and server agree on which version of TLS to use. The current version of TLS is 1.3 ({{RFC(8446)}}), and this is the most widely-used version, although TLS 1.2 is still used in some websites. TLS 1.1 and 1.0 should no longer be used.
```
hamishwillee:
```suggestion
- Client and server agree on which version of TLS to use. The current version of TLS is 1.3 ({{RFC(8446)}}), and this is the most widely-used version. TLS 1.2 is still used by some websites, and TLS 1.1 and 1.0 should no longer be used.
```

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
...
 
-- Remove unused and unsafe features of TLS 1.2.
-- Include strong security analysis in the design.
-- Improve privacy by encrypting more of the protocol.
-- Reduce the time needed to complete a handshake.
+For this reason, browsers don't allow secure pages to load insecure subresources. Instead, depending on the type of subresource, they either upgrade the load request to use HTTPS, or block the request entirely.
 
-TLS 1.3 changes much of the protocol fundamentals, but preserves almost all of the basic capabilities of previous TLS versions. For the web, TLS 1.3 can be enabled without affecting compatibility with some rare exceptions (see below).
+If it's not possible for you to update your code to load resources from HTTPS URLs (for example, because your HTML has been archived) your server can set a [content security policy](/en-US/docs/Web/HTTP/Guides/CSP) that contains the [`upgrade-insecure-requests`](/en-US/docs/Web/HTTP/Guides/CSP#upgrading_insecure_requests) directive, and the browser will automatically upgrade these requests to HTTPS.
```
hamishwillee:
```suggestion
If it's not possible for you to update your code to load resources from HTTPS URLs (for example, because your HTML has been archived), your server can set a [content security policy](/en-US/docs/Web/HTTP/Guides/CSP) that contains the [`upgrade-insecure-requests`](/en-US/docs/Web/HTTP/Guides/CSP#upgrading_insecure_requests) directive, and the browser will automatically upgrade these requests to HTTPS.
```

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
...
 
-To assist you in configuring your site, Mozilla provides a helpful [TLS configuration generator](https://ssl-config.mozilla.org/) that will generate configuration files for the following Web servers:
+## Mixed content
 
-- Apache
-- Nginx
-- Lighttpd
-- HAProxy
-- Amazon Web Services CloudFormation Elastic Load Balancer
+A website should use HTTPS for not only the main document but all subresources that it loads, such as scripts, stylesheets, images, and fonts. If a website loads the main document over HTTPS but then loads any of its subresources over HTTP, this is called _mixed content_.
```
hamishwillee: If you're not moving mixed content in (and I wouldn't) maybe link here.

```suggestion
A website should use HTTPS not only for the main document, but also for all subresources that it loads, such as scripts, stylesheets, images, and fonts. If a website loads the main document over HTTPS but then loads any of its subresources over HTTP, this is called _mixed content_.
```

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
...
+For example, if a document served from `https://example.org` includes the following content, then it will constitute mixed content:
 
-## TLS 1.3
+```html
+<img src="http://example.org/my-image.png" />
+```
 
-{{RFC("8446", "TLS 1.3")}} is a major revision to TLS. TLS 1.3 includes numerous changes that improve security and performance. The goals of TLS 1.3 are:
+Mixed content is unsafe because the subresources don't get the protection that HTTPS offers, so an attacker can not only read them but potentially modify them, and this can undermine the integrity of the page as a whole. For example, we could imagine an attacker modifying a script to behave harmfully. Other resources are less dangerous than scripts but still potentially dangerous: for example, an attacker could modify images so as to confuse or mislead users.
```
hamishwillee:
```suggestion
Mixed content is unsafe because the subresources don't get the protection that HTTPS offers, so an attacker can not only read them but potentially also modify them. This can undermine the integrity of the page as a whole! For example, we could imagine an attacker modifying a script to behave harmfully. Other resources are less dangerous than scripts, but still potentially dangerous: for example, an attacker could modify images so as to confuse or mislead users.
```

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
 sidebar: security
 ---
 
-The security of any connection using Transport Layer Security (TLS) is heavily dependent upon the cipher suites and security parameters selected. This article's goal is to help you make these decisions to ensure the confidentiality and integrity of communication between client and server. The Mozilla Operations Security (OpSec) team [maintains a wiki entry](https://wiki.mozilla.org/Security/Server_Side_TLS) with reference configurations for servers.
+Transport Layer Security (TLS) is a protocol which enables a client to communicate securely with a server across an untrusted network. Most notably, on the web, it's used to secure HTTP connections: the resulting protocol is called {{glossary("HTTPS")}}.
```
hamishwillee: IMO 
```suggestion
Transport Layer Security (TLS) is a protocol which enables a client to communicate securely with a server across an untrusted network. Most notably it's used to secure HTTP connections on the web: the resulting protocol is called {{glossary("HTTPS")}}.
```

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
...
 
-The Transport Layer Security (TLS) protocol is the standard for enabling two networked applications or devices to exchange information privately and robustly. Applications that use TLS can choose their security parameters, which can have a substantial impact on the security and reliability of data. This article provides an overview of TLS and the kinds of decisions you need to make when securing your content.
+TLS secures a network connection in three ways:
 
-## History
+- **Encryption**: the data exchanged between client and server is encrypted while in transit, so it can't be read by any attackers.
+- **Integrity**: an attacker can't, without detection, modify data while it is in transit between client and server.
```
hamishwillee: Maybe


```suggestion
- **Integrity**: an attacker can't secretly modify data (without detection) while it is in transit between client and server.
```

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
...
+In particular, HTTPS is the defense against a [manipulator in the middle (MITM)](/en-US/docs/Web/Security/Attacks/MITM) attack, in which the attacker inserts themselves between the user's browser and the server they are connecting to, and can read and modify the traffic exchanged.
 
-Despite the fact that the web now uses TLS for encryption, many people still refer to it as "SSL" out of habit.
+Browsers consider pages delivered over HTTPS as providing a [secure context](/en-US/docs/Web/Security/Secure_Contexts). Many powerful web APIs are only available to code running in a secure context.
 
-Although TLS can be used on top of any low-level transport protocol, the original goal of the protocol was to encrypt HTTP traffic. HTTP encrypted using TLS is commonly referred to as {{Glossary("HTTPS")}}. TLS-encrypted web traffic is by convention exchanged on port 443 by default, while unencrypted HTTP uses port 80 by default. HTTPS remains an important use case for TLS.
+**All websites should serve all their pages and subresources over HTTPS, and implement server authentication.**
```
hamishwillee, line 21: Perhaps important enough to be an aside.

files/en-us/web/security/defenses/transport_layer_security/index.md
```diff
@@ -6,97 +6,78 @@ page-type: guide
...
 
-To assist you in configuring your site, Mozilla provides a helpful [TLS configuration generator](https://ssl-config.mozilla.org/) that will generate configuration files for the following Web servers:
+## Mixed content
 
-- Apache
-- Nginx
-- Lighttpd
-- HAProxy
-- Amazon Web Services CloudFormation Elastic Load Balancer
+A website should use HTTPS not only for the main document, but also for all subresources that it loads, such as scripts, stylesheets, images, and fonts. If a website loads the main document over HTTPS but then loads any of its subresources over HTTP, this is called _mixed content_.
```
wbamberg:
```suggestion
A website should use HTTPS not only for the main document, but also for all subresources that it loads, such as scripts, stylesheets, images, and fonts. If a website loads the main document over HTTPS but then loads any of its subresources over HTTP, this is called [mixed content](/en-US/docs/Web/Security/Defenses/Mixed_content).
```

wbamberg:
Thank you for the quick review!

PR #1134, opened 2021-01-09
fix: typo in repeat CSS function docs

dbgb:
Just a quick typo fix! 🧐🔎  

"independant" -> "independent" 👍

PR #43268, opened 2026-02-27
Fix DOMPurify usage

xmcgcg:
Many examples use `DOMPurify.sanitize()` to create Trusted Type policies without unsetting the `RETURN_TRUSTED_TYPE` flag, in which case this method might return a `TrustedHTML` object directly. DOMPurify explicitly [requires](https://github.com/cure53/DOMPurify?tab=readme-ov-file#what-about-dompurify-and-trusted-types) unsetting the `RETURN_TRUSTED_TYPE` flag during policy creation.

Unlike [`TrustedTypePolicy.createHTML()`](https://developer.mozilla.org/en-US/docs/Web/API/TrustedTypePolicy/createHTML), the `createHTML` method of the policy option argument of [`TrustedTypePolicyFactory.createPolicy()`](https://developer.mozilla.org/en-US/docs/Web/API/TrustedTypePolicyFactory/createPolicy) should return a string instead of a `TrustedHTML` object.

This PR fixes all such usages, with some minor changes:
* Unifies all policy names to `"my-policy"`.
* Adds missing links to DOMPurify references.
* Replaces the `SAFE_FOR_JQUERY` flag in an example. The flag is already [removed](https://github.com/cure53/DOMPurify?tab=readme-ov-file#removed-configuration).

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
@@ -82,7 +82,9 @@ You can use any method to add the sanitized HTML to your DOM, for example jQuery
 
 ```js
 let elem = $("<div/>");
-let cleanHTML = DOMPurify.sanitize(externalHTML, { RETURN_TRUSTED_TYPE: false });
+let cleanHTML = DOMPurify.sanitize(externalHTML, {
+  RETURN_TRUSTED_TYPE: false,
+});
 elem.html(cleanHTML);
 ```
 
--- a/files/en-us/web/api/domparser/parsefromstring/index.md
+++ b/files/en-us/web/api/domparser/parsefromstring/index.md
@@ -107,7 +107,8 @@ Commonly, implementations of `createHTML()` use a library such as [DOMPurify](ht
 
 ```js
 const policy = trustedTypes.createPolicy("my-policy", {
-  createHTML: (input) => DOMPurify.sanitize(input, { RETURN_TRUSTED_TYPE: false }),
+  createHTML: (input) =>
+    DOMPurify.sanitize(input, { RETURN_TRUSTED_TYPE: false }),
 });
 ```
 
--- a/files/en-us/web/api/element/innerhtml/index.md
+++ b/files/en-us/web/api/element/innerhtml/index.md
@@ -108,7 +108,8 @@ Commonly implementations of `createHTML()` use a library such as [DOMPurify](htt
 
 ```js
 const policy = trustedTypes.createPolicy("my-policy", {
-  createHTML: (input) => DOMPurify.sanitize(input, { RETURN_TRUSTED_TYPE: false }),
+  createHTML: (input) =>
+    DOMPurify.sanitize(input, { RETURN_TRUSTED_TYPE: false }),
 });
 ```
 
--- a/files/en-us/web/api/element/outerhtml/index.md
+++ b/files/en-us/web/api/element/outerhtml/index.md
@@ -138,7 +138,8 @@ Commonly implementations of `createHTML()` use a library such as [DOMPurify](htt
 
 ```js
 const policy = trustedTypes.createPolicy("my-policy", {
-  createHTML: (input) => DOMPurify.sanitize(input, { RETURN_TRUSTED_TYPE: false }),
+  createHTML: (input) =>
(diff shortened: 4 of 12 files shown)


files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
xmcgcg, line 74: This example is unchanged, because `innerHTML` can accept `TrustedHTML` objects.

files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
```diff
@@ -78,11 +78,11 @@ let cleanHTML = DOMPurify.sanitize(externalHTML);
 elem.innerHTML = cleanHTML;
 ```
```
xmcgcg, line 80: jQuery's [`html()`](https://api.jquery.com/html/#html2) method doesn't accept `TrustedHTML` objects, so `RETURN_TRUSTED_TYPE` needs to be unset here.

hamishwillee:
@xmcgcg Did you **test** that setting `RETURN_TRUSTED_TYPE` to false is needed?

The text in https://github.com/cure53/DOMPurify?tab=readme-ov-file#what-about-dompurify-and-trusted-types is slightly ambiguous because the first part implies that `RETURN_TRUSTED_TYPE` is false by default when trusted types are enabled. The requirement could therefore be that it IS false, not that it is SET false - if you see what I mean.

Obviously being explicit is never going to be wrong but it is better to have minimum code, so if the flage is not required I'd leave it out and just add a note about this.

files/en-us/mozilla/add-ons/webextensions/safely_inserting_external_content_into_a_page/index.md
```diff
@@ -78,11 +78,13 @@ let cleanHTML = DOMPurify.sanitize(externalHTML);
...
 
-You can use any method to add the sanitized HTML to your DOM, for example jQuery's `.html()` function. Remember though that the `SAFE_FOR_JQUERY` flag needs to be used in this case:
+You can use any method to add the sanitized HTML to your DOM, for example jQuery's `.html()` function. Remember though that the `RETURN_TRUSTED_TYPE` flag needs to be unset in this case:
```
rebloor, line 81: The term "unset" might be unclear to readers. I suggest being more explicit:

```suggestion
You can use any method to add the sanitized HTML to your DOM, for example, jQuery's `.html()` function. Remember, though, that the `RETURN_TRUSTED_TYPE` flag needs to be set to `false` in this case:
```

hamishwillee: @rebloor As per https://github.com/mdn/content/pull/43268#issuecomment-3981235007 I'm not sure it really needs to be - it may be that this is really "must not be set to `true`". 

Unless you are of the opinion that it is safer to always unset this, I am hoping that this actually defaults to `false` so we don't modify the code everywhere. That makes far more sense to me as an implementation for DOMPurify.

xmcgcg:
@hamishwillee I checked the implementation of [`DOMPurify.sanitize()`](https://github.com/cure53/DOMPurify/blob/302b51de22535cc90235472c52e3401bedd46f80/src/purify.ts#L1516) yesterday. `RETURN_TRUSTED_TYPE` is `false` by default, and the method has a surprising behavior:
* If `DOMPurify.setconfig()` is never called before, calling `DOMPurify.sanitize(input, cfg)` will consider `cfg`.
* If `DOMPurify.setconfig(globalCfg)` is called before without any interleaving `DOMPurify.clearconfig()`, calling `DOMPurify.sanitize(input, cfg)` will **ignore** `cfg`. None of the flags in `cfg` will override those in `globalCfg`.

Therefore `DOMPurify.sanitize(input, { RETURN_TRUSTED_TYPE: false })` always has the same behavior as `DOMPurify.sanitize(input)`, regardless of any prior `DOMPurify.setconfig()` call.

The author is not planning for a fix on that (see [DOMPurify issue #1134](https://github.com/cure53/DOMPurify/issues/1134)) , so I will close this PR, and submit another one fixing `SAFE_FOR_JQUERY` only.

hamishwillee:
Thanks very much for taking the time to tess this fully!

PR #43270, opened 2026-02-27
Remove duplicated content from MITM guide

wbamberg:
https://github.com/mdn/content/pull/43253 added TLS guidance to the TLS guide, some of which was taken from the [MITM page](https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/MITM). This PR removes that content from the MITM page, replacing it with pointers to the TLS guide.

I should really have done this in https://github.com/mdn/content/pull/43253, sorry.


Samuel330ghjkm:
> https://github.com/mdn/content/pull/43253 added TLS guidance to the TLS guide, some of which was taken from the [MITM page](https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/MITM). This PR removes that content from the MITM page, replacing it with pointers to the TLS guide.
> 
> 
> 
> I should really have done this in https://github.com/mdn/content/pull/43253, sorry.

Issue #43226, opened 2026-02-23
Including the HTML Sanitizer API's built-in safe default configuration


evilpie:
The HTML Sanitizer API has a ["built-in safe default configuration"](https://wicg.github.io/sanitizer-api/#built-in-safe-default-configuration). This data might be useful to show to MDN users in a nicer representation than a JSON blob. I have an experimental Markdown representation of the data here: https://gist.github.com/evilpie/8c54c6d3c31447477189fc0891b67b7d

I really don't know where to put this and how to properly integrate this with the rest of the documentation. The Sanitizer API is shipping this week in Firefox. @hamishwillee I know you wrote a lot of the original documentation for the Sanitizer, is this something you could help me with?

Josh-Cena:
I think we've discussed it last time and decided that the exact implementation is uninteresting to developers and is subject to change, so the following paragraph should suffice:

> The default Sanitizer allows only XSS-safe input by default, omitting elements such as [`<script>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script), [`<frame>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/frame), [`<iframe>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe), [`<object>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/object), `<use>`, and event handler attributes from their respective allow lists, and disallowing data attributes, and comments. It is created if `"default"` or no object is passed to the constructor.

evilpie:
That isn't really true, what you seem to be describing is the element/attributes that are "unsafe" and always removed by removeUnsafe and part of https://wicg.github.io/sanitizer-api/#built-in-safe-baseline-configuration.

For example an `<img>`  is not an XSS vector (ignoring event handlers), but isn't part of the "safe default configuration" either. So a Sanitizer initialized with `"default"` wouldn't allow it by default.

Something being subject to change shouldn't be a hindrance to documenting it. There are probably only going to be changes when new elements or attributes are added.

hamishwillee:
@evilpie Thanks very much for getting in touch.

[...]

PR #43271, opened 2026-02-27
Add default santizer config

hamishwillee:
This add information on the `Sanitizer` default configuration and cross links it where needed.
@wbamberg @evilpie - would appreciate your review.

The information is specified in two places.
1. An overview in the API overview page. This provides the gist and links to ...
2. A separate guide page with a slightly expanded explanation, and a listing of the values.

I chose to do it in two docs , because the "full listing" is too much information for most users. I chose to put the "summary version" in the overview guide rather than in `Sanitizer()` because it is relevant to everything and worth knowing up front.

In addition, I made lots of minor spot changes that reflected a long-standing lack of understanding (mine) of the purpose of the default config vs the baseline remove config.

Essentially the default config removes more than just XSS-unsafe stuff - so this is more restrictive than what you get if you start with a custom config and call removeUnsafe on it. 
The implication of the docs before this chang was that removeUnsafe would get you back to the default config.
The changes might still allow you to infer that in some cases, but I think I've been pretty clear that they are different now.

Related docs work can be tracked in https://github.com/mdn/content/issues/42755

Fixes https://github.com/mdn/content/issues/43226

### DIFF
--- a/files/en-us/web/api/element/sethtml/index.md
+++ b/files/en-us/web/api/element/sethtml/index.md
@@ -31,9 +31,9 @@ setHTML(input, options)
     - `sanitizer`
       - : A {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} object which defines what elements of the input will be allowed or removed, or the string `"default"` for the default configuration.
         The method will remove any XSS-unsafe elements and attributes, even if allowed by the sanitizer.
-        If not specified, the [default `Sanitizer` configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#default_sanitizer_configuration) is used.
+        If not specified, the [default sanitizer configuration](/en-US/docs/Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration) is used.
 
-        Note that if you're using the same configuration multiple times, it's expected to be more efficient to use a `Santitizer` and modify it when you need to.
+        Note that if you're using the same configuration multiple times, it's expected to be more efficient to use a `Sanitizer` and modify it when you need to.
 
 ### Return value
 
@@ -55,7 +55,7 @@ The **`setHTML()`** method provides an XSS-safe method to parse and sanitize a s
 `setHTML()` drops any elements in the HTML input string that are invalid in the context of the current element, such as a {{htmlelement("col")}} element outside of a {{htmlelement("table")}}.
 It then removes any HTML entities that aren't allowed by the sanitizer configuration, and further removes any XSS-unsafe elements or attributes — whether or not they are allowed by the sanitizer.
 
-If no sanitizer is specified in the `options.sanitizer` parameter, `setHTML()` is used with the [default `Sanitizer` configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#default_sanitizer_configuration).
+If no sanitizer is specified in the `options.sanitizer` parameter, `setHTML()` is used with the [default sanitizer configuration](/en-US/docs/Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration).
 This configuration is suitable for the majority of use cases as it prevents XSS attacks, as well as other attacks like clickjacking or spoofing.
 
 A custom `Sanitizer` or `SanitizerConfig` can be specified to choose which elements, attributes, and comments are allowed or removed.
--- a/files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
+++ b/files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
@@ -1,28 +1,30 @@
 ---
-title: Default Sanitizer configuration
+title: Default sanitizer configuration
 slug: Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration
 page-type: guide
 ---
 
 {{DefaultAPISidebar("HTML Sanitizer API")}}
 
-The default sanitizer configuration defines the {{domxref("Sanitizer")}} that is used by default if you call the [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods), such as {{domxref("Element.setHTML()")}}, {{domxref("ShadowRoot.setHTML()")}}, and {{domxref("Document.parseHTML_static","Document.parseHTML()")}}, without specifying a custom sanitizer.
-It is also the default configuration returned by the [`Sanitizer()` constructor](/en-US/docs/Web/API/Sanitizer/Sanitizer) if no `configuration` is passed as an argument.
+The default sanitizer configuration defines the configuration returned by the [`Sanitizer()` constructor](/en-US/docs/Web/API/Sanitizer/Sanitizer) if no `configuration` is passed as an argument.
+This same configuration is implicitly used if you call the [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods), such as {{domxref("Element.setHTML()")}}, {{domxref("ShadowRoot.setHTML()")}}, and {{domxref("Document.parseHTML_static","Document.parseHTML()")}}, without specifying a custom sanitizer.
 
 This configuration removes the following sorts of items:
 
 1. Those that are known to be XSS-unsafe:
    - {{htmlelement("embed")}}, {{htmlelement("frame")}}, {{htmlelement("iframe")}}, {{htmlelement("object")}}, {{htmlelement("script")}}, and {{SVGElement("use")}}.
    - All event handler content attributes, such as `onafterprint`, `onbeforeinput`, and so on.
-2. Additional items that might be be used in clickjacking, spoofing, or other attacks.
+2. Additional items that might be used in clickjacking, spoofing, or other attacks.
 3. Comments and `data-*` attributes.
 
 It therefore provides a sanitizer with a minimal attack surface, which is still suitable for the majority of sanitization use cases.
 
 > [!NOTE]
-> The specification calls this configuration the [built-in safe default configuration](https://wicg.github.io/sanitizer-api/#built-in-safe-default-configuration).
+> Calling {{domxref("Sanitizer.removeUnsafe()")}}, or passing a custom sanitizer to the safe sanitization method, only removes the XSS-unsafe items.
+> It does not remove the additional items, comments, and `data-*` attributes.
 
-The following sections list all the elements, with a &check; mark indicating those that are allowed by the default configuration, and also listing the attributes that are allowed only for particular elements.
+The following sections list all the elements, with a &check; mark indicating those that are _allowed_ by the default configuration (the ❌ therefore indicates those that will be removed).
+The "Additional allowed attributes" column lists the attributes that are allowed for the corresponding elements; any other attributes on the element would be removed (unless allowed by the global attributes).
 The [Global attributes](#global_attributes) section lists the attributes that are allowed on all elements (the attributes that are not removed when the configuration is used).
 
 ## HTML elements
--- a/files/en-us/web/api/html_sanitizer_api/index.md
+++ b/files/en-us/web/api/html_sanitizer_api/index.md
@@ -24,15 +24,17 @@ These come in XSS-safe versions that enforce removal of all unsafe elements and 
 
 The HTML Sanitizer API provides XSS-safe and XSS-unsafe methods for injecting HTML strings into an {{domxref('Element')}} or a {{domxref('ShadowRoot')}}, and for parsing HTML into a {{domxref('Document')}}.
 
-- Safe methods: {{domxref('Element.setHTML()')}}, {{domxref('ShadowRoot.setHTML()')}}, and {{domxref('Document/parseHTML_static','Document.parseHTML()')}}.
-- Unsafe methods: {{domxref('Element.setHTMLUnsafe()')}}, {{domxref('ShadowRoot.setHTMLUnsafe()')}}, and {{domxref('Document/parseHTMLUnsafe_static','Document.parseHTMLUnsafe()')}}.
+- Safe methods
+  - : {{domxref('Element.setHTML()')}}, {{domxref('ShadowRoot.setHTML()')}}, and {{domxref('Document/parseHTML_static','Document.parseHTML()')}}.
+- Unsafe methods
+  - : {{domxref('Element.setHTMLUnsafe()')}}, {{domxref('ShadowRoot.setHTMLUnsafe()')}}, and {{domxref('Document/parseHTMLUnsafe_static','Document.parseHTMLUnsafe()')}}.
 
 All the methods take the HTML to be injected and an optional {{domxref("Sanitizer")}} or {{domxref("SanitizerConfig")}} as arguments.
 The sanitizer objects defines the HTML entities that will be filtered out of the input before it is injected.
 The {{domxref('Element')}} methods are context aware, and will additionally drop any elements that the HTML specification does not allow in the target element.
 
 The safe methods always remove XSS-unsafe elements and attributes.
-If no sanitizer is passed as a parameter they will use the [default `Sanitizer` configuration](#default_sanitizer_configuration), which removes both XSS-unsafe elements and attributes, such as {{htmlelement("script")}} elements and `onclick` event handlers, along with others that might be be used in other kinds of attacks if provided as user input.
+If no sanitizer is passed as a parameter they will use the [default sanitizer configuration](#default_sanitizer_configuration), which removes both XSS-unsafe elements and attributes, such as {{htmlelement("script")}} elements and `onclick` event handlers, along with others that might be used in other kinds of attacks if provided as user input.
 If a custom sanitizer is used with a safe method, it is implicitly updated to remove any elements and attributes that are not XSS-safe (note that the passed sanitizer is not modified, and might still allow unsafe entities if used with an unsafe method).
 
 The safe methods should be used instead of {{domxref("Element.innerHTML")}}, {{domxref("Element.outerHTML")}}, or {{domxref("ShadowRoot.innerHTML")}}, for injecting untrusted HTML content.
@@ -54,7 +56,7 @@ sanitizer.allowAttribute("onblur"); // Allow onblur
 someElement.setHTMLUnsafe(untrustedString, { sanitizer });
 ```
 
-### Sanitizer configuration
+### Sanitizer configurations
 
 A sanitizer configuration defines what HTML entities will be allowed, replaced, or removed when the sanitizer is used, including elements, attributes, `data-*` attributes, and comments.
 
@@ -193,23 +195,33 @@ So if you call `allowElement()` on an allow configuration and the specified elem
 But if the element is already present then the method would return `false`.
 Note that if you call the same method to set a per-element attribute, this will return `false` if called on a remove sanitizer, because the change cannot be made.
 
-#### Default `Sanitizer` configuration
+### Built-in configurations
 
-The default `Sanitizer` configuration defines the {{domxref("Sanitizer")}} that is used if you call {{domxref("Element.setHTML()")}} or the other [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods) without a custom sanitizer.
+#### XSS-safe baseline configuration
+
+The XSS-safe baseline configuration defines the items that must be removed from an input in order to make it XSS-safe:
+
+- {{htmlelement("embed")}}, {{htmlelement("frame")}}, {{htmlelement("iframe")}}, {{htmlelement("object")}}, {{htmlelement("script")}}, and {{SVGElement("use")}}.
+- All event handler content attributes, such as `onafterprint`, `onbeforeinput`, and so on.
+
+The configuration is automatically applied to {{domxref("Element.setHTML()")}} and the other [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods), and ensures that XSS-unsafe items are removed from the output even if they are allowed by a passed sanitizer.
+You can also call {{domxref("Sanitizer/removeUnsafe","removeUnsafe")}} on a {{domxref("Sanitizer")}} instance to apply this configuration and make it XSS-safe.
+
+#### Default sanitizer configuration
+
+The default sanitizer configuration is more restrictive than the XSS-safe baseline.
+It defines the sanitizer that is used if you call {{domxref("Element.setHTML()")}} or the other [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods) without passing a sanitizer object.
 It is also the configuration that is returned by the [`Sanitizer()` constructor](/en-US/docs/Web/API/Sanitizer/Sanitizer) when no configuration is set.
 
-This configuration removes the following sorts of items:
+The configuration removes the following sorts of items:
 
-1. Those that are known to be XSS-unsafe.
-2. Additional items that might be be used in clickjacking, spoofing, or other attacks.
+1. Those that are known to be XSS-unsafe (as specified in the [XSS-safe baseline configuration](#xss-safe_baseline_configuration)).
+2. Additional items that might be used in clickjacking, spoofing, or other attacks.
 3. Comments and `data-*` attributes.
 
-It therefore provides a sanitizer with a minimal attack surface, which is still suitable for the majority of sanitization use cases.
+Therefore it provides a sanitizer with a minimal attack surface, which is still suitable for the majority of sanitization use cases.
 
-Note that if you specify a _custom_ sanitizer that allows XSS-unsafe items to a safe sanitization method, these items will still be removed from the input.
-However, the safe sanitization methods do not automatically remove the additional items, comments, and `data-*` attributes, that would be removed by the default configuration.
-
-For a listing of the allowed elements and attributes, see [Default `Sanitizer` configuration](/en-US/docs/Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration).
+For a listing of the allowed elements and attributes, see [Default sanitizer configuration](/en-US/docs/Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration).
 
 ### Sanitization and Trusted Types
 
--- a/files/en-us/web/api/sanitizer/index.md
+++ b/files/en-us/web/api/sanitizer/index.md
@@ -43,8 +43,8 @@ It can be used with the following [sanitization methods](/en-US/docs/Web/API/HTM
 - Safe methods: {{domxref("Element.setHTML()")}}, {{domxref("ShadowRoot.setHTML()")}}, and [`Document.parseHTML()`](/en-US/docs/Web/API/Document/parseHTML_static).
 - Unsafe methods: {{domxref("Element.setHTMLUnsafe()")}}, {{domxref("ShadowRoot.setHTMLUnsafe()")}}, and [`Document.parseHTMLUnsafe()`](/en-US/docs/Web/API/Document/parseHTMLUnsafe_static).
 
-A `Sanitizer` instance can be constructed from a {{domxref("SanitizerConfig")}}, and is effectively a wrapper around that object. A `Sanitizer` and a `SanitizerConfig` can be used with the same methods, but if you're using the same configuration multiple times, it's expected to be more efficient to use a `Santitizer` and modify it when you need to.
-If no `SanitizerConfig` is passed to the constructor, the sanitizer is created with the [default `Sanitizer` configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#default_sanitizer_configuration), which removes XSS-unsafe elements and attributes, along with other elements and attributes that can potentially be used in other attacks, such as clickjacking and spoofing.
+A `Sanitizer` instance can be constructed from a {{domxref("SanitizerConfig")}}, and is effectively a wrapper around that object. A `Sanitizer` and a `SanitizerConfig` can be used with the same methods, but if you're using the same configuration multiple times, it's expected to be more efficient to use a `Sanitizer` and modify it when you need to.
+If no `SanitizerConfig` is passed to the constructor, the sanitizer is created with the [default sanitizer configuration](/en-US/docs/Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration), which removes XSS-unsafe elements and attributes, along with other elements and attributes that can potentially be used in other attacks, such as clickjacking and spoofing.
 
 Note that any `Sanitizer` can be made XSS-safe by calling {{domxref("Sanitizer.removeUnsafe()")}}, but other potentially dangerous elements and attributes — which are removed by the default configuration — may still be present.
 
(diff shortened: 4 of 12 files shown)


files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
```diff
@@ -0,0 +1,230 @@
+---
+title: Default Sanitizer configuration
+slug: Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration
+page-type: guide
+---
+
```
hamishwillee, line 6: @evilpie Note, this is your list, modified to:
1. Make all the elements internal links.
2. Elements listed first - generally I like to know about them first.
3. Introduction explaining the rationall. Somewhat duplicated in the Concepts in the overview doc.
4. Links to spec.
5. NOte on the existence of the baseline spec for removal
6. Global attributes somewhat grouped to make them easier to digest. This is arguable.

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
```diff
@@ -0,0 +1,230 @@
...
+It therefore provides a "more than XSS safe" sanitizer, that minimizes the attack surface while still being suitable for the majority of sanitization use cases.
+
+The HTML, SVG, and MathML elements and attributes allowed by the configuration are specified in the following sections.
+Comments and `data-*` attributes are removed by default.
+Note that this is derived from the specification, where it is referred to as the [built-in safe default configuration](https://wicg.github.io/sanitizer-api/#built-in-safe-default-configuration).
+
+> [!NOTE]
+> The specification also defines the [built-in safe baseline configuration](https://wicg.github.io/sanitizer-api/#built-in-safe-baseline-configuration), which lists the elements that must minimally be removed to block script-content: {{htmlelement("embed")}}, {{htmlelement("frame")}}, {{htmlelement("iframe")}}, {{htmlelement("object")}}, {{htmlelement("script")}}, and `use`.
```
mozfreddyb: Is there an svgelement macro to point to https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/use?

files/en-us/web/api/element/sethtml/index.md
```diff
@@ -55,10 +55,11 @@ The **`setHTML()`** method provides an XSS-safe method to parse and sanitize a s
 `setHTML()` drops any elements in the HTML input string that are invalid in the context of the current element, such as a {{htmlelement("col")}} element outside of a {{htmlelement("table")}}.
 It then removes any HTML entities that aren't allowed by the sanitizer configuration, and further removes any XSS-unsafe elements or attributes — whether or not they are allowed by the sanitizer configuration.
 
-If no sanitizer configuration is specified in the `options.sanitizer` parameter, `setHTML()` is used with the default {{domxref("Sanitizer")}} configuration.
-This configuration allows all elements and attributes that are considered XSS-safe, thereby disallowing entities that are considered unsafe; see the [`Sanitizer()`](/en-US/docs/Web/API/Sanitizer/Sanitizer) constructor for more information.
+If no sanitizer configuration is specified in the `options.sanitizer` parameter, `setHTML()` is used with the [default Sanitizer configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#default_sanitizer_configuration).
+This configuration is suitable for the majority of sanitization use cases (note that it removes more than just the XSS-unsafe elements and attributes).
```
mozfreddyb:
```suggestion
This configuration is suitable for the majority of use cases as it handles XSS as well as other issues like clickjacking or spoofing.
```

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
```diff
@@ -0,0 +1,230 @@
...
+slug: Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration
+page-type: guide
+---
+
+{{DefaultAPISidebar("HTML Sanitizer API")}}
+
+The default Sanitizer configuration defines the {{domxref("Sanitizer")}} that is used by default if you call the [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods), such as {{domxref("Element.setHTML()")}}, {{domxref("ShadowRoot.setHTML()")}}, and {{domxref("Document.parseHTML_static","Document.parseHTML()")}}, without specifying a custom sanitizer.
+It is also the default configuration returned by the [`Sanitizer()` constructor](/en-US/docs/Web/API/Sanitizer/Sanitizer) if no configuration is set.
+
+The set of elements and attributes [allowed](/en-US/docs/Web/API/HTML_Sanitizer_API#allow_configurations) by the default configuration omits those that are known to be XSS-unsafe, along with other elements and attributes that are considered problematic for other reasons.
```
mozfreddyb:
```suggestion
The set of elements and attributes [allowed](/en-US/docs/Web/API/HTML_Sanitizer_API#allow_configurations) by the default configuration omits those that are known to be XSS-unsafe, along with other elements and attributes that are considered unsafe for user inputs (e.g., {{htmlelement("iframe")}}).
```

files/en-us/web/api/sanitizer/sanitizer/index.md
```diff
@@ -20,7 +20,9 @@ new Sanitizer(configuration)
 ### Parameters
 
 - `configuration` {{optional_inline}}
-  - : A {{domxref("SanitizerConfig")}} defining a [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration), or the string `"default"` to indicate the default configuration.
+  - : A {{domxref("SanitizerConfig")}} defining a [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration), or the string `"default"` to indicate the [default Sanitizer configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#default_sanitizer_configuration).
+
+    If omitted, the constructor returns a `Sanitizer` with the default configuration.
```
mozfreddyb, line 26: Maybe also mention here that `{}` gives you an xss-safe sanitizer when using the safe function variants?

Maybe also mention that **any** kind of custom configuration (even when used with an unsafe function) can be made safe by calling `removeUnsafe()`?

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
hamishwillee: Thanks. I'm going to try a slightly bigger definitition using your other examples.

```suggestion
The set of elements and attributes [allowed](/en-US/docs/Web/API/HTML_Sanitizer_API#allow_configurations) by the default configuration omits those that are known to be XSS-unsafe, along with other elements and attributes that might be used for clickjacking or spoofing, or are otherwise potentially unsafe if provided by a user.
```

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -32,8 +32,8 @@ The configuration defines the HTML entities that will be filtered out of the inp
 The {{domxref('Element')}} methods are context aware, and will additionally drop any elements that the HTML specification does not allow in the target element.
 
 The safe methods always remove XSS-unsafe elements and attributes.
-If no sanitizer is passed as a parameter they will use the default sanitizer configuration, which allows all elements and attributes except those that are known to be unsafe, such as {{htmlelement("script")}} elements and `onclick` event handlers.
-If a custom sanitizer is used, it is implicitly updated to remove any elements and attributes that are not XSS-safe (note that the passed sanitizer is not modified, and might still allow unsafe entities if used with an unsafe method).
+If no sanitizer is passed as a parameter they will use the [default sanitizer configuration](#default_sanitizer_configuration), which removes both XSS-unsafe elements and attributes, such as {{htmlelement("script")}} elements and `onclick` event handlers, and others that are problematic for other reasons.
```
hamishwillee: FWIW, the default sanitiser config doc can then be expanded if we want to come up with a more extensive list of reasons why particular elements/attributes are not allowed.

```suggestion
If no sanitizer is passed as a parameter they will use the [default sanitizer configuration](#default_sanitizer_configuration), which removes both XSS-unsafe elements and attributes, such as {{htmlelement("script")}} elements and `onclick` event handlers, along with others that might be be used in other kinds of attacks if provided as user input.
```

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -193,6 +193,19 @@ So if you call `allowElement()` on an allow configuration and the specified elem
 But if the element is already present then the method would return `false`.
 Note that if you call the same method to set a per-element attribute, this will return `false` if called on a remove sanitizer, because the change cannot be made.
 
+#### Default Sanitizer configuration
+
+The default Sanitizer configuration defines the {{domxref("Sanitizer")}} that is used if you call {{domxref("Element.setHTML()")}} or the other [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods) without a custom sanitizer.
+It is also the configuration that is returned by the [`Sanitizer()` constructor](/en-US/docs/Web/API/Sanitizer/Sanitizer) when no configuration is set.
+
+This sanitizes elements and attributes that are known to be XSS-unsafe, along with other elements and attributes that are considered problematic for other reasons.
```
hamishwillee:
```suggestion
This sanitizes elements and attributes that are known to be XSS-unsafe, along with others that might be be used in clickjacking, spoofing, or other attacks, if provided as user input.
```

files/en-us/web/api/sanitizer/sanitizer/index.md
hamishwillee: @mozfreddyb I mentioned that `{}` can be passed because at this point it isn't clear whether it is a "valid config" - so worth saying.

However IMO constructor not appropriate level to discuss "how to make configurations safe".
That said, you wouldn't have mentioned it if this seemed well covered, so I have updated the  `Santizer` to better cover this.

I also made consistent updates to update the text you identified about elements removed "for other reasons" with the examples provided (good call).

YOu can see those changes inhttps://github.com/mdn/content/pull/43271/changes/9a2da1b21d289e0a0da6dbe4fa891f86c2520a15

I think this addresses your issues.

files/en-us/web/api/sanitizer/sanitizer/index.md
```diff
@@ -21,6 +21,7 @@ new Sanitizer(configuration)
 
 - `configuration` {{optional_inline}}
   - : A {{domxref("SanitizerConfig")}} defining a [valid configuration](/en-US/docs/Web/API/SanitizerConfig#valid_configuration), or the string `"default"` to indicate the [default Sanitizer configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#default_sanitizer_configuration).
+    The "empty configuration" (`{}`) can also be passed, and results in a [remove configuration](/en-US/docs/Web/API/HTML_Sanitizer_API#remove_configurations) with empty arrays.
```
mozfreddyb, line 24: Maybe this is too much jargon? I think we should say this ends up as a Sanitizer that only prevents XSS. Or if we want to use explanation referenced differently maybe say "allow list" and "block list"?

mozfreddyb:
This is much better, thank you!

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
```diff
@@ -0,0 +1,230 @@
...
+- {{SVGAttr("stroke-linecap")}}
+- {{SVGAttr("stroke-linejoin")}}
+- {{SVGAttr("stroke-miterlimit")}}
+- {{SVGAttr("stroke-opacity")}}
+- {{SVGAttr("stroke-width")}}
+- {{SVGAttr("text-anchor")}}
+- {{SVGAttr("text-rendering")}}
+- {{SVGAttr("vector-effect")}}
+
+### CSS properties
```
evilpie: This is not correct. These are SVG attributes that map to CSS properties. e.g. https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/clip-path

files/en-us/web/api/sanitizer/sanitizer/index.md
hamishwillee: Thanks, but I don't think so. 

1. On the topic of Jargon, yes it is a bit that way, but I still don't want to change it. The way Sanitizers work internally when you add and remove attributes is quite complicated and differs based on the configuration the sanitizer is created with. In other words, I think it _matters_ in a way that allow list and block list don't capture.
2. " this ends up as a Sanitizer that only prevents XSS." - That is true when a Sanitizer is used in safe method or removeUnsafe is used, but not otherwise. At this level though we're talking about what exactly we get back. IN other words, not the right place. But I have attempted to address as discussed in that other comment below https://github.com/mdn/content/pull/43271/changes#r2875603522

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
hamishwillee: Thanks. Fixed by merging into above list. https://github.com/mdn/content/pull/43271/changes/93e21aa79a84c40d66972db438358ae0c883af32

hamishwillee:
@wbamberg Do you think you might be able to look at this please?

wbamberg:
Yes, I'll look tomorrow. Sorry to be slow.

files/en-us/web/api/sanitizer/sanitizer/index.md
mozfreddyb: ok, sgtm :)

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -193,6 +193,19 @@ So if you call `allowElement()` on an allow configuration and the specified elem
 But if the element is already present then the method would return `false`.
 Note that if you call the same method to set a per-element attribute, this will return `false` if called on a remove sanitizer, because the change cannot be made.
 
+#### Default Sanitizer configuration
+
+The default Sanitizer configuration defines the {{domxref("Sanitizer")}} that is used if you call {{domxref("Element.setHTML()")}} or the other [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods) without a custom sanitizer.
```
wbamberg: We should be consistent about the capitalization of "sanitizer" when it's not the name of the interface. I prefer lower-case, but the current pages are not consistent.

files/en-us/web/api/html_sanitizer_api/index.md
```diff
@@ -193,6 +193,19 @@ So if you call `allowElement()` on an allow configuration and the specified elem
...
+It is also the configuration that is returned by the [`Sanitizer()` constructor](/en-US/docs/Web/API/Sanitizer/Sanitizer) when no configuration is set.
+
+This sanitizes elements and attributes that are known to be XSS-unsafe, along with others that might be be used in clickjacking, spoofing, or other attacks, if provided as user input.
+It therefore provides a sanitizer with a minimal attack surface, which is still suitable for the majority of sanitization use cases.
+
+Note that if you specify a custom sanitizer for a safe sanitization method, it will always be updated to ensure XSS-unsafe elements and event handler content attributes are sanitized.
+However, it will not be updated to remove the additional "problematic" elements that aren't allowed by the default configuration.
```
wbamberg: Not that keen on the scare-quoted "problematic". Why not say "potentially dangerous"?

But also, AIUI what this is saying is, line 201 says that there are 2 sorts of things removed in the default config: (1) things that are XSS-dangerous, and (2) things that are dangerous in other ways: "clickjacking, spoofing, or other attacks".

And if I make a custom config and use it with a safe method, then whatever I choose to put in it, the things in (1) are always removed, but the things in (2) are not. Is that right?

If so I think it could more clearly indicate that 'additional "problematic" elements' refers to the items listed in the first para, maybe even just by more aggressively separating the two types, with something like:

```suggestion
This configuration removes these sorts of items:

 1. Those that are known to be XSS-unsafe.
 2. Additional items that might be be used in clickjacking, spoofing, or other attacks.
 3. Comments and `data-*` attributes.

It therefore provides a sanitizer with a minimal attack surface, which is still suitable for the majority of sanitization use cases.

Note that if you specify a custom sanitizer for a safe sanitization method, it will always be updated to ensure XSS-unsafe items are removed.
However, it will not be updated to remove the additional items, comments or `data-*` attributes that aren't allowed by the default configuration.
```

? But I'm also a bit confused. Line 201 refers to "elements and attributes that are known to be XSS-unsafe", but line 204 refers to "XSS-unsafe elements and event handler content attributes". So are these actually talking about the same thing? Are there any XSS-unsafe attributes that aren't event handler attributes, and if so are these not removed in a custom config?

Finally, I think it would be very interesting and helpful to indicate which elements/attributes are removed for "XSS-unsafety" and which are removed for "clickjacking, spoofing, and other attacks" (and are therefore, AIUI, not removed automatically from a custom config).

files/en-us/web/api/sanitizer/index.md
```diff
@@ -43,6 +35,30 @@ Note that `Sanitizer` is expected to be more efficient to reuse and modify when
...
 
+## Description
+
+A `Sanitizer` is a reusable configuration object that specifies what elements, attributes and comments are allowed or should be removed when inserting strings of HTML into an {{domxref("Element")}} or {{domxref("ShadowRoot")}}, or when parsing an HTML string into a {{domxref("Document")}}.
+It can be used with the following [sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods):
+
+- Safe methods: {{domxref("Element.setHTML()")}}, {{domxref("ShadowRoot.setHTML()")}}, and [`Document.parseHTML()`](/en-US/docs/Web/API/Document/parseHTML_static).
+- Unsafe methods: {{domxref("Element.setHTMLUnsafe()")}}, {{domxref("ShadowRoot.setHTMLUnsafe()")}}, and [`Document.parseHTMLUnsafe()`](/en-US/docs/Web/API/Document/parseHTMLUnsafe_static).
+
+A `Sanitizer` instance can be constructed from a {{domxref("SanitizerConfig")}}, and is effectively a wrapper around that object (it can be used with the same methods as `SanitizerConfig`, but is expected to be more efficient to reuse and modify when needed).
```
wbamberg:
```suggestion
A `Sanitizer` instance can be constructed from a {{domxref("SanitizerConfig")}}, and is effectively a wrapper around that object. A `Sanitizer` and a `SanitizerConfig` can be used with the same methods, but if you're using the same configuration multiple times, it's expected to be more efficient to use a `Santitizer` and modify it when you need to.
```

(I think? I thought "is expected to be more efficient to reuse and modify when needed" was a bit unclear.)

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
```diff
@@ -0,0 +1,227 @@
...
+slug: Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration
+page-type: guide
+---
+
+{{DefaultAPISidebar("HTML Sanitizer API")}}
+
+The default Sanitizer configuration defines the {{domxref("Sanitizer")}} that is used by default if you call the [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods), such as {{domxref("Element.setHTML()")}}, {{domxref("ShadowRoot.setHTML()")}}, and {{domxref("Document.parseHTML_static","Document.parseHTML()")}}, without specifying a custom sanitizer.
+It is also the default configuration returned by the [`Sanitizer()` constructor](/en-US/docs/Web/API/Sanitizer/Sanitizer) if no configuration is set.
+
+The set of elements and attributes [allowed](/en-US/docs/Web/API/HTML_Sanitizer_API#allow_configurations) by the default configuration omits those that are known to be XSS-unsafe, along with other elements and attributes that might be used for clickjacking or spoofing, or are otherwise potentially unsafe if provided by a user.
```
wbamberg: As I said before, I'd love to know more detail here about which removals protect against which threats.

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
```diff
@@ -0,0 +1,227 @@
...
+It therefore provides a "more than XSS safe" sanitizer, that minimizes the attack surface while still being suitable for the majority of sanitization use cases.
+
+The HTML, SVG, and MathML elements and attributes allowed by the configuration are specified in the following sections.
+Comments and `data-*` attributes are removed by default.
+Note that this is derived from the specification, where it is referred to as the [built-in safe default configuration](https://wicg.github.io/sanitizer-api/#built-in-safe-default-configuration).
```
wbamberg:
```suggestion
The specification calls this configuration the [built-in safe default configuration](https://wicg.github.io/sanitizer-api/#built-in-safe-default-configuration).
```

(I mean, just about everything is derived from the specification)

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
```diff
@@ -0,0 +1,227 @@
...
+
+> [!NOTE]
+> The specification also defines the [built-in safe baseline configuration](https://wicg.github.io/sanitizer-api/#built-in-safe-baseline-configuration), which lists the elements that must minimally be removed to block script-content: {{htmlelement("embed")}}, {{htmlelement("frame")}}, {{htmlelement("iframe")}}, {{htmlelement("object")}}, {{htmlelement("script")}}, and {{SVGElement("use")}} .
+> "Safe" sanitization methods will remove these elements from a custom sanitizer automatically, along with any event handler content attributes (such as `onafterprint`, `onbeforeinput`, and so on).
+
+## Allowed elements
+
+The "Additional allowed attributes" columns list attributes permitted only on that specific element (the [global attributes](#global_attributes) listed below are allowed on any element).
+
+### HTML elements
```
wbamberg: I think you might disagree with this :) but I would consider listing _all_ HTML elements, and indicating in the table whether they are allowed or not. The reason is that it always gives me a positive answer about any given element - otherwise I have to wonder: is `<input>` not here because it's not allowed, or did Hamish forget to add it?

It's like the ongoing thing about MDN not indicating events that bubble, but only those that do not.

You could additionally (although I'm not sure this would work, because of the attributes) add a column for the baseline configuration, and AIUI this would help with my question above about describing which removals are the "XSS-unsafe" ones.

files/en-us/web/api/html_sanitizer_api/index.md
hamishwillee: @wbamberg Absolutely. I have attempted a fix inhttps://github.com/mdn/content/pull/43271/changes/9f8d36524f1fcb7b9cc14edc6eb4d3b178268d01 - note though that there are so many instances I might have missed some.

Essentially I have used `Sanitizer` or `SanitizerConfig` when I mean specifically that type of object, and otherwise it is lower case, and can mean either.

files/en-us/web/api/html_sanitizer_api/index.md
hamishwillee: @wbamberg 
- Yes, all correct. Yes, this captures the way it works more clearly. I have added item "3" to include comments and `data-*` attributes and I will accept.
- I used the term problematic mostly because that is what the spec used. Potentially dangerous is just as good or better. I suspect problematic conveys that it is an opinionated list that includes some elements that the spec authors just don't like.
- ... but line 204 refers to "XSS-unsafe elements and event handler content attributes". Yes, I think that this is JUST the content handler attributes - but I will check again when I propagate this structure to the main default sanitizer doc.

hamishwillee: @wbamberg FYI, actually a minor nuance. A custom sanitizer is not actually "updated" to remove the XSS unsafe elements when it is used with a safe method. It is more that the unsafe items are also removed from the output. 
The difference here is that an unsafe configuration remains unsafe if it is used in an unsafe method.

I'll tune that a little.

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
hamishwillee: So would I. We can say for sure what removals protect against XSS, but the spec doesn't cover the others, and I am not going to guess.

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
hamishwillee: I do disagree (a bit), but Claude makes testing plausible https://pr43271.review.mdn.allizom.net/en-US/docs/Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration

(Change https://github.com/mdn/content/pull/43271/changes/524441db22622466b1d023df80c73bbf86a15f19)

Can you have a look at the new version and tell me if you think that this is better? Some notes:
- Up the top I explicitly list the XSS removals in the list of things that are removed. 
- The Element sections are now level 2 and appear in the sidebar. This is an improvement. 
- The elements tables list "all elements" as grepped by claude from MDN element overview pages. My experience is that Claude is pretty good at these changes. Some confidence by the fact that the links all work.
- The elements tables have a column for the "allowed" elements with a checkbox. 
- The tables don't have an attacks column. The spec does not give specific reasons for anything that has been removed other than XSS.
- I have removed info about the safe baseline config and what is removed by `Sanitizer.removeUnsafe()` into that method in https://github.com/mdn/content/pull/43271/changes/12a228427384723fc24751672bfce4f38d46cc13

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
hamishwillee: Upshot, maybe some other iteration.

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
```diff
@@ -0,0 +1,334 @@
+---
+title: Default `Sanitizer` configuration
```
wbamberg: You can't put Markdown in titles, so this doesn't work.

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
```diff
@@ -0,0 +1,334 @@
...
+slug: Web/API/HTML_Sanitizer_API/Default_sanitizer_configuration
+page-type: guide
+---
+
+{{DefaultAPISidebar("HTML Sanitizer API")}}
+
+The default `Sanitizer` configuration defines the {{domxref("Sanitizer")}} that is used by default if you call the [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods), such as {{domxref("Element.setHTML()")}}, {{domxref("ShadowRoot.setHTML()")}}, and {{domxref("Document.parseHTML_static","Document.parseHTML()")}}, without specifying a custom sanitizer.
```
wbamberg:
```suggestion
The default sanitizer configuration defines the {{domxref("Sanitizer")}} that is used by default if you call the [safe sanitization methods](/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods), such as {{domxref("Element.setHTML()")}}, {{domxref("ShadowRoot.setHTML()")}}, and {{domxref("Document.parseHTML_static","Document.parseHTML()")}}, without specifying a custom sanitizer.
```

(would be my preference. Or "The default configuration for `Sanitizer` objects")

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
wbamberg: Up to you whether to use this version. I think it is better but I've said my piece. I'd also have ❌ or something for the ones that it blocks, for readability.

I didn't mean an attacks column.

As I understand it:

1. the [_built-in safe default baseline configuration_](https://wicg.github.io/sanitizer-api/#built-in-safe-baseline-configuration) blocks "XSS unsafe" elements like `<iframe>`, and event handler content attributes. 
2. the [_built-in safe default configuration_](https://wicg.github.io/sanitizer-api/#built-in-safe-default-configuration) blocks everything in (1), plus extra stuff that may allow other attacks.

| What you do | What you get |
| -----------|------- |
| create a sanitizer with no config    | 2    |
| call `setHTML()` (or any other "safe" method) with no specified sanitizer    | 2    |
| call `setHTML()` (or any other "safe" method) with a custom sanitizer    | 1 (in addition to your custom behavior)    |
| `setHTMLUnsafe()` (or any other "unsafe" method) with no specified sanitizer    | nothing    |
| `setHTMLUnsafe()` (or any other "unsafe" method) with a custom sanitizer    | nothing (beyond your custom behavior) (?)    |

Is that right? I don't know if there's a place in these docs that summarises this situation. It helps IMO to be able to name things, so you can refer to specific configs and be precise, so a reader knows exactly how a phrase like "XSS-unsafe items" pays out in terms of what's actually allowed. That means having a sharp definition of (1) and (2) that you can link to whenever it's relevant.

But also, it would be nice to see clearly what the delta is between (1) and (2) - because one thing that this means for a developer is, if I use an safe method with a custom config, I have to worry about what's in that delta. So I thought that including an extra column showing this would be helpful.

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
hamishwillee: Thanks. Keeping as upper case though, because I do mean `Sanitizer`.

```suggestion
title: Default Sanitizer configuration
```

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
hamishwillee: Essentially your summation is correct. Let me think on this

FYI only - just to be pendantic in case you care about the very little details in the spec

---

> the [built-in safe default baseline configuration](https://wicg.github.io/sanitizer-api/#built-in-safe-baseline-configuration) blocks "XSS unsafe" elements like <iframe>, and event handler content attributes.

Ish. Configuration is just the elements. The event handler attributes are separate in the spec because they are considered a list that is likely to change. Makes no real difference to your point or how I would present it.

---

One caveat is that both setHTML and setHTMLUnsafe WILL do parsing on the input and remove elements that aren't allowed in the context, such as a td element outside of a table. Again, just an FYI - but it does mean that setHTMLUnsafe doesn't do "nothing".

files/en-us/web/api/html_sanitizer_api/default_sanitizer_configuration/index.md
hamishwillee: Accepting, but not sure I will do everywhere.

hamishwillee: OK, everywhere it is a "default sanitizer configuration". Originally I wanted to define what the actual object was - i.e. this is not a `SanitizerConfig`. However when refering to the default in a safe method, it isn't a `Sanitizer` - under the hood it could be anything, even if it has the same result as that object.

hamishwillee:
@wbamberg Thanks for your review. I have made the changes suggested to the table and to use generally sanitizer in lower case everywhere.

I would have liked to include something like your table https://github.com/mdn/content/pull/43271#discussion_r2920135566 but I could not find a good way to integrate it. Instead I created a built in configurations section https://pr43271.review.mdn.allizom.net/en-US/docs/Web/API/HTML_Sanitizer_API#built-in_configurations to nest the XSS-safe baseline and the default configuration. This makes it more clear that they are different and how.

Might not be perfect but there is a lot of improvement. Merging.

mozfreddyb:
thank you!

PR #43256, opened 2026-02-26
docs: add animal idiom alternatives to inclusive language guide

stuckvgn:
## Summary

Closes #43242.

Following reviewer consensus in the issue thread, this is a narrowly scoped addition to the "Use inclusive language" section of the Writing Style Guide. It adds guidance to avoid figurative idioms that use animals in violent or harmful contexts, with plain-English alternatives for three common phrases:

- "kill two birds with one stone" → "solve two problems at once"
- "beating a dead horse" → "belaboring the point" or "going in circles"
- "more than one way to skin a cat" → "more than one way to do this"

**Explicitly excluded** per reviewer feedback in the issue: technical terms like `kill` (Unix command), `monkey-patching`, and `dogfooding` — these are well-established technical vocabulary and replacing them would introduce confusion without benefit.

PR #43273, opened 2026-02-27
fix: replace "Whack-a-Mole" idiom per inclusive language guide

stuckvgn:
## Summary

Follow-up to #43256 — applies the new inclusive language guidance added in that PR.

Replaces a single instance of "Whack-a-Mole" (a figurative idiom implying violence toward animals) with "chasing regressions," which is both more inclusive and more precise in context.

**File changed:** `files/en-us/games/anatomy/index.md`

This was the only instance found in the content repo after a full search for figurative animal idioms.


chrisdavidmills (approved):
LGTM.

PR #43274, opened 2026-02-27
Update the collation descriptions

hsivonen:
### Description

* Document that emoji and eor ignore the language (which makes sense for eor but may be an unwanted surprise with emoji).
* Use concrete phrasing for which language each collation attaches to, since only `trad` is really "such as".

### Motivation

The behavior of `eor` and `emoji` is likely surprising to readers. (It was surprising to me as a collator implementor!)

Vague "such as" isn't really useful, so while at it, I updated some other items to concretely say which languages they attach to. (`trad` is the only real "such as" here.)

### Additional details

The reviewer may convince themselves that the allegation of behavior is true by experimenting with `.compare("\u0587", "\u0584")` in the `eor` case and with `.compare("😄", "😁")` in the `emoji` case.


hsivonen:
Now that I'm editing this, I bundled slight precision update (s/CJK/Han/) and also was more specific about what non-Han effects `unihan` has.

sideshowbarker (approved):

PR #43283, opened 2026-02-27
Document Notification.navigate property

pepelsbey:
### Description

**New pages**

- `API/Notification/navigate` — The `navigate` property

**Updated pages**

- `API/Notification` — Added `navigate` to the list of instance properties
- `API/Notification/Notification` — Added `navigate` to the constructor’s `options`
- `API/ServiceWorkerRegistration/showNotification` — Added `navigate` to the `options` and the `actions`

### Motivation

Document property supported since Safari 18.4

### Additional details

- [BCD: Notification.navigate](https://github.com/mdn/browser-compat-data/blob/main/api/Notification.json#L827)
- [Spec: dom-notification-navigate](https://notifications.spec.whatwg.org/#dom-notification-navigate)

### DIFF
--- a/files/en-us/web/api/notifications_api/index.md
+++ b/files/en-us/web/api/notifications_api/index.md
@@ -11,7 +11,9 @@ spec-urls: https://notifications.spec.whatwg.org/
 
 {{DefaultAPISidebar("Web Notifications")}}{{securecontext_header}} {{AvailableInWorkers}}
 
-The Notifications API allows web pages to control the display of system notifications to the end user. These are outside the top-level browsing context viewport, so therefore can be displayed even when the user has switched tabs or moved to a different app. The API is designed to be compatible with existing notification systems, across different platforms.
+The Notifications API allows web pages to control the display of system notifications to the end user.
+These are outside the top-level browsing context viewport, so therefore can be displayed even when the user has switched tabs or moved to a different app.
+The API is designed to be compatible with existing notification systems, across different platforms.
 
 ## Concepts and usage
 
@@ -35,12 +37,13 @@ This will spawn a request dialog, along the following lines:
 
 ![A dialog box asking the user to allow notifications from that origin. There are options to never allow or allow notifications.](screen_shot_2019-12-11_at_9.59.14_am.png)
 
-From here the user can choose to allow notifications from this origin, or block them. Once a choice has been made, the setting will generally persist for the current session.
+From here the user can choose to allow notifications from this origin, or block them.
+Once a choice has been made, the setting will generally persist for the current session.
 
 ### Notification display and handling
 
 Notifications are created using the {{domxref("Notification.Notification","Notification()")}} constructor.
-This must be passed a title argument, and can optionally be passed an options object to specify options, such as text direction, body text, icon to display, notification sound to play, and more.
+This must be passed a title argument, and can optionally be passed a parameter to specify options such as text direction, body text, icon to display, notification sound to play, and more.
 
 For example, the following code shows how you might create a notification that sets the [`navigate`](/en-US/docs/Web/API/Notification/Notification#navigate) option, specifying a URL that will be opened if the notification is accepted (you can also defined click handlers to process notification actions).
 
@@ -71,7 +74,8 @@ The Notifications API supports two types of notifications:
 ## Interfaces
 
 - {{domxref("Notification")}}
-  - : Defines a notification object. When activated, a non-persistent notification fires a {{domxref("Notification.click_event", "click")}} event, unless a {{domxref("Notification.navigate", "navigate")}} URL is set, in which case the user agent navigates to that URL instead.
+  - : Defines a notification object.
+    When activated, a non-persistent notification fires a {{domxref("Notification.click_event", "click")}} event, unless a {{domxref("Notification.navigate", "navigate")}} URL is set, in which case the user agent navigates to that URL instead.
 - {{domxref("NotificationEvent")}}
   - : Represents a notification event dispatched on the {{domxref("ServiceWorkerGlobalScope")}} of a {{domxref("ServiceWorker")}}.
 
--- a/files/en-us/web/api/notification/notification/index.md
+++ b/files/en-us/web/api/notification/notification/index.md
@@ -10,11 +10,15 @@ browser-compat: api.Notification.Notification
 
 The **`Notification()`** constructor creates a new {{domxref("Notification")}} object instance, which represents a user notification.
 
-Trying to create a notification inside the {{domxref("ServiceWorkerGlobalScope")}} using the `Notification()` constructor will throw a `TypeError`. Use {{domxref("ServiceWorkerRegistration.showNotification()")}} instead.
+Trying to create a notification inside the {{domxref("ServiceWorkerGlobalScope")}} using the `Notification()` constructor will throw a `TypeError`.
+Use {{domxref("ServiceWorkerRegistration.showNotification()")}} instead.
 
-You must first get permission before being able to display notifications, using {{domxref("Notification.requestPermission_static", "Notification.requestPermission()")}}. The permission may not be grantable, for example if the page is in private browsing mode.
+You must first get permission before being able to display notifications, using {{domxref("Notification.requestPermission_static", "Notification.requestPermission()")}}.
+The permission may not be grantable, for example if the page is in private browsing mode.
 
-This constructor throws a {{jsxref("TypeError")}} when called in nearly all mobile browsers and this is unlikely to change, because web pages on mobile devices almost never "run in the background", which is the main use case for notifications. Instead, you need to register a service worker and use {{domxref("ServiceWorkerRegistration.showNotification()")}}. See [Chrome issue](https://crbug.com/481856) for more information.
+This constructor throws a {{jsxref("TypeError")}} when called in nearly all mobile browsers and this is unlikely to change, because web pages on mobile devices almost never "run in the background", which is the main use case for notifications.
+Instead, you need to register a service worker and use {{domxref("ServiceWorkerRegistration.showNotification()")}}.
+See [Chrome issue](https://crbug.com/481856) for more information.
 
 ## Syntax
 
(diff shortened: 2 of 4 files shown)


files/en-us/web/api/notification/navigate/index.md
```diff
@@ -0,0 +1,95 @@
...
+slug: Web/API/Notification/navigate
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.Notification.navigate
+---
+
+{{APIRef("Web Notifications")}}{{securecontext_header}}{{SeeCompatTable}} {{AvailableInWorkers}}
+
+The **`navigate`** read-only property of the {{domxref("Notification")}} interface contains the URL the user agent will navigate to when the user activates the notification, as specified in the `navigate` option of the {{domxref("Notification.Notification", "Notification()")}} constructor or {{domxref("ServiceWorkerRegistration.showNotification()")}}.
```
hamishwillee: A bit clunky. Maybe:
```suggestion
The **`navigate`** read-only property of the {{domxref("Notification")}} interface contains the URL the user agent will navigate to when the user activates the notification.

This is the resolved value of the URL, if any, that was specified in the `navigate` option of the {{domxref("Notification.Notification", "Notification()")}} constructor or {{domxref("ServiceWorkerRegistration.showNotification()")}}.
```

files/en-us/web/api/notification/navigate/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+When a notification with a navigation URL is activated, the user agent navigates to the specified URL instead of firing the {{domxref("Notification.click_event", "click")}} event or the {{domxref("ServiceWorkerGlobalScope.notificationclick_event", "notificationclick")}} event. This allows notifications to direct users to a specific page without requiring a click event handler.
```
hamishwillee: What does activated mean? This API suffers from the fact that there is not a single real example of a notification prompt for me to check. I assume though that this means each notifcation has a button. Is this true:

```suggestion
When a notification with a navigation URL is accepted (by clicking its acknowledge button) the user agent navigates to the specified URL instead of firing the {{domxref("Notification.click_event", "click")}} event or the {{domxref("ServiceWorkerGlobalScope.notificationclick_event", "notificationclick")}} event. This allows notifications to direct users to a specific page without requiring a click event handler.
```

files/en-us/web/api/notification/navigate/index.md
```diff
@@ -0,0 +1,95 @@
...
+        body: "Hey, are you coming to the party tonight?",
+        navigate: "/messages/alice",
+      });
+
+      // The navigate property reflects the resolved URL
+      console.log(notification.navigate);
+    }
+  });
+});
+```
```
hamishwillee, line 43: Can we make these live. It would make me happy.

files/en-us/web/api/notification/navigate/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Value
+
+A string containing a {{glossary("URL")}}, or an empty string if no navigation URL has been set.
+
+## Examples
+
+### Creating a notification with a navigation URL
+
+This example creates a notification that, when activated by the user, navigates to a specific page rather than firing a click event. Note that {{domxref("Notification.requestPermission_static", "Notification.requestPermission()")}} must be called from a user gesture such as a button click.
```
hamishwillee: Again, what is firing the click event? I am pretty sure you mean tha notification, but the only thing shown on the screen that might fire an event is notify button below. Be more precise?

files/en-us/web/api/notification/navigate/index.md
```diff
@@ -0,0 +1,95 @@
...
+      // The navigate property reflects the resolved URL
+      console.log(notification.navigate);
+    }
+  });
+});
+```
+
+### Using navigate with a service worker
+
+When using persistent notifications through a service worker, the `navigate` option allows the notification to open a page without needing to handle the {{domxref("ServiceWorkerGlobalScope.notificationclick_event", "notificationclick")}} event.
```
hamishwillee: Which event is fired - click, or notificationclick, or both? Here it looks like notificationclick only

files/en-us/web/api/notification/navigate/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+```js
+// Inside a service worker
+self.registration.showNotification("Order shipped!", {
+  body: "Your order #1234 has been shipped.",
+  navigate: "/orders/1234",
+});
+```
+
+### Checking the navigate property value
```
hamishwillee: Given the property is about `navigate` should this example go first?

hamishwillee: And in this context, are the other examples relevant?

files/en-us/web/api/notification/navigate/index.md
```diff
@@ -0,0 +1,95 @@
...
+title: "Notification: navigate property"
+short-title: navigate
+slug: Web/API/Notification/navigate
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.Notification.navigate
+---
+
+{{APIRef("Web Notifications")}}{{securecontext_header}}{{SeeCompatTable}} {{AvailableInWorkers}}
```
hamishwillee, line 11: As an aside, the value of the navigate feature isn't all that clear but that info probably belongs in the place where the property is set.

What I mean is that it is easy to set a click handler, so it isn't obvious why it is worth providing the feature in the interface. I suspect that perhaps there are nuances to the  handling in the service worker case that aren't obvious.

files/en-us/web/api/notification/navigate/index.md
hamishwillee: Ah, I get it - there is a different event in a worker. MIght be worth noting/highlighting that up in the intro.

files/en-us/web/api/notification/navigate/index.md
pepelsbey: Unfortunately, live examples aren't feasible here: cross-origin iframes on `live.mdnplay.dev` just won’t allow it.

pepelsbey:
@hamishwillee good points! Let me know if [it has gotten better](https://github.com/mdn/content/pull/43283/commits/c8185dbfa0b608278777f91a32686dca5a2527d1).

Unfortunately, no live examples here 🥲

hamishwillee:
> Unfortunately, no live examples here 🥲

Boo. Hiss. Thanks for explaining.

files/en-us/web/api/notification/navigate/index.md
```diff
@@ -0,0 +1,63 @@
...
+
+The **`navigate`** read-only property of the {{domxref("Notification")}} interface contains the URL the user agent will navigate to when the user activates the notification.
+
+This is the resolved value of the URL, if any, that was specified in the `navigate` option of the {{domxref("Notification.Notification", "Notification()")}} constructor or {{domxref("ServiceWorkerRegistration.showNotification()")}}.
+
+When a notification with a navigation URL is activated by the user, the user agent navigates to the specified URL instead of firing an event on the notification. Normally, activating a non-persistent notification fires the {{domxref("Notification.click_event", "click")}} event on its {{domxref("Notification")}} object, and activating a persistent notification fires the {{domxref("ServiceWorkerGlobalScope.notificationclick_event", "notificationclick")}} event on the {{domxref("ServiceWorkerGlobalScope")}}. Setting a navigation URL bypasses both, allowing notifications to direct users to a specific page without requiring an event handler.
```
hamishwillee: Yes, this really clear thanks. I'm splitting it by sentence, because I find that easier to review. 

```suggestion
When a notification with a navigation URL is activated by the user, the user agent navigates to the specified URL instead of firing an event on the notification.
Normally, activating a non-persistent notification fires the {{domxref("Notification.click_event", "click")}} event on its {{domxref("Notification")}} object, and activating a persistent notification fires the {{domxref("ServiceWorkerGlobalScope.notificationclick_event", "notificationclick")}} event on the {{domxref("ServiceWorkerGlobalScope")}}.
Setting a navigation URL bypasses both, allowing notifications to direct users to a specific page without requiring an event handler.
```

files/en-us/web/api/notification/navigate/index.md
```diff
@@ -0,0 +1,65 @@
...
+---
+
+{{APIRef("Web Notifications")}}{{securecontext_header}}{{SeeCompatTable}} {{AvailableInWorkers}}
+
+The **`navigate`** read-only property of the {{domxref("Notification")}} interface contains the URL the user agent will navigate to when the user activates the notification.
+
+This is the resolved value of the URL, if any, that was specified in the `navigate` option of the {{domxref("Notification.Notification", "Notification()")}} constructor or {{domxref("ServiceWorkerRegistration.showNotification()")}}.
+
+When a notification with a navigation URL is activated by the user, the user agent navigates to the specified URL instead of firing an event on the notification.
+Normally, activating a non-persistent notification fires the {{domxref("Notification.click_event", "click")}} event on its {{domxref("Notification")}} object, and activating a persistent notification fires the {{domxref("ServiceWorkerGlobalScope.notificationclick_event", "notificationclick")}} event on the {{domxref("ServiceWorkerGlobalScope")}}.
```
hamishwillee, line 17: As I said before this is much more clear. It did get me wondering what a persistent notifications was - in particular what that means for a window vs a servicew roker.

You might think this is out of scope, but I think we should do some updates, because this is not at all clear anywhere in the top level.
- https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API#interfaces 
  - `click` is not even mentioned - just [notificationclick](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/notificationclick_event)
  - This navigate url isn't mentioned
- https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API#notification_events 
  - doesn't include `notificationclick_event` but does include click.
  - The difference between persistent and non persistent isn't captured.

Does this make sense?

hamishwillee: PS at this point you'd hopefully link through to persistent/non-persistant definitions. 
I'd be OK with merging this and doing the other work as a post process. We should minimally file and issue

pepelsbey:
@hamishwillee I added a few fixed. Please let me know if it got better!

(sorry for delay, went on a quick vacation)

files/en-us/web/api/notifications_api/index.md
```diff
@@ -31,24 +37,48 @@ This will spawn a request dialog, along the following lines:
...
+  Their lifetime is tied to the lifetime of the page — if the page is closed, the notification can no longer be interacted with.
+
+  They are created using the {{domxref("Notification.Notification","Notification()")}} constructor and fire events such as {{domxref("Notification/click_event", "click")}} directly on the `Notification` instance
 
-In addition, the Notifications API spec specifies a number of additions to the [ServiceWorker API](/en-US/docs/Web/API/Service_Worker_API), to allow service workers to fire notifications.
+- **Persistent notifications** are created from a service worker, and can remain interactive beyond the lifetime of an individual page.
 
-> [!NOTE]
-> To find out more about using notifications in your own app, read [Using the Notifications API](/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API).
+  They are created using {{domxref("ServiceWorkerRegistration.showNotification()")}} from a service worker and fire {{domxref("ServiceWorkerGlobalScope/notificationclick_event", "notificationclick")}} and {{domxref("ServiceWorkerGlobalScope/notificationclose_event", "notificationclose")}} events on the {{domxref("ServiceWorkerGlobalScope")}}.
```
hamishwillee, line 72: @pepelsbey Your changes were good, but didn't address a problem, which is that we refer to persistant and non-persistent notifications but don't really define them. To address this I updated the Concepts section in https://github.com/mdn/content/pull/43283/changes/2d9ce43dad8d22ccb8a103170c03da2b26071814

This is kind of a minimal update. Had I time I'd go through the user guide and also show how to use the click handlers and so on.

Anyway, 
1. Please check you are happy with this update.
2. If you are, please link to the heading from the first use of persistent/non-persistent in other cases.

If you are happy then I think we can merge after that.

pepelsbey: Yeah, now I see what you meant. It looks better now. Thank you!

pepelsbey:
> Awesome. Thanks very much.

@hamishwillee Thanks for the suggestion! Although I don't think we should be rewriting each other's formatting style (a sentence per line, etc.) since both are commonly used on MDN.

I'd rather stick to one style, but it would take a lot of reformatting and probably not be worth it.

hamishwillee:
> Although I don't think we should be rewriting each other's formatting style (a sentence per line, etc.) since both are commonly used on MDN.

Sorry. Objectively we shouldn't - it makes the git blame less useful for tracking content changes. I just find it so much easier to review and read. Wish I could convince you all I'm right :-(

PR #43285, opened 2026-02-27
docs(Intl): Remove other Intl objects that are not top-level

tj-commits:
Specific Intl properties are not their own standard/global objects on the top-level scope, so removed them, just leaving global Intl.


Josh-Cena:
Thanks, this is intentional, because `Intl` itself is useless, while its members themselves are not trivial either, so this is slightly different from `Math` or `Atomics`.

PR #43287, opened 2026-02-27
Improve wording

vmxarya:
Made a change for better understanding between static and dynamic websites

### Description
making an explanation more understandable, and now I can better understand why sites need to be dynamic ( because clients made a change to them ! )

### Motivation,
when I was reading it, I didn't see myself in it! 

### Additional details
The dynamic part of the explanation is too technical; I didn't understand it when I first read it. It uses a phrase twice without providing any understanding of what the user is doing here.

### DIFF
--- a/files/en-us/learn_web_development/getting_started/web_standards/the_web_standards_model/index.md
+++ b/files/en-us/learn_web_development/getting_started/web_standards/the_web_standards_model/index.md
@@ -174,7 +174,7 @@ These technologies are not considered to be "web standards" — they are develop
 Another way that client-side and server-side languages are often described is **static** and **dynamic**:
 
 - A plain HTML file is stored on the server. When requested, it is delivered to the client, unchanged, and rendered by the browser. Because it does not change, it is referred to as "static".
-- When server-side code (for example, a Python script or an ASP.NET page) generates some HTML containing data and returns that HTML to the client, the contents of the HTML change depending on what the server-side code does. It is therefore referred to as "dynamic".
+- When server-side code (for example, a Python script or an ASP.NET page) generates some HTML containing data and returns that HTML to the client, the contents of the HTML change depending on what the client wants. It is therefore referred to as "dynamic".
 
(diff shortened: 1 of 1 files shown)


files/en-us/learn_web_development/getting_started/web_standards/the_web_standards_model/index.md
```diff
@@ -174,7 +174,7 @@ These technologies are not considered to be "web standards" — they are develop
 Another way that client-side and server-side languages are often described is **static** and **dynamic**:
 
 - A plain HTML file is stored on the server. When requested, it is delivered to the client, unchanged, and rendered by the browser. Because it does not change, it is referred to as "static".
-- When server-side code (for example, a Python script or an ASP.NET page) generates some HTML containing data and returns that HTML to the client, the contents of the HTML change depending on what the server-side code does. It is therefore referred to as "dynamic".
+- When server-side code (for example, a Python script or an ASP.NET page) generates some HTML containing data and returns that HTML to the client, the contents of the HTML change depending on what the client wants. It is therefore referred to as "dynamic".
```
chrisdavidmills:
```suggestion
- Server-side code (such as a Python script or an ASP.NET page) generates HTML content that varies depending on what data is passed to it, and then returns that HTML to the client. It is therefore referred to as "dynamic". For example, the same weather forecast page can show different data depending on whether the weather is sunny or rainy, what location the user is in, and what their preferences are (some users may want to see pollen count and humidity, whereas others might not care about that data and might select preferences to hide it).
```

chrisdavidmills (changes requested):
@vmxarya I agree that the sentence wasn't very well-written; thanks for bring this up!

I've suggested a better fix, as I thought your suggestion was still unclear. I've simplified the existing wording, and also added a specific example. Let me know what you think.

Josh-Cena:
@chrisdavidmills Perhaps you can make any change as you see fit so we can get this through. It's pure editorial preference anyway.

chrisdavidmills (approved):
@Josh-Cena, yeah, let's just get this one in.

PR #43300, opened 2026-03-01
Correct displaying event properties example to do what's claimed

whxvd:
The last example (Displaying event object properties) is meant to "to display all the properties of the onload event object and their values". But the example did not register the appropriate event handler; instead it had an erroneous last line.

### DIFF
--- a/files/en-us/web/api/document_object_model/index.md
+++ b/files/en-us/web/api/document_object_model/index.md
@@ -587,13 +587,12 @@ td {
 ```
 
 ```js
-function showEventProperties(e) {
+function showEventProperties(event) {
   function addCell(row, text) {
     const cell = row.insertCell(-1);
     cell.appendChild(document.createTextNode(text));
   }
 
-  const event = e || window.event;
   document.getElementById("eventType").textContent = event.type;
 
   const table = document.createElement("table");


files/en-us/web/api/document_object_model/index.md
```diff
@@ -619,7 +619,7 @@ function showEventProperties(e) {
   document.body.appendChild(table);
 }
 
-showEventProperties(event);
+onload = showEventProperties;
```
Josh-Cena:
```suggestion
window.addEventListener("click", showEventProperties);
```

whxvd: Ok, that would introduce three changes:

1. `window.addEventListener` instead of the `onload` property. That is fine. After all, `addEventListener` is the recommended way. I had used the `onload` property, because currently many pages on MDN use them, and I had thought that for brevity or in examples, event properties may be preferred.

2. Make the global object explicit (`window.`). That's fine, too.

3. `"click"` instead of `"load"`. That would not fit the preceding text that explains the example. IMO, that text and the load event are fine. I am in favor of sticking to the load event.

Josh-Cena: > 3\. `"click"` instead of `"load"`. That would not fit the preceding text that explains the example. IMO, that text and the load event are fine. I am in favor of sticking to the load event.

I prefer `click` because it contains non-trivial properties such as coordinates, plus it can fire many times. `load` events are also generally frowned upon in MDN because code examples used to abuse them while top-level code works just fine, so sometimes I go around and remove unnecessary `load` event listeners. It's best if we can avoid them where we can to make the auditing easier. You can update the preceding text.

whxvd: Ok, now the `click` event is used.

whxvd:
The text had a "double `domxref`" to `Window.load_event` and `event`. I found that a bit strange and switched to the simplest form. The documentation for `Event` is easily reacheble from `Element.click_event`. It does not need an extra link.

files/en-us/web/api/document_object_model/index.md
```diff
@@ -556,7 +556,7 @@ document.getElementById("t-daddy").addEventListener("click", () => {
 
 ### Displaying event object properties
 
-This example uses DOM methods to display all the properties of the {{domxref("Window.load_event", "onload")}} {{domxref("event")}} object and their values in a table. It also shows a useful technique of using a [`for...in`](/en-US/docs/Web/JavaScript/Reference/Statements/for...in) loop to iterate over the properties of an object to get their values.
+This example uses DOM methods to display all the properties of the {{domxref("Element.click_event")}} object and their values in a table. It also shows a useful technique of using a [`for...in`](/en-US/docs/Web/JavaScript/Reference/Statements/for...in) loop to iterate over the properties of an object to get their values.
```
Josh-Cena:
```suggestion
This example uses DOM methods to display all the properties of the {{domxref("Element.click_event", "click")}} event object and their values in a table. It also shows a useful technique of using a [`for...in`](/en-US/docs/Web/JavaScript/Reference/Statements/for...in) loop to iterate over the properties of an object to get their values.
```

PR #43307, opened 2026-03-02
Fix sanitizer usage in setHTMLUnsafe examples

dibery:
### Description

Fix sanitizer usage in `setHTMLUnsafe` examples. Current example doesn't work.

### Motivation

Fix usage example so future readers won't have to spend time figuring out why the example code doesn't work.

### Additional details

There are problems in these examples in the following sections:
https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API#sanitization_methods
https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API#using_an_allow_sanitizer_configuration

No.


Josh-Cena:
Sorry, I don't understand what this is trying to fix. The examples mean exactly what they represent.

PR #43308, opened 2026-03-02
Added missing dot

verheyenkoen:
Added missing end-of-sentence dot at line 30 to fix this:
image

### Description

One sentence was missing is missing a dot at the end. In markdown the next sentence is on the next line but in rendered HTML it comes right after which makes it odd.

### Motivation

Sentences are meant to be finished with a dot character.


chrisdavidmills (approved):
Thanks a lot for the fix, @verheyenkoen!

While I was here, I decided to make a couple of additional fixes.

PR #43314, opened 2026-03-02
WebDriver: Move Capabilities and Errors out of Classic

dipikabh:
### Description

Right now linking from BiDi pages to `Reference/Classic/Capabilities/` and `Reference/Classic/Errors/` is incorrect - it implies those resources belong to Classic. A shared `Reference/Capabilities/`  and `Reference/Errors/ ` location is more appropriate.

Eventually, each capability/error page can note if it is Classic-only, BiDi-only, or both, and if behavior is different across the two protocols.

This PR:
- Moves Capabilities and Errors to a shared location under `Reference`.
  (used `npm run content move` and `node scripts/update-moved-file-links.js`)
- Updates the sidebar (`webdriver.yaml`), so that Capabilities and Errors are now top-level entries alongside BiDi and Classic.

### Sidebar screenshots

| Before | After |
|--------|--------|
| <kbd>Screenshot 2026-03-02 at 9 46 20 AM</kbd> |<kbd>Screenshot 2026-03-02 at 9 46 01 AM</kbd> |

### DIFF
--- a/files/sidebars/webdriver.yaml
+++ b/files/sidebars/webdriver.yaml
@@ -3,12 +3,9 @@
 sidebar:
   - type: section
     link: /Web/WebDriver
-    title: WebDriver
   - type: section
     link: /Web/WebDriver/Reference
-    title: Reference
   - link: /Web/WebDriver/Reference/BiDi
-    title: BiDi
     details: open
     children:
       - type: listSubPages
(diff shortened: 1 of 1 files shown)


dipikabh:
Thanks for reviewing!

PR #43315, opened 2026-03-02
Fluidify wording of type of parseFromString `input` parameter

immjs:
Saying "string instance" might sound like the value needs to be an instance of the `String` class. One can quickly disprove this hypothesis by looking further in the document, nonetheless I still believe the proposed wording could eliminate some possible confusion.

### Description
I changed the wording of `parseFromString`'s `input` parameter to be described not as a "TrustedHTML or string instance" but as a "TrustedHTML instance or a string".

### Motivation

A "string instance" could remind some readers of the `String` constructor. This hypothesis is quickly disproven by looking at the usage examples of this function, nonetheless I believe the proposed change could discourage some confusion

### DIFF
--- a/files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
+++ b/files/en-us/web/api/shadowroot/sethtmlunsafe/index.md
@@ -30,7 +30,7 @@ setHTMLUnsafe(input, options)
 ### Parameters
 
 - `input`
-  - : A {{domxref("TrustedHTML")}} or string instance defining HTML to be parsed.
+  - : A {{domxref("TrustedHTML")}} instance or a string defining HTML to be parsed.
 - `options` {{optional_inline}}
   - : An options object with the following optional parameters:
     - `sanitizer` {{optional_inline}}


wbamberg:
There are lots of cases of this pattern, most (all?) of which can be found through the links listed here: https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API#injection_sink_interfaces. 

Note that many of them do already use "A [TrustedHTML](https://developer.mozilla.org/en-US/docs/Web/API/TrustedHTML) instance or string", for example [Element.setHTMLUnsafe()](https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTMLUnsafe#input), but some others say "string instance", such as [Document.parseHTMLUnsafe()](https://developer.mozilla.org/en-US/docs/Web/API/Document/parseHTMLUnsafe_static#input).

cc @hamishwillee .

immjs:
Ah alright, I understand; I was more asking in case it was a mistake. Apologies for the inconvenience :')

wbamberg:
Sorry I wasn't clear. I didn't mean your change isn't good. I think it is good (personally, although I'd be happy to go with what Hamish thinks). I'm just saying that if we want to make the change, there are some other pages that should have the same update.

So if you don't mind I think we should reopen the PR :).

hamishwillee:
Thanks @wbamberg Yes, I prefer this " A {{domxref("TrustedHTML")}} instance or a string ...". They should all have been this way.

@immjs Would you be willing to fix the other cases?

immjs:
Yes, of course! Give me a minute

immjs:
There are no more instances of "string instance" in the PR now :tada:

hamishwillee:
Thanks very much from me too!

PR #43318, opened 2026-03-02
Fix accidentally duplicated CSSMathValue description

nightpool:
### Description

The description for CSSMathSum was carried over to CSSMathProduct. This fixes that by creating an appropriate description for CSSMathProduct.

As a drive-by fix, I also made the description for CSSMathInvert clearer by explaining how it's used

### Motivation

I read the docs and saw that they were inaccurate! 

### Additional details

https://drafts.css-houdini.org/css-typed-om/#dom-cssnumericvalue-mul

I don't see any related issues.

### DIFF
--- a/files/en-us/web/api/css_typed_om_api/index.md
+++ b/files/en-us/web/api/css_typed_om_api/index.md
@@ -73,11 +73,11 @@ The {{domxref('CSSKeywordValue')}} interface of the CSS Typed Object Model API c
   - : An interface which creates an object to represent CSS keywords and other identifiers. When used where a string is expected, it will return the value of `CSSKeyword.value`.
 - {{domxref('CSSMathValue')}}
   - : A tree of subclasses representing numeric values that are more complicated than a single value and unit, including:
-    - {{domxref('CSSMathInvert')}} - represents a CSS {{cssxref("calc","calc()")}} value used as `calc(1 / <value>)`.
     - {{domxref('CSSMathMax')}} - represents the CSS {{cssxref("max","max()")}} function.
     - {{domxref('CSSMathMin')}} - represents the CSS {{cssxref("min","min()")}} function.
     - {{domxref('CSSMathNegate')}} - negates the value passed into it.
-    - {{domxref('CSSMathProduct')}} - represents the result obtained by calling {{domxref('CSSNumericValue.add','add()')}}, {{domxref('CSSNumericValue.sub','sub()')}}, or {{domxref('CSSNumericValue.toSum','toSum()')}} on {{domxref('CSSNumericValue')}}.
+    - {{domxref('CSSMathInvert')}} - represents a CSS {{cssxref("calc","calc()")}} value used as `calc(1 / <value>)`. This type is used internally by {{domxref('CSSNumericValue.div','div()')}}, to create an appropriate {{domxref('CSSMathProduct')}}.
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/css_typed_om_api/index.md
```diff
@@ -73,11 +73,11 @@ The {{domxref('CSSKeywordValue')}} interface of the CSS Typed Object Model API c
...
 - {{domxref('CSSMathValue')}}
   - : A tree of subclasses representing numeric values that are more complicated than a single value and unit, including:
-    - {{domxref('CSSMathInvert')}} - represents a CSS {{cssxref("calc","calc()")}} value used as `calc(1 / <value>)`.
     - {{domxref('CSSMathMax')}} - represents the CSS {{cssxref("max","max()")}} function.
     - {{domxref('CSSMathMin')}} - represents the CSS {{cssxref("min","min()")}} function.
     - {{domxref('CSSMathNegate')}} - negates the value passed into it.
-    - {{domxref('CSSMathProduct')}} - represents the result obtained by calling {{domxref('CSSNumericValue.add','add()')}}, {{domxref('CSSNumericValue.sub','sub()')}}, or {{domxref('CSSNumericValue.toSum','toSum()')}} on {{domxref('CSSNumericValue')}}.
+    - {{domxref('CSSMathInvert')}} - represents a CSS {{cssxref("calc","calc()")}} value used as `calc(1 / <value>)`. This type is used internally by {{domxref('CSSNumericValue.div','div()')}}, to create an appropriate {{domxref('CSSMathProduct')}}.
```
chrisdavidmills, line 79:
```suggestion
    - {{domxref('CSSMathInvert')}} - represents the value returned by {{cssxref("calc","calc(1 / <value>)")}}. This type is used internally by {{domxref('CSSNumericValue.div','div()')}} to create an appropriate {{domxref('CSSMathProduct')}}.
```

The phrasing seemed a bit odd here. Do you think this works?

chrisdavidmills (changes requested):
Thanks for the fix, @nightpool. I've suggested one minor tweak to the `CSSMathInvert` wording, but other than that, I think this is fine.

One other stretch suggestion would be that it would be good to turn the bulleted lists into nested DLs instead, which has the advantage that it turns all the DTs into links, so all the items can be linked. Do you fancy doing this? If not, I'm happy to update the PR.

files/en-us/web/api/css_typed_om_api/index.md
nightpool: It probably works. The reason I didn't reword it is because the way I read the spec, `1 / <value>` would actually be represented by `CSSMathProduct(CSS.number(1), CSSMathInvert(value))`, which is a little weird, so I sought to thread that needle here by keeping the "used as" language and then adding a clarifying example. It could be argued that the product + invert distinction is too granular for a basic overview though.

nightpool: note that the same "used as" language appears on the CSSMathInvert standalone doc page, so if we update it here it should probably be updated there as well.

chrisdavidmills: OK, we'll leave this for now.

chrisdavidmills (approved):
@nightpool, thanks for the explanation. Approving!

PR #43319, opened 2026-03-03
FF149 Relnote: HTMLInputElement.showPicker() support for a <datalist>

hamishwillee:
FF149 supports `HTMLInputElement.showPicker()` for a `<datalist>` in https://bugzilla.mozilla.org/show_bug.cgi?id=1998668

This adds a release note and a note in the document about why we don't have a live example.

Related docs work can be tracked in #43201

### DIFF
--- a/files/en-us/web/api/htmlinputelement/showpicker/index.md
+++ b/files/en-us/web/api/htmlinputelement/showpicker/index.md
@@ -135,7 +135,7 @@ button.addEventListener("click", () => {
 });
 ```
 
-A for the other pickers we can't show this code running as a live example because it runs in a cross-origin frame, and would cause a [`SecurityError`](#securityerror).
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/htmlinputelement/showpicker/index.md
```diff
@@ -135,6 +135,8 @@ button.addEventListener("click", () => {
 });
 ```
 
+A for the other pickers we can't show this code running as a live example because it runs in a cross-origin frame, and would cause a [`SecurityError`](#securityerror).
```
pepelsbey:
```suggestion
As for the other pickers, we can't show this code running as a live example because it runs in a cross-origin frame, and would cause a [`SecurityError`](#securityerror).
```

hamishwillee: Doh. Thank you.

PR #43324, opened 2026-03-03
Remove incorrect detail for drop-shadow syntax

mateuszJS:
### Description

Misleading information that filter: drop-shadow uses `0` for offset length if value is not provided. `drop-shadow` with one `<length>` is invalid. Additionally, on the start of the paragraph it's specified that 2 or 3 values are accepted, so it's impossible to omit x or y offset. In my opinion, the paragraph without that sentence describes the behaviour in a bit cleaner way.

### Additional details

I assume it was copy-pasted from [spec](https://www.w3.org/TR/filter-effects-1/#funcdef-filter-drop-shadow):
`The default value for omitted values is missing length values set to 0 and the missing used color is taken from the color property.`. From the spec perspective it might be technically correct that `0` is used by default and it's later interpreted as no gaussian blur but from developer friendly docs it introduces a bit more confusion than help.

Let me know if you agree.

### DIFF
--- a/files/en-us/mozilla/firefox/experimental_features/index.md
+++ b/files/en-us/mozilla/firefox/experimental_features/index.md
@@ -346,20 +346,6 @@ The {{domxref("CloseWatcher")}} interface allows developers to implement UI comp
 - `dom.closewatcher.enabled`
   - : Set to `true` to enable.
 
-### HTML Sanitizer API
-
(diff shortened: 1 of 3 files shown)


hamishwillee (approved):
@mateuszJS - I completely agree. Thank you.

Issue #43297, opened 2026-02-28
Promise can't resolve multiple times


ottoville:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/ImageDecoder/decode

### What specific section or headline is this issue about?

completeframesonly

### What information was incorrect, unhelpful, or incomplete?

The phrase:

“the promise returned by decode() will resolve exactly once for each new level of detail”

reads like one promise resolves multiple times.

But that would violate JavaScript semantics:

- A Promise settles once.
- After resolution, it cannot resolve again.
- It cannot “emit” multiple values.

### What did you expect to see?

Each call to decode() returns a promise that resolves once, and if additional detail becomes available later, calling decode() again will return another promise that resolves when the next level is ready.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/api/imagedecoder/decode`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/ImageDecoder/decode
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/imagedecoder/decode/index.md
* Last commit: https://github.com/mdn/content/commit/a7482281c4570bb7f932dce381f510d87ddf9924
* Document last modified: 2024-10-31T03:19:16.000Z

</details>

dataCenter430:
Hi, @Josh-Cena
I'd love to take this issue, if needed.
are you able to assign it to me?
Best regards

Josh-Cena:
We don't assign. Just make the PR.

PR #43326, opened 2026-03-03
Fix decode() promise wording: clarify one promise per call

dataCenter430:
### Description
Updates the ImageDecoder.decode() docs so the completeFramesOnly parameter is described in line with JavaScript promise semantics. The text now states that each call to decode() returns a new promise that settles once, and that further levels of detail are obtained by calling decode() again. Also fixes a typo in the progressive-decoding example comment and uses the correct variable name (imageDecoder) in the example.

### Motivation
The previous wording (“the promise returned by decode() will resolve exactly once for each new level of detail”) suggested that a single promise could resolve multiple times, which contradicts how promises work in JavaScript. A promise settles only once (either fulfilled or rejected). The change clarifies that multiple resolutions correspond to multiple calls to decode(), each returning its own promise.

### Additional details
Parameter description: Reworded so it’s explicit that “each call to decode() returns a new promise that resolves once” and that “call decode() again to receive further levels as they become available.”
Example: Comment typo “is won’t” → “won’t”; imageDecode → imageDecoder for consistency with the first example.

Fixes https://github.com/mdn/content/issues/43297 (Promise can't resolve multiple times)

### DIFF
--- a/files/en-us/web/api/imagedecoder/decode/index.md
+++ b/files/en-us/web/api/imagedecoder/decode/index.md
@@ -24,7 +24,10 @@ decode(options)
     - `frameIndex` {{optional_inline}}
       - : An integer representing the index of the frame to decode. Defaults to `0` (the first frame).
     - `completeFramesOnly` {{optional_inline}}
-      - : A {{jsxref("boolean")}} defaulting to `true`. When `false` indicates that for progressive images the decoder may output an image with reduced detail. When `false`, each call to `decode()` returns a new promise that resolves once with the next available level of detail; call `decode()` again to receive further levels as they become available.
+      - : A {{jsxref("boolean")}} defaulting to `true`.
+        When `true`, the `Promise` returned by the method resolves only when the image is fully decoded.
+        When `false`, the method will return a new `Promise` that may resolve with a partially decoded image.
+        The method can be called repeatedly until `result.complete` is true, with each step providing an image with the next available level of detail.
 
 ### Return value
 


dataCenter430:
Hi, @hamishwillee  @Josh-Cena 
could you pls review the PR?
Best regards

hamishwillee:
@dataCenter430 Thanks! Please see this text in https://developer.mozilla.org/en-US/docs/MDN/Community/Issues#general_triaging_tasks

> If a pull request has been opened to fix the issue but has not been reviewed for **a week**, give the reviewer a gentle ping to ask if they can get to it.

It happens we're a small team with many priorities. We really appreciate the contributions and we'll get to them as soon as priorities allow. A nudge after a week is appreciate though, as longer might mean I missed something.

dataCenter430:
Hi, @hamishwillee 
Thanks for your review.
I'm happy to contribute to content, 😄 
Thanks again.

files/en-us/web/api/imagedecoder/decode/index.md
```diff
@@ -24,7 +24,7 @@ decode(options)
     - `frameIndex` {{optional_inline}}
       - : An integer representing the index of the frame to decode. Defaults to `0` (the first frame).
     - `completeFramesOnly` {{optional_inline}}
-      - : A {{jsxref("boolean")}} defaulting to `true`. When `false` indicates that for progressive images the decoder may output an image with reduced detail. When `false`, the promise returned by `decode()` will resolve exactly once for each new level of detail.
+      - : A {{jsxref("boolean")}} defaulting to `true`. When `false` indicates that for progressive images the decoder may output an image with reduced detail. When `false`, each call to `decode()` returns a new promise that resolves once with the next available level of detail; call `decode()` again to receive further levels as they become available.
```
hamishwillee: Thanks @dataCenter430 

Do you think it worth being more complete? Something like

```suggestion
      - : A {{jsxref("boolean")}} defaulting to `true`.
        When `true`, the `Promise` returned by the method resolves only when the image is fully decoded.
        When `false`, the method will return a new `Promise` that may resolve with a partially decoded image.
        The method can be called repeatedly until `result.complete` is true, with each step providing an image with the next available level of detail.
```

dataCenter430: Thanks for your clarification, @hamishwillee  
I agree more detail would help here. 
I think it will work for users
thanks again, 🙏

PR #43327, opened 2026-03-03
Editorial review: Add information on clipboardchange event permissions

chrisdavidmills:
### Description

From Chrome 145, firing `clipboardchange` events requires sticky user activation or `clipboard-read` permission, to prevent unauthorized clipboard monitoring. See https://chromestatus.com/feature/5163741220044800.

This PR adds notes to relevant pages to inform the reader of these requirements. I don't think much more work is required, as the `clipboardchange` example works as-is (permission is requested when the user tries to copy text.) This is mostly just about making the permission requirements more explicit for the event.

See also https://github.com/mdn/browser-compat-data/pull/29170 for the related BCD change.

### DIFF
--- a/files/en-us/web/api/clipboard/clipboardchange_event/index.md
+++ b/files/en-us/web/api/clipboard/clipboardchange_event/index.md
@@ -11,7 +11,7 @@ browser-compat: api.Clipboard.clipboardchange_event
 
 The **`clipboardchange`** event of the {{domxref("Clipboard")}} interface is fired when the system clipboard contents are changed in any way, for example via a system copy command, or via an API method such as {{domxref("Clipboard.writeText()")}}.
 
-Firing `clipboardchange` events requires [sticky activation](/en-US/docs/Glossary/Sticky_activation) or the `clipboard-read` permission to be granted.
+The `clipboardchange` event is only fired with [sticky activation](/en-US/docs/Glossary/Sticky_activation) or after the `clipboard-read` permission is granted.
 
 ## Syntax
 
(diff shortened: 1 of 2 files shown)


files/en-us/web/api/clipboard/clipboardchange_event/index.md
```diff
@@ -11,6 +11,8 @@ browser-compat: api.Clipboard.clipboardchange_event
 
 The **`clipboardchange`** event of the {{domxref("Clipboard")}} interface is fired when the system clipboard contents are changed in any way, for example via a system copy command, or via an API method such as {{domxref("Clipboard.writeText()")}}.
 
+Firing `clipboardchange` events requires [sticky activation](/en-US/docs/Glossary/Sticky_activation) or the `clipboard-read` permission to be granted.
```
dipikabh: - should this instead be phrased as "The `clipboardchange` event is fired only when..."?
- should this perhaps be a note callout for chrome behavior (given that in BCD, the note is specific to Chrome)?

files/en-us/web/api/clipboard_api/index.md
```diff
@@ -75,6 +75,7 @@ Chromium browsers:
 - Writing requires either the `clipboard-write` permission or transient activation.
   If the permission is granted, it persists, and further transient activation is not required.
 - The HTTP [Permissions-Policy](/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy) permissions `clipboard-read` and `clipboard-write` must be allowed for {{HTMLElement("iframe")}} elements that access the clipboard.
+- Firing {{domxref("Clipboard.clipboardchange_event", "clipboardchange")}} events requires [sticky activation](/en-US/docs/Glossary/Sticky_activation) or the `clipboard-read` permission to be granted.
```
dipikabh: same comments if fixing in the other file

chrisdavidmills: On this page, I've taken the bit about the `clipboardchange` event out of the Chromium-specific details list and put it above, as it is spec-mandated behavior and not divergent from the spec.

files/en-us/web/api/clipboard/clipboardchange_event/index.md
chrisdavidmills: I've rephrased the line as suggested.

At the moment, this is Chromium-specific behavior because only Chromium browsers do this, but it is spec-compliant behavior. Maybe at some point, I need to add BCD notes to say that Firefox and Safari don't do this?

PR #43334, opened 2026-03-04
Remove SAFE_FOR_JQUERY usage in DOMPurify example

xmcgcg:
DOMPurify [removed](https://github.com/cure53/DOMPurify?tab=readme-ov-file#removed-configuration) the `SAFE_FOR_JQUERY` flag in [2.1.0](https://github.com/cure53/DOMPurify/releases/tag/2.1.0) (released in 2020) and claimed that it is “safe by default now for jQuery” since that version.


rebloor (approved):
Thanks @xmcgcg

PR #43343, opened 2026-03-05
[Editorial review] BiDi - add pages for session module's subscribe, unsubscribe commands and a how-to

dipikabh:
### Description

This PR is a continuation of https://github.com/mdn/content/pull/43251 to document the remaining commands of the session module:

- `session.subscribe`
- `session.unsubscribe`

It also adds a how-to page to explain creating a WebDriver BiDi connection.

Additionally, it updates the [WebDriver landing page](https://developer.mozilla.org/en-US/docs/Web/WebDriver) to rewrite the intro in simpler language, remove the usage example, and add BiDi to the Reference section.

### Additional details

The pages in this PR cross-link to `session.new`, `session.end`, `session` module, which are included in #43251.
Those links will resolve once that PR merges.

### Related issue

Doc issue: https://github.com/mdn/mdn/issues/339


pepelsbey (approved):
Looks good! Thank you 🙂

files/en-us/web/webdriver/how_to/index.md
```diff
@@ -0,0 +1,11 @@
+---
+title: WebDriver how to
+short-title: How to
+slug: Web/WebDriver/How_to
+page-type: listing-page
+sidebar: webdriver
+---
+
+This page lists how-to guides for WebDriver.
```
whimboo, line 9: Do you already have in mind how you want to structure the page by differentiating between WebDriver Classic and BiDi?

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
```diff
@@ -0,0 +1,51 @@
...
+slug: Web/WebDriver/How_to/Create_BiDi_connection
+page-type: how-to
+sidebar: webdriver
+---
+
+The client and the browser communicate using the WebDriver BiDi protocol over a WebSocket connection. This article explains how to establish that connection.
+
+## Launching the browser with remote debugging enabled
+
+To use WebDriver BiDi, you need to enable it in the browser by launching it with the `--remote-debugging-port` flag. The browser then listens for incoming WebSocket connections on the specified port. Port `9222` is a conventional default for browser debugging, but you can use any available port.
```
whimboo: Just for clarity the initial connection is based on HTTP and we run a WebSocket upgrade over the same connection when the client connects. We could just reference https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Protocol_upgrade_mechanism#upgrading_to_a_websocket_connection here.

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+On macOS, `firefox` may not be in your PATH. In that case, use the full path instead:
+
+```bash
+/Applications/Firefox.app/Contents/MacOS/firefox --remote-debugging-port 9222
+```
+
+> [!NOTE]
+> For Chrome, enabling WebDriver BiDi requires a different setup using ChromeDriver with the `webSocketUrl` capability.
```
whimboo: Based on that note we should most likely split the how to in the following two sections:

1. Request to enable WebDriver BiDi via the `webSocketUrl` capability when creating a new WebDriver classic session. With the capability set to `true` the webdriver client will make sure to start the relevant browser with the extra port open, which then can be used to upgrade to a WebSocket connection. This applies to all existing browsers including Firefox with geckodriver.

2. A BiDi-only connection can be established as you described above. It is the client that have to correctly start the browser with the required command line argument set to open the connection at the given port. For Firefox no other binary is needed but Chrome and its derivates require the [Chromium BiDi wrapper](https://github.com/GoogleChromeLabs/chromium-bidi) package. A client can also specify the port as `0` which will allow the browser to use a freely chosen port by the system to avoid port collisions in case multiple Firefox processes run in parallel. To know which port has been chosen the client needs to listen to stderr and check for `WebDriver BiDi listening on ws://127.0.0.1:46249` ([example from a Puppeteer job](https://treeherder.mozilla.org/logviewer?job_id=552877707&repo=mozilla-central&task=ePt8jsAIRuSazCRh-jMb2w.0&lineNumber=406)).

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Getting the WebSocket URL
```
whimboo, line 29: As pointed out in the other comment the real ws URL can be scraped from the stderr output in case of a BiDi-only connection.

If there is a BiDi-upgrade via a classic session the WebSocket URL is returned in the `webSocketUrl` capability. Note that in this case we return a URL specific to the WebDriver session as created by WebDriver classic.

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
```diff
@@ -0,0 +1,51 @@
...
+For example, if you launched Firefox with `--remote-debugging-port 9222`, the URL is `ws://127.0.0.1:9222/session`.
+
+> [!NOTE]
+> Firefox uses `127.0.0.1` rather than `localhost` for the BiDi WebSocket endpoint.
+
+## Connecting to the WebSocket
+
+With the WebSocket URL, use any WebSocket client to open a connection. Common options include the `ws` package for Node.js and the `websockets` package for Python.
+
+Once connected, you can send WebDriver BiDi commands as JSON messages and receive responses and events from the browser. See [`session.new`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/new) to create a BiDi session after connecting.
```
whimboo: Feel free to as well reference our online WebDriver BiDi demo client.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Parameters
+
+The input is an object with the following fields:
+
+- `events`
+  - : An array of one or more strings that specifies event names for subscription.
+    Each string can be either a specific event name (for example, `"log.entryAdded"`) or a module name (for example, `"log"`) that subscribes the client to all events in that module.
+- `contexts` {{optional_inline}}
+  - : An array of one or more browsing context IDs, each corresponding to a tab or frame.
```
whimboo: In terms of the current HTML spec these ids are actually for [Navigables](https://html.spec.whatwg.org/#navigables). We started our implementation before that term was added, so we weren't able to change the name of the parameter. As such we call it `context` but not `browsing context` internally.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
```diff
@@ -0,0 +1,119 @@
...
+    If specified, events are received only for those contexts and their descendants.
+    This field cannot be used if `userContexts` is also specified.
+- `userContexts` {{optional_inline}}
+  - : An array of one or more user context IDs, each corresponding to a browser profile or container.
```
whimboo:
```suggestion
  - : An array of one or more user context IDs, each corresponding to a browser context or container.
```

I wouldn't use the term `profile` here given that it would give a false indication. In terms of Chrome it is called a `Browser Context` while in Firefox we use `Container`.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
```diff
@@ -0,0 +1,119 @@
...
+    If specified, events are received only for those user contexts.
+    This field cannot be used if `contexts` is also specified.
+
+If neither `contexts` nor `userContexts` is provided, the subscription is global, so events are received for all browsing contexts.
+
+### Return value
+
+An object with the following fields:
```
whimboo:
```suggestion
An object with the following field:
```

... but if it's always used in plural form it will be fine as well.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+- `subscription`
+  - : A string that contains the unique identifier for this subscription.
+
+### Errors
+
+- [invalid argument](/en-US/docs/Web/WebDriver/Reference/Errors/InvalidArgument)
+  - : The `events` array is empty or omitted, an event name is not recognized, `contexts` or `userContexts` is provided but empty, or values for both `contexts` and `userContexts` were provided in the same request.
```
whimboo: This also includes invalid types or out of bounds values as used for any of the parameters.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Examples
+
+### Subscribing to an event globally
+
+With a WebDriver BiDi connection and an active session, send the following message to subscribe to the `log.entryAdded` event for all browsing contexts:
```
whimboo: Should we link the `WebDriver BiDi connection` to the HowTo?

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
```diff
@@ -0,0 +1,119 @@
...
+{
+  "id": 2,
+  "type": "success",
+  "result": {
+    "subscription": "c7b7b3a2-1f4b-4b4e-8a1f-2a3b4c5d6e7f"
+  }
+}
+```
+
+### Subscribing to multiple events
```
whimboo:
```suggestion
### Subscribing to multiple events globally
```

At least one other example would be nice when subscribing to a given context or user context.

whimboo (changes requested):
Thank you @dipikabh! Overall this looks great. From the technical point I have a couple of comments. Let me know if you need more details for some of them.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
```diff
@@ -0,0 +1,119 @@
...
+### Parameters
+
+The input is an object with the following fields:
+
+- `events`
+  - : An array of one or more strings that specifies event names for subscription.
+    Each string can be either a specific event name (for example, `"log.entryAdded"`) or a module name (for example, `"log"`) that subscribes the client to all events in that module.
+- `contexts` {{optional_inline}}
+  - : An array of one or more browsing context IDs, each corresponding to a tab or frame.
+    If specified, events are received only for those contexts and their descendants.
```
juliandescottes: One thing to clarify here, is that even if the command accepts frames as contexts, the subscription will be created for the top level navigable which owns the frame. 

See steps 8.1 & 8.2 of https://w3c.github.io/webdriver-bidi/#command-session-subscribe

juliandescottes (commented):
Looks great, just one comment on my side regarding the `session.subscribe` behavior.

files/en-us/web/webdriver/index.md
```diff
@@ -5,64 +5,32 @@ page-type: landing-page
...
+WebDriver is a browser automation interface that allows external programs to remotely inspect and control a browser. It is a platform- and language-neutral wire protocol, supporting both command-based automation and real-time, event-driven communication.
 
-To have the ability to write instruction sets that can be run interchangeably in many browsers on different platforms is critical to deliver a consistent experience to users. With the new wave of developments on the web platform, the increase diversity in devices and demands for real interoperability between the technologies, WebDriver provides tooling for [cross-browser testing](/en-US/docs/Learn_web_development/Extensions/Testing/Introduction).
+WebDriver provides tooling for [cross-browser testing](/en-US/docs/Learn_web_development/Extensions/Testing/Introduction), helping teams deliver consistent user experiences across many browsers and platforms. It is primarily intended to allow web authors to write tests to automate a browser from a separate controlling process, though it can also be used from scripts running within the browser itself.
```
juliandescottes: > though it can also be used from scripts running within the browser itself

I am not sure which scenario this corresponds to. There might be edge cases where you can achieve this setup but that's probably not what mainstream webdriver usage will look like and mentioning it here would be confusing.

Note: I see we previously had the following text: `but may also be used in such a way as to allow in-browser scripts to control a — possibly separate — browser.`, but similarly I think this is/was confusing and should probably not be part of the landing page.

dipikabh:
Rebased this branch on main after #43251 merging

files/en-us/web/webdriver/index.md
dipikabh: Ah ok, thanks for pointing. I've removed that bit.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
dipikabh: Got it. Added a line to make that clear.

> If the context ID corresponds to a frame, the subscription is created for the top-level context (tab) that owns the frame.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
dipikabh: Done. Added an example for a specific context: https://pr43343.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/subscribe#subscribing_to_events_scoped_to_a_tab

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
dipikabh: Yes! I've also added the link on other session command pages

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
dipikabh: I've added "invalid type".

I'm not sure I understand out-of-bound values - would these be for the IDs? We haven't defined a range for any of the values

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
dipikabh: You're right. I also applied style updates we implemented in `session.new`, `session.status` for this sentence

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
dipikabh: Updated

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
dipikabh: Noted, updated "browsing context IDs" to "context IDs".
Also updated other instances of "browsing context"

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
dipikabh: Did you mean https://juliandescottes.github.io/bidi-web-client/web/? If that's the one, we might want to copy it over to an official repo (such as under mdn) before adding a link to it.

Might make better to sense to add a link to this demo in the "See also" section of all our command pages.

Let me know what you think.

files/en-us/web/webdriver/how_to/index.md
dipikabh: I imagine the title of the guide will be good enough to highlight whether it is a Classic or BiDi guide.
Any guide in the "how_to" folder will automatically get listed here.

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
dipikabh: I've added this line to the [Connecting to the WebSocket endpoint](https://pr43343.review.mdn.allizom.net/en-US/docs/Web/WebDriver/How_to/Create_BiDi_connection#connecting_to_the_websocket_endpoint) section:

> The initial connection starts as HTTP and is automatically [upgraded to WebSocket](/en-US/docs/Web/HTTP/Guides/Protocol_upgrade_mechanism#upgrading_to_a_websocket_connection) when the client connects.

Question: Is it correct to assume that all browsers establish the connection via the HTTP upgrade?

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
dipikabh: Would it be okay to address this split in the future?

For now, I've updated the section title to clearly reflect the specific method covered here, added a mention of both the methods in the intro, and added a mention of port `0`.

> The client and the browser communicate using the WebDriver BiDi protocol over a WebSocket connection. There are two ways to establish this connection: you can manually launch the browser from the command line, or let the WebDriver framework handle the browser startup via the `webSocketUrl` capability when creating a classic WebDriver session. The second method works across browsers, including Firefox with `geckodriver` and Chrome with `ChromeDriver`.
> The following sections walk you through the manual approach.


Do you think these changes are sufficient for now and that the content is not misleading or inaccurate?

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
dipikabh: for the bidi-only connection, I've updated the sentence as:

> For example, if you launched Firefox with `--remote-debugging-port 9222`, the URL is `ws://127.0.0.1:9222/session`. If you specified port `0`, check the `stderr` output for a message like `WebDriver BiDi listening on ws://127.0.0.1:46249` to find the assigned port.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
whimboo: It was just an example which you will see in other (future) commands, eg. when we define a number from [1..100] but the client specifies `0` or `101`. Or we have a defined list of string values and the client specified an invalid one.

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
whimboo: Yes, that is the client. Julian moved it not that long ago under the firefox-devtools ownership. So it's now available at https://firefox-dev.tools/bidi-web-client/web/.

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
whimboo: Actually we always have a HTTP session first because the server in Firefox is driven by https://searchfox.org/firefox-main/source/netwerk/test/httpserver/httpd.sys.mjs. So the websocket client connects via the `ws://` URL and specifies the necessary headers as required for a websocket connection. And then internally we upgrade from HTTP to WS.

Given that these are actually internals we should probably not mention these details on MDN.

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
whimboo: The `webSocketUrl` capability is not used for handling the browser startup. It's only there to indicate that the WebDriver BiDi protocol should be enabled. Also starting the browser is always done by the appropriate client and not necessarily manually. Maybe you can update that text?

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
whimboo: That sounds good. Thanks.

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
dipikabh: Got it, I've removed that line

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
dipikabh: Oops, yes, that came out wrong. Can you please check this:

> The client and the browser communicate using the WebDriver BiDi protocol over a WebSocket connection. There are two ways a client can establish this connection.
>
> In one method, when creating a classic WebDriver session, the client can set the `webSocketUrl` capability to request BiDi be enabled; the client then starts the browser with the required port open.
>
> In the other method, the client starts the browser from the command line with the required flag and on the specified port. This method works with Firefox directly; Chromium-based browsers need the additional Chromium BiDi wrapper package. The sections in this article walk you through this method.

I've added the Chromium BiDi wrapper link in the "See also" section

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
dipikabh: Perfect! I've added the link to the "See also" sections on this page and on the [module landing page](https://pr43343.review.mdn.allizom.net/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#see_also).

dipikabh:
Thanks @whimboo for going through the updates. Could you take another look?

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
whimboo: Yes, mostly fine. Just two suggestions (feel free to use a better wording):

> Upgrade to WebDriver BiDi: When creating a classic WebDriver session, the `webSocketUrl` capability can be set to `true` to request WebDriver BiDi to be enabled; the webdriver client then starts the browser with the required WebSocket port open.

And equivalent:

> WebDriver BiDi-only: The browser gets started from the command line by passing the required flag and the wanted port. This method works with Firefox directly, but for Chromium-based browsers the additional Chromium BiDi wrapper package is needed. The sections in this article walk you through this method.

dipikabh: Sure. I've made the updates albeit with some tweaks, mainly to keep active voice in place so that the text is clear about who is doing the action (e.g., "the client sets the `webSocketUrl` capability" vs "the `webSocketUrl` capability is set",
"Chromium-based browsers require" vs "wrapper package is needed" - that phrasing prompts the question "needed by who").

I've updated the first one to:

> In one method, when creating a classic WebDriver session, the WebDriver client sets the `webSocketUrl` capability to `true` to request BiDi to be enabled; the client then starts the browser with the WebSocket port open.

And the second one to:

> In the other method, the WebDriver client starts the browser from the command line by passing the required flag and the desired port. This method works with Firefox directly; however, Chromium-based browsers require the additional Chromium BiDi wrapper package. The sections in this article walk you through this method using Firefox.

(note that I've added "using Firefox" in the last sentence to clarify upfront that the steps in this article specifically cover Firefox)

Does this look accurate?

whimboo: I think that this sounds good. Thanks.

whimboo (approved):
Thanks! Looks good to me.

@juliandescottes would you mind to skim over the additions as well in case you see something else?

files/en-us/web/webdriver/how_to/create_bidi_connection/index.md
dipikabh: Thanks for checking!

juliandescottes (approved):
Looks great, thanks @dipikabh !

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
```diff
@@ -0,0 +1,153 @@
...
+### Return value
+
+The `result` field in the response is an object with the following field:
+
+- `subscription`
+  - : A string that contains the unique identifier for this subscription.
+
+### Errors
+
+- [invalid argument](/en-US/docs/Web/WebDriver/Reference/Errors/InvalidArgument)
```
pepelsbey:
```suggestion
- [Invalid argument](/en-US/docs/Web/WebDriver/Reference/Errors/InvalidArgument)
```

files/en-us/web/webdriver/reference/bidi/modules/session/unsubscribe/index.md
```diff
@@ -0,0 +1,143 @@
...
+    Each string can be either a specific event name (for example, `"log.entryAdded"`) or a module name (for example, `"log"`) that unsubscribes the client from all events in that module.
+    Only global subscriptions can be removed using event names; those created using [`contexts`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/subscribe#contexts) or [`userContexts`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/subscribe#usercontexts) cannot be.
+
+### Return value
+
+The `result` field in the response is an empty object (`{}`).
+
+### Errors
+
+- [invalid argument](/en-US/docs/Web/WebDriver/Reference/Errors/InvalidArgument)
```
pepelsbey:
```suggestion
- [Invalid argument](/en-US/docs/Web/WebDriver/Reference/Errors/InvalidArgument)
```

pepelsbey (approved):
Looks good! Thank you 🙂

I would only suggest improving consistency with existing docs: start list items with capital letters, like other WebDriver docs do:

```md
### Errors

- [Session not created](…)

### Errors

- [Invalid session ID](…)
```

dipikabh:
Thanks for the review, Vadim! 🙌 

> I would only suggest improving consistency with existing docs: start list items with capital letters, like other WebDriver docs do:

Thanks for flagging this. The lowercase here is actually intentional - to exactly match the error string returned by the protocol (see first column in this [table of errors](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/Errors#table_of_errors)).
I plan to eventually update the existing pages to use lowercase for error strings.

But this is a good opportunity to wrap them in backticks to indicate they are literal strings. I'll do that here.

files/en-us/web/webdriver/reference/bidi/modules/session/subscribe/index.md
dipikabh:
```suggestion
- [`invalid argument`](/en-US/docs/Web/WebDriver/Reference/Errors/InvalidArgument)
```

files/en-us/web/webdriver/reference/bidi/modules/session/unsubscribe/index.md
dipikabh:
```suggestion
- [`invalid argument`](/en-US/docs/Web/WebDriver/Reference/Errors/InvalidArgument)
```

PR #43346, opened 2026-03-05
Fix backslash escape in String.raw example

guigueb:
### Description
Changed comment description to match what the outcome should be.

### Motivation
I came across this mistake while reading the docs and do not want it to confuse others.

### Additional details
The confusing part may be that the string in the comment is suppose to show what is displayed.
But showing what is displayed - 'A trailing backslash: \\' - may confuse someone who thinks that the backslash is escaping the quote character.

Or the examples confused me and they show what is being stored in the string - not what is being displayed.
In which case this change request is incorrect as 'A trailing backslash: \\\\' is being stored in the string.

Nothing is related to this that I am aware of.


guigueb:
I believe the MDN example is correct - it is showing what would be stored not what would be displayed.

PR #43353, opened 2026-03-06
Document CookieStore.set() maxAge option

chrisdavidmills:
### Description

Since version 145, Chrome has supported the [`CookieStore.set()`](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore/set) method's `maxAge` option; see https://chromestatus.com/feature/5190778418757632.

This PR documents the new option.

### DIFF
--- a/files/en-us/web/api/cookiestore/set/index.md
+++ b/files/en-us/web/api/cookiestore/set/index.md
@@ -59,7 +59,9 @@ A {{jsxref("Promise")}} that resolves with {{jsxref("undefined")}} when setting 
 - `SecurityError` {{domxref("DOMException")}}
   - : Thrown if the origin can not be {{glossary("Serialization", "serialized")}} to a URL.
 - {{jsxref("TypeError")}}
-  - : Thrown if setting the cookie with the given `name` and `value` or `options` fails.
+  - : Thrown if:
+    - Both the `expires` and `maxAge` properties are set.
+    - Setting the cookie with the given `name` and `value` or `options` fails in any other way.
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


files/en-us/web/api/cookiestore/set/index.md
```diff
@@ -34,6 +34,8 @@ Or
       - : A string containing the domain of the cookie. Defaults to `null`.
     - `expires` {{Optional_Inline}}
       - : A timestamp, given as {{glossary("Unix time")}} in milliseconds, containing the expiration date of the cookie. Defaults to `null`.
+    - `maxAge` {{Optional_Inline}}
+      - : A number representing the number of seconds until the cookie expires. A zero or negative number will expire the cookie immediately. If both `expires` and `maxAge` are set, the `set()` call fails with a `TypeError`. Defaults to `null`.
```
hamishwillee, line 38: Please add the new TypeError in the exceptions section.

files/en-us/web/api/cookiestore/set/index.md
```diff
@@ -88,21 +90,20 @@ async function cookieTest() {
 
 ### Setting a cookie with options
 
-This example sets a cookie by passing an `options` object with `name`, `value`, `expires`, and `partitioned`.
+This example sets a cookie by passing an `options` object with `name`, `value`, `maxAge`, and `partitioned`.
```
hamishwillee: This is problematic, in particular because maxAge is less supported that expires. Is there a BCD?

files/en-us/web/api/cookiestore/set/index.md
chrisdavidmills: Done!

files/en-us/web/api/cookiestore/set/index.md
chrisdavidmills: I've been going in circles about this issue a bit. You are right that it is problematic — afaics, it isn't that easy to feature detect whether a specific option from the options object is supported. For now, I've just put it back the way it was, to using `expires`.

chrisdavidmills: The BCD is at https://github.com/mdn/browser-compat-data/pull/29197.

chrisdavidmills:
> Perhaps also add an update to https://developer.mozilla.org/en-US/docs/Web/API/CookieStore#setting_cookies to note that you can also use maxAge on browsers where it is supported? (and maybe a link to the `set()` browser compat)?

I've added this.

PR #43359, opened 2026-03-06
fix: typo on volume property page

webbertakken:
### Description

Type on the volume property page

### DIFF
--- a/files/en-us/web/api/htmlmediaelement/volume/index.md
+++ b/files/en-us/web/api/htmlmediaelement/volume/index.md
@@ -13,8 +13,7 @@ which the media will be played.
 
 ## Value
 
-A double-precision floating-point value. Must fall between 0 and 1, where 0 is effectively muted
-and 1 is the loudest possible value.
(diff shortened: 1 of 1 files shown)


webbertakken:
This should be good to go once the last workflows are approved.

files/en-us/web/api/htmlmediaelement/volume/index.md
```diff
@@ -13,8 +13,8 @@ which the media will be played.
 
 ## Value
 
-A double values must fall between 0 and 1, where 0 is effectively muted and 1 is the
-loudest possible value.
+A double-precision floating-point value. Must fall between 0 and 1, where 0 is effectively muted
+and 1 is the loudest possible value.
```
hamishwillee: Thanks for this fix. However the underlying text is wrong as it uses spec language. JavaScript just has numbers. In MDN we (should) do this like so:


```suggestion
A number between 0 and 1, where 0 is effectively muted, and 1 is the loudest possible value.
```

hamishwillee (approved):
Thanks @webbertakken, and welcome to MDN

PR #43361, opened 2026-03-07
doc: fix example of content-length to correct value

samuel871211:

### DIFF
--- a/files/en-us/web/http/reference/status/205/index.md
+++ b/files/en-us/web/http/reference/status/205/index.md
@@ -17,7 +17,6 @@ The instruction to "reset content" can mean clearing the contents of a form, res
 
 Note that the response must not include any content, and browsers may reject responses that do so.
 The response should not include the {{HTTPHeader("Content-Length")}} header either, but if it does the value must be `0`.
- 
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/reference/status/205/index.md
samuel871211: I find this sentence strange because the server can generate `content-length: 0`, as is the example in line 47 below.

[RFC 9110 15.3.6. 205 Reset Content](https://www.rfc-editor.org/rfc/rfc9110.html#name-205-reset-content)
```
Since the 205 status code implies that no additional content will be provided, a server MUST NOT generate content in a 205 response.
```

Perhaps @hamishwillee could take a look at how to rephrase this sentence, thank you.

samuel871211:
> I think you should replace `!` with `%21` and set the length to `16` since it's `application/x-www-form-urlencoded`.

Yes, you are right. Thanks !

files/en-us/web/http/reference/status/205/index.md
hamishwillee: The spec says "should not".  I'll patch this here. Thanks.

> A user agent SHOULD NOT send a Content-Length header field when the request message does not contain content and the method semantics do not anticipate such data.

Issue #43367, opened 2026-03-08
`Document.querySelector()` vs. `Element.querySelector()`


whxvd:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model

### What specific section or headline is this issue about?

Examples

### What information was incorrect, unhelpful, or incomplete?

The first two examples start with notes that `Document.querySelector()` is used. But both examples use `document.body.querySelector`, i.e., `Element.querySelector()` instead `Document.querySelector()`.

### What did you expect to see?

The first two examples probably could use `document.querySelector` instead of `document.body.querySelector`.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/api/document_object_model`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/document_object_model/index.md
* Last commit: https://github.com/mdn/content/commit/85fccefc8066bd49af4ddafc12c77f35265c7e2d
* Document last modified: 2025-11-07T15:58:06.000Z

</details>

jyotsna05-22:
Hello maintainers,

I would like to work on this issue. The examples currently use `document.body.querySelector()` instead of `document.querySelector()`, and I would like to update the examples to reflect the correct usage.

Could you please assign this issue to me?

Thank you!

Josh-Cena:
We don't assign. Just send a PR.

PR #43387, opened 2026-03-09
Fix: replace document.body.querySelector with document.querySelector

jyotsna05-22:
Replaced `document.body.querySelector` with `document.querySelector` in the DOM examples to follow the correct usage mentioned in the issue.

Fixes https://github.com/mdn/content/issues/43367


pepelsbey:
There are more `document.body.querySelector` examples in the article. It would make sense to update them all.

PR #43390, opened 2026-03-09
Update CookieStoreManager.subscribe() docs with optional name and url

### LINKED ISSUE #43382: Change arguments to CookieStoreManager's subscribe method
### Proposal

Update documentation to reflect changes in https://github.com/whatwg/cookiestore/pull/291#issuecomment-3723024282 

The PR changes the arguments of `subscribe(name, url)`. `name` is now optional and matches to every available cookie in the service worker scope if omitted. The `url` of the subscription should default to be the service worker's scope url if omitted. 

### Browser support

_No response_

### Tasks

- [ ]

### Dependencies

- [ ]

### Additional information

this information should be on https://developer.mozilla.org/en-US/docs/Web/API/CookieStoreManager/subscribe 

can write/review

mohammed-ali-osman:
### Description

- Updated the `subscribe()` method docs to reflect the new optional `name` and `url` parameters.
-  Clarified that omitting name subscribes to all cookies in the scope, and omitting url defaults to the service worker registration scope.
- Added examples demonstrating:
  - subscribing to all cookies.
  - subscribing to all cookies with a specific name,
  - handling subscriptions for narrower URL paths, 
  - `TypeError` when URL is outside the service worker scope.
 - Noted that duplicate subscriptions are ignored.
 
 fixes: #43382


chrisdavidmills:
I'm sorry, but I don't have time to review this over the next few days. I'll see if someone else can look at this.

mohammed-ali-osman:
Thank you @chrisdavidmills for letting me know. I appreciate you checking if someone else might be able to review it.

files/en-us/web/api/cookiestoremanager/subscribe/index.md
```diff
@@ -20,9 +20,9 @@ subscribe(subscriptions)
 
 - `subscriptions`
   - : An array of objects, each of which has the following properties:
-    - `name`
+    - `name` {{optional_inline}}
       - : A string with the name of a cookie.
```
wbamberg:
```suggestion
      - : A string with the name of a cookie. If `name` is omitted, the service worker is subscribed to change events for all cookies that are in scope.
```

files/en-us/web/api/cookiestoremanager/subscribe/index.md
```diff
@@ -20,9 +20,9 @@ subscribe(subscriptions)
...
-    - `url`
+    - `url` {{optional_inline}}
       - : A string with the url of a cookie scope. This may be narrower than the scope of the service worker registration.
```
wbamberg:
```suggestion
      - : A string with the url of a cookie scope. This may be narrower than the scope of the service worker registration. If `url` is omitted, it defaults to the scope of the service worker registration.
```

files/en-us/web/api/cookiestoremanager/subscribe/index.md
```diff
@@ -38,19 +38,41 @@ A {{jsxref("Promise")}} that resolves with {{jsxref("undefined")}} when the subs
 
 In this example, the {{domxref("ServiceWorkerRegistration")}} represented by `registration` is subscribing to change events on the cookie named `"cookie1"` with a scope of `"/path1"`.
 
+Duplicate subscriptions (same `name` and `url`) are ignored; only unique subscriptions are added.
```
wbamberg: I don't think this is terribly clear, and it doesn't really seem in scope for this change. We could say something like:

> Duplicate subscriptions are ignored: that is, if a service worker subscribes more than once to the same cookie, it will only receive change notifications once.

However, this doesn't belong in "Examples". Usually we'd put more information in a separate "Description" section, but this doesn't seem like enough to justify that. Maybe it could go under "Return value", as it's somewhat about the result of the call?

files/en-us/web/api/cookiestoremanager/subscribe/index.md
```diff
@@ -38,19 +38,41 @@ A {{jsxref("Promise")}} that resolves with {{jsxref("undefined")}} when the subs
...
+
 ```js
 const subscriptions = [{ name: "cookie1", url: `/path1` }];
 await registration.cookies.subscribe(subscriptions);
 ```
 
+If the `name` property is omitted in a subscription entry, the subscription applies to all cookies within the specified URL scope. Similarly, if the `url` property is omitted, the subscription applies to the entire scope of the service worker registration.
```
wbamberg: The "Examples" section as a whole is a bit disorganized. It would help to introduce some H3 subheadings for the different examples. You could have:

- Setting name and url (this could combine the first example and the example that starts "The URL passed to the subscribe() method, may be narrower", since they seem to be talking about the same case really).
- Setting name only
- Setting url only
- Service worker scope restriction

wbamberg (dismissed):
Thanks for your PR!

files/en-us/web/api/cookiestoremanager/subscribe/index.md
mohammed-ali-osman: That is fair, but i am thinking of adding that under section "Return value" section not as value but as behavior so it feels odd to add it casually but what if as "note block"?

mohammed-ali-osman:
Thanks for the suggestions, feel free to edit this further to align with MDN style guidelines if needed. I'm happy for you to adjust structure, wording, or formatting to better fit consistency with surrounding documentation.

files/en-us/web/api/cookiestoremanager/subscribe/index.md
wbamberg: > under section "Return value" section not as value but as behavior so it feels odd

Yeah, I do agree about that. In that case I'd have it at the start, as another paragraph after:

> The subscribe() method of the [CookieStoreManager](https://pr43390.review.mdn.allizom.net/en-US/docs/Web/API/CookieStoreManager) interface subscribes a [ServiceWorkerRegistration](https://pr43390.review.mdn.allizom.net/en-US/docs/Web/API/ServiceWorkerRegistration) to cookie change events.

I don't think a note is appropriate, notes are for "aside" type things that aren't part of the normal flow of the document. This is just part of its behavior.

mohammed-ali-osman: > The subscribe() method of the [CookieStoreManager](https://pr43390.review.mdn.allizom.net/en-US/docs/Web/API/CookieStoreManager) interface subscribes a [ServiceWorkerRegistration](https://pr43390.review.mdn.allizom.net/en-US/docs/Web/API/ServiceWorkerRegistration) to cookie change events.

That makes sense, thanks! I’ll move it into the main description right under this paragraph.

files/en-us/web/api/cookiestoremanager/subscribe/index.md
mohammed-ali-osman: I totally agree with that.

wbamberg:
> Thanks for the suggestions, feel free to edit this further to align with MDN style guidelines if needed. I'm happy for you to adjust structure, wording, or formatting to better fit consistency with surrounding documentation.

OK, I'll push some changes.

wbamberg:
@chrisdavidmills , since I ended up pushing a bunch of my own changes to this PR, perhaps you could review? Thank you!

chrisdavidmills:
> @chrisdavidmills , since I ended up pushing a bunch of my own changes to this PR, perhaps you could review? Thank you!

Sure thing, @wbamberg.

files/en-us/web/api/cookiestoremanager/subscribe/index.md
```diff
@@ -20,10 +22,10 @@ subscribe(subscriptions)
 
 - `subscriptions`
   - : An array of objects, each of which has the following properties:
-    - `name`
-      - : A string with the name of a cookie.
-    - `url`
-      - : A string with the url of a cookie scope. This may be narrower than the scope of the service worker registration.
+    - `name` {{optional_inline}}
+      - : A string with the name of a cookie. If `name` is omitted, the service worker is subscribed to change events for all cookies that are in scope.
```
chrisdavidmills:
```suggestion
      - : A string equal to the name of a cookie. If `name` is omitted, the service worker is subscribed to change events for all cookies that are in scope.
```

"with" sounded a bit odd to me.

files/en-us/web/api/cookiestoremanager/subscribe/index.md
```diff
@@ -20,10 +22,10 @@ subscribe(subscriptions)
...
+    - `url` {{optional_inline}}
+      - : A string with the url of a cookie scope. This may be narrower than the scope of the service worker registration. If `url` is omitted, it defaults to the scope of the service worker registration.
```
chrisdavidmills:
```suggestion
      - : A string equal to the URL of a cookie scope. This may be narrower than the scope of the service worker registration. If `url` is omitted, it defaults to the scope of the service worker registration.
```

files/en-us/web/api/cookiestoremanager/subscribe/index.md
chrisdavidmills: This isn't quite right, as it says above that the URL can be narrower than the SW's scope. Looking at the spec, this should say that a `TypeError` is thrown if `url` is not a valid URL, or does not start with the SWs reg's scope.

files/en-us/web/api/cookiestoremanager/subscribe/index.md
```diff
@@ -36,19 +38,45 @@ A {{jsxref("Promise")}} that resolves with {{jsxref("undefined")}} when the subs
...
+### Setting name and URL
+
 In this example, the {{domxref("ServiceWorkerRegistration")}} represented by `registration` is subscribing to change events on the cookie named `"cookie1"` with a scope of `"/path1"`.
 
 ```js
+// Subscribe to a specific cookie and URL
 const subscriptions = [{ name: "cookie1", url: `/path1` }];
 await registration.cookies.subscribe(subscriptions);
+
+// Subscribe to all cookie changes within the entire registration scope
```
chrisdavidmills: This is a separate example, so should have a separate H3?

files/en-us/web/api/cookiestoremanager/subscribe/index.md
```diff
@@ -36,19 +38,45 @@ A {{jsxref("Promise")}} that resolves with {{jsxref("undefined")}} when the subs
...
+await registration.cookies.subscribe([{}]);
 ```
 
-The URL passed to the `subscribe()` method, may be narrower than the service worker registration scope. In the following example the subscription is for `/path/one/`, so it will receive change events for changes on the first cookie, but not the second.
+### Setting name only
+
+In this example we set only `name`, and omit `url`: the subscription applies to all cookies names `cookie1` that are under the service worker's scope.
```
chrisdavidmills:
```suggestion
In this example, we set only `name` and omit `url`: the subscription applies to all cookies named `cookie1` within the service worker's scope.
```

chrisdavidmills (changes requested):
@wbamberg, @mohammed-ali-osman, a few comments for you, but this is mostly looking good.

files/en-us/web/api/cookiestoremanager/subscribe/index.md
mohammed-ali-osman: I see why a separate section was suggested, since `subscribe([{}])` behaves differently from the other examples.

Maybe we could add a new section titled “Subscribing to all cookies” under "Setting URL only" section, and then explain that this is done by passing an empty object (`{}`).

Let me know if that is ok.

files/en-us/web/api/cookiestoremanager/subscribe/index.md
mohammed-ali-osman: Good catch, that makes sense. “Does not match the scope” is misleading since the URL can be a subpath of the service worker’s scope.

I’ll update this to reflect the spec more accurately, something like:

- `TypeError`
    - : Thrown if the `url` is not a valid URL, or if it does not start with the service worker registration’s `scope`.

Let me know if that wording looks right to you.

chrisdavidmills: Looks good to me; you could shorten the description slightly to:

> Thrown if the `url` is not a valid URL, or doesn't start with the service worker registration’s scope.

files/en-us/web/api/cookiestoremanager/subscribe/index.md
chrisdavidmills: I think the suggested heading "Subscribing to all cookies" is fine, but I would put it in its own section, not under "Setting URL only" the logic being that you are not setting anything, you are just using the default scope.

Maybe put it just above the "Setting a URL outside the service worker's scope" example?

mohammed-ali-osman: > Maybe we could add a new section titled “Subscribing to all cookies” under "Setting URL only" section, and then explain that this is done by passing an empty object (`{}`).

The fun part is that the section above the "Setting a URL outside the service worker's scope" is under "Setting URL only" section that i was talking about.

My fault i didn't mention it's separate or independent section.

mohammed-ali-osman:
@chrisdavidmills I've applied the suggested changes. Please let me know if further tweaks are needed.

chrisdavidmills (approved):
@mohammed-ali-osman awesome, this looks ready to merge. Thank you, and thanks to @wbamberg!

PR #43396, opened 2026-03-10
Correct Event: isTrusted exception for elem.click()

mustaqahmed:
### Description

The MDN page for `Event.isTrusted` property suggests that the click event should always have the property set to false. This seems incorrect.

### Motivation

All of Blink, Gecko and WebKit fire clicks with regular `.isTrusted` values (trusted if and only if the event came from UA).

### Additional details

Test page: https://codepen.io/editor/mustaqahmed/pen/019cd7f5-bed2-7130-9ef7-74821cee52d3

See https://github.com/w3c/pointerevents/issues/642.


mustaqahmed:
FYI @smaug----.

PR #43407, opened 2026-03-11
Clarify offsetParent behavior for fixed-position elements

### LINKED ISSUE #43394: offsetParent is not always null for position: fixed elements
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent

The Note box

The Note box states that:

> offsetParent returns null in the following situations
> ... 
> The element has the position property set to fixed (Firefox returns `<body>`).

This is stated as if it's always correct; however, that's not the case. 

In both Chrome and Firefox, offsetParent is null for fixed-position elements only if the containing block for this fixed-position element is the viewport. Otherwise, it's not null; it's the nearest ancestor that has styles like `transform` set.

The sentence in the note section should clarify this.

There's a WPT test to verify this: https://wpt.fyi/results/css/cssom-view/offsetParent-fixed.html?label=experimental&label=master&aligned

_No response_

mohammed-ali-osman:
### Description

Updated Note box in `HTMLElement.offsetParent` to clarify the behavior for elements with `position: fixed`.

- [x] `display: none` → null (No layout box)
- [x] `position: fixed` → null (Unless inside a containing block)
- [x] `Root elements` → null (<html> and <body>)

### Comparison

before;

> [!NOTE]
> `offsetParent` returns `null` in the following
> situations:
>
> - The element or any ancestor has the `display` property set to
>   `none`.
> - The element has the `position` property set to `fixed`
>   (Firefox returns `<body>`).
> - The element is `<body>` or `<html>`.

after:

> [!NOTE]
> `offsetParent` returns `null` in the following situations:
>
> - The element or any ancestor has the `display` property set to
>   `none`.
> - The element has the `position` property set to `fixed`. In this case,
>   `offsetParent` is `null` if the containing block is the viewport; otherwise
>   it returns the nearest ancestor that forms the containing block
>   (e.g., an ancestor with `transform`, `perspective`, or `filter`).
> - The element is `<body>` or `<html>`.

Fixes #43394

### DIFF
--- a/files/en-us/web/api/htmlelement/offsetparent/index.md
+++ b/files/en-us/web/api/htmlelement/offsetparent/index.md
@@ -25,8 +25,8 @@ If there is no positioned ancestor element, the `body` is returned.
 >
 > - The element or any ancestor has the `display` property set to
 >   `none`.
-> - The element has the `position` property set to `fixed` and it's the containing block is the viewport.
->   If the containing block is not the viewport, `offsetParent` it returns the nearest ancestor that forms
+> - The element has the `position` property set to `fixed` and its containing block is the viewport.
+>   If the containing block is not the viewport, `offsetParent` returns the nearest ancestor that forms
 >   a containing block, for example, an ancestor with `transform`, `perspective`, or `filter` styles set.
 > - The element is `<body>` or `<html>`.
 


files/en-us/web/api/htmlelement/offsetparent/index.md
```diff
@@ -21,13 +21,14 @@ A positioned ancestor might be:
...
 > [!NOTE]
-> `offsetParent` returns `null` in the following
-> situations:
+> `offsetParent` returns `null` in the following situations:
 >
 > - The element or any ancestor has the `display` property set to
 >   `none`.
-> - The element has the `position` property set to `fixed`
->   (Firefox returns `<body>`).
+> - The element has the `position` property set to `fixed`. In this case,
```
chrisdavidmills: I've got a couple of issues with this phrasing:

1. It's repetitive. The lead-in line already says "`offsetParent` returns `null` in the following situations".
2. Someone flicking through the information quickly might just read the first sentence, "The element has the `position` property set to `fixed`.", and end up being misled.

I would suggest rewriting the bullet to something like:

> - The element has the `position` property set to `fixed`, but only if the containing block is the viewport. Otherwise, it returns the nearest ancestor that forms a containing block, for example, an ancestor with `transform`, `perspective`, or `filter` styles set.

mohammed-ali-osman:
Thanks @chrisdavidmills for suggestion and the guidance.

Updated the second bullet list to match your request, 

- removed repeated ``offsetParent`` return ``null``, it's already mentioned in the Note title.
- Simplified the first sentence to avoid misleading readers who skim quickly, and made the condition bold for easier scanning.

### Preview

> [!NOTE]
> `offsetParent` returns `null` in the following situations:
>
> - The element or any ancestor has the `display` property set to
>   `none`.
> - The element has the `position` property set to `fixed`, **if the containing block is the viewport**.
>   Otherwise, it returns the nearest ancestor that forms a containing block, for example, an ancestor with
>   `transform`, `perspective`, or `filter` styles set.
> - The element is `<body>` or `<html>`.

files/en-us/web/api/htmlelement/offsetparent/index.md
```diff
@@ -21,13 +21,13 @@ A positioned ancestor might be:
...
 > [!NOTE]
-> `offsetParent` returns `null` in the following
-> situations:
+> `offsetParent` returns `null` in the following situations:
 >
 > - The element or any ancestor has the `display` property set to
 >   `none`.
-> - The element has the `position` property set to `fixed`
->   (Firefox returns `<body>`).
+> - The element has the `position` property set to `fixed`, if the containing block is the viewport.
```
chrisdavidmills:
```suggestion
> - The element has the `position` property set to `fixed` and its containing block is the viewport.
```

files/en-us/web/api/htmlelement/offsetparent/index.md
```diff
@@ -21,13 +21,13 @@ A positioned ancestor might be:
...
+>   Otherwise, it returns the nearest ancestor that forms a containing block, for example, an ancestor with
```
chrisdavidmills:
```suggestion
>   If the containing block is not the viewport, `offsetParent` returns the nearest ancestor that forms a containing block, for example, an ancestor with
```

files/en-us/web/api/htmlelement/offsetparent/index.md
mohammed-ali-osman: That is true, reading it now feels more good than using `comma`.

I guess we should use a conjunction.

mohammed-ali-osman:
@chrisdavidmills Nice addition `If the containing block is not the viewport, offsetParent`  instead of `Otherwise`. it adds more information and it feel more understandable.

mohammed-ali-osman:
@chrisdavidmills You're welcome! I didn’t do much, the credit goes to the suggestions; I just followed them.

PR #43413, opened 2026-03-12
Clarify that (lang="") means explicitly unknown, not default.

### LINKED ISSUE #43259: MDN "lang" page says incorrectly says "the default value of lang is the empty string", which is wrong -- explicit-empty-string vs. missing-attribute have different semantics
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang

"Note: The default value of lang is the empty string, which means that the language is unknown."

"Note: The default value of lang is the empty string,"

Either this statement being removed, or it expanded to be more accurate.

https://html.spec.whatwg.org/multipage/dom.html#language

Notes on why the statement is incorrect:

https://html.spec.whatwg.org/multipage/dom.html#language has the details, and it is conditioned on **whether the attribute is set**.  (which means the attribute being set to the empty string vs. being unset are semantically different)

If the `lang` attribute is explicitly set to the empty string, then the spec says that the language is **explicitly unknown** per [1].

But when when the `lang` attribute is not set on an element, then the spec defines a variety of other sources that browsers should use to establish the language (including the element's parent node, or shadow root, or "a [pragma-set default language](https://html.spec.whatwg.org/multipage/semantics.html#pragma-set-default-language)", or language information from another involved protocol like from an HTTP header.

[1] Here's the spec text that says an explicit empty-string value results in the language being explicitly-unknown:
> To determine the language of a node, user agents must use the first appropriate step in the following list:
> [...]
> * If the node is an [HTML element](https://html.spec.whatwg.org/multipage/infrastructure.html#html-elements) or an element in the [SVG namespace](https://infra.spec.whatwg.org/#svg-namespace), and it has a [lang](https://html.spec.whatwg.org/multipage/dom.html#attr-lang) in no namespace attribute set
>     Use the value of that attribute.
> [...]
>  If the resulting value is the empty string, then it must be interpreted as meaning that the language of the node is explicitly unknown.

It looks like this note was last modified in #32604 /  https://github.com/mdn/content/commit/1a8480354c678666ab33ec87259c3363f971db05 which replaced another bit of incorrect information ("The default value of `lang` is `unknown`"), but the replacement about the empty string is also incorrect. 

### MDN metadata

<details>

* Folder: `en-us/web/html/reference/global_attributes/lang`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/reference/global_attributes/lang/index.md
* Last commit: https://github.com/mdn/content/commit/0754cd805a8e010d2e3a2a065f634a3bcf358252
* Document last modified: 2025-07-09T21:49:04.000Z

</details>

mohammed-ali-osman:
### Description

The previous note on the `lang` attribute said the default value is the empty string.

That was incorrect, the **empty string** only applies when the attribute is explicitly set, which signals the language is explicitly `unknown`.

Spec reference:
```text
 If the resulting value is the empty string, then it must be interpreted as meaning that the language of the node is explicitly unknown.
```

 This PR updates the note to be spec-accurate and complements the [inheritance section](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang#inheritance).

### Preview

> [!NOTE]
> If the value of `lang` is set to empty string the language is explicitly unknown. Therefore, it is recommended to always specify an appropriate value for this attribute.

fixes #43259

### DIFF
--- a/files/en-us/web/html/reference/global_attributes/lang/index.md
+++ b/files/en-us/web/html/reference/global_attributes/lang/index.md
@@ -10,7 +10,7 @@ sidebar: htmlsidebar
 The **`lang`** [global attribute](/en-US/docs/Web/HTML/Reference/Global_attributes) helps define the language of an element: the language that non-editable elements are written in, or the language that the editable elements should be written in by the user. The attribute contains a single {{glossary("BCP 47 language tag")}}.
 
 > [!NOTE]
-> If the value of `lang` is set to empty string the language is explicitly unknown. Therefore, it is recommended to always specify an appropriate value for this attribute.
+> If the value of `lang` is set to an empty string, the language is explicitly unknown. Therefore, it is recommended to always specify an appropriate value for this attribute.
 
 {{InteractiveExample("HTML Demo: lang", "tabbed-shorter")}}
 


files/en-us/web/html/reference/global_attributes/lang/index.md
```diff
@@ -10,7 +10,7 @@ sidebar: htmlsidebar
 The **`lang`** [global attribute](/en-US/docs/Web/HTML/Reference/Global_attributes) helps define the language of an element: the language that non-editable elements are written in, or the language that the editable elements should be written in by the user. The attribute contains a single {{glossary("BCP 47 language tag")}}.
 
 > [!NOTE]
-> The default value of `lang` is the empty string, which means that the language is unknown. Therefore, it is recommended to always specify an appropriate value for this attribute.
+> If the value of `lang` is set to empty string the language is explicitly unknown. Therefore, it is recommended to always specify an appropriate value for this attribute.
```
pepelsbey:
```suggestion
> If the value of `lang` is set to an empty string, the language is explicitly unknown. Therefore, it is recommended to always specify an appropriate value for this attribute.
```

mohammed-ali-osman:
Thanks for suggestions and the review @pepelsbey!

PR #40627, opened 2025-08-04
fix issue: Backdrop Root concept undocumented

kb42:
### Description

Clarify that `backdrop-filter` only affects the backdrop image up to its nearest ancestor Backdrop Root as stated in the issue, and add a concise note listing what triggers a Backdrop Root which was obtained from the source listed in this PR.

### Motivation

The existing text (“area behind an element”) implied that backdrop-filter applies to all pixels behind an element, which contradicts real-world behavior and the Filter Effects Level 2 specification. Developers encountered “missing” blur outside the nearest Backdrop Root (for example, a semi-transparent ancestor), leading to confusion. This update aligns MDN’s description with the spec and addresses user reports in the GitHub issue.

### Additional details

- **Filter Effects Level 2 (Backdrop Root definition)**  
  https://drafts.fxtf.org/filter-effects-2/#BackdropRoot  
- **Live demo showing blur limited to the nearest Backdrop Root**  
  https://jsfiddle.net/xer31ywj/  

### Related issues and pull requests

Fixes #40602

PR #43421, opened 2026-03-12
Document backdrop root concept on backdrop-filter page

### LINKED ISSUE #40602: Backdrop Image and Backdrop Root are not mentioned and it causes confusion.
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter

First description of the backdrop-filter

> "to the area behind an element"

This is statement in the first paragraph is not correct

According to the specification the filter is applied to the area behind the element up until the nearest ancestor Backdrop Root element.

So that in this example:

```
<div style="background-color:red"> <!-- grand-parent -->
  <div style="color: yellow; will-change:opacity"> <!-- nearest ancestor Backdrop Root -->
    Yellow Text
    <div style="backdrop-filter: blur(2px); position:absolute;top:0;left:0;"></div> <!-- element with backdrop filter
  </div>
</div>
```

In this case only yellow text pixels will be blurrred and NOT red pixels of the grand-parent.

I think we should at least mention it, because it was not clear for me when I've encountered "bug" with not all behind pixels being blurred.

What we can add:

```
NOTE: Filter do not applies to the pixels which are behind the element with "backdrop-filter" property in case if they are produced by elements outside
of the nearest backdrop root ancestor of the element with "backdrop-filter".

The Backdrop Root element:
    - The root element of the document (HTML).
    - An element with a filter property other than "none".
    - An element with an opacity value less than 1.
    - An element with mask, mask-image, mask-border, or clip-path properties with values other than “none”.
    - An element with a backdrop-filter value other than "none".
    - An element with a mix-blend-mode value other than "normal".
    - An element with a will-change value specifying any property that would create a Backdrop Root on non-initial value.


```

And maybe some example:
https://jsfiddle.net/xer31ywj/

Specification of the backdrop filter:
https://drafts.fxtf.org/filter-effects-2/#BackdropRoot

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/css/backdrop-filter`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/backdrop-filter/index.md
* Last commit: https://github.com/mdn/content/commit/0cc9980e3b21c83d1800a428bc402ae1865326b2
* Document last modified: 2025-07-14T14:43:58.000Z

</details>

kovan:
## Summary
- Add a Description section to the `backdrop-filter` CSS property page explaining that effects are bounded by the nearest ancestor backdrop root
- List the properties that create a backdrop root: `filter`, `opacity`, `mask`, `mask-image`, `mask-border`, `clip-path`, `backdrop-filter`, `mix-blend-mode`, and `will-change`
- Explain the common confusion when `backdrop-filter` appears to have no effect due to an ancestor backdrop root

## Motivation
Developers are frequently confused when `backdrop-filter` doesn't affect all pixels behind an element. The concept of a "backdrop root" (from Filter Effects Level 2) is not documented on MDN. A previous PR (#40627) attempted this but was closed due to merge conflicts; this is a fresh implementation incorporating reviewer feedback (using a dedicated Description section instead of a note).

Fixes #40602.

## Test plan
- [ ] Page renders correctly with new Description section
- [ ] All CSS property links resolve

### DIFF
--- a/files/en-us/web/css/reference/properties/backdrop-filter/index.md
+++ b/files/en-us/web/css/reference/properties/backdrop-filter/index.md
@@ -124,51 +124,63 @@ The following example demonstrates how backdrop roots affect `backdrop-filter`. 
 ```
 
 ```css
+body {
+  display: flex;
+  column-gap: 16px;
+  padding: 16px;
+  background-image: conic-gradient(
+    gray 90deg,
+    silver 90deg 180deg,
+    gray 180deg 270deg,
+    silver 270deg
+  );
+  background-size: 32px 32px;
+}
+
 .parent {
+  position: relative;
   width: 256px;
   height: 256px;
-  position: relative;
-  display: inline-block;
 }
 
 .backdrop-root {
-  will-change: opacity;
   outline: 2px solid crimson;
+  will-change: opacity;
 }
 
 .square {
+  position: absolute;
+  top: 35px;
+  left: 40%;
   width: 25%;
   height: 25%;
-  position: absolute;
   border: 10px solid white;
-  left: 40%;
-  top: 35px;
 }
 
 .text {
+  position: absolute;
+  left: 40%;
+  color: white;
   font-size: 32px;
   font-weight: 500;
   text-align: center;
   line-height: 256px;
-  position: absolute;
-  left: 40%;
   filter: blur(1px);
-  color: white;
 }
 
 .overlay {
-  border-radius: 9999px;
-  outline: 1px solid #eaeaea;
-  width: 50%;
-  height: 50%;
   position: absolute;
   top: 25%;
   left: 50%;
+  width: 50%;
+  height: 50%;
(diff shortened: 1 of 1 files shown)


pepelsbey (changes requested):
Thank you for the help! A few suggestions:

- The description section usually goes under the Values section, see the [background property](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/background#description).
- There was [a good demo](https://jsfiddle.net/xer31ywj/) mentioned in the issue, it might be worth including it as well.

pepelsbey:
@whiteand what do you say? Does it fix the issue you’ve raised? Thank you!

pepelsbey (approved):
Looks good! Thank you 🙂

I updated the demo to have a checkered background, like the original one.

PR #43424, opened 2026-03-13
Add a guide on fetch metadata

wbamberg:
This adds a standalone guide on fetch metadata headers. I thought it fit better in HTTP than in Security.

### DIFF
--- a/files/en-us/web/http/guides/fetch_metadata/index.md
+++ b/files/en-us/web/http/guides/fetch_metadata/index.md
@@ -7,20 +7,13 @@ sidebar: http
 
 **Fetch metadata** is the term for a group of HTTP request headers that give the server information about the context in which the request is being made.
 
-For example, this enables the server to know:
+Among other things, fetch metadata allows the server to know:
 
 - Whether the request represents a navigation between documents, or a request for a subresource, or was explicitly made from JavaScript, for example using the {{domxref("window.fetch()", "fetch()")}} API.
 
 - The relationship between the requester of the resource and the resource being requested: whether they are same-{{glossary("origin")}}, or same-{{glossary("site")}}, or from completely different sites.
 
-By using the information in these headers to allow or deny specific requests, a server can implement a defense against _cross-origin attacks_: that is, attacks in which a request is made from the attacker's origin to a resource belonging to the target origin. This includes [cross-site request forgeries (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF) and various [cross-site leaks](/en-US/docs/Web/Security/Attacks/XS-Leaks).
-
-> [!NOTE]
-> We use the term _cross-origin_ attack in this guide, although many attacks are conventionally called _cross-site_ attacks.
->
-> An {{glossary("origin")}} is a more restrictive concept than a {{glossary("site")}}. In particular, a site includes a domain's subdomains, and an origin does not: so `https://example.org` and `https://login.example.org` are the same site, but different origins.
->
-> This means that while all cross-site attacks are cross-origin attacks, some cross-origin attacks are _not_ cross-site attacks. For example, if an attacker gains control of a subdomain of a site, then they can attack the site using _cross-origin_, _same-site_ requests. To include these attacks, we use the more restrictive term.
+By using the information in these headers to allow or deny specific requests, a server can implement a defense against [_cross-origin attacks_](#cross-origin_attacks) such as [cross-site request forgeries (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF) and various [cross-site leaks](/en-US/docs/Web/Security/Attacks/XS-Leaks).
 
 ## Fetch metadata headers
 
@@ -41,7 +34,7 @@ We could think of it, roughly, as the way the returned resource would be used.
 
 For most {{glossary("replaced elements")}}, the value of the header names the element that this resource will be used for, such as `iframe`, `object`, `audio`, or `video`. A value of `image` indicates that the resource will be used as an image referenced by a replaced element such as an HTML {{htmlelement("img")}} element, a CSS {{cssxref("background-image")}} property, an SVG {{svgelement("image")}}, or any other place in the web platform that use images loaded from subresources.
 
-For the complete set of possible values, see the {{HTTPHeader("Sec-Fetch-Site", "reference page", "", "nocode")}} for this header, but other interesting values include:
+Some other interesting destination values include:
 
 - `document`
   - : The request is for a new document that is the target of a top-level navigation (for example, the user clicking a link in the page or submitting a form).
@@ -54,9 +47,11 @@ For the complete set of possible values, see the {{HTTPHeader("Sec-Fetch-Site", 
 - `empty`
   - : The request does not have a defined destination: among other possible causes, this is the value given if the request is the result of a {{domxref("Window.fetch()", "fetch()")}} call.
 
+For the complete set of possible values, see the {{HTTPHeader("Sec-Fetch-Site", "reference page", "", "nocode")}} for this header.
+
 ### Sec-Fetch-Mode
 
-This header indicates the _mode_ of the request. Like _destination_, the mode is defined in the Fetch API, where it is exposed as the {{domxref("Request.mode")}} property.
+This header indicates the _mode_ of the request. Like _destination_, the mode is defined in the [Fetch API](/en-US/docs/Web/API/Fetch_API), where it is exposed as the {{domxref("Request.mode")}} property.
 
 Its most commonly used values are:
 
@@ -64,7 +59,9 @@ Its most commonly used values are:
   - : The request represents a navigation between documents (for example, the user clicking a link).
 
 - `no-cors`
-  - : The request was made in `no-cors` mode, which means that it is allowed cross-origin without the server sending the appropriate [CORS](/en-US/docs/Web/HTTP/Guides/CORS) headers, but the response is _opaque_, meaning that its headers and bodies can't be accessed by JavaScript running in the client.
+  - : The request was made in `no-cors` mode.
+
+    This means that it is allowed cross-origin without the server sending the appropriate [CORS](/en-US/docs/Web/HTTP/Guides/CORS) headers, with the restriction that the response can't be accessed by JavaScript running in the client (it is _opaque_).
 
     This is the default mode for pages that load subresources such as images, fonts, scripts, and stylesheets, and explains why a different site is by default allowed to use your site's subresources, even if you haven't configured CORS to allow it.
 
@@ -110,6 +107,13 @@ This header is included only if the request was initiated by a user action (such
 
 Fetch metadata is especially useful as a defense against _cross-origin attacks_. These attacks typically target a user who has an account with a legitimate site, and is signed into this site. The attacker creates a website that makes a _cross-origin request_ to the legitimate site, and then tricks the user into executing that request.
 
+> [!NOTE]
+> We use the term _cross-origin_ attack in this guide, although many attacks are conventionally called _cross-site_ attacks.
+>
+> An {{glossary("origin")}} is a more restrictive concept than a {{glossary("site")}}. In particular, a site includes a domain's subdomains, and an origin does not: so `https://example.org` and `https://login.example.org` are the same site, but different origins.
+>
+> This means that while all cross-site attacks are cross-origin attacks, some cross-origin attacks are _not_ cross-site attacks. For example, if an attacker gains control of a subdomain of a site, then they can attack the site using _cross-origin_, _same-site_ requests. To include these attacks, we use the more restrictive term.
+
 For example, the attacker's site might contain a {{htmlelement("form")}} element that submits to the legitimate site. For some cross-origin attacks, no user interaction is needed at all: the attacker's page can just execute a {{domxref("Window.fetch()", "fetch()")}} request to the legitimate site on page load, and then the user only has to open the attacker's page for the cross-origin request to be executed.
 
 Because the request came from the user's browser, it will include any cookies set for the user by the legitimate site, including cookies that the legitimate site uses to identify users. The request will therefore be given the privileges for that user.
@@ -136,7 +140,8 @@ For example, the following [Express](/en-US/docs/Learn_web_development/Extension
 
 ```js
 function isAllowed(req) {
-  // Allow same-origin and directly user-initiated requests
+  // Allow same-origin requests
+  // Allow directly user-initiated requests (from bookmarks, address bar etc.)
   const secFetchSite = req.headers["sec-fetch-site"];
   if (secFetchSite === "same-origin" || secFetchSite === "none") {
     return true;


files/en-us/web/http/reference/headers/sec-fetch-dest/index.md
```diff
@@ -7,7 +7,7 @@ browser-compat: http.headers.Sec-Fetch-Dest
 sidebar: http
 ---
 
-The HTTP **`Sec-Fetch-Dest`** {{Glossary("fetch metadata request header")}} indicates the request's _destination_.
+The HTTP **`Sec-Fetch-Dest`** [fetch metadata request header](/en-US/docs/Web/HTTP/Guides/Fetch_metadata) indicates the request's _destination_.
```
hamishwillee, line 10: 1. The glossary needs to be updated to point to this new doc.
2. Why link to fetch metadata rather than the glossary? Arguments for linking to the glossary  might be that it actually matches the link title, and also that you might reasonably only be interested in a succinct link in this case - providing information about the headers themselves.

   Ditto for all the other "exactly the same" cases.

   Obviously I could argue the other view, so this is really an "are you sure" question.

files/en-us/web/http/guides/fetch_metadata/index.md
```diff
@@ -0,0 +1,175 @@
+---
+title: Fetch metadata
+slug: Web/HTTP/Guides/Fetch_metadata
+page-type: guide
+sidebar: http
+---
+
+**Fetch metadata** is the term for a group of HTTP request headers that give the server information about the context in which the request is being made.
+
+For example, this enables the server to know:
```
hamishwillee: I'd be tempted to remove "For example" and expand out with the a line on cors vs no cors and the sec-fetch user. 

Assuming you keep it, I'm also having trouble working out whether "this enables" is correct.  Probably is, but this is easier to read.

```suggestion
Among other things, fetch metadata allows the server to know:
```

files/en-us/web/http/guides/fetch_metadata/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+- Whether the request represents a navigation between documents, or a request for a subresource, or was explicitly made from JavaScript, for example using the {{domxref("window.fetch()", "fetch()")}} API.
+
+- The relationship between the requester of the resource and the resource being requested: whether they are same-{{glossary("origin")}}, or same-{{glossary("site")}}, or from completely different sites.
+
+By using the information in these headers to allow or deny specific requests, a server can implement a defense against _cross-origin attacks_: that is, attacks in which a request is made from the attacker's origin to a resource belonging to the target origin. This includes [cross-site request forgeries (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF) and various [cross-site leaks](/en-US/docs/Web/Security/Attacks/XS-Leaks).
```
hamishwillee: Consider that if this is in HTTP you probably shouldn't go too far into attacks at this level. Perhaps link down rather than explain the specific issues.

```suggestion
By using the information in these headers to allow or deny specific requests, a server can implement a defense against [cross-origin attacks](LINK), [cross-site request forgeries (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF) and various [cross-site leaks](/en-US/docs/Web/Security/Attacks/XS-Leaks).
```

files/en-us/web/http/guides/fetch_metadata/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+> [!NOTE]
+> We use the term _cross-origin_ attack in this guide, although many attacks are conventionally called _cross-site_ attacks.
+>
+> An {{glossary("origin")}} is a more restrictive concept than a {{glossary("site")}}. In particular, a site includes a domain's subdomains, and an origin does not: so `https://example.org` and `https://login.example.org` are the same site, but different origins.
+>
+> This means that while all cross-site attacks are cross-origin attacks, some cross-origin attacks are _not_ cross-site attacks. For example, if an attacker gains control of a subdomain of a site, then they can attack the site using _cross-origin_, _same-site_ requests. To include these attacks, we use the more restrictive term.
```
hamishwillee: Ditto for this - could move down into the section on cross origin attacks, since the focus is not XSS in HTTP, it is "providing context"

files/en-us/web/http/guides/fetch_metadata/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Sec-Fetch-Dest
+
+This header indicates the _destination_ of the request. This attribute is defined in the Fetch API, where it is exposed as the {{domxref("Request.destination")}} property.
+
+We could think of it, roughly, as the way the returned resource would be used.
+
+For most {{glossary("replaced elements")}}, the value of the header names the element that this resource will be used for, such as `iframe`, `object`, `audio`, or `video`. A value of `image` indicates that the resource will be used as an image referenced by a replaced element such as an HTML {{htmlelement("img")}} element, a CSS {{cssxref("background-image")}} property, an SVG {{svgelement("image")}}, or any other place in the web platform that use images loaded from subresources.
+
+For the complete set of possible values, see the {{HTTPHeader("Sec-Fetch-Site", "reference page", "", "nocode")}} for this header, but other interesting values include:
```
hamishwillee: Consider splitting this. It is a matter of taste - I figure if you wanted to highlight these few, then nice not to distract with the link to the complete reference.

So before:

> Some other interesting destination values include:


And after the list:

> For the complete set of possible values, see the {{HTTPHeader("Sec-Fetch-Site", "reference page", "", "nocode")}} for this header..

files/en-us/web/http/guides/fetch_metadata/index.md
```diff
@@ -0,0 +1,175 @@
...
+  - : The resource will be used as a script loaded from an HTML {{htmlelement("script")}} element or a call to {{domxref("WorkerGlobalScope.importScripts()", "importScripts()")}} in a web worker.
+
+    More specific values are used to indicate other places where the resource is used as a script, such as worklets (`audioworklet` and `paintworklet`) and workers (`sharedworker`, `serviceworker`, and `worker`).
+
+- `empty`
+  - : The request does not have a defined destination: among other possible causes, this is the value given if the request is the result of a {{domxref("Window.fetch()", "fetch()")}} call.
+
+### Sec-Fetch-Mode
+
+This header indicates the _mode_ of the request. Like _destination_, the mode is defined in the Fetch API, where it is exposed as the {{domxref("Request.mode")}} property.
```
hamishwillee: We link to the fetch() method above. Never sure if it is better to link to https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API - you could choose to do that link here.

files/en-us/web/http/guides/fetch_metadata/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+Its most commonly used values are:
+
+- `navigate`
+  - : The request represents a navigation between documents (for example, the user clicking a link).
+
+- `no-cors`
+  - : The request was made in `no-cors` mode, which means that it is allowed cross-origin without the server sending the appropriate [CORS](/en-US/docs/Web/HTTP/Guides/CORS) headers, but the response is _opaque_, meaning that its headers and bodies can't be accessed by JavaScript running in the client.
```
hamishwillee: Maybe choppier sentences.


```suggestion
  - : The request was made in `no-cors` mode, which means that it is allowed cross-origin without the server sending the appropriate [CORS](/en-US/docs/Web/HTTP/Guides/CORS) headers. In this case the response is _opaque_, meaning that its headers and bodies can't be accessed by JavaScript running in the client.
```

files/en-us/web/http/guides/fetch_metadata/index.md
```diff
@@ -0,0 +1,175 @@
...
+- Top-level navigational requests from another origin, so users can get to your site by clicking links in other sites.
+- Requests to specific endpoints that are meant to be accessed cross-origin, including any that use [CORS](/en-US/docs/Web/HTTP/Guides/CORS).
+
+For example, the following [Express](/en-US/docs/Learn_web_development/Extensions/Server-side/Express_Nodejs) code allows only same-origin requests, directly user-initiated requests, and navigations.
+
+```js
+function isAllowed(req) {
+  // Allow same-origin and directly user-initiated requests
+  const secFetchSite = req.headers["sec-fetch-site"];
+  if (secFetchSite === "same-origin" || secFetchSite === "none") {
```
hamishwillee, line 146: None? Quite possible I'm missing something, but the comment indicates "user initiated request", so you might be missing `Sec-Fetch-User` request here.

files/en-us/web/http/reference/headers/sec-fetch-dest/index.md
wbamberg: > Why link to fetch metadata rather than the glossary? 

No, I'm not sure. One thing is that I don't really like the way the HTTP docs use glossary entries here. I think the canonical case of glossary entries are things that span technologies, or at least have a lot of application in other tech areas, even if they are only defined in one. But "what a fetch metadata header is", or ["what a forbidden request header"](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_request_header) is specifically HTTP documentation and belongs there.

But that's a bigger issue.

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: > I'd be tempted to remove "For example" and expand out with the a line on cors vs no cors and the sec-fetch user.

Do you mean that by doing that, the list is exhaustive and therefore I don't need "for example"?

I'm not sure that's really true though, especially for `Sec-Fetch-Mode`, which doesn't only cover cors versus no-cors. And I want this part, not to list all the things that the headers expose, but to give people a motivation for the feature. That is, to express how you could use the feature to do things that are fairly obviously likely to be useful, which I'm not sure is true of something like "determine whether the request was made in cors or no-cors mode". I don't think most people even know what no-cors really is (see e.g. Mike's excellent post at https://stackoverflow.com/questions/43262121/trying-to-use-fetch-and-pass-in-mode-no-cors).

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: Fair, I started this as a guide in security and then moved it into HTTP.

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: I also love shorter sentences but I worry that this weakens the link between "you don't need to use CORS" and "you don't get access to the response". People (as seen in the SO thread linked above) sometimes think that no-cors is a fix for their problem but it almost never is.

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: No, `none` is a value for `Sec-Fetch-Site` which indicates that the request was "directly user-initiated". See https://pr43424.review.mdn.allizom.net/en-US/docs/Web/HTTP/Guides/Fetch_metadata#sec-fetch-site and https://w3c.github.io/webappsec-fetch-metadata/#directly-user-initiated. We don't need `Sec-Fetch-User` here.

files/en-us/web/http/reference/headers/sec-fetch-dest/index.md
hamishwillee: I hadn't thought about that, but I agree specifically that these are so HTTP specific that they should be HTTP docs. 

So _perhaps_ update the links there to the header section https://pr43424.review.mdn.allizom.net/en-US/docs/Web/HTTP/Guides/Fetch_metadata#fetch_metadata_headers

I'd be tempted in this PR to fix up all the links to this new doc and remove the glossary. Add an issue to fix the others.

files/en-us/web/http/guides/fetch_metadata/index.md
hamishwillee: Yes, the suggestion was at least to make it more exhaustive. I can see why you wouldn't do this. 

Mostly I spent far too long wondering if the "this" in "this enables" is correct. It may be, but the words in my suggestion above read clearly without me having to think about it.

files/en-us/web/http/guides/fetch_metadata/index.md
hamishwillee: It does. There are many ways you could do this if you think it worth it. Here is one option.

```suggestion
  - : The request was made in `no-cors` mode.
     This means that it is allowed cross-origin without the server sending the appropriate [CORS](/en-US/docs/Web/HTTP/Guides/CORS) headers, with the restriction that the response can't be accessed by JavaScript running in the client (it is  _opaque_).
```

files/en-us/web/http/guides/fetch_metadata/index.md
hamishwillee: Sorry, missed that. 
1. I am not advocating, because I should have seen this ... but it could be added to the list in the linked doc "It indicates whether the requester is from:" as "`none`: no origin at all, if the user navigates directly from ....". 
2. Perhaps "directly user-initiated requests" could be "browser-navigated requests (bookmarks, address bar, etc.). I think I glanced over that and assumed it wasn't covered by this case.

wbamberg: I can see how it's easy to miss given the organization of that section. Addressing it would take a little more than just adding it to the list though because the following paras outline a case of the user clicking a link in a page, which will never result in the value `none`. So it's weird to introduce `none` here, then say "For example..." and have an example that doesn't cover `none`.

I could add a bit to the example to include `none`, but things are still weird because we also say "Similar mappings apply [to] Requests for subresources", which they don't for `none` - `none` is only for documents as far as I can tell.

We could perhaps add an H4 heading for the final paragraph that talks about `none`, just to make it more obvious?

> Perhaps "directly user-initiated requests" could be "browser-navigated requests (bookmarks, address bar, etc.). I think I glanced over that and assumed it wasn't covered by this case.

This is spec language though and I want to keep it because it's really only defined (to the extent that it is defined at all) in the spec, here: https://w3c.github.io/webappsec-fetch-metadata/#directly-user-initiated. I think if we use different language this will be very confusing.

hamishwillee: I accept your arguments for not modifying the linked doc. I don't want to add an H4 - it would work but feels like way overkill. 

I think it might be sufficient to just provide the specific examples here.
```
  // Allow same-origin and directly user-initiated requests (from bookmarks, address bar etc.)
```

Basically that makes it a bit more of a hint about what these are, and that they are not likely to mean the `Sec-Fetch-User`. YOu could maybe add a line after this one in `Sec-Fetch-User`:

>  This header is included only if the request was initiated by a user action (such as a click on a link), and if included, always has the value ?1.

That says something like:

> Note that this differs from direct user interactions via bookmarks or the address bar, which are signaled via `Sec-Fetch-Site` [LINK TO ABOVE section]

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: Yeah, nor am I as a rule. I think in this case though Fetch API is better. I'll add a link to https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API here.

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: Hm, I tried but could not find a reasonably concise way to express this. Basically, all `Sec-Fetch-Site: none` requests will also have `Sec-Fetch-User` set, the inverse does not apply. For example, if the user clicks on a link in a web page, this is a user-initiated request, but since the link is controlled by a website, it does not count as _directly user-initiated_, according the [specification](https://w3c.github.io/webappsec-fetch-metadata/#directly-user-initiated).

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: -> https://github.com/mdn/content/pull/43424/commits/e3c651682fde2b11ab9cb60349bf98d3e1251784 removes "enables"

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: -> https://github.com/mdn/content/pull/43424/commits/e3c651682fde2b11ab9cb60349bf98d3e1251784

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: Moved in https://github.com/mdn/content/pull/43424/commits/e3c651682fde2b11ab9cb60349bf98d3e1251784.

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: 👍 https://github.com/mdn/content/pull/43424/commits/e3c651682fde2b11ab9cb60349bf98d3e1251784

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: -> https://github.com/mdn/content/pull/43424/commits/e3c651682fde2b11ab9cb60349bf98d3e1251784

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: 👍 https://github.com/mdn/content/pull/43424/commits/e3c651682fde2b11ab9cb60349bf98d3e1251784

files/en-us/web/http/guides/fetch_metadata/index.md
wbamberg: Updated comment in https://github.com/mdn/content/pull/43424/commits/e3c651682fde2b11ab9cb60349bf98d3e1251784.

files/en-us/web/http/reference/headers/sec-fetch-dest/index.md
wbamberg: > I'd be tempted in this PR to fix up all the links to this new doc and remove the glossary. Add an issue to fix the others.

I don't really want to do that in this PR, because I do think it's worth having something glossary-ish for these concepts (i.e. something shorter than this guide) but I'm not that clear yet where to put it, and solving that seems to be part of solving it for the other glossary entries.

files/en-us/web/http/guides/fetch_metadata/index.md
hamishwillee: IMO better than it was. Thanks.

files/en-us/web/http/reference/headers/sec-fetch-dest/index.md
hamishwillee: Fair enough. There is certainly nothing "worse" about what you have done here.

wbamberg:
Thank you Hamish! I'm happy to merge this :).

PR #43440, opened 2026-03-14
Clarify extractable parameter behavior for asymmetric keys in SubtleC…

### LINKED ISSUE #39493: Clarify 'extractable' parameter behavior when 'SubtleCrypto.generateKey' is used with asymmetric key algorithms
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey

Parameters, 'extractable'

Throughout the page, references are made to 'the key'. This can be misleading because the `generateKey` function can create pairs of public and private keys, instead of a single key.

It is unclear in such cases, importantly, whether the `extractable` parameter affects the private key, the public key, or both. The `CryptoKeyPair` page doesn't clarify this, either.

In Deno 2.3.1, the public key appears to always be extractable.

Ambiguous or incorrect references to 'the key' should be clarified to reflect the standard behavior more precisely.

_No response_

_No response_

jyotsna05-22:
### Description
Clarified how the `extractable` parameter behaves when SubtleCrypto.generateKey is used with asymmetric algorithms.

Fixes #39493

### DIFF
--- a/files/en-us/web/api/subtlecrypto/generatekey/index.md
+++ b/files/en-us/web/api/subtlecrypto/generatekey/index.md
@@ -33,8 +33,8 @@ generateKey(algorithm, extractable, keyUsages)
     - For [X25519](/en-US/docs/Web/API/SubtleCrypto/deriveKey#x25519): pass the string `X25519` or an object of the form `{ name: "X25519" }`.
 
 - `extractable`
-  - : A boolean value indicating whether it will be possible to export the key using {{domxref("SubtleCrypto.exportKey()")}} or {{domxref("SubtleCrypto.wrapKey()")}}. 
-     Note that when generating asymmetric key pairs (such as RSA or ECDSA), the `extractable` parameter determines whether the private key can be exported; the public key is always extractable.
+  - : A boolean value indicating whether it will be possible to export the key using {{domxref("SubtleCrypto.exportKey()")}} or {{domxref("SubtleCrypto.wrapKey()")}}.
+    Note that when generating asymmetric key pairs (such as RSA or ECDSA), the `extractable` parameter determines whether the private key can be exported; the public key is always extractable.
 - `keyUsages`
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/subtlecrypto/generatekey/index.md
```diff
@@ -33,7 +33,7 @@ generateKey(algorithm, extractable, keyUsages)
     - For [X25519](/en-US/docs/Web/API/SubtleCrypto/deriveKey#x25519): pass the string `X25519` or an object of the form `{ name: "X25519" }`.
 
 - `extractable`
-  - : A boolean value indicating whether it will be possible to export the key using {{domxref("SubtleCrypto.exportKey()")}} or {{domxref("SubtleCrypto.wrapKey()")}}.
+  - : A boolean value indicating whether it will be possible to export the key using {{domxref("SubtleCrypto.exportKey()")}} or {{domxref("SubtleCrypto.wrapKey()")}}. When generating asymmetric key pairs (such as RSA or ECDSA), the `extractable` parameter determines whether the private key can be exported, while the public key in the generated key pair is generally extractable.
```
hamishwillee: Thanks very much.
I'm doing a minor modification because all the assymetric key definitions follow the same pattern, and they all set extractable to true for the public key - such as https://w3c.github.io/webcrypto/#ecdsa-operations-generate-key. Note that your approach of using "generally" was safer (which is usually a good thing) - its just that given there are no exceptions the intent here is pretty clear.


```suggestion
  - : A boolean value indicating whether it will be possible to export the key using {{domxref("SubtleCrypto.exportKey()")}} or {{domxref("SubtleCrypto.wrapKey()")}}. 
     Note that when generating asymmetric key pairs (such as RSA or ECDSA), the `extractable` parameter determines whether the private key can be exported; the public key is always extractable.
```

PR #43445, opened 2026-03-15
Fix the mix-blend-mode compositing spec links to the csswg site

tristanphease:
### Description

Currently the links for explaining the plus-darker and plus-lighter values link to https://drafts.fxtf.org/compositing/#porterduffcompositingoperators_plus_darker which gives a 404 page not found. This updates it to the csswg site spec at https://drafts.csswg.org/compositing-2/#porterduffcompositingoperators_plus_darker

### Motivation

Making this change so readers can see the correct specification for the compositing in mix-blend-mode

### Additional details

Do we need to update all the drafts.fxtf.org links on the site? I happened to notice this one here but seems like there's others on the site that all give 404s or redirect.

### DIFF
--- a/files/en-us/web/css/reference/properties/mix-blend-mode/index.md
+++ b/files/en-us/web/css/reference/properties/mix-blend-mode/index.md
@@ -79,9 +79,9 @@ mix-blend-mode: unset;
 - {{cssxref("&lt;blend-mode&gt;")}}
   - : The blending mode that should be applied.
 - `plus-darker`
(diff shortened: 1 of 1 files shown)


tristanphease:
Should we also look to fix the other links that don't work like these? https://github.com/search?q=repo%3Amdn%2Fcontent%20drafts.fxtf.org&type=code

pepelsbey:
> Should we also look to fix the other links that don't work like these? https://github.com/search?q=repo%3Amdn%2Fcontent%20drafts.fxtf.org&type=code

That would be great, thank you!

PR #43449, opened 2026-03-15
Deduplicate content from CSS gradient color hint guidance

pzhlkj6612:
### Description

In the "*Using CSS gradients*" guidance, I've combined duplicate content about gradient hints.

### Motivation

Hi.

> ### Gradient hints
> By default, the gradient transitions evenly from one color to the next. You can include a color-hint to move the midpoint of the transition value to a certain point along the gradient. In this example, we've moved the midpoint of the transition from the 50% mark to the 10% mark.
> 
> ...
> 
> ### Controlling the progression of a gradient
> 
> By default, a gradient evenly progresses between the colors of two adjacent color stops, with the midpoint between those two color stops being the midpoint color value. You can control the {{Glossary("interpolation")}}, or progression, between two color stops by including a color hint location. In this example, the color reaches the midpoint between lime and cyan 20% of the way through the gradient rather than 50% of the way through. The second example does not contain the hint to highlight the difference the color hint can make: ...
> 
> *from: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Images/Using_gradients*

I've thought that the above two paragraphs are talking about the really same thing. We can make that useful article a bit shorter by removing duplicate thing.

A screenshot of the rendered webpage:

image

Of course, we can preserve the "Controlling the progression of a gradient" heading instead.

### Additional details

I use AI tools. I know what I've written.

none

### DIFF
--- a/files/en-us/web/css/guides/images/using_gradients/index.md
+++ b/files/en-us/web/css/guides/images/using_gradients/index.md
@@ -187,7 +187,7 @@ div {
 
 ### Gradient hints
 
-By default, the gradient transitions evenly from one color to the next. You can include a color-hint to move the midpoint of the transition value to a certain point along the gradient. In this example, we've moved the midpoint of the transition from the 50% mark to the 10% mark.
+By default, a gradient transitions evenly from one color to the next. You can include a color-hint to move the midpoint of the transition value to a certain point along the gradient, letting you control the {{Glossary("interpolation")}}, or progression, between two color stops. In this example, we've moved the midpoint of the transition from the 50% mark to the 10% mark. To highlight the difference the color hint can make, the second example does not contain the hint:
 
 ```html hidden
 <div class="color-hint"></div>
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/guides/images/using_gradients/index.md
```diff
@@ -187,7 +187,7 @@ div {
 
 ### Gradient hints
```
dipikabh: how about we use a combined version
```suggestion
### Controlling the progression of a gradient using color hints
```

files/en-us/web/css/guides/images/using_gradients/index.md
```diff
@@ -280,36 +280,6 @@ In the second example, the second color stop for each color is at the same locat
 
 In both examples, the gradient is written twice: the first is the CSS Images Level 3 method of repeating the color for each stop and the second example is the CSS Images Level 4 multiple color stop method of including two color-stop-lengths in a linear-color-stop declaration.
 
-### Controlling the progression of a gradient
```
dipikabh, line 283: Instead of removing this section, I would suggest removing "### Gradient hints" section because this section explains it better:
- "between the colors of two adjacent color stops"
- "the midpoint between those two color stops being the midpoint color value"
- "the color reaches the midpoint between lime and cyan 20% of the way through the gradient rather than 50% of the way through"

dipikabh (changes requested):
Thanks @pzhlkj6612 for spotting some repeated content and sending a PR for improvement.
Please take a look at my suggestion.

files/en-us/web/css/guides/images/using_gradients/index.md
pzhlkj6612: The combined title looks better! And there is no "Gradient hint" in the Reference page [^mdn] [^source], but "color hint" or "color-hint". Consistency is important.

[^mdn]: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/gradient/linear-gradient
[^source]: https://github.com/mdn/content/blob/0abb70602b0b3b11a2909c417a03e10eabd607a8/files/en-us/web/css/reference/values/gradient/linear-gradient/index.md

files/en-us/web/css/guides/images/using_gradients/index.md
dipikabh: Hi @pzhlkj6612, I would still recommend the reverse approach: keep the second section at L283 ("Controlling the progression of a gradient") and remove the first "Gradient hints" section (now renamed) on L188 instead. The description in the second section is clearer and its example better illustrates the effect.

Two changes required apart from swapping the deleted section:
- Rename line 283's heading to "Controlling the progression of a gradient using color hints".
- Update the ID in the `EmbedLiveSample` macro on L311 to match the updated heading

pzhlkj6612: Hi, @dipikabh! Sorry, I really misunderstood your previous suggestions. Now the "Gradient hints" section has been removed and the "Controlling ..." section has been preserved. The previous renaming was a wrong move.

dipikabh (approved):
Perfect! Thank you.

PR #43454, opened 2026-03-16
FF149 Relnote: CSS font-family: math and MathML

hamishwillee:
FF149 enables the CSS `font-family: math` property and applies it by default on the MathML `<math>` element. This adds a release note and tweaks the property text.

Related docs work can be tracked in #43209


pepelsbey (approved):
Looks good! Thank you 🙂

PR #43460, opened 2026-03-16
Editorial review: Wasm SIMD Load instructions

chrisdavidmills:
### Description

This PR adds reference docs for WebAssembly SIMD load instructions.

### DIFF
--- a/files/en-us/webassembly/reference/simd/load/index.md
+++ b/files/en-us/webassembly/reference/simd/load/index.md
@@ -10,7 +10,7 @@ WebAssembly SIMD load instructions.
 ## Load and load lane
 
 - [`load`](/en-US/docs/WebAssembly/Reference/SIMD/load/load)
-  - : Loads all lanes of a [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretation with the value at a given memory address.
+  - : Loads all lanes of a [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretation with values from a given memory address.
 - [`load8_lane`](/en-US/docs/WebAssembly/Reference/SIMD/load/load8_lane)
   - : Loads a single value from a given memory address into the specified lane of a `v128` `i8x16` value interpretation.
 - [`load16_lane`](/en-US/docs/WebAssembly/Reference/SIMD/load/load16_lane)
--- a/files/en-us/webassembly/reference/simd/load/load/index.md
+++ b/files/en-us/webassembly/reference/simd/load/load/index.md
@@ -7,7 +7,7 @@ browser-compat: webassembly.simd.load
 sidebar: webassemblysidebar
 ---
 
-The **`load`** [SIMD load instruction](/en-US/docs/WebAssembly/Reference/SIMD/load) loads all lanes of a [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretation from a given memory address.
+The **`load`** [SIMD load instruction](/en-US/docs/WebAssembly/Reference/SIMD/load) loads all lanes of a [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretation with values from a given memory address.
 
 {{InteractiveExample("Wat Demo: load", "tabbed-taller")}}
 


files/en-us/webassembly/reference/simd/load/load/index.md
```diff
@@ -0,0 +1,69 @@
...
+- `memory_address`
+  - : an integer representing the memory address to load from.
+- `output`
+  - : The output `v128` value interpretation.
+
+### Binary encoding
+
+| Instruction | Binary equivalent | Example text => binary     |
+| ----------- | ----------------- | -------------------------- |
+| `v128.load` | `0xfd 0:u32`      | `v128.load` => `0xfd 0x00` |
```
yurydelendik: Some fields are missing. Full syntax will look like: v128.load _memarg_, where the memarg is two u32 numbers. In binary form it will be: `0xfd 0:u32 align:u32 offset:u32`. Example needs to reflect that: `v128.load offset=0` => `0xfd 0x00 0x00 0x00`

There are multimemory proposal that will add optional memory index immediate. Also there 64-bit memories proposal -- for such memories `o` will become u64.

The memory/load/index.md does not have encodings -- so the above nuances are not exposed yet in MDN content. Same problem in load*_lane below

chrisdavidmills: I wasn't sure about this; thanks for the information.

In the spec, the memarg syntax is written as "(𝑥, ao):memarg"; are you saying that I should just replace this with "align:u32 offset:u32"?

What about instructions that also require an index — written in the spec as "𝑖:laneidx". Should I just write this as "𝑖:laneidx" in the binary form? Or something else?

chrisdavidmills: Also, in the text format version, I should just write `0` for the offset, not `offset=0`, right? As that is the actual text format syntax?

yurydelendik: > in the text format version, I should just write 0 for the offset, not offset=0, right?

There are two offsets (they will be added as resulting offset):

```
i32.const 4
local.get $arg1
i32.mul
f32.load offset=8
;; loaded f32 from (4 * $arg1 + 8) is on the stack
```
or s-expression equivalent:`(f32.load offset=8 (i32.mul (i32.const 4) (local.get $arg1)))`

yurydelendik: > In the spec, the memarg syntax is written as "(𝑥, ao):memarg"; are you saying that I should just replace this with "align:u32 offset:u32"?

Personally, I have no preference. Whatever is easier to understand.

> What about instructions that also require an index — written in the spec as "𝑖:laneidx". Should I just write this as "𝑖:laneidx" in the binary form? Or something else?

IMHO if it is the part of the example (last column), we need to provide bytes for it. It can be abstract as "𝑖:laneidx" in the encoding column (second column).

yurydelendik: There is also align attribute, but as "offset=" part it is optional and can be omited. Here is another example with align present:
```
      i32.const 0                                                                                                       
      v128.load offset=8 align=4
```
More complex example with load8_lane, memory64, and multi-memory (just for clarify where all future numbers will go):
```
 (module                                                                                                               
    (memory $mem0 1)                    ;; memory 0, 32-bit                                                             
    (memory $mem1 i64 1)                ;; memory 1, 64-bit                                                             
                                                                                                                        
    (data (memory 1) (i64.const 0) "\01\02\03\04\05\06\07\08\09\0a\0b\0c\0d\0e\0f\10")                                  
                                                                                                                        
    (func (export "load") (result i32)                                                                                  
      i64.const 2                       ;; 64-bit address (i64, not i32)
      v128.const i8x16 10 4 6 5 8 7 11 3 8 12 9 7 2 31 9 1                                                              
      v128.load8_lane $mem1 offset=0 align=1 9                                                                          
      i8x16.extract_lane_s 9                                                                                            
    )                                                                                                                   
  )                                                                                                                     
```

chrisdavidmills: I don't understand the offset and align values. From reading the proposal doc, I thought instructions like `load` just took a single integer value as an input, which is a memory address, hence the way I've written the type sections in the docs.

I've also written the examples as looking something like this:

```
(module
  (import "console" "log" (func $log (param i32)))
  (memory $memory 1)
  (data (i32.const 0) "\00\01\02\03\04\05\06\07\08\09\0a\0b\0c\0d\0e\0f\00\01\02\03")
  (func $main
    (v128.load (i32.const 0))
    i32x4.extract_lane 0
    call $log
  )
  (start $main)
)
```

I've just tried the following, which also works:

```
(module
  (import "console" "log" (func $log (param i32)))
  (memory $memory 1)
  (data (i32.const 0) "\00\01\02\03\04\05\06\07\08\09\0a\0b\0c\0d\0e\0f\00\01\02\03")
  (func $main
    (v128.load offset=0 align=2 (i32.const 0))
    i32x4.extract_lane 0
    call $log
  )
  (start $main)
)
```

So, questions:

1. Was my previous example working because offset and align just use default values if you don't specify them? 
2. What are offset and align, and how should I describe them?
3. Where should I include them in the reference pages — in the syntax, or in the types?

yurydelendik: > * Was my previous example working because offset and align just use default values if you don't specify them?

That's correct. (They are default in text format but present in binary)

> * What are offset and align, and how should I describe them?

Default values are good for general use and examples -- I provided them so you will be aware of their presence (hence extra bytes in binary encoding)

> 
> * Where should I include them in the reference pages — in the syntax, or in the types?

It is not really important at this stage. Maybe as TODO or footnote somewhere to note presence of extra fields/data.

files/en-us/webassembly/reference/simd/load/load/index.md
```diff
@@ -0,0 +1,69 @@
...
+- `memory_address`
+  - : An integer representing the memory address to load from.
+- `output`
+  - : The output `v128` value interpretation.
+
+### Binary encoding
+
+| Instruction | Binary equivalent                 | Example text => binary                 |
+| ----------- | --------------------------------- | -------------------------------------- |
+| `v128.load` | `0xfd 0:u32 align:u32 offset:u32` | `v128.load 0` => `0xfd 0x00 0x00 0x00` |
```
yurydelendik: In some toolchains that 0 in the example may stay for memory index, some choose `v128.load (memory 0)`. If the 0 means offset, the "offset=" prefix must be used: `v128.load offset=0`

files/en-us/webassembly/reference/simd/load/load/index.md
chrisdavidmills: I really think we should include them as possible options. To not do so would result in MDN looking incomplete.

files/en-us/webassembly/reference/simd/load/load/index.md
chrisdavidmills: If this means different things to different toolchains, then that ambiguity is a problem. Is there a "full" or "longhanded" form that we can write out that will be unambiguous? We should write documentation that reflects the base language, not a particular toolchain interpretation or variant.

yurydelendik: for `offset=0` it is the same on all toolchains AFIAK. I am pointing out on 0 in `v128.load 0`, which can be a memory index. In MDN, you are using it as `i32.load (memory 0)` in non-SIMD loads

chrisdavidmills:
@hamishwillee another WebAssembly thing for you to put a rubber stamp on, if that's OK sir? This has been thoroughly reviewed, and also put through Claude quite a bit to check for inaccuracies.

files/en-us/webassembly/reference/index.md
```diff
@@ -31,6 +31,8 @@ This page lists references for [WebAssembly](/en-US/docs/WebAssembly).
   - : WebAssembly memory instructions.
 - [WebAssembly numeric instructions](/en-US/docs/WebAssembly/Reference/Numeric)
   - : WebAssembly numeric instructions.
+- [WebAssembly SIMD instructions](/en-US/docs/WebAssembly/Reference/SIMD)
+  - : The SIMD instructions available in Wasm.
```
hamishwillee: Worth spelling out here.

```suggestion
  - : The SIMD (single instruction, multiple data) instructions available in Wasm.
```

files/en-us/webassembly/reference/simd/load/index.md
```diff
@@ -0,0 +1,54 @@
...
+page-type: landing-page
+sidebar: webassemblysidebar
+---
+
+WebAssembly SIMD load instructions.
+
+## Load and load lane
+
+- [`load`](/en-US/docs/WebAssembly/Reference/SIMD/load/load)
+  - : Loads a [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretation from a given memory address.
```
hamishwillee: What's a value interpretation? 

I suspect this means "Load all lanes of a v128 from a memory address", and that is probably clearer.

files/en-us/webassembly/reference/simd/load/index.md
```diff
@@ -0,0 +1,54 @@
...
+- [`load8_lane`](/en-US/docs/WebAssembly/Reference/SIMD/load/load8_lane)
+  - : Loads a single value from a given memory address into the specified lane of a `v128` `i8x16` value interpretation.
```
hamishwillee, line 15: For all these I don't understand what value interpretation. means, but I otherwise cool.

files/en-us/webassembly/reference/simd/load/index.md
```diff
@@ -0,0 +1,54 @@
...
+  - : Loads a single value from a given memory address into the specified lane of a `v128` `i16x8` value interpretation.
+- [`load32_lane`](/en-US/docs/WebAssembly/Reference/SIMD/load/load32_lane)
+  - : Loads a single value from a given memory address into the specified lane of a `v128` `i32x4` value interpretation.
+- [`load64_lane`](/en-US/docs/WebAssembly/Reference/SIMD/load/load64_lane)
+  - : Loads a single value from a given memory address into the specified lane of a `v128` `i64x2` value interpretation.
+
+## Load and extend
+
+- [`load8x8_s`](/en-US/docs/WebAssembly/Reference/SIMD/load/load8x8_s)
+  - : Loads eight 8-bit integers from a given memory address and sign extends each one to a 16-bit lane, outputting a `v128` `i16x8` value interpretation.
```
hamishwillee, line 26: What does sign-extend mean? It isn't guessable.

You might need to explain this everywhere.

files/en-us/webassembly/reference/simd/load/index.md
```diff
@@ -0,0 +1,54 @@
...
+- [`load8x8_u`](/en-US/docs/WebAssembly/Reference/SIMD/load/load8x8_u)
```
hamishwillee, line 27: What does zero-extend mean - I assumed zero extend means "zero fill the extra bytes in the new lanes"?

files/en-us/webassembly/reference/simd/load/load16_lane/index.md
```diff
@@ -0,0 +1,88 @@
+---
+title: "load16_lane: Wasm SIMD load instruction"
+short-title: load16_lane
+slug: WebAssembly/Reference/SIMD/load/load16_lane
+page-type: webassembly-instruction
+browser-compat: webassembly.simd.load16_lane
+sidebar: webassemblysidebar
+---
+
+The **`load16_lane`** [SIMD load instruction](/en-US/docs/WebAssembly/Reference/SIMD/load) loads a single value from a given memory address into the specified lane of a [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) type `i16x8` value interpretation.
```
hamishwillee, line 10: What happes in all the other lanes? 0 filled?

files/en-us/webassembly/reference/simd/load/load16_lane/index.md
```diff
@@ -0,0 +1,88 @@
...
+WebAssembly.instantiateStreaming(fetch("{%wasm-url%}"), { console });
+```
+
+## Syntax
+
+```plain
+;; Common usage
+v128.load16_lane lane_value
+
+;; With optional immediates
```
hamishwillee, line 40: "Immediates" is a new term. Is it defined anywhere?

files/en-us/webassembly/reference/simd/load/index.md
chrisdavidmills: What a value interpretation is, is briefly explained on the `v128` type page that I have linked from everywhere. Ideally, we would explain this kind of thing on a guide page somewhere, and we will eventually provide some guides. But this is the best I've got for now.

"loads all lanes" is better, yes. I've updated that in both places.

files/en-us/webassembly/reference/simd/load/index.md
chrisdavidmills: See above comment

files/en-us/webassembly/reference/simd/load/index.md
chrisdavidmills: See https://en.wikipedia.org/wiki/Sign_extension for information on sign- and zero-extend.

The trouble is, this kind of definition isn't the sort of thing we normally include on MDN, and the target audience of this reference is likely to know what they mean already.

Do you think I should just add very brief glossary entries for each one, and then link to them from everywhere?

files/en-us/webassembly/reference/simd/load/index.md
chrisdavidmills: See above.

files/en-us/webassembly/reference/simd/load/load16_lane/index.md
chrisdavidmills: No; the other lanes are unchanged. Only the specified lane's content is changed, to the value found at the given memory address.

files/en-us/webassembly/reference/simd/load/load16_lane/index.md
chrisdavidmills: No, but this is the kind of detail obvious to those using the reference material that would be explained in a guide somewhere, which we will add later.

Wasm instructions effectively have two types of input value that they use to function: values taken from the stack, which are generally used to perform some kind of calculation, and immediates, which are kind of like flags that further modify the instruction behavior if included.

They are both sort of like parameters in JS, but not exactly.

files/en-us/webassembly/reference/simd/load/index.md
hamishwillee: If it were me I would add a "See also" to that wikipedia entry on these cases. But you make a good point about audience, so I leave to your discretion (non blocking suggestion)

files/en-us/webassembly/reference/simd/load/load16_lane/index.md
hamishwillee: Doh, thanks. For some reason I wasn't thinking "load" but "create". 

I was going to suggest adding "... replacing any exising content in the lane". But that seems like too bleedingly obvious.

files/en-us/webassembly/reference/simd/load/load16_lane/index.md
hamishwillee: Ah, I recall now. Fair enough.

files/en-us/webassembly/reference/simd/load/index.md
```diff
@@ -0,0 +1,54 @@
...
+page-type: landing-page
+sidebar: webassemblysidebar
+---
+
+WebAssembly SIMD load instructions.
+
+## Load and load lane
+
+- [`load`](/en-US/docs/WebAssembly/Reference/SIMD/load/load)
+  - : Loads all lanes of a [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretation from a given memory address.
```
hamishwillee: Could this be misinterpreted as reading the memory and populating all the lanes from sequential memory rather than one value at the address used in all langes.

```suggestion
  - : Loads all lanes of a [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretation with the value at a given memory address.
```

files/en-us/webassembly/reference/simd/load/index.md
chrisdavidmills: Yeah, I think I'll leave this one and add links or more information later if we see the need.

PR #43468, opened 2026-03-17
FF149 ExprFeat/Relnote: CSS text-box and longhand props

hamishwillee:
FF149 adds support for the shorthand CSS property [`text-box`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-box) , corresponding longhand properties [`text-box-trim`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-box-trim) and [`text-box-edge`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-box-edge), and the value [`text-edge`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/text-edge) behind the preference `layout.css.text-box.enabled` in https://bugzilla.mozilla.org/show_bug.cgi?id=2013458.

This adds an experimental features entry and a release note entry that are essentially the same.
Tried to catch a flavour of what this does, which is make it possible to align text within a box so that it is visually positioned consistently for different fonts.

Related docs work can be tracked in #43208

### DIFF
--- a/files/en-us/mozilla/firefox/experimental_features/index.md
+++ b/files/en-us/mozilla/firefox/experimental_features/index.md
@@ -319,6 +319,22 @@ The CSS `text-decoration-trim` property allows you to specify {{cssxref("text-de
 - `layout.css.text-decoration-trim.enabled`
   - : Set to `true` to enable.
 
+### `text-box`, `text-box-edge`, `text-box-trim`
+
+The CSS {{cssxref("text-box")}} shorthand property, or the corresponding {{cssxref("text-box-trim")}} and {{cssxref("text-box-edge")}} longhand properties, allow you to trim space from the block-start and block-end edges of a text element's block container, aligning the container edges with the actual ink bounds of the text rather than the abstract box edges defined by font metrics.
+This makes it possible to achieve consistent visual spacing around text regardless of which font is used.
+([Firefox bug 2013458](https://bugzil.la/2013458)).
+
+| Release channel   | Version added | Enabled by default? |
+| ----------------- | ------------- | ------------------- |
+| Nightly           | 149           | No                  |
+| Developer Edition | 149           | No                  |
+| Beta              | 149           | No                  |
+| Release           | 149           | No                  |
+
+- `layout.css.text-box.enabled`
+  - : Set to `true` to enable.
+
 ### `@custom-media` at-rule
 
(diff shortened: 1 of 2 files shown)


files/en-us/mozilla/firefox/experimental_features/index.md
```diff
@@ -319,6 +319,22 @@ The CSS `text-decoration-trim` property allows you to specify {{cssxref("text-de
 - `layout.css.text-decoration-trim.enabled`
   - : Set to `true` to enable.
 
+### `text-box`, `text-box-edge`, `text-box-trim`
```
dipikabh, line 322: since we're ordering by release, this update should go after the "### `@custom-media` at-rule" entry

dipikabh:
```suggestion
### `text-box`, `text-box-edge`, and `text-box-trim`
```

files/en-us/mozilla/firefox/experimental_features/index.md
```diff
@@ -319,6 +319,22 @@ The CSS `text-decoration-trim` property allows you to specify {{cssxref("text-de
...
+
+The CSS {{cssxref("text-box")}} shorthand property, or the corresponding {{cssxref("text-box-trim")}} and {{cssxref("text-box-edge")}} longhand properties, allow you to trim space from the block-start and block-end edges of a text element's block container, aligning the container edges with the actual ink bounds of the text rather than the abstract box edges defined by font metrics.
```
dipikabh, line 324: - Does `<text-edge>` also need to be mentioned?
- This sentence is too long, can probably be split:
  - List features that have been added (that's where text-edge can probably be included)
  - "These allow you to.." (also "aligning" can be replaced to add another pause: "...block container; as a result, the container edges align with..."

Some other nits: 
- The first "or" should be "and"
- `text-box-trim` and `text-box-edge` can be swapped to match the order in the heading

files/en-us/mozilla/firefox/experimental_features/index.md
```diff
@@ -319,6 +319,22 @@ The CSS `text-decoration-trim` property allows you to specify {{cssxref("text-de
...
+This makes it possible to achieve consistent visual spacing around text regardless of which font is used.
```
dipikabh, line 325:
```suggestion
This makes it possible to achieve consistent visual spacing around text regardless of the font used.
```

files/en-us/mozilla/firefox/releases/149/index.md
dipikabh, line 1: same edits can be applied here

hamishwillee:
@dipikabh Thanks for this. I just got a note from the developer that this is too early to be tested. I'm therefore going to not merge this!

PR #43478, opened 2026-03-17
Add Baseline banner to Fullscreen API page

dfabulich:
This page was referencing the `browser-compat` data for the deprecated `api.Document.fullscreen` feature, which doesn't have a corresponding feature ID in `web-features`. As a result, it doesn't display a Baseline banner.

By removing that key from `browser-compat`, the page will show the correct Baseline status banner.

### DIFF
--- a/files/en-us/web/api/fullscreen_api/index.md
+++ b/files/en-us/web/api/fullscreen_api/index.md
@@ -7,7 +7,6 @@ browser-compat:
   - api.Document.fullscreenEnabled
   - api.Document.exitFullscreen
   - api.Element.requestFullscreen
-  - api.Document.fullscreen
 ---
 
 {{DefaultAPISidebar("Fullscreen API")}}


files/en-us/web/api/fullscreen_api/guide/index.md
```diff
@@ -3,8 +3,10 @@ title: Guide to the Fullscreen API
 slug: Web/API/Fullscreen_API/Guide
 page-type: guide
 browser-compat:
+  - api.Document.fullscreenElement
   - api.Document.fullscreenEnabled
-  - api.Document.fullscreen
+  - api.Document.exitFullscreen
+  - api.Element.requestFullscreen
```
hamishwillee: All of these can be removed. We don't usually include these in guides

files/en-us/web/api/fullscreen_api/index.md
```diff
@@ -7,7 +7,6 @@ browser-compat:
   - api.Document.fullscreenEnabled
   - api.Document.exitFullscreen
   - api.Element.requestFullscreen
-  - api.Document.fullscreen
```
hamishwillee, line 10: Thanks. It works but I'm not sure this is the "right" fix. 

@caugner Is there any reason we can't add  "a corresponding feature ID in web-features" for a`pi.Document.fullscreen`?

dfabulich: I think my PR is the right fix (though I'm not 100% sure there isn't a better solution)… but I know that adding a `web-features` ID for `api.Document.fullscreen` is _not_ the right fix.

1. The `web-features` build will not allow you to create a feature for `api.Document.fullscreen`; if you try, its CI will fail with this error.
    > error: fullscreen: contains deprecated compat feature api.Document.fullscreen. This is forbidden for non-discouraged published features.

    (I did try that approach before filing this PR!)
2. Let's suppose we somehow made an exception for `api.Document.fullscreen`, giving it its own separate `legacy_fullscreen` feature. Well, even then, `fred` would refuse to show a Baseline banner on this page, because pages whose `browser-compat` section lists references multiple different `web-features` never show Baseline banners. https://github.com/mdn/fred/issues/1391

hamishwillee: @dfabulich Sorry, perhaps I should have just been more clear: it is not the right fix.
Baseline needs to be made to work with the docs: we don't mess with docs to get baseline to work. 

The right thing to to do is raise an issue against that infrastructure. I've only left this open to keep it visible until @caugner can look. IF you want to create an issue against the https://github.com/mdn/fred (I think?) then we can close this.

dfabulich: I strongly disagree with your characterization of this PR as "messing with docs to get baseline to work." Removing `api.Document.fullscreen` from these pages is the right thing to do, regardless of Baseline.

* https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API/Guide doesn't mention or document `api.Document.fullscreen` at all. It is completely wrong for `api.Document.fullscreen` to appear in the `browser-compat` list on this guide page; removing it is the right thing to do, and it would be the right thing to do even if that had no effect on the Baseline banner.
* https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API only links to `api.Document.fullscreen` to say that it's legacy/deprecated. This page isn't about `api.Document.fullscreen`; it should be removed from this page also.

Cleaning up invisible metadata wouldn't really be worth filing. The fact that removing `api.Document.fullscreen` _also_ makes Baseline work is what made this cleanup PR worth filing.

More deeply, Baseline is not allowed to show a banner when there are multiple web features, one of which is deprecated.

* I'm not allowed to add `api.Document.fullscreen` to `web-features`, so as long as `api.Document.fullscreen` appears in the `browser-compat` of this page, it won't have a Baseline banner. There is no reasonable issue anyone could file on `fred` that could possibly make that work.
* I've already filed https://github.com/mdn/fred/issues/1391 and fixed it in https://github.com/mdn/rari/pull/582 but it's labeled "needs content decision" so it's not allowed to merge.
* But even that PR would only affect pages with multiple web-features where all of them are "Widely Available." There is no issue I can correctly file on `fred` that would show a banner on pages referring to multiple `web-features` IDs where one of them is "Limited Availability" and one of them is deprecated.

In https://github.com/mdn/content/pull/43480#issuecomment-4095672217 you wrote:

> * if a page has more than one browser compat entry then the compatibility is not shown. I'd argue that it would be useful to be able to note that parts of an API are very stable and parts are not.

Good thinking! But, I already tried that, in https://github.com/mdn/rari/pull/527 and it was closed without merging, with this explanation:

> As written, this would show a "Limited availability" banner on https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API, which is not really what I think we want.

I can't fix Baseline because it would show the wrong results on Fetch API, but I can't fix Fetch API because I should have fixed Baseline instead.

I'm not sure it's possible to fix bugs in this system. Every part of the system says that it must not be solved _here_ it must be solved somewhere else, anywhere but here.

files/en-us/web/api/fullscreen_api/guide/index.md
```diff
@@ -2,9 +2,7 @@
 title: Guide to the Fullscreen API
 slug: Web/API/Fullscreen_API/Guide
 page-type: guide
-browser-compat:
-  - api.Document.fullscreenEnabled
-  - api.Document.fullscreen
+browser-compat: api.Document.fullscreenEnabled
```
caugner: Is it normal for guides to have compat tables? And if it is, shouldn't this guide have the same `browser-compat` values as the parent page?

files/en-us/web/api/fullscreen_api/index.md
caugner: I don't have concerns about removing the deprecated feature from `browser-compat` here, but it would be good to have a second opinion from @LeoMcA.

files/en-us/web/api/fullscreen_api/guide/index.md
hamishwillee: No. These should be removed. But I'll continue that discussion on the other comment.

files/en-us/web/api/fullscreen_api/index.md
hamishwillee: @caugner (@LeoMcA) This was more a query about 
- "is there anything that BCD/Rari/Fred can do to help in cases where a page has multiple browser compat entries". It sounds from @dfabulich comment above is that this is a known issue that you consider not worth the cost/effort of fixing - https://github.com/mdn/rari/pull/527.
- It seems crazy that the existance of a well supported but deprecated API would stop the baseline from appearing.  Is the reason that `api.Document.fullscreen` can't be added to web-features that it is deprecated, so it isn't part of the platform? What are the other reasons something can't be a web-feature? 
  - What I'm hoping is that perhaps we could exclude this from the baseline calculation, instead of using it as a reason to not display it. 
- A better example of where the inability to characterise mixed support is a problem (from another PR @dfabulich posted) is the Fetch API. This is well established API but has been extended with deferred fetch, which is poorly supported. As a result there is no no baseline. It's all one API so it would be good to be able to provide a more nuanced overview. I wonder if we might consider a new document/content type for a sub API?

Anyway ...

@dfabulich In principle my characterisation is correct for the overview page, which is allowed and expected to have browser compat data. Baseline builds on BCD. It is hard to reflect the inconvenient reality of mixed support levels like this, but if the team owning baseline don't think it is worth fixing then  what you want to do is not supported.

Pragmatically you're right - removal of this key from the overview page is only going to help. Note, no guarantee automated tooling or "someone" won't put it back.

Guide pages generally do not have browser compat data - though this is "allowed" for newer APIs where support is sketchy. This  data is definititely going to be removed when someone notices. Let's remove it now.

dfabulich: > It seems crazy that the existence of a well supported but deprecated API would stop the baseline from appearing.  Is the reason that `api.Document.fullscreen` can't be added to web-features that it is deprecated, so it isn't part of the platform? What are the other reasons something can't be a web-feature?

https://github.com/web-platform-dx/web-features/blob/c256e6d56a609f69b8d9f8ddb315a7185967592b/scripts/dist.ts#L279

This seems to be the only blocked thing in `web-features`. The block was added here https://github.com/web-platform-dx/web-features/commit/e51f69af605dbaabef508619f750e919da16e665

FWIW, I don't think `web-features` considers either the legacy fullscreen or the modern Fullscreen API to be "well supported." https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenElement is listed as "Limited Availability" because of its mobile Safari support, where it's only available on iPad, not iPhone, and because it "shows an overlay button which can not be disabled. Swiping down exits fullscreen mode, making it unsuitable for some use cases like games." If `web-features` did define a feature for the legacy fullscreen API, it would be considered "Limited Availability" as well.

dfabulich: I've filed a separate PR to remove browser-compat data from the guide. 

* https://github.com/mdn/content/pull/43522 

And I've force-pushed this PR to only remove the deprecated `api.Document.fullscreen` feature from the fullscreen API page.

hamishwillee: Thanks for that. I've merged your other PR. Let's see what Claus and Leo say about the rest before we merge this.

LeoMcA:
> I don't have concerns about removing the deprecated feature from `browser-compat` here, but it would be good to have a second opinion from @LeoMcA.

In this instance, I think this is the correct thing to do.

PR #42693, opened 2026-01-08
Integrate fetchLater() API into Fetch

wbamberg:
Fixes https://github.com/mdn/content/issues/42606. Integrates the "fetchLater API" into Fetch.

In this PR I've:

- created a new guide "Using Deferred Fetch", under [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)
- integrated https://developer.mozilla.org/en-US/docs/Web/API/fetchLater_API into the Fetch API docs by relocating "concepts and usage" into the new "Using Deferred Fetch" guide, and the interface listing into the [Fetch API overview page](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
- integrated the [quotas guide](https://developer.mozilla.org/en-US/docs/Web/API/fetchLater_API/fetchLater_quotas) into the Fetch API docs by making it a section of the new "Using Deferred Fetch" guide.
- deleted the [fetchLater API](https://developer.mozilla.org/en-US/docs/Web/API/fetchLater_API) page and its child guide, and redirected to the relevant new places in the Fetch API.
- updated groupdata.json to merge fetchLater into Fetch.
- fixed up all the links and the sidebars.

I **have not** changed any content except where it was essential for things to still make sense. Except I removed the BCD and spec tables from the "quotas" guide because we don't generally include these sections in guides.

Issue #42606, opened 2026-01-02
fetchLater() ought to be documented as part of the Fetch API


wbamberg:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/fetchLater_API

### What information was incorrect, unhelpful, or incomplete?

The `fetchLater()` function is part of the Fetch API: it's defined in the Fetch spec. But in MDN it's documented as its own standalone `fetchLater()` API, and not even linked from the Fetch API docs.

This is bad for discoverability and coherence.

### What did you expect to see?

`fetchLater()` and associated interfaces documented as part of the Fetch API.

The page at https://developer.mozilla.org/en-US/docs/Web/API/fetchLater_API should be split up, with "concepts and usage" probably made into a separate guide page under https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API, and the interfaces and HTTP headers listed inside https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API.

The groupdata entry should be removed and its contents included in the entry for Fetch.

MUSKANNISHAD:
This makes sense. Happy to help with documentation changes once there’s agreement on whether fetchLater() should live under Fetch API docs or remain standalone.

PR #43480, opened 2026-03-17
Factor out Deferred Fetch documentation, shows Baseline on Fetch API

### LINKED ISSUE #43479: "Fetch API" has no Baseline banner, because it also documents `fetchLater`
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

Baseline banner

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API is primarily about the good old `fetch` API, which is widely available, but it's currently serving double duty, documenting the `fetchLater` "Deferred Fetch" API as well.

The `fetchLater` API has limited availability https://developer.mozilla.org/en-US/docs/Web/API/Window/fetchLater and is a completely different feature.

We can only show one Baseline banner at the top of the page; since the page refers to two different features with two different Baseline statuses, the page just doesn't show a banner.

The `fetchLater` API should be documented separately. This page should exclusively be about the regular Fetch API, and should show a Baseline: Widely Available banner.

_No response_

_No response_

dfabulich:
### Description

Factor out Deferred Fetch documentation from the main "Fetch API" page.

### Motivation

After this change, we now show a "Baseline: Widely Available" banner on the "Fetch API" page, and a "Baseline: Limited Availability" banner on the "Using Deferred Fetch" page.

Fixes #43479

### DIFF
--- a/files/en-us/web/api/fetch_api/index.md
+++ b/files/en-us/web/api/fetch_api/index.md
@@ -2,9 +2,7 @@
 title: Fetch API
 slug: Web/API/Fetch_API
 page-type: web-api-overview
-browser-compat:
-  - api.fetch
-  - api.Window.fetchLater
+browser-compat: api.fetch
 ---
 
 {{DefaultAPISidebar("Fetch API")}}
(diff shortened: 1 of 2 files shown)


Josh-Cena:
It was explicitly decided that fetchLater should be part of fetch because it's the same spec and reuses the same machinery: https://github.com/mdn/content/pull/42693 and https://github.com/mdn/content/issues/42606.

dfabulich:
My primary goal here is to put a big green "Baseline: Widely Available" banner on https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API.

There's kinda no reasonable way to show a banner on that page when it's documenting a mix of `fetch` and `fetchLater`.

Maybe the moral of this story is to just wait until `fetchLater` reaches baseline (a minimum of three years from now…?) but that feels like a counterintuitive conclusion.

Fetch API is hardly alone in having this issue.

* https://developer.mozilla.org/en-US/docs/Web/API/File_System_API doesn't have a banner, because it's a mix of OPFS (widely available) and File System Access API (limited availability, [opposed by WebKit](https://github.com/WebKit/standards-positions/issues/28) and [opposed by Mozilla](https://github.com/mozilla/standards-positions/issues/154)).
* https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API doesn't have a banner, because it's a mix of classic WebSockets (widely available) and WebSocketStream (experimental, non-standard; obsoleted by WebTransport API)

Josh-Cena:
But then we don't have a consistent editorial definition of what an "API" is. Currently our definition is more or less "one spec, one API"; the only exceptions I know are HTML and DOM which are split because of how big their scopes are, and some specs that are coalesced into a single API because of how unified their scopes are. Most specs can be extended, and it would be really awkward if each extension has to become its own API just for the sake of preserving the old API's baseline. After all, API groups don't have the concept of "baseline" anyway; only individual interfaces do.

hamishwillee:
Thanks @dfabulich.

As @Josh-Cena says, we can't have these changes; 
- it's all one API, parts of which are broadly supported and parts that are not.  
- Using Deferred Fetch is a guide page. Guide pages don't have those API overview sections. They should not, but sometimes do have `browser-compat`.

The intent is great, but baseline needs to change to fit the docs, not the docs to the baseline.

It would be good is to create an issue in https://github.com/mdn/fred/issues to note that
- if a page has more than one browser compat entry then the compatibility is not shown. I'd argue that it would be useful to be able to note that parts of an API are very stable and parts are not.
- there are pages that might benefit from the banner, such as Using Deferred Fetch, but that should not have browser-compat. It may be they will say "add the compat key but omit the spec and compat sections", or they might say "it isn't an overview for an API so it doesn't get the banner".

Closing this though.

PR #43493, opened 2026-03-19
Editorial review: Document paintTime and presentationTime

chrisdavidmills:
### Description

Chrome 145 adds support for the `presentationTime` and `paintTime` to relevant performance entries (`PerformancePaintTiming`, `LongestContentfulPaint`, `PerformanceElementTiming`, and `PerformanceLongAnimationFrameTiming`). See https://chromestatus.com/feature/5162859838046208.

This PR documents these properties on each separate interface and adds a few other appropriate mentions besides. Let me know if you feel other mentions or coverage are needed.

### DIFF
--- a/files/en-us/web/api/largestcontentfulpaint/presentationtime/index.md
+++ b/files/en-us/web/api/largestcontentfulpaint/presentationtime/index.md
@@ -14,7 +14,7 @@ The `presentationTime` is optional — some browsers may choose always to return
 
 ## Value
 
-A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("null")}} if the value is not exposed.
+A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("Operators/null", "null")}} if the value is not exposed.
 
 ## Examples
 
--- a/files/en-us/web/api/performanceelementtiming/presentationtime/index.md
+++ b/files/en-us/web/api/performanceelementtiming/presentationtime/index.md
@@ -14,7 +14,7 @@ The `presentationTime` is optional — some browsers may choose always to return
 
 ## Value
 
-A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("null")}} if the value is not exposed.
+A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("Operators/null", "null")}} if the value is not exposed.
 
 ## Examples
 
--- a/files/en-us/web/api/performancelonganimationframetiming/presentationtime/index.md
+++ b/files/en-us/web/api/performancelonganimationframetiming/presentationtime/index.md
@@ -8,13 +8,13 @@ browser-compat: api.PerformanceLongAnimationFrameTiming.presentationTime
 
 {{APIRef("Performance API")}}
 
-The **`presentationTime`** read-only property of the {{domxref("PerformanceLongAnimationFrameTiming")}} interface returns the {{domxref("DOMHighResTimeStamp","timestamp")}} when the element was actually drawn on the screen.
+The **`presentationTime`** read-only property of the {{domxref("PerformanceLongAnimationFrameTiming")}} interface returns the {{domxref("DOMHighResTimeStamp","timestamp")}} when the UI update was actually drawn on the screen.
 
 The `presentationTime` is optional — some browsers may choose always to return `0` or not to expose the value at all. The value is also implementation-dependent — it may differ across browsers that choose to expose it.
 
 ## Value
 
-A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("null")}} if the value is not exposed.
+A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("Operators/null", "null")}} if the value is not exposed.
 
 ## Examples
 
--- a/files/en-us/web/api/performancepainttiming/presentationtime/index.md
+++ b/files/en-us/web/api/performancepainttiming/presentationtime/index.md
@@ -14,7 +14,7 @@ The `presentationTime` is optional — some browsers may choose always to return
 
 ## Value
 
-A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("null")}} if the value is not exposed.
+A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("Operators/null", "null")}} if the value is not exposed.
 
 ## Examples
 


files/en-us/web/api/largestcontentfulpaint/painttime/index.md
```diff
@@ -0,0 +1,33 @@
+---
+title: "LargestContentfulPaint: paintTime property"
+short-title: paintTime
+slug: Web/API/LargestContentfulPaint/paintTime
+page-type: web-api-instance-property
+browser-compat: api.LargestContentfulPaint.paintTime
+---
+
+{{APIRef("Performance API")}}{{AvailableInWorkers}}
```
mmocny: I notice in the preview that this api "is" available in workers-- but I think it isn't-- at least indirectly in that LCP itself wont ever be available in workers.

But Im not sure how to fix this since it seems automated.

files/en-us/web/api/largestcontentfulpaint/presentationtime/index.md
```diff
@@ -0,0 +1,33 @@
...
+slug: Web/API/LargestContentfulPaint/presentationTime
+page-type: web-api-instance-property
+browser-compat: api.LargestContentfulPaint.presentationTime
+---
+
+{{APIRef("Performance API")}}{{AvailableInWorkers}}
+
+The **`presentationTime`** read-only property of the {{domxref("LargestContentfulPaint")}} interface returns the {{domxref("DOMHighResTimeStamp","timestamp")}} when the painted pixels were actually drawn on the screen.
+
+The `presentationTime` is implementation-dependant, therefore the value may differ across browsers.
```
mmocny: This value is also "optional" and some implementations return 0 always

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -61,9 +48,24 @@ It also supports the following properties:
...
   - : If the element is an image, the request url of the image.
 
+It also extends the following {{domxref("PerformanceEntry")}} properties, qualifying and constraining them as described:
+
+- {{domxref("PerformanceEntry.entryType")}} {{ReadOnlyInline}} {{Experimental_Inline}}
+  - : Returns `"largest-contentful-paint"`.
+- {{domxref("PerformanceEntry.name")}} {{ReadOnlyInline}} {{Experimental_Inline}}
+  - : Always returns an empty string.
+- {{domxref("PerformanceEntry.startTime")}} {{ReadOnlyInline}} {{Experimental_Inline}}
+  - : Returns the value of this entry's {{domxref("LargestContentfulPaint.renderTime", "renderTime")}} if it is not `0`, otherwise the value of this entry's {{domxref("LargestContentfulPaint.loadTime", "loadTime")}}.
```
mmocny: This technically matches spec wording, but afaik renderTime can never be 0 any more on any modern browser implementation.

It was 0 on historical versions on Chromium.

@noamr if we should fix the spec wording at this point?

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -89,6 +91,32 @@ const observer = new PerformanceObserver((list) => {
...
+    console.log(
+      `LCP paint time: ${lastEntry.paintTime} milliseconds; presentation time: ${lastEntry.presentationTime} milliseconds.`,
+    );
+    // Logs:
+    // LCP paint time: 473.30000001192093 milliseconds;
+    // presentation time: 516 milliseconds.
+  } else {
+    console.log(`LCP start time: ${lastEntry.startTime} milliseconds.`);
+    // Logs:
+    // LCP start time: 516 milliseconds.
```
mmocny: Nit: the comment of being 516ms matches the presetnation time of 516ms (above).  This can be confusing, since this fallback will only be used when there is no available `presentationTime`, and that implies this will be `paintTime` value instead.

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -89,6 +91,32 @@ const observer = new PerformanceObserver((list) => {
...
+  const entries = list.getEntries();
+  const lastEntry = entries[entries.length - 1]; // Use the latest LCP candidate
+  if (lastEntry.paintTime && lastEntry.presentationTime) {
+    console.log(
+      `LCP paint time: ${lastEntry.paintTime} milliseconds; presentation time: ${lastEntry.presentationTime} milliseconds.`,
+    );
+    // Logs:
+    // LCP paint time: 473.30000001192093 milliseconds;
+    // presentation time: 516 milliseconds.
+  } else {
```
mmocny, line 115: Nit: This fallback will be used when:

- Non-chromium browsers that don't have `presentationTime`, or
- Older chromium browsers that don't have PaintTimingMixin at all

This makes it a bit complex what the fallback values are...

If you want perfect coverage, you would do something like:

```js
if (lastEntry.presentationTime) {
  console.log("LCP.paintTime:", lastEntry.paintTime, ".presentationTime:", lastEntry.presentationTime);
} else if (lastEntry.paintTime) {
  console.log("LCP.paintTime:", lastEntry.paintTime);
} else if (lastEntry.startTime != lastEntry.loadTime) {
  console.log("LCP:", lastEntry.startTime);
} else {
  console.log("LCP.loadTime:", lastEntry.loadTime);
}
```

(The last 2 cases are historical.)

If you only want to describe modern usage, I would just do:

```js
if (lastEntry.presentationTime) {
  console.log("LCP.paintTime:", lastEntry.paintTime, ".presentationTime:", lastEntry.presentationTime);
} else {
  console.log("LCP.paintTime:", lastEntry.paintTime);
}
```

files/en-us/web/api/performance_api/long_animation_frame_timing/index.md
```diff
@@ -112,7 +118,7 @@ The timestamps provided in the {{domxref("PerformanceLongAnimationFrameTiming")}
 
 | Timing                            | Calculation                                                              |
 | --------------------------------- | ------------------------------------------------------------------------ |
-| Start time                        | `startTime`                                                              |
+| Start time                        | `startTime` (or `presentationTime`)                                      |
```
mmocny: I think you should drop the "or presentation time" since start time for LoAF is task/rendering start, unlike paint timings.

perhaps this was supposed to go to "End time" instead?

start + duration should == paintTime or presentationTime

files/en-us/web/api/performanceelementtiming/index.md
```diff
@@ -90,6 +94,33 @@ observer.observe({ type: "element", buffered: true });
 
 Two entries will be output to the console. The first containing details of the image, the second with details of the text node.
 
+### Observing separate paint and presentation timings
+
+The `paintTime` and `presentationTime` properties enable you to retrieve specific timings for the paint phase starting and the element being drawn on the screen. The `paintTime` is broadly interoperable, whereas the `presentationTime` is implementation-dependant.
+
+This example uses a `PerformanceObserver` to observe all performance entries of type `"element"` (remember that, to be observed, elements need to have `elementtiming` attributes set on them). We check for `paintTime` and `presentationTime` support and retrieve those values if they are available. In non-supporting browsers, the code retrieves the `startTime`.
+
+```js
```
mmocny, line 103: (same as above, just noting).

files/en-us/web/api/largestcontentfulpaint/index.md
noamr: Sure!

files/en-us/web/api/largestcontentfulpaint/painttime/index.md
chrisdavidmills: Ah, good spot — this was caused by me copy/pasting the original page I used as a template from elsewhere. I've fixed this by removing the `{{AvailableInWorkers}}` macro call from all the property pages.

files/en-us/web/api/largestcontentfulpaint/presentationtime/index.md
chrisdavidmills: OK, good to know. The spec doesn't seem to say anything about "optional" or `0`; it does indicate that the value can be `null`, although it isn't very clear about when `null` might be returned.

I thought about this a bit, and updated the commented sentence with:

> The `presentationTime` is implementation-dependent, therefore the value may differ across implementing browsers. Some browsers may choose to always return `0` or not to expose the value.

I also updated the "Value" section to say:

> A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("null")}} if the value is not exposed.

Is this fair to say? I'm really using educated guesses here.

files/en-us/web/api/largestcontentfulpaint/index.md
chrisdavidmills: This doesn't relate to my work on this PR, but I'm happy to change this in a future PR. Let me know when the spec wording gets updated.

files/en-us/web/api/largestcontentfulpaint/index.md
chrisdavidmills: I think I will just delete the log comments to avoid the potential for confusion. People know what a console.log looks like.

noamr:
What @mmocny wrote is sufficient. Thanks!

files/en-us/web/api/largestcontentfulpaint/index.md
chrisdavidmills: This makes sense; thanks for the useful info. 

I've actually gone for updating the three examples (on `PerformanceElementTiming`, `PerformancePaintTiming`, and `LongestContentfulPaint`) to use the "perfect coverage" example you've provided — I think this tells a better story for the history of the available properties, and for the browser compat data we've currently got available.

If there are older browser versions that supported `paintTiming` but not `presentationTiming`, then we probably ought to update the compat data to more accurately reflect this. I'm not sure what the data should be, but flagging this anyway for some future work, perhaps.

files/en-us/web/api/performance_api/long_animation_frame_timing/index.md
chrisdavidmills: Ah, good point. I've moved it to "End time" and pointed out that it could be `paintTime` or `presentationTime`.

files/en-us/web/api/performanceelementtiming/index.md
chrisdavidmills: updated

files/en-us/web/api/largestcontentfulpaint/index.md
mmocny: I [updated the spec wording](https://www.w3.org/TR/largest-contentful-paint/#sec-largest-contentful-paint-interface) to just:

> The [startTime](https://w3c.github.io/performance-timeline/#dom-performanceentry-starttime) attribute’s getter must return the value of [this](https://webidl.spec.whatwg.org/#this)’s [renderTime](https://www.w3.org/TR/largest-contentful-paint/#dom-largestcontentfulpaint-rendertime).

files/en-us/web/api/largestcontentfulpaint/index.md
mmocny: Perfect, ty!

Also, I found a slightly better way to do it:

```js
if (lastEntry.presentationTime) {
  console.log("LCP.paintTime:", lastEntry.paintTime, ".presentationTime:", lastEntry.presentationTime);
} else if (lastEntry.paintTime) {
  console.log("LCP.paintTime:", lastEntry.paintTime);
} else if (lastEntry.renderTime) {
  console.log("LCP.renderTime:", lastEntry.renderTime);
} else {
  console.log("LCP.loadTime:", lastEntry.loadTime);
}
```

But don't sweat it.

files/en-us/web/api/largestcontentfulpaint/index.md
chrisdavidmills: Cool, I didn't realize it would get done so quickly ;-)

I've updated the wording on the page to:

> Returns the value of this entry's {{domxref("LargestContentfulPaint.renderTime", "renderTime")}}.

files/en-us/web/api/largestcontentfulpaint/index.md
chrisdavidmills: That makes sense. I've updated all three examples to use this pattern, or as close to as possible.

chrisdavidmills:
> (still cannot formally approve, but this LGTM at this point-- ty so much!)

No worries — this is good enough for me. I'll move this along to the editorial review stage. 

Thanks for taking the time to review this.

files/en-us/web/api/largestcontentfulpaint/presentationtime/index.md
```diff
@@ -0,0 +1,33 @@
...
+
+{{APIRef("Performance API")}}
+
+The **`presentationTime`** read-only property of the {{domxref("LargestContentfulPaint")}} interface returns the {{domxref("DOMHighResTimeStamp","timestamp")}} when the painted pixels were actually drawn on the screen.
+
+The `presentationTime` is implementation-dependent, therefore the value may differ across implementing browsers. Some browsers may choose to always return `0` or not to expose the value.
+
+## Value
+
+A {{domxref("DOMHighResTimeStamp")}} or {{jsxref("Operators/null", "null")}} if the value is not exposed.
```
hamishwillee, line 17: FYI, last commit fixed up 5 "null" links that were redirecting.

files/en-us/web/api/performancelonganimationframetiming/presentationtime/index.md
```diff
@@ -8,13 +8,13 @@ browser-compat: api.PerformanceLongAnimationFrameTiming.presentationTime
 
 {{APIRef("Performance API")}}
 
-The **`presentationTime`** read-only property of the {{domxref("PerformanceLongAnimationFrameTiming")}} interface returns the {{domxref("DOMHighResTimeStamp","timestamp")}} when the element was actually drawn on the screen.
+The **`presentationTime`** read-only property of the {{domxref("PerformanceLongAnimationFrameTiming")}} interface returns the {{domxref("DOMHighResTimeStamp","timestamp")}} when the UI update was actually drawn on the screen.
```
hamishwillee, line 11: FYI element change to UI update. This was a claude catch. This matches the parent doc.

files/en-us/web/api/largestcontentfulpaint/painttime/index.md
```diff
@@ -0,0 +1,33 @@
...
+title: "LargestContentfulPaint: paintTime property"
+short-title: paintTime
+slug: Web/API/LargestContentfulPaint/paintTime
+page-type: web-api-instance-property
+browser-compat: api.LargestContentfulPaint.paintTime
+---
+
+{{APIRef("Performance API")}}
+
+The **`paintTime`** read-only property of the {{domxref("LargestContentfulPaint")}} interface returns the {{domxref("DOMHighResTimeStamp","timestamp")}} when the rendering phase ended and the paint phase started.
```
hamishwillee, line 11: I though rendering and painting were the same thing :-). 

Is there a link we can add for each of these phases - and all the other cases?

files/en-us/web/api/largestcontentfulpaint/presentationtime/index.md
```diff
@@ -0,0 +1,33 @@
...
+slug: Web/API/LargestContentfulPaint/presentationTime
+page-type: web-api-instance-property
+browser-compat: api.LargestContentfulPaint.presentationTime
+---
+
+{{APIRef("Performance API")}}
+
+The **`presentationTime`** read-only property of the {{domxref("LargestContentfulPaint")}} interface returns the {{domxref("DOMHighResTimeStamp","timestamp")}} when the painted pixels were actually drawn on the screen.
+
+The `presentationTime` is implementation-dependent, therefore the value may differ across implementing browsers. Some browsers may choose to always return `0` or not to expose the value.
```
hamishwillee: By " or not to expose" do you mean "not implement"? SHouldn't this be marked as optional then?

files/en-us/web/api/largestcontentfulpaint/index.md
```diff
@@ -32,38 +32,40 @@ Additional key paint moments are provided by the {{domxref("PerformancePaintTimi
...
+It also extends the following {{domxref("PerformanceEntry")}} properties, qualifying and constraining them as described:
+
+- {{domxref("PerformanceEntry.entryType")}} {{ReadOnlyInline}} {{Experimental_Inline}}
+  - : Returns `"largest-contentful-paint"`.
+- {{domxref("PerformanceEntry.name")}} {{ReadOnlyInline}} {{Experimental_Inline}}
+  - : Always returns an empty string.
+- {{domxref("PerformanceEntry.startTime")}} {{ReadOnlyInline}} {{Experimental_Inline}}
+  - : Returns the value of this entry's {{domxref("LargestContentfulPaint.renderTime", "renderTime")}}.
+- {{domxref("PerformanceEntry.duration")}} {{ReadOnlyInline}} {{Experimental_Inline}}
+  - : Returns `0`, as `duration` is not applicable to this interface.
```
hamishwillee, line 67: Note to self - these are already up to date.

files/en-us/web/api/performanceelementtiming/presentationtime/index.md
```diff
@@ -0,0 +1,33 @@
...
+slug: Web/API/PerformanceElementTiming/presentationTime
+page-type: web-api-instance-property
+browser-compat: api.PerformanceElementTiming.presentationTime
+---
+
+{{APIRef("Performance API")}}
+
+The **`presentationTime`** read-only property of the {{domxref("PerformanceElementTiming")}} interface returns the {{domxref("DOMHighResTimeStamp","timestamp")}} when the element was actually drawn on the screen.
+
+The `presentationTime` is implementation-dependent, therefore the value may differ across implementing browsers. Some browsers may choose to always return `0` or not to expose the value.
```
hamishwillee: Just a thought. As before, "not expose the value is fine", but I wonder if we need to otherwise show that this is optional.

files/en-us/web/api/largestcontentfulpaint/presentationtime/index.md
chrisdavidmills: Lovely, thanks!

files/en-us/web/api/largestcontentfulpaint/presentationtime/index.md
chrisdavidmills: I thought about this a bit, and updated the sentence to:

> The `presentationTime` is optional — some browsers may choose always to return `0` or not to expose the value at all. The value is also implementation-dependent — it may differ across browsers that choose to expose it.

files/en-us/web/api/performanceelementtiming/presentationtime/index.md
chrisdavidmills: See above comment

files/en-us/web/api/largestcontentfulpaint/painttime/index.md
chrisdavidmills: The rendering phase is when the browser determines the layout and positioning — where everything should be placed on the screen, how big it should be, what colors to use, etc.

The paint phase is when it is all drawn on the screen. The paintTime indicates the start of this phase, and the presentationTime indicates the end of this phase - when the pixels actually appear on the screen. I'm not convinced we need links for this.

hamishwillee: OK. I slightly disagree, because I didn't know this (at least for sure). I'll probably fix this to my taste if I ever touch the docs.

PR #43496, opened 2026-03-19
clarify visual authoring order vs mathematical application for transform property

### LINKED ISSUE #43180: Contradictory information about composite transforms in `transform` docs
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/transform

[Values](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/transform#values) and [Transform order](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/transform#transform_order)

In the docs for `transform`, under [Values](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/transform#values), it says that composite transforms are effectively applied in order from right to left. I interpret this to mean that `transform: translateX(200px) rotate(135deg);` effectively rotates first by 135deg *then* translates along the X axis by 200px.

If this is the case, the only way this could happen is if the translation axes **do not** rotate with the element and the transform origin **does not** translate with the element during these transforms.

If they do, that means composite transforms are read and applied left to right (for the above, translating *then* rotating).

See this [codepen example] for an animated demonstration of what I mean (forgive my laziness in not animating the rotation for result 2 right-to-left along a curved path - hopefully it's clear enough what I'm getting at).

The contradiction comes in the [Transform order](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/transform#transform_order) section. It says 
> When an element is rotated before being translated, the translate direction is on the rotated axis. The axis as indicated with the dotted lines.

"Translation direction is on the rotated axis" should only be true if composite transforms are applied left to right, but as above, the Values section states they are applied right to left, meaning the axis should not rotate.

That part also implies it's talking about `.two` which is `transform: rotate(135deg) translateX(200px);` but the text says "when an element is rotated *before* being translated". This is either suddenly talking about the written order (which is confusing when inconsistent with the Values section talking about applied order), or composite transforms are actually applied left to right and the Values section needs to be amended.

---

I'm happy to PR fixed wording for this page but I need to clarify exactly what the behaviour is. Unfortunately, I haven't found a way to view the actual axes/transform origin during the transforms in devtools, so I cannot directly confirm the behaviour myself.

Either the Values section says composite transforms are applied left to right

OR

The Transform order section is amended according to right-to-left composite transform behaviour, and is clear and consistent about what order it's referring to (ideally applied order, not mixing with written order).

_No response_

_No response_

low-perry:
### Description

This PR clarifies the relationship between the written order of transform functions (left-to-right) and their mathematical application (right-to-left). It emphasizes that each function modifies the local coordinate system for subsequent functions, which is why the visual result follows the authoring order

### Motivation

The current wording in the Values section ("effectively applied in order from right to left") creates a mental mismatch for developers. While mathematically true for point-mapping , it contradicts the visual intuition described in the Transform order section, where a rotation "tilts" the axes for a following translation. These changes bridge that gap by explaining that functions establish cumulative coordinate spaces.

### Additional details

This clarification is based on the CSS Transforms Module Level 1 and 2 specifications:
- [Matrix Composition (L-to-R)](https://www.w3.org/TR/css-transforms-1/#transform-rendering:~:text=Multiply%20by%20each%20of%20the%20transform%20functions%20in%20transform%20property%20from%20left%20to%20right): The transformation matrix is computed by multiplying functions from left to right.
- [Coordinate Systems](https://www.w3.org/TR/css-transforms-1/#transform-rendering:~:text=Specifying%20a%20value%20other%20than%20none%20for%20the%20transform%20property%20establishes%20a%20new%20local%20coordinate%20system%20at%20the%20element%20that%20it%20is%20applied%20to.%20The%20mapping%20from%20where%20the%20element%20would%20have%20rendered%20into%20that%20local%20coordinate%20system%20is%20given%20by%20the%20element%E2%80%99s%20transformation%20matrix.): Each function establishes a new local coordinate system for the element.
- [Mathematical Application (R-to-L)](https://www.w3.org/TR/css-transforms-1/#intro:~:text=Each%20%3Ctransform%2Dfunction%3E%20can%20get%20represented%20by%20a%20corresponding%204x4%20matrix.%20To%20map%20a%20point%20from%20the%20coordinate%20space%20of%20the%20div%20box%20to%20the%20coordinate%20space%20of%20the%20parent%20element%2C%20these%20transforms%20get%20multiplied%20in%20the%20reverse%20order%3A): Mapping a point from the element to the parent effectively happens in reverse order.

Fixes #43180


mao-sz (commented):
If the practical reality (as far as a developer would be concerned in order to know how to read composite transforms) is left-to-right, then [animation-composition](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/animation-composition#formal_syntax:~:text=behavior%2E-,With,30px,-%2E) will require amending accordingly.

I do not know if similar right-to-left descriptions exist in other pages.

low-perry:
@mao-sz I think you are 100% correct, but I'd love to keep this PR focused on the `transform` property as our 'source of truth' first, though.
Once we land this, we’ll have a clear standard to align the other pages as a follow-up

files/en-us/web/css/reference/properties/transform/index.md
```diff
@@ -101,7 +101,7 @@ The `transform` property may be specified as either the keyword value `none` or
 
 - {{cssxref("&lt;transform-function&gt;")}}
   - : One or more of the [CSS transform functions](/en-US/docs/Web/CSS/Reference/Values/transform-function) to be applied.
-    The transform functions are multiplied in order from left to right, meaning that composite transforms are effectively [applied in order from right to left](#transform_order).
+    The transform functions are multiplied in order from left to right. Each function establishes a new coordinate system for the ones following it, so the visual result follows the written order. Mathematically, this is equivalent to applying them in reverse order (right to left).
```
dipikabh: - suggesting replacing "multiplied" to make it clearer
- using "coordinate space" for consistency with the term used in L10
- rephrased 2nd sentence to draw clearer distinction between the two ways to look at the result

```suggestion
    The transform functions are combined from left to right – each function establishes a new coordinate space for the next function and so on – so the visual result matches the written order of the functions. Alternatively, keeping the parent coordinate space fixed, the same transformation can be described as applying the functions in reverse order (right to left).
```

files/en-us/web/css/reference/properties/transform/index.md
```diff
@@ -152,7 +152,9 @@ div {
 
 ### Transform order
```
dipikabh: let's make this title more explicit


```suggestion
### Comparing the order of transform functions
```

files/en-us/web/css/reference/properties/transform/index.md
```diff
@@ -152,7 +152,9 @@ div {
...
 
-The order of transform functions matters. In this example, two boxes are rotated and translated by the same values; only the transform function order is different.
+The order of transform functions matters. Visually, transforms follow the written order (left to right) because each function establishes a new coordinate system for the functions that follow it.
+
+In this example, two boxes are rotated and translated by the same values; only the transform function order is different.
```
dipikabh: we can skip repeating this part and instead, add explanation to the "Result" instead.

```suggestion
The order of transform functions matters.

In this example, two boxes are rotated and translated by the same values but the functions are in the opposite order. The dotted lines mark the X axis before and after rotation.
```

files/en-us/web/css/reference/properties/transform/index.md
dipikabh: Let's expand on this a bit more. The second sentence can be moved alongside example intro.

```suggestion
- Box 1 (first `translateX()`, then `rotate()`): The coordinate space first shifts `200px` along the X axis, then rotates `135deg` within that shifted space, so the element ends up to the right of its original position, rotated.
- Box 2 (first `rotate()`, then `translateX()`): The coordinate space first rotates `135deg`, so the element then moves `200px` along the rotated axis, in the direction shown by the dotted lines.
  
Alternatively, when mapping a point to the parent coordinate space, the functions apply in the reverse order (right to left):
- Box 1: The element first rotates `135deg`, then moves `200px` along the original X axis.
- Box 2: The element first moves `200px` along the original X axis, then rotates `135deg`.
```

dipikabh (changes requested):
Thanks @low-perry for sending a PR to clarify the language to highlight the two perspectives.

I've made a few suggestions that I hope add more clarity. Let me know what you think.

files/en-us/web/css/reference/properties/transform/index.md
mao-sz: Do we need this page to expand on *both* perspectives? i.e. is there high practical value in showing people how it would work if they wanted to read it RtL?

From my perspective, that feels confusing and I'm wondering why someone would want or need to read composite transforms RtL since that requires a different approach to coordinate space, even if it's technically correct.

To me, it feels more practical to only talk and demonstrate how the authored order (LtR) works, then there's consistency that's fully practical. They can see that each transform establishes a new coordinate space (i.e. rotating also rotates the translation axes/translating also translates the transform origin) and it's clear from there. I feel like introducing the idea of mapping a point to the parent coordinate space is out of this doc's scope and more of a "technically this is what happens under the hood mathematically" (which is suitable for the official spec of course). And I think it may have been that part of the spec that led to the current confusing wording in the first place.

dipikabh: You have a fair point, @mao-sz, but I see no harm in providing an alternate mental mapping for readers.

With the proposed changes, the original confusion that came from mixing both framings without explaining how they are related will be resolved.

Also note that there are other pages like [`transform-origin`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/transform-origin) and [`transform-function`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/transform-function) that use the R-to-L language on MDN. So it would be good to have that explanation (with an example) as well.

mao-sz: That's fair, thanks for pointing that out :+1:

files/en-us/web/css/reference/properties/transform/index.md
low-perry: @dipikabh This sounds much better. Thank you

files/en-us/web/css/reference/properties/transform/index.md
dipikabh: Hi @low-perry, thanks a lot for applying the other changes!

Are you intentionally choosing to not use the "Alternatively,..." explanation text for the example

dipikabh (approved):
Thanks for the updates, @low-perry!

PR #43500, opened 2026-03-20
Update compareDocumentPosition index.md

holloway:
### Description

The `compareDocumentPosition` API can also return `0` if the position is identical, ie comparing a node to itself.

### DIFF
--- a/files/en-us/web/api/node/comparedocumentposition/index.md
+++ b/files/en-us/web/api/node/comparedocumentposition/index.md
@@ -8,8 +8,7 @@ browser-compat: api.Node.compareDocumentPosition
 
 {{APIRef("DOM")}}
 
-The **`compareDocumentPosition()`** method of the {{domxref("Node")}} interface
-reports the position of its argument node relative to the node on which it is called.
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/node/comparedocumentposition/index.md
```diff
@@ -26,7 +26,7 @@ compareDocumentPosition(otherNode)
 
 An integer value representing `otherNode`'s position relative to `node`
 as a [bitmask](<https://en.wikipedia.org/wiki/Mask_(computing)>) combining the
-following constant properties of {{domxref("Node")}}:
+following constant properties of {{domxref("Node")}} or `0` if the position is the same:
```
hamishwillee: Thanks very much. This is true, but it's confusing way to say that the nodes are the same.

```suggestion
following constant properties of {{domxref("Node")}} or `0` if `otherNode` is the same as this node:
```

hamishwillee (approved):
Thanks @holloway - good catch - and welcome to MDN!

PR #43502, opened 2026-03-20
FF149 @container <container-query> optional

hamishwillee:
FF149 adds support for [`@container`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@container) selections where the [`<container-query>`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@container#container-query) part is optional in https://bugzilla.mozilla.org/show_bug.cgi?id=2016474

What this means is that you can just declare that a container maps to a particular named element.

This change updates the @container docs to make it clear that either or both, but not neither, of the conditions must be present. 

Related docs work can be tracked in https://github.com/mdn/content/issues/43212


pepelsbey (approved):
Looks good! Thank you 🙂

PR #43505, opened 2026-03-20
Add type: "text" import attribute

eemeli:
### Description

Adds documentation for the upcoming `type: 'text'` import attribute.

### Motivation

I'm the proposal champion, and would like to advance it.

### Additional details

The proposal is specified in:
- [tc39/proposal-import-text](https://github.com/tc39/proposal-import-text)
- whatwg/html#11933
- whatwg/fetch#1898
- w3c/webappsec-csp#794

Related MDN work tracked in 
- #44166


files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -127,6 +127,19 @@ document.adoptedStyleSheets.push(exampleStyles);
 
 Note that importing CSS modules into workers is usually not supported, because the CSSOM specification only exposes `CSSStyleSheet` in the window context.
 
+#### Text Modules (`{ type: "text" }`)
+
+The `text` type allows importing a file as a UTF-8 string value.
```
Josh-Cena:
```suggestion
The `text` type allows importing a module's source as a string value.
```

There's no such thing as "UTF-8 string" so even if practically it makes sense we should avoid using such terminology. This would be controlled by the underlying system and the JavaScript environment only ever sees a "string". (I do believe you can only transmit UTF-8 modules on the web, but I'm less certain about non-browser environments. What about importing a GB18030 file in Node.js?)

files/en-us/web/javascript/reference/statements/import/with/index.md
```diff
@@ -127,6 +127,19 @@ document.adoptedStyleSheets.push(exampleStyles);
...
+You can load text from a file into the `text` string using the following code:
+
+```js
+import text from "https://example.com/file.txt" with { type: "text" };
+```
+
+The file will be requested with an `{{HTTPHeader("Accept")}}: text/plain` header,
+but the value of the response's `{{HTTPHeader("Content-Type")}}` header is ignored,
+and all files are parsed as UTF-8.
```
Josh-Cena:
```suggestion
The file will be requested with an `{{HTTPHeader("Accept")}}: text/plain` header, but the value of the response's `{{HTTPHeader("Content-Type")}}` header is ignored, and all files are parsed as UTF-8. It can contain any textual data, even JavaScript code (which is treated as plain text).
```

files/en-us/web/javascript/reference/statements/import/with/index.md
eemeli: It would be good to mention the parsing-as-UTF-8 in some way, as that's what browsers will do. I've raised WinterTC55/proposal-minimum-common-api#117 to clarify the situation for non-browser runtimes.

eemeli: Ah! I forgot that's already mentioned below.

Josh-Cena (approved):
LGTM

hamishwillee (changes requested):
@eemeli Is this supported by any browser? I was expecting to see a text key in the browser compatibility section of each of the places you added the addition - such as https://pr43505.review.mdn.allizom.net/en-US/docs/Web/JavaScript/Reference/Statements/import/with#browser_compatibility

This is important because MDN only documents things that are available in browsers (though not necessarily released). 

I'm (very) happy with the docs updatd. I've added request changes though, because we shouldn't merge this until the corresponding browser compatibility data has merged. Can you link that in the description?

hamishwillee:
@eemeli As per my last comment, is it in any browser? If so the corresponding browser compatibility data should be updated and we can look at this/gt it in.

eemeli:
Ah, sorry, lost track of this earlier. I've been working on a [Firefox implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=2024854), which I'm hoping to land into Nightly next week. So no, this feature is not _yet_ available in any browser.

hamishwillee:
Thanks @eemeli ! I've converted to draft so it falls off my review queue. Can you please reenable once the feature is in nightly. It is OK if you forget because likely I'll end up with the MDN release tasks when we get to FF152 anyway.

Josh-Cena:
This can now move forward: https://github.com/mdn/content/issues/44166

eemeli:
I merged from `main` to resolve conflicts, and intend in [bug 2039881](https://bugzilla.mozilla.org/show_bug.cgi?id=2039881) to enable this by default in Firefox 153.

It's currently available in Nightly behind the `javascript.options.experimental.import_text` pref.

hamishwillee (approved):
Thanks @eemeli - this is excellent.

PR #43486, opened 2026-03-19
Testing backticks in front matter title on two HTML pages

dipikabh:
### Description

For standardizing our titles on reference pages, backtick support has been added in mdn/rari#552, which is now available in mdn/content.

In this PR, I'm trying out the backticks on two reference pages in HTML.

`rari` is correctly outputting `titleHTML` in the JSON:
- `"titleHTML"`: `"<code>is</code> HTML global attribute"`
- `"titleHTML"`: `'<code>&lt;input type="number"&gt;</code> HTML attribute value'`

### Additional details

`fred` still needs to be updated to read `titleHTML` from rari's output and render it in page `<h1>`.
Also, after `fred` starts rendering `<code>` tags in the title, they'd need to be styled appropriately.

Meanwhile, title updates need not be blocked by `fred` work and can proceed in parallel.

### Related issues and pull requests

- https://github.com/mdn/content/issues/39894
- https://github.com/mdn/mdn/issues/811

PR #43508, opened 2026-03-20
Remove backticks from the title

pepelsbey:
### Description

They are not currently rendered anyway, but they soon will be, but the logic is different, see:

- https://github.com/mdn/content/pull/43486
- https://github.com/mdn/fred/pull/1398


Josh-Cena:
But isn't it a good idea to render them anyway? This is the verbatim error message of Firefox. I don't like us arbitrarily changing things that have a character-to-character reference.

pepelsbey:
> I don't like us arbitrarily changing things that have a character-to-character reference.

@Josh-Cena Maybe it wasn't a good idea to put them in the title in the first place. But in any case, it's a temporary fix to avoid code rendering in this particular case while we're figuring out styling for code in heading. And it was the only title getting the way of that.

Josh-Cena:
If it's only temporary and we plan to eventually revert it, then it's alright :)

PR #43510, opened 2026-03-20
ci(pr-check-lint_content): run reviewdog in separate workflow

caugner:
### Description

- Extracts the reviewdog portion of the `pr-check-lint_content` workflow into a separate `pr-reviewdog` workflow (privileged, with status reporting), by passing diff and Markdownlint log as an artifact.
- Makes the `pr-check-lint_content` workflow unprivileged, and simplifies it by removing now-obsolete safeguards.

### Motivation

- Eliminate risk of code injection.
- Make it easier to simplify the lint workflow in the future.

### Additional details

Tested on my personal fork:

- Pushed [this branch/commit to caugner:main](https://github.com/caugner/content/commit/3bf3046e761f767d43caf2e26bf21188f1b15cab).
- Opened [a PR](https://github.com/caugner/content/pull/1) with changes causing lint failures.
- The [Lint content](https://github.com/caugner/content/actions/runs/23639785325/job/68857350929?pr=1) workflow ran, and uploaded both the diff and the Markdownlint log as an artifact.
- The [PR Reviewdog](https://github.com/caugner/content/actions/runs/23639804817) workflow picked it up.

Note: Reviewdog refuses to use `github-pr-review` reporter when run in a non-PR workflow (`workflow_run`), so we need to pretend we're not in GitHub CI by hiding `GITHUB_ACTIONS` from it.

Same as:

- https://github.com/mdn/translated-content/pull/34668
- https://github.com/mdn/translated-content-de/pull/237

### DIFF
--- a/.github/workflows/pr-check-lint_content.yml
+++ b/.github/workflows/pr-check-lint_content.yml
@@ -1,4 +1,4 @@
-name: Lint and review content files
+name: Lint content
 
 on:
   pull_request:
@@ -16,7 +16,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint-and-review-docs:
+  lint:
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/pr-reviewdog.yml
+++ b/.github/workflows/pr-reviewdog.yml
@@ -3,7 +3,7 @@ name: PR Reviewdog
 on:
   workflow_run:
     workflows:
-      - "Lint and review content files"
+      - "Lint content"
     types:
       - completed
 


.github/workflows/pr-reviewdog.yml
```diff
@@ -0,0 +1,67 @@
...
+        if: steps.check.outputs.HAS_ARTIFACT == 'true'
+        env:
+          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          reviewdog \
+            -name="mdn-linter" \
+            -f=diff \
+            -f.diff.strip=1 \
+            -filter-mode=diff_context \
+            -reporter=github-pr-review < lint-results/lint.diff
```
caugner: Need to tell reviewdog somehow what PR this is about, since this is not obvious from the context.

caugner:
Converting to draft, as I should test this in a fork before.

**Edit**: Done.

.github/workflows/pr-reviewdog.yml
caugner: Fixed via 03435792c83.

.github/workflows/pr-reviewdog.yml
```diff
@@ -0,0 +1,68 @@
...
+        env:
+          CI_PULL_REQUEST: ${{ steps.identify-pr.outputs.number }}
+          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          reviewdog \
+            -name="mdn-linter" \
+            -f=diff \
+            -f.diff.strip=1 \
+            -filter-mode=diff_context \
+            -reporter=github-pr-review < lint-results/lint.diff
```
caugner: Unfortunately, reviewdog fails when the run was not triggered by the PR directly:

```
reviewdog: this is not PullRequest build.
```

See: https://github.com/caugner/content/actions/runs/23594470828/job/68707566609#step:6:13

caugner: I was able to resolve the issues, and made it run successfully on https://github.com/caugner/content/pull/1.

.github/workflows/pr-check-lint_content.yml
```diff
@@ -215,40 +201,18 @@ jobs:
...
-            -reporter="github-pr-review"
+          mkdir -p lint-results
+          git diff HEAD > lint-results/lint.diff
+
+      - name: Upload lint results
+        if: always() && steps.lint-diff.outputs.FILES_MODIFIED == 'true'
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        with:
+          name: lint-results
+          path: lint-results/
```
LeoMcA: Might it be a little safer to just upload the `lint.diff` file, rather than the whole folder (which a malicious user may add content to in their PR), with the new feature available in upload-artifact v7: https://github.com/actions/upload-artifact/releases/tag/v7.0.0

.github/workflows/pr-reviewdog.yml
```diff
@@ -0,0 +1,67 @@
...
+      - completed
+
+permissions:
+  # Download artifact from lint workflow.
+  actions: read
+  # Post inline review comments via reviewdog.
+  pull-requests: write
+
+concurrency:
+  group: ci-${{ github.workflow }}-${{ github.event.workflow_run.id }}
```
LeoMcA: Will this always be unique? Is it therefore not needed?

.github/workflows/pr-reviewdog.yml
```diff
@@ -0,0 +1,67 @@
...
+        id: identify-pr
+        env:
+          BASE_REPO: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ github.token }}
+          HEAD_REPO: ${{ github.event.workflow_run.head_repository.full_name }}
+          HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+        run: |
+          PR_NUMBER=$(gh api "repos/$HEAD_REPO/commits/$HEAD_SHA/pulls" \
+            --jq ".[] | select(.base.repo.full_name == \"$BASE_REPO\") | .number")
+          echo "number=$PR_NUMBER" >> $GITHUB_OUTPUT
```
LeoMcA, line 46: Thought for later: I remember us using a similar step to this across a bunch of workflows, short of publishing our own actions, could we have a library of "actions" we can import across a range of workflows (as I'm sure there's other similar steps we need across workflows).

.github/workflows/pr-check-lint_content.yml
```diff
@@ -215,40 +201,18 @@ jobs:
...
-          git diff >"${TMPFILE}"
-          git stash -u && git stash drop
-          reviewdog \
-            -name="mdn-linter" \
-            -f=diff \
-            -f.diff.strip=1 \
-            -filter-mode=diff_context \
-            -reporter=github-pr-review < "${TMPFILE}"
-
-      - name: Add reviews for markdownlint errors
```
LeoMcA, line 239: Do we lose this step in the new workflow?

caugner: Good catch, I don't think it had any effect before, so I removed it.

.github/workflows/pr-check-lint_content.yml
caugner: Good idea, done via f7e182f4245.

.github/workflows/pr-reviewdog.yml
caugner: Good point, removed via 87dc13f5042.

.github/workflows/pr-reviewdog.yml
caugner: The problem with actions (to be used as a step) is that afaik you need one repository per action, hence why https://github.com/Mozilla-Actions/ exists.

.github/workflows/pr-check-lint_content.yml
caugner: Oh, I might have missed something: The "Add reviews for markdownlint errors" step pipes the output of running `npx markdownlint-cli2 --fix "${files_to_lint[@]}" 2>&1` into reviewdog, and I thought this doesn't make sense.

But it looks like `-efm="%f:%l:%c %m" -efm="%f:%l %m"` allows to make non-autofixable markdownlint errors appear in the review.

caugner: Restored via 0b631022c3ee22d7c9705fdc0b6e8ea2761b9619.

.github/workflows/pr-reviewdog.yml
fkiriakos07: not sure if that's required, you can host multiple reusable actions in the same repo, MozCloud have reusable actions in this repo for example: https://github.com/mozilla-it/deploy-actions

caugner: IIUC composite actions run as jobs, not as steps. This could run as a separate job, but it would add overhead and slow things down.

caugner: Hm, looks like that repo indeed has actions that can run as a step. All actions in one repo would share the same tag, but that might be fine for our own actions.

PR #43521, opened 2026-03-22
Update dispatchEvent documentation for clarity

### LINKED ISSUE #43519: Misleading sentence in dispatchEvent() page
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent

Unlike "native" events, which are fired by the browser and invoke event handlers asynchronously via the [event loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Execution_model),

I think this sentence:

> Unlike "native" events, which are fired by the browser and invoke event handlers asynchronously via the [event loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Execution_model)

Is misleading.

AFAIK, Native events puts "dispatchEvent" calls on the event loop. Once those are called, the rest is still synchronous, meaning: the event handling itself run synchronously, including checking for registered event handlers and running them - synchronously.  

The way that sentence is phrased, could be interpreted as though the event handler functions themselves run "asynchronously" - or are put on the Event Loop - but they are registered/saved on the element implementing EventTarget - and run "from there".

Maybe:
Unlike "native" events, which are fired by the browser and invoke dispatchEvent asynchronously via the [event loop]

_No response_

_No response_

yuval-a:
### Description
Clarified the difference between native events and manually dispatched events in the dispatchEvent documentation.

### Motivation
The previous phrasing implied that the actual event handler functions themselves are scheduled asynchronously on the Event Loop, but only the dispatch is.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent

Fixes https://github.com/mdn/content/issues/43519


dipikabh:
Thanks for the PR, @yuval-a!

Could you add to your PR description the line "Fixes #43519"?
That way they'll be linked and the issue will close when this PR is merged.

files/en-us/web/api/eventtarget/dispatchevent/index.md
```diff
@@ -19,8 +19,7 @@ should have already been created and initialized using an {{domxref("Event/Event
 > [!NOTE]
 > When calling this method, the {{domxref("Event.target")}} property is initialized to the current `EventTarget`.
 
-Unlike "native" events, which are fired by the browser and invoke event handlers
-asynchronously via the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model),
+Unlike calling `dispatchEvent()` manually - "native" events, which are fired by the browser, schedule an asynchronous dispatch via the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model). This dispatch uses the same process as `dispatchEvent()`.
```
dipikabh: Your suggestion in the issue you raised was on the right track. The change here is complicating the sentence structure.

Try this with the original phrasing: "fired by the browser and dispatched asynchronously via the event loop"

dipikabh (changes requested):

yuval-a (commented):
@dipikabh how's this?

files/en-us/web/api/eventtarget/dispatchevent/index.md
```diff
@@ -19,10 +19,7 @@ should have already been created and initialized using an {{domxref("Event/Event
 > [!NOTE]
 > When calling this method, the {{domxref("Event.target")}} property is initialized to the current `EventTarget`.
 
-Unlike "native" events, which are fired by the browser and invoke event handlers
-asynchronously via the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model),
-`dispatchEvent()` invokes event handlers _synchronously_. All applicable event
-handlers are called and return before `dispatchEvent()` returns.
+Unlike calling `dispatchEvent()` manually, "native" events are fired by the browser and dispatched asynchronously via the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model). The dispatch process itself is similar in both cases and invokes event handlers _synchronously_. All applicable event handlers are called and return before `dispatchEvent()` returns.
```
dipikabh: This rewrite ends up comparing two different things - "calling" and "events".
I suggest we either keep the original phrasing ("Unlike "native" events...")
OR
break it down into two sentences and also change the voice to active to differentiate between browser and developer actions. ("when you call" is a subtle hint that you call this method manually)

What do you think about:

```suggestion
The browser queues "native" events on the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model) and fires them asynchronously. In contrast, when you call `dispatchEvent()`, it invokes event handlers synchronously. Note that `dispatchEvent()` returns only after all applicable event handlers have executed.
```

@Josh-Cena could you cross-check the technical accuracy of my suggestion here. Thanks!

dipikabh (changes requested):

files/en-us/web/api/eventtarget/dispatchevent/index.md
Josh-Cena: This looks right to me :) I would suggest:


```suggestion
The browser queues "native" events on the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model#job_queue_and_event_loop), so each event handler is executed asynchronously in a separate job. In contrast, when you call `dispatchEvent()`, it invokes event handlers synchronously and returns only after all applicable event handlers have executed.
```

yuval-a: @Josh-Cena - but that phrasing brings back the technical confusion/inaccuracy that this PR aims to correct from the first place.

Correct me if I'm wrong.

The difference to "native" - is that the call for the dispatch itself is async, but during it, event listeners run synchronously - don't go up the event loop.

The statement 
> "each event handler is executed asynchronously[...]"  

makes it sound as if each of them is scheduled asynchronously on the event loop.

Interestingly, I tried to physically verify this by running some code in the console:

```javascript
const btn = document.createElement('button'); 
btn.textContent = 'click me'; 
document.body.appendChild(btn); 
btn.addEventListener('click', () => { 
   console.log('listener 1 start'); 
   Promise.resolve().then(() => { 
      console.log('microtask from listener 1'); 
   }); 
   setTimeout(() => { console.log('timeout from listener 1'); }, 0); 
   console.log('listener 1 end'); 
}); 
btn.addEventListener('click', () => { console.log('listener 2'); }); 
```

The microtask log **does** run between listeners, the timeout log only after both run.
When running the same principle with a manual `dispatchEvent` and a custom event - the microtask log only shows **after** both listeners run. In native - there is a microtask checkpoint between listeners. But, still - the listeners themselves doesn't run async, no separate async queuing for each...
Or are they...? I mean, this test shows at least that no async timers run between (tried with other types of asyncs as well - same result) - my only doubt is that maybe they DO - and have a different "async priority" / are queued as a batch.
There is no place in the specs that explicitly says they run async in native, anyway.

What is the definite proof...?

EDIT: I'm learning to read the specs / distinguish between the terms, and it definitely says the handlers are "called" and not queued.

Josh-Cena: Wait, now I'm not sure what the issue is claiming.

For a "native" event:

1. The browser receives a mouse click; it enqueues all registered event handlers on the task queue.
2. At some indefinite time in the future, the event handler is successfully dequeued from the task queue and gets executed. It's asynchronous in the sense that it doesn't reuse the job of any other JavaScript execution—it's the entrypoint of a whole JavaScript job. It's synchronous in the sense that the browser doesn't "await" it, whatever that should mean when there's no relevant return value anyway.

Imagine this:

```js
btn.addEventListener("click", function onclick1() { foo(); });
btn.addEventListener("click", function onclick2() { bar(); });
```

The job queue:

```
job 1: whatever code is currently running
job 2: click event handler 1
  stack frame 1: onclick1()
  stack frame 2: foo()
job 3: click event handler 2
  stack frame 1: onclick2()
  stack frame 2: bar()
```

When job 2 executes, `onclick1` may schedule more jobs via `setTimeout`, `new Promise`, etc. Because `new Promise` makes a microtask with higher priority, when job 2 finishes, the next task the browser picks out will be the microtask instead of job 3, which has to wait a bit more. This is the speculation behavior you have observed.

For `dispatchEvent`:

1. The function gets called. All registered event handlers get called synchronously.
2. Each event handler executes in the current JavaScript job. It's definitely synchronous by every definition of synchronicity.

The job queue:

```
job 1: whatever code is currently running
  stack frame 1: dispatchEvent()
  stack frame 2: onclick1()
  stack frame 3: foo()
  ...
  stack frame 2: onclick2()
  stack frame 3: bar()
```

Nothing is allowed to execute between `onclick1()` and `onclick2()` because it's one synchronous block.

My acceptance of the original issue was that the word "asynchronous" is ambiguous in the context of native events because an event handler cannot observe its asynchronicity anyway, when _by definition_ it has no relevant caller or callee. The key point here is that the event handler executes in a separate JavaScript job.

Now I look back at your suggestion, I don't think it makes a lot of sense either:

> Unlike "native" events, which are fired by the browser and invoke dispatchEvent asynchronously via the [event loop]

You are implying the following:

```
job 1: whatever code is currently running
job 2: all click event handlers
  stack frame 1: dispatchEvent()
  stack frame 2: onclick1()
  stack frame 3: foo()
  ...
  stack frame 2: onclick2()
  stack frame 3: bar()
```

This would mean no other job is allowed to execute between `onclick1()` and `onclick2()`. This is false.

So I think there's something to be rewritten about this paragraph, but not your suggestion verbatim. Sorry I didn't read your suggestion in detail when I triaged the issue.

yuval-a: As far as I understand, the mental model (for native events) is more like:
job 1: start internal event dispatch process, which are these steps:
https://dom.spec.whatwg.org/#dispatching-events
these include, for each of the 3 phases: capturing, target, bubbling:
cloning the event handlers registered on each EventTarget object along the "event path" - and calling them one-by-one (like in a for-loop. Specs say: "For each"): 
https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke

the key point/question here, again, is if they are called "asynchronously" (as in what JS developers usually expect when they see this term) or not;
They are called in what the specs define as "callback invocation"
https://webidl.spec.whatwg.org/#js-invoking-callback-functions

which, as I'm researching thus far, is not the same as what variations of "running asynchronously" usually mean. It is a specific mechanism, that includes cleanup and microtasks checkpoint after each, but not a "macrotask" checkpoint.
i.e. Promises/Microtasks queued from a listener will run when it's done, but any other "async macrotask" will only run after ALL listeners finished running.

I've seen all sorts of discussions surrounding this, and I also saw this:
https://github.com/whatwg/dom/issues/1308
(A proposal for "asynchronous event listeners" - which I think strengthens my point that event listeners are not considered "asynchronous" in the common/usual way).

so, I think what's actually happening is something that's between the first model you describe in your comment and the second one (that you describe as what I was implying).

And in that model: queued microtasks will run in-between, queued macrotasks will not run in-between, only after.

The issue may be mostly "technical semantics"; 
I still think the current phrasing:

> Unlike "native" events, which are fired by the browser and invoke event handlers asynchronously via the [event loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Execution_model)

does not reflect what is technically happening; Event handlers do not go up the event loop the same way other "usual" async tasks usually do.

yuval-a: For further clarity, your (Josh) statement on your comment:
> The browser receives a mouse click; it enqueues all registered event handlers on the task queue.

I'm sure that it's more like:
The browser receives a mouse click; it enqueues a "dispatch mouse click event" on the task queue.
If, during when this task runs - registered event handlers are enqueued on the task queue - that I'm not sure of. I was pretty sure they don't when I opened the PR, but doubts arose from the discussion. As per my last comment, I think they are called in a unique way - adding to what I wrote it seems each event handler in native do get its own stack frame (and in manual they run in a single frame).

I wish someone who touched the actual code implementation could clarify this - is there someone we can tag?

Josh-Cena: Not only does each event handler get called in its stack frame (it doesn't make sense to not to); it's also called in its own job.

> And in that model: queued microtasks will run in-between, queued macrotasks will not run in-between, only after.

That's because other queued macrotasks necessarily have lower priority than these alraedy-queued event handlers. I'm not sure what that's proving.

I'm saying that there's no single job called "dispatch all event listeners". Basically, here's `dispatchEvent()`:

```js
function dispatchEvent(e) {
  for (const onclick of this.clickHandlers) {
    onclick(e);
  }
}
```

Here's the code that's run whenever a native click happens (this code itself can be executed synchronously or asynchronously—doesn't matter because there's no one to observe its synchronicity):

```js
for (const onclick of element.clickHandlers) {
  queueTask(() => onclick(e));
}
```

yuval-a: When the specs mention something that is supposed to be queued on the event loop - they specifically use the term "Queue a task" - linking to the part detailing the steps for it:
https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task

For event handlers they say they are called by callback invocation, linking to this:
https://webidl.spec.whatwg.org/#invoke-a-callback-function
and not by queuing a task.

Josh-Cena: It's because in https://webidl.spec.whatwg.org/#call-a-user-objects-operation, when the callback finishes, the browser performs cleanup which includes clearing the microtask queue. This is analogous in every way (that I can think of) in userland to having a separate macrotask for every callback; it's just specified this way

yuval-a: do they go up on the event loop? If not then at least "via the event loop" should be omitted :)

Josh-Cena: Now that I read the spec, I actually don't think there's a big difference between `dispatchEvent` and the "fire an event" operation! So yes I think we should remove event loop and also try to rewrite it more significantly.

yuval-a: You're going through a similar rabbit hole I went through. I thought at first there is almost no difference.

But, did the experiment from a few comments back, comparing both:
two listeners, the first one schedules both a microtask and a macrotask (timeout), with logs.

In manual (calling `dispatchEvent`): two listeners finished, then microtask, them macrotask.
In native: one listener finished, microtask, listener 2 finished, macrotask.

So there was a significant difference :) dispatch seems to be 100% synchronous. Native... there's a microtask checkpoint between - so again I think the actual event handlers invocation is different in native - not synchronous, but not fully "async" as well. 
In manual they all run sequentially 100% - nothing in between goes through  until dispatch returns (it's a "chunk"), in native microtasks get through.

yuval-a: See this thread I raised on WHATWG chat room and the replies.
https://matrix.to/#/!AGetWbsMpFPdSgUrbs:matrix.org/$J5oBFMPGzUjAGpP1jhoTbulgJNktJ1RZKOpr_NMEi78?via=matrix.org&via=mozilla.org&via=igalia.com

My suggestion currently is to change the paragraph to something along these lines:

> Unlike calling `dispatchEvent()` manually, when "native" events are dispatched by the browser - applicable event handlers run via "Callback Invocation": each runs on its own stack frame, with a microtasks queue checkpoint after each handler finishes running. When calling `dispatchEvent()` - all applicable event handlers are called synchronously and return before `dispatchEvent()` returns.

yuval-a: I broadened my knowledge.
I also ran tests that show that rendering is "blocked" - deferred until all handlers run, even if making a visual change from one handler, and then blocking the thread from another. Even if making visual changes from within a queued microtask from within a handler.

As I now know about Blink as an outer host environment for Javascript execution threads - I'm pretty sure that the actual technical difference between native and manual is - manual triggers the dispatch "from within V8", native triggers it from "outside" - from Blink, schedules on the event loop, but it runs as a "single unit", not yielding back to the main event loop - or anyway it's "atomic". Microtasks run because there is also a microtasks checkpoint defined after each script execution (as well as at the end of an event loop iteration).

This is as I thought in the beginning, only with more actual implementation details.

I suggest this edit:

> Unlike calling `dispatchEvent()` manually, which triggers event handlers to run synchronously in the same stack frame - "native" events are dispatched by the browser, triggering each event handler to run on a separate stack execution frame; any queued microtasks will run after each event handler, but other asynchronous browser operations, such as rendering, will only run after all event handlers finished running.

Josh-Cena: This looks good to me

yuval-a: @dipikabh you're also okay with this?

dipikabh: Sorry @yuval-a, this fell off my radar.

Since @Josh-Cena has approved this latest version, I'm happy to go with it albeit with some tweaks and to break up the long sentence:

> Unlike calling `dispatchEvent()` manually, which triggers event handlers to run synchronously in the same stack frame, "native" events are dispatched by the browser, which triggers each event handler to run on a separate stack execution frame; any queued microtasks run after each event handler. However, other asynchronous browser operations, such as rendering, run only after all event handlers have finished running.

Kaiido: Not all native events are fired in their own task (and thus spawn a new stack per callback, which itself will enter the microtask checkpoint when cleaning). 
For instance the `<iframe>`'s load event will fire synchronously at document insertion if its `src` is  `about:blank`: https://jsfiddle.net/0cus135z/

yuval-a: @Kaiido Interesting.
BUT... your code calls `document.body.append(i);` - which is **conceptually** like calling `element.click()` (it's a different kind of "dispatch" - it spawns from JS - like an "artificial click" - meaning it's not a "native event" "conformally" - it doesn't come from "outside" (e.g. an actual mouse click detected by the OS );

Note that once I put the `iframe` in the html (without `src`) - and even set `src` to `about:blank` from JS - the expected behavior happens.
See variant of your jsfiddle: https://jsfiddle.net/eqt29b6x/2/

Or it could be that the `createElement` of the iframe from JS is what's making the difference - this can make sense because in the first variation JS is the source of creation of the iframe, and in the second variation it's the HTML parser - which is in the host/Blink layer - so this also might be the reason for the `load` behaving differently between variations - it's like a "native load".
Either way I don't think it shows an exception.

Also I saw this:
https://github.com/whatwg/html/issues/4965
But not sure is related or even relevant anymore as my first paragraph above suggests.

I'm going to try to commit the PR with dipikabh phrasing from yesterday - in about two more days (to allow some more time for any other "possible exceptions").

Kaiido: I'm not sure everyone will understand this term that way. For me at least, a "native event" is an event fired from the UA itself, and not one generated by the user code that would be dispatched by a direct call to `dispatchEvent` (or by a wrapper like `HTMLElement.click()`).   

Other such events I know of, that would fall into *my* definition of "synchronous native events": `<form>`'s *formdata* and `<input>` *invalid* events when the form is submitted programmatically, `AbortSignal`'s *abort* event (even when tied by `AbortSignal.any()`), `<dialog>` before toggle when triggered from `close()`, *popstate*  when setting `location.hash`. But there may be many more in APIs I don't know. Events have to be fired from a task only when "the algorithm that triggers the event could be running on a different thread or process" (https://w3ctag.github.io/design-principles/)

Also, while I believe this might be a bug, in Firefox, *beforeinput*  and *input*  events do fire synchronously, even though they'd fall in *your* definition.

yuval-a: I would define "native event" as "anything that does not originate from within the runtime of Javascript".
This also aligns with the quote you mention from the design specs - and also with your examples.

Meaning: "if you do it from JS - it's not a "native event"". I think this definition holds - also aligns with your examples - e.g. "submit with JS" vs. the browser's "native submit" (a button with type="submit" inside a form).

I would also argue that possibly any API implementation parts that ARE in JS actually call `dispatchEvent`.

"native events" are "system events" --- the obvious examples are things like mouse click - but are also true for something that happens in the "outer host browser level" - "browsery things".

The most important insight and anecdote I've learned from all of this is to remember that Javascript still keeps its original intention of acting as an "interface" to "outer" or "native" capabilities (it started as a "scripting language to interface with the browser" after all) - and those widened incredibly far over the years - but this truth still holds.

The mental model is seeing this single V8 thread as the "inside", and the host browser and beyond as the "outside". If it's triggered "programmatically" with JS === not "native".

yuval-a: Come to think of it - there is no "official definition" in MDN of what a "native event" is (at least, I asked chat GPT to check and it says there isn't), while "fired by the user agent" IS sometimes used in specs, so maybe we should use:

> Unlike calling `dispatchEvent()` manually, which triggers event handlers to run synchronously in the same stack frame, "native" events - originating from the browser environment, triggers each event handler to run on a separate stack execution frame; any queued microtasks run after each event handler. However, other asynchronous browser operations, such as rendering, run only after all event handlers have finished running.

files/en-us/web/api/eventtarget/dispatchevent/index.md
```diff
@@ -19,10 +19,7 @@ should have already been created and initialized using an {{domxref("Event/Event
 > [!NOTE]
 > When calling this method, the {{domxref("Event.target")}} property is initialized to the current `EventTarget`.
 
-Unlike "native" events, which are fired by the browser and invoke event handlers
-asynchronously via the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model),
-`dispatchEvent()` invokes event handlers _synchronously_. All applicable event
-handlers are called and return before `dispatchEvent()` returns.
+Unlike calling `dispatchEvent()` manually, which triggers event handlers to run synchronously in the same stack frame, "native" events (dispatched by the browser) triggers each event handler to run on a separate stack execution frame; any queued microtasks run after each event handler. However, other asynchronous browser operations, such as rendering, run only after all event handlers have finished running.
```
hamishwillee: @Josh-Cena @dipikabh Looking at this text ^^^ it seems pretty unhelpful to me - should I as a user of this method need to understand stack frames or microtask? No, no I should not - certainly not for this method where knowing this makes no difference.

This was also discussed in #43973, where we were converging towards something like this:

```suggestion
Unlike "native" events, which the browser fires by queuing a task on the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model), `dispatchEvent()` invokes all applicable event handlers _synchronously_ before returning.
```

It is possibly not quite correct, but it captures the main point of the original, which is that these events are synchronous, whereas events generated by the browser are not.

Anyway, my take on this is that something simple here is better or deleting the text entirely, as largely irrelevant to readers.

I'll leave to you, and close the other issue as it was "late to the party".

hamishwillee: ... or if you leave it to me, I'll just delete this section altogether :-)

Josh-Cena: The fundamental question is whether there's race—i.e., if it's possible for other jobs to execute in between two event handler executions.

yuval-a: Following @hamishwillee comment about not being too "low-level-ish" (though I argue about his remark regarding microtasks - modern JS developers are familiar with the concept, especially as `queueMicrotask` is an official API, MDN also has a guide on it (added a link to it in my suggestion), plus the event loop is mentioned - it's just the same amount of "low-level", if not more) I ) and in the spirit of this paragraph being about "native events VS. user events", I now suggest changing to adding this note (while removing that paragraph)

```suggestion
> [!NOTE]
> For events dispatched with this method, the [`isTrusted` property of the Event object]
> (/en-US/docs/Web/API/Event/isTrusted) will be `false`. This is also true for events
> triggered programmatically by other means (for example, by calling `element.click()`).
> By contrast, events dispatched by the browser will have `isTrusted` set to `true`.
>
> Additionally, [microtasks](/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide) scheduled
> from event handlers of an event dispatched programmatically will run only after all
> handlers have finished executing. For browser-dispatched events, microtasks may run
> between individual event listener invocations during the dispatch process.
>
> Other browser work, such as rendering, does not occur until after the entire event dispatch and handling has completed, in both cases.
```

yuval-a: > @Josh-Cena @dipikabh Looking at this text ^^^ it seems pretty unhelpful to me - should I as a user of this method need to understand stack frames or microtask? No, no I should not - certainly not for this method where knowing this makes no difference.
> 
> This was also discussed in #43973, where we were converging towards something like this:
> 
> It is possibly not quite correct, but it captures the main point of the original, which is that these events are synchronous, whereas events generated by the browser are not.
> 
> Anyway, my take on this is that something simple here is better or deleting the text entirely, as largely irrelevant to readers.
> 
> I'll leave to you, and close the other issue as it was "late to the party".

@hamishwillee I disagree about the comment about should they be familiar with what a `microtask` is. Yes - they should, `queueMicrotask` is official API. MDN has this guide: https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide.
The event loop is mentioned - which is an even "lower-level" concept.

I think the questions regarding about whether this paragraph should exist, and if so - is this the right place, are:
* Would a reader of the `dispatchEvent` page will likely be interested in a quick summary of the differences between "custom and native"? I say: yes.
* What would represent the practical differences? I say: mainly the microtasks difference, the non render between handlers, then I also think the `isTrusted`.

On the other hand - your last suggestion is short and concise - and does "fix" the technical semantic "mistake" that triggered me to open this PR.

hamishwillee: > The fundamental question is whether there's race—i.e., if it's possible for other jobs to execute in between two event handler executions.

@Josh-Cena If this is the fundamental question/problem then IMO should be addressed directly rather than highlighting that some things are sync and others are async. I.e. start with the problem.

@yuval-a As a general principle I think we will have to agree to disagree on microtasks - most programmers most of the time do not need to care about that level of detail.  Of course it may be that this is a case where the problem can't be explained without them. Fortunately I'm not the reviewer, so I don't need to assess that.

The  paragraph on isTrusted seems useful YOu might consider a few words on why it matters (in addition to the link)

yuval-a: @hamishwillee I thought about just mentioning Promises but it sounds too partial.
Anyway, I think the subject of "do we go there?" about Microtasks - is like a "standards/specs" question. I'm not a long-running contributor to MDN - I came here because I was triggered by the phrasing of that paragraph - but do MDN have a formal "standards" document? like a "style guide"? If not, it should - I was going per what I know already exists. But anyway I emphasis again that I think it's very legit to use that term;
1. As I mentioned there's actually an official global method with that term in its name (for me that alone is enough to justify it).
2. Microtasks is the de facto mechanism of "asynchronicity" in Javascript - before it existed it was just "offload to the host browser" for async. To the best of my knowledge adding it was a design choice spawned from: "We need a way to allow these "clean-up/just-before" tasks inside Javascript. It's used for Promise resolve callbacks, but also for Custom Elements reactions, MutationObserver callbacks and likely other stuff. It's a thing in Javascript, like, a major thing. 

Anyway, a compromise would be to mention Promises instead and add "Microtasks" parenthesised. Also, I'm leaning more and more towards a note that explains the two different types of events.

This is my next suggestion:

```suggestion
> [!NOTE]
> Events dispatched programmatically using `dispatchEvent()` (as well as by other
> means, such as calling [`click()`](/en-US/docs/Web/API/HTMLElement/click))
> are informally known as "synthetic events", while events dispatched by
> the browser are often referred to as "native events".
>
> Synthetic events (including custom events) have their
> [`isTrusted`](/en-US/docs/Web/API/Event/isTrusted) property set to `false`, while
> browser-generated events have it set to `true`. This allows applications 
> and browser features to distinguish between browser-generated events 
> and script-dispatched events.
>
> Additionally, [Promise](en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) 
> callbacks (and other microtasks) scheduled from event handlers of
> synthetic events will run only after all event handlers have completed running. For
> browser-dispatched events, microtasks may run between individual event handler 
> invocations during dispatch. Other browser work, such as rendering, occurs only after
> the entire event dispatch has completed in both cases.
```

hamishwillee: @yuval-a Thanks. I think this is getting htere ^^^. I wouldn't necessarily have (all of) it as a note because it isn't really all aside material. You would want to tidy this bit "scheduled from event handlers of "synthetic events will run only after all event handlers have completed running." because it compares event handlers and event handlers without differentiation of the types you mean.

But I am leaving this to @dipikabh and @Josh-Cena

dipikabh: My 2 cents here: The line just before Line 22 is already a note callout. Also, the proposed text in https://github.com/mdn/content/pull/43521#discussion_r3205015741 would introduce a large block of text before "Syntax" - @hamishwillee is that normal for the API reference area, or should more details like this go in a "Description" section instead?

@yuval-a I am okay going with following the earlier suggestion from Hamish, it focuses on the original issue and keeps it minimal:

"Unlike "native" events, which the browser fires by queuing a task on the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model), `dispatchEvent()` invokes all applicable event handlers _synchronously_ before returning."

@Josh-Cena, can you confirm if this version looks okay to you?

hamishwillee: FWIW perfect is the enemy of good. The minimal fix might be imperfect, but we seem to be stalled here.

yuval-a: @dipikabh I am okay with the phrasing you presented in your comment as it represents loyalfully the technicality. Sorry for the delay in the response.

yuval-a: I commited a change with your phrasing verbatim, plus a sentence about `isTrusted`. I don't think it needs further explanation, because it's mostly just that - to flag a differentiation between 'native' and 'synthetic' (plus there's a link to the page about it).

files/en-us/web/api/eventtarget/dispatchevent/index.md
```diff
@@ -19,10 +19,7 @@ should have already been created and initialized using an {{domxref("Event/Event
 > [!NOTE]
 > When calling this method, the {{domxref("Event.target")}} property is initialized to the current `EventTarget`.
 
-Unlike "native" events, which are fired by the browser and invoke event handlers
-asynchronously via the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model),
-`dispatchEvent()` invokes event handlers _synchronously_. All applicable event
-handlers are called and return before `dispatchEvent()` returns.
+Unlike "native" events, which the browser fires by queuing a task on the [event loop](/en-US/docs/Web/API/Document_Object_Model/Events#creating_and_dispatching_events), `dispatchEvent()` invokes all applicable event handlers synchronously before returning. [Event](/en-US/docs/Web/API/Event) objects for native events have the read-only [`isTrusted`](/en-US/docs/Web/API/Event/isTrusted) property set to `true`, while events dispatched with `dispatchEvent()` have it set to `false`.
```
dipikabh: - Any reason for changing the link? The originally linked page looks more appropriate for "event loop". In fact, we could link to this section `/en-US/docs/Web/JavaScript/Reference/Execution_model#job_queue_and_event_loop`
- Tweaked the 2nd sentence

```suggestion
Unlike "native" events, which the browser fires by queuing a task on the [event loop](/en-US/docs/Web/JavaScript/Reference/Execution_model#job_queue_and_event_loop), `dispatchEvent()` invokes all applicable event handlers synchronously before returning. The read-only [`isTrusted`](/en-US/docs/Web/API/Event/isTrusted) property is `true` for native events and `false` for events dispatched with `dispatchEvent()`.
```

dipikabh (changes requested):
Thanks @yuval-a.

If these suggestions look good, we can land it with these updates

dipikabh (approved):
Thanks everyone for your inputs and discussion on shaping this update! I'll be merging it now. @Josh-Cena, feel free to follow up if anything looks off.

Thanks @yuval-a for your persistence in seeing this through 🙏

PR #43528, opened 2026-03-23
Editorial review: Document PerformanceNavigationTiming.confidence

chrisdavidmills:
### Description

Chrome 145 adds support for the `PerformanceNavigationTiming.confidence` property, and the associated `PerformanceTimingConfidence` interface. See https://chromestatus.com/feature/5186950448283648.

This PR adds documentation for both features mentioned above.


hamishwillee:
@chrisdavidmills This says "technical review". Is it ready for me to look at?

chrisdavidmills:
> @chrisdavidmills This says "technical review". Is it ready for me to look at?

@hamishwillee. Not yet; I requested a tech review from the browser engineers yesterday. Once it is ready, I'll flip it to "Editorial review".

files/en-us/web/api/performancenavigationtiming/confidence/index.md
```diff
@@ -0,0 +1,52 @@
...
+slug: Web/API/PerformanceNavigationTiming/confidence
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.PerformanceNavigationTiming.confidence
+---
+
+{{APIRef("Performance API")}}{{SeeCompatTable}}
+
+The **`confidence`** read-only property of the {{domxref("PerformanceNavigationTiming")}} interface returns a {{domxref("PerformanceTimingConfidence")}} object containing information that indicates whether the user agent considers returned navigation metrics to be representative of the current user's device.
```
mmocny: > "representative of the current user's device"

I see that this wording comes from the navigation timing spec itself.  Personally, I don't like that phrasing.

It's more like "known to have adverse conditions affecting performance" or something.

I think its fine to leave to match spec, but maybe Mike Jackson wants to take the chance to wordsmith?

files/en-us/web/api/performancenavigationtiming/confidence/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+For example, if the browser has launched from a "cold start" before loading a website or has resource-intensive extensions running, or if other applications running on the user's device are taking up a lot of resource cycles, web pages may load more slowly as a result. In such cases, a `low` confidence {{domxref("PerformanceTimingConfidence.value", "value")}} would be returned.
```
mmocny: Some of these are aspirational right now.

Should we make it clearer what is actually (I think cold startup?) vs might one day be affecting?

Mike is the expert.

mmocny (commented):
The docs look great, thanks for doing it!

I'm not the primary engineering contact for this, so hopefully Mike Jackson at Msft will have a change to take a look.

One detail that is missing from the docs: how should you use this value and interpret the data on the server?  This feels like the most important part of the API, and also the harder to understand for developers.

Mike has done some presentations on this, and I see that he added a NOTE to the very bottom of this spection of the spec: https://www.w3.org/TR/navigation-timing-2/#sec-PerformanceNavigationTiming

> This section is intended to help RUM providers and developers interpret confidence

...that section might be worth including in docs here?

Cheers.

files/en-us/web/api/performancenavigationtiming/confidence/index.md
mwjacksonmsft: Maybe something like: returned navigation metrics free from external system load unrelated to the page.

files/en-us/web/api/performancenavigationtiming/confidence/index.md
mwjacksonmsft: At the moment, this works for cold start, and session restore. I'd prefer that we only highlight those items (or call out that the others are potential future cases and not currently implemented).

files/en-us/web/api/performancenavigationtiming/index.md
```diff
@@ -40,6 +40,8 @@ The interface also supports the following properties:
 
 - {{domxref('PerformanceNavigationTiming.activationStart')}} {{ReadOnlyInline}} {{experimental_inline}}
   - : A {{domxref("DOMHighResTimeStamp")}} representing the time between when a document starts prerendering and when it is activated.
+- {{domxref('PerformanceNavigationTiming.confidence')}} {{ReadOnlyInline}} {{experimental_inline}}
+  - : A {{domxref("PerformanceTimingConfidence")}} object containing information that indicates whether the user agent considers returned navigation metrics to be representative of the current user's device.
```
mwjacksonmsft: Should we adjust the language here as well to match confidence/index.md?

files/en-us/web/api/performancetimingconfidence/value/index.md
```diff
@@ -0,0 +1,33 @@
...
+slug: Web/API/PerformanceTimingConfidence/value
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.PerformanceTimingConfidence.value
+---
+
+{{APIRef("Performance API")}}{{SeeCompatTable}}
+
+The **`value`** read-only property of the {{domxref("PerformanceTimingConfidence")}} interface is an enumerated value indicating a broad confidence measure of whether the user agent considers returned navigation metrics to be representative of the current user's device.
```
mwjacksonmsft: Same question here - should we update this to match the verbiage in confidence/index.md

chrisdavidmills:
> Mike has done some presentations on this, and I see that he added a NOTE to the very bottom of this section of the spec: https://www.w3.org/TR/navigation-timing-2/#sec-PerformanceNavigationTiming
> 
> > This section is intended to help RUM providers and developers interpret confidence
> 
> ...that section might be worth including in docs here?

This makes sense. For the moment, I've gone for including all the text in this section in the `PerformanceTimingConfidence` page, under a heading of "Interpreting confidence data". I've not made many changes, except for a few tweaks, and adding links to the different values.

Anyway, I'll include that in my next commit.

files/en-us/web/api/performancenavigationtiming/confidence/index.md
chrisdavidmills: I like @mwjacksonmsft's wording; I'll update the wording to that in the places where it appears.

files/en-us/web/api/performancenavigationtiming/confidence/index.md
chrisdavidmills: I've updated the wording on both pages it appears on to:

> For example, if a website has loaded after a browser "cold start" or session restore, its pages may load more slowly as a result.

files/en-us/web/api/performancenavigationtiming/index.md
chrisdavidmills: Good call; done.

files/en-us/web/api/performancetimingconfidence/value/index.md
chrisdavidmills: done

mwjacksonmsft (commented):
These changes LGTM. Thanks!

chrisdavidmills:
Cool, thanks, @mwjacksonmsft. I'll move this to the editorial review stage.

@hamishwillee, ready for you to have a look, if you've still got time early next week.

files/en-us/web/api/performancetimingconfidence/index.md
```diff
@@ -0,0 +1,101 @@
...
+browser-compat: api.PerformanceTimingConfidence
+---
+
+{{APIRef("Performance API")}}
+
+The **`PerformanceTimingConfidence`** interface provides access to information that indicates whether the user agent considers returned navigation metrics to be free from external system load unrelated to the page.
+
+For example, if a website has loaded after a browser "cold start" or session restore, its pages may load more slowly as a result. In such cases, a `low` confidence {{domxref("PerformanceTimingConfidence.value", "value")}} would be returned for an associated performance record. On the other hand, if the browser determines a returned performance record to be representative of typical application performance, a `high` confidence value is returned.
+
+This confidence measure is useful for developers when trying to determine whether a performance issue is a legitimate concern, or an outlier being caused by external factors.
```
hamishwillee: If you have a lot to say, interface pages tend to push the discussion down to a description.
I would push this and the "Interpreting confidence data".

files/en-us/web/api/performancetimingconfidence/index.md
```diff
@@ -0,0 +1,101 @@
...
+     > [!NOTE]
+     > `w` may be negative for some records; you should keep every record.
+   - Let `weighted_duration = duration * w` (see {{domxref("PerformanceEntry.duration", "duration")}}).
+3. Let `total_weighted_duration` be the sum of the `weighted_duration` values across all records.
+4. Let `sum_weights` be the sum of the `w` values across all records.
+5. Let `debiased_mean = total_weighted_duration / sum_weights`, provided `sum_weights` is not near zero.
+
+To compute debiased percentiles for both `high` and `low`:
+
+1. Follow the _computing debiased means_ steps to compute a per-record weight `w`.
```
hamishwillee: This is unambiguous in the spec because " _computing debiased means_" would be a link.

files/en-us/web/api/performancetimingconfidence/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+This confidence measure is useful for developers when trying to determine whether a performance issue is a legitimate concern, or an outlier being caused by external factors.
+
+The `PerformanceTimingConfidence` object for each navigation timing entry is accessed via the {{domxref("PerformanceNavigationTiming")}} interface's {{domxref("PerformanceNavigationTiming.confidence", "confidence")}} property.
+
+## Interpreting confidence data
+
+Since the {{domxref("PerformanceTimingConfidence.randomizedTriggerRate", "randomizedTriggerRate")}} can vary across records, per-record weighting is needed to recover unbiased aggregates. The procedure below illustrates how weighting based on {{domxref("PerformanceTimingConfidence.value", "value")}} can be applied before computing summary statistics.
```
hamishwillee: There is always an argument around what you should expect a developer to reasonably interpret and how much hand holding you should do. To my mind though, basic questions like "why do I need to do this have not been answered".

Let's start backwards. Why are summary statistics needed and how do I use those?
What is an unbiased aggregate? Why do I care to recover one? Blah blah. 

To put it another way, I can see myself following these instructions and generating the data, but then not knowing what to do with it.

files/en-us/web/api/performancenavigationtiming/confidence/index.md
```diff
@@ -0,0 +1,52 @@
...
+slug: Web/API/PerformanceNavigationTiming/confidence
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.PerformanceNavigationTiming.confidence
+---
+
+{{APIRef("Performance API")}}{{SeeCompatTable}}
+
+The **`confidence`** read-only property of the {{domxref("PerformanceNavigationTiming")}} interface returns a {{domxref("PerformanceTimingConfidence")}} object containing information that indicates whether the user agent considers returned navigation metrics to be free from external system load unrelated to the page.
```
hamishwillee: IMO it's not so much "free from" as "likely to be affected by". Something like

```suggestion
The **`confidence`** read-only property of the {{domxref("PerformanceNavigationTiming")}} interface returns a {{domxref("PerformanceTimingConfidence")}} object containing information that indicates whether the user agent considers returned navigation metrics to be affected by external system load (unrelated to the page).
```

YOu could also steal your own words from further down "considers a performance record to reflect typical application performance, or possibly be affected by external factors."

hamishwillee: It's a mouthful anyway, and it might be reasonable to assume that it is the user-agent's assumptions. If we can make that assumption maybe



```suggestion
The **`confidence`** read-only property of the {{domxref("PerformanceNavigationTiming")}} interface returns a {{domxref("PerformanceTimingConfidence")}} object containing information that indicates the likelihood that navigation metrics have been been affected by factors unrelated to the page, such as external system load.
```

files/en-us/web/api/performancenavigationtiming/index.md
```diff
@@ -40,6 +40,8 @@ The interface also supports the following properties:
 
 - {{domxref('PerformanceNavigationTiming.activationStart')}} {{ReadOnlyInline}} {{experimental_inline}}
   - : A {{domxref("DOMHighResTimeStamp")}} representing the time between when a document starts prerendering and when it is activated.
+- {{domxref('PerformanceNavigationTiming.confidence')}} {{ReadOnlyInline}} {{experimental_inline}}
+  - : A {{domxref("PerformanceTimingConfidence")}} object containing information that indicates whether the user agent considers returned navigation metrics to be free from external system load unrelated to the page.
```
hamishwillee: As above, not so sure of "Free from".

files/en-us/web/api/performancetimingconfidence/randomizedtriggerrate/index.md
```diff
@@ -0,0 +1,33 @@
...
+slug: Web/API/PerformanceTimingConfidence/randomizedTriggerRate
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.PerformanceTimingConfidence.randomizedTriggerRate
+---
+
+{{APIRef("Performance API")}}{{SeeCompatTable}}
+
+The **`randomizedTriggerRate`** read-only property of the {{domxref("PerformanceTimingConfidence")}} interface is a number representing a percentage value that indicates how often noise is applied when exposing the {{domxref("PerformanceTimingConfidence.value")}}.
```
hamishwillee: This is very complete and accurate, but it is  hard to parse. Possibly it is not necessary to capture this all in one sentence here, since you should to that in the value.

```suggestion
The **`randomizedTriggerRate`** read-only property of the {{domxref("PerformanceTimingConfidence")}} indicates how often noise is applied when exposing the {{domxref("PerformanceTimingConfidence.value")}}.
```

hamishwillee: Either way 
- why do we add noise? We should say, and also state what a high rate actually means vs a low rate.
- So 100% (1) would mean every `PerformanceTimingConfidence` has noise applied to the value. 
- If noise is applied does that flip the `value`. 

Just trying to get a feel for what a developer might do or not do with this knowledge.

files/en-us/web/api/performancetimingconfidence/randomizedtriggerrate/index.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Value
+
+A number inside the interval `0` to `1` inclusive.
```
hamishwillee: How you do this depends on whether you do anything like the suggestion in https://github.com/mdn/content/pull/43528/changes#r3007055024

But, I would use wording like this if you don't change anything above.
```suggestion
A number between `0` and `1`, inclusive.
```

files/en-us/web/api/performancetimingconfidence/value/index.md
```diff
@@ -0,0 +1,33 @@
...
+slug: Web/API/PerformanceTimingConfidence/value
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.PerformanceTimingConfidence.value
+---
+
+{{APIRef("Performance API")}}{{SeeCompatTable}}
+
+The **`value`** read-only property of the {{domxref("PerformanceTimingConfidence")}} interface is an enumerated value indicating a broad confidence measure of whether the user agent considers returned navigation metrics to be free from external system load unrelated to the page.
```
hamishwillee: A above...
```suggestion
The **`value`** read-only property of the {{domxref("PerformanceTimingConfidence")}} interface is an enumerated value indicating a broad confidence measure of whether the user agent considers returned navigation metrics to be affected by external system load unrelated to the page.
```

files/en-us/web/api/performancetimingconfidence/index.md
```diff
@@ -0,0 +1,101 @@
...
+     - Let `lower_duration` be `duration` for `sorted_records[idx-1]`.
+     - Let `upper_duration` be `duration` for `sorted_records[idx]`.
+     - Return `lower_duration + (upper_duration - lower_duration) * ifrac`.
+
+## Instance properties
+
+- {{domxref("PerformanceTimingConfidence.randomizedTriggerRate")}} {{ReadOnlyInline}}
+  - : A number representing a percentage value that indicates how often noise is applied when exposing the `value`.
+- {{domxref("PerformanceTimingConfidence.value")}} {{ReadOnlyInline}}
+  - : An enumerated value indicating a broad confidence measure of whether the user agent considers returned navigation metrics to be representative of the current user's device.
```
hamishwillee: This isn't right. The spec explicitly states that the confidence must not consider specifics fo the user device. YOu might mean "of the application".

Note, I do think you should clearly state that what contributes to the assessement, and make it clear that CPU etc do not count.

hamishwillee (commented):
Looks pretty good. I have questions. 

It might be nice to mention this in https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Navigation_timing

files/en-us/web/api/performancetimingconfidence/index.md
chrisdavidmills: Good call; done in next commit.

files/en-us/web/api/performancetimingconfidence/index.md
chrisdavidmills: I've added separate headings for each algorithm, and linked this text to the first heading to remove the ambiguity.

files/en-us/web/api/performancetimingconfidence/index.md
chrisdavidmills: @mwjacksonmsft can you provide a short paragraph that answers these questions, which I can edit into some finished prose? I don't know the answers to these questions.

files/en-us/web/api/performancenavigationtiming/confidence/index.md
chrisdavidmills: I preferred my bit of text that you hinted at using, but I edited it down to:

> indicates whether a performance record reflects typical application performance, or is likely affected by external factors.

I've used this wording in all the places...

files/en-us/web/api/performancenavigationtiming/index.md
chrisdavidmills: updated

files/en-us/web/api/performancetimingconfidence/randomizedtriggerrate/index.md
chrisdavidmills: @mwjacksonmsft can you provide answers to these questions?

files/en-us/web/api/performancetimingconfidence/randomizedtriggerrate/index.md
chrisdavidmills: This is better wording regardless. Updated in my next commit.

files/en-us/web/api/performancetimingconfidence/value/index.md
chrisdavidmills: updated as previously discussed.

files/en-us/web/api/performancetimingconfidence/index.md
chrisdavidmills: I've changed it to the same wording used to describe the overall interface:

> An enumerated value indicating a broad confidence measure of whether a performance record reflects typical application performance, or is likely affected by external factors.

On the main `PerformanceTimingConfidence` page, I've added the following note:

> [!NOTE]
> Device factors such as CPU do not contribute to the performance assessment. Other factors than browser "cold start" and session restore may be taken into account in future updates.

I only wanted to add it in one place to make for less maintenance work.

chrisdavidmills:
> It might be nice to mention this in https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Navigation_timing

Good point; I've added a section to cover it.

files/en-us/web/api/performancenavigationtiming/confidence/index.md
hamishwillee: Yeah, that's good.

hamishwillee (commented):
Those changes look excellent. Can you ping me again direclty when you've got the remaining answers and integrated them?

chrisdavidmills:
> Those changes look excellent. Can you ping me again direclty when you've got the remaining answers and integrated them?

Yes, will do—cheers mate.

hamishwillee:
Note, I haven't been pinged back on this one AFAIK.

chrisdavidmills:
@mmocny @mwjacksonmsft, there are a couple of outstanding questions that came up in the editorial review that are blocking publication of this documentation. Can you look at them and help me with some answers? I've closed all the resolved comments, so they should be easy to find. Thanks!

files/en-us/web/api/performancetimingconfidence/index.md
mmocny: I didn't want to write new prose for @mwjacksonmsft here, but went looking for existing text on the subject.

The spec doesn't really elaborate, but the original [design doc](https://docs.google.com/document/d/1D6DqptsCEd3wPRsZ0q1iwVBAXXmhxZuLV-KKFI0ptCg/edit?tab=t.0#heading=h.7nki9mck5t64) does.

Stealing from there:

> ### **Summary**
> 
> Web applications may suffer from bimodal distribution in page load performance, due to factors outside of the web application’s control. For example:
> 
> * When a user agent first launches (a "cold start" scenario), it must perform many expensive initialization tasks that compete for resources on the system.  
> * Browser extensions can affect the performance of a website. For instance, some extensions run additional code on every page you visit, which can increase CPU usage and result in slower response times.  
> * When a machine is busy performing intensive tasks, it can lead to slower loading of web pages.
> 
> In these scenarios, content the web app attempts to load will be in competition with other work happening on the system. This makes it difficult to detect if performance issues exist within web applications themselves, or because of external factors.
> 
> Teams we have worked with have been surprised at the difference between real-world dashboard metrics and what they observe in page profiling tools. Without more information, it is challenging for developers to understand if (and when) their applications may be misbehaving or are simply being loaded in a contended period.
> 
> A new ‘confidence’ field on the PerformanceNavigationTiming object will enable developers to discern if the navigation timings are representative for their web application.

mmocny:
I spot only a single unresolved comment in the patch at this point-- but github UI claims there are two unresolved comments.  (Perhaps one is stale from a line in the patch that has been removed, not sure).

Let me know if I haven't found all the questions that need answering.

files/en-us/web/api/performancetimingconfidence/index.md
mmocny: Also, re-reading this patch, the description in the main navigation timing doc section `Performance timing confidence` seems to answer these questions already, so may just need linking to?

---

If the question here is even more broad, such as: why do developers measure performance in the field? Then I would also point to existing docs on the subject rather than answer them here.

The point of this feature (`confidence`) is to help segment field data into two distinct groups, with the observation that the high-confidence results are more stable over time, but the relative distribution between the groups can change.

chrisdavidmills: OK, thanks. I've done some research of my own, and added a bit of information on why unbiased aggregates are needed.

It seems to me that summary statistics just refers to the statistics you produce from the raw data, which you actually give people to read. I therefore don't think this needs a huge amount of explanation, but I have added a few more words to indicate that they are statistics based on the confidence data.

@hamishwillee, let me know what you think.

mwjacksonmsft: The comment @mmocny left captures the why. 

> Teams we have worked with have been surprised at the difference between real-world dashboard metrics and what they observe in page profiling tools. Without more information, it is challenging for developers to understand if (and when) their applications may be misbehaving or are simply being loaded in a contended period.

Developers can debias the data, and then focus on measuring and improving perf for things under their control.

files/en-us/web/api/performancetimingconfidence/randomizedtriggerrate/index.md
chrisdavidmills: @hamishwillee I've done a bit of research here too, and added more details about the randomized trigger rate and noise. Let me know if that answers your questions.

files/en-us/web/api/performancetimingconfidence/index.md
chrisdavidmills: Thanks, @mwjacksonmsft. I've added a couple more sentences to capture some of these thoughts.

files/en-us/web/api/performancetimingconfidence/index.md
```diff
@@ -0,0 +1,112 @@
...
+> Device factors such as CPU do not contribute to the performance assessment. Other factors than browser "cold start" and session restore may be taken into account in future updates.
+
+This confidence measure is useful for developers when trying to determine whether a performance issue is a legitimate concern, or an outlier being caused by external factors. There is often a significant difference between real-world dashboard metrics and performance observations in page profiling tools.
+
+### Interpreting confidence data
+
+Since the {{domxref("PerformanceTimingConfidence.randomizedTriggerRate", "randomizedTriggerRate")}} can vary across records, per-record weighting is needed to recover unbiased aggregates, to improve consistency of data, cut down on compound errors, and generally produce more realistic and reliable results. The procedures below illustrate how weighting based on {{domxref("PerformanceTimingConfidence.value", "value")}} can be applied before computing summary statistics based on the confidence data.
+
+Once you have debiased the data and computed realistic summary statistics, you can focus on measuring and improving performance for issues under your control.
```
hamishwillee: @chrisdavidmills Same problem as I highlighted before - I didn't understand the "point" from this text and what you would do when you have the debiased data. 
I asked Claude if this paragraph was just marketing and apparently it isn't :-). Apparently  the term "recover unbiased aggregates" means something :-)

The point that is a bit buried is that `value` is not deterministic. The browser uses randomization to assign "high" or "low". When `p = 0.1` (say), it means 10% of the time the value you see was randomly assigned regardless of actual conditions.

So you can't just filter out "low" records and average the "high" ones to work out your real performance — you'd be throwing away records that were actually fine but happened to get a random "low", and keeping records that were actually bad but got a random "high".

The debiasing math corrects for the random noise so that your aggregate statistics (mean, p75, etc.) are statistically valid.  This is what the paragraph above did not make clear to me. Perhaps I am dim.

Claude says that what I'd do with the data if collecting navigation timing data (e.g. for a real-user monitoring dashboard):

1. Collect records via `PerformanceObserver` as normal.
2. For each record, also grab `entry.confidence.value` and `entry.confidence.randomizedTriggerRate`.
3. When computing your p75 LCP or mean load time, apply the weighting formulas instead of a plain average — this gives you separate, corrected metrics for "typical" loads vs. "degraded" loads.
4. Use the `"high"` confidence mean/percentile as your "real" performance baseline, and use the `"low"` one to understand how bad things get in cold-start scenarios.

This last bit is what I meant by "what do you do with the data" - use it as a new baseline.

Does my problem now make sense?

chrisdavidmills: This kind of makes more sense now. I've plugged the stuff you've provided above into my doc and committed it. Better?

files/en-us/web/api/performancetimingconfidence/index.md
```diff
@@ -0,0 +1,116 @@
...
+
+This confidence measure is useful for developers when trying to determine whether a performance issue is a legitimate concern, or an outlier being caused by external factors. There is often a significant difference between real-world dashboard metrics and performance observations in page profiling tools.
+
+### Using the data
+
+Since the `randomizedTriggerRate` can vary across records, per-record weighting is needed to recover unbiased aggregates, to improve consistency of data, cut down on compound errors, and generally produce more realistic and reliable results. You should use the data as follows, to get effective results:
+
+1. When collecting {{domxref("PerformanceNavigationTiming")}} records, collect {{domxref("PerformanceTimingConfidence.randomizedTriggerRate", "randomizedTriggerRate")}} and {{domxref("PerformanceTimingConfidence.value", "value")}} for each record.
+2. When computing statistics such as p75 {{glossary("Largest_contentful_paint", "LCP")}} or mean {{glossary("page load time")}}, apply the weighting formulas explained below instead of a plain average — this gives you separate, corrected metrics for "typical" loads vs. "degraded" loads.
```
hamishwillee: Jargon 75p, and LCP undefined here.

```suggestion
2. When computing statistics such as 75th-percentile {{glossary("Largest_contentful_paint", "Largest contentful paint (LCP)")}} or mean {{glossary("page load time")}}, apply the weighting formulas explained below instead of a plain average — this gives you separate, corrected metrics for "typical" loads vs. "degraded" loads.
```

files/en-us/web/api/performancetimingconfidence/index.md
```diff
@@ -0,0 +1,116 @@
...
+  - : Returns a JSON representation of the `PerformanceTimingConfidence` object.
+
+## Description
+
+If a website has loaded after a browser "cold start" or session restore, its pages may load more slowly as a result. In such cases, a `low` confidence {{domxref("PerformanceTimingConfidence.value", "value")}} is returned for an associated performance record. On the other hand, if the browser determines a returned performance record to be representative of typical application performance, a `high` confidence value is returned.
+
+> [!NOTE]
+> Device factors such as CPU do not contribute to the performance assessment. Other factors than browser "cold start" and session restore may be taken into account in future updates.
+
+This confidence measure is useful for developers when trying to determine whether a performance issue is a legitimate concern, or an outlier being caused by external factors. There is often a significant difference between real-world dashboard metrics and performance observations in page profiling tools.
```
hamishwillee: Later on you already need to know about randomisation - see https://github.com/mdn/content/pull/43528/changes#r3246340693

Lets improve this so that "using the data" makes sense. This is mostly your own words restructured.


```suggestion
If a website has loaded after a browser "cold start" or session restore, its pages may load more slowly as a result. 
This can cause a significant difference between real-world dashboard metrics and performance observations in page profiling tools, making it hard for a developer to understand whether a performance issue is a legitimate concern or an outlier caused by external factors.

The `PerformanceTimingConfidence` interface allows developers to compensate for this problem by returning a browser estimate (in the {{domxref("PerformanceTimingConfidence.value", "value")}} property) of the likelihood that a returned performance record represents typical application performance.
This is a value of either `"low"` or `"high"`, indicating the browser's confidence in the measurement.

> [!NOTE]
> Device factors such as CPU do not contribute to the performance assessment. Other factors than browser "cold start" and session restore may be taken into account in future updates.

To reduce the possibility of using the value for fingerprinting, noise is added to the estimate, meaning the `value` will deliberately be wrong for some proportion of results.
The trigger rate for the noise is given in the {{domxref("PerformanceTimingConfidence.randomizedTriggerRate", "randomizedTriggerRate")}} property.

Since this can vary across records, per-record weighting is needed to recover unbiased aggregates, to improve data consistency, reduce the number of compound errors, and generally to produce a baseline against which the measured results can be evaluated.
```

files/en-us/web/api/performancetimingconfidence/index.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Using the data
+
+Since the `randomizedTriggerRate` can vary across records, per-record weighting is needed to recover unbiased aggregates, to improve consistency of data, cut down on compound errors, and generally produce more realistic and reliable results. You should use the data as follows, to get effective results:
```
hamishwillee: Nearly!

At this point the user doesn't know about `randomizedTriggerRate` - other than it has noise.  I have proposed an into above in https://github.com/mdn/content/pull/43528/changes#r3246332436

Then ...

```suggestion
You should use the data as follows to extract meaninful information from the randomized values:
```

files/en-us/web/api/performancetimingconfidence/index.md
chrisdavidmills: Well, I linked to the LCP page for those who don't know, and even I know what 75p means, being someone who has very limited knowledge of statistics. 

But I'm not going to argue with you. I just want to see this one put to bed now.

files/en-us/web/api/performancetimingconfidence/index.md
chrisdavidmills: OK, I've edited this suggestion a bit, but it reads well. Thanks.

files/en-us/web/api/performancetimingconfidence/index.md
hamishwillee: Fair enough on LCP, but clearly my limited knowledge is even more limited, because I didn't know/remember what 75p was. Thanks for accepting.

hamishwillee (approved):
Thanks very much @chrisdavidmills - this looks good.

chrisdavidmills:
Cool, cheers @hamishwillee!

PR #43530, opened 2026-03-23
Documentation clarity improvement

sohilsshah:
Style declarations are filtered by a condition and applied to the container if the condition is true.

The current wording says that styles are “applied to the container if the condition is true,” which seems wrong.

From my understanding, container queries evaluate the container’s size or style, but the rules inside @container apply to descendant elements, not the container itself.

I think it should say "Style declarations are filtered by a condition and applied to the elements if the condition is true."


files/en-us/web/css/reference/at-rules/@container/index.md
```diff
@@ -7,7 +7,7 @@ sidebar: cssref
 ---
 
 The **`@container`** [CSS](/en-US/docs/Web/CSS) [at-rule](/en-US/docs/Web/CSS/Guides/Syntax/At-rules) is a conditional group rule that applies styles to a [containment context](/en-US/docs/Web/CSS/Guides/Containment/Container_queries#naming_containment_contexts).
-Style declarations are filtered by a condition and applied to the container if the condition is true.
+Style declarations are filtered by a condition and applied to the elements if the condition is true.
```
dipikabh: this can be clarified further:
```suggestion
Style declarations are filtered by a condition and applied to the elements within the container if the condition is true.
```

files/en-us/web/css/reference/at-rules/@container/index.md
dipikabh: Noticed the same issue here. Also moved `ident` closer to what it's describing:
```suggestion
      - : The name of the container to query; it is specified as an {{cssxref("ident")}}. If the query evaluates to `true`, the declared styles are applied to the container's descendant elements.
```

sohilsshah: Thank you, @dipikabh, for the warm welcome and for diving into the requested change so quickly and thoroughly. I truly appreciate it—this feels like a great positive start to being a part of MDN. 🎉

dipikabh: Feel free to apply suggestions directly from the review using "Commit suggestion".
Let me know if you need help with resolving conflicts.

PR #43534, opened 2026-03-23
Update deprecated margin attributes on body

zcorpan:
### Description

Update the descriptions for these attributes to match the spec change.

### Motivation

MDN should match the spec.

### Additional details

See https://github.com/whatwg/html/pull/11881

### DIFF
--- a/files/en-us/web/html/reference/elements/body/index.md
+++ b/files/en-us/web/html/reference/elements/body/index.md
@@ -79,23 +79,21 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
   - : Background color for the document.
     Use the CSS {{cssxref("background-color")}} property instead.
 - `bottommargin` {{deprecated_inline}}
-  - : The margin of the bottom of the body.
-    Use the CSS {{cssxref("margin-bottom")}} property (or the logical {{cssxref("margin-block-end")}} property) instead.
+  - : Ignored.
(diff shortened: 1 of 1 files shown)


pepelsbey:
> MDN should match the spec.

Not exactly 😬 We document browser behavior for real-world usage, that’s only _based_ on specs. But the implementation might be different.

So I would test the implementation and only then update the docs.

pepelsbey:
@sideshowbarker have you seen my comment?

sideshowbarker:
> @sideshowbarker have you seen my comment?

Sorry no, I somehow missed it

zcorpan:
@pepelsbey fair, but the recommendation for these attributes is "don't use". Is it helpful to explain how these attributes were previously non-interoperable?

pepelsbey:
> Is it helpful to explain how these attributes were previously non-interoperable?

@zcorpan I'm only saying that if they _are_ supported in browsers for backward compatibility reasons, we should reflect that.

zcorpan:
I agree, and this PR aligns with that. Chrome and Safari already ignore `marginright` and when https://bugzilla.mozilla.org/show_bug.cgi?id=2025580 is fixed Firefox will also ignore `marginright` (for better web compat).

pepelsbey:
> I agree, and this PR aligns with that.

Perfect! Thank you

PR #43543, opened 2026-03-24
FF149 ShadowRoot referenceTarget and friends minor update

hamishwillee:
FF149 added support for `referenceTarget` and `shadowRootReferenceTarget` attributes behind a pref. These are already mostly documented in #42323. This is pretty basic, but it was and remains about the right level because this isnt supported in releases or preview yet. At that point we should create guide material and cross link (this can be generated from the [explainer](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/reference-target-explainer.md).

What this does is get us a little bit closer so that when we have to update there is less to do.
1. Update the docs to use correct templates
2. Add linked individual property property docs rather than just the top level doc.

I have put a little bit more work into `ShadowRoot.referenceTarget` docs because it ONLY affects this property. So we could in theory add examples and so on to this without being confusing for other docs. See inline comments.

Related docs work can be tracked in #43453

### DIFF
--- a/files/en-us/web/api/element/attachshadow/index.md
+++ b/files/en-us/web/api/element/attachshadow/index.md
@@ -428,5 +428,6 @@ The example below should show the content of the slots displayed in the appropri
 - {{domxref("ShadowRoot.mode")}}
 - {{domxref("ShadowRoot.delegatesFocus")}}
 - {{domxref("ShadowRoot.slotAssignment")}}
+- {{domxref("ShadowRoot.referenceTarget")}}
 - Declaratively attach a shadow root with the [`shadowrootmode`](/en-US/docs/Web/HTML/Reference/Elements/template#shadowrootmode) attribute of the [`<template>` element](/en-US/docs/Web/HTML/Reference/Elements/template)
 - [Declarative shadow DOM](https://web.dev/articles/declarative-shadow-dom) on web.dev (2023)
--- a/files/en-us/web/api/htmltemplateelement/shadowrootreferencetarget/index.md
+++ b/files/en-us/web/api/htmltemplateelement/shadowrootreferencetarget/index.md
@@ -1,0 +1,33 @@
+---
+title: "HTMLTemplateElement: shadowRootReferenceTarget property"
+short-title: shadowRootReferenceTarget
+slug: Web/API/HTMLTemplateElement/shadowRootReferenceTarget
+page-type: web-api-instance-property
+browser-compat: api.HTMLTemplateElement.shadowRootReferenceTarget
+---
+
+{{APIRef("Web Components")}}
+
+The **`shadowRootReferenceTarget`** property of the {{domxref("HTMLTemplateElement")}} interface reflects the value of the [`shadowrootreferencetarget`](/en-US/docs/Web/HTML/Reference/Elements/template#shadowrootreferencetarget) attribute of the associated [`<template>`](/en-US/docs/Web/HTML/Reference/Elements/template) element.
+
+> [!NOTE]
+> This property is not useful for developers, and is only documented for completeness.
+> If a `<template>` element is used to declaratively create a [`ShadowRoot`](/en-US/docs/Web/API/ShadowRoot), then this object and property do not exist.
+> Otherwise, if an `HTMLTemplateElement` is created, the value of this property is irrelevant because the object is not a shadow root and cannot subsequently be changed to a shadow root.
+
+## Value
+
+Reflects the value of the [`shadowrootreferencetarget`](/en-US/docs/Web/HTML/Reference/Elements/template#shadowrootreferencetarget) attribute of the associated [`<template>`](/en-US/docs/Web/HTML/Reference/Elements/template) element.
+
+## Specifications
+
+{{Specifications}}
+
+## Browser compatibility
+
+{{Compat}}
+
+## See also
+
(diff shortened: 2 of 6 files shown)


files/en-us/web/api/element/attachshadow/index.md
```diff
@@ -8,42 +8,7 @@ browser-compat: api.Element.attachShadow
 
 {{APIRef("Shadow DOM")}}
 
-The **`Element.attachShadow()`** method attaches a shadow DOM tree to the specified element and returns a reference to its {{domxref("ShadowRoot")}}.
```
hamishwillee: This change just moves us to Web API template, moving the following stuff into a description.

files/en-us/web/api/element/attachshadow/index.md
```diff
@@ -112,6 +75,51 @@ Returns a {{domxref("ShadowRoot")}} object.
...
 
+## Description
+
+The **`attachShadow()`** method attaches a {{domxref("ShadowRoot")}} to the specified element and returns a reference to it.
+
+The returned {{domxref("ShadowRoot")}} is the root node of a shadow tree, which can then be populated with its own DOM content.
+The element to which the shadow root is attached is referred to as the "shadow host", and remains in its original DOM.
+
+The `ShadowRoot` acts as a boundary between the original DOM and the shadow tree.
+It controls encapsulation — for example, styles defined inside the shadow tree do not affect the original DOM.
```
hamishwillee: This is new. IMO it is worth saying that the shadow root is what is created, initially empty.

files/en-us/web/api/htmltemplateelement/shadowrootreferencetarget/index.md
```diff
@@ -0,0 +1,33 @@
+---
```
hamishwillee, line 1: FYI "a nothing doc". Not useful but following the pattern of other docs in this this interface.

files/en-us/web/api/shadowroot/referencetarget/index.md
```diff
@@ -0,0 +1,106 @@
...
+
+      const input = this.shadowRoot_.getElementById("input");
+
+      log("aria-checked:", input.getAttribute("aria-checked"));
+      log("aria-label:", input.getAttribute("aria-label"));
+      log("aria-labelledby:", input.getAttribute("aria-labelledby"));
+    }
+  },
+);
+```
```
hamishwillee, line 92: Still not sure what example to show - this is a modification of the explainer. It is hard because you can't propgrammatically log the supposedly affected attributes, such as `aria-labelledby`, to show the effect. 
I explained this above in the description - the browser doesn't change the element, it computes it when needed.

Open to ideas that don't require a screen reader or other accessibility tools.

Also not sure the right way to "feature test".

hamishwillee: @meyerweb Any ideas on a good live example at this early stage of the feature?

Note, I'm not planning on going much further with docs on this until it is in preview release/FF nightly. What you did before was sufficient, this is just a bit of tidy.

files/en-us/web/api/shadowroot/referencetarget/index.md
```diff
@@ -0,0 +1,106 @@
...
+slug: Web/API/ShadowRoot/referenceTarget
+page-type: web-api-instance-property
+browser-compat: api.ShadowRoot.referenceTarget
+---
+
+{{APIRef("Shadow DOM")}}
+
+The **`referenceTarget`** property of the {{domxref("ShadowRoot")}} interface is a nullable string that represents the effective target of any element reference made against the shadow host from outside the host element.
+
+Some elements are associated with other elements using attributes that reference their `id` attribute.
```
hamishwillee:
```suggestion
Some elements are associated with other elements using attributes that take _element references_: strings that specify the other element's `id`.
```

files/en-us/web/api/shadowroot/referencetarget/index.md
meyerweb: It’s a good question!  I’ll ask around internally and see what we can come up with.

(And many thanks for the tidying!)

alice: For feature testing, I think this should work:

```js
ShadowRoot.prototype.hasOwnProperty("referenceTarget");
```

For demonstrating the behaviour - yeah, it's tricky because a _lot_ of the attributes it affects are ARIA attributes which by definition don't have any effect on the DOM.

However, `popoverTarget`, `commandFor` and `for` can show it pretty well, I think. In this example, you could log `label.control` and `input.labels` to show that the label association working. Note that the value of `label.control` will be retargeted to the shadow host.

With `popoverTarget`, you could show an element inside a shadow root with the `popover` attribute, and a button with `popoverTarget` pointing at the host with referenceTarget pointing at the inner popover element, to show that only the element with `popover` is shown when the button is pressed. Likewise, `commandFor` could point at a `<dialog>` inside a shadow root.

alice: You could also look through the WPT tests for reference target for more ideas: https://wpt.fyi/results/shadow-dom/reference-target/tentative?label=experimental&label=master&aligned

hamishwillee: Thanks. I'll be looking at this again either later in the week or next week!

hamishwillee:
OK. I have decided to close this. I might come back to it when this ships in FF - at least in nightly.

PR #43571, opened 2026-03-25
Add 'scroll' to timeline-range-name options

estelle:
the value is in the spec.

https://drafts.csswg.org/scroll-animations/#typedef-timeline-range-name

### DIFF
--- a/files/en-us/web/css/reference/values/timeline-range-name/index.md
+++ b/files/en-us/web/css/reference/values/timeline-range-name/index.md
@@ -38,8 +38,9 @@ Valid `<timeline-range-name>` values:
 
 - `exit-crossing`
   - : Represents the range where the principal box crosses the start border edge. The range start (0% progress) occurs when the start border edge of the element's principal box coincides with the start edge of its view progress visibility range. The range end (100% progress) is the point at which the end border edge of the element's principal box coincides with the start edge of its view progress visibility range. The size of the range is the size of the element's principle box in the scroll direction.
+
 - `scroll`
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/reference/values/timeline-range-name/index.md
```diff
@@ -38,10 +38,12 @@ Valid `<timeline-range-name>` values:
 
 - `exit-crossing`
   - : Represents the range where the principal box crosses the start border edge. The range start (0% progress) occurs when the start border edge of the element's principal box coincides with the start edge of its view progress visibility range. The range end (100% progress) is the point at which the end border edge of the element's principal box coincides with the start edge of its view progress visibility range. The size of the range is the size of the element's principle box in the scroll direction.
+- `scroll`
+  - : Represents the full range of the {{glossary("scroll container")}} on which the view progress timeline is defined. The range start (0% progress) occurs at the startmost scroll position of the scroll container underlying the view progress timeline. The range end (100% progress) is the endmost scroll position of the scroll container underlying the view progress timeline.
```
hamishwillee: @estelle This isn't wrong, but startmost and endmost are odd/spec terms that appear nowhere else in MDN. They aren't defined except in terms of progress. Anyway, how about this:

```suggestion

- `scroll`
  - : Represents the full range of the {{glossary("scroll container")}} on which the view progress timeline is defined. The range start (0% progress) and end (100% progress) occur at the very start and end positions of the scroll container underlying the view progress timeline.
```

Note, suggestion adds a line of blankspace to match the other entries.

PR #43573, opened 2026-03-25
Add rgbToHSL and hslToRGB functions to using classes encapsulation code example 

BrettKulp:
…de example

### Description

The code example of encapsulation calls rgbToHSL and hslToRGB functions that are not implemented
### Motivation

The code example will not run without those functions implemented


Josh-Cena:
Thanks. The example is not designed to be executable, just for illustration, and as you can see these functions are so long that they are distracting.

PR #43575, opened 2026-03-25
Editorial review: Document local network access

chrisdavidmills:
### Description

Between versions 142 and 145, Chrome enables support for various features of the [Local Network Access spec](https://wicg.github.io/local-network-access/). See https://chromestatus.com/feature/5152728072060928 and https://chromestatus.com/feature/5068298146414592.

Specifically, these are:

- The `targetAddressSpace` `Request` property, and the equivalent `RequestInit` property.
- The `local-network-access` permission/`Permissions-Policy` directive.
- The more granular `local-network` and `loopback-network` permissions/`Permissions-Policy` directives.

This PR:

- Adds reference pages for the above features.
- Adds a new "Local network access" guide inside the MDN security section (I didn't add a regular API page for this, because it doesn't really feel like it warrants an API page — it is just a single property and some new permissions, in terms of web developer footprint.)
- Adds links to the guide from several related places.


files/en-us/web/security/defenses/local_network_access/index.md
```diff
@@ -0,0 +1,190 @@
...
+- A website connecting to a local site (such as an intranet) to provide authentication information, or as part of test environment.
+
+Allowing local network access comes with risks — attackers can use such requests to initiate [cross-site request forgery (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF) attacks against local network devices such as routers and printers, and fingerprint the user's local network. Local network access mitigates these risks, controlling local network access via a set of permission prompts and {{httpheader("Permissions-Policy")}} directives.
+
+## Address spaces
+
+Local network access defines three different **address spaces**, which all network addresses are categorised under:
+
+- Local
+  - : A local address is only accessible on the local network: Its target will differ on different networks. For example, `http://192.168.0.1`.
```
christhompson:
```suggestion
  - : A local address is only accessible on the local network: Its target will differ on different networks. For example, `192.168.0.1`.
```

I think better to keep these as just IP addresses, rather than URLs.

files/en-us/web/security/defenses/local_network_access/index.md
```diff
@@ -0,0 +1,190 @@
...
+- Loopback
+  - : A loopback address is only accessible on the local device: Its target will differ on every device. For example, `127.0.0.1`, which is generally known as `localhost`.
+- Public
+  - : A public web address is available from anywhere on the internet: Its target is the same for all devices globally on the IP network. For example, `https://example.com`.
```
christhompson:
```suggestion
  - : A public address is available from anywhere on the internet: Its target is the same for all devices globally on the Internet. For example, `104.18.27.120` (the IP address of `example.com`).
```

Thinking about how to make this more consistent with the other address spaces... Maybe remove "web" as this is a lower layer concept (don't want to confuse with URLs). Swap domain name with the public IP address it resolves to.

files/en-us/web/security/defenses/local_network_access/index.md
```diff
@@ -0,0 +1,190 @@
...
+  method: "get",
+  mode: "cors",
+  targetAddressSpace: "loopback",
+});
+
+console.log(req.targetAddressSpace);
+// loopback
+```
+
+This is needed in cases where a URL is a public domain address, but ends up resolving to a local network address, such as `http://example.com`.
```
christhompson: Do we want this example to be a real world case of a "public domain name that resolves to a local IP address"? One such case is `routerlogin.net` when the user is behind an Netgear router; another is plex.direct URLs (like `192-168-0-7.###.plex.direct`). Or we could still use the example domain but make it a bit more obviously a non-real-world example like `internal.example.com`?
```suggestion
This is needed in cases where a URL is a public domain address, but ends up resolving to a local network address, such as `http://internal.example.com`.
```

files/en-us/web/security/defenses/local_network_access/index.md
```diff
@@ -0,0 +1,190 @@
...
+  allow="local-network https://example.com https://example2.com"></iframe>
+```
+
+You could also specify `allow="local-network *"` to allow any origin loaded in the `<iframe>` to make local network requests. For example, this can be useful in cases where an `<iframe>` makes arbitrary redirects to another origin (such as for SSO) before redirecting back to `localhost`.
+
+## The `local-network-access` alias
+
+Local network access was originally implemented with the {{httpheader('Permissions-Policy/local-network-access','local-network-access')}} permission, which was used to control network requests to both local and loopback addresses together. This was updated to the more granular {{httpheader('Permissions-Policy/local-network','local-network')}} and {{httpheader('Permissions-Policy/loopback-network','loopback-network')}} permissions, which you should use going forward.
+
+However, the `local-network-access` permission continues to be supported for backwards-compatibility, as an alias of `local-network` and `loopback-network`.
```
christhompson:
```suggestion
However, the `local-network-access` permission continues to be supported for backwards-compatibility in Chromium, as an alias of `local-network` and `loopback-network`.
```

The alias is only supported in Chromium. Firefox prefers to not need to support it, and we will hopefully be able to remove it entirely 🤞

files/en-us/web/http/reference/headers/permissions-policy/index.md
```diff
@@ -149,6 +149,15 @@ You can specify
 - {{httpheader('Permissions-Policy/local-fonts','local-fonts')}} {{Experimental_Inline}}
   - : Controls whether the current document is allowed to gather data on the user's locally-installed fonts via the {{DOMxRef("Window.queryLocalFonts()")}} method (see also the {{domxref("Local Font Access API", "Local Font Access API", "", "nocode")}}).
 
+- {{httpheader('Permissions-Policy/local-network','local-network')}} {{Experimental_Inline}}
+  - : Controls whether the current document is allowed to make network requests to local addresses.
+
+- {{httpheader('Permissions-Policy/local-network-access','local-network-access')}} {{Experimental_Inline}}
+  - : Controls whether the current document is allowed to make network requests to local and loopback addresses. This policy directive is an alias for the more granular `local-network` and `loopback-network` directives.
```
christhompson, line 179:
```suggestion
  - : Controls whether the current document is allowed to make network requests to local and loopback addresses. This policy directive is an alias for the more granular `local-network` and `loopback-network` directives. Only supported in Chromium browsers.
```

christhompson: Similarly as to the main article, the alias is only supported by Chromium (and we hope to remove it eventually).

files/en-us/web/http/reference/headers/permissions-policy/local-network/index.md
```diff
@@ -0,0 +1,47 @@
...
+  - experimental
+browser-compat: http.headers.Permissions-Policy.local-network
+sidebar: http
+---
+
+{{SeeCompatTable}}
+
+The HTTP {{HTTPHeader("Permissions-Policy")}} header `local-network` directive controls whether the current document is allowed to make network requests to local addresses.
+
+A local address is only accessible on the local network: Its target will differ on different networks. For example, `http://192.168.0.1`.
```
christhompson:
```suggestion
A local address is only accessible on the local network: Its target will differ on different networks. For example, `192.168.0.1`.
```

Same as in main article, prefer using IP address form over URL form I think

files/en-us/web/http/reference/headers/permissions-policy/local-network-access/index.md
```diff
@@ -0,0 +1,48 @@
...
+page-type: http-permissions-policy-directive
+status:
+  - experimental
+browser-compat: http.headers.Permissions-Policy.local-network-access
+sidebar: http
+---
+
+{{SeeCompatTable}}
+
+The HTTP {{HTTPHeader("Permissions-Policy")}} header `local-network-access` directive controls whether the current document is allowed to make network requests to local and loopback addresses. This policy directive is an alias for the newer {{httpheader('Permissions-Policy/local-network','local-network')}} and {{httpheader('Permissions-Policy/loopback-network','loopback-network')}} directives.
```
christhompson, line 14:
```suggestion
The HTTP {{HTTPHeader("Permissions-Policy")}} header `local-network-access` directive controls whether the current document is allowed to make network requests to local and loopback addresses. This policy directive is an alias for the newer {{httpheader('Permissions-Policy/local-network','local-network')}} and {{httpheader('Permissions-Policy/loopback-network','loopback-network')}} directives. It is only supported in Chromium browsers.
```

files/en-us/web/http/reference/headers/permissions-policy/local-network-access/index.md
```diff
@@ -0,0 +1,48 @@
...
+
+- A local address is only accessible on the local network: Its target will differ on different networks. For example, `http://192.168.0.1`.
```
christhompson:
```suggestion
- A local address is only accessible on the local network: Its target will differ on different networks. For example, `192.168.0.1`.
```

files/en-us/web/api/requestinit/index.md
```diff
@@ -214,6 +214,16 @@ You can also construct a `Request` with a `RequestInit`, and pass the `Request`
   - : A string that sets a policy for the {{httpheader("Referer")}} header. The syntax and semantics of this option are exactly the same as for the {{httpheader("Referrer-Policy")}} header.
 - `signal` {{optional_inline}}
   - : An {{domxref("AbortSignal")}}. If this option is set, the request can be canceled by calling {{domxref("AbortController.abort()", "abort()")}} on the corresponding `AbortController`.
+- `targetAddressSpace` {{optional_inline}}
+  - : An enumerated value that specifies whether the request is intended to be a loopback, local, or public request. Valid values are:
+    - `local`
+      - : The request URL is a local address, which is only accessible on the local network: Its target will differ on different networks. For example, `http://192.168.0.1`.
+    - `loopback`
+      - : The request URL is a loopback address, which is only accessible on the local device: Its target will differ on every device. For example, `127.0.0.1`, which is generally known as `localhost`.
```
christhompson:
```suggestion
      - : The request URL is a loopback address, which is only accessible on the local device: Its target will differ on every device. For example, `http://127.0.0.1`, which is generally known as `localhost`.
```

christhompson: In this context we are talking about URLs explicitly, so let's make all of these be fully specified URLs for consistency?

files/en-us/web/api/request/targetaddressspace/index.md
```diff
@@ -0,0 +1,65 @@
...
+The read-only **`targetAddressSpace`** property of the {{DOMxRef("Request")}} interface returns the request's target address space, which indicates whether it is intended to be a loopback, local, or public request.
+
+## Value
+
+An enumerated value, which can be one of:
+
+- `local`
+  - : The request URL is a local address, which is only accessible on the local network: iIts target will differ on different networks. For example, `http://192.168.0.1`.
+- `loopback`
+  - : The request URL is a loopback address, which is only accessible on the local device: Its target will differ on every device. For example, `127.0.0.1`, which is generally known as `localhost`.
```
christhompson:
```suggestion
  - : The request URL is a loopback address, which is only accessible on the local device: Its target will differ on every device. For example, `http://127.0.0.1`, which is generally known as `localhost`.
```

files/en-us/web/security/defenses/local_network_access/index.md
chrisdavidmills: OK. I've fixed all these across the PR in my next commit.

files/en-us/web/security/defenses/local_network_access/index.md
chrisdavidmills: Yeah, I like this. I've tweaked it slightly, to:

A public address is available from anywhere on the internet; its target is the same for all devices globally. For example, `104.18.27.120` (the IP address of `example.com`).

I'll go through and make sure similar fixes are made across the PR.

files/en-us/web/security/defenses/local_network_access/index.md
chrisdavidmills: I've updated it to `http://internal.example.com` — we tend to avoid using "real" examples that we don't control, with `example.com` being the exception.

files/en-us/web/security/defenses/local_network_access/index.md
chrisdavidmills: Makes sense. I've tweaked it to the following:

> However, the `local-network-access` permission continues to be supported for backwards-compatibility [where already implemented](/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy/local-network-access#browser_compatibility), as an alias of `local-network` and `loopback-network`.

Does that work for you?

We're strict about not including specific browser support info in content (rather than in the browser compat data), but this should be allowed. There is no compat data at that link yet, but this will change when https://github.com/mdn/browser-compat-data/pull/29331 is merged.

Once you remove this permission, we should be able to delete this section and mark the compat data with the removal version, and we should be all set.

files/en-us/web/http/reference/headers/permissions-policy/index.md
chrisdavidmills: See above comment — readers can look at the compat data for the directive to see the support story.

files/en-us/web/http/reference/headers/permissions-policy/local-network/index.md
chrisdavidmills: updated

files/en-us/web/http/reference/headers/permissions-policy/local-network-access/index.md
chrisdavidmills: Again, see above comments.

files/en-us/web/http/reference/headers/permissions-policy/local-network-access/index.md
chrisdavidmills: updated

files/en-us/web/api/requestinit/index.md
chrisdavidmills: In light of your previous comments, I ended up changing these to talk about addresses rather than URLs (and on the `targetAddressSpace` property page). Have a look at the updated pages after my next commit and see what you think. I can happily change them again, if you think we'd be better off talking specifically about URLs here.

files/en-us/web/api/request/targetaddressspace/index.md
chrisdavidmills: See above comment

files/en-us/web/security/defenses/local_network_access/index.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Background
+
+There are a number of scenarios in which a site might want to make a request to a local network. For example:
+
+- A website accessing a local device to retrieve diagnostic information when troubleshooting it.
+- A website accessing a local device to provide configuration details during a setup procedure.
+- A website connecting to a local site (such as an intranet) to provide authentication information, or as part of test environment.
+
+Allowing local network access comes with risks — attackers can use such requests to initiate [cross-site request forgery (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF) attacks against local network devices such as routers and printers, and fingerprint the user's local network. Local network access mitigates these risks, controlling local network access via a set of permission prompts and {{httpheader("Permissions-Policy")}} directives.
```
wbamberg: I'd love a bit more of an unpacking of this. How does the CSRF attack work, and how are existing CSRF defenses not adequate to protect us?

chrisdavidmills: @christhompson, what would you advise adding as an answer to this? I agree that a bit more info would be good, although we don't want it to take over the article.

chrisdavidmills: @christhompson, @hubertchao, apologies, I let this content PR slip through the cracks. 

Can you provide me with an answer to the above question? In the meantime, I'll look at the BCD comments over at https://github.com/mdn/browser-compat-data/pull/29331#issuecomment-4126475451 and see if/how they affect the content.

Thanks!

chrisdavidmills: OK, I've made the above check, and it doesn't look like the content needs any changes due to the linked comments.

christhompson: Sorry for missing this until now. Server-side CSRF defenses are also important, but many local devices don't expect to be subject to malicious traffic and there is a long history of vulnerable devices (many that are still running, unpatched). We call out the necessity of fixing this on both sides in https://wicg.github.io/local-network-access/#scope-of-mitigation

chrisdavidmills: Thanks for the info, @christhompson.

@wbamberg, I've added some information about the types of CSRF attacks this mitigates.

files/en-us/web/api/request/targetaddressspace/index.md
```diff
@@ -0,0 +1,65 @@
...
+title: "Request: targetAddressSpace property"
+short-title: targetAddressSpace
+slug: Web/API/Request/targetAddressSpace
+page-type: web-api-instance-property
+browser-compat: api.Request.targetAddressSpace
+---
+
+{{APIRef("Fetch API")}}{{AvailableInWorkers}}
+
+The read-only **`targetAddressSpace`** property of the {{DOMxRef("Request")}} interface returns the request's target address space, which indicates whether it is intended to be a loopback, local, or public request.
```
hamishwillee:
```suggestion
The **`targetAddressSpace`** read-only property of the {{DOMxRef("Request")}} interface returns the request's target address space, which indicates whether it is intended to be a loopback, local, or public request.
```

files/en-us/web/api/request/targetaddressspace/index.md
```diff
@@ -0,0 +1,65 @@
...
+- `unknown`
+  - : No `targetAddressSpace` was set on the request.
+
+## Description
+
+In browsers that support [Local Network Access](/en-US/docs/Web/Security/Defenses/Local_network_access), loopback and local requests are gated behind specific permissions — `local-network` and `loopback-network`, respectively. Requiring user permissions for such requests mitigates the risk of [cross-site request forgery (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF) attacks against local network devices such as routers and printers, and reduces the ability of sites to use these requests to fingerprint the user's local network.
+
+This permission is restricted to secure contexts. If granted, the permissions additionally relax [mixed content](/en-US/docs/Web/Security/Defenses/Mixed_content) blocking for local network requests. This is useful because many local devices are not able to obtain publicly trusted TLS certificates for various reasons. For example, it allows public websites to access local testing servers or devices running on HTTP.
```
hamishwillee:
```suggestion
These permissions are restricted to secure contexts. If granted, the permissions additionally relax [mixed content](/en-US/docs/Web/Security/Defenses/Mixed_content) blocking for local network requests. This is needed because many local devices are not able to obtain publicly trusted TLS certificates for various reasons. For example, it allows public websites to access local testing servers or devices running on HTTP.
```

files/en-us/web/api/requestinit/index.md
```diff
@@ -214,6 +214,16 @@ You can also construct a `Request` with a `RequestInit`, and pass the `Request`
   - : A string that sets a policy for the {{httpheader("Referer")}} header. The syntax and semantics of this option are exactly the same as for the {{httpheader("Referrer-Policy")}} header.
 - `signal` {{optional_inline}}
   - : An {{domxref("AbortSignal")}}. If this option is set, the request can be canceled by calling {{domxref("AbortController.abort()", "abort()")}} on the corresponding `AbortController`.
+- `targetAddressSpace` {{optional_inline}}
+  - : An enumerated value that specifies whether the request is intended to be a loopback, local, or public request. Valid values are:
+    - `local`
+      - : The request is to a local address, which is only accessible on the local network; its target will differ on different networks. For example, `192.168.0.1`.
```
hamishwillee, line 221: The explainer also mentions `.local hostnames`. If these are supported we should mention them somewhere.

files/en-us/web/security/defenses/local_network_access/index.md
```diff
@@ -0,0 +1,196 @@
...
+    adviseUserOnRemediation();
+  }
+});
+```
+
+The permissions are restricted to secure contexts. On non-secure contexts, all requests will fail.
+
+## Relaxing mixed content
+
+If granted, the permissions additionally relax [mixed content](/en-US/docs/Web/Security/Defenses/Mixed_content) blocking for local network requests. This is useful because many local devices are not able to obtain publicly trusted TLS certificates for various reasons. For example, it allows public websites to access local testing servers or devices running on HTTP.
```
hamishwillee: "This is useful" > "this is needed"?

files/en-us/web/api/requestinit/index.md
```diff
@@ -214,6 +214,16 @@ You can also construct a `Request` with a `RequestInit`, and pass the `Request`
   - : A string that sets a policy for the {{httpheader("Referer")}} header. The syntax and semantics of this option are exactly the same as for the {{httpheader("Referrer-Policy")}} header.
 - `signal` {{optional_inline}}
   - : An {{domxref("AbortSignal")}}. If this option is set, the request can be canceled by calling {{domxref("AbortController.abort()", "abort()")}} on the corresponding `AbortController`.
+- `targetAddressSpace` {{optional_inline}}
+  - : An enumerated value that specifies whether the request is intended to be a loopback, local, or public request. Valid values are:
```
hamishwillee: It might be worth saying here ...


```suggestion
  - : An enumerated value that specifies whether the request is intended to be a loopback, local, or public request. This controls how the user agent will handle mixed-content.
  Valid values are:
```

files/en-us/web/security/defenses/local_network_access/index.md
```diff
@@ -0,0 +1,196 @@
...
+  method: "get",
+  mode: "cors",
+  targetAddressSpace: "loopback",
+});
+
+console.log(req.targetAddressSpace);
+// loopback
+```
+
+This is needed in cases where a URL is a public domain address, but ends up resolving to a local network address, such as `http://internal.example.com`.
```
hamishwillee: I'm not sure this is quite right, or perhaps that it is fully explained. 

As I understand it, if you were not to specify `targetAddressSpace: "loopback",` you might still be able to access the loopback address? Or will you be blocked?
But assuming you can and you get permissions, if you don't specify this then the mixed-content checks are not disabled. That means that an HTTPS site can't load an HTTP resource.

So what you're really saying here is that we need to specify the loopback or if we end up accessing an HTTP address everything will fall over. Right - or I am missing something?

The point being that it is not clear what "This" in "This is needed", or whether the access will still try happen if you don't put in `targetAddressSpace`. But it seesm likely that if it does happen the result if you omit this is that  the request will be blocked as missed content.

files/en-us/web/security/defenses/local_network_access/index.md
```diff
@@ -0,0 +1,196 @@
...
+
+For example, if `https://example.com` navigated the `<iframe>` to `https://example2.com`, and both `https://example.com` and `https://example2.com` made local network access requests, you would need to explicitly delegate the permission to both origins as follows:
+
+```html
+<iframe
+  src="https://example.com"
+  allow="local-network https://example.com https://example2.com"></iframe>
+```
+
+You could also specify `allow="local-network *"` to allow any origin loaded in the `<iframe>` to make local network requests. For example, this can be useful in cases where an `<iframe>` makes arbitrary redirects to another origin (such as for SSO) before redirecting back to `localhost`.
```
hamishwillee:
```suggestion
You could also specify `allow="local-network *"` to allow any origin loaded in the `<iframe>` to make local network requests. This can be useful in cases where an `<iframe>` makes arbitrary redirects to another origin (such as for SSO) before redirecting back to `localhost`.
```

files/en-us/web/security/defenses/local_network_access/index.md
```diff
@@ -0,0 +1,196 @@
...
+
+## The `local-network-access` alias
```
hamishwillee, line 129: Maybe 


```suggestion
### The `local-network-access` alias
```

hamishwillee: Ignore that. There's quite a lot to say here.

files/en-us/web/http/reference/headers/permissions-policy/local-network-access/index.md
```diff
@@ -0,0 +1,48 @@
...
+{{SeeCompatTable}}
+
+The HTTP {{HTTPHeader("Permissions-Policy")}} header `local-network-access` directive controls whether the current document is allowed to make network requests to local and loopback addresses. This policy directive is an alias for the newer {{httpheader('Permissions-Policy/local-network','local-network')}} and {{httpheader('Permissions-Policy/loopback-network','loopback-network')}} directives.
+
+- A local address is only accessible on the local network; its target will differ on different networks. For example, `192.168.0.1`.
+- A loopback address is only accessible on the local host; its target will differ on every device. For example, `127.0.0.1`, which is generally known as `localhost`.
+
+Specifically, where a defined policy blocks use of this feature, requests to local and loopback addresses will always fail. If you want more granular control over local and loopback addresses, you should use the newer directives linked above.
+
+See [Local Network Access](/en-US/docs/Web/Security/Defenses/Local_network_access) for more details.
```
hamishwillee: Link to the specific section

files/en-us/web/http/reference/headers/permissions-policy/local-network/index.md
```diff
@@ -0,0 +1,47 @@
...
+```http
+Permissions-Policy: local-network=<allowlist>;
+```
+
+- `<allowlist>`
+  - : A list of origins for which permission is granted to use the feature. See [`Permissions-Policy` > Syntax](/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy#syntax) for more details.
+
+## Default policy
+
+The default allowlist for `local-network` is `self`.
```
hamishwillee: Would be really good to duplicate pattern of examples like https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy/bluetooth#examples

files/en-us/web/http/reference/headers/permissions-policy/loopback-network/index.md
```diff
@@ -0,0 +1,47 @@
...
+Permissions-Policy: loopback-network=<allowlist>;
+```
+
+- `<allowlist>`
+  - : A list of origins for which permission is granted to use the feature. See [`Permissions-Policy` > Syntax](/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy#syntax) for more details.
+
+## Default policy
+
+The default allowlist for `loopback-network` is `self`.
+
```
hamishwillee, line 63: Would be really good to duplicate pattern of examples like https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy/bluetooth#examples

files/en-us/web/http/reference/headers/permissions-policy/local-network-access/index.md
hamishwillee: Or even perhaps pull all that into this doc in a description? On one hand that makes sense, because we hope people won't use this, and pulling the information here makes it less visible in the main overview. On the other hand, a lot of that material is about the other permissions. If it were me I think I would, but for you it's a suggestion.

files/en-us/web/http/reference/headers/permissions-policy/local-network-access/index.md
```diff
@@ -0,0 +1,48 @@
...
+Permissions-Policy: local-network-access=<allowlist>;
+```
+
+- `<allowlist>`
+  - : A list of origins for which permission is granted to use the feature. See [`Permissions-Policy` > Syntax](/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy#syntax) for more details.
+
+## Default policy
+
+The default allowlist for `local-network-access` is `self`.
+
```
hamishwillee, line 35: Would be somewhat good to duplicate pattern of examples like https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Permissions-Policy/bluetooth#examples

I'm saying somewhat here and not "Definitely" as in the other cases, because this one isn't one we really want people to use.

hamishwillee (commented):
I think it makes sense. A few minor things.
The biggest is this one https://github.com/mdn/content/pull/43575#discussion_r3331426190 - it isn't clear exactly what happens if I do/do not set the targetAddressSpace

Do we know the standards position of other vendors on this?

files/en-us/web/api/request/targetaddressspace/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+The `targetAddressSpace` property, when set to `local` or `loopback` on new requests (via the {{domxref("Request.Request", "Request()")}} constructor or directly in the {{domxref("Window.fetch", "fetch()")}} method), explicitly tells supporting browsers to skip mixed content checks. This is needed in cases where a URL is a public domain address, but ends up resolving to a local network address, such as `http://internal.example.com`.
+
+Some addresses such as private IP literals (for example, `192.168.0.1`) and `local` addresses (such as `http://router.local`) are exempt from mixed content checks, so don't need the `targetAddressSpace` property to be set.
```
hubertchao: This could be misconstrued. These IP's aren't exempt from mixed content checks, its that if a request to a private IP literal is considered an LNA request with permission, the mixed content requirement will be relaxted.

If LNA isn't relevant, then the mixed content check still applies (e.g. https://192.168.1.1 tried to use a resource from http://192.168.1.12 would still be blocked by mixed content )

files/en-us/web/security/defenses/local_network_access/index.md
hubertchao: > So what you're really saying here is that we need to specify the loopback or if we end up accessing an HTTP address everything will fall over. Right - or I am missing something?

Correct. Well, you'd need to specify a `targetAddressSpace: loopback` or `targetAddressSpace: local` depending on where the resource is served from. 

See: https://wicg.github.io/local-network-access/#mixed-content-lna

files/en-us/web/api/requestinit/index.md
chrisdavidmills: These are mentioned in the guide, and in the `Request.targetAddressSpace` ref page:

> Some addresses such as private IP literals (for example, `192.168.0.1`) and `local` addresses (such as `http://router.local`) are exempt from mixed content checks, so don't need the `targetAddressSpace` property to be set.

files/en-us/web/security/defenses/local_network_access/index.md
chrisdavidmills: Updated

files/en-us/web/security/defenses/local_network_access/index.md
chrisdavidmills: I think this makes sense to me now. I've updated that paragraph to:

> Setting `targetAddressSpace: "loopback"` is required in cases where a URL is a public domain address, but ends up resolving to a local network address, such as `http://internal.example.com`. If this is omitted, a request from an HTTPS site to access the HTTP resource will fail mixed content checks.

Does that work?

files/en-us/web/http/reference/headers/permissions-policy/local-network-access/index.md
chrisdavidmills: I thought about this a bit and decided to link to the specific section. I think removing that would make the structure a bit weird, although, granted, I can see your logic.

files/en-us/web/http/reference/headers/permissions-policy/local-network-access/index.md
chrisdavidmills: Yeah, I'm just going to leave this one.

files/en-us/web/http/reference/headers/permissions-policy/local-network/index.md
chrisdavidmills: done

files/en-us/web/http/reference/headers/permissions-policy/loopback-network/index.md
chrisdavidmills: done

chrisdavidmills:
> Do we know the standards position of other vendors on this?

For Mozilla, the standards position for private network access (the precursor to this) was positive: https://mozilla.github.io/standards-positions/#cors-and-rfc1918

For Safari, it is positive also: https://webkit.org/standards-positions/#position-163

files/en-us/web/api/request/targetaddressspace/index.md
chrisdavidmills: OK, I've tried rewriting this as :

> Some addresses, such as private IP literals (for example, `192.168.0.1`) and `.local` addresses (such as `http://router.local`), have mixed content checks relaxed in cases where local network access permissions are granted, so don't need the `targetAddressSpace` property to be set. However, if local network access permissions are not granted, mixed content checks still apply, regardless of whether `targetAddressSpace` is set or not.

Does that make sense?

hubertchao: Yep, that makes sense!

files/en-us/web/security/defenses/local_network_access/index.md
hamishwillee: It definitely captures the loopback case. 

FWIW I suspect that there is a similar explanation for local case.

files/en-us/web/api/request/targetaddressspace/index.md
```diff
@@ -0,0 +1,65 @@
...
+- `loopback`
+  - : The request is to a loopback address, which is only accessible on the local device; its target will differ on every device. For example, `127.0.0.1`, which is generally known as `localhost`.
+- `public`
+  - : The request is to an address available from anywhere on the internet; its target is the same for all devices globally. For example, `104.18.27.120` (the IP address of `example.com`)
+- `unknown`
+  - : No `targetAddressSpace` was set on the request.
+
+## Description
+
+In browsers that support [Local Network Access](/en-US/docs/Web/Security/Defenses/Local_network_access), loopback and local requests are gated behind specific permissions — `local-network` and `loopback-network`, respectively. Requiring user permissions for such requests mitigates the risk of [cross-site request forgery (CSRF)](/en-US/docs/Web/Security/Attacks/CSRF) attacks against local network devices such as routers and printers, and reduces the ability of sites to use these requests to fingerprint the user's local network.
```
hamishwillee: This just caught my eye:

> In browsers that support [Local Network Access](/en-US/docs/Web/Security/Defenses/Local_network_access), loopback and local requests are gated behind specific permissions

Here "Local Network Access" is a specification, and it is correct. 
But in English it reads as "In browsers that let you access the local network" - and that is very wrong, because it would mean LNA was implemented everywhere.

Something to think about..

hamishwillee (approved):
@chrisdavidmills This looks good to me. Thank you.  Just one nit to consider https://github.com/mdn/content/pull/43575/changes#r3359739157

files/en-us/web/api/request/targetaddressspace/index.md
chrisdavidmills: I totally agree that this could create a bit of confusion. I've gone through the PR and updated this instance and ones like it.

PR #43579, opened 2026-03-25
Node.js server without a framework: decodeURI

ratijas:
### Description

In general case, decoding URI is needed to transform it to a usable file path without %percent encoded whitespaces etc.


ratijas:
Might also wanna strip ?query= params as well. Those are sometimes used with a dummy value just to bypass caching.

hamishwillee (approved):
Thanks @ratijas - and congrats on landing your first PR merge.

> Might also wanna strip ?query= params as well. Those are sometimes used with a dummy value just to bypass caching.

We could, but I don't think it is necessary. Would consider a PR, but not sure it would add much as a "teaching opportunity".

PR #43581, opened 2026-03-26
Changes example in CSS invert() to use appropriate classes.

elooly10:
### Description

Applies the css invert to the item in the example labelled `invert() filter` instead of the SVG invert for the `SVG filter equivalent`, and fixes the SVG filter equivalent with a class as the document's text explains rather than as an inline SVG.

### Motivation

The changes keep the content accurate and make the playground more useable.

### DIFF
--- a/files/en-us/web/css/reference/values/filter-function/invert/index.md
+++ b/files/en-us/web/css/reference/values/filter-function/invert/index.md
@@ -130,6 +130,7 @@ td {
   padding: 5px;
 }
 ```
+
 ```html hidden
 <table>
   <thead>
--- a/files/en-us/web/css/reference/values/filter-function/invert/index.md
+++ b/files/en-us/web/css/reference/values/filter-function/invert/index.md
@@ -151,7 +151,7 @@ td {
       <td>
         <img
           class="svgFilter"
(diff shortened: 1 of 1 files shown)


files/en-us/web/css/reference/values/filter-function/invert/index.md
```diff
@@ -144,24 +144,15 @@ td {
     <tr>
       <td>
         <img
-          class="svgFilter"
+          class="filter"
```
chrisdavidmills, line 147: This is the only change that is needed to make the fix that you've proposed. The other changes below have broken the SVG equivalent version — it no longer displays properly.

chrisdavidmills (changes requested):
Hi @elooly10. What you are proposing to fix makes sense, but you've gone too far and broken the SVG equivalent — see the comment below.

Josh-Cena:
@chrisdavidmills There's something obviously non-sensical about this example because the `svgFilter` class isn't even used. In fact, all these CSS `filter-function` pages are written so inconsistently I don't even know what the intended approach is: SVG inlined in the table, or as a separate markup. I suggest we pick a "canonical page structure" for all these pages (such as `hue-rotate()`) and then update all other pages.

PR #43583, opened 2026-03-26
Document Wasm SIMD arithmetic instructions

chrisdavidmills:
### Description

This PR adds reference docs for WebAssembly SIMD arithmetic instructions (and possibly some more bitwise ones too).

Note: I've updated the general `shl` page to cover SIMD and non-SIMD values. I've also deleted the previously published SIMD-specific `shl` page and redirected it to the general one.


files/en-us/webassembly/reference/numeric/abs/index.md
```diff
@@ -31,17 +31,107 @@ await WebAssembly.instantiateStreaming(fetch(url), { console });
...
+```plain
+[input] -> [output]
+```
+
+- `input`
+  - : The input value.
+- `output`
+  - : The output value.
+
+For a non-SIMD `abs`, the `input` and `output` will be basic numeric values such as `-3.5` or `3.5`.
```
hamishwillee: This reads as though the output can be -3.5.

files/en-us/webassembly/reference/numeric/abs/index.md
```diff
@@ -31,17 +31,107 @@ await WebAssembly.instantiateStreaming(fetch(url), { console });
...
+
+## Examples
+
+### SIMD abs
+
+In this example, we demonstrate running the `abs` instruction on a SIMD value and outputting one of the resulting lane values.
+
+#### JavaScript
+
+In our script, we grab a reference to a {{htmlelement("p")}} element that we will output our result to, then define an object for import into Wasm containing a single function that writes a value to the output `<p>`. We then compile and instantiate our Wasm module using the [`WebAssembly.instantiateStreaming()`](/en-US/docs/WebAssembly/Reference/JavaScript_interface/instantiateStreaming_static) method, importing the object in the process.
```
hamishwillee, line 88: Any reason you chose "in our" rather than "in the following", "in the", or "in this"? Feels inconsistent/odd.

files/en-us/webassembly/reference/numeric/abs/index.md
```diff
@@ -31,17 +31,107 @@ await WebAssembly.instantiateStreaming(fetch(url), { console });
...
+```
+
+- `input`
+  - : The input value.
+- `output`
+  - : The output value.
+
+For a non-SIMD `abs`, the `input` and `output` will be basic numeric values such as `-3.5` or `3.5`.
+
+For a [SIMD](/en-US/docs/WebAssembly/Reference/SIMD) `abs`, the `input` and `output` will be [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretations, for example `i32x4 4 -8 12 -16`.
```
hamishwillee: Ditto. So for a SIMD all the values will be made absolute after the instruction?

files/en-us/webassembly/reference/numeric/add/index.md
```diff
@@ -32,20 +32,119 @@ await WebAssembly.instantiateStreaming(fetch(url), { console });
...
+- `input1`
+  - : The first value to be added.
+- `input2`
+  - : The second value to be added.
+- `output`
+  - : The sum of the two input values.
+
+For a non-SIMD `add`, these will be basic numeric values such as `3` or `3.5`.
+
+For a [SIMD](/en-US/docs/WebAssembly/Reference/SIMD) `add`, these will be [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretations, for example `f32x4 0x9 0xa 0xb 0xc`.
```
hamishwillee: I think you need to say what numbers are added in the SIMD case since there are multiple lanes in the input and the description is "is used for adding up two numbers".
Also in this case, where does the value "go"? Is it written into the v128 value or onto the stack?

hamishwillee: Ah, this is answered by the example. I think it is "For SIMD this adds two v128 values of the same interpretation type, returning another v128 on the stack where each lane is the addition of the corresponding lanes in the original blah blha.

It isn't clear if they have to be the same type of data, though I would assume so.

We should capture all this here.

hamishwillee: And similar for all the other instructions

files/en-us/webassembly/reference/numeric/and/index.md
```diff
@@ -44,18 +44,105 @@ function numToBin(num) {
...
+- `input1`
+  - : The first input value.
+- `input2`
+  - : The second input value.
+- `output`
+  - : The output value.
+
+For a non-SIMD `and`, these will be basic numeric values such as `14` or `3`.
+
+For a [SIMD](/en-US/docs/WebAssembly/Reference/SIMD) `and`, these will be [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretations, for example `i32x4 9 4 -16 100`.
```
hamishwillee: As above https://github.com/mdn/content/pull/43583/changes#r3061408253 we need to make it clear if the types are the same and that the comparison is by corresponding lane, returning a new value to the stack of the same type.

files/en-us/webassembly/reference/numeric/ceil/index.md
hamishwillee: YOu changed this in all the other cases to instruction/is

hamishwillee (commented):
It is all very consistent. I haven't got the energy or time to review every case, so I have read a few carefully and added a few points for discussion.

files/en-us/webassembly/reference/simd/arithmetic/min_u/index.md
```diff
@@ -0,0 +1,73 @@
...
+```plain
+[input1, input2] -> [output]
+```
+
+- `input1`
+  - : The first input value.
+- `input2`
+  - : The second input value.
+- `output`
+  - : The output value. A new `v128` of the same type as the inputs, with each lane set to the greater of that lane index's value on the two inputs.
```
hamishwillee, line 60:
```suggestion
  - : The output value. A new `v128` of the same type as the inputs, with each lane set to the lower of that lane index's value on the two inputs.
```

hamishwillee: This was a claude check ^^^

Here is the rest of the comments:

  2. simd/bitwise/index.md — wrong URL for any_true

  - [`any_true`](/en-US/docs/WebAssembly/Reference/SIMD/bitwise/andnot)

  any_true links to the andnot page. Should be .../bitwise/any_true.

  ---

  Grammar errors

  3. simd/index.md — "This instructions"

  ▎ "This instructions in this section are SIMD-specific."

  Should be: "The instructions in this section are SIMD-specific."

  4. simd/bitwise/any_true/index.md — article before v128

  ▎ "tests whether an v128 input value contains any non-zero bits."

  Should be: "tests whether a v128 input value..." (v is a consonant sound).

  5. simd/bitwise/all_true/index.md — type section says input1 but signature has [input]
  6. 
```
  ### Type

  [input] -> [output]

  - `input1`
    - : The input `v128` value interpretation.
```

  The parameter name is input1 but the type signature uses input. Should be consistent — either input1/input2 pattern
  (if that's the convention, but all_true only takes one input) or just input. This is the same pattern issue as
  bitmask.

  6. simd/bitwise/bitmask/index.md — same input1 vs [input] inconsistency

  The type signature shows [input] -> [output] but the parameter description uses input1.

  ---

  Spelling errors

  7. simd/bitwise/bitmask/index.md — "equivent"

  ▎ "The example's output value is 4096, which is the decimal equivent of the above binary pattern."

  Should be: "equivalent"

  8. simd/bitwise/bitmask/index.md — duplicate word "means means"

  ▎ "while an MSB of 0 means means the value is less than 128."

  Should be: "while an MSB of 0 means the value is less than 128."

  ---

  Wrong section heading (content accuracy)

  9. simd/arithmetic/index.md — ## Comparison heading is misleading

  The section under ## Comparison lists max_s, max_u, min_s, min_u — these are arithmetic min/max instructions, not   comparison instructions (comparisons return booleans). A heading like ## Min/Max would be more accurate.

  ---

files/en-us/webassembly/reference/numeric/abs/index.md
chrisdavidmills: Fair point. I've removed the negative value in my next commit.

files/en-us/webassembly/reference/numeric/abs/index.md
chrisdavidmills: No reason at all. I don't think this is a big enough issue to worry about. It feels slightly more conversational than usual, but otherwise OK.

files/en-us/webassembly/reference/numeric/abs/index.md
chrisdavidmills: Yes. I changed this one to all positive values too.

files/en-us/webassembly/reference/numeric/add/index.md
chrisdavidmills: Some of this is a bit too obvious for the target audience. For example, if the "Type" section shows an output, you know that a value is pushed onto the stack as a result.

However, I think it is useful to explicitly describe the nature of each output in the case of SIMD versions of the instructions. I've gone through and added a descriptive sentence for each one, for example:

> For a [SIMD](/en-US/docs/WebAssembly/Reference/SIMD) `add`, these will be [`v128`](/en-US/docs/WebAssembly/Reference/Types/v128) value interpretations, for example `f32x4 0x9 0xa 0xb 0xc`. Each lane of the output pushed to the stack is the addition of the corresponding lanes in the input values.

eqrion:
Just commenting that I did a quick technical review of this as well. It's a lot so I didn't look at everything. I added some commits that Chris merged on to his branch.

files/en-us/webassembly/reference/numeric/and/index.md
chrisdavidmills: I've added an explanation, as per above. The types will always be the same in Wasm.

files/en-us/webassembly/reference/numeric/ceil/index.md
chrisdavidmills: Good catch; changed.

files/en-us/webassembly/reference/simd/arithmetic/min_u/index.md
chrisdavidmills: OK, cool, I've corrected all the ones that were not already fixed by @eqrion's PR.

files/en-us/webassembly/reference/numeric/abs/index.md
hamishwillee: I certainly wouldn't block on it, but I'd probably update it if I got this feedback (depending on the day). Up to you.

hamishwillee (approved):
Thanks @chrisdavidmills (and @eqrion !). Fixes look good to me. Approving so you can merge when you are ready.

PR #42853, opened 2026-01-21
New guide: timeline range names

estelle:
Creates a new guide
Links to the guide
added two guides to the sidebar
crosslinking

This guide is really about names.  I am currently working on another guide, which should be ready in a few weeks. ~~https://github.com/mdn/content/pull/42105~~ https://github.com/mdn/content/pull/43587 that covers insetting view timelines. I pulled out the named ranges because it was getting overwhelming.

PR #43587, opened 2026-03-26
New guide: animation timeline range insets

estelle:
adding a new guide. 
Other guide is https://github.com/mdn/content/pull/42853


dipikabh:
Hi @estelle, I'm off next week. Could you request someone else for a review?

files/en-us/web/css/guides/scroll-driven_animations/index.md
```diff
@@ -143,6 +143,8 @@ Scroll the element in the inline direction to see its background color change. S
   - : Scroll-driven animation timelines and creating scroll-driven animations.
 - [Timeline range names](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_names)
   - : The {{cssxref("timeline-range-name")}} data type: Understanding the various timeline range names.
+- [Insetting view progress timelines](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_insets)
+  - : Insetting animation attachment ranges of scroll driven animations.
```
chrisdavidmills:
```suggestion
  - : Insetting the animation attachment ranges of scroll-driven animations.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
+---
+title: Understanding timeline insets
+slug: Web/CSS/Guides/Scroll-driven_animations/Timeline_insets
+page-type: guide
+sidebar: cssref
+---
+
+By default, [view progress timelines](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timelines#view_progress_timelines) track elements across the entire [animation attachment range](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_names#the_animation_attachment_range), and the `0%` progress point is at the start and `100%` progress point is at the end of that range, respectively. The animation attachment range can be changed setting a [timeline range name](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/timeline_range_name), and the location of the `0%` and `100%` progress points can be inset along the range by setting length or percentage-based insets.
```
chrisdavidmills:
```suggestion
By default, [view progress timelines](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timelines#view_progress_timelines) track elements across the entire [animation attachment range](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_names#the_animation_attachment_range). The `0%` progress point is at the start of the range, while the `100%` progress point is at the end. The animation attachment range can be changed by setting a [timeline range name](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/timeline_range_names), and the location of the `0%` and `100%` progress points along the range can be adjusted by setting length or percentage-based inset values.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+The various timeline range names, their meanings, and how they are used are discussed in the [timeline range name guide](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/timeline_range_name). This guide explains limiting the animation timeline to a specific portion of the animation timeline range, effectively _insetting_ the animation timeline, using length or percentage values.
```
chrisdavidmills:
```suggestion
This guide explains how to limit the animation timeline to a specific portion of the animation timeline range using length or percentage inset values.
```

chrisdavidmills: I didn't feel like you needed to state the first line; you've already linked to the timeline range names guide in the previous paragraph.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+## Animation timelines: a primer
```
chrisdavidmills, line 12: It seems to me like quite a bit of this primer (and the section that follows it) is duplicated from https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_names#view_progress_timeline_primer.

Couldn't you just cut this stuff down a bit, and just add a note that says "if you are new to timeline range names and insets, be sure to read [Understanding timeline range names](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_names) before reading this article." or similar?

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+{{EmbedLiveSample("svg_view", "100%", "720")}}
+
+The yellow subject elements represents the position of the element when the `from` keyframe is applied, which is the animation range's `0%` progress mark. The red represents the location of the animated element relative to the scrollport when the `to` keyframe is applied, which is the end of the animation, or the `100%` progress mark. The grey represents the scrollport.
+
+By default, the element animates while it is "in view", but this default definition of "in view" may not fit your needs. Fortunately, animation range properties enable us to control which edges define the ends of the edges of the animation attachment range and then the start and end of that range by a {{cssxref("length")}} or a {{cssxref("percentage")}} relative to named or default timeline range.
+
+### Animation range properties
+
+The {{cssxref("animation-range")}} properties enable setting a named timeline range and then insetting animations from the animation attachment ranges using {{cssxref("length-percentage")}} inset values, such as `20%` or `100px`.
```
chrisdavidmills: In this section, I'd first explain that the default timeline range is `cover`, with start and end positions of 0% and 100%, and what that means. Then say you can adjust the named timeline range and its offsets using the `animation-range*` properties. 

Next, say that we've already covered setting these values to timeline range names in the previous article — this changes the timeline range used, but keeps the start and end as 0% and 100% of that range.

Finally, explain that these properties can also take `<length-percentage>` values, or a combination of timeline range name and `<length-percentage>`, and say what both of those choices do.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+The {{cssxref("animation-range-start")}} and {{cssxref("animation-range-end")}} properties, which can both be set using the `animation-range` shorthand, define an animation's attachment range, limiting the keyframe's active interval to that specific portion of the range.
+
+These properties specify the start and end of the animations's attachment range, shifting the start time and/or end of the animation, where the `0%` and `100%` keyframes occur when the {{cssxref("animation-iteration-count")}} is set or defaults to `1`. Both the start and end offsets are from the start of the attachment range.
```
chrisdavidmills: This paragraph is hard to parse.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+Setting an inset is similar to setting the {{cssxref("animation-delay")}} on time-based {{domxref("DocumentTimeline")}} animations. Just like the `animation-delay` property can delay the animation start until `500ms` after the animation is otherwise applied, the `animation-range-start` property can be used to delay a view progress timeline animation until `50px` of the subject are in view. The `animation-range-end` property can similarly end the animation early. We also have the `animation-range` shorthand property, which can both move the `from` animation progress point to a scroll position occurring after the default `0%` progress point and move the `to` animation progress point to a scroll position occurring before the default `100%` progress point.
```
chrisdavidmills:
```suggestion
Setting an inset is similar to setting the {{cssxref("animation-delay")}} on time-based {{domxref("DocumentTimeline")}} animations. Just like the `animation-delay` property can delay the animation start until `500ms` after the animation is otherwise applied, the `animation-range-start` property can be used to delay a view progress timeline animation until `50px` of the subject is in view. The `animation-range-end` property can similarly end the animation early. We also have the `animation-range` shorthand property, which can both move the `from` animation progress point to a scroll position occurring after the default `0%` progress point and the `to` animation progress point to a scroll position occurring before the default `100%` progress point.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+See the [timeline range name guide](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/timeline_range_name) to learn about the predefined named timeline ranges. The focus of this guide is insetting the starts and ends of the animation attachment range using {{cssxref("length-percentage")}} inset values, regardless of which animation timeline range is set.
```
chrisdavidmills:
```suggestion
See the [timeline range name guide](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/timeline_range_name) to learn about the different named timeline ranges. The focus of this guide is insetting the start and end of the animation attachment range using {{cssxref("length-percentage")}} inset values.
```

"predefined" implies that it is possible to create custom named timeline ranges. I'd stick to "different".

I'm also suggesting removing "regardless of which animation timeline range is set", because I don't think it is needed, and `<length-percentages>` values do have different effects depending on the named timeline range set.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+## Setting insets using lengths
+
+The `animation-range-start` and `animation-range-end` properties each accept a named animation range, a {{cssxref("length-percentage")}} offset value, or both. Any length or percentage offset is from the _start_ of the animation attachment range.
```
chrisdavidmills:
```suggestion
The `animation-range-start` and `animation-range-end` properties each accept a named animation range, a {{cssxref("length-percentage")}} offset value, or both. Any length or percentage offset is measured from the _start_ of the animation attachment range.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+When a {{cssxref("length")}} is set, the offset is fairly intuitive.
+Here we use the `animation-range-start` and `animation-range-end` properties to inset the animation timeline. In this way, we define a subsection of the element's full animation attachment range as the active interval, set the distance from the start of the animation attachment range.
```
chrisdavidmills: It feels to me like this should explain that we are insetting the default `cover` timeline range.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+    transparent calc(var(--start) + 1px) calc(var(--end) - 1px),
+    #ccc calc(var(--end) - 1px) calc(var(--end) + 1px),
+    transparent calc(var(--end) + 1px)
+  );
+}
+```
+
+{{EmbedLiveSample("inset_length", "100%", "400")}}
+
+We've added a lines `1em` and `125px` from the block end edge of the scroll container. The animation starts when the block start edge of the subject element reaches the line `1em` line and ends when it reaches the `125px` line.
```
chrisdavidmills:
```suggestion
We've added lines `1em` and `125px` from the block end edge of the scroll container. The animation starts when the block start edge of the subject element reaches the `1em` line and ends when it reaches the `125px` line.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+In this case, as the animation attachment range is the same for both properties, the location of the insets was fairly straightforward.
```
chrisdavidmills:
```suggestion
In this case, as the animation attachment range is the same for both the start and end inset values, the location of the insets is fairly straightforward.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+## Lengths and ranges
```
chrisdavidmills: This title doesn't really mean very much, and sounds too similar to the previous title. This section is really all about the effect of `<length-percentage>` values when the named timeline range differs, so I've just use the "Length offsets with varying ranges" title here, and delete the instance of it below.

chrisdavidmills: You could also do with a bit of explanation beforehand that explains that when you use a range value that is a named timeline range plus a `<length-percentage>`, it sets an offset on a specific named timeline range, not the default `cover` range.

This could be earlier on (I inserted a comment about it previously), or here.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+</main>
+```
+
+{{EmbedLiveSample("different_length", "100%", "310")}}
+
+As the start edge of both the `normal` and `entry` ranges is the container's end edge, the animation begins when the subject start edge is `50px` from the bottom of the scrollport and ends, reaching `100%` progress, when the subject start edge is `50px` from the bottom of the scrollport, irrespective of the subject size. While the size of the `entry` range is different for the three different subject size, in this case the size of the underlying range didn't matter.
+
+### Length offsets with varying ranges
+
+The size of range matters when the range doesn't start at the element's end edge, as is the case with `exit` and `exit-crossing`, or if the offset is a percentage value. This fact, and the fact that you can mix and match animation range names, make [understanding timeline range names](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_name) important while making view progress timeline offsets a bit more complicated.
```
chrisdavidmills:
```suggestion
The size of range matters when the range doesn't start at the element's end edge, as is the case with `exit` and `exit-crossing`, or if the offset is a percentage value. This fact, and the fact that you can mix and match animation range names, make view progress timeline offsets a bit more complicated to understand than non-offsetted [timeline range names](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_names).
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+For example, when setting `exit` as the timeline range name, the subject size matters as it determines where the location of the end edge of the range.
```
chrisdavidmills:
```suggestion
For example, when setting `exit` as the timeline range name, the subject size matters as it determines the location of the range's end edge.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+```css live-sample___exit_length
+.animated_element {
+  animation-range-start: entry 50px;
+  animation-range-end: exit 75px;
+}
+```
+
+The offset position is relative to the start of the declarations animation range. With `animation-range-start` set to `entry 50px`, the animation's `0%` progress occurs `50px` from the container's end edge (denoted by the blue line) for all three subject sizes; this is the same as in the previous example.
```
chrisdavidmills:
```suggestion
The offset position is relative to the start of the declaration's animation range. With `animation-range-start` set to `entry 50px`, the animation's `0%` progress occurs `50px` from the container's end edge (denoted by the blue line) for all three subject sizes; this is the same as in the previous example.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+        text-align: center;
+      }
+    }
+  }
+}
+```
+
+{{EmbedLiveSample("exit_length", "100%", "310")}}
+
+The `animation-range-end` is where things get interesting. We've set the end of the range to be `75px` from the start edge of the `exit` range. The `exit` attachment range is the opposite of `entry`. With `exit`, the range being the size of the subject, up to the size of the scrollport, with the `0%` progress occurring when the animated element's start edge crosses the scrollport's start edge and the `100%` progress occurring when the subject's end edge crosses the scrollport's start edge.
```
chrisdavidmills: The second sentence here is very long and hard to parse. Can you rephrase? Also, it would help to have a line that demonstrates where the point "`75px` from the start edge of the `exit` range" is, to help readers that are not sure. 

The two larger boxes end their animations, but the 50px one doesn't seem to, which makes this extra confusing.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+### Beyond the scrollport edges
+
+Offsets that are larger than the subject create `animations-range-start` or `animation-range-end` points that occur when no part of the subject is visible.
+
+When the subject is `50px` tall, the `exit` range is `50px`. The point `75px` from the start of the `exit` range past the start edge of the scrollport. The animation only ends when the subject's start edge is `75px` past the scrollport's start edge and the end edge is `25px` past.
```
chrisdavidmills: This is very hard to follow. I didn't realize we were still talking about the previous live example until I got to the "denoted with a red line" phrase. And the animations don't seem to end there. They end about 10px or so above the red line.

Also, I am colorblind, so it is hard to see which one is the red line (I think it's the highest one). Could we use solid and dashed instead?

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+With our `250px` tall container, when the subject is `250px` or `500px` tall, the `exit` range is the size of the container, with the start being the scroll container's end edge. With a `75px` offset, the end of the animation occurs when the end edge of the subject is `75px` from the end edge of the scroll container (denoted with a red line).
+
+The animation ends, meaning animation continues to the `to` keyframe and the {{domxref("Element/animationend_event", "animationend")}} event occurs, even if the animation range end is outside of the scrollport, as long as there is room to scroll to that point. Had we set `animation-range-end: exit 250px`, the animation would have ended when the end edge of the medium and tall subjects exited the scrollport at the container's start edge. The small subject's animation may not end as there may not be `450px` worth of content after the subject in our example for the user to scroll until the animation end is reached.
```
chrisdavidmills: This is really hard to follow. Brain melting out of ears.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+    linear-gradient(
+      to top,
+      transparent 0 calc(var(--total) * 0.6 - 1px),
+      red calc(var(--total) * 0.6 - 1px) calc((var(--total) * 0.6) + 1px),
+      transparent calc(var(--total) * 0.6 + 1px)
+    );
+}
+```
+
+This defines the active interval to begin `20%` into the default attachment range and end `60%` through that same range. The default `normal` animation attachment range, which behaves as [`cover`](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_names#cover), is the height of the scroll container plus the height of the subject element, meaning the range will differ based on which radio button is selected.
```
chrisdavidmills, line 395: OK, this paragraph is really helpful. I now understand what is going on with this percentage inset stuff much better than I did before.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+{{EmbedLiveSample("inset_percent", "100%", "400")}}
+
+For illustrative purposes, there are grey lines crossing the container at the 20% and 60% points of the scrollport's height, measured from from the bottom of the scrollport. As the `animation-range-*` percentages are relative to the timeline range, not the scrollport, these lines only show how the percentages **don't** align. Based on the height of our subjects, the `20%` mark is either `60px`, `100px`, or `150px` from the end edge of the scrollport (marked with a red line), and the `60%` mark is `180px`, `300px`, or `450px` from the same point (marked with a green line).
```
chrisdavidmills: This is all really confusing, however. I don't think it is very helpful to have the grey lines that show where the start and end positions aren't.

I also don't understand what the lines inside the subject are supposed to be either. They also don't seem to match up with anything useful.

I'd honestly simplify this all, remove the other lines, and just show where the start and end insets actually are. Explain why they are where they are in terms of being a percentage of the scrollport + subject height. Show the calculation of the pixel numbers.

Maybe also include a different example where the start and end offset are both inside the scrollport

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+The animation only begins when the element reaches the 20% mark along the animation attachment range. This point is that is `60px`, `100px`, or `150px` from the bottom edge of the scroll port, depending on the the size of the element, is the animation's `0%` progress point. The location of the subject element at this point, representing the position of the element when the `from` or `0%` keyframe is applied, is shown in yellow.
+
+The red represents the location of the animated element relative to the scrollport when the `to` or `100%` keyframe is applied, which is the end of the animation. This point is either `180px`, `300px`, or `450px` from the bottom edge of the scrollport, depending on the subject size. The animation occurs when the element is between the `to` and the `from` positions; between the positions along the animation attachment range defined.
```
chrisdavidmills: OK, so you have explained some of the pixel amount details down here, as I asked for earlier. But it seems very hard to follow down here. It would be better explained right next to the live example.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+You may have noticed something interesting about the dashed horizontal lines: when the animation starts, the line that is 20% from the end edge of the viewport is 20% from the _top_ of the subject element and the line that is 60% from the end edge of the viewport is 60% from the _top_ of the subject element when the animation ends.
```
chrisdavidmills: I found this paragraph very hard to follow, and I'm not sure how it helps.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+
+### Subject size matters
```
chrisdavidmills, line 455: This example is also really hard to follow. I think the issue is trying to understand why the "subject smaller than scrollport" behaves as it does, and then also trying to understand why the "subject larger than scrollport" example behaves as it does, both while trying to imagine how the 0%/100% equivalent would behave. It also includes a negative percentage just to make things even harder.

I would split this up into two examples, one showing "subject smaller than scrollport" with the -20%/20% shown alongside the 0%/100% for comparison, and one showing "subject larger than scrollport" with the -20%/20% shown alongside the 0%/100% for comparison.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+}
+```
+
+{{EmbedLiveSample("exit_percent", "100%", "400")}}
+
+The animation lasts `40%` of the animation-attachment range. As you scroll, note how the larger the subject, the longer the range.
+
+### Percentages equal to the scrollport
+
+For the start and end percentages to be relative to the scrollport and the subjects, we use the `<timeline-range-name>` keyword `contain` along with the percentage for each of our `animation-range-*` values.
```
chrisdavidmills: Do you mean "relative to the height of the scrollport plus the height of the subject", or something else? This is a big vague.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,746 @@
...
+    linear-gradient(black, black), linear-gradient(black, black);
+  background-size: 1px 1px;
+  background-position:
+    center 25%,
+    center 75%;
+  background-repeat: repeat-x;
+}
+```
+
+Without needing to know the size of the container or the subjects, we are able to limit our animation to the middle of scrollport.
```
chrisdavidmills: How does this relate to the text at the start of the section? Especially the bit about subject being same size as scrollport? This needs more explanation of why the animation is limited to the middle of the scrollport.

chrisdavidmills (changes requested):
@estelle first pass, bunch of comments for you.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
estelle: I cut it down a little bit, and it's linked, but I want to keep the intro in this page.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
estelle: done

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
estelle: they're gradients.  I changed it to read bottom red and top blue.

I also prefaced the section with an intro and links to each subsection of this example so it's more clear

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
estelle: reorganized the section, so hopefully it's easier to follow.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
estelle: removed the lines and explanation, explained in px in the next paragraph, but not adding math equations

estelle: Put it back in. It's important to understand and related to the dashed line explanation below.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
estelle: it helps, it heavily relates to percentages with other properties.

estelle (commented):
i think i've addressed all your concerns.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+sidebar: cssref
+---
+
+By default, [view progress timelines](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timelines#view_progress_timelines) track elements across the entire [animation attachment range](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_names#the_animation_attachment_range). The `0%` progress point is at the start of the range, while the `100%` progress point is at the end. The animation attachment range can be changed by setting a [timeline range name](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/timeline_range_names), and the location of the `0%` and `100%` progress points along the range can be adjusted by setting length or percentage-based inset values.
+
+This guide explains how to limit the animation timeline to a specific portion of the animation timeline range using length or percentage inset values.
+
+## Animation timelines: a primer
+
+[CSS animations](/en-US/docs/Web/CSS/Guides/Animations) are created by defining named {{cssxref("@keyframes")}} animations, which define an animation's behavior, and then attaching the keyframe animation to an element using the animation's name.
```
chrisdavidmills:
```suggestion
[CSS animations](/en-US/docs/Web/CSS/Guides/Animations) are created by defining named {{cssxref("@keyframes")}} animations, which specify an animation's behavior, and then attaching the keyframe animation to an element using the animation's name.
```

To get rid of "defining ... define"

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+The element's animation timeline, defined by the {{cssxref("animation-timeline")}} property, determines how and when the element progresses through those keyframes. By default, the timeline is time-based, using the document's default time-based {{domxref("DocumentTimeline")}}.
+
+The [CSS scroll-driven animation](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations) module defines scroll-based and view-based timelines, which are methods of animating property values along a scroll-based timeline rather than the default time-based document timeline.
```
chrisdavidmills:
```suggestion
The [CSS scroll-driven animation](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations) module defines scroll progress and view progress timelines, which are methods for animating property values along a scroll-based timeline rather than the default time-based document timeline. In this article, we'll only discuss view progress timelines, as [scroll progress timelines](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timelines#scroll_progress_timelines) are not relevant to timeline insets.
```

chrisdavidmills: It seemed a bit weird to talk about both, and then never mention scroll-based again throughout the entire article.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+.animated_element {
+  animation-name: nameOfAnimation;
+  animation-timeline: view();
+}
+```
+
+Setting an {{cssxref("animation-name")}} applies the animation to the selected element.
+
+> [!NOTE]
+> The `animation-timeline` property should always come after any `animation` shorthand declarations. While the shorthand property can not be used to set the `animation-timeline` property, it does reset the timeline to the default time-based document timeline.
```
chrisdavidmills:
```suggestion
> The `animation-timeline` property should always come after any `animation` shorthand declarations. While the shorthand property can't be used to set the `animation-timeline` property, it does reset the timeline to the default time-based document timeline.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+As you scroll up, the animation progresses. As you scroll down, the animation reverses.
+
+{{EmbedLiveSample("initial", "100%", "400")}}
+
+In this example, note how any time any part of the subject element is visible in the scrollport, the animation is occurring. By default, view progress animations begin just as the top edge of the subject element aligns with the bottom edge of the scroll container and end, reaching `100%` progress, when the end edge aligns with the start edge of the container, no matter the size of the subject element. By default, the animation is applied when any part of the the subject is visible within the scrollport.
```
chrisdavidmills:
```suggestion
In this example, whenever any part of the subject element is visible in the scrollport, the animation occurs. By default, view progress animations begin just as the top edge of the subject element aligns with the bottom edge of the scroll container and end, reaching `100%` progress, when the end edge aligns with the start edge of the container, no matter the size of the subject element. By default, the animation is applied when any part of the subject is visible within the scrollport.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+### Animation attachment ranges
+
+In a [view timeline progress](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timelines#view_progress_timelines), when no animation range properties are defined, the `<timeline-range-name>` is `normal`, which defaults to `cover`. The animation is applied the entire time any portion of the subject element is visible, meaning the default **animation attachment range** is the sum of the height of the scroll container and the height of the subject element, with that extra height being at the scroll end edge. In our example, as the scroll container is `250px` tall, and the subject is `50px`, `250px`, or `500px` tall, with the vertical animation attachment range being `300px`, `500px`, or `750px` respectively.
```
chrisdavidmills:
```suggestion
In a [view progress timeline](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timelines#view_progress_timelines), when no animation range properties are defined, the `<timeline-range-name>` is `normal`, which defaults to `cover`. The animation is applied whenever any portion of the subject element is visible, meaning the default **animation attachment range** is the sum of the height of the scroll container and the height of the subject element, with that extra height being at the scroll end edge. In our example, as the scroll container is `250px` tall, and the subject is `50px`, `250px`, or `500px` tall, with the vertical animation attachment range being `300px`, `500px`, or `750px` respectively.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+The `0%` progression occurs when the subject element's start edge intersects the scrollport at the end edge, reaching `100%` progress when the subject's end edge exits via the scrollport's start edge. These are the subject and scrollport's top and bottom edges when scrolling vertically, and the left and right or right and left edges when scrolling horizontally, depending on the writing mode.
+
+We can visualize the position of the subject at the `0%` and `100%` progress points for the three subject sizes:
```
chrisdavidmills:
```suggestion
The following diagram illustrates the position of the subject at the `0%` and `100%` progress points for the three subject sizes:
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+{{EmbedLiveSample("svg_view", "100%", "720")}}
+
+The yellow subject elements represents the position of the element when the `from` keyframe is applied, which is the animation range's `0%` progress mark. The red represents the location of the animated element relative to the scrollport when the `to` keyframe is applied, which is the end of the animation, or the `100%` progress mark. The grey represents the scrollport.
+
+By default, the element animates while it is "in view", but this default definition of "in view" may not fit your needs. Fortunately, we can control which edges define the edges of the animation attachment range and then offset the start and end of that range with the animation range properties.
+
+### Animation range properties
+
+The {{cssxref("animation-range")}} properties enable defining a named timeline range and then insetting animations from the animation attachment ranges by setting a {{cssxref("timeline-range-name")}}, such as `contain` or `exit-crossing`, and {{cssxref("length-percentage")}} inset values from the start of the range, with percentages being relative to that named or default timeline range.
```
chrisdavidmills:
```suggestion
The {{cssxref("animation-range")}} properties enable specifying a named timeline range, such as `contain` or `exit-crossing`, which changes the range used from the default `cover` range. You can also include a {{cssxref("length-percentage")}} value, which insets the attachment range from the start of the range. Percentages are relative to the named or default timeline range.
```

chrisdavidmills: I needed to do a bit of work on this paragraph. All of the content was there, but it was very difficult to parse.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+Named timeline ranges define the portions of a {{domxref("ViewTimeline")}} that define an animation's range, specifying the start and end of the animations's attachment range.
+
+The `animation-range` property is a shorthand property, defining the {{cssxref("animation-range-start")}} and {{cssxref("animation-range-end")}} properties. The `animation-range-start` defines the position of the subject element when the animation starts. The `animation-range-end` defines the position of the subject element when the animation ends.
+
+See the [timeline range name guide](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/timeline_range_name) to learn about the different named timeline ranges. The focus of this guide is insetting the start and end of the animation attachment range using {{cssxref("length-percentage")}} inset values.
```
chrisdavidmills:
```suggestion
See the [timeline range names guide](/en-US/docs/Web/CSS/Guides/Scroll-driven_animations/Timeline_range_names) to learn about the different named timeline ranges. This guide focuses on how the {{cssxref("length-percentage")}} inset values work.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+## Setting insets using lengths
+
+The `animation-range-start` and `animation-range-end` properties each accept a named animation range, a {{cssxref("length-percentage")}} offset value, or both. Any length or percentage offset is measured from the _start_ of the animation attachment range.
+
+When a {{cssxref("length")}} is set, the offset is fairly intuitive.
+Here we use the `animation-range-start` and `animation-range-end` properties to inset the animation timeline. In this way, we define a subsection of the element's full animation attachment range as the active interval, set the distance from the start of the default `normal` animation attachment range.
```
chrisdavidmills:
```suggestion
Here we use the `animation-range-start` and `animation-range-end` properties to inset the animation timeline. This defines a subsection of the element's full animation attachment range as the active interval, with the `<length>` values specifying distances from the start of the default `normal` animation attachment range.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+```css live-sample___inset_length
+.animated_element {
+  animation-range-start: 1em;
+  animation-range-end: 125px;
+}
+```
+
+This defines the start and end of the animation as being `1em` and `125px` from the start of the animation attachment range, respectively. Because the timeline range default is `normal`, which resolves to `cover`, the start of the animation attachment range is the block end edge of the container.
```
chrisdavidmills:
```suggestion
The start and end of the animation range are `1em` and `125px` from the start of the animation attachment range, respectively. Because the timeline range default is `normal`, which resolves to `cover`, the start of the animation attachment range is the block end edge of the container.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+    transparent calc(var(--start) + 1px) calc(var(--end) - 1px),
+    #ccc calc(var(--end) - 1px) calc(var(--end) + 1px),
+    transparent calc(var(--end) + 1px)
+  );
+}
+```
+
+{{EmbedLiveSample("inset_length", "100%", "400")}}
+
+We've added lines `1em` and `125px` from the block end edge of the scroll container. The animation starts when the block start edge of the subject element reaches the `1em` line and ends when it reaches the `125px` line.
```
chrisdavidmills, line 124: I can't see any lines in the rendered example.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+In this case, as the animation attachment range resolves to `cover` for both the start and end inset values, the location of the insets is fairly straightforward.
```
chrisdavidmills, line 126: You've said all of this above. Delete this paragraph?

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+### Effect of named ranges on length offsets
+
+The offset distance is always from the start of the associated animation range. In this example, we set the `animation-range-start` to be 50px from the start of the default `normal` range and set the `animation-range-end` to be `100px` from the start of the explicitly set `entry` range:
```
chrisdavidmills:
```suggestion
The offset distance is always from the start of the associated animation range. In this example, we set the `animation-range-start` to be `50px` from the start of the default `normal` range and set the `animation-range-end` to be `100px` from the start of the explicitly set `entry` range:
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+    <p>&nbsp;</p>
+    <p>&nbsp;</p>
+    <p>Scroll up ⇧</p>
+  </article>
+</main>
+```
+
+{{EmbedLiveSample("different_length", "100%", "310")}}
+
+As the start edge of both the `normal` and `entry` ranges is the container's end edge, the animation begins when the subject start edge is `50px` from the bottom of the scrollport and ends, reaching `100%` progress, when the subject start edge is `50px` from the bottom of the scrollport, irrespective of the subject size. While the size of the `entry` range is different for the three different subject size, in this case the size of the underlying range didn't matter.
```
chrisdavidmills:
```suggestion
As the start edge of both the `normal` and `entry` ranges is the container's end edge, the animation begins when the subject start edge is `50px` from the bottom of the scrollport and ends, reaching `100%` progress, when the subject start edge is `100px` from the bottom of the scrollport, irrespective of the subject size. While the size of the `entry` range is different for the three different subject sizes, in this case, the size of the underlying range didn't matter.
```

This explanation makes sense, but the live sample is a bit confusing because you expect the lines to mark exactly where the animation starts and finishes. But the happen slightly after each line. Do they need adjusting a bit?

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+  animation-range-start: entry 60px;
+  animation-range-end: exit 75px;
+}
+```
+
+With both `entry` and `exit`, the range is the size of the subject, with the size clamped to the size of the scrollport. This means that the height of the `entry` and `exit` ranges are the height of the box in both the `50px` and `250px` examples, while in the `500px` example, the range is clamped to the height of the scrollport, which is `250px` tall.
+
+{{EmbedLiveSample("exit_length", "100%", "310")}}
+
+We added a few lines to facilitate the explanations that follow: the bottom blue line is `60px` from the end edge of the scrollport, and the top red line is `75px` from that same edge. The start and the ends of the range are [measured from their respective named ranges](#measured_from_the_start_edge_of_the_range), potentially occurring [beyond the edges of the scrollport](#beyond_the_scrollport_edges), remembering the size of the named range may be [clamped, effecting the offset's starting point](#effects_of__clamping).
```
chrisdavidmills: OK, so the animation starts in each case when the start edge of each subject crosses the top red line. That makes sense. What relevance does the blue line have? None of the subject's animations start or stop anywhere near it.

Also, the last sentence here is way too much to understand in one go. I'm reading it, thinking, "OK, so to understand this sentence, I have to follow three links elsewhere to other information sources, and then come back here, to piece together what it means?" Also, the third link doesn't work.

I would just delete that sentence, and say that in the following sections, we will explain in detail what is going on here.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+#### Measured from the start edge of the range
+
+As the offset position is always relative to the start of the declaration's animation range, the start of the animation for all three elements occurs when the elements start edge crosses the point that is `60px` from the start of the `entry` range.
+
+Don't be confused by the fact the `entry` range for the `50px` subject is only `50px` tall: the size of the named animation range is clamped by the size of the viewport. We are defining the element's animation range - which goes from the scroll position defined by the `animation-range-start` value to the position defined by the animation-range-end value.
```
chrisdavidmills: I don't get the relevance of this paragraph here. It sounds like it should be in the "Effects of clamping" section.

chrisdavidmills: Maybe instead, say something like this:

> The position where the animation ends is defined by the `animation-range-end` value — it is when `75px` of the subject has exited the top edge of the viewport. This varies for each subject:
>
> - For the `50px`-high subject, this only occurs `25px` after it has left the scrollport.
> - For the `250px`-high subject, this occurs when `75px` of it has left the viewport.
> - For the `500px`- high subject, this occurs at the same time as the `250px`-high subject. Why? Because the maximum size of the named animation range is clamped to the size of the scrollport. This subject, therefore, acts as if it is `250px` for this purpose.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+#### Beyond the scrollport edges
+
+For our `50px` tall subject, the `exit` range is 50px tall abutting the start edge of the scroll port. Setting `animation-range-end: exit 75px` for any element less than `75px` tall means end of the range is outside the scrollport as the point `75px` from the start of the `exit` range is past the start edge of the scrollport. In our example, the end of the animation range for the `50px` subject occurs when the subject's start edge is `75px` past the scrollport's start edge. The animation ends, reaching the `to` keyframe and the {{domxref("Element/animationend_event", "animationend")}} event only occuring, if the element is scrolled `25px` out of view.
```
chrisdavidmills:
```suggestion
For our `50px` tall subject, the `exit` range is 50px tall abutting the start edge of the scroll port. Setting `animation-range-end: exit 75px` for any element less than `75px` tall means the end of the range is outside the scrollport, as the point `75px` from the start of the `exit` range is past the start edge of the scrollport. In our example, the end of the animation range for the `50px` subject occurs when the subject's start edge is `75px` past the scrollport's start edge. The animation ends when the element is scrolled `25px` out of view.
```

chrisdavidmills: So, this goes into what my previous comment says, in much more detail. I would still cut it down a bit, as suggested, because it is already such a lot of information to absorb in one paragraph. If you still want to say the bit about the animation ending being where the `to` keyframe is reached and the {{domxref("Element/animationend_event", "animationend")}} event fires, do it somewhere else.

You could then maybe link from my suggestion above to this section from the first bullet, saying this provides a bit more detail. Or maybe just cut it down to a short bulleted list, as I'm not sure the reader needs this much detail?

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+The animation ends, meaning animation continues to the `to` keyframe and the {{domxref("Element/animationend_event", "animationend")}} event occurs, even if the animation range end is outside of the scrollport, as long as there is room to scroll to that point. Had we set `animation-range-end: exit 250px`, the animation would have ended when the end edge of the medium and tall subjects exited the scrollport at the container's start edge. The small subject's animation may not end as there may not be `450px` worth of content after the subject in our example for the user to scroll until the animation end is reached.
```
chrisdavidmills: Again, cut it down a bit:


```suggestion
The animation ends even if the animation range end is outside of the scrollport, as long as there is room to scroll to that point. Had we set `animation-range-end: exit 250px`, the animation would have ended when the end edge of the medium and tall subjects exited the scrollport at the container's start edge.

With the end set to `exit 250px`, the small subject's animation might not end, as there might not be `450px` of content after the subject for the user to scroll to before the end point is reached.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+#### Effects of clamping
+
+With our `250px` tall container, when the subject is `250px` or `500px` tall, the `exit` range is the size of the container, with the start being the scroll container's end edge. With a `75px` offset, the end of the animation occurs when the end edge of the subject is `75px` from the end edge of the scroll container (denoted by the top red line).
+
+As the offset position is always relative to the start of the named or default animation range, in our example, clamping impacts the large subject's `animation-range-end`. We set the end of the range to `exit 75px` which is `75px` from the start edge of the `exit` range. When the subject is the same size as the scrollport (our `250px` subject) or larger (our `500px` subject), the animation range end is `75px` from the end edge of the scrollport, `75px` from the start of the scrollport-clamped range.
```
chrisdavidmills:
```suggestion
As the offset position is always relative to the start of the named or default animation range, in our example, clamping impacts the large subject's `animation-range-end`. We set the end of the range to `exit 75px` which is `75px` from the start edge of the `exit` range. When the subject is the same size as the scrollport (our `250px` subject) or larger (our `500px` subject), the animation range end is `75px` from the end edge of the scrollport, which is `75px` from the start of the scrollport-clamped `exit` range.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+#B::after {
+  content: " (0)";
+}
+```
+
+## Setting insets using percentages
+
+Like length values, percentage values define offsets from the _start_ of the animation attachment range. The percentage offsets are relative to the timeline range dimension, not relative to the scrollport. For this reason, percentage values are not as intuitive as length values for most people (realizing length values weren't that intuitive either).
+
+Here use `animation-range-start` and `animation-range-end` to inset the animation timeline. While we are using the same properties, we set `<percentage>` values instead of `<length>` values:
```
chrisdavidmills:
```suggestion
Here, we use `animation-range-start` and `animation-range-end` to inset the animation timeline. While we are using the same properties, we set `<percentage>` values instead of `<length>` values:
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+{{EmbedLiveSample("inset_percent", "100%", "400")}}
+
+For illustrative purposes, there are to dark lines crossing the container at the `20%` and `60%` points of the full animation range. The animation starts when the block-start edge reaches the `20%` point, which is the bottom green line. The animation ends when the start block edge is 60% of the way through the normal range, which is the top red line. Only when the element is `50px` tall is the top of the subject still in the scrollport when the end of the animation is reached; there are no top red lines when `250px` or `500px` are selected, as the end of the animation range is outside of the scrollport. Based on the height of our subjects, the `20%` mark is either `60px`, `100px`, or `150px` from the end edge of the scrollport (marked with a red line, which is always in the scrollport), and the `60%` mark is `180px`, `300px`, or `450px` from the same point (marked with a green line, but only visible for the 50px subject).
```
chrisdavidmills:
```suggestion
For illustrative purposes, there are two dark lines crossing the container at the `20%` and `60%` points of the full animation range. The animation starts when the block-start edge reaches the `20%` point, which is the bottom green line. The animation ends when the start block edge is 60% of the way through the normal range, which is the top red line.

Only when the element is `50px` tall is the top of the subject still in the scrollport when the end of the animation is reached; there are no top red lines when `250px` or `500px` are selected, as the end of the animation range is outside of the scrollport.

Based on the height of our subjects, the `20%` mark is either `60px`, `100px`, or `150px` from the end edge of the scrollport (marked with a red line, which is always in the scrollport), and the `60%` mark is `180px`, `300px`, or `450px` from the same point (marked with a green line, but only visible for the `50px` subject).
```

This still feels like a huge, impenetrable wall of text that is very hard to understand. I think part of my trouble is that I am color blind, so it is hard to understand which line is which (red versus green). I constantly find myself going "OK, so which line are we referring to now?" But I'm not sure what to suggest to make it better. Maybe just leave it for now and see what readers think? We could always iterate on it later on.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+    <text y="760" x="390">0%</text>
+  </svg>
+</div>
+```
+
+{{EmbedLiveSample("svg_insets2", "100%", "710")}}
+
+As before, the yellow represents the position of the element when the `from` keyframe is applied, the red represents the location when the `to` keyframe is applied, and the grey represents the scrollport. The striped areas are where the red and yellow element representations overlap. For illustrative purposes, we've added dashed black horizontal lines `20%` and `60%` way through the scrollport, starting from the bottom.
+
+The animation only begins when the element reaches the 20% mark along the animation attachment range. This point is `60px`, `100px`, or `150px` from the bottom edge of the scroll port, depending on the the size of the element. The location of the subject element at this point, representing the position of the element when the `from` or `0%` keyframe is applied, is shown in yellow.
```
chrisdavidmills:
```suggestion
The animation only begins when the element reaches the `20%` mark along the animation attachment range. This point is `60px`, `100px`, or `150px` from the bottom edge of the scroll port, depending on the the size of the element. The location of the subject element at this point, representing the position of the element when the `from` or `0%` keyframe is applied, is shown in yellow.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+The red represents the location of the animated element relative to the scrollport when the `to` or `100%` keyframe is applied, which is the end of the animation. This point is either `180px`, `300px`, or `450px` from the bottom edge of the scrollport, depending on the subject size. The animation occurs when the element is between the `to` and the `from` positions.
+
+You may have noticed something interesting about the dashed horizontal lines: when the animation starts, the line that is 20% from the end edge of the viewport is 20% from the _top_ of the subject element and the line that is 60% from the end edge of the viewport is 60% from the _top_ of the subject element when the animation ends. This is what was illustrated by the very light grey lines in the live demo for this example.
```
chrisdavidmills:
```suggestion
You may have noticed something interesting about the dashed horizontal lines: when the animation starts, the line that is `20%` from the end edge of the viewport is `20%` from the _top_ of the subject element and the line that is `60%` from the end edge of the viewport is `60%` from the _top_ of the subject element when the animation ends. This is what was illustrated by the very light grey lines in the live demo for this example.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+### Subject size matters
+
+As we saw when we [set insets with lengths](#setting_insets_using_lengths), the size of the subject can make a difference. When setting animation ranges, percentage values are relative to the size of animation attachment range, not the scrollport. For most named ranges, the size of the attachment range depends partially on the subject size. As percentages are based on the size of the range, the named range impacts the resolved size of the insets. Depending on the name, the start position may also change, impacting the location of the range and therefore the location of progress points.
+
+In this example, we define an active range that is 40% of the size of the subject:
```
chrisdavidmills:
```suggestion
In this example, we define an active range that is `40%` of the size of the subject:
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+```
+
+{{EmbedLiveSample("exit_percent", "100%", "400")}}
+
+The animation lasts `40%` of the animation-attachment range. As you scroll, note how the larger the subject, the longer the range. With exit-crossing, the animation range is not cropped; it is the size of the subject even if the subject is larger than the viewport, with the range abutting the start edge of the scrollport, and extending off the end edge if the subject is larger than the scrollport.
+
+With the `-20%` and `20%` insets, the `50px` subject's will animation over `20px`: the animation starts when the subject's end is `-10px` from range start, or `60px` from exiting the screen, and ends when the subject's end is `40px` from exiting the screen. The middle subject will animate over `100px`: the animation starts when the subject end is `-50px` from range start, which is `50px` off of the scrollport's end edge, and ends when the subject's end is `50px` into the scrollport. The large subject animates over `200px`, starting when the bottom is `600px` from the the container's start edge, with only `150px` in view, and ends when the bottom is 400px from that start edge, when `100px` have scrolled off the start edge.
```
chrisdavidmills, line 481:
```suggestion
With the `-20%` and `20%` insets, the `50px` subject will animate over `20px`: the animation starts when the subject's end is `-10px` from the range start, or `60px` from exiting the screen, and ends when the subject's end is `40px` from exiting the screen.

The middle subject will animate over `100px`: the animation starts when the subject's end is `-50px` from the range start, which is `50px` off of the scrollport's end edge, and ends when the subject's end is `50px` into the scrollport.

The large subject animates over `200px`, starting when the bottom is `600px` from the the container's start edge, with only `150px` in view, and ends when the bottom is 400px from that start edge, when `100px` has scrolled off the start edge.
```

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
```diff
@@ -0,0 +1,758 @@
...
+
+### Percentages equal to the scrollport
+
+When it comes to offsetting with percentages, the least complicated named timeline range is `contain`. With `contain`, the animation range is the size of the scrollport, meaning the start and end percentages are relative to the scrollport. For this reason, when using offsets, you may want to use `contain` instead of letting the range default and resolve to `cover`.
+
+The `contain` range fully _contains_ the animation within the scrollport. It represents the range during which the principal box is either fully contained by, or fully covers, its view progress visibility range within the scrollport. With `contain`, if the subject is the same size or smaller than the scrollport, it can be fully visible, but if the element is the same size as the container, the animation is over `0px`, so happens, but is not visible to the user.
```
chrisdavidmills:
```suggestion
The `contain` range fully _contains_ the animation within the scrollport. It represents the range during which the principal box is either fully contained by, or fully covers, its view progress visibility range within the scrollport. With `contain`, if the subject is the same size or smaller than the scrollport, it can be fully visible. If the element is the same size as the container, however, the animation occurs over `0px`. This means that it runs, but it is not visible to the user.
```

chrisdavidmills (approved):
@estelle, some more comments. I think we are getting there, and these are now mostly grammatical improvements, or requests for simplification in the name of adding a bit more clarity.

There are still a few bits that are really hard to understand, but I think you should just get on and publish this now. It is a super complex subject, you have given it your best shot, and I've reviewed this three times now.

chrisdavidmills (approved):
@estelle, some more comments. I think we are getting there, and these are now mostly grammatical improvements, or requests for simplification in the name of adding a bit more clarity.

There are still a few bits that are really hard to understand, but I think you should just get on and publish this now. It is a super complex subject, you have given it your best shot, and I've reviewed this three times now.

files/en-us/web/css/guides/scroll-driven_animations/timeline_insets/index.md
estelle: good addition!

PR #43591, opened 2026-03-27
FF150Relnote: Document.caretRangeFromPoint() supported

hamishwillee:
FF150 supports [` Document.caretRangeFromPoint()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/caretRangeFromPoint). This is non-standard, so primarily for parity reasons.
Developers should prefer [`caretPositionFromPoint()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/caretPositionFromPoint) when supported "generally" because it is shadow root aware (but not on FF, which doesn't support the option according to BCD).

This adds a release note, and also updates the docs to explain why you should prefer `caretPositionFromPoint()`.

Related docs work can be tracked in #43550

### DIFF
--- a/files/en-us/web/api/document/caretrangefrompoint/index.md
+++ b/files/en-us/web/api/document/caretrangefrompoint/index.md
@@ -12,9 +12,9 @@ browser-compat: api.Document.caretRangeFromPoint
 
 The **`caretRangeFromPoint()`** method of the {{domxref("Document")}} interface returns a {{domxref("Range")}} object for the document fragment under the specified coordinates.
 
-The method predates the definition of shadow DOMs, and will return unpredictable and implementation-dependent results in a document that contains {{domxref("ShadowRoot")}} objects.
+This method predates the concept of shadow DOM, and will return unpredictable and implementation-specific results in a document that contains {{domxref("ShadowRoot")}} objects.
 
-{{domxref("Document/caretPositionFromPoint", "caretPositionFromPoint()")}} should be preferred on supporting browsers, as it is a standardized method that can return the caret position from within {{domxref("ShadowRoot")}} instances — provided they are passed in its `options` parameter.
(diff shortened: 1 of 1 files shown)


files/en-us/web/api/document/caretrangefrompoint/index.md
```diff
@@ -10,11 +10,11 @@ browser-compat: api.Document.caretRangeFromPoint
 
 {{APIRef("DOM")}}{{Non-standard_header}}
 
-The **`caretRangeFromPoint()`** method of the
-{{domxref("Document")}} interface returns a {{domxref("Range")}} object for the document
-fragment under the specified coordinates.
+The **`caretRangeFromPoint()`** method of the {{domxref("Document")}} interface returns a {{domxref("Range")}} object for the document fragment under the specified coordinates.
 
-This method is the WebKit-proprietary implementation of the {{domxref("Document.caretPositionFromPoint")}} method.
+The method predates the definition of shadow DOMs, and will return unpredictable and implementation-dependent results in a document that contains {{domxref("ShadowRoot")}} objects.
```
dipikabh:
```suggestion
This method predates the concept of shadow DOM, and will return unpredictable and implementation-specific results in a document that contains {{domxref("ShadowRoot")}} objects.
```

files/en-us/web/api/document/caretrangefrompoint/index.md
```diff
@@ -10,11 +10,11 @@ browser-compat: api.Document.caretRangeFromPoint
...
+
+{{domxref("Document/caretPositionFromPoint", "caretPositionFromPoint()")}} should be preferred on supporting browsers, as it is a standardized method that can return the caret position from within {{domxref("ShadowRoot")}} instances — provided they are passed in its `options` parameter.
```
dipikabh: - I'd suggest to switch to active voice, so "Use caretPositionFromPoint() instead on.."
- qualify "they'

```suggestion
Use {{domxref("Document/caretPositionFromPoint", "caretPositionFromPoint()")}} instead on supporting browsers because it is a standard method that can return the caret position from within {{domxref("ShadowRoot")}} instances, provided the relevant shadow roots are passed in its `options` parameter.
```

hamishwillee:
Thanks for the review @dipikabh !

PR #43593, opened 2026-03-27
Element.attachShadow() - doc for options.customElementRegistry

### LINKED ISSUE #38266: Scoped Custom Element Registries
This is being standardized here:

- https://github.com/whatwg/dom/pull/1341
- https://github.com/whatwg/html/pull/10869

This is not shipping yet, but there are experimental implementations.

### LINKED ISSUE #43237: Missing scoped custom element registries docs
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/CustomElementRegistry

Multiple

Scoped custom element registries are very useful for consumers of web components, especially for those who use their own component libraries, or that import 3rd-party libraries.

Today, all components are registered into a single, global registry, which means that two different components can't use the same tagname (like `<fancy-button>`) without the page breaking.

Scoped custom element registries allow consumers to _scope_ different custom elements to different documents, elements, or shadow roots.

This is useful, for example, when importing a 3rd-party component library, or when testing multiple versions of the same library side by side.

The scoped custom element registry feature is documented partially on MDN:

* Existing docs:
  * [`CustomElementRegistry()` constructor](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/CustomElementRegistry), to create a new registry.
  * [`initialize()` method](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/initialize), to scope an element, shadow root, or document to a registry.
  * [`shadowrootcustomelementregistry` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/template#shadowrootcustomelementregistry), for scoping a registry to a declarative shadow DOM.
  * [`customElementRegistry` property on DOM elements](https://developer.mozilla.org/docs/Web/API/Element/customElementRegistry), [documents](https://developer.mozilla.org/en-US/docs/Web/API/Document/customElementRegistry), and [shadow roots](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/customElementRegistry), to check which registry is a _thing_ associated to.

The following docs (at least) seem to be missing:

* Missing docs:
  * The `customElementRegistry` option for the `attachShadow()` method: https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow does not mention it 
    - From @hamishwillee - will be added by #43593
  * The `customElementRegistry` option for the `document.createElement()` method: https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement#options does not mention it.
    - From @hamishwillee - was added in https://github.com/mdn/content/pull/43350

Adequate docs for the missing pieces listed above.

_No response_

I think scoped custom element registries are already supported in Safari, and coming to Edge and Chrome 146. Not sure when individual pieces of the API shipped.

hamishwillee:
Adds   `customElementRegistry` option for the `Element.attachShadow()`

Fixes #43237
Fixes #38266

Related docs work can be tracked in #43555

### DIFF
--- a/files/en-us/web/api/window/customelements/index.md
+++ b/files/en-us/web/api/window/customelements/index.md
@@ -21,6 +21,7 @@ A {{domxref("CustomElementRegistry")}}.
 ### Basic usage
 
 The most common example you'll see of this property being used is to get access to the {{domxref("CustomElementRegistry.define()")}} method to define and register a new custom element.
+
 For example:
 
 ```js
@@ -28,7 +29,7 @@ let customElementRegistry = window.customElements;
 customElementRegistry.define("my-custom-element", MyCustomElement);
 ```
 
-However, it is usually shortened to something like the following:
+Note that the custom element class is commonly defined directly inside the `define()` call, as shown:
 
 ```js
 customElements.define(


files/en-us/web/api/element/attachshadow/index.md
```diff
@@ -79,6 +79,10 @@ attachShadow(options)
     - `clonable` {{Optional_Inline}}
       - : A boolean that specifies whether the shadow root is clonable: when set to `true`, the shadow host cloned with {{domxref("Node.cloneNode()")}} or {{domxref("Document.importNode()")}} will include shadow root in the copy. Its default value is `false`.
 
+    - `customElementRegistry` {{Optional_Inline}}
```
chrisdavidmills, line 81: As an FYI, looking at the preview page, the nested DL formatting is broken.

chrisdavidmills: Also, there are a couple of big sections up top that need dumping into a "Description" section further down. Not sure if you fancy fixing that while you are here? I won't block on it, if not ;-)

files/en-us/web/api/element/attachshadow/index.md
```diff
@@ -79,6 +79,10 @@ attachShadow(options)
...
+      - : A {{DOMxRef('CustomElementRegistry')}} that will be used as the [scoped custom element registry](/en-US/docs/Web/API/Web_components/Using_custom_elements#scoped_custom_element_registries) of the attached shadow root.
+        If `null` or `undefined`, the shadow root will use the global registry provided by {{domxref("Window.customElements")}}.
```
chrisdavidmills:
```suggestion
        If `null` or `undefined`, the shadow root will use the global registry referenced by {{domxref("Window.customElements")}}.
```

files/en-us/web/api/element/attachshadow/index.md
```diff
@@ -99,6 +103,9 @@ attachShadow(options)
           - : Elements are not automatically assigned to {{HTMLElement("slot")}} elements. Instead, they must be manually assigned with {{domxref("HTMLSlotElement.assign()")}}.
             Its default value is `named`.
 
+  Note that if you pass a {{DOMxRef('ShadowRoot')}} the method will read the corresponding properties as though it was an [`options`](#options) object.
```
chrisdavidmills: Feels like this should be formatted as a note, and begin the sentence with "If..."

Also, it is a bit confusing — it is not clear _where_ you pass this `ShadowRoot`. Is it one of the option values? Does it take the place of the options object? It feels like the note is referencing one of the option descriptions, but I'm not sure which one.

files/en-us/web/api/element/attachshadow/index.md
```diff
@@ -111,6 +118,8 @@ Returns a {{domxref("ShadowRoot")}} object.
     - where the element definition static property `disabledFeatures` has been given a value of `"shadow"`.
     - that already has a shadow root that was not created declaratively.
     - that has a [declarative shadow root](/en-US/docs/Web/HTML/Reference/Elements/template#declarative_shadow_dom) but the specified `mode` does not match the existing mode.
+    - passing a `customElementRegistry` value that isn't `null` or a locally scoped registry (that you created using `new CustomElementRegistry()`).
```
chrisdavidmills:
```suggestion
    - while passing a `customElementRegistry` value that isn't `null` or a locally scoped registry (that you created using `new CustomElementRegistry()`).
```

files/en-us/web/api/window/customelements/index.md
```diff
@@ -8,11 +8,20 @@ browser-compat: api.Window.customElements
...
+
+A {{domxref("CustomElementRegistry")}}.
 
 ## Examples
 
-The most common example you'll see of this property being used is to get access to the {{domxref("CustomElementRegistry.define()")}} method to define and register a new custom element, e.g.:
+### Basic usage
+
+The most common example you'll see of this property being used is to get access to the {{domxref("CustomElementRegistry.define()")}} method to define and register a new custom element.
+For example:
```
chrisdavidmills, line 25: Maybe put this "For example:" on a new line?

Also, the line later on — "However, it is usually shortened to something like the following:" — makes very little sense, as the code snippet that follows it is longer than the first code snippet ;-)

If you fancy it, maybe replace that line with something like "However, the custom element class is usually defined directly inside the `define()` call, something like this:"

files/en-us/web/api/element/attachshadow/index.md
hamishwillee: @chrisdavidmills Thanks. Fixed the formatting.

I have already fixed the description stuff in another PR, so won't do it in this one to avoid clashes. That PR is lower priority so won't be finished for a week or so.

files/en-us/web/api/element/attachshadow/index.md
hamishwillee: You're not wrong to be confused.

So the options are supposed to be just an object, but developers keep passing an existing shadow root. The trigger for me to do this work was actually a spec modification that allows a ShadowRoot to be passed, because it is just treated as an object. Hence the note.

Given that this isn't actually something we want developers to do, though we allow, I'll just remove the note.

files/en-us/web/api/window/customelements/index.md
hamishwillee: Thanks. Yes, put on new line. Updated the text similar to your suggestion - didn't even see that as it predates my specific changes.

PR #43599, opened 2026-03-27
‎`browser_specific_settings` version dot releases rejected by AMO

D4n2021:
Added info about dot releases being invalid

### Description

Added info that dot releases are invalid

### Motivation

Information on this was missing, it costs us addon-devs time to stumble across validation issues while submitting an addon.

### Additional details

grafik

### DIFF
--- a/files/en-us/mozilla/add-ons/webextensions/manifest.json/browser_specific_settings/index.md
+++ b/files/en-us/mozilla/add-ons/webextensions/manifest.json/browser_specific_settings/index.md
@@ -89,7 +89,7 @@ The `gecko` sub-key supports these properties:
     See [Extensions and the Add-on ID](https://extensionworkshop.com/documentation/develop/extensions-and-the-add-on-id/) for more information about setting extension IDs.
 
 - `strict_min_version`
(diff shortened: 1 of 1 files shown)


Rob--W:
For future reference:
In the past dot releases were included, as one can see at the versions listed at https://addons.mozilla.org/api/v5/applications/firefox/

These days, however, only major versions with `.0` and `.*` are included in practice.

121.0a1 was the last version that also supported `.0a1`(https://github.com/mozilla/addons/issues/15284).

Due to the root CA expiration in Firefox, in practice the minimum supported version capable of receiving updates is 115.0 (ESR), or if ESR versions are not included, 128.0 is the minimum supported versions. Older versions do not recognize the signatures any more, as explained at https://support.mozilla.org/en-US/kb/root-certificate-expiration

D4n2021:
@Rob--W what, ".*" is accepted? The text specifically says that the "*" character isn't working either

Rob--W:
`strict_min_version: "*"` is invalid. Even if you tried to create such an extension, Firefox will refuse to load it.

`strict_max_version: "*"` is meaningless.

D4n2021:
Oh, true, sorry, yes it already said * is invalid, not that it's valid, oops. Thanks!

rebloor:
@D4n2021 do you want to update this PR to add a note including the information provided by @Rob--W

PR #43601, opened 2026-03-27
Synchronize with BCD v7.3.9

OnkarRuikar:
[BCD release v7.3.9](https://github.com/mdn/browser-compat-data/releases/tag/v7.3.9)

/cc cc/ @bsmth @wbamberg

### DIFF
--- a/files/en-us/web/api/cssfontfacedescriptors/index.md
+++ b/files/en-us/web/api/cssfontfacedescriptors/index.md
@@ -2,12 +2,10 @@
 title: CSSFontFaceDescriptors
 slug: Web/API/CSSFontFaceDescriptors
 page-type: web-api-interface
-status:
-  - experimental
 browser-compat: api.CSSFontFaceDescriptors
 ---
 
-{{APIRef("CSSOM")}}{{SeeCompatTable}}
+{{APIRef("CSSOM")}}
 
 The **`CSSFontFaceDescriptors`** interface represents a CSS declaration block for an {{cssxref("@font-face")}} [at-rule](/en-US/docs/Web/CSS/Guides/Syntax/At-rules).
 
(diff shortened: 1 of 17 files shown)


Josh-Cena:
@OnkarRuikar btw, I think you should take Brian off the cc list.

OnkarRuikar:
> @OnkarRuikar btw, I think you should take Brian off the cc list.

Along with you, who should I add to the list?  No one was there to merge v7.3.7. Had to scrap the PR and trigger v7.3.8.

pepelsbey (approved):
Looks good! Thank you 🙂

pepelsbey:
> Along with you, who should I add to the list?  No one was there to merge v7.3.7. Had to scrap the PR and trigger v7.3.8.

I would suggest adding @mdn/content-team. There are other teams as well, which is more flexible than tagging specific people.

Josh-Cena:
I'd say the assigned reviewer is a good baseline, because this does constitute a content change and could have complications when displayed to the reader.

OnkarRuikar:
> I would suggest adding @mdn/content-team. There are other teams as well, which is more flexible than tagging specific people.

@pepelsbey Without being the mdn [organisation member](https://docs.github.com/en/organizations/managing-membership-in-your-organization/inviting-users-to-join-your-organization) on GitHub I can't tag @mdn/content-team . 

> I'd say the assigned reviewer is a good baseline

De facto the code owners have rarely approved or merged these PRs. Also, there is limited time frame to merge these PRs because BCD release happens every Friday or Saturday(weekend), in past on Tuesdays as well. We can't force everyone to merge before the next BCD release. 
That is why looking for dedicated reviewers for such maintenance type of PRs. Some repo maintainer who glances at each content PR and issue ever;y morning.
How about the bcd person who does BCD release?

pepelsbey:
@caugner do you think it could be automated somehow? I don't see why we wouldn't auto-merge such PRs.

caugner:
> @caugner do you think it could be automated somehow? I don't see why we wouldn't auto-merge such PRs.

It makes sense to automate, but IIRC when we discussed this internally in January after I filed https://github.com/mdn/rari/issues/503 the general opinion was that we should get rid of those inline macros, rather than automating the sync.

pepelsbey:
> we should get rid of those inline macros, rather than automating the sync.

I believe it's a project that's been going on for a while now. I would suggest we implement auto merge feature while it's not ready. Otherwise, we're presenting outdated/incorrect information because workflow is painfully slow: first, manual BCD release, then semi (of fully, not sure) automated PR, then another manually merged PR.

PR #43606, opened 2026-03-28
Add note about YUV to RGB support in Chromium 127

ashrafzunaira18:
Added missing note about YUV to RGB conversion support starting from Chromium 127.


ashrafzunaira18:
A review is highly appreciated

chrisdavidmills:
Closing, as this looks confused, and is probably AI slop. The title and the support note don't seem to relate to the page the content is added to. This information should be in the browser compat data, not in the page content.

PR #43609, opened 2026-03-28
Remove duplicated description

mfuji09:
### Description

Remove duplicated description.
It is same to the line 101, and it is no sense here.


hamishwillee (approved):
Good catch. Thanks very much

PR #43621, opened 2026-03-30
FF150 Relnote: CSSFontFaceDescriptors

hamishwillee:
FF150 adds support for `CSSFontFaceRule.style` being a `CSSFontFaceDescriptors`. This adds a release note and minor updates to docs.

- [x] Release note
- [x] CSSFontFaceRule - make into live example

Related docs work can be tracked in #43564

### DIFF
--- a/files/en-us/web/api/cssfontfacerule/index.md
+++ b/files/en-us/web/api/cssfontfacerule/index.md
@@ -24,7 +24,12 @@ _Inherits methods from its ancestor {{domxref("CSSRule")}}._
 
 ## Examples
 
-This example uses the CSS found as an example on the {{cssxref("@font-face")}} page. The first {{domxref("CSSRule")}} returned will be a `CSSFontFaceRule`.
+### Accessing @font-face properties
+
+This example defines a {{cssxref("@font-face")}} rule and then iterates the rules on the page to the associated `CSSFontFaceRule`.
+It then logs some of the properties.
+
+#### CSS
 
 ```css
 @font-face {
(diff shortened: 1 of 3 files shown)


files/en-us/web/api/cssfontfacerule/index.md
```diff
@@ -24,7 +24,12 @@ _Inherits methods from its ancestor {{domxref("CSSRule")}}._
 
 ## Examples
 
-This example uses the CSS found as an example on the {{cssxref("@font-face")}} page. The first {{domxref("CSSRule")}} returned will be a `CSSFontFaceRule`.
+### Accessing @font-face properties
+
+This example defines a {{cssxref("@font-face")}} rule and then iterates the rules on the page to the associated `CSSFontFaceRule`.
```
chrisdavidmills:
```suggestion
This example defines a {{cssxref("@font-face")}} rule and then iterates over the rules on the page until the associated `CSSFontFaceRule` is found.
```

files/en-us/web/api/cssfontfacerule/index.md
```diff
@@ -36,11 +41,47 @@ This example uses the CSS found as an example on the {{cssxref("@font-face")}} p
...
+    log(` type: ${rule.type}`);
+    log(` style: ${rule.style}`);
+  }
+}
 ```
 
+#### Result
+
+{{EmbedLiveSample("Accessing @font-face properties", "100%", "250px")}}
+
```
chrisdavidmills, line 84: The live example looks fine. Just one thing — in the BCD table, what I assume is the enrty for this change is worded as "Type changed to `CSSFontFaceDescriptors`". It doesn't say supported by Fx150 yet, and the wording itself is a bit confusing because it implies that the interface's `type` property will return `CSSFontFaceDescriptors`, but actually it is the `style` property that will return that. It would be less confusing if it were phrased something like "`style` returns a `CSSFontFaceDescriptors` object."

I understand you are working on this, but I'm flagging it anyway just in case.

chrisdavidmills (dismissed):
@hamishwillee couple of bits for you to look at.

files/en-us/web/api/cssfontfacerule/index.md
hamishwillee: @chrisdavidmills Thanks for the review. I've taken most suggestions. Also updated the `style` page too to make it a live example.

I don't plan to do anything about this, because it is a BCD issue.
1. The table has since updated so it does say FF150
2. The "Type changed to" is nested under style, and that is shown visually. I don't disagree that it isn't hugely obvious, but that's out of scope for me there.

Image

hamishwillee:
@chrisdavidmills Can you look at this again please.

files/en-us/web/api/cssfontfacerule/index.md
chrisdavidmills: OK, fair enough.

chrisdavidmills (approved):
@hamishwillee, this is looking GTG. Merging!

PR #43623, opened 2026-03-30
docs: fix typo in [\w-] regex example

sierra9999:
### Description
Updated the regex character class example for `[\w-]` to correctly identify the hyphen (`-`) as the intended match in the string "non-profit".

### Motivation
The previous text stated that the pattern matches the "n" in "non-profit". Since `\w` already matches "n", the original example failed to demonstrate the actual effect of adding the hyphen to the character set. Changing it to match the "-" makes the technical distinction clear for learners.

### Additional details
N/A

N/A


Josh-Cena:
This is intended. It lists the very first character that's matched, like the behavior of `exec()`.

PR #43626, opened 2026-03-30
Add NavigationTransition.to docs

tunetheweb:
### Description

Adds the `NavigationTransition.to` docs, as this was added in Chrome 145.

Also adds a small note about when this `NavigationTransition` data is available as it wasn't always clear to me that this wouldn't be available outside of an interception.

### Motivation

Noticed this was missing today when looking it up.

### Additional details

See: https://chromestatus.com/feature/6008183302782976

Added to BCD already in https://github.com/mdn/browser-compat-data/pull/28833


files/en-us/web/api/navigationtransition/index.md
```diff
@@ -9,7 +9,7 @@ browser-compat: api.NavigationTransition
 
 The **`NavigationTransition`** interface of the {{domxref("Navigation API", "Navigation API", "", "nocode")}} represents an ongoing navigation, that is, a navigation that hasn't yet reached the {{domxref("Navigation/navigatesuccess_event", "navigatesuccess")}} or {{domxref("Navigation/navigateerror_event", "navigateerror")}} stage.
 
-It is accessed via the {{domxref("Navigation.transition")}} property.
+It is accessed via the {{domxref("Navigation.transition")}} property. This is only available during a [navigation interception](/en-US/docs/Web/API/Navigation/navigate_event#handling_a_navigation_using_intercept).
```
hamishwillee: 1. What is the alternative to a navigation interception - a scroll intercept? If so, perhaps something like

```suggestion
It is accessed via the {{domxref("Navigation.transition")}} property.
This is only available during a [navigation interception](/en-US/docs/Web/API/Navigation/navigate_event#handling_a_navigation_using_intercept) using [`intercept()`](/en-US/docs/Web/API/NavigateEvent/intercept)).
Within a [Scroll navigation](/en-US/docs/Web/API/Navigation/navigate_event#handling_scrolling_using_scroll) the value is XXXXX.
```

2. What happens in the case it is not available - is it `null`? This would be updated in XXXXX above.

files/en-us/web/api/navigation/transition/index.md
```diff
@@ -8,7 +8,7 @@ browser-compat: api.Navigation.transition
 
 {{APIRef("Navigation API")}}
 
-The **`transition`** read-only property of the {{domxref("Navigation")}} interface returns a {{domxref("NavigationTransition")}} object representing the status of an in-progress navigation, which can be used to track it.
+The **`transition`** read-only property of the {{domxref("Navigation")}} interface returns a {{domxref("NavigationTransition")}} object representing the status of an in-progress navigation, which can be used to track it. This is only available during a [navigation interception](/en-US/docs/Web/API/Navigation/navigate_event#handling_a_navigation_using_intercept).
```
hamishwillee: Comments as for https://github.com/mdn/content/pull/43626/changes#r3012877744 - mostly "precisely what does available mean and in what methods do we see that case".

files/en-us/web/api/navigationtransition/to/index.md
```diff
@@ -0,0 +1,37 @@
...
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.NavigationTransition.to
+---
+
+{{APIRef("Navigation API")}}{{seecompattable}}
+
+The **`to`** read-only property of the
+{{domxref("NavigationTransition")}} interface returns the {{domxref("NavigationHistoryEntry")}} that the transition is navigating to.
```
hamishwillee: Remove spurious line break

```suggestion
The **`to`** read-only property of the {{domxref("NavigationTransition")}} interface returns the {{domxref("NavigationHistoryEntry")}} that the transition is navigating to.
```

files/en-us/web/api/navigationtransition/to/index.md
```diff
@@ -0,0 +1,36 @@
...
+slug: Web/API/NavigationTransition/to
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.NavigationTransition.to
+---
+
+{{APIRef("Navigation API")}}{{seecompattable}}
+
+The **`to`** read-only property of the {{domxref("NavigationTransition")}} interface returns the {{domxref("NavigationHistoryEntry")}} that the transition is navigating to.
```
hamishwillee: `NavigationHistoryEntry` has mentions that it is most commonly accessed via [a number of APIs]. Should we add this one to the list?

files/en-us/web/api/navigationtransition/to/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Value
+
+A {{domxref("NavigationHistoryEntry")}} object.
+
+## Examples
+
+```js
+console.log(navigation.transition.to);
```
hamishwillee: Can we have a more real world example either here or in the `intercept()` doc (and linked from here)?

files/en-us/web/api/navigationtransition/index.md
tunetheweb: The alternative is the navigation is not intercepted. Ie. You are passively observing the `navigate` event but not taking over the navigation to prevent it from happening and handling it manually (e.g. by updating content in place in an SPA).

Also some navigations (e.g. cross-origin navigations) cannot be intercepted so will always be passively observed.

So this is not set  (i.e. `null`) except in the `intercept()` handler.

files/en-us/web/api/navigationtransition/to/index.md
tunetheweb: We don’t mention `from` in that list either, so we should either add both or neither.

My vote is neither as I think the two it mentions are the more common use cases of how to get at these.

files/en-us/web/api/navigationtransition/to/index.md
tunetheweb: I just copied this from the `from` doc. Will try to get better examples for both.

files/en-us/web/api/navigationtransition/to/index.md
```diff
@@ -0,0 +1,36 @@
...
+title: "NavigationTransition: to property"
+short-title: to
+slug: Web/API/NavigationTransition/to
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.NavigationTransition.to
+---
+
+{{APIRef("Navigation API")}}{{seecompattable}}
```
hamishwillee, line 11: @tunetheweb The chromestatus says this:

> It is especially useful when using precommit handlers, as during precommit the current URL has not yet switched to the destination.

I don't think this update currently captures this point "why and when this is useful". That's why I asked for a real example in https://github.com/mdn/content/pull/43626/changes#r3012891457

I assume you might change information you fetch on the basis of both the from and to destinations but readers shouldn't have to make that assumption.

files/en-us/web/api/navigationtransition/to/index.md
```diff
@@ -0,0 +1,36 @@
...
+
```
hamishwillee, line 12: PS FMI perhaps a dumb question highlighting my limited knowledge of this interface but ....

This returns a [`NavigationHistoryEntry`](https://developer.mozilla.org/en-US/docs/Web/API/NavigationHistoryEntry) right? If we're in a precommit we haven't actually ended our navigation, so how can we have a history entry?

files/en-us/web/api/navigationtransition/index.md
hamishwillee: Thanks! So how about we are very precise?

```suggestion
It is accessed via the {{domxref("Navigation.transition")}} property.
However it is only populated within the [`intercept()`](/en-US/docs/Web/API/NavigateEvent/intercept) handler (i.e. during a [navigation interception](/en-US/docs/Web/API/Navigation/navigate_event#handling_a_navigation_using_intercept)), and is otherwise `null`.
```

files/en-us/web/api/navigationtransition/to/index.md
tunetheweb: Oh you’re totally right! This returns a NavigationDestination not a history entry! Will update.

files/en-us/web/api/navigationtransition/to/index.md
tunetheweb: Actually this is not longer relevant as it should be the navigation destination not the history entry. Resolving.

files/en-us/web/api/navigationtransition/index.md
tunetheweb: Let me double check tomorrow whether it’s null or undefined before committing this. Also it’s the whole `transition` parent that is null or undefined. Not just the `to` property.

hamishwillee: Thanks. FYI, I'm away until maybe Easter Monday after today, so won't be looking again until then.

files/en-us/web/api/navigationtransition/to/index.md
```diff
@@ -0,0 +1,42 @@
...
+## Examples
+
+```js
+  navigation.addEventListener("navigate", e => {
+    e.intercept({
+        async precommitHandler() {
+            console.log(`About to transition to ${navigation.transition.to.url}`);
+        }
+    });
+  });
```
tunetheweb:
```suggestion
});
```

files/en-us/web/api/navigationtransition/to/index.md
noamr: It's better perhaps to show an example that doesn't check `navigation.transition` from within the navigation event.

e.g.
```js
requestAnimationFrame(() => {
  if (navigation.transition?.to === "some-destination-url") {
     /* change the way the frame is rendered */
  }
})
```

tunetheweb: Thanks. Used `onnavigateerror` as discussed offline as less contrived example.

files/en-us/web/api/navigationtransition/index.md
```diff
@@ -9,7 +9,7 @@ browser-compat: api.NavigationTransition
 
 The **`NavigationTransition`** interface of the {{domxref("Navigation API", "Navigation API", "", "nocode")}} represents an ongoing navigation, that is, a navigation that hasn't yet reached the {{domxref("Navigation/navigatesuccess_event", "navigatesuccess")}} or {{domxref("Navigation/navigateerror_event", "navigateerror")}} stage.
```
hamishwillee: This is not wrong but feels a little awkward. I think this is a little better
```suggestion
The **`NavigationTransition`** interface of the {{domxref("Navigation API", "Navigation API", "", "nocode")}} represents an ongoing navigation — a navigation that hasn't yet reached the {{domxref("Navigation/navigatesuccess_event", "navigatesuccess")}} or {{domxref("Navigation/navigateerror_event", "navigateerror")}} stage.
```

files/en-us/web/api/navigationtransition/index.md
```diff
@@ -21,6 +21,8 @@ It is accessed via the {{domxref("Navigation.transition")}} property.
   - : Returns the {{domxref("NavigationHistoryEntry")}} that the transition is coming from.
 - {{domxref("NavigationTransition.navigationType", "navigationType")}} {{ReadOnlyInline}}
   - : Returns the type of the ongoing navigation.
+- {{domxref("NavigationTransition.to", "to")}} {{ReadOnlyInline}} {{Experimental_Inline}}
+  - : Returns the {{domxref("NavigationHistoryEntry")}} that the transition is navigating to.
```
hamishwillee: Same fix


```suggestion
  - : Returns the {{domxref("NavigationDestination")}} that the transition is navigating to.
```

files/en-us/web/api/navigationtransition/index.md
```diff
@@ -9,7 +9,7 @@ browser-compat: api.NavigationTransition
 
 The **`NavigationTransition`** interface of the {{domxref("Navigation API", "Navigation API", "", "nocode")}} represents an ongoing navigation, that is, a navigation that hasn't yet reached the {{domxref("Navigation/navigatesuccess_event", "navigatesuccess")}} or {{domxref("Navigation/navigateerror_event", "navigateerror")}} stage.
 
-It is accessed via the {{domxref("Navigation.transition")}} property.
+It is accessed via the {{domxref("Navigation.transition")}} property. However, `Navigation.transition` is only populated while the [`intercept()`](/en-US/docs/Web/API/NavigateEvent/intercept) handler is unresolved (i.e., during a [navigation interception](/en-US/docs/Web/API/Navigation/navigate_event#handling_a_navigation_using_intercept)), and is otherwise `null`.
```
hamishwillee: Good information. I'm being pedantic, but I think this is a little easier to read.


```suggestion
It is accessed via the {{domxref("Navigation.transition")}} property.
Note that this property is only populated while the [`intercept()`](/en-US/docs/Web/API/NavigateEvent/intercept) handler is unresolved (i.e., during a [navigation interception](/en-US/docs/Web/API/Navigation/navigate_event#handling_a_navigation_using_intercept)), and is otherwise `null`.
```

files/en-us/web/api/navigationtransition/to/index.md
```diff
@@ -0,0 +1,42 @@
...
+slug: Web/API/NavigationTransition/to
+page-type: web-api-instance-property
+status:
+  - experimental
+browser-compat: api.NavigationTransition.to
+---
+
+{{APIRef("Navigation API")}}{{seecompattable}}
+
+The **`to`** read-only property of the {{domxref("NavigationTransition")}} interface returns the {{domxref("NavigationDestination")}} that the transition is navigating to.
```
hamishwillee, line 13:
```suggestion
The **`to`** read-only property of the {{domxref("NavigationTransition")}} interface returns the {{domxref("NavigationDestination")}} that the transition is intending to navigate to.
```

tunetheweb: Not so sure about this one to be honest.

I agree it can be intercepted so it may not be there (which is why I think you went with “intending”), but it feels a bit weird to give something inanimate “intent”.

And it IS navigating to that (the navigation has started and will continue there unless something stops it - like an intercept). Whereas “intending to” gives the impression that at some future point it will get round to navigating there. Rather than it being in the process of that navigation.

WDYT?

files/en-us/web/api/navigationtransition/to/index.md
```diff
@@ -0,0 +1,42 @@
...
+
+This mirrors the {{domxref("NavigateEvent.destination")}} property but, unlike that, is available outside of the {{domxref("Navigation.navigate_event", "navigate")}} event handler. It is especially useful when calling functions prior to the URL changing (e.g., during precommit or on error).
+
+## Value
+
+A {{domxref("NavigationDestination")}} object.
+
+## Examples
+
```
hamishwillee, line 22: Web API template suggests using  a level 3 heading. 

```suggestion
## Examples

### Handling a navigation error

```

files/en-us/web/api/navigationtransition/to/index.md
hamishwillee: Yes, I wanted to capture that this is the expected destination but that you might change it in precommit or that you might not get to that destination due to an error (note, I'm not sure what happens in the case of error - i.e. in your example what "do something" might be)

Perhaps we should leave this bit as is, and modify `It is especially useful when calling functions prior to the URL changing (e.g., during precommit or on error).` to be more specific about WHY it is useful - that should cover that either the intercept might change the URL, or you might fail to get there due to error.

Does that seem reasonable.

tunetheweb: Well so basically it's only available during an 'intercept()` at which point you've already stopped the navigation. BUT the point of that `intercept()' is for you to handle it locally (i.e. as an SPA navigation). So it IS still navigating there, just not as a full page load navigation.

The navigation is basically in two parts: the precommit that happens before URL changes. This can be used to [set up two pages view transitions for example](https://www.bram.us/2025/11/17/faking-two-phase-view-transitions-with-the-navigation-apis-precommithandler/) or any other "pre" work, and then the actual navigation after.

So it's not so much that you stop the navigation (or that an error will stop the navigation) and more so that you have alternative ways of handling the navigation. I guess technically you could stop it, or send it somewhere else, but that's not really the common use case. The common use case is the SPA option of "don't handle this as a full page load navigation please browser, I got this and will update it in place, but the user will still see it as having navigated to that page". 

Sp personally I'd just leave it as is.

PR #43434, opened 2026-03-13
Add loading attribute for audio and video

tunetheweb:
### Description

We're adding `loading=lazy` to `<audio>` and `<video>` attributes!

Coming to Chrome in 147 (behind a flag): https://chromestatus.com/feature/5200068565139456

FYI: @scottjehl @hjanuschka

### Motivation

More details:

https://scottjehl.com/posts/lazy-media/ (with a demo of it working)
https://www.januschka.com/chromium-video-lazy-loading.html

### Additional details

Test results and supporting details
Audio:
https://wpt.fyi/results/html/semantics/embedded-content/the-audio-element?label=experimental&label=master&aligned

Video:
https://wpt.fyi/results/html/semantics/embedded-content/the-video-element?label=experimental&label=master&aligned

https://chromestatus.com/feature/5200068565139456

Implementation bugs:

- Chrome: https://crbug.com/469111735 -
- Firefox: https://bugzil.la/2005072 (being implemented at: https://phabricator.services.mozilla.com/D278547)
- Safari: https://webkit.org/b/303995 (being implemented at: https://github.com/WebKit/WebKit/pull/58220)

### Related issues and pull requests

Makes progress on: https://github.com/mdn/mdn/issues/809

- [x] BCD PR: https://github.com/mdn/browser-compat-data/pull/29256
- [x] Spec PR: https://github.com/whatwg/html/pull/11980

PR #43627, opened 2026-03-30
Lazy loading audio controls

### LINKED ISSUE #43624: Specify that Lazy Audio elements need a controls attr
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/audio#loading

loading + controls

In order for audio elements to intersect visibly with the viewport, they need to be visible. UAs use the controls attribute to make audio elements visible, so it's needed for lazy loading.

`<audio loading=lazy controls src=...></audio>`

https://html.spec.whatwg.org/multipage/media.html#the-audio-element
Image

_No response_

tunetheweb:
Fixes #43624

### Description

Adds a note that for audio to be lazy loaded the `controls` attribute needs to be set.

I've also intended the lazy-loading stuff under the `loading=lazy` attribute as they don't apply to the `loading` attribute overall.

### Motivation

Clarification in case this is not obvious

Small update to #43434 when I added these docs.


scottjehl:
This PR looks great. Nice work, @tunetheweb

Issue #43604, opened 2026-03-28
The Forms and buttons 2 test needs an explanation.


jmj073:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Test_your_skills/Forms_and_buttons

### What specific section or headline is this issue about?

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Test_your_skills/Forms_and_buttons#forms_and_buttons_2

### What information was incorrect, unhelpful, or incomplete?

Please understand that the writing may be awkward as I am using a translator.

I am currently studying web development using MDN. While I am receiving a lot of help, there is one area that is confusing. There is a discrepancy between the explanations and test code in [*Forms and Buttons*](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/HTML_forms#checkboxes). It states that checkboxes generally have different names, but the tests in [*Forms and Buttons 2*](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Test_your_skills/Forms_and_buttons#forms_and_buttons_2) do not. Consequently, it is difficult to correctly identify the answers in the tests.

> The main difference (apart from the type value!) is that each checkbox has a different name value, and they generally aren't given value attributes. Behavior-wise, this means they represent different data values, whereas a radio button set only represents one. On submission, each value is submitted with a value of on if the checkbox was checked — yoga=on, balloon=on, etc.

It would be helpful to have additional explanations regarding checkboxes. Thank you.

### What did you expect to see?

```HTML
<form>
    <fieldset>
      <legend>Who is your favorite pony?</legend>
      <ul>
        <li>
          <label for="pinkie">Pinkie Pie</label>
          <input type="radio" id="pinkie" name="pony" value="pinkie" checked />
        </li>
        <li>
          <label for="rainbow">Rainbow Dash</label>
          <input type="radio" id="rainbow" name="pony" value="rainbow" />
        </li>
        <li>
          <label for="twilight">Twilight Sparkle</label>
          <input type="radio" id="twilight" name="pony" value="twilight" />
        </li>
      </ul>
    </fieldset>
    <fieldset>
      <legend>Hotdog preferences</legend>
      <ul>
        <li>
          <label for="vegan">Vegan</label>
          <input type="checkbox" id="vegan" name="hotdog_vegan" />
        </li>
        <li>
          <label for="onions">Onions</label>
[...]

PR #43636, opened 2026-03-31
Make checkbox solution consistent with tutorial

chrisdavidmills:
### Description

The solution for the second task is inconsistent with the guidance we give in the associated tutorial.

Fixes https://github.com/mdn/content/issues/43604


chrisdavidmills:
> Personally I disagree with this statement:
> 
> > However, there is not much point doing this. A checkbox is either submitted with a single value if checked, or it isn't submitted at all. It doesn't really matter what value is sent to the server.
> 
> And the example you have here is exactly one such counterexample. `hotdog=onions&hotdog=mustard&hotdog=ketchup` specifies an array which is much more handleable on the server than having a bunch of discrete fields. Even for an intro guide where we choose to limit our scope I don't think we should actively dismiss cases we didn't mention. But this fix does fit in the context of what was introduced.

I agree; I've just removed the sentence altogether, in https://github.com/mdn/content/pull/43703.

PR #43649, opened 2026-04-02
Fix RFC link for brotli

mathewhodson:
Fixes: 58ea518e521a ("CompressionStream/DecompressionStream support brotli")


mathewhodson:
The old link was duplicating the one for gzip at https://www.rfc-editor.org/rfc/rfc1952

PR #43658, opened 2026-04-03
Add page on integrity; revise SRI guide

### LINKED ISSUE #43302: SRI examples are incomplete
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Subresource_Integrity

Using Subresource Integrity

The standard describes multiple integrity values, but the current node doesn't mention which delimiter is used.

It's a space; see also https://www.w3.org/TR/sri/#agility

I'd love to have all examples include multiple hashes. According to the standard, the browser should choose a hash or hash-function. Therefore, a server should probably offer ALL digest functions.

Furthermore, you'd love to add an example of the same digest function being used twice. Since you can add the same twice, in case you respond differently depending on request headers.

https://www.w3.org/TR/sri/#agility

Affected pages:

https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Subresource_Integrity
https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement/integrity

wbamberg:
Fixes https://github.com/mdn/content/issues/43302.

We didn't have a standalone page for `integrity`, and since it's a bit involved and used in more than one element. I thought it was worth creating one.

I've talked about including multiple hash values, and the behavior as I understand it.

TBH I don't understand why anyone would provide more than one hash function rather than just use SHA-512. Maybe I'm missing something?

I haven't systematically rewritten the SRI guide, just made some changes that seemed necessary.

I would be tempted to remove a lot of the detail about "how to generate hashes" and just point people at tools like OpenSSL.


files/en-us/web/html/reference/elements/link/index.md
```diff
@@ -229,10 +229,8 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
-- `integrity`
-  - : Contains inline metadata — a base64-encoded cryptographic hash of the resource (file) you're telling the browser to fetch.
-    The browser can use this to verify that the fetched resource has been delivered without unexpected manipulation.
-    The attribute must only be specified when the `rel` attribute is specified to `stylesheet`, `preload`, or `modulepreload`.
+- [`integrity`](/en-US/docs/Web/HTML/Reference/Attributes/integrity)
+  - : This attribute contains one or more {{glossary("hash function", "hashes")}} of the resource, and is used to ensure that the content of the resource is what the developer expects it to be, and has not been replaced with a malicious copy in a [supply chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks). The attribute must only be specified when the `rel` attribute is specified to `stylesheet`, `preload`, or `modulepreload`.
```
hamishwillee: I want to go short and choppy and tidy double use of specified

```suggestion
  - : This attribute contains one or more {{glossary("hash function", "hashes")}} of the resource. It is used to ensure that the content of the resource is what the developer expects it to be, and has not been replaced with a malicious copy in a [supply chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks). The attribute must only be specified when the `rel` attribute is set to `stylesheet`, `preload`, or `modulepreload`.
```

files/en-us/web/html/reference/elements/script/index.md
```diff
@@ -86,8 +86,8 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
       - : Don't set a preference for the fetch priority.
         This is the default.
         It is used if no value or an invalid value is set.
-- `integrity`
-  - : This attribute contains inline metadata that a user agent can use to verify that a fetched resource has been delivered without unexpected manipulation. The attribute must not be specified when the `src` attribute is absent. See [Subresource Integrity](/en-US/docs/Web/Security/Defenses/Subresource_Integrity).
+- [`integrity`](/en-US/docs/Web/HTML/Reference/Attributes/integrity)
+  - : This attribute contains one or more {{glossary("hash function", "hashes")}} of the script, and is used to ensure that the content of the script is what the developer expects it to be, and has not been replaced with a malicious script in a [supply chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks). The attribute must not be specified when the `src` attribute is absent. See also [Subresource Integrity](/en-US/docs/Web/Security/Defenses/Subresource_Integrity).
```
hamishwillee: Consider

```suggestion
  - : This attribute contains one or more {{glossary("hash function", "hashes")}} of the script. It is used to ensure that the content of the script is what the developer expects it to be, and has not been replaced with a malicious script in a [supply chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks). The attribute must not be specified when the `src` attribute is absent. See also [Subresource Integrity](/en-US/docs/Web/Security/Defenses/Subresource_Integrity).
```

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -21,37 +15,63 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
 <script src="https://not-example.com/script.js"></script>
 ```
 
-This comes with a risk, in that if an attacker gains control of the third-party host, the attacker can inject arbitrary malicious content into its files (or replace the files completely) and thus can also potentially attack sites that fetch files from it.
+This comes with a risk, in that if an attacker gains control of the third-party host, the attacker can inject arbitrary malicious content into its files (or replace the files completely) and thus can also potentially attack sites that fetch files from it, in a [supply chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks).
```
hamishwillee: I would do this as below to remove some of the unnecessary repetition

```suggestion
This comes with a risk, in that if an attacker gains control of the third-party host, then they can inject arbitrary malicious content into its files (or replace the files completely). This is referred to as a [supply chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks).
```

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -21,37 +15,63 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
...
 
-Subresource Integrity enables you to mitigate some risks of attacks such as this, by ensuring that the files your web application or web document fetches have been delivered without an attacker having injected any additional content into those files — and without any other changes of any kind at all having been made to those files.
+Subresource Integrity is a defense against attacks such as this, by ensuring that the files your web application fetches have not been modified.
```
hamishwillee:
```suggestion
Subresource Integrity is a defense against attacks such as this, by ensuring that the files your web application fetches are exactly the version that it expects.
```

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -21,37 +15,63 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
...
 
 ## Using Subresource Integrity
 
-You use the Subresource Integrity feature by specifying a base64-encoded cryptographic hash of a resource (file) you're telling the browser to fetch, in the value of the `integrity` attribute of a {{HTMLElement("script")}} element or a {{HTMLElement("link")}} element with [`rel="stylesheet"`](/en-US/docs/Web/HTML/Reference/Attributes/rel#stylesheet), [`rel="preload"`](/en-US/docs/Web/HTML/Reference/Attributes/rel/preload), or [`rel="modulepreload"`](/en-US/docs/Web/HTML/Reference/Attributes/rel/modulepreload).
+You can use Subresource Integrity with:
 
-An `integrity` value begins with at least one string, with each string including a prefix indicating a particular hash algorithm (currently the allowed prefixes are `sha256`, `sha384`, and `sha512`), followed by a dash, and ending with the actual base64-encoded hash.
+- {{HTMLElement("script")}} elements
```
hamishwillee: Consistency
```suggestion
- {{HTMLElement("script")}} elements.
```

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -21,37 +15,63 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
...
-> [!NOTE]
-> An **integrity** value may contain multiple hashes separated by whitespace. A resource will be loaded if it matches one of those hashes.
+### Setting the `integrity` attribute
+
+To use the feature, add the [`integrity`](/en-US/docs/Web/HTML/Reference/Attributes/integrity) attribute to the element. The value of `integrity` is a whitespace-separated list of {{glossary("hash function", "cryptographic hashes")}} of the content of the linked resource, where each hash is prefixed with an identifier for the hash algorithm used, followed by a dash, and ending with the actual base64-encoded hash value.
+
+Currently the allowed prefixes are `sha256`, `sha384`, and `sha512`.
 
-Example `integrity` string with base64-encoded sha384 hash:
+For example, the following markup adds an `integrity` attribute to a `<script>` element. The attribute contains two SHA-384 hashes, and two SHA-512 hashes:
```
hamishwillee, line 35: My understanding is that
- you specify multiple hashes of the same strength for the file if you have multiple versions of the file that you can accept - for example: minor variations for localisation, potential for different versions being available during some kind of update, or perhaps literally different versions that might be served based on your client. 
- You specify different hashes of the same file for compatibility across browser versions. I _think_ there is no reason to do anything but the highest SHA as browsers have supported the current version for ages. No saying that they won't support more in future though.

You might want to capture some flavour of the first point.

hamishwillee: Ah, I see you described what happens with alternate versions below. YOu might still want to explain why you have alternates, but up to you.

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -21,37 +15,63 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
...
+### Cross-Origin Resource Sharing and Subresource Integrity
+
+For subresource-integrity verification of a resource served from an origin other than the document in which it's embedded, browsers additionally check the resource using [Cross-Origin Resource Sharing (CORS)](/en-US/docs/Web/HTTP/Guides/CORS), to ensure the origin serving the resource allows it to be shared with the requesting origin. Therefore, the resource must be served with an [`Access-Control-Allow-Origin`](/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin) header that allows the resource to be shared with the requesting origin; for example:
+
+```http
+Access-Control-Allow-Origin: *
+```
 
-### Tools for generating SRI hashes
+## Tools for generating SRI hashes
```
hamishwillee, line 139: I'm OK with keeping the tools, but I would push them to the end, after the Integrity-Policy.

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -92,24 +112,6 @@ shasum -b -a 384 FILENAME.js | awk '{ print $1 }' | xxd -r -p | base64
...
-
-   For subresource-integrity verification of a resource served from an origin other than the document in which it's embedded, browsers additionally check the resource using [Cross-Origin Resource Sharing (CORS)](/en-US/docs/Web/HTTP/Guides/CORS), to ensure the origin serving the resource allows it to be shared with the requesting origin.
-
-2. If the script or stylesheet doesn't match its associated `integrity` value, the browser must refuse to execute the script or apply the stylesheet, and must instead return a network error indicating that fetching of that script or stylesheet failed.
-
 ## Integrity policy
 
 The {{httpheader("Integrity-Policy")}} and {{httpheader("Integrity-Policy-Report-Only")}} HTTP headers enable a document to enforce a policy regarding the integrity metadata requirements on loaded script and stylesheet subresources.
```
hamishwillee: Tempted to add

```suggestion
The {{httpheader("Integrity-Policy")}} and {{httpheader("Integrity-Policy-Report-Only")}} HTTP headers enable a document to enforce a policy regarding the integrity metadata requirements on loaded script and stylesheet subresources. In other words, the policy allows a website to _require_ that the `integrity` attribute is specified for loaded resources.
```

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,81 @@
...
+  - html.elements.link.integrity
+  - html.elements.script.integrity
+sidebar: htmlsidebar
+---
+
+The **`integrity`** attribute provides a mechanism for a developer to assert that a linked script or stylesheet must have a particular value. The browser will check that the resource does in fact have that value, and refuse to load the resource if it does not.
+
+This is a defense against a [supply-chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks) in which an attacker gets access to the domain that serves the script or stylesheet, and changes the expected resource for a malicious one.
+
+## Description
```
hamishwillee, line 16: FWIW "Description" is good. HTML is a mess of "Overview", "Usage notes" etc. I think we should globally make it match the Web API and move this below value etc. A job for another day.

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,81 @@
...
+short-title: integrity
+slug: Web/HTML/Reference/Attributes/integrity
+page-type: html-attribute
+browser-compat:
+  - html.elements.link.integrity
+  - html.elements.script.integrity
+sidebar: htmlsidebar
+---
+
+The **`integrity`** attribute provides a mechanism for a developer to assert that a linked script or stylesheet must have a particular value. The browser will check that the resource does in fact have that value, and refuse to load the resource if it does not.
```
hamishwillee: "particular value"? It's a hash, so this doesn't feel right

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+This is a defense against a [supply-chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks) in which an attacker gets access to the domain that serves the script or stylesheet, and changes the expected resource for a malicious one.
+
+## Description
+
+The attribute can be applied to {{htmlelement("script")}} or {{htmlelement("link")}} elements only.
+
+The attribute consists of zero or more components, each of which consists of:
+
+- An identifier for a {{glossary("hash function", "cryptographic hash" function)}}. Three hash functions are supported. In increasing order of strength, these are: SHA-256, SHA-384, and SHA-512.
```
hamishwillee: Maybe reverse - AIUI most should use SHA-512 and I like to present most-useful-first.

hamishwillee (commented):
Looks good

> TBH I don't understand why anyone would provide more than one hash function rather than just use SHA-512. Maybe I'm missing something?

AIUI https://github.com/mdn/content/pull/43658/changes#r3037760065

> I haven't systematically rewritten the SRI guide, just made some changes that seemed necessary.
> I would be tempted to remove a lot of the detail about "how to generate hashes" and just point people at tools like OpenSSL.

Reasonable. I've suggested at least moving the tools to the end after the integrity policy. Also consisider forward referencing integrity-policy in the subresource integrity doc introduction.

files/en-us/web/html/reference/attributes/integrity/index.md
hamishwillee: But either way

```suggestion
- An identifier for a {{glossary("hash function", "cryptographic hash function")}}. Three hash functions are supported. In increasing order of strength, these are: SHA-256, SHA-384, and SHA-512.
```

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+The **`integrity`** attribute provides a mechanism for a developer to assert that a linked script or stylesheet must have a particular value. The browser will check that the resource does in fact have that value, and refuse to load the resource if it does not.
+
+This is a defense against a [supply-chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks) in which an attacker gets access to the domain that serves the script or stylesheet, and changes the expected resource for a malicious one.
+
+## Description
+
+The attribute can be applied to {{htmlelement("script")}} or {{htmlelement("link")}} elements only.
+
+The attribute consists of zero or more components, each of which consists of:
```
hamishwillee: Can it really be zero? I only ask because I think 0 would be meaningless.

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,81 @@
...
+- `sha384-HASH_VALUE`
+- `sha512-HASH_VALUE`
+
+In each case, the part preceding `-` identifies the {{glossary("hash function")}} used, and `HASH_VALUE` is the {{glossary("base64")}} encoding of the result of hashing the resource using the specified hash function.
+
+## Examples
+
+### Including `integrity` on `<script>` elements
+
+The following {{htmlelement("script")}} element includes an `integrity` attribute containing two values calculated using SHA-384, and two calculated using SHA-512. If the browser supports SHA-512 values then it will only use the SHA-512 values, and if the hash of the script at `https://cdn.example.com/script.js` matches either of them, the browser will load it.
```
hamishwillee: I think all browsers support SHA512 so "if the browser supports" is going to be true. Perhaps "because the browser supports (as all do)" text.

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Including `integrity` on `<link>` elements
+
+The following {{htmlelement("link")}} element loads a stylesheet and includes an `integrity` attribute containing a single value, calculated using SHA-512.
+
+```html
+<link
+  rel="stylesheet"
+  href="https://cdn.example.com/style.css"
+  integrity="sha512-MzViNmUxNTFiMDUzZGIwZjcwZDIyYTc5NTA4ZmQyNTYzNjUxYm" />
```
hamishwillee: Claude tells me you're omitting a `crossorigin="anonymous"`

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -21,37 +15,63 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
...
+### How browsers handle Subresource Integrity
 
-> [!NOTE]
-> An `integrity` value's "hash" part is, strictly speaking, a **_cryptographic_ _digest_** formed by applying a particular hash function to some input (for example, a script or stylesheet file). But it's common to use the shorthand "hash" to mean _cryptographic_ _digest_, so that's what's used in this article.
+When a browser encounters a {{HTMLElement("script")}} or {{HTMLElement("link")}} element with an `integrity` attribute, before executing the script or before applying any stylesheet specified by the {{HTMLElement("link")}} element, the browser must first compare the script or stylesheet to the expected hashes given in the `integrity` value.
+
+The different hash functions have different strengths: from weaker to stronger, the order is SHA-256, SHA-384, SHA-512. When the browser downloads a resource with the integrity attribute set, it will first select the set of hashes that were generated using the strongest hash function present. That is, if the attribute contains values generated with SHA-256 and SHA-384, it will only use the hashes generated using SHA-384. It will ignore all other hashes.
```
hamishwillee:
```suggestion
The different hash functions have different strengths: from weaker to stronger, the order is SHA-256, SHA-384, SHA-512. When the browser downloads a resource with the `integrity` attribute set, it will first select the set of hashes that were generated using the strongest hash function present. That is, if the attribute contains values generated with SHA-256 and SHA-384, it will only use the hashes generated using SHA-384. It will ignore all other hashes.
```

files/en-us/web/html/reference/elements/link/index.md
hamishwillee: (or just "when the `rel` attribute is"

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: Why is that needed?

It does appear in some other examples I have seen but I don't know why it is needed.

hamishwillee: This is from Claude - I find it convincing, but I haven't tested it:

---

`crossorigin="anonymous"` is required when using integrity on a cross-origin resource.

When you add an integrity attribute, the browser requires the response to be CORS-eligible so it can inspect the response body to verify the hash. Without crossorigin="anonymous", the browser fetches the resource in no-cors mode, which intentionally makes the response opaque — meaning the browser can't read the bytes to check them against the hash, so it blocks the resource entirely.

wbamberg: I don't understand that. Of course the browser can read the bytes of a no-cors response, how else could it (for example) evaluate it as a script?

This seems more likely: https://w3c.github.io/webappsec-subresource-integrity/#cross-origin-data-leakage - that SRI could allow an attacker to get information about the content of a resource that the site has not opted to share with it, by making a no-cors request with a specific hash and seeing if the resource loads.

Either way, we should explain this!

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: I think we should keep this order because the spec lists them in ascending order of strength, and it would be confusing to make people mentally reverse the order if they look at the spec.

files/en-us/web/security/defenses/subresource_integrity/index.md
wbamberg: I don't like version because it makes me think of semver and stuff, but this is both more specific than that and more general. 

```suggestion
Subresource Integrity is a defense against attacks such as this, by ensuring that the files your web application fetches have exactly the contents that you expect them to have.
```

maybe?

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: I'm not sure why not. The developer thinks the file should contain `alert("hi Hamish")`. That's the particular value. The integrity attribute assert that the file actually does have that value.

But I'd accept alternative suggestions...

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: no you are right, I misread the spec.

-> https://github.com/mdn/content/pull/43658/commits/a5a0a50c8340443030f4cb51c643469d00ef0feb

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: That feels weird. We're talking in the abstract here, i.e. about the specced behavior, not the current actual situation. If we say "because the browser supports (as all do)", the obvious question is, why does the spec allow any other choice?

We could perhaps compromise on:

```suggestion
The following {{htmlelement("script")}} element includes an `integrity` attribute containing two values calculated using SHA-384, and two calculated using SHA-512. If the browser supports SHA-512 values (as all current browsers do) then it will only use the SHA-512 values, and if the hash of the script at `https://cdn.example.com/script.js` matches either of them, the browser will load it.
```

files/en-us/web/security/defenses/subresource_integrity/index.md
wbamberg: -> https://github.com/mdn/content/pull/43658/commits/3cc00abc9f286d1b68aea99138d54ed80b6b1f39

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: I also don't fully understand that though:

> Attackers would attempt to load the resource with a known digest, and watch for load failures.

... because how can the attacker watch for load failure? I thought part of the point about no-cors was that the attacker didn't get information about the success/failure of the request. Or is this just some defense in depth, in case that fails?

files/en-us/web/html/reference/attributes/integrity/index.md
hamishwillee: IMO it is the "if supported" case that is weird. Doing this forces readers to check the compat table when there is no need.

Happy enough with the compromise though.

> , the obvious question is, why does the spec allow any other choice?

No idea. I'd say they expect to remove and add algorithms in future though.

files/en-us/web/html/reference/attributes/integrity/index.md
hamishwillee: Technically you are right - but value is a much overloaded term and I suspect most people will find this as grating as I do. Anyway , the minimal

```suggestion
The **`integrity`** attribute provides a mechanism to assert that a linked script or stylesheet has a cryptographic hash that matches one or more specified values. The browser will check that the resource does in fact have one of the matched values, and refuse to load the resource if it does not.
```

files/en-us/web/security/defenses/subresource_integrity/index.md
hamishwillee: Fair enough. I think what I don't like about "have not been modified" is that the expression really has to be relative to something "have not been modified from the expected contents" or "have not been modified since you uploaded them" or whatever.

Your proposed version works for me.

files/en-us/web/html/reference/attributes/integrity/index.md
hamishwillee: You're right. Yes that makes more sense, and yes we should explain this.

> I also don't fully understand that though:
> 
> > Attackers would attempt to load the resource with a known digest, and watch for load failures.
> 
> ... because how can the attacker watch for load failure? I thought part of the point about no-cors was that the attacker didn't get information about the success/failure of the request. Or is this just some defense in depth, in case that fails?

My understanding is that for a no-cors request you can still get a load or an error event for a loaded resource - albeit with little information about the failure. So I think this is saying that IF no-cors mode was allowed you could distinguish between success or fail of your integrity check and guess the content. However with cors required, a no-cors request be would be blocked and you'd only get the error (never load) so nothing can be inferred about the SRI.

wbamberg: Yeah, I think that might be it. I'll try to write this up.

wbamberg: All right I have had a go at explaining this in https://github.com/mdn/content/pull/43658/commits/c3ff58e2f5a8e62be571f90db665a5f40a1de369.

We ought to have somewhere central to talk about `no-cors` and the other modes.

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -21,98 +15,96 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
...
-```batch
-@echo off
-set bits=384
-openssl dgst -sha%bits% -binary %1% | openssl base64 -A > tmp
-set /p a= < tmp
-del tmp
-echo sha%bits%-%a%
-pause
-```
+If you set the `integrity` attribute on a resource, and the resource is being served from a different {{glossary("origin")}} to the requester, then the site serving the resource must also use [Cross-Origin Resource Sharing (CORS)](/en-US/docs/Web/HTTP/Guides/CORS) to indicate that it has explicitly allowed the resource to be loaded by the requester, by setting the appropriate {{httpheader("Access-Control-Allow-Origin")}} response header.
```
hamishwillee: Very long sentence, that I find hard to parse. Maybe


```suggestion
If you set the `integrity` attribute on a resource, and the resource is being served from a different {{glossary("origin")}} to the requester, then the site serving the resource must also use [Cross-Origin Resource Sharing (CORS)](/en-US/docs/Web/HTTP/Guides/CORS) (set the appropriate {{httpheader("Access-Control-Allow-Origin")}} response header) to indicate that it has explicitly allowed the resource to be loaded by the requester.
```

wbamberg: yeah I know but I struggled to improve it.

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -21,98 +15,96 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
...
+```http
+Access-Control-Allow-Origin: *
+```
 
-> [!NOTE]
-> If OpenSSL is not installed on your system, visit the [OpenSSL project website](https://www.openssl.org/) for information about downloading and installing it. The OpenSSL project does not itself host binary distributions of OpenSSL, but does maintain an informal list of third-party distributions: https://github.com/openssl/openssl/wiki/Binaries.
+### `no-cors` mode and the `crossorigin` attribute
 
-#### Using shasum
+As a consequence of the requirement to use CORS when requesting a resource with integrity, the resource must not be requested in `no-cors` mode.
```
hamishwillee: If it were me I'd show the rule then explain the reasoning. I think people learn better that way. So something like below. Up to you though - I mean it all follows.


````suggestion
As a consequence of the requirement to use CORS when requesting a resource with integrity, the resource must not be requested in `no-cors` mode.
In other words, an element that specifies the `integrity` attribute must always include the [`crossorigin`](/en-US/docs/Web/HTML/Reference/Attributes/crossorigin) attribute as shown:

```html example-good
<script
  src="https://cdn.example.com"
  integrity="sha512-abcde"
  crossorigin="anonymous"></script>
```
````

Then "the reasonf or this is that ...

Comment (hamishwillee) [reply files/en-us/web/security/defenses/subresource_integrity/index.md]: OR, put this bit above "as well" in the intro were we are discussing this https://github.com/mdn/content/pull/43658#discussion_r3042772445 

So "the server must use CORs, so there is a requirement that the request is not make in no-cors mode and should include ^^^^. That's actually not a bad idea since then you can keep this bit all the same, but you've presented the issue and the "solution" all in the previous bit

File: files/en-us/web/security/defenses/subresource_integrity/index.md (hunk 16 above)
Comment (hamishwillee) [reply files/en-us/web/security/defenses/subresource_integrity/index.md]: Yeah. What I have is a bit more readable I think. Anyway, also suggesting an addition here in https://github.com/mdn/content/pull/43658#discussion_r3042783145

Comment (wbamberg) [reply files/en-us/web/security/defenses/subresource_integrity/index.md]: OK, alternative: https://github.com/mdn/content/pull/43658/commits/e7e937e79b234d868ae0cb1e103e1e06c2bccfa9.

Comment (hamishwillee) [reply files/en-us/web/security/defenses/subresource_integrity/index.md]: That works. Also see my comment above though.

File: files/en-us/web/security/defenses/subresource_integrity/index.md (hunk 17 above)
Comment (wbamberg) [reply files/en-us/web/security/defenses/subresource_integrity/index.md]: I agree with you that showing the rule first is better.

I'm not sure about introducing no-cors at the start though, it seems like that is a lot to hit people with right away. Might be easier for people to take in if we first talk about CORS then get into no-cors.

Comment (wbamberg) [reply files/en-us/web/security/defenses/subresource_integrity/index.md]: had a go at a version of this -> https://github.com/mdn/content/pull/43658/commits/d8c142a8dffc333020bf031f5032c5bafa1ea263. 

I appreciate that the explanation is pretty long-winded, but like you say I always want to know why and I assume others will, too.

File: files/en-us/web/security/defenses/subresource_integrity/index.md [hunk 18]
```diff
@@ -21,98 +15,95 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
...
+  crossorigin="anonymous"></script>
 ```
 
-- The pipe-through `xxd` step takes the hexadecimal output from `shasum` and converts it to binary.
-- The pipe-through `awk` step is necessary because `shasum` will pass the hashed filename in its output to `xxd`. That can have disastrous consequences if the filename happens to have valid hex characters in it — because `xxd` will also decode that and pass it to `base64`.
-
-### Cross-Origin Resource Sharing and Subresource Integrity
-
-For subresource-integrity verification of a resource served from an origin other than the document in which it's embedded, browsers additionally check the resource using [Cross-Origin Resource Sharing (CORS)](/en-US/docs/Web/HTTP/Guides/CORS), to ensure the origin serving the resource allows it to be shared with the requesting origin. Therefore, the resource must be served with an [`Access-Control-Allow-Origin`](/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin) header that allows the resource to be shared with the requesting origin; for example:
+The reason for this is that, by default, a resource loaded from a document's HTML is loaded in `no-cors` mode:
```
hamishwillee:
```suggestion
This is needed because, by default, a resource loaded from a document's HTML is loaded in `no-cors` mode:
```

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -21,98 +15,95 @@ Websites sometimes choose to rely on a third party such as a {{glossary("CDN", "
...
 
-```http
-Access-Control-Allow-Origin: *
+```html
+<script src="https://cdn.example.com"></script>
+<!-- loaded in no-cors mode -->
 ```
 
-## How browsers handle Subresource Integrity
+In `no-cors` mode, a cross-origin request will succeed even if the owner of the resource does not send the appropriate CORS headers, but the content of the response will not be shared with the requester. So a document can _use_ a resource that it requested using `no-cors`, but isn't able to _read_ it.
```
hamishwillee, line 89: This is all much more to my taste. 

A suggestion though, is that we could be more explicit about what is to follow - an explanation of the reason `no-cors` is a problem. That makes it very clear the following information is reasoning not instruction.

So perhaps prefix with something like

```suggestion
Browsers will not allow `no-cors` requests to use subresource integrity to avoid attacks that can use SRI to indirectly determine user information.
In `no-cors` mode, a cross-origin request will succeed even if the owner of the resource does not send the appropriate CORS headers, but the content of the response will not be shared with the requester. So a document can _use_ a resource that it requested using `no-cors`, but isn't able to _read_ it.
```

Just a thought.

wbamberg:
@codingjoe , I would be happy to know if this fixes https://github.com/mdn/content/issues/43302 in your opinion. It turned into a more extensive fix than I had expected.

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: I'm not very keen on this because I think it misses out the high-level purpose of SRI. It's not to check that the resource has a particular hash. It's to check that the contents of the resource are what the developer expects. The hash is just the means of doing that.

If "value" is the problem, maybe:

```suggestion
The **`integrity`** attribute provides a mechanism for a developer to assert that a linked script or stylesheet has the contents that the developer expects it to have, and has not been modified. The browser will check that the resource does in fact have the asserted contents, and refuse to load the resource if it does not.
```

?

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+This is a defense against a [supply-chain attack](/en-US/docs/Web/Security/Attacks/Supply_chain_attacks) in which an attacker gets access to the domain that serves the script or stylesheet, and changes the expected resource for a malicious one.
+
+## Description
+
+The attribute can be applied to {{htmlelement("script")}} or {{htmlelement("link")}} elements only.
+
+The attribute consists of one or more components, each of which consists of:
+
+- An identifier for a {{glossary("hash function", "cryptographic hash function")}}. Three hash functions are supported. In increasing order of strength, these are: SHA-256, SHA-384, and SHA-512.
```
codingjoe, line 22: Isn't SHA-384 just a truncated version with the same randomness as SHA-512? Strengths might be ambiguous, nor do I think this article needs to classify SHA algorithms unless one has become known to have a vulnerability.

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,83 @@
...
+- The result of hashing the resource contents using the specified hash function.
+
+When the browser downloads a resource with the `integrity` attribute set, it will first select the set of hashes that were generated using the strongest hash function present. That is, if the attribute contains values generated with SHA-256 and SHA-384, it will only use the hashes generated using SHA-384.
```
codingjoe, line 25: Is that the case? I don't remember the standard to detail this. What about caching? If your resource was cached but the browser at the time used SHA-256, will it rehash the cached resource?

codingjoe: Is that what's implied here? https://www.w3.org/TR/sri/#get-the-strongest-metadata

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,83 @@
...
+```html
+<script
+  src="https://cdn.example.com/script.js"
+  integrity="
+  sha384-Tk2Yjg3YmYzMWNkZTdhMTFkM2FlNDg4ZjE3MzEzNTk3ZDlh
+  sha384-DEzZmZhMGFkMGQ0OTQ3MzZkNGY0OTg4NGIwN2ZiMMTM3YmQ
+  sha512-ZmQ5NjNiYWJjYTM3MjRhMGI4MTQzNWRmZTZkZGYyMzQyOGYYTZkYjBm
+  sha512-OGUwYThkZDc2YzFlZGI5MDEzZmZhMGFkMGQ0OTQ3MzZkNGYZTEzODk2"
+  crossorigin="anonymous"></script>
+```
```
codingjoe: IMHO, this section needs a little more explanation.  Having two values allows for two different payloads. But the description is about picking the stronger algo, which doesn't match the example. Or at least the example could be more narrow / two examples for each different behavior.

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Including `integrity` on `<link>` elements
+
+The following {{htmlelement("link")}} element loads a stylesheet and includes an `integrity` attribute containing a single value, calculated using SHA-512.
```
codingjoe: Sidenote: I'd love to see the spec expand integrity values to hyperlinks (a-tags) and img/picture tags.

How cool would it be if I didn't need to compare hashes every time I downloaded software?

files/en-us/web/security/defenses/subresource_integrity/index.md
```diff
@@ -145,18 +136,72 @@ A typical report might look like this
...
+
+You can generate SRI hashes using [**shasum**](https://linux.die.net/man/1/shasum) with a command invocation such as:
+
+```bash
+shasum -b -a 384 FILENAME.js | awk '{ print $1 }' | xxd -r -p | base64
+```
+
+- The pipe-through `xxd` step takes the hexadecimal output from `shasum` and converts it to binary.
+- The pipe-through `awk` step is necessary because `shasum` will pass the hashed filename in its output to `xxd`. That can have disastrous consequences if the filename happens to have valid hex characters in it — because `xxd` will also decode that and pass it to `base64`.
+
```
codingjoe, line 185: Using Node

```javascript
import crypto from 'node:crypto'

/**
 * Yield integrity hashes for the given data using each of the specified algorithms.
 *
 * @param data {string|Buffer} - The data to hash.
 * @param algorithms {string[]} - The hashing algorithms to use.
 * @yields {string} - A base64-encoded hash string with algorithm prefix.
 */
export function* integrityHashes(data, algorithms = ['sha256', 'sha384', 'sha512']) {
  for (const algorithm of algorithms) {
    const hash = crypto.createHash(algorithm).update(data).digest('base64')
    yield `${algorithm.toLowerCase().replace(/-/, '')}-${hash}`
  }
}
```

codingjoe (commented):
Thanks! Yes, that's a great improvement! I left a couple of minor notes. I hope they help. - Joe

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: The spec does this though: https://w3c.github.io/webappsec-subresource-integrity/#valid-sri-hash-algorithm-token-set

I suppose "with stronger algorithms appearing later in the set" is a little bit of a weaker claim, and we could say that here.

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: That's my reading of it. Also from https://w3c.github.io/webappsec-subresource-integrity/#agility:

> the user agent will choose the strongest hash function in the list, and use that metadata to validate the response (as described below in the [§ 3.3.2 Parse metadata](https://w3c.github.io/webappsec-subresource-integrity/#parse-metadata-section) and [§ 3.3.3 Get the strongest metadata from set](https://w3c.github.io/webappsec-subresource-integrity/#get-the-strongest-metadata) algorithms).

I don't know what happens with caching. I guess the integrity would not be checked again (because the resource has already been fetched) but I don't know.

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: I think the example tries to combine both (multiple hash functions, and multiple hash values) and does mention the latter "it will only use the SHA-512 values, and **if the hash of the script at https://cdn.example.com/script.js matches either of them**"?

But I could see it might be clearer to separate into two examples.

wbamberg: I suppose the reason I combined was that I wanted people to understand that you can combine them. So it might be better to have:
- one example showing multiple hash functions
- another showing multiple hash functions and multiple hash values

Having said that though, I'd like to know better what the best practice guidance is for multiple hash functions, so we can embody that in the examples. I'll try asking.

codingjoe: Since the example repeats in another file, maybe we can have two here to add variety and detail.

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: I think I'm going to keep this, because the spec does define a "strength-ordering" and uses it to select an algorithm, and we do need to explain this.

files/en-us/web/html/reference/attributes/integrity/index.md
wbamberg: OK, I just pushed https://github.com/mdn/content/pull/43658/commits/4547ad38a0576dd5bdc48a3bbab799a3f06dc0fd, in which we have:

- two examples for the `<script>` case, one with multiple hash functions, one with multiple file values, and talks about the behaviour in each case.
- one example for the `<link>` case, which combines both, to show that this is possible too.

wbamberg: > Having said that though, I'd like to know better what the best practice guidance is for multiple hash functions, so we can embody that in the examples. I'll try asking.

I have asked about this, but don't want to block on an answer (IME it can take a while to get an answer). I'll file an update later if it seems to be needed.

files/en-us/web/html/reference/attributes/integrity/index.md
hamishwillee: I'm happier with that ^^^

Perhaps the simplest update is this. Select which you prefer.


```suggestion
The **`integrity`** attribute provides a mechanism for a developer to assert that a linked script or stylesheet must have a particular content. The browser will check that the resource does in fact have that content, and refuse to load the resource if it does not.
```

files/en-us/web/html/reference/attributes/integrity/index.md
```diff
@@ -40,32 +40,62 @@ In each case, the part preceding `-` identifies the {{glossary("hash function")}
 
 ## Examples
 
-### Including `integrity` on `<script>` elements
+### Including different hash functions
```
hamishwillee, line 43: FYI these example changes are great

hamishwillee (approved):

files/en-us/web/html/reference/attributes/integrity/index.md
codingjoe: Yes, that seems correct in relation to the spec. Thanks for checking.

Issue #43673, opened 2026-04-06
include_subdomains description for NEL header is inaccurate


Slonser:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Network_Error_Logging

### What specific section or headline is this issue about?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Network_Error_Logging#include_subdomains

### What information was incorrect, unhelpful, or incomplete?

The current description of `include_subdomains` in the [Network Error Logging guide](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Network_Error_Logging) states:

> If true, the policy applies to all subdomains under the origin that the policy header is set.

### What did you expect to see?

Per the [NEL spec](https://w3c.github.io/network-error-logging/#include-subdomains) and [w3c/network-error-logging#201](https://github.com/w3c/network-error-logging/issues/201), `include_subdomains` only enables the NEL policy for subdomains for `dns` phase reports. If the report phase is not `dns`, the policy will not apply to subdomains regardless of this flag.

The description should be updated to reflect this limitation.

PR #43674, opened 2026-04-06
Fix `include_subdomains` description for NEL header

Slonser:
### Description

Fix `include_subdomains` description for the NEL header to clarify that it only applies to dns phase reports, not all report types.

### Motivation

The previous description stated that `include_subdomains` applies the policy to "all subdomains under the origin", without mentioning that this is limited to dns phase reports only. This was inaccurate per the spec and could mislead developers.

### Additional details

Per the [NEL spec](https://w3c.github.io/network-error-logging/#include-subdomains), the `include_subdomains` member enables the policy for subdomains only for `dns` phase reports. If the member is not present, is false, or the report phase is not `dns`, the policy will not apply to subdomains.

Fixes https://github.com/mdn/content/issues/43673

### DIFF
--- a/files/en-us/web/http/guides/network_error_logging/index.md
+++ b/files/en-us/web/http/guides/network_error_logging/index.md
@@ -33,7 +33,7 @@ The following object keys can be specified in the NEL header:
 - max_age
   - : Specifies the lifetime of the policy, in seconds (in a similar way to e.g., HSTS policies are time-restricted). The referenced reporting group should have a lifetime at least as long as the NEL policy.
 - include_subdomains
-  - : If `true`, this NEL policy is also enabled for all subdomains of the origin, but only for network errors that occur during DNS resolution.  The NEL policy will not be enabled for subdomains if `include_subdomains` is not present, is `false`, or for other (non-DNS-related) network errors. The reporting group must also be set to include subdomains for this option to have effect.
+  - : If `true`, this NEL policy is also enabled for all subdomains of the origin, but only for network errors that occur during DNS resolution. The NEL policy will not be enabled for subdomains if `include_subdomains` is not present, is `false`, or for other (non-DNS-related) network errors. The reporting group must also be set to include subdomains for this option to have effect.
 - success_fraction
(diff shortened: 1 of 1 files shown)


files/en-us/web/http/guides/network_error_logging/index.md
```diff
@@ -33,7 +33,7 @@ The following object keys can be specified in the NEL header:
 - max_age
   - : Specifies the lifetime of the policy, in seconds (in a similar way to e.g., HSTS policies are time-restricted). The referenced reporting group should have a lifetime at least as long as the NEL policy.
 - include_subdomains
-  - : If true, the policy applies to all subdomains under the origin that the policy header is set. The reporting group should also be set to include subdomains, if this option is to be enabled.
+  - : If true, enables this NEL policy for all subdomains of the origin (to an unlimited subdomain depth), but only for `dns` phase reports. If `include_subdomains` is not present, is false, or the report phase is not `dns`, the NEL policy will not be enabled for subdomains. The reporting group should also be set to include subdomains if this option is to be enabled.
```
hamishwillee: - "(to an unlimited subdomain depth)" can be assumed.
- "`dns` phase reports" doesn't necessarily mean anything to a reader as they don't have the spec background.
- The reporting group must (not should) enable include subdomains. 

How about?

```suggestion
  - : If `true`, this NEL policy is also enabled for all subdomains of the origin, but only for network errors that occur during DNS resolution.  The NEL policy will not be enabled for subdomains if `include_subdomains` is not present, is `false`, or for other (non-DNS-related) network errors. The reporting group must also be set to include subdomains for this option to have effect.
```

As an aside, 
- Report-To is deprecated, and I wonder if the implementation has been updated to use the newer headers. 
- Most of this information should move into the NEL header doc. Perhaps as a post-process.

Slonser: First of all, yes, these remarks are fair (merged). 
As for the question about Report-To, NEL works correctly together with it in WebKit, Chromium, and Firefox.

hamishwillee: I'm sure it does. The question is whether it also works with the `Reporting-Endpoint`. That's a question for another day though.

PR #43684, opened 2026-04-07
Ff150 `<img>` attribute `sizes` takes value `auto`

hamishwillee:
FF150 adds `HTMLImageElement.sizes` value of `auto` and the corresponding attribute on [``](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#sizes) in https://bugzilla.mozilla.org/show_bug.cgi?id=1819581. 

This updates the docs to 
- make the attribute easier to understand in the `` element
- Better cross link from the HTMLIMageElement.sizes property, and add an example to make it easier to see what this does.

Related docs work can be tracked in #43562


files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +104,97 @@ break50.addEventListener(
...
 
+The page is best {{LiveSampleLink('Selecting an image to fit window width', 'viewed in its own window')}}, so you can adjust the sizes fully (otherwise it is constrained by its containing frame).
+
+1. Enable the developer tools and change the width of the page — you should see the image change size at the sizes media query trigger points: 480px (30em), and 800px (50em).
+2. Set the width between 50em (800px) and 60em (960px) so that the last media query is selected. Then alternately press each of the buttons and note how the layout size of the image is changed.
+
 {{EmbedLiveSample("Selecting an image to fit window width", "", 1050)}}
 
-The page is best {{LiveSampleLink('Selecting an image to fit window width', 'viewed in its own window')}}, so you can adjust the sizes fully.
+### Ex - Slider
```
hamishwillee: This and the images are test code (at this point) shows a slider that allows you to see the selection of a new image.

Thing is, only works on FF because chrome only loads the image ONCE. FF seems to load a better image if the media query changes.

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
+        Source size values specify the intended display size of the image.
+        {{glossary("User agent", "User agents")}} use the current source size to select one of the sources supplied by the `srcset` attribute, when those sources are described using width (`w`) descriptors.
+        The selected source size affects the {{glossary("intrinsic size")}} of the image (the image's display size if no {{glossary("CSS")}} styling is applied).
 
-    The `auto` keyword can replace the whole list of sizes or the first entry in the list. It is only valid when combined with `loading="lazy"`, and resolves to the [concrete size](/en-US/docs/Web/CSS/Reference/Values/image) of the image. Since the intrinsic size of the image is not yet known, `width` and `height` attributes (or CSS equivalents) should also be specified to prevent the browser from assuming the default image width of 300px.
```
hamishwillee, line 210: FYI I didn't think this was quite right. The browser might now know intrinsic sizes at some points but it doesn't need to - it falls back to the default intrinsic size for images in the ua stylesheet. Also this is not a factor of "auto" so much as the fallback when auto can't work. Such as if you have a lazy loaded image but it is displayed in the viewport on load (so it isn't lazy loaded)

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
 
-    A **source size** consists of:
-    1. A [media condition](/en-US/docs/Web/CSS/Guides/Media_queries/Using#syntax), omitted for the last item in the list.
-    2. A source size value.
+        For example, the following source size proposes using an image source of 1000px width if the _viewport_ height is 500px or less.
 
-    Media conditions describe properties of the _viewport_, not the _image_. For example, `(height <= 500px) 1000px` proposes using an image source of 1000px width if the _viewport_ height is 500px or less. Because a source size descriptor specifies the width to use for the image during layout, the media condition is typically (but not necessarily) based on the {{cssxref("@media/width")}}.
+        ```css
+        (height <= 500px) 1000px
+        ```
```
hamishwillee, line 209: This section is the same information put into a definition list. The only change in structure otherwise is pulling out this example of what a query looks like.

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -8,17 +8,24 @@ browser-compat: api.HTMLImageElement.sizes
...
+The **`sizes`** property of the {{domxref("HTMLImageElement")}} interface allows you to specify the layout width of the [image](/en-US/docs/Web/HTML/Reference/Elements/img) for each of a list of [media queries](/en-US/docs/Web/CSS/Guides/Media_queries), or `auto` for lazy-loaded images to allow the browser to automatically select an image to display based on the layout size of the element.
+This provides the ability to automatically select among different images specified in the element {{domxref("HTMLImageElement/srcset","srcset")}} — even images of different orientations or aspect ratios — as the document state changes to match different media conditions.
```
chrisdavidmills:
```suggestion
This allows the browser to choose between different images specified in the element {{domxref("HTMLImageElement/srcset", "srcset")}} — even images with different orientations or aspect ratios — as the document state changes to match different media conditions.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -8,17 +8,24 @@ browser-compat: api.HTMLImageElement.sizes
...
+
+It reflects the `<img>` element's [`sizes`](/en-US/docs/Web/HTML/Reference/Elements/img#sizes) content attribute.
```
chrisdavidmills:
```suggestion
The `sizes` property reflects the `<img>` element's [`sizes`](/en-US/docs/Web/HTML/Reference/Elements/img#sizes) content attribute.
```

Felt a bit weird to have two sentences in a row beginning with "It..."

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -8,17 +8,24 @@ browser-compat: api.HTMLImageElement.sizes
...
+It can only be present when `srcset` uses width descriptors.
 
 ## Value
 
-A string. For more information about the syntax of the `sizes` attribute, see the HTML [`<img>`](/en-US/docs/Web/HTML/Reference/Elements/img#sizes) reference.
+A string containing one or more comma-separated values, which can be the `auto` keyword and/or any number of _source sizes_.
```
chrisdavidmills:
```suggestion
A string containing one or more comma-separated values, each of which can be the `auto` keyword and/or a _source size_.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
 
 #### Result
 
+The page is best {{LiveSampleLink('Selecting an image to fit window width', 'viewed in its own window')}}, so you can adjust the sizes fully (otherwise it is constrained by its containing frame).
```
chrisdavidmills:
```suggestion
The page is best {{LiveSampleLink('Selecting an image to fit window width', 'viewed in its own window')}}, so you can adjust the sizes fully, and the example is not constrained by its containing frame.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+
+1. Enable the developer tools and change the width of the page — you should see the image change size at the sizes media query trigger points: 480px (30em), and 800px (50em).
```
chrisdavidmills: All the length values should be put in code font.

chrisdavidmills: Also, I've got a couple of issues with this explanation:

1. You don't mention that the image itself changes, as well as the size. This is covered in the example HTML section, but it would be helpful to clarify it here.
2. You don't get to see how the width changes at 480px wide, as the width only goes down to 500px. At least, this was my experience on Mac, on both Firefox and Chrome. Maybe worth experimenting with using a different trigger point that is more likely to be seen on desktop computers?

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+2. Set the width between 50em (800px) and 60em (960px) so that the last media query is selected. Then alternately press each of the buttons and note how the layout size of the image is changed.
+
 {{EmbedLiveSample("Selecting an image to fit window width", "", 1050)}}
 
-The page is best {{LiveSampleLink('Selecting an image to fit window width', 'viewed in its own window')}}, so you can adjust the sizes fully.
+### Automatic image selection for lazy loaded images
+
+This example demonstrates how setting the `sizes` value to `auto` affects selection of the image to load from the [`srcset`](/en-US/docs/Web/HTML/Reference/Elements/img#srcset) of lazy-loaded {{htmlelement("img")}} elements.
```
chrisdavidmills:
```suggestion
This example demonstrates how setting the `sizes` value to `auto` affects the selection of the image to load from the [`srcset`](/en-US/docs/Web/HTML/Reference/Elements/img#srcset) when {{htmlelement("img")}} elements are lazy-loaded.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+It also allows you to see the effect of changing the size of a container on the loaded image.
+
+#### HTML
+
+In order to demonstrate the effect of lazy loading the images need to be initially hidden from the visual viewport, and then scrolled into view.
+This is achieved by having an outer `scroll-container` {{htmlelement("div")}} that nests `spacer` and `demo-wrap` containers.
+The images are nested within containers inside the `demo-wrap`, which is pushed out of the visual viewport by the `spacer`.
```
chrisdavidmills:
```suggestion
The images are contained inside the `demo-wrap` container, which is pushed out of the visual viewport by the height set on the `spacer` container.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+
+The three {{htmlelement("img")}} elements have different `alt` properties, but are otherwise identical:
```
chrisdavidmills:
```suggestion
The three {{htmlelement("img")}} elements have different `alt` attribute values, but are otherwise identical:
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+
+- `srcset` defines three images and indicates that they are 600px, 400px, and 200px wide.
+- `src` specifies the image that will be used if `srcset` is not supported or it can't be parsed.
+  We use the largest image in the `srcset` as this will almost always downscale better than the smallest image will upscale.
```
chrisdavidmills:
```suggestion
  This is the largest image in the `srcset` as it will almost always downscale better than the smallest image will upscale.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+  height: 100px;
+  overflow: scroll;
+  padding: 0.5rem;
+  border: 1px solid black;
+}
+```
+
+#### CSS
+
+Here we show the size of CSS that sets the size of the image containers.
```
chrisdavidmills:
```suggestion
Here we show the CSS classes that set the size of the different image containers.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+    root: document.getElementById("scroll-container"),
+    rootMargin: "0px",
+    threshold: 0.1,
+  },
+);
+
+images.forEach((img) => observer.observe(img));
+```
+
+The remaining CSS and the code for the slider, logging, and so on are not shown (if you are interested select "Play" to view the whole example in the interactive playground).
```
chrisdavidmills:
```suggestion
The remaining CSS and the JavaScript that powers the slider, logging, and so on, are not shown (if you are interested in examining these, select "Play" to view the whole example in the interactive playground).
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+
+#### Result
+
+Scroll the frame to display the three images.
+The browser should have selected a different image for each based on the constrained layout size.
```
chrisdavidmills:
```suggestion
The browser should have selected a different image for each based on the different width constraints.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+You can use the slider to modify the size of the container for the first image.
+Note that the browser may or may not select a new image to display as the size of the container changes.
```
chrisdavidmills: Is this due to browser implementation differences, or something else?

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+
+{{EmbedLiveSample("Automatic image selection for lazy loaded images", "", 600)}}
+
+The log provides information when a `load` event fires for each image, and when an image intersects the visible viewport.
+Note that the images are lazy loaded, so the load event should be fired just before the image enters the viewport.
```
chrisdavidmills:
```suggestion
Note that the images are lazy-loaded, so the `load` event should be fired just before the image enters the viewport.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+Also note that the event fires as you modify the container size for the first image, indicating when the browser has recalculated the layout (not necessarily that a new image has been loaded).
```
chrisdavidmills:
```suggestion
Also note that the `load` event also fires as you modify the container size for the first image, indicating when the browser has recalculated the layout (not necessarily that a new image has been loaded).
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
+The remaining CSS and the code for the slider, logging, and so on are not shown (if you are interested select "Play" to view the whole example in the interactive playground).
+
+#### Result
+
+Scroll the frame to display the three images.
+The browser should have selected a different image for each based on the constrained layout size.
+You can use the slider to modify the size of the container for the first image.
+Note that the browser may or may not select a new image to display as the size of the container changes.
+
+{{EmbedLiveSample("Automatic image selection for lazy loaded images", "", 600)}}
```
chrisdavidmills, line 404: I'm assuming this demo works. The images aren't loading on the preview link, which is a bit frustrating.

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
+    The spec requires that the `sizes` attribute only be present when `srcset` uses width descriptors.
+    - **source size**
+      - : A **source size** consists of:
+        1. A [media condition](/en-US/docs/Web/CSS/Guides/Media_queries/Using#syntax), omitted for the last item in the list.
+        2. A source size value.
 
-    A **source size** consists of:
-    1. A [media condition](/en-US/docs/Web/CSS/Guides/Media_queries/Using#syntax), omitted for the last item in the list.
-    2. A source size value.
+        For example, the following source size proposes using an image source of 1000px width if the _viewport_ height is 500px or less.
```
```suggestion
        For example, the following source size proposes using a `1000px`-wide image source if the _viewport_ height is 500px or less.
```

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
 
-    Media conditions describe properties of the _viewport_, not the _image_. For example, `(height <= 500px) 1000px` proposes using an image source of 1000px width if the _viewport_ height is 500px or less. Because a source size descriptor specifies the width to use for the image during layout, the media condition is typically (but not necessarily) based on the {{cssxref("@media/width")}}.
+        ```css
+        (height <= 500px) 1000px
+        ```
 
-    Source size values specify the intended display size of the image. {{glossary("User agent", "User agents")}} use the current source size to select one of the sources supplied by the `srcset` attribute, when those sources are described using width (`w`) descriptors. The selected source size affects the {{glossary("intrinsic size")}} of the image (the image's display size if no {{glossary("CSS")}} styling is applied).
+        Media conditions describe properties of the _viewport_, not the _image_.
+        Because a source size descriptor specifies the width to use for the image during layout, the media condition is typically (but not necessarily) based on the {{cssxref("@media/width")}}.
```
chrisdavidmills, line 212: In which case, it seems weird to show a `height` viewport example earlier on. I mean, you are nearly never going to want to do that.

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
 
-    A source size value can be any non-negative [length](/en-US/docs/Web/CSS/Reference/Values/length). It must not use CSS functions other than the [math functions](/en-US/docs/Web/CSS/Reference/Values/Functions#math_functions). Units are interpreted in the same way as [media queries](/en-US/docs/Web/CSS/Guides/Media_queries), meaning that all relative length units are relative to the document root rather than the `<img>` element. For example, an `em` value is relative to the root font size, not the font size of the image. [Percentage](/en-US/docs/Web/CSS/Reference/Values/percentage) values are not allowed. If the `sizes` attribute is not provided, it has a default value of `100vw` (the viewport width).
+        Source size values specify the intended display size of the image.
+        {{glossary("User agent", "User agents")}} use the current source size to select one of the sources supplied by the `srcset` attribute, when those sources are described using width (`w`) descriptors.
+        The selected source size affects the {{glossary("intrinsic size")}} of the image (the image's display size if no {{glossary("CSS")}} styling is applied).
```
chrisdavidmills: "affects" seems like a weird word to use here — you can't change an image's intrinsic size. I'm assuming what this really means is that it specifies the intrinsic size of the image that should be chosen.

So maybe something like this:


```suggestion
        The selected source size specifies the {{glossary("intrinsic size")}} of the image that should be chosen (the image's display size if no {{glossary("CSS")}} styling is applied).
```

?

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
-        swing-800.jpg   800w,
-        swing-1600.jpg 1600w
-      "
-      src="swing-400.jpg"
-      alt="Kettlebell Swing" />
-    ```
+    - `auto`
+      - : The `auto` keyword indicates that the browser should use the expected layout width of the element to select the image to display.
+        That is, it should use the [concrete size](/en-US/docs/Web/CSS/Reference/Values/image) of the image, calculated after layout from HTML and CSS has been applied.
+        This is only valid when combined with `loading="lazy"`, as the page can be expected to already have CSS and other layout information by the time the image is loaded.
```
```suggestion
        This is only valid when combined with `loading="lazy"`, as the page is expected to already have CSS and other layout information by the time the image loads.
```

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
+
+        Using `auto` saves you having to specify your layout media conditions twice: once for layout, and once for selection of an appropriate image to fetch and display.
+
+        If `auto` cannot resolve — either because the browser does not support it, or because the image has no layout size yet — the browser falls through to the fallback _source sizes_ in the list to determine the width, then to `width`/`height` attributes defined on the element, and finally to the default intrinsic size for `<img>` elements defined in the user agent stylesheet (300px by 150px).
```
```suggestion
        If `auto` cannot resolve — either because the browser does not support it, or because the image has no layout size yet — the browser falls back to the _source sizes_ in the list to determine the width, then to `width`/`height` attributes defined on the element, and finally to the default intrinsic size for `<img>` elements defined in the user agent stylesheet (300px by 150px).
```

chrisdavidmills (changes requested):
@hamishwillee nice work, man. Bunch of comments for you to consider.

files/en-us/web/api/htmlimageelement/sizes/index.md
SebastianZ: I actually have an issue with the example itself. It uses `em` units, which depend on the font size chosen by the user. So 30em are only 480px in the default case of 16px = 1em.

The concept behind the `sizes` attribute is already complicated enough.
Therefore, I'd rather replace the current example by a simpler (contrived) one that

* uses pixels instead of `em`s.
* uses only two instead of three size options.
* uses different images for the size options (like you did for the `auto` value).
* works for a huge range of devices by choosing the size option ranges wisely.

Or maybe instead of completely replacing the example, it should be moved to the end and the simpler one described here be added as the first one.

files/en-us/web/api/htmlimageelement/sizes/index.md
SebastianZ: It might be unclear what is meant by "downscale better", so I'd rather suggest

```suggestion
  This is the largest image in the `srcset` as the visual result from downscaling it for smaller layout sizes will almost always be better than upscaling the smallest image for larger layout sizes.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
SebastianZ: Yeah, it's obviously up to the browsers when to run the algorithm, as [Hamish pointed out in the related Bugzilla bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1819581#c18), see https://html.spec.whatwg.org/multipage/images.html#reacting-to-environment-changes. (Though the spec also states: "User agents are encouraged to run this algorithm in particular when the user changes the viewport's size".)

@hamishwillee Maybe the sentence should clarify that circumstance.

```suggestion
Note that the browser may or may not select a new image to display as the size of the container changes as implementations are not required to react to dynamic changes.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +111,240 @@ break50.addEventListener(
...
 
-The page is best {{LiveSampleLink('Selecting an image to fit window width', 'viewed in its own window')}}, so you can adjust the sizes fully.
+### Automatic image selection for lazy loaded images
+
+This example demonstrates how setting the `sizes` value to `auto` affects selection of the image to load from the [`srcset`](/en-US/docs/Web/HTML/Reference/Elements/img#srcset) of lazy-loaded {{htmlelement("img")}} elements.
+It also allows you to see the effect of changing the size of a container on the loaded image.
+
+#### HTML
+
+In order to demonstrate the effect of lazy loading the images need to be initially hidden from the visual viewport, and then scrolled into view.
```
SebastianZ:
```suggestion
In order to demonstrate the effect of lazy loading the images need to be initially hidden from the {{glossary("visual viewport")}}, and then scrolled into view.
```

files/en-us/web/html/reference/elements/img/index.md
SebastianZ:
```suggestion
      - : Loads the image immediately, regardless of whether or not the image is currently within the {{glossary("visual viewport")}} (this is the default value).
```

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
-    2. A source size value.
+        For example, the following source size proposes using an image source of 1000px width if the _viewport_ height is 500px or less.
 
-    Media conditions describe properties of the _viewport_, not the _image_. For example, `(height <= 500px) 1000px` proposes using an image source of 1000px width if the _viewport_ height is 500px or less. Because a source size descriptor specifies the width to use for the image during layout, the media condition is typically (but not necessarily) based on the {{cssxref("@media/width")}}.
+        ```css
+        (height <= 500px) 1000px
+        ```
 
-    Source size values specify the intended display size of the image. {{glossary("User agent", "User agents")}} use the current source size to select one of the sources supplied by the `srcset` attribute, when those sources are described using width (`w`) descriptors. The selected source size affects the {{glossary("intrinsic size")}} of the image (the image's display size if no {{glossary("CSS")}} styling is applied).
+        Media conditions describe properties of the _viewport_, not the _image_.
```
SebastianZ:
```suggestion
        Media conditions describe properties of the _{{glossary("viewport")}}_, not the _image_.
```

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
-        swing-400.jpg   400w,
-        swing-800.jpg   800w,
-        swing-1600.jpg 1600w
-      "
-      src="swing-400.jpg"
-      alt="Kettlebell Swing" />
-    ```
+    - `auto`
+      - : The `auto` keyword indicates that the browser should use the expected layout width of the element to select the image to display.
+        That is, it should use the [concrete size](/en-US/docs/Web/CSS/Reference/Values/image) of the image, calculated after layout from HTML and CSS has been applied.
```
SebastianZ: Maybe link to the description section of the `<image>` value, where "concrete size" is defined.

```suggestion
        That is, it should use the [concrete size](/en-US/docs/Web/CSS/Reference/Values/image#concrete_size) of the image, calculated after layout from HTML and CSS has been applied.
```

SebastianZ (changes requested):

files/en-us/web/api/htmlimageelement/sizes/index.md
hamishwillee: Thanks, slight modification because this wasn't quite correct (originally). The way it read the selection adjusts as the document changes _state_ for media conditions.
In fact this is what happens for FF and partially for chrome, but it is not required by the spec. That basically says you evaluate the selection based on the media query at the point the image is loaded. It doesn't require that the load is revaluated as the media conditions change.  I don't clarify the detail but this makes the requirement more clear.

```suggestion
This allows the browser to choose between different images specified in the element {{domxref("HTMLImageElement/srcset", "srcset")}} to match different media conditions — even images with different orientations or aspect ratios.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
hamishwillee: No, it can be just auto, or start with auto and have any number of source sizse, or be any number of source sizes and not the auto keyword. You can only specify auto at the start an only once 


```suggestion
A string containing that is can be the `auto` keyword (optionally followed by any number of _source sizes_), or one or more _source sizes_.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
hamishwillee: It does!

files/en-us/web/html/reference/elements/img/index.md
hamishwillee: No, this actually means what it says, though it's a bit confusing. Generally we think about the intrinsic size as the raw pixel dimensions. But when using a srcset the width in the `sizes` takes this role. So if you specify a w value of 500 px that many pixels are reserved, and whatever image you download is scaled to fit into that space.

Perhaps we can avoid overloading the term intrinsic size - better?

```suggestion
        The `w` value defined in sizes determines the image’s default layout width.
        In the absence of {{glossary("CSS")}}, the browser will render the image at this size, regardless of the physical pixel dimensions of the downloaded file.
```

files/en-us/web/html/reference/elements/img/index.md
hamishwillee: It's a good idea. I added an explicit heading as a link target and modified your suggestion ^^^. Thank you.

files/en-us/web/html/reference/elements/img/index.md
hamishwillee: This is good, but github is showing it to me as an addition to an empty line (i.e. some kind of github bug). Setting resolved but I will make this change manually.

files/en-us/web/html/reference/elements/img/index.md
hamishwillee: Yeah, that predates my changes.  MOdified the query to condition on the width (but did not change the width) in https://github.com/mdn/content/pull/43684/changes/59f86f56d36a3ddf134e32172cbad50e75c50fe9

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
+    The spec requires that the `sizes` attribute only be present when `srcset` uses width descriptors.
+    - **source size**
+      - : A **source size** consists of:
+        1. A [media condition](/en-US/docs/Web/CSS/Guides/Media_queries/Using#syntax), omitted for the last item in the list.
+        2. A source size value.
 
-    A **source size** consists of:
-    1. A [media condition](/en-US/docs/Web/CSS/Guides/Media_queries/Using#syntax), omitted for the last item in the list.
-    2. A source size value.
+        For example, the following source size proposes using a `1000px`-wide image source if the _viewport_ height is 500px or less.
```
hamishwillee:
```suggestion
        For example, the following source size proposes using a `1000px`-wide image source if the _viewport_ width is 500px or less.
```

files/en-us/web/html/reference/elements/img/index.md
```diff
@@ -195,36 +195,57 @@ This element includes the [global attributes](/en-US/docs/Web/HTML/Reference/Glo
...
 
-    Media conditions describe properties of the _viewport_, not the _image_. For example, `(height <= 500px) 1000px` proposes using an image source of 1000px width if the _viewport_ height is 500px or less. Because a source size descriptor specifies the width to use for the image during layout, the media condition is typically (but not necessarily) based on the {{cssxref("@media/width")}}.
+        ```css
+        (height <= 500px) 1000px
```
hamishwillee:
```suggestion
        (width<= 500px) 1000px
```

files/en-us/web/api/htmlimageelement/sizes/index.md
hamishwillee: @chrisdavidmills Thanks! Fixed up as you suggested. I moved this example to the end.

@SebastianZ Thanks, I have created a new "first" example which uses sizes media queries. Please check it out. 

I found this hard because the behaviour depends on your particular hardware - on my computer you end up with a double-size image being selected to what you might expect because it has a DPR of 2 (on FF). Chrome behaves differently again, as it seems to detect a DPR of 1.5.
I have explained this briefly, but it means you end up with more images that you would expect in the srcset and media query triggers that select different images on different browsers. Not good for examples.

Anyhow, see what you think.

files/en-us/web/html/reference/elements/img/index.md
chrisdavidmills: This is much better, thanks! I'm going to commit this suggestion.

(I've not looked at this stuff for ages, and had forgotten how confusing it is)

files/en-us/web/html/reference/elements/img/index.md
chrisdavidmills: Perfect, this makes a lot more sense now. I edited the file to add a space to make the code a bit neater.

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -8,17 +8,423 @@ browser-compat: api.HTMLImageElement.sizes
...
+The **`sizes`** property of the {{domxref("HTMLImageElement")}} interface allows you to specify the layout width of the [image](/en-US/docs/Web/HTML/Reference/Elements/img) for each of a list of [media queries](/en-US/docs/Web/CSS/Guides/Media_queries), or `auto` for lazy-loaded images to allow the browser to automatically select an image to display based on the layout size of the element.
+This allows the browser to choose between different images specified in the element {{domxref("HTMLImageElement/srcset", "srcset")}} to match different media conditions — even images with different orientations or aspect ratios.
+
+The `sizes` property reflects the `<img>` element's [`sizes`](/en-US/docs/Web/HTML/Reference/Elements/img#sizes) content attribute.
+It can only be present when `srcset` uses width descriptors.
 
 ## Value
 
-A string. For more information about the syntax of the `sizes` attribute, see the HTML [`<img>`](/en-US/docs/Web/HTML/Reference/Elements/img#sizes) reference.
+A string containing that is can be the `auto` keyword (optionally followed by any number of _source sizes_), or one or more _source sizes_.
```
chrisdavidmills:
```suggestion
A string containing that can be the `auto` keyword (optionally followed by any number of _source sizes_), or one or more _source sizes_.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -8,17 +8,423 @@ browser-compat: api.HTMLImageElement.sizes
...
+
+#### HTML
+
+In order to demonstrate the effect of lazy loading the images need to be initially hidden from the {{glossary("visual viewport")}}, and then scrolled into view.
+This is achieved by having an outer `scroll-container` {{htmlelement("div")}} that nests `spacer` and `demo-wrap` containers.
+The image is contained inside the `demo-wrap` container, which is pushed out of the visual viewport by the height set on the `spacer` container.
+
+The {{htmlelement("img")}} element has the following attributes:
+
+- `srcset` defines four images and indicates that they are 600px, 900px, 1200px, and 1500px wide.
```
chrisdavidmills: code font for the values?

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -8,17 +8,423 @@ browser-compat: api.HTMLImageElement.sizes
...
+
+The CSS and JavaScript are not shown (if you want to examine these, select "Play" to view the whole example in the interactive playground).
+
+#### Result
+
+The example is best {{LiveSampleLink('Selecting an image to fit window width', 'viewed in its own window')}}, so you can adjust the sizes fully, and the example is not constrained by its containing frame.
+
+1. Scroll the frame to display the image.
+   The label at the bottom of the image shows the current container width.
+2. Resize the window — you should see the image change at the `sizes` attribute's media query trigger points.
```
chrisdavidmills:
```suggestion
2. Resize the window — you should see the image change at the `sizes` attribute's media query break points.
```

For consistency?

Or am I wrong?

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -8,17 +8,423 @@ browser-compat: api.HTMLImageElement.sizes
...
+Note that the image is lazy-loaded, so the `load` event should be fired just before the image enters the viewport.
+
+### Automatic image selection for lazy loaded images
+
+This example demonstrates how setting the `sizes` value to `auto` affects the selection of the image to load from the [`srcset`](/en-US/docs/Web/HTML/Reference/Elements/img#srcset) when {{htmlelement("img")}} elements are lazy-loaded.
+It also allows you to see the effect of changing the size of a container on the loaded image.
+
+#### HTML
+
+In order to demonstrate the effect of lazy loading the images need to be initially hidden from the {{glossary("visual viewport")}}, and then scrolled into view.
```
chrisdavidmills: Given that this example is very similar to the previous one, you can probably cut down on some of this explanation, and just say that it is similar to the previous example, blah blah.

chrisdavidmills (approved):
@hamishwillee looking great, thanks @hamishwillee! I had a few more comments for you, but I don't think I need to see this again. Approving.

files/en-us/web/api/htmlimageelement/sizes/index.md
hamishwillee:
```suggestion
- `srcset` defines four images and indicates that they are `600px`, `900px`, `1200px`, and `1500px` wide.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
hamishwillee: No, I think that is the normal terminology. Getting old :-)

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -8,17 +8,423 @@ browser-compat: api.HTMLImageElement.sizes
...
+   The label at the bottom of the image shows the current container width.
+2. Resize the window — you should see the image change at the `sizes` attribute's media query trigger points.
+
+   Note that the the selected image may be larger than the container width alone suggests.
+   Many displays, if not most, have a [device pixel ratio (DPR)](/en-US/docs/Web/API/Window/devicePixelRatio) greater than one.
+   In order to render a sharp image at the physical pixel density of the display, a browser will multiply the matched `sizes` hint by the DPR before selecting from `srcset`.
+   For example, on a 2× display with a viewport of ~500px, the matched hint is `600px`, but the browser looks for a ~1200px image and selects `1200.png` as the closest available size and then scales it to fit in the available space.
+
+   > [!NOTE]
+   > As a result, some of the images in the `srcset` may not be reachable on a particular display at some trigger points, and this may be browser dependent.
```
hamishwillee:
```suggestion
   > As a result, some of the images in the `srcset` may not be reachable on a particular display at some break points, and this may be browser dependent.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -104,9 +510,13 @@ break50.addEventListener(
 
 #### Result
 
-{{EmbedLiveSample("Selecting an image to fit window width", "", 1050)}}
+The page is best {{LiveSampleLink('Blog example', 'viewed in its own window')}}, so you can adjust the sizes fully, and the example is not constrained by its containing frame.
+
+1. Enable the developer tools and change the width of the page — you should see the image change (and jump in size) at the sizes media query trigger points: `640px` (`40em`), and `800px` (`50em`).
```
hamishwillee:
```suggestion
1. Enable the developer tools and change the width of the page — you should see the image change (and jump in size) at the sizes media query break points: `640px` (`40em`), and `800px` (`50em`).
```

files/en-us/web/api/htmlimageelement/sizes/index.md
hamishwillee:
```suggestion
A string that can be the `auto` keyword (optionally followed by any number of _source sizes_), or one or more _source sizes_.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
```diff
@@ -8,17 +8,423 @@ browser-compat: api.HTMLImageElement.sizes
...
+The three {{htmlelement("img")}} elements have different `alt` attribute values, but are otherwise identical:
+
+- `srcset` defines three images and indicates that they are 600px, 400px, and 200px wide.
+- `src` specifies the image that will be used if `srcset` is not supported or it can't be parsed.
+  This is the largest image in the `srcset` as the visual result from downscaling it for smaller layout sizes will almost always be better than upscaling the smallest image for larger layout sizes.
+- `loading` is `lazy`.
+- `sizes` is `auto`.
+  This tells the browser to select the image that is most appropriate based on the layout information it has at the time the image intersects with the visual viewport.
+
+The three images are constrained within containers that are sized to select different images.
```
hamishwillee:
```suggestion
The HTML is similar to that in the previous example, except that it defines three near-identical {{htmlelement("img")}} elements, each with a `scrset` indicating 3 images that are `600px`, `400px`, and `200px` wide, and with a `sizes` value of `auto`.
These are constrained within containers that are sized to select the different images.
```

files/en-us/web/api/htmlimageelement/sizes/index.md
hamishwillee: Not certain if it was worth it, but done in https://github.com/mdn/content/pull/43684/changes#r3076728088 :-)

hamishwillee:
Thanks very much @chrisdavidmills and @SebastianZ  - appreciate your help and reviews. I'm going to merge, but can iterate in a follow up if needed.

PR #43712, opened 2026-04-08
Separate active and legacy Firefox platforms in UA reference

### LINKED ISSUE #43642: Removing or reducing legacy info on Firefox UA string reference
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent/Firefox

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent/Firefox#firefox_for_fire_tv and below

There's a lot of content for products no longer shipping, as FirefoxOS, or Firefox for Fire TV & Echo Show that might have a historic value, but given are long deprecated and unsupported ideally should not be hierarchically on the same level as the active products (and maybe even little pruned for brevity for today's needs)?

I expected the local nav sidebar be more concise, without long list of abandoned products just next to the current products:

> sidebar

_No response_

Not sure if completely removing the sections, or keeping them shorter and bundled together under one chapter as legacy UAs is better.

To keep the old fragments working, would nesting the headings to a lower level still work for permalinks with anchor ids, but hiding them from the sidebar navigation?

mohammed-ali-osman:
### Description

- Added a deprecation note for historical and compatibility purposes.
- Consolidated Firefox for Fire TV, Echo Show, and Firefox OS (including their device-specific subsections) under the Legacy Firefox Platforms section for clarity and consistency.

Fixes #43642


files/en-us/web/http/reference/headers/user-agent/firefox/index.md
```diff
@@ -136,7 +136,13 @@ Mozilla/5.0 (iPhone; CPU iPhone OS 12_1 like Mac OS X) AppleWebKit/605.1.15 (KHT
 
 Note: this user agent was retrieved from an iPhone XR simulator and may be different on device.
 
-## Firefox for Fire TV
+## Legacy Firefox platforms
+
+> [!NOTE] These platforms are deprecated and no longer supported.
```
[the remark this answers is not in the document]
mohammed-ali-osman: This is interesting, i just followed the guidelines to **how you write**?

> [!NOTE]
> This is how you write a note.
>
> It can have multiple lines.

in the [Writing guidelines](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines) > [Notes example](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#examples)

at this point how to fix this linting problem?

Josh-Cena:
```suggestion
> [!NOTE]
> These platforms are deprecated and no longer supported.
```

Josh-Cena (commented):
These platforms have been dead for more than 5 years. I don't think we need to keep them at all.

mohammed-ali-osman:
Hi @Josh-Cena,

Thanks for taking a look and making the update.

I agree and i thought to removed it all cause these platforms have been deprecated for quite a while. That said, MDN generally preserves historical and compatibility information rather than removing it entirely especially for things like user agent formats that may still appear in logs or legacy systems.

This change keeps the content but clearly marks it as deprecated and moves it under a dedicated Legacy Firefox Platforms section, which reduces its appearance in the sidebar while keeping it available as a reference.

Josh-Cena:
> That said, MDN generally preserves historical and compatibility information rather than removing it entirely

Not true though, we have the policy (for web features, at least) to remove stuff after they become obsolete for 2 years. Of course that dictates primarily what you _send_ to your user, not what you _receive_ from your user, but we can safely assume that these user agents are well outside the support range of "MDN-documented user agents" anyway, so they are irrelevant.

mohammed-ali-osman:
> we have the policy (for web features, at least) to remove stuff after they become obsolete for 2 years.

I agree these platforms are obsolete and shouldn’t be presented as current or recommended.

My concern is that UA strings are not something developers only “use”, but also something they may still encounter in logs or legacy systems. That makes them a bit different from spec-defined features covered by the [removal guidelines](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Experimental_deprecated_obsolete#guidelines_for_removing_content).

That said, I think a more selective approach makes sense and would reduce maintenance overhead:

- Fire TV / Echo Show could likely be removed or reduced, given their limited and short-lived usage
- Firefox OS might still be worth keeping as a minimal, clearly deprecated reference due to its broader historical footprint

hamishwillee:
Thanks @mohammed-ali-osman - I strongly agree with @Josh-Cena that we should delete all these. MDN is supposed to document the evolving web platform. All of this is well past its expiry date in HTTP header docs. 

PS:
- yes, I did read the reasons for keeping above - not sufficient justification for any of it
- I do appreciate that you have thought about this though.

files/en-us/web/http/reference/headers/user-agent/firefox/index.md
mohammed-ali-osman: Thanks @hamishwillee, @Josh-Cena for the clarification that makes sense.

I agree removing these is the right call. I appreciate you both taking the time to review and discuss this.

For reference, this change removes the section from L138 to R211.

hamishwillee (approved):
Thank very much @mohammed-ali-osman (@Josh-Cena ). Much appreciated!

PR #43721, opened 2026-04-09
Mention default receiver behavior for Reflect.get()

jespertheend:
### Description

Mention the default `Reflect.get()` behavior when the optional `receiver` parameter is not provided.

### Motivation

I was looking for this information but couldn't find it on MDN.

### Additional details

https://tc39.es/ecma262/multipage/reflection.html#sec-reflect.get


Josh-Cena (approved):
Nice, thank you

PR #43728, opened 2026-04-10
FF150 Relnote: img sizes attribute supports auto option

hamishwillee:
FF150 adds `HTMLImageElement.sizes` value of `auto` and the corresponding attribute on [``](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#sizes) in https://bugzilla.mozilla.org/show_bug.cgi?id=1819581. 

This updates the release note.

Related docs work can be tracked in #43562


pepelsbey (approved):
Looks good! Thank you 🙂

PR #43744, opened 2026-04-11
Custom Highlight methods accept any AbstractRange

narasimhan-lakshmi:
### Description

Updated the CSS Custom Highlight API documentation to use AbstractRange (instead of Range) across the Highlight interface and related pages.
Fixed cross-references by replacing Range links with AbstractRange in the Highlight docs subtree.

### Motivation

Made the changes according to the CSS Custom Highlight API documentation, making users to use proper methods without hassles.

 Fixes #43704


Josh-Cena (approved):
Very nice work. Thank you!

PR #43750, opened 2026-04-11
Remove 504'ed WebKit Trac link

Buenaventura0:
### Description

The pseudo element ::-webkit-inner-spin-button page property has a reference link below the See also section:

- [Styling Form Controls – WebKit

As part of Apple's restructuring documentation, this link doesn't work anymore and is returning a 504 server error.
I propose removing it from the page.

### Motivation

Clean up the documentation and avoid reference confusion.

### Additional details

(https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::-webkit-inner-spin-button)


Josh-Cena:
Personally I think we should confirm if it's gone for good, and if so, we need to remove 3 other occurrences of `trac.webkit.org` as well.

Buenaventura0:
> Personally I think we should confirm if it's gone for good, and if so, we need to remove 3 other occurrences of `trac.webkit.org` as well.

Hi Josh, I searched the web and the WebKit present documentation for more references, but I couldn't find any yet. Is there someone from Apple here at the MDN we could consult on the availability of this domain?

If we confirm it's gone for good, I can include the other occurrences as well.

Josh-Cena:
Yeah unfortunately I don't know who to ping. I'll ask.

hamishwillee:
@Josh-Cena I don't know who to ask, but I am pretty sure this is gone for good. The interesting stuff migrated to https://github.com/WebKit/WebKit - but this did not (just FYI, you can see what it was on https://web.archive.org/web/20250101203208/https://trac.webkit.org/wiki/Styling%20Form%20Controls).

I think we should delete all those links too. But looks like your owning this, so do as you will :-)

chrisdavidmills:
> > Personally I think we should confirm if it's gone for good, and if so, we need to remove 3 other occurrences of `trac.webkit.org` as well.
> 
> Hi Josh, I searched the web and the WebKit present documentation for more references, but I couldn't find any yet. Is there someone from Apple here at the MDN we could consult on the availability of this domain?
> 
> If we confirm it's gone for good, I can include the other occurrences as well.

Hi @jdatapple and @jensimmons; can you confirm this?

hamishwillee (approved):
These are better links than the originals. Happy to merge without waiting on status of trac. Thanks @Buenaventura0 @Josh-Cena

PR #43751, opened 2026-04-11
Improve replace() try-it example

evgenykolotov:
### Description

Updates the `Try it` example for `String.prototype.replace()` so it more clearly shows the difference between string replacement, case-insensitive regex replacement, and global regex replacement.

### Motivation

The previous example did not clearly demonstrate that when `pattern` is a string, only the first occurrence is replaced. It also did not make the effect of the regex `i` flag very clear.

### Additional details

The new example uses a more natural sentence and shows three distinct cases:
- string replacement, which replaces only the first matching substring
- regex replacement with `i`, which performs a single case-insensitive match
- regex replacement with `gi`, which replaces all matches

Fixes #43720


Josh-Cena (approved):
Thanks

PR #43752, opened 2026-04-12
fix example to better show replace only affects first match

Olabodela:
the previous example only had one "Ruth's" so it didnt really show the behaviour clearly

added another occurrence so its obvious only the first one gets replaced


Josh-Cena:
Duplicate of https://github.com/mdn/content/pull/43751

PR #43759, opened 2026-04-13
FF150 Expr Feature - multiple JS import maps

hamishwillee:
FF150 adds support for [multiple import maps](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/importmap#merging_multiple_import_maps) behind a pref (`dom.multiple_import_maps.enabled`). This adds an experimental feature and release note entry.

It also adds a minor clarification to the JS doc.

Related docs work can be tracked in #43551

### DIFF
--- a/files/en-us/web/html/reference/elements/script/type/importmap/index.md
+++ b/files/en-us/web/html/reference/elements/script/type/importmap/index.md
@@ -166,6 +166,8 @@ For example, the map below defines integrity metadata for the `square.js` module
 
 ### Merging multiple import maps
 
+Supporting browsers can declare one or more import maps anywhere in the document, provided they are defined before any module that depends on them is loaded (earlier specifications allowed only a single declaration that must be made before any module is loaded).
+
 Internally, browsers maintain a single global import map representation. When multiple import maps are included in a document, their contents are merged into the global import map when they are registered.
 
 For example, consider the following two import maps:
--- a/files/en-us/web/html/reference/elements/script/type/importmap/index.md
+++ b/files/en-us/web/html/reference/elements/script/type/importmap/index.md
@@ -166,7 +166,7 @@ For example, the map below defines integrity metadata for the `square.js` module
 
 ### Merging multiple import maps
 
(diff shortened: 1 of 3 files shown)


files/en-us/web/html/reference/elements/script/type/importmap/index.md
```diff
@@ -166,6 +166,8 @@ For example, the map below defines integrity metadata for the `square.js` module
 
 ### Merging multiple import maps
 
+Supporting browsers can declare one or more import maps anywhere in the document, provided they are defined before any module that depends on them is loaded (earlier specifications allowed only a single declaration that must be made before any module is loaded).
```
hamishwillee: The single import map model was that you had to declare all your mappings before loading any module - whether or not the module used any of those mappings. Now we have multiple import maps you can spread them around, before and after modules are loaded.

I _think_ it is worth adding a note that while you can split up your import maps, you still need to make sure that  all definitions needed by a module are provided in a map before that module is loaded.

@Josh-Cena Too obvious? I can drop the commit if you feel strongly about this.

As an aside, the docs talk about registering an import map. Is that the same as declaring in your HTML?

Josh-Cena: Yeah I like this addition. Although instead of "earlier specifications" I would prefer to see actual browser support in the table, if possible.

> the docs talk about registering an import map. Is that the same as declaring in your HTML?

AFAIK, yes

hamishwillee: Thanks, and fair enough. I'll link the compat table, which includes a key for this.

```suggestion
Supporting browsers can declare one or more import maps anywhere in the document, provided they are defined before any module that depends on them is loaded (some [browser versions](#browser_compatibility) allow only a single import map declaration, which must appear before any module is loaded).
```

pepelsbey (approved):
Looks good! Thank you 🙂

Just to double-check: not enabled in Nightly by default?

hamishwillee:
Thanks @pepelsbey - definitely false always in the PR. I checked searchfox at the time and also false there (searchfox now misbehaving so can't reverify).

PR #43775, opened 2026-04-13
docs: change inline style example to avoid confusion

wesleyboar:
### Description

On [Content-Security-Policy: style-src-attr directive > Examples > Violation cases](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-attr#violation_cases), change the style attribute value to something that does not use a keyword.

### Motivation

I intend to avoid possible confusion. A user might think `style-src-attr 'none'` blocks `<div style="display:none">Foo</div>` because it has "none" in the attribute value. (At least one user was confused… me.)

### Additional details

N/A

N/A


files/en-us/web/http/reference/headers/content-security-policy/style-src-attr/index.md
wesleyboar:
```suggestion
document.querySelector("div").style.display = "inline";
```

pepelsbey (approved):
Good catch! Thank you 🙂

PR #43776, opened 2026-04-13
FF150 Relnote: Devtools response tab specific message on redirect

hamishwillee:
FF150 modifies devtools Network panel such that the response tab has a specific message if the response is empty due to a redirect in https://bugzilla.mozilla.org/show_bug.cgi?id=2016679. This adds a release note.

Related docs work can be tracked in #43554


pepelsbey (approved):
Looks good! Thank you 🙂

PR #43781, opened 2026-04-14
Fixed typo in getDay() range - was 1-7 now 0-6

coryannj:
### Description

getDay() range incorrectly mentioned as one-indexed (1-7) instead of zero-indexed (0-6)

### Motivation

Fixing typo/incorrect content

### Additional details
Change suggested based on getDay() docs:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay#:~:text=An%20integer%2C%20between%200%20and%206%2C%20representing%20the%20day%20of%20the%20week%20for%20the%20given%20date%20according%20to%20local%20time%3A%200%20for%20Sunday%2C%201%20for%20Monday%2C%202%20for%20Tuesday%2C%20and%20so%20on

### DIFF
--- a/files/en-us/web/api/indexeddb_api/checking_when_a_deadline_is_due/index.md
+++ b/files/en-us/web/api/indexeddb_api/checking_when_a_deadline_is_due/index.md
@@ -147,7 +147,7 @@ function checkDeadlines() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


hamishwillee (approved):
Thanks very much. Verified with
```js
let current = new Date(); // Start at today
for (let i = 0; i < 7; i++) {
  // Increment the day by 1
  current.setDate(current.getDate() + 1);
  console.log(current.getDay());
}
```
... and congratulations on landing your first MDN PR!

PR #43782, opened 2026-04-14
ReportingAPI: List of reports - draft

### LINKED ISSUE #43688: Reporting: there's no complete list of possible report types (or named reporting endpoints)
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API

Report types [but also others]

I could not find a complete list report types, or at least the current text doesn't tell me if the listed report types are exhaustive. By report types, I mean:

On this page, I expected to find a complete list of all the possible values for a report body's `type` property. Short of that, I expected to be able to piece together the list from the [ReportingObserver: observe() method](https://developer.mozilla.org/en-US/docs/Web/API/ReportingObserver/observe) page and the [Reporting-Endpoints header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints) page.

Likewise, I expected to find a complete list of named reporting endpoints (e.g., `Reporting-Endpoints: crash-reporting="…"`), plus the `default` endpoint.

Without this information, it's very difficult to know what reports one might expect to receive via the `ReportingObserver` API or via `POST` to a `Reporting-Endpoints: default="…"` header URL.

The [Report types section](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API#report_types) says:

> The type of report can be determined from its `type` property, which for the reports above would be `coep`, `deprecation`, `integrity-violation`, `intervention` and `csp-violation`.

It does not say whether "the reports above" is an exhaustive list of report types.

The relevant specs are in a deeply confused state, so it's not easy to determine this from the specifications (see https://github.com/w3c/reporting/issues/216#issuecomment-2967834156 and that issue in general for more information).

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/api/reporting_api`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/reporting_api/index.md
* Last commit: https://github.com/mdn/content/commit/09d8ff096be97b28ea415fc4c68fb1cff0ff8af9
* Document last modified: 2026-04-06T19:43:24.000Z

</details>

hamishwillee:
This is in progress work to provide a list of all reports and how they are accessed for reporting API.
Fixes #43688

This will have the parent grouping. Still working on the pages that it needs before this can go in though. This is all being tested still.

- [ ] Document-Policy violations https://wicg.github.io/document-policy/#configure-reporting and https://chromestatus.com/feature/5756689661820928 and https://github.com/mdn/mdn/issues/394 and https://github.com/WICG/document-policy/blob/main/document-policy-explainer.md
- [ ] Permission policy violations
  - [ ] https://github.com/mdn/content/pull/43783
  - [ ] https://github.com/mdn/browser-compat-data/pull/29500
- [ ] Network errors - to be done - see https://w3c.github.io/network-error-logging/#concept-network-error-reports https://w3c.github.io/network-error-logging/#nel-response-header [Network Error Logging](/en-US/docs/Web/HTTP/Guides/Network_Error_Logging)
- [ ] Crash reports 
  - [ ] #43791
  - [ ] https://github.com/mdn/browser-compat-data/pull/29475
- [ ] coop to be done - ([spec](https://html.spec.whatwg.org/multipage/browsers.html#coop-reporting))
- [ ] csp-next - https://wicg.github.io/csp-next/scripting-policy.html - appears dormant and not implementd

### DIFF
--- a/files/en-us/web/api/reporting_api/index.md
+++ b/files/en-us/web/api/reporting_api/index.md
@@ -75,6 +75,21 @@ The only difference is that server reports are JSON serializations of the object
 
 A list of documented report types and their corresponding report dictionary are given in the [`options.types`](/en-US/docs/Web/API/ReportingObserver/ReportingObserver#types) parameter passed to the `ReportingObserver()` constructor.
 
+##### XXXXX
+
+| Type                           | Report object                                   | Local | Server | Notes                                                                                            |
+| ------------------------------ | ----------------------------------------------- | ----- | ------ | ------------------------------------------------------------------------------------------------ |
+| `coep`                         | {{domxref("COEPViolationReport")}}              | Y     | Y      | {{httpheader("Cross-Origin-Embedder-Policy")}} (COEP) violations                                 |
+| `coop`                         | {{domxref("COOPViolationReport")}}              | Y     | Y      | {{httpheader("Cross-Origin-Opener-Policy")}} (COOP) violations                                   |
+| `crash`                        | {{domxref("CrashReport")}}                      | N     | Y      | Browser crash reports                                                                            |
+| `csp-violation`                | {{domxref("CSPViolationReport")}}               | Y     | Y      | [Content Security Policy (CSP)](/en-US/docs/Web/HTTP/Guides/CSP) violations                      |
+| `deprecation`                  | {{domxref("DeprecationReport")}}                | Y     | Y      | Deprecated features used by the site.                                                            |
+| `document-policy-violation`    | {{domxref("DocumentPolicyViolationReport")}}    | ??    | Y      | {{httpheader("Document-Policy")}} violations                                                     |
+| `integrity-violation`          | {{domxref("IntegrityViolationReport")}}         | Y     | Y      | {{httpheader("Integrity-Policy")}} violations                                                    |
+| `intervention`                 | {{domxref("InterventionReport")}}               | Y     | Y      | Features blocked by the user agent, for example, if an ad significantly impacts page performance |
+| `network-error`                | {{domxref("NetworkErrorReport")}}               | Y     | N      | Network error report                                                                             |
+| `permissions-policy-violation` | {{domxref("PermissionsPolicyViolationReport")}} | Y     | Y      | {{httpheader("Permissions-Policy")}} violations                                                  |
+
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
@hamishwillee, as an FYI, I'm working on the Crash Reporting API docs right now.

One thing I am confused about with the docs for the reporting APIs — the dictionary pages are titled `xReport`, for example, `DeprecationReport`, but the corresponding API entries in the specs are called `xReportBody`, for example, `DeprecationReportBody`. I'm assuming that this is because the `ReportingObserver` returns report objects that the body objects are contained within, but I really don't get where or how this is specified.

Crash reports are not observed by reporting observers; they are just sent to the server endpoints directly when a crash occurs. This being the case, should I still call the corresponding dictionary page `CrashReport` rather than `CrashReportBody`? This stuff is confusing as hell.

I'll get on with it anyway, and produce a draft to look at.

hamishwillee:
>Crash reports are not observed by reporting observers; they are just sent to the server endpoints directly when a crash occurs. This being the case, should I still call the corresponding dictionary page CrashReport rather than CrashReportBody? This stuff is confusing as hell.

@chrisdavidmills This has a long and still messy history. The original Reporting API defined Report and ReportBody interfaces, and implementations were supposed to derive from ReportBody. The spec changed a ciouple of years ago to make Report and ReportBody into dictonaries, and updated some of the derived report bodies to match. This is still all a mess where some have been migrated, some have not, and some don't derive from anything at all and therefore have no spec name. Following the suggestion of BCD we agreed to assume that we should document as though everything had migrated.

So the short version is, we document reports as dictionaries. Following MDN practise we don't document intermediate dictionaries, we just document the top level dictionary. Hence the pattern you see were we document the whole derived report and assume nothing is derived. Make sense?

hamishwillee:
> Crash reports are not observed by reporting observers; they are just sent to the server endpoints directly when a crash occurs. This being the case, should I still call the corresponding dictionary page CrashReport rather than CrashReportBody? This stuff is confusing as hell.

Yes. You have done it as I would. The point of THIS PR is to list all the reports you might get, and also to track the missing ones.

PR #43783, opened 2026-04-14
ReportingAPI: PermissionPolicyViolationReport

hamishwillee:
This adds docs for Permissions-Policy violation reports (`PermissionsPolicyViolationReport`), along with changes I thought necessary to make it work.

This contributes to #43782 (and  #43688)

PR #43791, opened 2026-04-14
Technical review: Document crash reporting api

chrisdavidmills:
### Description

Chrome 145 adds support for the Crash Reporting API; see https://chromestatus.com/feature/6228675846209536.

I have documented this as part of the Reporting API documentation, rather than in its own API docs tree, as I feel that this makes much sense, even though it is specified in a separate spec.


files/en-us/web/api/crashreport/index.md
```diff
@@ -0,0 +1,97 @@
...
+browser-compat: api.ReportingObserver.ReportingObserver.options_parameter.types_property.crash
+---
+
+{{APIRef("Reporting API")}}
+
+The `CrashReport` dictionary of the [Reporting API](/en-US/docs/Web/API/Reporting_API) represents a crash report.
+
+## Instance properties
+
+- `age`
```
hamishwillee, line 17: FYI I don't/haven't documented the age and user agent properties - normally I document all the others and then here say  "same for server version but also has ...".

files/en-us/web/api/crashreport/index.md
```diff
@@ -0,0 +1,97 @@
...
+## Description
+
+The [Crash Reporting API](https://wicg.github.io/crash-reporting) extension to the [Reporting API](/en-US/docs/Web/API/Reporting_API) specifies a mechanism allowing arbitrary information to be recorded and made available in crash reports. This is useful because we can store detailed diagnostic information throughout the lifetime of an application and use the crash reports to debug crashes more effectively.
+
+The information is stored in a special key-value store, manipulated using the document's {{domxref("CrashReportContext")}} object, which is accessed via the {{domxref("Window.crashReport")}} property.
+
+When the browser crashes, the information stored in the key-value store is added to a {{domxref("CrashReport")}} and sent to the default [reporting server endpoint](/en-US/docs/Web/API/Reporting_API#reporting_server_endpoints), if it is defined. The reporting server endpoint and its mapping to a particular URL are set using the {{httpheader("Reporting-Endpoints")}} header.
+
+> [!NOTE]
+> It is not possible to retrieve {{domxref("CrashReport")}}s using a {{domxref("ReportingObserver")}}.
```
hamishwillee: It is worth saying why, and I would put this as a note up in the description - i.e. this only supports server reports.

files/en-us/web/api/crashreport/index.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Examples
+
+### Sending a report to a reporting endpoint
+
+Configuring a web page to send a crash report requires that you set a [reporting server endpoint](/en-US/docs/Web/API/Reporting_API#reporting_server_endpoints) named "default" using the {{httpheader("Reporting-Endpoints")}} header, for example `https://example.com/default`.
```
hamishwillee: Permissions-Policy is similar -https://github.com/mdn/content/pull/43783 - I nearly got this in. Reason I bring tihs up is that it would be good to follow the same pattern here in terms of the link to the new heading (added in that PR) for the default reporting endpoint.

hamishwillee: I'm on other stuff today - this was fly-by. It looks good, but is different than the others so I will need to look at this in detail on Friday.

Can you tell me how you induced a crash for testing? I have my own test code for this and would like to add it https://github.com/hamishwillee/various_webapi_tests/tree/main/ReportingAPI

files/en-us/web/api/crashreport/index.md
chrisdavidmills: Yeah, I know. I did it this way because, in this case, the report type isn't retrieved via reporting observers, so the comparison can't be made.

files/en-us/web/api/crashreport/index.md
chrisdavidmills: Good call; I've updated the note to:

> [!NOTE]
> It is not possible to retrieve {{domxref("CrashReport")}}s using a {{domxref("ReportingObserver")}} — the reports are only generated when the browser crashes, at which point the observer code isn't available to run.

I didn't understand the second part of this sentence — the note is in the description already.

files/en-us/web/api/crashreport/index.md
chrisdavidmills: > Permissions-Policy is similar -#43783 - I nearly got this in. Reason I bring tihs up is that it would be good to follow the same pattern here in terms of the link to the new heading (added in that PR) for the default reporting endpoint.

Makes sense. I've updated the links in the next commit.

chrisdavidmills: > Can you tell me how you induced a crash for testing? I have my own test code for this and would like to add it https://github.com/hamishwillee/various_webapi_tests/tree/main/ReportingAPI

I didn't actually test it; I just trusted the explainer. I intended to ask the engineers for a test case rather than write my own, to save a bit of time.

files/en-us/web/api/crashreportcontext/delete/index.md
```diff
@@ -0,0 +1,60 @@
...
+  - : Thrown if:
+    - The calling document is not fully active.
+    - The crash report key-value store is not yet initialized via an {{domxref("CrashReportContext.initialize", "initialize()")}} call.
+
+## Examples
+
+### Basic usage
+
+```js
+window.crashReport.set(1024).then(() => {
```
jcscottiii: nit
```suggestion
window.crashReport.initialize(1024).then(() => {
```

files/en-us/web/api/crashreportcontext/set/index.md
```diff
@@ -0,0 +1,64 @@
...
+    - The crash report key-value store is not yet initialized via an {{domxref("CrashReportContext.initialize", "initialize()")}} call.
+- `NotAllowedError` {{domxref("DOMException")}}
+  - : The size of the serialized key-value pair is greater than the [`length`](/en-US/docs/Web/API/CrashReportContext/initialize#length) value set when the store was first initialized.
+
+## Examples
+
+### Basic usage
+
+```js
+window.crashReport.set(1024).then(() => {
```
jcscottiii:
```suggestion
window.crashReport.initialize(1024).then(() => {
```

files/en-us/web/api/crashreportcontext/initialize/index.md
```diff
@@ -0,0 +1,57 @@
...
+## Syntax
+
+```js-nolint
+initialize(length)
+```
+
+### Parameters
+
+- `length`
+  - : A number indicating the maximum number of bytes that can be stored in the key-value store by each individual `set()` call. The maximum allowable value is `5 * 1024 * 1024` (5MB), which is also the maximum amount of crash report memory allowed for one origin.
```
jcscottiii: The limit is actually 64KB

Source: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/frame/frame.mojom;l=261;drc=ad1df7654e8a72f3267dd10890f157ea639b989d

files/en-us/web/api/crashreportcontext/initialize/index.md
```diff
@@ -0,0 +1,57 @@
...
+A {{jsxref("Promise")}} that fulfills with {{jsxref("undefined")}}.
+
+### Exceptions
+
+- `InvalidStateError` {{domxref("DOMException")}}
+  - : Thrown if:
+    - The calling document is not fully active.
+    - A crash report key-value store was already initialized by a previous `initialize()` call.
+- `NotAllowedError` {{domxref("DOMException")}}
+  - : `length` is more than `5 * 1024 * 1024`.
```
jcscottiii: See note about updated limit

jcscottiii (commented):
Small nits but LGTM overall

files/en-us/web/api/crashreport/index.md
hamishwillee: The note update is good. I'm suggesting this is important enough to move it below the introductory sentence at the top of the page, rather than here. Mostly because this behaviour is a real outlier for the interface.

files/en-us/web/api/crashreport/index.md
```diff
@@ -0,0 +1,97 @@
...
+{{APIRef("Reporting API")}}
+
+The `CrashReport` dictionary of the [Reporting API](/en-US/docs/Web/API/Reporting_API) represents a crash report.
+
+## Instance properties
+
+- `age`
+  - : The age of the report in milliseconds.
+- `type`
+  - : The string `"crash"` indicating that this is a deprecation report.
```
hamishwillee:
```suggestion
  - : The string `"crash"` indicating that this is a crash report.
```

files/en-us/web/api/crashreport/index.md
hamishwillee: Works for me. Shouldn't they be alphabetic ordering? I'm "OK "with this being an exception since this is the order it is serialized, but my preference is to do it the same as all other APIs

files/en-us/web/api/crashreport/index.md
```diff
@@ -0,0 +1,97 @@
...
+        - `visible`
+          - : The document content is at least partially visible.
+        - `hidden`
+          - : The document content is completely hidden.
+    - `crash_report_api` {{experimental_inline}} {{optional_inline}}
+      - : An object containing the key-value pairs set via the {{domxref("CrashReportContext.set()")}} method, if any.
+
+## Description
+
+The [Crash Reporting API](https://wicg.github.io/crash-reporting) extension to the [Reporting API](/en-US/docs/Web/API/Reporting_API) specifies a mechanism allowing arbitrary information to be recorded and made available in crash reports. This is useful because we can store detailed diagnostic information throughout the lifetime of an application and use the crash reports to debug crashes more effectively.
```
hamishwillee: "The [Crash Reporting API](https://wicg.github.io/crash-reporting) extension to the" is unusual. MDN doesn't recognise extensions like this or link direct to spec - that would be added to browser compat or a spec url. For this I would do


```suggestion
Crash reports containing arbitrary information can be sent to a server endpoint using the [Reporting API](/en-US/docs/Web/API/Reporting_API) .
This is useful because we can store detailed diagnostic information throughout the lifetime of an application and use the reports to debug crashes more effectively.
```

files/en-us/web/api/crashreport/index.md
```diff
@@ -0,0 +1,97 @@
...
+
+The information is stored in a special key-value store, manipulated using the document's {{domxref("CrashReportContext")}} object, which is accessed via the {{domxref("Window.crashReport")}} property.
```
hamishwillee:
```suggestion
The diagnostic information is stored in a special key-value store that can be manipulated using the document's {{domxref("CrashReportContext")}} object. 
This is accessed via the {{domxref("Window.crashReport")}} property.
```

files/en-us/web/api/window/crashreport/index.md
```diff
@@ -0,0 +1,31 @@
...
+title: "Window: crashReport property"
+short-title: crashReport
+slug: Web/API/Window/crashReport
+page-type: web-api-instance-property
+browser-compat: api.Window.crashReport
+---
+
+{{securecontext_header}}{{APIRef("Reporting API")}}
+
+The **`crashReport`** read-only property of the {{domxref("Window")}} interface returns a {{domxref("CrashReportContext")}} object that enables arbitrary data to be recorded for the current top-level browsing context, which is then added to a {{domxref("CrashReport")}} and sent to a reporting endpoint when a browser crash occurs.
```
hamishwillee: Consider making this into shorter sentences. 


```suggestion
The **`crashReport`** read-only property of the {{domxref("Window")}} interface returns a {{domxref("CrashReportContext")}} object that enables arbitrary data to be recorded for the current top-level browsing context.
The data is then included in {{domxref("CrashReport")}} objects that are sent to a reporting endpoint when a browser crash occurs.
```

files/en-us/web/api/crashreportcontext/index.md
```diff
@@ -0,0 +1,83 @@
...
+- {{domxref("CrashReportContext.delete()")}}
+  - : Deletes a previously-stored key-value pair.
+- {{domxref("CrashReportContext.initialize()")}}
+  - : Initializes a portion of memory to be used for storing crash report data specified by {{domxref("CrashReportContext.set", "set()")}}. This must be called before any other method is invoked on the object.
+- {{domxref("CrashReportContext.set()")}}
+  - : Stores a key-value pair in the memory initialized by {{domxref("CrashReportContext.initialize", "initialize()")}}.
+
+## Description
+
+The [Crash Reporting API](https://wicg.github.io/crash-reporting) extension to the [Reporting API](/en-US/docs/Web/API/Reporting_API) specifies a mechanism allowing arbitrary information to be recorded and made available in crash reports. This is useful because we can store detailed diagnostic information throughout the lifetime of an application and use the crash reports to debug crashes more effectively.
```
hamishwillee: Same comment as before - we don't reference specs in page like this.

files/en-us/web/api/crashreportcontext/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+The information is stored in a special key-value store, which is similar to [Web Storage](/en-US/docs/Web/API/Web_Storage_API) except that its scope is restricted to the current top-level origin, and it doesn't have a getter method available. It allows document-specific state information to be recorded and deleted. The top-level document will record diagnostic information relating to it and any embedded documents, and send any related crash reports.
+
+To use this API, a document first needs to call {{domxref("CrashReportContext.initialize", "window.crashReport.initialize()")}}, which takes as an argument a number specifying the maximum number of bytes that can be recorded in the key-value store by each individual {{domxref("CrashReportContext.set", "window.crashReport.set()")}} call. Values are then recorded using `set()`, and deleted using {{domxref("CrashReportContext.delete", "window.crashReport.delete()")}}.
```
hamishwillee, line 30: FWIW I was going to say "what happens if you run out of memory when calling set", but this is covered in set()

files/en-us/web/api/reporting_api/index.md
hamishwillee, line 114: Add the `CrashReport` here? I think so.

files/en-us/web/api/crashreport/index.md
hamishwillee: I can't trigger OOM, but it is possible to trigger an unresponsive message on Chrome. The following is what I got out of [my crash here](https://github.com/hamishwillee/various_webapi_tests/tree/main/ReportingAPI). If you want to try it you need to install Caddy and run it from ReportingAPI with `npm run start:crash`

```
[0] [
[0]   {
[0]     age: 5,
[0]     body: {
[0]       crash_report_api: { 'test-scenario': 'unresponsive-trigger' },
[0]       is_top_level: true,
[0]       reason: 'unresponsive',
[0]       stack: 'Website owner has not opted in for JS call stacks in crash reports.',
[0]       visibility_state: 'visible'
[0]     },
[0]     type: 'crash',
[0]     url: 'https://localhost:8443/',
[0]     user_agent: 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36'
[0]   }
[0] ]
```

files/en-us/web/api/crashreport/index.md
chrisdavidmills: I've updated the order to put the top-level props alphabetically, except for `body`, which is at the end. Inside `body`, I've put the sub-props alphabetically. I think it's easier to read with `body` at the end like this.

files/en-us/web/api/crashreport/index.md
chrisdavidmills: Ah, gotcha! Yup, that makes sense; I've moved the note up top as suggested.

files/en-us/web/api/crashreportcontext/initialize/index.md
chrisdavidmills: Updated. Thanks for the info, @jcscottiii.

files/en-us/web/api/crashreportcontext/initialize/index.md
chrisdavidmills: Updated here too; thanks!

files/en-us/web/api/crashreportcontext/index.md
chrisdavidmills: Updated, as with the other article.

files/en-us/web/api/reporting_api/index.md
chrisdavidmills: done

chrisdavidmills:
@jcscottiii thanks for the review! I've fixed the issues you brought up. Let me know if you spot anything else.

files/en-us/web/api/crashreport/index.md
hamishwillee: In cases like this I tend to just follow the alphabetic rule. I do agree with you though, so not requiring a fix. I won't be surprised if tooling or someone else update it in future though.

files/en-us/web/api/crashreport/index.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Description
+
+Crash reports containing arbitrary information can be sent to a server endpoint using the [Reporting API](/en-US/docs/Web/API/Reporting_API) .
+This is useful because we can store detailed diagnostic information throughout the lifetime of an application and use the reports to debug crashes more effectively.
+
+The diagnostic information is stored in a special key-value store that can be manipulated using the document's {{domxref("CrashReportContext")}} object.
+This is accessed via the {{domxref("Window.crashReport")}} property.
+
+When the browser crashes, the information stored in the key-value store is added to a `CrashReport` and sent to the [default reporting server endpoint](/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#default_reporting_endpoint), if it is defined. The reporting server endpoint and its mapping to a particular URL are set using the {{httpheader("Reporting-Endpoints")}} header.
```
hamishwillee: @chrisdavidmills Last thing, just discovered that this needs to be updated. See spec: https://wicg.github.io/crash-reporting/#delivery-priority

Basically there are two possible endpoints `crash-reporting` or `default`, used in that order. We need to documnt this (i.e.  `crash-reporting` ). 

Also, below in the first example, we should show this being set.

```http
Reporting-Endpoints: crash-reporting="https://example.com/reports"
```

Note /reports endpoint might be better, if it can be used.

chrisdavidmills: Added!

hamishwillee (approved):
LGTM! Thanks very much @chrisdavidmills

PR #43793, opened 2026-04-14
[Editorial review] BiDi - Add page for log module event

dipikabh:
### Description

This PR adds a page for the `log.entryAdded` event, the only one event in the `log` module (it has no commands).

Also:
- Updated the `log.entryAdded` example in the modules landing page.
- Added a list of the event notification fields to the overview [Events](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#events) section.
This now makes it parallel to the command message fields in the [Commands](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#commands) section.

### Spec link

https://w3c.github.io/webdriver-bidi/#module-log

### For editorial review

Since this is the first page of its kind (a BiDi event), I've added the `webdriver-event` page-type to the `front-matter-config.json` file.

### Related issue

Doc issue: mdn/mdn#339

### DIFF
--- a/files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
+++ b/files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
@@ -48,7 +48,7 @@ All log entry objects include the following fields:
 In addition to the [common fields](#common-fields), log entry objects with `"type": "console"` also include:
 
 - `args`
-  - : An array of objects that represent the arguments passed to the console method. Each object has a `type` field and optional `value`, `handle`, and `internalId` fields.
+  - : An array of objects that represent the arguments passed to the console method. Each object has a `type` field (such as `"string"`, `"number"`, `"boolean"`, or `"array"`) and optional `value`, `handle`, and `internalId` fields.
 - `method`
   - : A string that contains the name of the console method that was called (for example, `"log"`, `"error"`, `"assert"`, `"debug"`, `"trace"`, `"warn"`).
 
@@ -56,7 +56,7 @@ In addition to the [common fields](#common-fields), log entry objects with `"typ
 
 ### Receiving an event for a console log
 
-With a [WebDriver BiDi connection](/en-US/docs/Web/WebDriver/How_to/Create_BiDi_connection) and a [subscription](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/subscribe) to `log.entryAdded` active, the browser sends a `log.entryAdded` event when a script evaluates `console.log("hello", [1, 2, 3])`:
+With a [WebDriver BiDi connection](/en-US/docs/Web/WebDriver/How_to/Create_BiDi_connection) and a [subscription](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/subscribe) to `log.entryAdded` active, the browser sends a `log.entryAdded` event when a script evaluates `console.log("hello", [1, true, "foo"])`:
 
 ```json
 {
@@ -78,13 +78,13 @@ With a [WebDriver BiDi connection](/en-US/docs/Web/WebDriver/How_to/Create_BiDi_
         "type": "array",
         "value": [
           { "type": "number", "value": 1 },
-          { "type": "number", "value": 2 },
-          { "type": "number", "value": 3 }
+          { "type": "boolean", "value": true },
+          { "type": "string", "value": "foo" }
         ]
       }
     ],
     "level": "info",
-    "text": "hello 1,2,3",
+    "text": "hello 1,true,foo",
     "timestamp": 1712345678901,
     "stackTrace": {
       "callFrames": [
--- a/files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
+++ b/files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
@@ -11,7 +11,7 @@ The `log.entryAdded` [event](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#ev
 
 ## Event data
 
-The `params` field in the event notification is a log entry object. Based on the source of the log, the log entry object has different types: `"console"`, `"javascript"`, or any other string. Each type may provide additional fields specific to that source.
+The `params` field in the event notification is a log entry object. Based on the source of the log, the log entry object has different types: `"console"` or `"javascript"`. Each type may provide additional fields specific to that source.
 
 ### Common fields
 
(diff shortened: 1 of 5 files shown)


caugner (approved):
Front matter config change LGTM.

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
```diff
@@ -0,0 +1,165 @@
...
+page-type: webdriver-event
+browser-compat: webdriver.bidi.log.entryAdded_event
+sidebar: webdriver
+---
+
+The `log.entryAdded` [event](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules#events) of the [`log`](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/log) module fires when a new log entry is created in the browser, from either a console API call or an unhandled JavaScript error.
+
+## Event data
+
+The `params` field in the event notification is a log entry object. Based on the source of the log, the log entry object has different types: `"console"`, `"javascript"`, or any other string. Each type may provide additional fields specific to that source.
```
whimboo: At the moment we do not have a generic error type anywhere. So maybe we extend it later when it's clear which other types are actually allowed?

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
```diff
@@ -0,0 +1,165 @@
...
+    - `"console"`: Indicates that the log entry was generated from a call to a console API method (for example, {{domxref("console/log_static", "console.log()")}}, {{domxref("console/warn_static", "console.warn()")}}). Log entry objects of this type include [additional fields](#console-entry-fields).
+    - `"javascript"`: Indicates that the log entry was generated from an unhandled JavaScript error.
+    - Any other string: Indicates an implementation-defined log source.
+
+### `"console"` log entry fields
+
+In addition to the [common fields](#common-fields), log entry objects with `"type": "console"` also include:
+
+- `args`
+  - : An array of objects that represent the arguments passed to the console method. Each object has a `type` field and a `value` field.
```
whimboo: Not every remote value actually has a `value` field, so we should make it optional. But every such remote value has as well an optional `handle` and `internalId`, which can be used later on to reference the same object in following commands.

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
```diff
@@ -0,0 +1,165 @@
...
+- `method`
+  - : A string that contains the name of the console method that was called (for example, `"log"`, `"error"`, `"assert"`, `"debug"`, `"trace"`, `"warn"`).
+
+## Examples
+
+### Receiving an event for a console log
+
+With a [WebDriver BiDi connection](/en-US/docs/Web/WebDriver/How_to/Create_BiDi_connection) and a [subscription](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/session/subscribe) to `log.entryAdded` active, the browser sends a `log.entryAdded` event when a script evaluates `console.log("hello", "world")`:
```
whimboo: It might be good to maybe use an object or array as second argument to have a more complex scenario here.

whimboo (approved):
Thank you @dipikabh. Overall this looks good. Just some smaller additions from my side.

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
```diff
@@ -0,0 +1,165 @@
...
+    - `realm`
+      - : A string that contains the ID of the realm.
+    - `context` {{optional_inline}}
+      - : A string that contains the ID of the context in which the log entry was created.
+    - `userContext` {{optional_inline}}
+      - : A string that contains the ID of the user context in which the script-related event occurred.
+- `stackTrace` {{optional_inline}}
+  - : An object with a `callFrames` array that represents the [JavaScript stack](/en-US/docs/Web/WebDriver/Reference/BiDi/Modules/script/stackTrace) at the point the entry was created. Each item in the array is a stack frame with the following fields: `columnNumber`, `functionName`, `lineNumber`, and `url`.
+- `text`
+  - : A string that contains the log message, or `null` if not available.
```
juliandescottes: Some thoughts about the `text` property.

For console API calls, `text` will be generated by concatenating all the stringified arguments passed to the console method. Except for primitives, the actual way an argument is stringified is up to the browser. 

For errors etc... the `text` is usually the error message, but is not standardized. 

Not sure if we have to go into the details of all the nuances here, but I would at least call out that this property should not be used to expect/assert anything, because it's likely to change depending on the browser.

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
```diff
@@ -0,0 +1,165 @@
...
+          "functionName": "",
+          "lineNumber": 3,
+          "url": "https://example.com/app.js"
+        },
+        {
+          "columnNumber": 18,
+          "functionName": "",
+          "lineNumber": 3,
+          "url": "https://example.com/app.js"
+        }
```
juliandescottes, line 171: We should actually have a stacktrace in most cases, but this was missing in Firefox until @lutien partially implemented it in https://bugzilla.mozilla.org/show_bug.cgi?id=1941813.

Which means you should also have a stacktrace property in the two first examples (`log` and `warn`). If you use Nightly to retrieve the payloads, you will however only have one frame in the stacktrace, whereas Chrome should consistently emit the whole stacktrace.

The spec still mentions that stacktrace can be null, but I don't think it's the case in practice for console log entries or javascript errors. (but we probably want to keep the field as potentially null if we extend this to other messages where stacktraces make no sense)

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
dipikabh: Sure, dropping "or any other string"

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
dipikabh: Updated this to:

>   - : An array of objects that represent the arguments passed to the console method. Each object has a `type` field and optional `value`, `handle`, and `internalId` fields.

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
dipikabh: Like this `console.log("hello", [1, 2, 3])`?

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
dipikabh: How about this:

>   - : A string that contains the log message or `null` if not available. For console entries, it is the concatenation of all stringified arguments joined by spaces, and for JavaScript errors, it is generally the error message.
    The exact format is browser-dependent, so don't rely on this value for assertions in tests.

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
dipikabh: Done

dipikabh:
Thanks @whimboo and @juliandescottes for the review. PTAL to check that the updates are good.

I'll also submit this for an editorial review now.

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
whimboo: Like that, yes. Or feel free to as well use different member types of the array: `[1, true, "foo"]`.

pepelsbey (approved):
Looks good! Thank you 🙂

files/en-us/web/webdriver/reference/bidi/modules/log/entryadded/index.md
dipikabh: Got it. Done in https://github.com/mdn/content/pull/43793/commits/a00b8371671bb97c0b2043b9bba0d849f87e3540

dipikabh:
Thanks, Vadim!

PR #43801, opened 2026-04-15
fix typo on `RegExp.escape` page

gameroman:


Josh-Cena (approved):
Nice catch, thank you

PR #43805, opened 2026-04-16
Use borderify name for borderify example extension

MeAmAnUsername:
### Description

beastify@mozilla.org seems to be a copy-paste from the second extension at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Your_second_WebExtension

Feel free to close if it intentionally uses the same ID as that one.

I did not test this tutorial, either before or after the changes in this PR. The explanation says this ID is used for publishing and the tutorial does not include publishing, so I assume this ID is unused for this tutorial.

### Motivation

As a reader it is unclear why this ID should be beastify@mozilla.org

### Additional details

N.A.

N.A.


rebloor (approved):
Thanks @MeAmAnUsername for spotting and fixing that. As you can see, I also noticed another issue: I hadn't updated the manifest version.

